PR #5, opened 2018-08-20
Add a simple Dockerfile

ironcladlou:


ironcladlou:
Follows https://github.com/openshift/machine-config-operator/blob/master/Dockerfile.machine-config-operator

PR #9, opened 2018-08-21
Add product image build support files

ironcladlou:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,14 @@ all: build
 
 APP_NAME=cluster-ingress-operator
 BIN=cluster-ingress-operator
-MAIN_PKG=github.com/openshift/$(APP_NAME)/cmd/cluster-ingress-operator
+MAIN_PKG=github.com/openshift/$(APP_NAME)/cmd/$(APP_NAME)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
+all: build
+
+APP_NAME=cluster-ingress-operator
+BIN=cluster-ingress-operator
+MAIN_PKG=github.com/openshift/$(APP_NAME)/cmd/cluster-ingress-operator
```
Miciah: Should that be `MAIN_PKG=github.com/openshift/$(APP_NAME)/cmd/$(APP_NAME)`? Or is it useful to parameterize `APP_NAME` at all?

Makefile
```diff
@@ -0,0 +1,16 @@
...
+ENVVAR=GOOS=linux GOARCH=amd64 CGO_ENABLED=0
+GOOS=linux
+GO_BUILD_RECIPE=GOOS=$(GOOS) go build -o $(BIN) $(MAIN_PKG)
+
+build: $(BIN)
+
+$(BIN): $(SRC)
```
Miciah: Where is `SRC` defined?

PR #15, opened 2018-08-22
Add clean option to Makefile

imcsk8:
Add output dir and clean option to Makefile.
Also modified the Makefile to be more flexible by moving
commands to the build section.


imcsk8:
PTAL @ironcladlou @pravisankar  @Miciah @openshift/sig-network-edge

Miciah:
You'll need to update the path in [cp command in the Dockerfile](https://github.com/openshift/cluster-ingress-operator/blob/593c2eed1d76b375ca570bfe467f52e612dca6a7/images/cluster-ingress-operator/Dockerfile#L12) too.

Two questions, probably for @ironcladlou:

1. Where did `GO_BUILD_RECIPE` come from? Is it used by CI or OSBS or prescribed by any convention?
2. Should we add a `make clean` to the Dockerfile?

ironcladlou:
@Miciah 
>Where did GO_BUILD_RECIPE come from? Is it used by CI or OSBS or prescribed by any convention?

Copypasta from the cluster-monitoring-operator Makefile. The only requirement from the CI/build system perspective is a Dockerfile.

>Should we add a make clean to the Dockerfile?

Refactor at will. We essentially have total control over this.

Dockerfile
```diff
@@ -0,0 +1 @@
+images/cluster-ingress-operator/Dockerfile
```
ironcladlou: Would prefer to avoid symlinks... because other out-of-tree configurations need to reference our Dockerfiles, I'd rather encourage pointing to physical files since symlinks can be problematic depending on the context.

Makefile
```diff
@@ -1,15 +1,17 @@
 all: build
 
 APP_NAME=cluster-ingress-operator
-BIN=cluster-ingress-operator
+OS_OUTPUT_BINPATH=_output/bin
```
ironcladlou: I don't see the need for an output directory... the only product of the build is a single binary which is already ignored by git

Makefile
```diff
@@ -1,15 +1,17 @@
...
 
-build: $(BIN)
+build: 
+	mkdir -p $(OS_OUTPUT_BINPATH)
+	GOOS=$(GOOS) go build -o $(BIN) $(MAIN_PKG)
 
-$(BIN):
-	$(GO_BUILD_RECIPE)
+clean:
+	rm -rf $(OS_OUTPUT_BINPATH)
```
ironcladlou: I think `rm cluster-ingress-operator` would  have been enough... `rm -rf $path` always makes me nervous on a local system

Makefile
ironcladlou: If you want the binary on PATH a more conventional approach would be to use `go install`

Dockerfile
imcsk8: the problem is when making the copy of the source to the image filesystem while building since dockerfiles don't allow access to higher level directories. It might then be better to create hack/build-image.sh script

Makefile
imcsk8: I think that it's a good practice to store the binaries in specific directories to avoid future confusions

hack/build-image.sh
```diff
@@ -0,0 +1,19 @@
...
+# CIO repository.
+
+export TAG=$(git describe --abbrev=0 2>&1)
+
+if [[ ${PWD} = *"hack"* ]]; then
+    cd ..
+fi
+
+echo "Building Cluster Ingress Operator Image"
+cp images/cluster-ingress-operator/Dockerfile .
```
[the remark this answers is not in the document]
imcsk8: yes, the Dockerfile assumes that the current directory is the CIO source when it copies it to the image, so the Dockerfile should be there.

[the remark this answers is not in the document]
Miciah: The Dockerfile doesn't need to be in the build context:
```bash
docker build -f ./images/cluster-ingress-operator/Dockerfile .
```

hack/build-image.sh
```diff
@@ -0,0 +1,19 @@
...
+
+export TAG=$(git describe --abbrev=0 2>&1)
+
+if [[ ${PWD} = *"hack"* ]]; then
+    cd ..
+fi
+
+echo "Building Cluster Ingress Operator Image"
+cp images/cluster-ingress-operator/Dockerfile . 
+docker build  -t openshift/cluster-ingress-operator:latest .
```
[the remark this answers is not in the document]
imcsk8: The only reason i can think of is that you have to use sudo to use it:

sudo buildah bud -f Dockerfile -t test-cio
and push it
sudo buildah push test-cio docker-daemon:test-cio:latest

images/cluster-ingress-operator/Dockerfile
```diff
@@ -2,14 +2,15 @@ FROM centos:7
...
 RUN mkdir $GOPATH
 
-COPY . $GOPATH/src/github.com/openshift/cluster-ingress-operator
+COPY . $GOPATH/$CIO_PATH
 
-RUN cd $GOPATH/src/github.com/openshift/cluster-ingress-operator \
+RUN cd $GOPATH/$CIO_PATH \
     && make build \
-    && cp $GOPATH/src/github.com/openshift/cluster-ingress-operator/cluster-ingress-operator /usr/bin/
+    && cp $GOPATH/$CIO_PATH/_output/bin/cluster-ingress-operator /usr/bin/
```
ironcladlou: `_output/bin` isn't correct, seems like this Dockerfile should fail to build in its current state

imcsk8: oh yeah i forgot to modify this after the rebase

PR #16, opened 2018-08-24
Deploy routers to a non-operator namespace

ironcladlou:
Deploy routers to a non-operator namespace

* Deploy routers to a new namespace separate from the operator itself to
facilitate clean teardown.
* Refactor manifest handling to use YAML files converted to bindata to enable reuse of assets.
* Configure RBAC for routers.


pkg/stub/handler.go
```diff
@@ -2,50 +2,97 @@ package stub
...
+
+	service, err := h.manifestFactory.RouterServiceCloud(ci)
+	if err != nil {
+		return fmt.Errorf("couldn't build service: %v", err)
+	}
+	err = sdk.Create(service)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create service: %v", err)
+	}
+
```
imcsk8: Maybe later this could be compacted in a sdkCreate function

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
...
+    metadata:
+      labels:
+        app: router
+    spec:
+      serviceAccountName: router
+      nodeSelector:
+        "node-role.kubernetes.io/infra": "true"
+      containers:
+        - name: router
+          image: "docker.io/openshift/origin-haproxy-router:v3.11.0"
```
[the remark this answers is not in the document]
ironcladlou: This and many other fields in the static YAML will be replaced at runtime in the factory code based on operator configuration (I'm not adding the image config with this PR though)

pkg/stub/handler.go
ironcladlou: Yeah, lots of boilerplate here

ironcladlou:
@openshift/sig-network-edge this is tested on GCP and ready for review

ironcladlou:
I want to do a little refactoring to the namespace handling and also make a YAML for namespace...

/hold

assets/router/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: openshift-cluster-ingress-router
+  annotations:
+    openshift.io/node-selector: ""
```
ironcladlou: Lack of this annotation was the reason my DaemonSet (which uses a node selector) wasn't scaling up after the isolated namespace refactor... lesson learned

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
+
+	s.ObjectMeta = metav1.ObjectMeta{
+		Name:      name,
+		Namespace: f.routerNamespace,
+		Labels: map[string]string{
+			"app": "router",
```
Miciah: Maybe it doesn't strictly matter for the service's label (as opposed to its selector), but shouldn't this be `"app": name`?

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
...
+    matchLabels:
+      app: router
+  template:
+    metadata:
+      labels:
+        app: router
+    spec:
+      serviceAccountName: router
+      nodeSelector:
+        "node-role.kubernetes.io/infra": "true"
```
Miciah: Presumably the factory may override this in the future depending on `ClusterIngress.Spec.NodePlacement`.

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: router
+  labels:
+    app: router
```
Miciah: Was it a conscious decision to use `app: router` where the current router uses `router: router`?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
+
+	ds.ObjectMeta = metav1.ObjectMeta{
+		Name:      name,
+		Namespace: f.routerNamespace,
+		Labels: map[string]string{
+			"app": "router",
```
Miciah: Is this deliberately hard-coded as `"app": "router"` rather than `"app": name`?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.DaemonSet, error) {
+	ds, err := NewDaemonSet(MustAssetReader(RouterDaemonSet))
+	if err != nil {
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
+
+	ds.ObjectMeta = metav1.ObjectMeta{
```
Miciah: I'd rather override individual fields rather than `ds.ObjectMeta` in its entirety.

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
...
+            protocol: TCP
+          - name: https
+            containerPort: 443
+            protocol: TCP
+          - name: stats
+            containerPort: 1936
+            protocol: TCP
+          env:
+          - name: STATS_PORT
+            value: "1936"
```
Miciah: This list will grow, right?

assets/router/daemonset.yaml
Miciah: The router deliberately sets `imagePullPolicy: IfNotPresent` (https://github.com/openshift/origin/blob/b054a7bf5b11c760e5dd0f870fc852cc9b8b748b/pkg/oc/cli/admin/router/router.go#L905, https://github.com/openshift/origin/pull/1779#issuecomment-96407925), so I was wondering whether we should do the same.  It looks like `imagePullPolicy: IfNotPresent` is the default if the image tag is not "latest" (https://github.com/openshift/origin/blob/bce7b76872bbb233087ec0f9641c1f84e5115db0/vendor/k8s.io/kubernetes/pkg/apis/core/v1/defaults.go#L85-L86), so the default should be fine if we don't want to be explicit.

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,44 @@
...
+          livenessProbe:
+            initialDelaySeconds: 10
+            httpGet:
+              path: /healthz
+              port: 1936
+          readinessProbe:
+            initialDelaySeconds: 10
+            httpGet:
+              path: /healthz
```
Miciah: Why isn't this using `healthz/ready` as the router currently does? https://github.com/openshift/origin/blob/b054a7bf5b11c760e5dd0f870fc852cc9b8b748b/pkg/oc/cli/admin/router/router.go#L559

(Although it probably makes even more sense to use `/healthz/ready`.)

assets/router/daemonset.yaml
```diff
@@ -0,0 +1,45 @@
...
+              port: 1936
```
Miciah: Are we planning to add `resources` at some point?

assets/router/namespace.yaml
Miciah: I thought you were setting this explicitly in the Go code before.  Did that not work for some reason, or do you mean the Go code that did this got lost in the isolated namespace refactor?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
+
+	ds.Name = name
+
+	ds.Labels = map[string]string{
+		"app": "router",
```
Miciah: Should this be `"app": name`?

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
+
+	s.Name = name
+
+	s.Labels = map[string]string{
+		"app": "router",
```
Miciah: Should this be `"app": name`? Maybe it doesn't really matter for the service's label, as opposed to its selector, but it looks a little strange, like possibly an oversight.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,169 @@
...
+
+	ds.Name = name
+
+	ds.Labels = map[string]string{
+		"app": "router",
+	}
+
+	ds.Spec.Selector = &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app": name,
```
Miciah: Was it a deliberate decision to use `"app": name` instead of `"router": name` as the router current does?

assets/router/namespace.yaml
ironcladlou: Previously I was testing everything in the `default` namespace, which already had the annotation

assets/router/daemonset.yaml
ironcladlou: I think that makes sense

assets/router/daemonset.yaml
ironcladlou: Since it's the way we configure the routers, I assume so

assets/router/daemonset.yaml
ironcladlou: Updated to set an explicit `imagePullPolicy` just because it's clearer

pkg/manifests/manifests.go
ironcladlou: Similar discussion with the daemonset... might make sense to add a label to each that lets us establish they're "owned by" a given ClusterIngress via the name since this namespace could have multiple sets of resources per ClusterIngress.

Worth doing in a followup.

pkg/manifests/manifests.go
ironcladlou: Not really; I think you raised good points throughout about labels generally and we should do a followup to refactor them all to ensure we can deal with ClusterIngress->(Resource Set) matching and such.

PR #18, opened 2018-09-04
Support multiple routers

ironcladlou:
Add new fields to the ClusterIngress type based on the API design. Add just enough configuration and labeling to support multiple routers.


pkg/stub/handler.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/sirupsen/logrus"
+
```
imcsk8: is there a reason not to use "github.com/golang/glog"?

ironcladlou: I was only using logrus because the SDK uses it; no objection to switching to glog for k8s/origin consistency (although for the record I don't like glog at all). I can switch in a followup if everyone agrees on glog.

ironcladlou:
Latest commit adds enough functionality that independent router clusters work, e.g.:

```yaml
apiVersion: ingress.openshift.io/v1alpha1
kind: ClusterIngress
metadata:
  name: default
spec:
  ingressDomain: apps.dmace.origin-gce.dev.openshift.com
  routeSelector:
    matchExpressions:
    - {key: environment, operator: NotIn, values: [internal]}
  highAvailability:
    type: Cloud
```

```yaml
apiVersion: ingress.openshift.io/v1alpha1
kind: ClusterIngress
metadata:
  name: internal
spec:
  ingressDomain: apps.dmace-internal.origin-gce.dev.openshift.com
  routeSelector:
    matchLabels:
      environment: internal
  highAvailability:
    type: Cloud

```

Tested this on GCP. Will be demonstrating more tomorrow morning. We might want to find a place for me to start putting these test fixtures.

pkg/apis/ingress/v1alpha1/types.go
```diff
@@ -22,8 +22,72 @@ type ClusterIngress struct {
...
+	// an unsupported cluster ingress configuration.
+	UnsupportedExtensions *[]string `json:"unsupportedExtensions"`
+}
+
+// NodePlacement wraps node scheduling constructs like selectors
+// and affinity rules.
+type NodePlacement struct {
+	// NodeSelector is a label selector used to inform pod
+	// scheduling.
+	NodeSelector map[string]string `json:"nodeSelector"`
```
Miciah: Should use type `metav1.LabelSelector`.  Since we intend to add additional fields and to support additional rules, it would even make sense to use `*metav1.LabelSelector` so that `ClusterIngressSpec.NodePlacement` could be non-specified without specifying a `ClusterIngressSpec.NodePlacement.NodeSelector`.

pkg/apis/ingress/v1alpha1/types.go
```diff
@@ -22,8 +22,72 @@ type ClusterIngress struct {
...
+	// implemented by fronting the cluster ingress implementation with
+	// a keepalived cluster.
+	//
+	// When using this HA type, a VRRP resource must exist to provide
+	// the configuration and state management for the keepalived
+	// cluster.
+	//
+	// OpenShift will create IPFailover resources as necessary to
+	// materialize the keepalived cluster based on existing managed
+	// ClusterIngress instances.
```
Miciah: Should note that this is not implemented yet.

pkg/manifests/manifests.go
```diff
@@ -78,17 +79,26 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
 
-	ds.Labels = map[string]string{
-		"app": "router",
-	}
+	ds.Spec.Template.Labels["router"] = name
 
 	ds.Spec.Selector = &metav1.LabelSelector{
 		MatchLabels: map[string]string{
-			"app": name,
+			"app":    "router",
```
Miciah: We could use `append` here, or else delete `spec.selector` from `assets/router/daemonset.yaml` if we are going to override `spec.selector.matchLabels` entirely and unconditionally.

pkg/manifests/manifests.go
```diff
@@ -78,17 +79,26 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
 	}
 
-	ds.Spec.Template.Labels["app"] = name
+	routeSelector, err := metav1.LabelSelectorAsSelector(cr.Spec.RouteSelector)
+	if err != nil {
+		return nil, fmt.Errorf("clusteringress %q has invalid spec.routeSelector: %v", cr.Name, err)
+	}
+	env := []corev1.EnvVar{
+		{Name: "ROUTER_SERVICE_NAME", Value: cr.Name},
+		{Name: "ROUTER_SERVICE_NAMESPACE", Value: ds.Namespace},
```
Miciah: Here the use of external assets makes the code confusing—it isn't obvious where `ds.Namespace` is set, or whether an assignment was omitted through oversight.  I think the code would be clearer with an explicit assignment `namespace := ds.Namespace` or (inclusive) comment that we are deliberately using the namespace that is hard-coded in the asset.  Thoughts?

Alternatively, since you are appending `env` to `ds.Spec.Template.Spec.Containers[0].Env`, you could just as well hard-code the environment variable in the yaml.

pkg/manifests/manifests.go
```diff
@@ -78,17 +79,26 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
+		{Name: "ROUTER_CANONICAL_HOSTNAME", Value: cr.Spec.IngressDomain},
+		{Name: "ROUTE_LABELS", Value: routeSelector.String()},
```
Miciah: The `oc adm router` command sets `ROUTE_LABELS` only if the selector is nonempty.  It looks like the router handles an empty value fine though, so it's probably not worth bothering to conditionalize exporting the variable.

pkg/manifests/manifests.go
ironcladlou: Do you think `ROUTER_SERVICE_NAME` makes more sense equal to ClusterIngress.Name or DaemonSet.Name? I'm not sure I have a strong argument either way yet.

Settings the namespace in the static asset seems okay. The confusion between the external assets and code is unfortunately inevitable, so any suggestions for keeping it sane are welcome.

pkg/manifests/manifests.go
ironcladlou: This also speaks to the external asset/code confusion. One argument I can imagine for managing the selector itself in code is that it contains dynamic values and also must match the labels to produce a valid DaemonSet. Maybe that could be incorporated into the manifest unit tests?

That said I guess I prefer removing `spec.selector` from the YAML.

pkg/apis/ingress/v1alpha1/types.go
```diff
@@ -22,8 +22,72 @@ type ClusterIngress struct {
 }
 
 type ClusterIngressSpec struct {
-	// Fill me
+	// IngressDomain is a DNS name suffix serviced by the
+	// ClusterIngress. This value is added automatically to the
+	// status of Routes to inform downstream systems. Required.
```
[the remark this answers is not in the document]
ironcladlou: Made it optional

pkg/apis/ingress/v1alpha1/types.go
ironcladlou: Changed to `*LabelSelector`

pkg/apis/ingress/v1alpha1/types.go
ironcladlou: Removed the whole field until we're ready to implement it

pkg/manifests/manifests.go
ironcladlou: I went ahead and made both `spec.selector` and `spec.template.labels` managed entirely in code for now

pkg/manifests/manifests.go
ironcladlou: I also went ahead and moved the service namespace to YAML since it's static

pkg/manifests/manifests.go
ironcladlou: I made this and canonical hostname optionally assigned

pkg/apis/ingress/v1alpha1/types.go
```diff
@@ -22,8 +22,59 @@ type ClusterIngress struct {
...
 type ClusterIngressSpec struct {
-	// Fill me
+	// IngressDomain is a DNS name suffix serviced by the
+	// ClusterIngress. This value is added automatically to the
+	// status of Routes to inform downstream systems.
+	IngressDomain *string `json:"ingressDomain"`
+
+	// NodePlacement provides additional control over the
+	// scheduling of ingress implementation pods.
+	NodePlacement *NodePlacement `json:"nodePlacement"`
```
Miciah: No reason we couldn't wire this up now—should be as simple as the following:
```go
if cr.Spec.NodePlacement != nil {
    if cr.Spec.NodePlacement.NodeSelector != nil {
        ds.Spec.Template.Spec.NodeSelector = cr.Spec.NodePlacement.NodeSelector
    }
}
```
Right?

Miciah: Well, I guess we'd need some `metav1.LabelSelectorAsSelector` conversion in there as well:
```go
if cr.Spec.NodePlacement != nil {
    if cr.Spec.NodePlacement.NodeSelector != nil {
        nodeSelector, err := metav1.LabelSelectorAsSelector(cr.Spec.RouteSelector)
	if err != nil {
	    return nil, fmt.Errorf("clusteringress %q has invalid spec.routeSelector: %v", cr.Name, err)
	}
        ds.Spec.Template.Spec.NodeSelector = nodeSelector
    }
}
```

pkg/apis/ingress/v1alpha1/types.go
```diff
@@ -22,8 +22,59 @@ type ClusterIngress struct {
...
+
+	// NamespaceSelector is a label selector which filters the
+	// namespaces covered by this ClusterIngress.
+	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector"`
```
Miciah: This isn't wired up yet.  Should be very similar to `RouteSelector`.

pkg/apis/ingress/v1alpha1/types.go
ironcladlou: Please do this in a followup

deploy/cr.yaml
```diff
@@ -2,3 +2,7 @@ apiVersion: "ingress.openshift.io/v1alpha1"
 kind: "ClusterIngress"
 metadata:
   name: "default"
+spec:
+  ingressDomain: cluster.local
```
Miciah: If we set a default ingress domain, it should be `${ROUTER_SERVICE_NAME.${ROUTER_SERVICE_NAMESPACE}.cluster.local`, which means we'll need to patch it in using Go code.

We can do this in a follow-up.

Miciah:
/lgtm

There are a few follow-ups for future work, as noted in the comments.

deploy/cr.yaml
ironcladlou: This file is just some autogenerated thing that's not very useful... it should  be deleted and replaced with test fixtures somewhere

pkg/apis/ingress/v1alpha1/types.go
Miciah: Followup: #39.

pkg/apis/ingress/v1alpha1/types.go
Miciah: Done in #39.

PR #19, opened 2018-09-06
 Do not duplicate default values in yaml

pravisankar:
- { app: router } label is already defaulted in the yaml file
- Removed some fields from yaml which we know will be replaced at runtime.
  We can keep only known default values in yaml manifest and apply cluster
  ingress specific values at runtime.


Makefile
```diff
@@ -15,8 +15,8 @@ GO_BUILD_RECIPE=GOOS=$(GOOS) go build -o $(BIN) $(MAIN_PACKAGE)
 build:
 	$(GO_BUILD_RECIPE)
 
+# Using "-modtime 1" to make generate target deterministic. It sets all file time stamps to unix timestamp 1
 generate: $(GOBINDATA_BIN)
-	# Using "-modtime 1" to make generate target deterministic. It sets all file time stamps to unix timestamp 1
 	go-bindata -mode 420 -modtime 1 -pkg manifests -o $(BINDATA) assets/...
```
ramr: `420`?  share some of what you are having! ;^)

pkg/manifests/manifests.go
```diff
@@ -79,17 +79,15 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
 
 	ds.Name = name
 
-	ds.Spec.Template.Labels = map[string]string{
-		"app":    "router",
-		"router": name,
+	if ds.Spec.Template.Labels == nil {
+		ds.Spec.Template.Labels = map[string]string{}
```
ramr: `app` not required?  in here and the other maps below.

pkg/manifests/manifests.go
```diff
@@ -79,17 +79,15 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
 	}
+	ds.Spec.Template.Labels["router"] = name
```
Miciah: We should update `assets/router/daemonset.yaml` to indicate that only the one field, and not the entirety of `.spec.template.metadata.labels`, is replaced at uptime.

pkg/manifests/manifests.go
```diff
@@ -79,17 +79,15 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
 
-	ds.Spec.Selector = &metav1.LabelSelector{
-		MatchLabels: map[string]string{
-			"app":    "router",
-			"router": name,
-		},
+	if ds.Spec.Selector.MatchLabels == nil {
+		ds.Spec.Selector.MatchLabels = map[string]string{}
 	}
+	ds.Spec.Selector.MatchLabels["router"] = name
```
Miciah: We should update `assets/router/service-cloud.yaml` to reflect that only the one field, and not the entirety of `.spec.selector`, is replaced at runtime.

pkg/manifests/manifests.go
```diff
@@ -122,15 +120,15 @@ func (f *Factory) RouterServiceCloud(cr *ingressv1alpha1.ClusterIngress) (*corev
 
 	s.Name = name
 
-	s.Labels = map[string]string{
-		"app":    "router",
-		"router": name,
+	if s.Labels == nil {
+		s.Labels = map[string]string{}
 	}
+	s.Labels["router"] = name
```
Miciah: We should update `assets/router/service-cloud.yaml` to reflect that only the one field, and not the entirety of `.metadata.labels`, is replaced at runtime.  Ditto for `spec.selector`.

pkg/manifests/manifests.go
Miciah: Whoops, I commented in the wrong place.

imcsk8:
Previous comments aside, LGTM

pkg/manifests/manifests.go
pravisankar: Yes, `app` label is already defaulted in the yaml file:
https://github.com/openshift/cluster-ingress-operator/blob/f0ab7cc3c0e9b6cda2b565f1acdfd0456ffb505c/assets/router/daemonset.yaml#L17

pkg/manifests/manifests.go
pravisankar: righto

pravisankar:

assets/router/daemonset.yaml
```diff
@@ -1,23 +1,20 @@
...
+  # Name is set at runtime.
   namespace: openshift-cluster-ingress-router
   labels:
     app: router
 spec:
-  # Replaced at runtime.
   selector:
     matchLabels:
       app: router
-      router: router
```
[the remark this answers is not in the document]
pravisankar: I understood your initial suggestion but I thought commenting each field may be hard to maintain and  we can easily figure out what additional fields are set/appended at runtime by just inspecting the manifests.go file.

Makefile
```diff
@@ -1,4 +1,4 @@
-all: build
+all: build generate
```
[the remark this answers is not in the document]
pravisankar: Good catch. I need to fix this in cluster-dns-operator repo as well.

Makefile
ironcladlou: Not all of us get to live in California! 😭

ironcladlou:
Let me add some additional context by trying to lay out what I think are the principles of the current approach:

1. Dummy values used in the YAML in places where the operator will override the values.
    * Pro: The YAML by itself is relatively complete/useful and could actually be applied outside the operator and work for a simple use case (and could more easily be post-processed by a script to replace the dummy values if desired)
    * Con: Requires maintenance discipline to keep up to date with manifest.go changes. (On the other hand, in the absence of dummy values, if we want comments where the removed fields are, this may be a wash.)
2. In the specific case of labels and associated match selectors, avoid merging partial sets spread across YAML and Go: specify the full set in YAML (dummy values) and replace the entire set in Go
    * Pro: Prevents mistakes where the labels and matcher could become mismatched, invalidating the deployment/daemonset or causing subtle pod adoption bugs. (to be fair, we could enforce this kind of invariant in unit tests)
    * Pro: Same legibility/standalone pro as above re: dummy values
    * Con: Same potential maintenance concern as above re: dummy values

If we don't want either dummy values OR comments in the YAML file, and instead want to rely entirely on inspection of the Go code (which may sprawl out beyond manifests.go as time progresses), we should be sure to:

1. Agree that we don't care about the YAML files being useful outside the operator
2. Agree on a consistent approach for labels and selectors specifically

I think I'm ultimately fine with whichever approach at the moment, as long as we all agree on an explicit set of principles/style guidelines to apply consistently going forward.

Thoughts?

pravisankar:
My 2 cents:
- For testing, we can always get complete yaml file with a simple script that calls methods in manifests.go
- Yes, there shouldn't be a risk of label/slector mismatch as we will be appending additional labels or selectors. This is similar to additional envs merged with existing default envs.
- Dummy values obscures reality. Upstream (k8s and coredns repos) approach is to maintain template file and the variables are filled at runtime. 
- If we can maintain runtime variable handling in known places (not sprawled but limited to couple of files), then either yaml templates or yaml with only defaults might be easy to maintain.

ironcladlou:
Tested with my manual "e2e" suite, looks good!

/lgtm

PR #21, opened 2018-09-10
Add operator manifest

imcsk8:
This manifest is needed in order to run the
Cluster Ingress Operator as a container inside
OpenShift.

Also added running information to the documentation.

### DIFF
--- a/deploy/cluster-ingress-operator.yaml
+++ b/deploy/cluster-ingress-operator.yaml
@@ -1,0 +1,31 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-ingress-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: cluster-ingress-operator
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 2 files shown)


imcsk8:
PTAL @openshift/sig-network-edge @ironcladlou @pravisankar @Miciah

deploy/cluster-ingress-operator.yaml
```diff
@@ -0,0 +1,31 @@
...
+      containers:
+        - name: cluster-ingress-operator
+          image: openshift/cluster-ingress-operator
+          ports:
+          - containerPort: 60000
+            name: metrics
+          command:
+          - cluster-ingress-operator
+          imagePullPolicy: Always
+          env:
```
ironcladlou: Isn't `KUBERNETES_CONFIG` also needed?

imcsk8: I haven't needed it in my tests. I just run it inside a openshift environment as the system user.
```
oc create -f deploy/cluster-ingress-operator.yaml
```

PR #23, opened 2018-09-13
Simplify running integration tests

pravisankar:
- CRD is created as part of the test
- WATCH_NAMESPACE env is now optional

Command to run integration tests:
```
KUBECONFIG=<admin.kubeconfig> CLUSTER_NAME=<gcp-cluster-name> make test-integration
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,11 @@ $(GOBINDATA_BIN):
 test:
 	go test ./...
 
+test-integration:
+	go test -v -tags integration ./test/integration
+
 clean:
 	go clean
 	rm -f $(BIN)
 
-.PHONY: all build generate test clean
+.PHONY: all build generate test test-integration clean
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


pravisankar:

imcsk8:
Code looks ok

Makefile
```diff
@@ -12,6 +12,7 @@ GOBINDATA_BIN=$(GOPATH)/bin/go-bindata
 ENVVAR=GOOS=linux GOARCH=amd64 CGO_ENABLED=0
 GOOS=linux
 GO_BUILD_RECIPE=GOOS=$(GOOS) go build -o $(BIN) $(MAIN_PACKAGE)
+GOCACHE ?= off
```
ironcladlou: I think disabling test caching for integration/extended/e2e tests absolutely makes sense (otherwise the flake detection aspect is lost); however, I'm not sure why we'd want to disable test caching for unit tests.

test/integration/integration_test.go
```diff
@@ -21,43 +22,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-var operatorNamespace string
-var clusterName string
+func TestMultipleIngresses(t *testing.T) {
```
ironcladlou: I was using subtests deliberately to support setup/teardown around (theoretical) multiple test cases involving the same operator deployment. Now every test case in this file will spin up a new competing operator goroutine.

test/integration/integration_test.go
```diff
@@ -200,3 +173,68 @@ func testMultipleIngresses(t *testing.T) {
...
+
+	clusterName := os.Getenv("CLUSTER_NAME")
+	if len(clusterName) == 0 {
+		t.Fatalf("CLUSTER_NAME is required")
+	}
+	config.clusterName = clusterName
+
+	namespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		namespace = "default"
```
ironcladlou: Just a note: I want this to go away ASAP; we should be applying /manifests which will put the operator into its own namespace with all the requisite RBAC resources rather than piggybacking on `default`.

test/integration/integration_test.go
```diff
@@ -200,3 +173,68 @@ func testMultipleIngresses(t *testing.T) {
...
+
+func (tc *TestConfig) createCRD() {
+	tc.runShellCmd(fmt.Sprintf("oc apply -f ../../deploy/crd.yaml -n %s", tc.operatorNamespace), "create cluster ingress CRD")
+}
+
+func (tc *TestConfig) deleteCRD() {
+	tc.runShellCmd(fmt.Sprintf("oc delete crd clusteringresses.ingress.openshift.io -n %s", tc.operatorNamespace), "delete cluster ingress CRD")
+}
+
+func (c *TestConfig) runShellCmd(command, msg string) {
```
ironcladlou: This is fine for now, but I think we should consider wrapping this in a script that invokes our tests, or compiling the manifests into bindata and using our Go client to apply them.

Makefile
pravisankar: I ran 'make test/test-integration' multiple times and the tests ran once and subsequent test results came from cache. Now I think this should be the expected behavior as there is no change in code. Any code change is going to invalidate the cache. I will remove the GOCACHE env.

test/integration/integration_test.go
pravisankar: Sure. Till we land that change, this will not force users to specify another env (WATCH_NAMESPACE)

test/integration/integration_test.go
pravisankar: Any yaml that doesn't need runtime change, we can run directly using cli command and for others we can generate objects.

pravisankar:

ironcladlou:
Thanks!

/lgtm

PR #24, opened 2018-09-14
Add Namespace for Cluster Ingress Operator

imcsk8:
The Cluster Ingress Operator Service Account needs a namespace
for isolating permissions.

### DIFF
--- a/deploy/namespace.yaml
+++ b/deploy/namespace.yaml
@@ -1,0 +1,4 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+    name: openshift-cluster-ingress-operator


imcsk8:
PTAL @openshift/sig-network-edge

deploy/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+    name: openshift-cluster-ingress-operator
+    labels:
+        name: openshift-cluster-ingress-operator
```
Miciah: Why is the label needed? I don't see similar labels on other operators' namespaces.

imcsk8: I added it just for information purposes, i'll remove it so we have the basic namespace configuration.

PR #25, opened 2018-09-18
Set deployment order

imcsk8:
In order to deploy a cluster ingress operator some
resource have to be deployed in a specific order.
The point of this patch is to allow user to issue a
oc create -f deploy/.

### DIFF
--- a/deploy/05_cr.yaml
+++ b/deploy/05_cr.yaml
@@ -1,0 +1,8 @@
+apiVersion: "ingress.openshift.io/v1alpha1"
+kind: "ClusterIngress"
+metadata:
+  name: "default"
+spec:
+  ingressDomain: cluster.local
+  highAvailability:
+    type: Cloud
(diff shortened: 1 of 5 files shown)


imcsk8:
PTAL @openshift/sig-network-edge @ironcladlou @Miciah @pravisankar

deploy/05_cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: "ingress.openshift.io/v1alpha1"
```
ironcladlou: This one should be deleted; the installer and/or end-user will create it

ironcladlou:
Please delete 05_cr.yaml in a followup

/lgtm

ironcladlou:
Let's also rename the directory `manifests` and add it to the Dockerfile in a followup?

Miciah:
This PR added new files; was the intention to rename the existing files? If so, can we do a follow-up PR to remove the new files (with `git rm`) and then rename the old files (with `git mv`) in order to preserve history?

ironcladlou:
Feel free to reshuffle in whatever way is helpful for history preservation. Technically our manifests can live in `deploy`, but it seems to make sense to rename it `manifests` since that's what the Cluster Version Operator expects for our images anyway.

PR #26, opened 2018-09-18
Makefile: Fix handling of GOPATH

Miciah:
Instead of interpolating `GOPATH` directly into the pathname for `go-bindata`, treat it as a search path.

Before this commit, if `GOPATH` had more than one entry, make would fail with the following error message:

    Makefile:20: *** target pattern contains no `%'.  Stop.


imcsk8:
LGTM

ironcladlou:
>Before this commit, if GOPATH had more than one entry, make would fail with the following error message:

I'd recommend against bothering to support multiple GOPATH entries. GOPATH itself is going away soon.

That said,
/lgtm

PR #27, opened 2018-09-19
Remove custom resource creation

imcsk8:
There's no need to have the custom resource creation
in the deploy directory.


imcsk8:
PTAL @openshift/sig-network-edge @ironcladlou @Miciah @pravisankar

PR #29, opened 2018-09-20
Handle delete events

imcsk8:
Handle delete events for ClusterIngress objects.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -12,6 +12,10 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
 )
 
 func NewHandler() sdk.Handler {
@@ -25,12 +29,18 @@ type Handler struct {
 }
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	if event.Deleted {
-		return nil
-	}
 	switch o := event.Object.(type) {
 	case *ingressv1alpha1.ClusterIngress:
-		return h.syncIngressUpdate(o)
+		if event.Deleted {
+			logrus.Infof("Deleting ClusterIngress object: %s", o.Name)
+			err := h.deleteIngress(o)
+			if err != nil {
+				return fmt.Errorf("error deleting ClusterIngress %s", err)
+			}
+			return nil
+		} else {
+			return h.syncIngressUpdate(o)
+		}
 	}
 	return nil
 }
@@ -98,6 +108,12 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 			if err != nil {
 				return fmt.Errorf("couldn't build service: %v", err)
 			}
+			dsRef, err := getDaemonSetOwnerRef(ds)
+			// Don't create the service unless the DaemonSet is ready
+			if err != nil {
+				return fmt.Errorf("failed to create service, could not get DaemonSet ownerReference: %s", err)
+			}
+			service.SetOwnerReferences(append(service.GetOwnerReferences(), dsRef))
 			err = sdk.Create(service)
 			if err == nil {
 				logrus.Infof("created router service %s/%s", service.Namespace, service.Name)
@@ -108,4 +124,41 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 	}
 
 	return nil
+}
+
+func (h *Handler) deleteIngress(ci *ingressv1alpha1.ClusterIngress) error {
+	ds, err := h.manifestFactory.RouterDaemonSet(ci)
+	if err != nil {
+		return fmt.Errorf("couldn't build DaemonSet object for deletion: %v", err)
+	}
+	return sdk.Delete(ds)
+}
+
+// getDaemonSetOwnerRef returns an object as OwnerReference
+func getDaemonSetOwnerRef(ds *appsv1.DaemonSet) (metav1.OwnerReference, error) {
+	var or metav1.OwnerReference
+	d := &appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "DaemonSet",
+			APIVersion: "apps/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ds.Name,
+			Namespace: ds.Namespace,
+		},
+	}
+
+	err := sdk.Get(d)
+	if err != nil {
+		return or, fmt.Errorf("couldn't get DaemonSet %s Namespace %s : %s", ds.Name, ds.Namespace, err)
+	}
+	trueVar := true
+	or = metav1.OwnerReference{
+		APIVersion: d.APIVersion,
+		Kind:       d.Kind,
+		Name:       d.Name,
+		UID:        d.UID,
+		Controller: &trueVar,
+	}
+	return or, nil
 }


imcsk8:
Starting conversation PTAL @openshift/sig-network-edge @ironcladlou @Miciah @knobunc @pravisankar

pkg/stub/handler.go
```diff
@@ -25,18 +27,28 @@ type Handler struct {
...
 func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
+	// Get the Owner Reference object for the current ClusterIngress
+	ciOwnerRef := getCIOwnerRef(ci)
+
 	ns, err := h.manifestFactory.RouterNamespace()
+	/* Set the Cluster Ingress Operator Object as Owner of the namespace
+	 * this allows the cascading deletion of all the objects when the
+	 * Cluster Ingress Operator object is deleted
+	 */
+	setOwnerRef(ns, ciOwnerRef)
```
pravisankar: When more than one router is deployed (e.g. default and internal), deleting one router should not remove the entire router namespace. Multiple routers are sharing the same router namespace, serviceaccount, cluster roles and role bindings. 
I think you need to remove router namespace if there is only one router otherwise only remove daemonset and service which is specific to the deleted clusteringress custom resource. Is owner ref solves this use case?

ironcladlou: Yeah... and then, why bother deleting the namespace anyway?

ironcladlou: Each router would need its own namespace

ironcladlou: GC doesn't allow inter-namespace relationships, so owner refs are of no use when it comes to ClusterIngress-><resource set in router namespace>

imcsk8: I think this is enough for now since we only have one router per ClusterIngress.

The use of ownerReferences pattern is the recommended way to delete objects in kubernetes: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

When we have other use cases we can easily refactor to deal with them using this pattern.

ironcladlou: We MUST support multiple routers in that namespace, it's a core feature and we already support it; if we stop supporting it that's a regression at this point. Regardless of the user guide, I know from working on GC with upstream that the use case may work sometimes but isn't supported. See recent commentary on this use case (namespaced owner, cluster scoped dependent): https://github.com/kubernetes/kubernetes/issues/65200

Relying on unsupported GC behavior caused one of our teams trouble last year when the gap was plugged, let's not set ourselves up for trouble.

pkg/stub/handler.go
```diff
@@ -109,3 +113,35 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 
 	return nil
 }
+
+func (h *Handler) deleteIngress(ci *ingressv1alpha1.ClusterIngress) error {
```
ironcladlou: Should probably add a TODO to refactor this to use GC so we need only delete the DaemonSet

pkg/stub/handler.go
```diff
@@ -25,12 +29,15 @@ type Handler struct {
 }
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	if event.Deleted {
-		return nil
-	}
 	switch o := event.Object.(type) {
 	case *ingressv1alpha1.ClusterIngress:
-		return h.syncIngressUpdate(o)
+		if event.Deleted {
```
pravisankar: Prefer using switch and case that will give clear picture of which events are handled or ignored (general pattern used in openshift for handling events)
```
switch event.Type {
case event.Added, event.Modified: (or event.Synced, check what is exposed by sdk.Event)
...
case event.Deleted:
}
```

pkg/stub/handler.go
```diff
@@ -25,12 +29,15 @@ type Handler struct {
...
+			logrus.Infof("Deleting ClusterIngress object: %s", o.Name)
```
pravisankar: glog? I'm fine with logrus, we just need to be consistent everywhere.

pkg/stub/handler.go
```diff
@@ -109,3 +118,50 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 
 	return nil
 }
+
+func (h *Handler) deleteIngress(ci *ingressv1alpha1.ClusterIngress) error {
```
pravisankar: Error returned by deleteIngress is ignored by the caller. We are not going to propagate the error to the user (which is the case here), just log the error instead of returning the error or at least the caller need to log the error.

pkg/stub/handler.go
```diff
@@ -109,3 +118,50 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ds.Name,
+			Namespace: ds.Namespace,
+		},
+	}
+
+	err := sdk.Get(d)
+	if err != nil {
+		logrus.Infof("Couldn't get DaemonSet %s Namespace %s : %s", ds.Name, ds.Namespace, err)
```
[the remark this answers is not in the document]
pravisankar: This is one of the places where instead of creating &appsv1.DaemonSet{} and calling sdk.Get(), using kubeClient directly would have been simpler! I'm fine with this.

pkg/stub/handler.go
```diff
@@ -109,3 +118,50 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
...
+		Kind:       d.Kind,
+		Name:       d.Name,
+		UID:        d.UID,
+		Controller: &trueVar,
+	}
+}
+
+// setOwnerRef sets the desired OwnerReference to an object
+func setOwnerRef(obj metav1.Object, ownerRef metav1.OwnerReference) {
+	logrus.Infof("Setting Owner Reference for %s to: %s", obj.GetName(), ownerRef.Name)
```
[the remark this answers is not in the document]
pravisankar: yes, don't log this one.

pkg/stub/handler.go
pravisankar: sdk.Event is different from watch.Event, it only sets Deleted flag in case of deletion and Object is the latest in case of add or modify event (no other Added, Modified or Synced events).
Ignore my comments, your implementation looks good.

pkg/stub/handler.go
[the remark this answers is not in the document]
imcsk8: i would like to stick using the sdk, if we need to use the k8s api directly we should fix it on a PR for the sdk

pkg/stub/handler.go
imcsk8: logrus seems to be the logger that the sdk uses

pkg/stub/handler.go
```diff
@@ -25,12 +29,18 @@ type Handler struct {
...
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	if event.Deleted {
-		return nil
-	}
 	switch o := event.Object.(type) {
 	case *ingressv1alpha1.ClusterIngress:
-		return h.syncIngressUpdate(o)
+		if event.Deleted {
+			logrus.Infof("Deleting ClusterIngress object: %s", o.Name)
+			err := h.deleteIngress(o)
```
pravisankar: return h.deleteIngress(o)?

pkg/stub/handler.go
```diff
@@ -98,6 +108,13 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 			if err != nil {
 				return fmt.Errorf("couldn't build service: %v", err)
 			}
+			dsRef, err := getDaemonSetOwnerRef(ds)
+			// Don't create the service unless the DaemonSet is ready
+			if err != nil {
+				return fmt.Errorf("failed to create service, could not get DaemonSet ownerReference: %s", err)
+			} else {
```
pravisankar: else is redundant, we don't follow this convention, just call service.SetownerReferences() after error handling.

pkg/stub/handler.go
```diff
@@ -109,3 +126,45 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 
 	return nil
 }
+
+func (h *Handler) deleteIngress(ci *ingressv1alpha1.ClusterIngress) error {
+	ds, err := h.manifestFactory.RouterDaemonSet(ci)
+	if err != nil {
+		return fmt.Errorf("couldn't build DaemonSet object for deletion: %v", err)
+	}
+	err = sdk.Delete(ds)
```
pravisankar: you could do: return sdk.Delete(ds)

PR #31, opened 2018-09-21
 Updated docs to match resource files in manifests dir

pravisankar:


pravisankar:

pravisankar:
Also noticed openshift-cluster-ingress operator complaining with this error in the logs:
```
time="2018-09-21T19:36:30Z" level=info msg="Watching ingress.openshift.io/v1alpha1, ClusterIngress, openshift-cluster-ingress-operator, 5"
ERROR: logging before flag.Parse: E0921 19:36:30.276183       1 reflector.go:205] github.com/openshift/cluster-ingress-operator/vendor/github.com/operator-framework/operator-sdk/pkg/sdk/informer.go:84: Failed to list *unstructured.Unstructured: clusteringresses.ingress.openshift.io is forbidden: User "system:serviceaccount:openshift-cluster-ingress-operator:cluster-ingress-operator" cannot list clusteringresses.ingress.openshift.io in the namespace "openshift-cluster-ingress-operator": no RBAC policy matched
```
We need to patch RBAC policy rules.

imcsk8:
I just noticed that error, i thought it was my setup.
For what this patch is concerned LGTM

manifests/04_cluster-ingress-operator.yaml
```diff
@@ -21,7 +21,7 @@ spec:
             name: metrics
           command:
           - cluster-ingress-operator
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
ironcladlou: If the default is a `latest` tag, would `Always` be more convenient for development?

pravisankar: Yes, helps in development but I don't think we want to ship with 'Always' image pull policy.

ironcladlou: Good point, let's remove the key

imcsk8:
LGTM

pravisankar:

ironcladlou:
We might want to add some example back in (but not in /manifests)

/lgtm

PR #33, opened 2018-09-24
Update to the latest operator-sdk and client libs

ironcladlou:
Update to the latest operator-sdk and client libs.


ironcladlou:
I still need to run the tests.

pravisankar:
@ironcladlou can you summarize what additional capabilities we can leverage with the new sdk changes?

Miciah:
The packages include changelogs:

https://github.com/openshift/cluster-ingress-operator/pull/33/files#diff-d2967e420acbe593b0690388caa03e87
https://github.com/openshift/cluster-ingress-operator/pull/33/files#diff-2d21dfc01633a2d9f479f2aef7832a81

Here are the items that jumped out for me:

• The `operator-sdk build` command is now `operator-sdk new`.
• operator-sdk now includes a test framework.
• `sdk.Watch` now expects a `time.Duration` for the resync interval (this PR makes the necessary change in the operator).
• client-go now has `NewFilteredSharedInformerFactory`, which filters by namespace.
• client-go now has an API for conditions on StatefulSets and DaemonSets.
•  The default garbage collection policy for deployments, daemonsets, replicasets, and statefulsets from `OrphanDependents` to `DeleteDependents`.

Miciah:
I may have misread the `operator-sdk build` change.  Here is the changelog entry:
> Moved the rendering of `deploy/operator.yaml` to the `operator-sdk new` command instead of `operator-sdk build

ironcladlou:
I'm doing this primarily as an exercise in understanding how we'll keep up to date with SDK changes (especially bugfixes); the main change I was interested in was upgrading our kube client to 1.10 (which of course is still not where we'd like to be, which is 1.11).

imcsk8:
Is this type of big code addition is going to happen to all the operators?
I don't want to be a pesimist but it feels a little wrong.

Miciah:
> Is this type of big code addition is going to happen to all the operators?

Presumably so.  Is it vendoring packages that bothers you, or is there something different and distinctly wrong with how we do it compared to other Go projects?

PR #34, opened 2018-09-25
Do not use 'operator' name for router cluster role and role bindings

pravisankar:
Otherwise it will create confusion with roles used in cluster-ingress-operator.


pravisankar:

imcsk8:
LGTM i got confused by this

ironcladlou:
The names like `cluster-ingress-operator:router` were meant to be disambiguating by using the `cluster-ingress-operator:` prefix. For example, when we roll out keepalived deployments, we may want `cluster-ingress-operator:keepalived`.

pravisankar:

Now the actual operator will use 'cluster-ingress-operator' name for its ClusterRole/ClusterRoleBindings and ingress operator driven components like router, keepalived, etc. will use 'cluster-ingress:' prefix (e.g. cluster-ingress:router, cluster-ingress:keepalived) for their cluster roles.

ironcladlou:
I like the way this looks, thanks

/lgtm

PR #35, opened 2018-09-25
Fix rbac roles for cluster ingress operator

pravisankar:


pravisankar:

Miciah:
The operator needs to be able to create the router's namespace, service account, cluster role, cluster rolebinding, daemonset, and service and watch, update, and delete the daemonset and service.  Doesn't that all belong in `03_rbac.yaml`?

Were the changes to `04_cluster-ingress-operator.yaml` accidentally included in this PR?

pravisankar:

@Miciah  Creation of router's namespace, service account, cluster role and role bindings are part of binary run by operator image. These roles are for the operator itself. 
Operator will be deployed as a pod and the binary running in the operator will watch for any clusteringress objects and creates router related components on demand.

Segregated rbac changes with ingress operator manifest cleanup.

imcsk8:
I aggre with @Miciah the operator needs to create the namespace and other elements for the router, creation privileges is a must.

ironcladlou:
The operator will need permissions to work with everything [here](https://github.com/openshift/cluster-ingress-operator/tree/master/assets/router).

pravisankar:

PR #36, opened 2018-09-27
Add examples directory

imcsk8:
Some manifests are needed to manage an test the Cluster Ingress Operator, this patch adds the
custom resource creation example.


imcsk8:
PTAL @openshift/sig-network-edge @ironcladlou @Miciah @pravisankar @knobunc

PR #38, opened 2018-09-27
Publish ClusterIngress status

Miciah:
Update the status conditions of ClusterIngress objects to indicate synchronization status and availability.

Watch Services in cluster-ingress-operator's namespace in order to reflect their status and any associated load-balancer ingresses to the corresponding ClusterIngress objects.

* `cmd/cluster-ingress-operator/main.go`: Watch services and endpoints in the operator's namespace.
* `manifests/00-cluster-role.yaml`: Allow operator to get services and endpoints.
* `pkg/apis/ingress/v1alpha1/status.go`: Add new functions for setting status on ClusterIngress objects: `setStatusCondition`, `SetStatusSyncCondition`, `SetStatusAvailableCondition`, `clusterIngressStatusEqual`, `UpdateStatus`.
* `pkg/apis/ingress/v1alpha1/types.go`: Add fields to `ClusterIngressStatus` for operator and load balancer status.
* `pkg/apis/ingress/v1alpha1/status_test.go`: Add tests.
* `pkg/stub/handler.go`: Use `SetStatusSyncCondition`, `SetStatusAvailableCondition`, and `UpdateStatus` in `syncIngressUpdate`.
Add new functions: `getClusterIngress`, `syncServiceUpdate`, `syncEndpointsUpdate`, and `syncServiceOrEndpointsUpdate`.
Use these new functions in `Handle` to handle Service and Endpoints events and to update the corresponding ClusterIngress status according to the Service and Endpoints status.

---

This PR depends on #47.


Miciah:
Rebased.

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,202 @@
...
+
+// UpdateStatus updates the receiver ClusterIngress object's status if it
+// differs from the provided ClusterIngress object's status.
+func (ci *ClusterIngress) UpdateStatus(oldCi *ClusterIngress) {
+	if clusterIngressStatusEqual(oldCi, ci) {
+		return
+	}
+
+	err := sdk.Update(ci)
+	for tries := 3; tries > 0; tries-- {
```
imcsk8: Why is the loop needed?
the err variable is updated outside the loop or am i missing something?

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,202 @@
...
+	if !available {
+		condition.Status = operatorapi.ConditionFalse
+		condition.Reason = "ServiceNotReady"
+		condition.Message = message
+	}
+
+	ci.SetStatusCondition(condition)
+}
+
+// clusterIngressStatusEqual returns true iff the status conditions (ignoring
```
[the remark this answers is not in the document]
Miciah: It's short for "if and only if".  I see it used a lot in included packages, but I can expand it.

pkg/apis/ingress/v1alpha1/status.go
Miciah: Whoops! `err := sdk.Update(ci)` needs to be inside the loop.  Thanks!

Miciah: I deleted the loop.  Now `UpdateStatus` returns an error value, and callers return any error from `UpdateStatus` to the handler so that the event gets reprocessed.

pkg/apis/ingress/v1alpha1/status.go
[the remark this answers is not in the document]
Miciah: Fixed.

Miciah:
Latest push deletes the retry logic in `UpdateStatus`; instead, `UpdateStatus` returns an error value, and its callers return any error from `UpdateStatus` to the handler so that the operator will retry processing the event.

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,190 @@
...
+		if conditionA.Type != conditionB.Type ||
+			conditionA.Status != conditionB.Status ||
+			conditionA.Reason != conditionB.Reason ||
+			conditionA.Message != conditionB.Reason {
+			return false
+		}
+	}
+
+	numIngresses := len(ci.Status.LoadBalancer.Ingress)
+	if numIngresses != len(oldCi.Status.LoadBalancer.Ingress) {
```
pravisankar: Perform load balancer checks only when ci.Spec.HighAvailability.Type == ingressv1alpha1.CloudClusterIngressHA ?
When ci.Spec.HighAvailability.Type  != CloudClusterIngressHA , need to clear old Status.LoadBalancer field.

pkg/stub/handler.go
```diff
@@ -162,3 +214,98 @@ func getDaemonSetOwnerRef(ds *appsv1.DaemonSet) (metav1.OwnerReference, error) {
...
+		}
+	}
+	if !serviceReady {
+		ci.SetStatusAvailableCondition(false,
+		                               "service endpoints not ready")
+
+		return nil
+	}
+
+	if len(service.Status.LoadBalancer.Ingress) == 0 {
```
pravisankar: Check valid only in case of ci.Spec.HighAvailability.Type == ingressv1alpha1.CloudClusterIngressHA

pkg/stub/handler.go
```diff
@@ -162,3 +214,98 @@ func getDaemonSetOwnerRef(ds *appsv1.DaemonSet) (metav1.OwnerReference, error) {
...
+		ci.SetStatusAvailableCondition(false,
+		                               "service has no ingresses")
+
+		return nil
+	}
+
+	ci.SetStatusAvailableCondition(true, "")
+	ci.Status.LoadBalancer.Ingress = service.Status.LoadBalancer.Ingress
```
pravisankar: Set this field only when ci.Spec.HighAvailability.Type == ingressv1alpha1.CloudClusterIngressHA ?

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,190 @@
...
+
+// clusterIngressStatusEqual returns true if and only if the status conditions
+// (ignoring LastTransitionTime) and ingresses of the provided ClusterIngress
+// objects are equal.
+func clusterIngressStatusEqual(oldCi, ci *ClusterIngress) bool {
+	numConditions := len(ci.Status.Conditions)
+	if numConditions != len(oldCi.Status.Conditions) {
+		return false
+	}
+	for i := 0; i < numConditions; i++ {
```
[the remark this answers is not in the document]
Miciah: Fixed.  Thanks!

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,190 @@
...
+			conditionA.Message != conditionB.Reason {
+			return false
+		}
+	}
+
+	numIngresses := len(ci.Status.LoadBalancer.Ingress)
+	if numIngresses != len(oldCi.Status.LoadBalancer.Ingress) {
+		return false
+	}
+	for i := 0; i < numIngresses; i++ {
```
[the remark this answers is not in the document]
Miciah: Fixed.  Thanks!

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,190 @@
...
+func (ci *ClusterIngress) CreateOrUpdateOperatorStatus(condition osv1.OperatorStatusCondition) error {
+	operatorNamespace, err := k8sutil.GetWatchNamespace()
+	if err != nil {
+		logrus.Fatalf("Failed to get watch namespace: %v", err)
+	}
+
+	operatorName, err := k8sutil.GetOperatorName()
+	if err != nil {
+		logrus.Fatalf("Failed to get operator name: %v", err)
+	}
```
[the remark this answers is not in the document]
Miciah: I added the fields to `Handler`.  Since this means `CreateOrUpdateOperatorStatus` needs the Handler, and since it didn't really need the ClusterIngress, I changed its receiver to `h *Handler`, moved it to `handler.go`, and changed it to `createOrUpdateOperatorStatus`.

pkg/stub/handler.go
```diff
@@ -38,14 +42,57 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+			Type:    osv1.OperatorStatusConditionTypeDone,
+			Message: fmt.Sprintf("done applying state: %#v", o),
+		}
+		if err := o.CreateOrUpdateOperatorStatus(cond); err != nil {
+			return err
+		}
+
+		return nil
+	case *corev1.Service:
+		isDeleted := o.GetDeletionTimestamp() != nil
```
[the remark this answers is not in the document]
Miciah: My understanding is that `event.Deleted` will go away that and `o.GetDeletionTimestamp()` will be the source of truth.

pkg/apis/ingress/v1alpha1/status.go
Miciah: As you point out in another comment, the logic in `syncServiceUpdate` needs to be fixed for `ci.Spec.HighAvailability.Type != CloudClusterIngressHA`, but once that is fixed, I think the equality check should still always compare `.Status.LoadBalancer.Ingress`.  If we ever support changing from `CloudClusterIngressHA` to some other value, we want the equality check to return false so that we update with the status with the empty list of ingresses.  Am I missing anything?

pkg/stub/handler.go
Miciah: Good point.  Thanks!

Miciah:
Latest push makes the following changes per Ravi's comments:

• Move `ClusterIngressStatusEqual` from `status.go` to `handler.go`, rename it to `clusterIngressStatusEqual`, and change its receiver from `ci *ClusterIngress` to `h *Handler`
• Add `operatorNamespace` and `operatorName` to `Handler` and initialize them in `NewHandler`.
• Use `h.operatorNamespace` and `h.operatorName` in `clusterIngressStatusEqual`.
• Change the equality check in `clusterIngressStatusEqual` to be agnostic to the ordering of conditions and ingresses.
• Change `syncServiceUpdate` to set the ClusterIngress's "Available" condition to "True" as soon as endpoints are ready without checking ingresses if the ClusterIngress's type is not `CloudClusterIngressHA`

pkg/stub/handler.go
```diff
@@ -162,3 +230,149 @@ func getDaemonSetOwnerRef(ds *appsv1.DaemonSet) (metav1.OwnerReference, error) {
...
+// syncServiceUpdate synchronizes the status of the ClusterIngress that
+// corresponds to the service with the service's status.
+func (h *Handler) syncServiceUpdate(service *corev1.Service, deleted bool) (ret error) {
+	if len(service.Name) < 8 || service.Name[0:7] != "router-" {
+		return nil
+	}
+
+	ciName := service.Name[7:]
+
+	ciNamespace, err := k8sutil.GetWatchNamespace()
```
pravisankar: Use h.operatorNamespace instead?

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,137 @@
...
+
+	ci.setStatusCondition(condition)
+}
+
+// clusterIngressStatusEqual returns true if and only if the status conditions
+// (ignoring LastTransitionTime) and ingresses of the provided ClusterIngress
+// objects are equal.
+func clusterIngressStatusEqual(oldCi, ci *ClusterIngress) bool {
+	numConditions := len(ci.Status.Conditions)
+	if numConditions != len(oldCi.Status.Conditions) {
```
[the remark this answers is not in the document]
Miciah: Fixed.

pkg/apis/ingress/v1alpha1/status.go
```diff
@@ -0,0 +1,137 @@
...
+				break
+			}
+		}
+
+		if !foundMatchingCondition {
+			return false
+		}
+	}
+
+	numIngresses := len(ci.Status.LoadBalancer.Ingress)
```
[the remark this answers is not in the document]
Miciah: Fixed.

pkg/stub/handler.go
Miciah: Fixed.  Thanks!

Miciah:
Latest push makes the following changes per Ravi's comments:

• Delete `numConditions` and `numIngresses` in `clusterIngressStatusEqual`.
• Use `h.operatorNamespace` in `syncServiceUpdate`.

I made the following additional changes:

• Added some unit tests in `status_test.go`.
• Renamed `createOrUpdateOperatorStatus` to `syncOperatorStatus`.

After some further consideration, I think we should discuss during the next planning call how the operator should handle the "Available" condition when `ci.Spec.HighAvailability.Type` is not `CloudClusterIngressHA`.

Miciah:
Latest push removes the OperatorStatus code; this PR now depends on #47 to import github.com/openshift/cluster-version-operator, which has some required types.  This push also adds a watch on endpoints and event handling for the same in order to resolve a race condition that would leave the "Available" condition as false with the message "service has no endpoints" when a service's ingresses are ready before its endpoints.

imcsk8:
Looks ok, just rebase and squash.

ironcladlou:
This is probably hopelessly out of date; going to close it and revisit later.

/close

PR #40, opened 2018-09-27
[WIP DO NOT MERGE] Handle update events

imcsk8:
Since the operator sdk does not natively handles update
events we have to check for differences in the ClusterIngress
structure by getting its md5 signatureits

This commit is work in progress does not handle updates on the
deployment and service elements of the ClusterIngress.

Code cleanup

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -28,12 +28,7 @@ type Handler struct {
 	manifestFactory *manifests.Factory
 }
 
-var ci_md5 map[string][16]byte
-
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
-	if ci_md5 == nil {
-		ci_md5 = make(map[string][16]byte)
-	}
 	switch o := event.Object.(type) {
 	case *ingressv1alpha1.ClusterIngress:
 		if event.Deleted {
@@ -128,13 +123,7 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 			err = sdk.Create(service)
 			if err == nil {
 				logrus.Infof("created router service %s/%s", service.Namespace, service.Name)
-			} else if errors.IsAlreadyExists(err) {
-				err = sdk.Update(service)
-				if err != nil {
-					return fmt.Errorf("couldn't update router service %s/%s", service.Namespace, service.Name)
-				}
-				logrus.Infof("updated router service %s/%s", service.Namespace, service.Name)
-			} else {
+			} else if !errors.IsAlreadyExists(err) {
 				return fmt.Errorf("failed to create service %s/%s: %v", service.Namespace, service.Name, err)
 			}
 		}
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -2,6 +2,8 @@ package stub
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/json"
 	"fmt"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


imcsk8:
PTAL @openshift/sig-network-edge @ironcladlou @Miciah @pravisankar @knobunc

ironcladlou:
See [topic 7](https://docs.google.com/document/d/1XwNLgig9BKzptumC80o0Ajl_69W8IbUYQPjdvqPoOsk/edit#) for some discussion on this, as well as the [upstream controller guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) on managing resources with `Generation`. Might be worth having a quick call with @deads2k or somebody in @openshift/sig-master for some concrete advice.

For now, I'd be happy if we just unconditionally re-built the resource and issued a `PATCH` which may no-op. Diffing is going to be a more complex issue.

soltysh:
@ironcladlou @tnozicka will know the most about generations.

pkg/stub/handler.go
```diff
@@ -28,7 +28,12 @@ type Handler struct {
 	manifestFactory *manifests.Factory
 }
 
+var ci_md5 map[string][16]byte
+
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
+	if ci_md5 == nil {
+		ci_md5 = make(map[string][16]byte)
```
pravisankar: Where is this used? probably stale code

pravisankar: Difference in MD5 checksum of the object to denote change in custom resource may not be useful in our case.
Once we populate status field, status is updated periodically and md5 checksum on the object will change even though there is no change in CR spec.
I think we should store old spec (as annotation or something mechanism) and then compare with new spec to determine if update is needed. This mechanism will also provide what exactly has changed and can react accordingly.
We could get more pointers from `oc apply` implementation.

pkg/stub/handler.go
```diff
@@ -97,7 +102,13 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 	err = sdk.Create(ds)
 	if err == nil {
 		logrus.Infof("created router daemonset %s/%s", ds.Namespace, ds.Name)
-	} else if !errors.IsAlreadyExists(err) {
+	} else if errors.IsAlreadyExists(err) {
+		err = sdk.Update(ds)
```
pravisankar: How do we know this is an update event and not resync event? (may be we want to update anyway during resync?)

pkg/stub/handler.go
```diff
@@ -117,7 +128,13 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
 			err = sdk.Create(service)
 			if err == nil {
 				logrus.Infof("created router service %s/%s", service.Namespace, service.Name)
-			} else if !errors.IsAlreadyExists(err) {
+			} else if errors.IsAlreadyExists(err) {
+				err = sdk.Update(service)
```
pravisankar: I don't think we need to update service, none of the spec changes in CR will change service object.

pkg/stub/handler.go
imcsk8: When @Miciah PR gets merged we will know when there's an update. I think this pattern is good for now.

pkg/stub/handler.go
imcsk8: you're right, removing this code

ironcladlou:
Needs redesigned per https://github.com/openshift/cluster-ingress-operator/pull/49; should probably close this one

PR #41, opened 2018-09-28
Changed informer resync period to 10 minutes

pravisankar:
- Old resync period of 5 secs is very aggressive and unnecessary.

### DIFF
--- a/cmd/cluster-ingress-operator/main.go
+++ b/cmd/cluster-ingress-operator/main.go
@@ -31,7 +31,7 @@ func main() {
 	if err != nil {
 		logrus.Fatalf("Failed to get watch namespace: %v", err)
 	}
-	resyncPeriod := 5 * time.Second
+	resyncPeriod := 10 * time.Minute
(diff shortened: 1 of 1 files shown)


pravisankar:

cmd/cluster-ingress-operator/main.go
```diff
@@ -31,7 +31,7 @@ func main() {
 	if err != nil {
 		logrus.Fatalf("Failed to get watch namespace: %v", err)
 	}
-	resyncPeriod := 5 * time.Second
+	resyncPeriod := 10 * time.Minute
```
ironcladlou: I don't think this will work at the current operator-sdk version because `resyncPeriod` unit is expected to be seconds... https://play.golang.org/p/4BKjIShbuYm

Might need to change this to `resyncPeriod := 600` until you update to the latest sdk which accepts a `time.Duration`.

ironcladlou: Disregard, thought I was looking at the cluster-dns-operator repo which is still on the previous version...

PR #43, opened 2018-10-03
Integrate operator with CVO

ironcladlou:


images/cluster-ingress-operator/Dockerfile
```diff
@@ -22,3 +22,5 @@ USER 1001
 LABEL io.k8s.display-name="OpenShift cluster-ingress-operator" \
       io.k8s.description="This is a component of OpenShift Container Platform and manages the lifecycle of cluster ingress components." \
       maintainer="Dan Mace <dmace@redhat.com>"
+
+LABEL io.openshift.release.operator true
```
pravisankar: Will this enable CVO to install cluster-ingress-operator as part of the cluster deployment?
If yes, don't we want some basic CI gating tests for new prs that validates interaction with CVO?

ironcladlou: Yeah, I'm doing testing now and will probably have to make a PR to fix some RBAC stuff

pkg/manifests/manifests_test.go
```diff
@@ -29,30 +29,6 @@ func TestManifests(t *testing.T) {
 		},
 	}
 
-	if _, err := f.OperatorNamespace(); err != nil {
-		t.Fatal(err)
-	}
-	if _, err := f.OperatorCustomResourceDefinition(); err != nil {
-		t.Fatal(err)
-	}
-	if _, err := f.OperatorRole(); err != nil {
```
pravisankar: Looks like we can keep f.OperatorRole() and f.OperatorRoleBinding()?

ironcladlou: We need to revisit RBAC holistically after this

PR #44, opened 2018-10-04
Owner ref for ingress service should only call sdk.Get() on ds if needed.

pravisankar:
- sdk.Create(ds) will populate all the needed info in 'ds'.
- set owner ref instead of append, otherwise we may
end up with same owner ref multiple times.


pravisankar:

PR #45, opened 2018-10-09
Fix installation in latest 4.0 environment

ironcladlou:
A variety of fixes related to admission and scheduling in the latest
4.0 cluster environments.


ironcladlou:
Tested with https://gist.github.com/ironcladlou/d3a7af553130f710a08475adc4096624.

PR #47, opened 2018-10-12
Publish operator status

Miciah:
#### Add openshift/cluster-version-operator dependency

* `Gopkg.lock`:
* `Gopkg.toml`:
* `vendor/github.com/openshift/cluster-version-operator`: Import package in order to have the necessary API types for updating ClusterOperator.

#### Publish operator status

Create (if necessary) and update a ClusterOperator object to indicate the operator's status.

* `cmd/cluster-ingress-operator/main.go`: Add a watch on daemonsets in the application namespace to trigger reconciliation on status updates.
* `manifests/00-cluster-role.yaml`: Allow the operator to create, get, and update clusteroperators.
* `pkg/util/clusteroperator/status.go`: Add new functions: `SetStatusCondition` and `ConditionsEqual`.
* `pkg/util/clusteroperator/status_test.go`: Add tests.
* `pkg/stub/status.go`: Add new functions: `syncOperatorStatus`, `getOperatorState`, and `computeStatusConditions`.
Use `getOperatorState` in `syncOperatorStatus` to get the prerequisite state for computing the operator's status.
Use `computeStatusConditions` in `syncOperatorStatus` to compute the status.
Use `clusteroperator.ConditionsEqual` in `syncOperatorStatus` to determine whether an update is needed.
Use `clusteroperator.SetStatusCondition` in `computeStatusConditions` to compute new conditions.
* `pkg/stub/handler.go`: Use `syncOperatorStatus` in `Handle` to update ClusterOperator.
* `pkg/stub/handler_test.go`: Add tests.


pkg/stub/handler.go
```diff
@@ -16,75 +19,158 @@ import (
...
+
+		cond := osv1.OperatorStatusCondition{
+			Type:    osv1.OperatorStatusConditionTypeWorking,
+			Message: fmt.Sprintf("working towards state: %#v", o),
 		}
+		if err := h.syncOperatorStatus(cond); err != nil {
+			return err
+		}
+
+		if err := h.syncIngressUpdate(o); err != nil {
```
ironcladlou: The `syncOperatorStatus` calls woven throughout the update logic seem more imperative than level driven to me... would it be possible to simply it to "if `syncIngressUpdate` returns an error, set a degraded status"? Or take the error into account when doing a state-observation driven computation unconditionally after the update, like:

1. Attempt update
2. Compute status by observing the state of the cluster after the update
    1. Take any error from (1) into account when computing in (2)

The point being it should be possible to re-compute status outside the context of the update function. That is, there could be a `Handler.ComputeOperatorStatus()` function which could be called at any time to compute the correct status absent any event to observe.

This also closes the current gap with status computation based on deletes: the current code will only handle deletes in response to a delete event (something which needs refactored); if any delete processing (or status processing during delete processing) fails, it should be possible to call a general status recompilation function later on (outside the context of a delete event) and get the right status.

Miciah:
> The `syncOperatorStatus` calls woven throughout the update logic seem more imperative than level driven to me...

Yeah, it's a bit icky.

> would it be possible to simply it to "if `syncIngressUpdate` returns an error, set a degraded status"?

No, not all errors should cause "degraded" status.  In particular, errors from which the operator can  recover without intervention should not cause "degraded" status.

> Or take the error into account when doing a state-observation driven computation unconditionally after the update, like:
> 
> 1. Attempt update
> 2. Compute status by observing the state of the cluster after the update
>   i. Take any error from (1) into account when computing in (2)
>
> The point being it should be possible to re-compute status outside the context of the update function. That is, there could be a `Handler.ComputeOperatorStatus()` function which could be called at any time to compute the correct status absent any event to observe.

This approach is complicated by the fact that `syncIngressUpdate` decorates its errors, which means the error value that it returns does not respond to such introspective methods as `errors.IsForbidden`.  We could stop decorating errors in order to preserve the information needed to decide whether an error were recoverable, use string matching to infer this information, or decorate using richer error types that included that information.  However, the simplest and cleanest approach I can think of would be to add a Boolean return value `retry` and let `Handle` use that value to decide whether to return an error (in order to cause a retry) or log and set "degraded" status.  This approach would still be edge-driven but would be simpler and less imperative.  Does this approach seem reasonable to you?

Maybe it would be helpful to list out error conditions, whether "degraded" status makes sense for each, and how we could otherwise infer "degraded" status from state:

- Error getting the watch namespace or operator name in `NewHandler`: This can only happen on startup, so the operator just errors out, which seems reasonable enough.
- Error building a resource: This could be caused either by some fault or by an erroneous name or selector in the ClusterIngress, either of which would require intervention.  Testing should catch the former, and validation should catch the latter.  Once we add validation, do we even need to handle errors from the manifest factory?
- Error creating, deleting, getting, or updating a resource, which could be any of the following:
  - Timeout, conflict, or other intermittent issue that we know is recoverable, in which case we should retry.
  - "Forbidden", which could be caused by a failure in CVO, a bad manifest, or possibly a misconfigured cluster, any of which requires intervention.  We could use SARs to identify when we will get permissions errors, and maybe that would be a reasonable thing to do on startup, but there is still the possibility that RBAC configuration will change while the operator is running, in which case we still need to handle the errors.
  - "Bad request" or similar error, or some generic error, in which case we really have no clue whether it is recoverable.  Right now we retry on these errors—should we default to assuming errors are unrecoverable?

> This also closes the current gap with status computation based on deletes: the current code will only handle deletes in response to a delete event (something which needs refactored); if any delete processing (or status processing during delete processing) fails, it should be possible to call a general status recompilation function later on (outside the context of a delete event) and get the right status.

In order to identify and remove orphaned resources, we must list them, list ClusterIngress resources, and then reconcile.  To this end, it would be helpful to disentangle the resync logic from event handling.  Ultimately if the overhead of listing everything can be mostly mitigated by informers' caching, we may want to replace the edge-triggered event handling (ClusterIngress was added/modified/deleted, so add/modify/delete the appropriate resources) with edge-triggered reconciliation (some event involving some object in one of our namespaces happened, so get all relevant state and reconcile).  This approach deviates from the pattern we see in operator-sdk and other operators though, and it is beyond the scope of this PR, I believe.

ironcladlou:
@Miciah 
>Maybe it would be helpful to list out error conditions, whether "degraded" status makes sense for each, and how we could otherwise infer "degraded" status from state:>
> ...
> <snip>

It might be useful to decouple the error handling and status discussions entirely. They aren't necessarily related. Error taxonomy as pertains to work queue handling (e.g. is this error transient? Is it recoverable?) is useful but not generally relevant to status computation.

>In order to identify and remove orphaned resources, we must list them, list ClusterIngress resources, and then reconcile. To this end, it would be helpful to disentangle the resync logic from event handling. Ultimately if the overhead of listing everything can be mostly mitigated by informers' caching, we may want to replace the edge-triggered event handling (ClusterIngress was added/modified/deleted, so add/modify/delete the appropriate resources) with edge-triggered reconciliation (some event involving some object in one of our namespaces happened, so get all relevant state and reconcile). This approach deviates from the pattern we see in operator-sdk and other operators though, and it is beyond the scope of this PR, I believe.

*Edge triggered reconciliation* is indeed the approach we should be striving for, and is the only reliable approach to authoring a controller of this nature. We should be able to, in a single thread,

1. Try and reconcile
2. Compute status

Any errors that occurred during step 1 are relevant only indirectly insofar as their side effects on the cluster state which is observed to drive step 2. Categorization/analysis of errors during step 1 may drive things like retries, but the status function shouldn't use errors (a kind of edge event) as inputs.

If that means we need to bite the bullet now and refactor `syncIngressUpdate` and `deleteIngress` out into a single level driven reconcile function, let's just do it and get it over with. I'm afraid of accumulating serious technical debt in the core controller at this stage.

Miciah:
Can you give an example state from which we could infer "degraded" status?

Would it be all right if I removed "degraded" status from this PR (that is, delete all calls to `setOperatorStatusDegraded`) and made a follow-up PR to implement reconciliation? That would enable us to keep moving, fulfill the requirement of publishing OperatorStatus without incurring significant additional technical debt, and then focus on reconciliation.

ironcladlou:
@Miciah 
>Can you give an example state from which we could infer "degraded" status?

For the operator itself, how about something something as simple as a failure to ensure the `openshift-cluster-ingress-router` namespace exists (if we can't create that, then we can't satisfy any ClusterIngress resources at all).

Expressed as a level driven status calc algorithm:

1. There exist more than zero ClusterIngress resources in the `openshift-cluster-ingress-operator` namespace
2. After a reconciliation, there does not exist a `openshift-cluster-ingress-router` namespace

Some similar analysis would be useful for the router ClusterRole.

ironcladlou:
https://github.com/openshift/cluster-ingress-operator/pull/47#issuecomment-430005153 May also blur the line a bit between operator and ClusterIngress status, but it's a start. I think there's a case to be made that the operator should be trying to ensure the router namespace exists during reconciliation independent of any ClusterIngress existing, which may further clarify the boundaries.

Miciah:
What about transient errors from creating the namespace?

Should we implement our own retry loop as part of reconciliation? We could even implement our own work queue à la CVO, but I thought operators generally were supposed to be moving away from that paradigm towards embracing operator-sdk and its patterns.

ironcladlou:
@Miciah 
>What about transient errors from creating the namespace?

>Should we implement our own retry loop as part of reconciliation? We could even implement our own work queue à la CVO, but I thought operators generally were supposed to be moving away from that paradigm towards embracing operator-sdk and its patterns.

The operator-sdk doesn't give us any help for re-queueing yet, [but it soon will](https://github.com/operator-framework/operator-sdk/issues/382). Don't spend too much time working around it. We can assume the happy path for a little while longer. The operator-sdk team has explicitly acknowledged the deficiencies of the current handler paradigm in these regards and has been actively working on refactoring to use the [controller-runtime](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler) upstream framework as the underlying basis, which will provide us with first-class reconciliation features including retry/work queue control.

Keep in mind the operator-sdk (at the version we started with) doesn't necessarily represent the defacto best practices for k8s controller design and the project is evolving to catch up. We must avoid at all costs going to prod with an edge-driven design: that particular mistake was extremely costly to repair in the original OpenShift v3 controllers.

deads2k:
/hold

This appears to be setting an older version of status that has been superseded by https://github.com/openshift/cluster-version-operator/pull/31

Miciah:
Latest push is a rewrite to be level-driven and to use the new ClusterOperator API.  Labeled "WIP" because I have not been able to test it (although most new code has unit test coverage).

pkg/stub/handler.go
```diff
@@ -247,3 +255,179 @@ func (h *Handler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) error
 	}
 	return nil
 }
+
+// syncOperatorStatus computes the operator's current status and therefrom
+// creates or updates the ClusterOperator resource for the operator.
+func (h *Handler) syncOperatorStatus() {
```
ironcladlou: Maybe move all this stuff and the new stuff in util into a new file (`status.go`)?

pkg/stub/handler.go
```diff
@@ -247,3 +255,179 @@ func (h *Handler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) error
...
+
+	if err := sdk.Update(co); err != nil {
+		logrus.Errorf("syncOperatorStatus: failed to update status of ClusterOperator %s/%s: %v",
+			co.Namespace, co.Name, err)
+	}
+}
+
+// getOperatorState gets and returns the resources necessary to compute the
+// operator's current state.
+func (h *Handler) getOperatorState() (*corev1.Namespace, *ingressv1alpha1.ClusterIngressList, *appsv1.DaemonSetList, error) {
```
ironcladlou: How about returning plain arrays (e.g. `[]ClusterIngress`) instead? The list wrappers are cumbersome to work with and provide no additional value in this case

pkg/stub/handler.go
```diff
@@ -247,3 +255,179 @@ func (h *Handler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) error
...
+		Status: osv1.ConditionUnknown,
+	}
+	if ns == nil {
+		failingCondition.Status = osv1.ConditionTrue
+		failingCondition.Reason = "NoNamespace"
+		failingCondition.Message = "router namespace does not exist"
+	} else {
+		failingCondition.Status = osv1.ConditionFalse
+	}
+	clusteroperator.SetStatusCondition(co, failingCondition)
```
ironcladlou: Style issue: mutating input is troublesome. This function is close to stateless; how about just returning the new Status object based on immutable inputs?

pkg/stub/handler.go
```diff
@@ -247,3 +255,179 @@ func (h *Handler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) error
...
+				numDaemonsets, numIngresses)
+		}
+	}
+	clusteroperator.SetStatusCondition(co, progressingCondition)
+
+	availableCondition := &osv1.ClusterOperatorStatusCondition{
+		Type:   osv1.OperatorAvailable,
+		Status: osv1.ConditionUnknown,
+	}
+	if ingressList != nil && daemonsetList != nil {
```
ironcladlou: I'm not sure this correlation is correct, the DaemonSet name is a function of (not equal to) the ClusterIngress name... also, I think you have to compare availability of daemonsets relative to their own status: e.g.:

* for each ingress
  * get daemonset by ingress
    * let fullyAvailable = (daemonset.desired == daemonset.ready)

I think a more natural way to do this sort of stuff would be to build up some indexed caches from the k8s stdlib. For example:

* DaemonSet cache with index:
  * by ClusterIngress.Name where .Status.Desired == .Status.Ready

Then given a ClusterIngress you can do an O(1) lookup to see if there exists a ready DaemonSet associated with the ClusterIngress.

ironcladlou: Maybe it would be better to start small and slightly incorrect and build up to unblock deployment? For example, if the count of daemonsets matches the count of ingresses, at least mark us as "Progressing" and not failed, and we can refine the algorithms in followups using strategies like in 
https://github.com/openshift/cluster-ingress-operator/pull/47/commits/5a46915b18c90fd4883fb95d7672607fef25c4e8#r226467331.

Would that help?

Miciah: > I'm not sure this correlation is correct, the DaemonSet name is a function of (not equal to) the ClusterIngress name...

Fixed (I think).  Thanks!

> also, I think you have to compare availability of daemonsets relative to their own status: e.g.:
> 
> * for each ingress
>   * get daemonset by ingress
>     * let fullyAvailable = (daemonset.desired == daemonset.ready)

What is `daemonset.desired`, and what is `daemonset.ready`? I chose to use `ds.Status.NumberAvailable` based on my reading the docstrings for `OperatorAvailable` and `DaemonSetStatus` and a literal interpretation of "available".  If my interpretation is not correct, we should specify what we mean by "available".

> I think a more natural way to do this sort of stuff would be to build up some indexed caches from the k8s stdlib.

I've been trying to get by using operator-sdk.  Presumably its client API will grow richer, so I don't want to move away from it prematurely.

Miciah:
Latest push renames `setStatusConditions` to `computeStatusConditions` and makes it functional, changes `getOperatorState` and `computeStatusConditions` to use slices instead of list types, moves `syncOperatorStatus`, `getOperatorState`, and `computeStatusConditions` to `pkg/stub/status.go`, makes `SetStatusCondition` functional, renames `StatusEqual` to `ConditionsEqual` and makes it work on slices of conditions instead of on statuses, and fixes the mapping of ClusterIngress name to DaemonSet name.

pkg/stub/handler.go
Miciah: Fixed.

pkg/stub/handler.go
Miciah: Fixed.  My only concern is that I don't want to copy large structures accidentally; hopefully I've avoided doing so...

pkg/stub/handler.go
ironcladlou: Available is probably right, I didn't look closely at the docs for those fields.

Regarding indexers, no need to go outside the operator-sdk. [Check this out](https://github.com/kubernetes/client-go/blob/master/informers/apps/v1/daemonset.go#L50). Given just a `kubernetes.Interface` (easily available at our entrypoint in `main.go`) you can create any number of indexes to pass along to the handler.

There are a lot of examples to help you prototype in unit tests, which should help.

* [shared_informer_test.go](https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer_test.go)
* [index_test.go](https://github.com/kubernetes/client-go/blob/master/tools/cache/index_test.go)
* [controller_test.go](https://github.com/kubernetes/client-go/blob/master/tools/cache/controller_test.go) 

Making liberal use of caches/indexes/appropriate data structures is going to be very useful here to avoid an ever-expanding rat's nest of imperative analysis code.

ironcladlou:
Threw together a quick contextual example of what I'm talking about:

```go
package stub

import (
	"testing"

	"k8s.io/client-go/tools/cache"

	appsv1 "k8s.io/api/apps/v1"

)

func TestComputeOperatorStatus(t *testing.T) {
	byStatus := func(obj interface{}) ([]string, error) {
		ds := obj.(*appsv1.DaemonSet)
		if ds.Status.NumberUnavailable == 0 {
			return []string{"available"}, nil
		}
		return []string{"unavailable"}, nil
	}

	mkds := func(name string, unavailable int32) *appsv1.DaemonSet {
		return &appsv1.DaemonSet{ObjectMeta: metav1.ObjectMeta{Name: name}, Status: appsv1.DaemonSetStatus{NumberUnavailable: unavailable}}
	}

	index := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{"byStatus": byStatus})

	index.Add(mkds("a", 0))
	index.Add(mkds("b", 0))
	index.Add(mkds("c", 1))
	index.Add(mkds("d", 2))
	index.Add(mkds("e", 1))

	unavailable, err := index.ByIndex("byStatus", "unavailable")
	if err != nil {
		t.Fatal(err)
	}
	if e, a := 3, len(unavailable); e != a {
		t.Errorf("expected %d unavailable, got %d", e, a)
	}
}
```

Lots of places to go from there to answer different questions. Status updates could launched in another goroutine with a pretty frequent resync period provided they remain cache-driven.

Miciah:
> Regarding indexers, no need to go outside the operator-sdk. [Check this out](https://github.com/kubernetes/client-go/blob/master/informers/apps/v1/daemonset.go#L50). Given just a `kubernetes.Interface` (easily available at our entrypoint in `main.go`) you can create any number of indexes to pass along to the handler.

You mean using `k8sclient.GetKubeClient()`? Looks like k8sclient is gone upstream, so we will need to change to the controller-runtime client eventually; do we have plans for that work?

> There are a lot of examples to help you prototype in unit tests, which should help.

I reviewed the examples, thanks!

> Making liberal use of caches/indexes/appropriate data structures is going to be very useful here to avoid an ever-expanding rat's nest of imperative analysis code.

Not sure it makes the code simpler, but I can give it a try.  Note that operator-sdk seems to be moving towards using controller-runtime and promoting use of the split client, which would give us caching for free, which means that using an indexer would probably not reduce API usage, although it could reduce CPU usage or LOC (although I'm skeptical about that).

ironcladlou:
David has warned us against relying on Indexer going forward; apparently it has consistency issues and has bitten us before in ways he can elaborate on if we're interested. His recommendation was to run filters on lists, not far from what you were doing.

pkg/stub/status.go
```diff
@@ -0,0 +1,200 @@
...
+		sdk.WithListOptions(&metav1.ListOptions{}))
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf(
+			"failed to list ClusterIngresses: %v", err)
+	}
+
+	daemonsetList := &appsv1.DaemonSetList{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "DaemonSet",
+			APIVersion: "extensions/v1beta1",
```
ironcladlou: This should be `apps/v1`

pkg/stub/status.go
```diff
@@ -0,0 +1,200 @@
...
+		failingCondition.Status = osv1.ConditionFalse
+	}
+	conditions = clusteroperator.SetStatusCondition(conditions,
+		failingCondition)
+
+	progressingCondition := &osv1.ClusterOperatorStatusCondition{
+		Type:   osv1.OperatorProgressing,
+		Status: osv1.ConditionUnknown,
+	}
+	if ingresses != nil && daemonsets != nil {
```
ironcladlou: I think this should always be computed; remove the `nil` checks (`len` is safe on all arrays, but let's consider passing `nil` arrays an error).

pkg/stub/status.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+	conditions = clusteroperator.SetStatusCondition(conditions,
+		progressingCondition)
+
+	availableCondition := &osv1.ClusterOperatorStatusCondition{
+		Type:   osv1.OperatorAvailable,
+		Status: osv1.ConditionUnknown,
+	}
+	if ingresses != nil && daemonsets != nil {
```
ironcladlou: I think this should always be computed; remove the `nil` checks (`len` is safe on all arrays, but let's consider passing `nil` arrays an error).

pkg/stub/status.go
```diff
@@ -0,0 +1,200 @@
...
+		dsAvailable := map[string]bool{}
+		for _, ds := range daemonsets {
+			dsAvailable[ds.Name] = ds.Status.NumberAvailable > 0
+		}
+
+		unavailable := []string{}
+		for _, ingress := range ingresses {
```
ironcladlou: Make a TODO here to revisit this block... the association logic needs cleaned up (e.g. using a function in `manifests` get associated resource names, or labels, or owner references, or something).

pkg/stub/status.go
```diff
@@ -0,0 +1,200 @@
...
+					ingress.Name)
+				unavailable = append(unavailable, msg)
+			}
+		}
+
+		if len(unavailable) == 0 {
+			availableCondition.Status = osv1.ConditionTrue
+		} else {
+			availableCondition.Status = osv1.ConditionFalse
+			availableCondition.Reason = "IngressUnavailable"
```
ironcladlou: Could this be inverted to `IngressAvailable` to avoid double-negatives?

pkg/util/clusteroperator/status.go
```diff
@@ -0,0 +1,65 @@
+package clusteroperator
```
ironcladlou: No upstream utils for this stuff?

Miciah: I found https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/lib/resourcemerge/os.go#L36 which could replace `SetStatusCondition`.  Is that something we can use (from a packaging and stability perspective)?

pkg/stub/status.go
Miciah: The `OperatorAvailable` condition comes from the CVO API definitions, so we cannot invert it.

pkg/util/clusteroperator/status.go
ironcladlou: Maybe worth a shot assuming transitive deps don't become an issue... duplicating these functions across our operators will be annoying but not the end of the world for now

Miciah:
Latest push changes the API version specified for listing DaemonSets from "extensions/v1beta1" to "apps/v1"; deletes unnecessary nil checks on slices in `computeStatusConditions`; adds a TODO to use manifests, labels, or owner references for associate ingresses with routers; renames `handler_test.go` to `status_test.go` (which I forgot to do when splitting `status.go` out of `handler.go`), and updates tests and bindata.

pkg/stub/status.go
Miciah: Fixed.  Thanks!

pkg/stub/status.go
Miciah: Deleted.

pkg/stub/status.go
Miciah: Deleted.

pkg/stub/status.go
Miciah: Added.

Miciah:
Rebased.

pkg/stub/status_test.go
```diff
@@ -0,0 +1,393 @@
...
+	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
+	"github.com/openshift/cluster-ingress-operator/pkg/util/clusteroperator"
+	osv1 "github.com/openshift/cluster-version-operator/pkg/apis/operatorstatus.openshift.io/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSetStatusConditions(t *testing.T) {
```
ironcladlou: 1. Given inputs, the output should be some expected `[]ClusterOperatorStatusCondition`. Existing condition isn't an input to the status computation function. When would `old` ever be relevant?
2. What about doing a few tweaks to allow expressing the cases as one-liners for easier comprehension?

imcsk8:
Looks good, there's a lot of stuff that could be shared between operators

pkg/stub/status_test.go
Miciah: > Existing condition isn't an input to the status computation function. When would old ever be relevant?

It is, and I think it needs to be if for no other reason so that we update `LastTransitionTime` if and only if it appropriate to do so (although the unit tests ignore `LastTransitionTime` presently).  As far as testing, having the old conditions lets us test to make sure `SetStatusCondition` doesn't add duplicate or fail to update an existing condition, for example. 

> What about doing a few tweaks to allow expressing the cases as one-liners for easier comprehension?

Maybe something like this?
```go
type testInputs struct {
  numWanted, numAvailable, numMissing, numUnavailable int
}
type testOutputs struct {
  failing, progressing, available bool
}
// ...
  testCases := []struct {
    description string
    haveNamespace bool
    in testInputs
    out testOutputs
  }{
    { "1/2 ingresses available", testInputs{2, 1, 0, 1}, testOutputs{false, true, false} },
    // ...
```
and ignore condition reason and message? (That example is >80 characters but technically a one-liner...)

ironcladlou: >>Existing condition isn't an input to the status computation function. When would old ever be relevant?

>It is, and I think it needs to be if for no other reason so that we update LastTransitionTime if and only if it appropriate to do so (although the unit tests ignore LastTransitionTime presently). 

I guess it could make sense in this case. I'm not sure how important that is.

>As far as testing, having the old conditions lets us test to make sure SetStatusCondition doesn't add duplicate or fail to update an existing condition, for example.

Assuming our operator owns that field, whenever the operator wants to update the field it could atomically replace the entire field with newly computed contents. Adding complexity to diff and patch seems like overkill.

pkg/stub/status.go
```diff
@@ -0,0 +1,196 @@
...
+	conditions = clusteroperator.SetStatusCondition(conditions,
+		failingCondition)
+
+	progressingCondition := &osv1.ClusterOperatorStatusCondition{
+		Type:   osv1.OperatorProgressing,
+		Status: osv1.ConditionUnknown,
+	}
+	numIngresses := len(ingresses)
+	numDaemonsets := len(daemonsets)
+	if numIngresses == numDaemonsets {
```
pravisankar: When an Ingress CR is deleted, corresponding daemonset may not be deleted yet (pending op) and the length of numIngresses and numDaemonsets will defer, don't we want to say OperatorProgressing = True in this case?

pkg/util/clusteroperator/status.go
Miciah: Transitive deps are becoming an issue.  Seems like importing "github.com/openshift/cluster-version-operator/lib/resourcemerge" requires vendoring "k8s.io/kube-aggregator" version 1.11.1 (CVO needs APIs that are not in 1.10.1), and pulling in kube-aggregator 1.11.1 will require updating operator-sdk, which has substantial API changes.

pkg/stub/status_test.go
Miciah: I implemented something very similar to what I outlined in my last comment.

Miciah:
Latest push renames `TestSetStatusCondition` to `TestComputeStatusConditions`, simplifies the test cases therein, and adds `TestSetStatusCondition` to `pkg/util/clusteroperator/status_test.go` to make up for coverage lost in the `TestComputeStatusConditions` simplification.

pkg/stub/status_test.go
Miciah: Just saw your last two comments.  I like using additional types such as your `daemonset` type for clarity, but I don't think it matters in the test case how we name the daemonsets or what specific value (besides zero or non-zero) we use for `NumberAvailable`.

Gopkg.toml
```diff
@@ -36,3 +36,7 @@ required = [
 [[constraint]]
   name = "github.com/openshift/api"
   branch = "master"
+
+[[constraint]]
+  name = "github.com/openshift/cluster-version-operator"
+  revision = "fe673cb712fa5e27001488fc088ac91bb553353d"
```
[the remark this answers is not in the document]
Miciah: I didn't see any warning, but I think I accidentally added the newline anyway.

Miciah:
Latest push adds a TODO to use [cluster-version-operator's `SetOperatorStatusCondition`](https://github.com/openshift/cluster-version-operator/blob/fe673cb712fa5e27001488fc088ac91bb553353d/lib/resourcemerge/os.go#L36-L54) once we update to a version of cluster-version-operator that has it.

ironcladlou:
Just needs tested against an e2e cluster (using `hack/local-release.sh` and `make test-integration`, etc.)

Miciah:
Latest push adds a watch on daemonsets in the "openshift-ingress" namespace so that status updates to daemonsets trigger status updates to clusteroperators, fixes the namespace that `getOperatorState` uses to list daemonsets, and changes `%s` to `%q` for interpolating ingress name into status. 

Integration tests are blocked on https://github.com/openshift/installer/issues/530.

Miciah:
Dan and I discussed testing of this PR, and we concluded that it is sufficient to test it manually using `make release-local`, `hack/uninstall.sh`, and `oc apply /tmp/...` and verifying that the operator runs and creates the service and daemonset correctly.  I have done this manual testing, so this PR is good to merge.

ironcladlou:
Great work, thanks.

/lgtm

imcsk8:
LGTM

PR #48, opened 2018-10-15
 Handle ingress domain updates to default cluster ingress

pravisankar:
- Delete duplicate finalizers section in default cluster ingress cr.
- Current code fetches needed info from installer configmap and tries to create default cluster ingress cr and ignores in case of already exists error. It won't handle the case where default cluster ingress cr exists but the ingress domain got changed. This pr handles this case.


pravisankar:

pkg/stub/handler.go
```diff
@@ -62,16 +62,48 @@ func (h *Handler) EnsureDefaultClusterIngress() error {
...
+		},
+	}
+	err := sdk.Get(oldci)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, false, fmt.Errorf("failed to fetch existing default cluster ingress %s/%s, %v", ci.Namespace, ci.Name, err)
+		}
+		return false, true, nil
+	}
+	if oldci.Spec.IngressDomain != ci.Spec.IngressDomain {
```
ironcladlou: I think you need to deference the pointers to do what you mean here: https://play.golang.org/p/N1-AT4yXgH-

pravisankar: yes, thanks

pravisankar:
@ironcladlou updated, ptal

ironcladlou:
How was it tested?

pravisankar:
Tested both create and update cases on 4.0 cluster
/hold cancel

pravisankar:

pkg/stub/handler.go
```diff
@@ -62,16 +62,49 @@ func (h *Handler) EnsureDefaultClusterIngress() error {
...
+		},
+	}
+	err := sdk.Get(oldci)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, nil, fmt.Errorf("failed to fetch existing default cluster ingress %s/%s, %v", ci.Namespace, ci.Name, err)
+		}
+		return false, nil, nil
+	}
+	if *oldci.Spec.IngressDomain != *ci.Spec.IngressDomain {
```
ironcladlou: If either of this is `nil` will the function panic?

pravisankar: Let me fix this

ironcladlou:
Would like to test this once https://github.com/openshift/cluster-ingress-operator/pull/51 lands

pkg/stub/handler.go
```diff
@@ -69,16 +69,56 @@ func (h *Handler) EnsureDefaultClusterIngress() error {
 		return err
 	}
 
-	err = sdk.Create(ci)
-	if err == nil || errors.IsAlreadyExists(err) {
-		if err == nil {
-			logrus.Infof("created default cluster ingress %s/%s", ci.Namespace, ci.Name)
+	changed, nci, err := checkClusterIngress(ci)
+	if err != nil {
```
imcsk8: I think this can use: clusterIngressStatusEqual
from PR: #38

Miciah: Maybe not `clusterIngressStatusEqual`, but the part that compares `.loadBalancer.Ingress` could be factored into a function that could be used both by `checkClusterIngress` and by `clusterIngressStatusEqual`.  Now that #47 is merged, I'll rebase #38 Tuesday and factor out a `LoadBalancerStatusEqual` function.  No need for that to block this PR though; I can change `checkClusterIngress` to use `LoadBalancerStatusEqual` if #48 merges before #38 is ready.

Miciah: Err, actually, `checkClusterIngress` is comparing only `.spec.ingressDomain` (a string) whereas `clusterIngressStatusEqual` is comparing `.status.loadBalancer` (a slice of a struct that has two string values), so there is really no opportunity for code re-use here as far as I see.

imcsk8: I might be confused but if the LoadBalancerIngress struct are different as being checked in your PR [1]  it means there's an update it should be enough for this purpouse.

[1] https://github.com/openshift/cluster-ingress-operator/pull/38/files#diff-9ae69f8dcadd8df44e0b10265f3a91a8R107

ironcladlou:
Superseded by https://github.com/openshift/cluster-ingress-operator/pull/48

Miciah:
> Superseded by #48

For the record, I think that was supposed to be #89.

PR #49, opened 2018-10-16
Refactor to level driven design

ironcladlou:
Refactor the handler to be level driven.

### DIFF
--- a/pkg/stub/handler.go
+++ b/pkg/stub/handler.go
@@ -14,6 +14,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 const (
@@ -24,27 +25,26 @@ const (
 	clusterConfigResource = "cluster-config-v1"
 )
 
-func NewHandler() *Handler {
+func NewHandler(namespace string, manifestFactory *manifests.Factory) *Handler {
 	return &Handler{
-		manifestFactory: manifests.NewFactory(),
+		namespace:       namespace,
+		manifestFactory: manifestFactory,
 	}
 }
 
 type Handler struct {
+	namespace       string
 	manifestFactory *manifests.Factory
 }
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
+	// TODO: This should be adding an item to a rate limited work queue, but for
+	// now correctness is more important than performance.
 	switch o := event.Object.(type) {
 	case *ingressv1alpha1.ClusterIngress:
-		if event.Deleted {
-			logrus.Infof("Deleting ClusterIngress object: %s", o.Name)
-			return h.deleteIngress(o)
-		} else {
-			return h.syncIngressUpdate(o)
-		}
+		logrus.Infof("reconciling for update to clusteringress %q", o.Name)
 	}
-	return nil
+	return h.reconcile()
 }
 
 // EnsureDefaultClusterIngress ensures that a default ClusterIngress exists.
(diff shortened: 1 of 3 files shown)


ironcladlou:
Holy cow, not sure why the dep update dragged in so much new junk. I want to revert that and try again.

ironcladlou:
Okay, I decided to copy and paste a slightly modified `slice.RemoveString` rather than dragging in all of Kubernetes. 😁

ironcladlou:
Tested end-to-end in a 4.0 AWS cluster. Ensured defaulting and delete handling works.

imcsk8:
nice! 
it feels more on the operators reconciliation pattern

pkg/stub/handler.go
```diff
@@ -24,27 +25,26 @@ const (
...
-		if event.Deleted {
-			logrus.Infof("Deleting ClusterIngress object: %s", o.Name)
-			return h.deleteIngress(o)
-		} else {
-			return h.syncIngressUpdate(o)
-		}
+		logrus.Infof("reconciling for update to clusteringress %q", o.Name)
 	}
-	return nil
+	return h.reconcile()
```
pravisankar: perform h.reconcole() only when we are dealing with ingressv1alpha1.ClusterIngress? This seems to call this reconciler for non cluster ingress objects.

pkg/stub/handler.go
```diff
@@ -74,7 +74,74 @@ func (h *Handler) EnsureDefaultClusterIngress() error {
...
+		if ingress.DeletionTimestamp != nil {
+			// Destroy any router associated with the clusteringress.
+			err := h.ensureRouterDeleted(&ingress)
+			if err != nil {
+				errors = append(errors, fmt.Errorf("couldn't delete clusteringress %q: %v", ingress.Name, err))
+				continue
+			}
+			// Clean up the finalizer to allow the clusteringress to be deleted.
+			ingress.Finalizers = RemoveString(ingress.Finalizers, "ingress.openshift.io/default-cluster-ingress")
+			err = sdk.Update(&ingress)
```
pravisankar: Once we update ingress object, don't we see another event for the same object? if yes, that will have DeletionTimestamp but no finalizers section?

pkg/stub/handler.go
```diff
@@ -74,7 +74,74 @@ func (h *Handler) EnsureDefaultClusterIngress() error {
...
 }
 
-func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
+// Reconcile performs a full reconciliation loop for ingress, including
+// generalized setup and handling of all clusteringress resources in the
+// operator namespace.
+func (h *Handler) reconcile() error {
+	// Ensure we have all the necessary scaffolding on which to place router
+	// instances.
+	err := h.ensureRouterNamespace()
```
pravisankar: Do we really want to do this for every event? Seems like one time op.

pkg/stub/handler.go
```diff
@@ -163,10 +225,34 @@ func (h *Handler) syncIngressUpdate(ci *ingressv1alpha1.ClusterIngress) error {
...
+	err = sdk.Delete(ds)
+	if !errors.IsNotFound(err) {
+		return err
+	}
+	return nil
+}
+
+// RemoveString returns a newly created []string that contains all items from slice that
+// are not equal to s.
+func RemoveString(slice []string, s string) []string {
```
pravisankar: removeString() if we don't expect to be used in other files.

pkg/stub/handler.go
```diff
@@ -24,27 +25,26 @@ const (
...
+		manifestFactory: manifestFactory,
 	}
 }
 
 type Handler struct {
+	namespace       string
 	manifestFactory *manifests.Factory
 }
 
 func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
```
pravisankar: What does ctx hold with this new scheme (not useful at this time)?

pkg/stub/handler.go
ironcladlou: It'll trigger for any object we're actively watching in our entry point, which for now is just ClusterIngress, but in the future should be other types of resources that we might want to trigger a reconciliation.

pkg/stub/handler.go
ironcladlou: Yes, and although that still works now, we're doing an additional pointless update. So I think we're at least still correct/reentrant/idempotent... I'll fix

pkg/stub/handler.go
ironcladlou: Probably not, but it seems advantageous to keep it simple for now. We could do a lookup before bothering to try creating the namespace rather than relying on a superfluous create attempt.

pkg/stub/handler.go
ironcladlou: Need to look at operator-sdk code, not sure. Nothing important for now AFAIK.

pkg/stub/handler.go
ironcladlou: I wish the upstream `slice` package were in a better place so we could use it and some other nice stuff I'll have to copy in 😞

pkg/stub/handler.go
ironcladlou: Fixed

Miciah:
I think `ensureRouterForIngress` should not get the ClusterIngress resource if it already exists because doing so means reconciliation will use the old resource's `.spec` instead of the new one's when creating the Service resource.

pkg/stub/handler.go
Miciah: Why do we log conditionally and reconcile unconditionally?

ironcladlou: No reason, logging can be tweaked

ironcladlou:
>I think ensureRouterForIngress should not get the ClusterIngress resource if it already exists because doing so means reconciliation will use the old resource's .spec instead of the new one's when creating the Service resource.

What line is this in reference to?

Miciah:
> > I think ensureRouterForIngress should not get the ClusterIngress resource if it already exists because doing so means reconciliation will use the old resource's .spec instead of the new one's when creating the Service resource.
> 
> 
> What line is this in reference to?

Line 199.  I would have commented inline but its outside of the diff context.

PR #53, opened 2018-10-18
Read default values from install config

ironcladlou:
The kco config is gone forever; get default values from the install config
instead.


ironcladlou:
Still need to test.

/hold

ironcladlou:
I don't have a clean integration test run yet because ELBs aren't being created for the LB service in `us-west-1` for some reason, but I'm pretty confident everything's working otherwise.

/hold cancel

ironcladlou:
e2e fix is here: https://github.com/openshift/installer/pull/496

PR #55, opened 2018-10-25
Move Dockerfile for product builds

ironcladlou:
Move the Dockerfile back to root since the product build requires it.


ironcladlou:
Depends on https://github.com/openshift/release/pull/2023/files

knobunc:
/lgtm
Thanks Dan

PR #58, opened 2018-11-05
 Do no generate bindata for manifests used by CVO

pravisankar:
Also fix gofmt for pkg/util/clusteroperator/status_test.go


pravisankar:
@openshift/sig-network-edge @ironcladlou PTAL

ironcladlou:
Thanks!

/lgtm

PR #59, opened 2018-11-09
Manage routing wildcard domain

ironcladlou:
Add routing wildcard DNS management for AWS. This should be replaced ASAP by something more generalized (e.g. [external-dns](https://github.com/kubernetes-incubator/external-dns)).

The records are managed in a way that's compatible with the installer and are correctly cleaned up during a cluster destroy.

*Does not yet handle updates or deletes.*


pkg/manifests/manifests.go
```diff
@@ -49,7 +49,7 @@ func (f *Factory) DefaultClusterIngress(ic *util.InstallConfig) (*ingressv1alpha
 	if err != nil {
 		return nil, err
 	}
-	ingressDomain := fmt.Sprintf("%s.%s", ic.Metadata.Name, ic.BaseDomain)
+	ingressDomain := fmt.Sprintf("*.apps.%s.%s", ic.Metadata.Name, ic.BaseDomain)
```
Miciah: This will break `TestDefaultClusterIngress`.  Also, do we want to publish the "*." in the canonical name in routes?

smarterclayton: No, canonical name is the name of the DNS wildcard root domain and API consumers should assume that it is `("%s.%s", myName, canonicalHostname)`.  I.e. canonical should always be reported as `a.b.c` and consumers must assume their name is `d.a.b.c`

Miciah: So we should add the "*." in `ensureDNSForLoadBalancer`.  Not sure about the "apps." though—probably should add "apps." here and adjust `TestDefaultClusterIngress` to keep it passing.

ironcladlou:
Extracted https://github.com/openshift/cluster-ingress-operator/pull/60 and https://github.com/openshift/cluster-ingress-operator/pull/61. Will rebase this soon.

pkg/manifests/manifests.go
ironcladlou: We don't yet have any cluster config for the app domain, so I'm just making things up here. Do agree that the `*.` should be prefixed at runtime in `ensureDNSForLoadBalancer()`.

ironcladlou:
@smarterclayton @abhinavdahiya thoughts on how we can approach hosted zone config/discovery here? We need to know:

1. The zone in which the alias should live
2. The zone of the alias target

I'm not entirely clear how 1 is managed in the first place, and for 2 all we have to start with is the hostname attached to the Service status.

abhinavdahiya:
here's how hive deletes r53 zones https://github.com/openshift/hive/tree/master/contrib/pkg/awstagdeprovision#L1366-L1421

smarterclayton:
It should be the same zone as the apiserver masters unless we have a really
clever use case.  Or we just make the installer create it.

On Nov 9, 2018, at 4:42 PM, Dan Mace <notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton> @abhinavdahiya
<https://github.com/abhinavdahiya> thoughts on how we can approach hosted
zone config/discovery here? We need to know:

   1. The zone in which the alias should live
   2. The zone of the alias target

I'm not entirely clear how 1 is managed in the first place, and for 2 all
we have to start with is the hostname attached to the Service status.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/cluster-ingress-operator/pull/59#issuecomment-437504119>,
or mute the thread

ironcladlou:
Did some more work on this to auto-discover all the inputs... probably worth another review at this point before I go any further. PTAL @openshift/sig-network-edge @smarterclayton

pkg/stub/handler.go
```diff
@@ -242,6 +242,12 @@ func (h *Handler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) err
 				} else if !errors.IsAlreadyExists(err) {
 					return fmt.Errorf("failed to create router service %s/%s: %v", service.Namespace, service.Name, err)
 				}
+				if ci.Spec.IngressDomain != nil {
+					err = h.ensureDNSForLoadBalancer(ci, service)
```
smarterclayton: we definitely need to think about how to rate limit how often this can be ensured (i.e. should we have a rate limiter that requeues us when the limit expires, should we have a periodic timer, etc).

pkg/stub/dns.go
```diff
@@ -0,0 +1,134 @@
...
+		},
+	}
+	err := sdk.Get(awsCreds)
+	if err != nil {
+		return fmt.Errorf("failed to get aws creds from %s/%s: %v", awsCreds.Namespace, awsCreds.Name, err)
+	}
+	creds := credentials.NewStaticCredentials(string(awsCreds.Data["aws_access_key_id"]), string(awsCreds.Data["aws_secret_access_key"]), "")
+	sess, err := session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(h.InstallConfig.Platform.AWS.Region),
```
smarterclayton: TODO here to pull this from the Infrastructure config, not InstallConfig.

joelddiaz:
The private zones do get some tagging applied to them (eg tectonicClusterID)... Not sure how helpful that is (or whether we want to embed this kind of tag lookup in the operator/controllers).

I'd feel more comfortable with the tags if the hosted zones were being tagged with the kubernetes.io/cluster/$clustername tag. There's an old discussion on this topic here https://github.com/openshift/installer/issues/458 

The uninstall code can also be updated to not require the matching private zone, but the question then becomes how would it unambiguously identify the records that need removal in the shared public zone?

ironcladlou:
Thanks, @joelddiaz. I didn't realize the zone was tagged. I think lookup using the `tectonicClusterID` tag on the private zone and creating a private zone record will have to be good enough for now.

ironcladlou:
@openshift/sig-network-edge @smarterclayton okay, I think this is ready for near final review... I've been using some uncommitted tests for it; we probably need some e2e coverage.

Overall I'm extremely uncomfortable maintaining this code and it's probably unsustainable beyond like 2 platforms (aws/gcp), but I'm hoping it's good enough to keep us moving...

ironcladlou:
I added some cheesy logic to cache updates operator-side to stop spamming AWS with UPSERT calls. Can come up with a better way later.

ironcladlou:
@smarterclayton 
```
error: error reading /tmp/cluster/telemeter-token: no such file or directory
error: no objects passed to apply
deployment.apps/telemeter-client created
/usr/bin/openshift-tests
which: no extended.test in (/usr/libexec/origin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
must provide TEST_SUITE variable
```
Looks like something's up with our job

ironcladlou:
@csrwng you might want to review what we're doing here as part of the external-dns use case analysis.

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+		return fmt.Errorf("failed to find public zone for cluster")
+	}
+	if privateZone == nil {
+		return fmt.Errorf("failed to find private zone for cluster")
+	}
+
+	// Find the target hosted zone of the load balancer attached to the service.
+	// TODO: cache it?
+	var targetHostedZoneID string
+	loadBalancers, err := m.ELB.DescribeLoadBalancers(&elb.DescribeLoadBalancersInput{})
```
csrwng: You'll need the elbv2 version of this call to catch the new network load balancers we're using

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to list hosted zones: %v", err)
+	}
+
+	m.zones = zones
+	return nil
+}
+
+// findClusterPublicZone finds the public zone given the base domain.
+func (m *Manager) findClusterPublicZone() *route53.HostedZone {
```
csrwng: I was hoping to send something like this upstream to external-dns ... add the ability to lookup zones by tags. In the meantime, the code will live in the operator.

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	}
+	if len(targetHostedZoneID) == 0 {
+		return fmt.Errorf("couldn't find hosted zone ID of target ELB with domain name %s", target)
+	}
+
+	// Create or update an alias from the wildcard domain to the service load
+	// balancer hostname in both zones.
+	// TODO: only make this call if the records don't exist.
+	updateRecord := func(zoneID string) error {
```
csrwng: This is the part that external-dns can do for you right now. Btw, if I set the --registry=noop flag, I don't have it create the ugly TXT records.

pkg/dns/aws/dns.go
ironcladlou: Can you explain further? I'm having no issues w/ the ELBs created by the k8s cloud provider for our services.

ironcladlou:
P.S., current code won't handle deletes or updates to the ClusterIngress. Can address that in a followup; need to get some sort of e2e coverage on this.

pkg/stub/handler.go
```diff
@@ -246,12 +249,35 @@ func (h *Handler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) err
...
 		}
 	}
 
 	return nil
 }
 
+// ensureDNSForLoadBalancer configures a wildcard DNS alias for a ClusterIngress
+// targeting the given service.
+func (h *Handler) ensureDNSForLoadBalancer(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service) error {
+	// Ensure DNS is configured for the load balancer.
```
Miciah: This comment doesn't seem helpful.

cmd/cluster-ingress-operator/main.go
```diff
@@ -38,34 +43,74 @@ func main() {
...
+	switch {
+	case ic.Platform.AWS != nil:
+		awsCreds := &corev1.Secret{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "Secret",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "aws-creds",
+				Namespace: "kube-system",
```
Miciah: Could use `metav1.NamespaceSystem` here.

cmd/cluster-ingress-operator/main.go
```diff
@@ -38,34 +43,74 @@ func main() {
...
 	sdk.Watch(resource, kind, namespace, resyncPeriod)
 	// TODO Use a named constant for the router's namespace or get the
 	// namespace from config.
 	sdk.Watch("apps/v1", "DaemonSet", "openshift-ingress", resyncPeriod)
+	sdk.Watch("v1", "Service", "openshift-ingress", resyncPeriod)
 	sdk.Handle(handler)
 	sdk.Run(context.TODO())
 }
+
+func createHandler(namespace string) (*stub.Handler, error) {
```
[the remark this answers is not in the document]
Miciah: Sure, it's only a suggestion.

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+	err := m.Route53.ListHostedZonesPages(&route53.ListHostedZonesInput{}, f)
+	if err != nil {
+		return fmt.Errorf("failed to list hosted zones: %v", err)
+	}
+
+	m.zones = zones
+	return nil
+}
+
+// findClusterPublicZone finds the public zone given the base domain.
```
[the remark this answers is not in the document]

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+	defer m.lock.Unlock()
+	for _, zwt := range m.zones {
+		if aws.StringValue(zwt.Zone.Name) == m.config.BaseDomain && !aws.BoolValue(zwt.Zone.Config.PrivateZone) {
+			return zwt.Zone
+		}
+	}
+	return nil
+}
+
+// findClusterPrivateZone finds the private zone given the base domain and
```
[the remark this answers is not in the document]

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+			targetHostedZoneID = aws.StringValue(lb.CanonicalHostedZoneNameID)
+			break
+		}
+	}
+	if len(targetHostedZoneID) == 0 {
+		return fmt.Errorf("couldn't find hosted zone ID of target ELB with domain name %s", target)
+	}
+
+	// Create or update an alias from the wildcard domain to the service load
+	// balancer hostname in both zones.
```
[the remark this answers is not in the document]
ironcladlou: Extracted the anonymous function to a method and tried to improve the comments.

cmd/cluster-ingress-operator/main.go
```diff
@@ -38,34 +43,75 @@ func main() {
...
+	}
+
+	ic, err := util.GetInstallConfig(k8sclient.GetKubeClient())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get installconfig: %v", err)
+	}
+
+	var dnsManager dns.Manager
+	switch {
+	case ic.Platform.AWS != nil:
```
Miciah: We could have an `InstallConfig` with a nil `Platform`, so we should check that here.

cmd/cluster-ingress-operator/main.go
```diff
@@ -38,34 +43,75 @@ func main() {
...
+		}
+		err := sdk.Get(awsCreds)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get aws creds from %s/%s: %v", awsCreds.Namespace, awsCreds.Name, err)
+		}
+		manager, err := awsdns.NewManager(awsdns.Config{
+			AccessID:   string(awsCreds.Data["aws_access_key_id"]),
+			AccessKey:  string(awsCreds.Data["aws_secret_access_key"]),
+			Region:     ic.Platform.AWS.Region,
+			BaseDomain: ic.BaseDomain + ".",
```
Miciah: As far as I can tell, `BaseDomain` is allowed to have a final dot, so it might be prudent to do a `TrimSuffix` to avoid ending up with two dots.

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+var _ dns.Manager = &Manager{}
+
+// Manager is a DNSManager for AWS which is tightly coupled to the DNS zone
```
Miciah: I guess that should be "is a dns.Manager" now.

pkg/dns/aws/dns.go
```diff
@@ -0,0 +1,248 @@
...
+							DNSName:              aws.String(target),
+							EvaluateTargetHealth: aws.Bool(false),
+						},
+					},
+				},
+			},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("couldn't update DNS record for in zone %s: %v", zoneID, err)
```
Miciah: You have two prepositions and only one prepositional object: "for in zone %s".

cmd/cluster-ingress-operator/main.go
ironcladlou: Fixed

pkg/dns/aws/dns.go
ironcladlou: Revised comment

pkg/dns/aws/dns.go

cmd/cluster-ingress-operator/main.go
Miciah: Ah, right, sorry.  Never mind!

Miciah:
Are the changes under `hack/` intended to be included in this PR? If so, let me know, and I'll go ahead and tag this PR.

ironcladlou:
Yeah, the `hack/` changes are just some bonus fixes/features

PR #60, opened 2018-11-09
Enable dependency pruning

ironcladlou:
Enable pruning to downsize the vendor directory.


Miciah:
`dep check` reports a couple problems:

    % dep check
    # vendor is out of sync:
    github.com/gogo/protobuf: hash of vendored tree not equal to digest in Gopkg.lock
    github.com/googleapis/gnostic: hash of vendored tree not equal to digest in Gopkg.lock

ironcladlou:
1. We shouldn't gate on manual tests anymore; let's incorporate `dep check` into CI separately to rule out "works for me" issues
2. If `dep check` fails for you in your environment (which it does not for me), and all our existing tests pass, what exactly is `dep check` protecting us against?

Miciah:
> We shouldn't gate on manual tests anymore; let's incorporate `dep check` into CI separately to rule out "works for me" issues

The only other operator I found that gates on `dep check` is prometheus-operator (actually, it does a `dep ensure` and then a `diff`):

https://github.com/openshift/prometheus-operator/blob/master/scripts/golang-dep-ensure.sh
https://github.com/openshift/prometheus-operator/blob/master/.travis.yml#L18-L19

> If `dep check` fails for you in your environment (which it does not for me), and all our existing tests pass, what exactly is `dep check` protecting us against?

Presumably it will prevent a PR that discomfits `dep` from necessitating additional effort from the next person who needs to deal with `dep`.

Miciah:
I'll add gating on `dep` later.

/lgtm

PR #61, opened 2018-11-09
Minor improvements

ironcladlou:
* Only create resources if they don't exist. Simplifies some code and brings us
closer to update diffing.
* Switch to the Recreate deployment strategy because we don't do leader elections
yet. Using Recreate will help us avoid contentious operators during rollouts.


ironcladlou:
Expecting CI to fail once since I missed the RBAC change... let's see how our e2e test does catching it.

ironcladlou:
@Miciah looks like e2e failed as expected; pushed the fix and log message cleanup, PTAL.

Miciah:
/retest

Error from e2e-aws looks like a flake:

    * aws_route.to_nat_gw.5: Error finding route after creating it: Unable to find matching route for Route Table (rtb-095684f69012ea8d1) and destination CIDR block (0.0.0.0/0).

PR #67, opened 2018-11-14
Add HTTPS endpoint to router service

ironcladlou:
Add an HTTPS endpoint to the router service. This fixes rencrypt routes and is
also a prerequisite for edge encryption.


Miciah:
/lgtm

> fixes rencrypt routes and is also a prerequisite for edge encryption

It's sufficient (if no custom certificate is needed) and necessary for both, isn't it?

PR #68, opened 2018-11-16
[NE-68] Add support for cluster ingress edge encryption support

ramr:
Allows a reference to a tls secret on the cluster ingress object - which can used to specify the wildcard certificate. If no secret is specified, the default cluster generated certificate (via an annotation) will be used.

Ref:  https://jira.coreos.com/browse/NE-68
 
Caveat: This is still work in progress (and untested - well apart from the unit tests), so use with care ... 

The secret can be created in manner something like ... 
```oc create secret tls ne68-router-secret --cert=./test-cert.pem --key=./test-key.pem```

and used in the ClusterIngress config instance 
`.Spec.DefaultCertificateSecret` = `"ne68-router-secret"`
Alternatively we could just `stuff` all the tls config into the Spec (as a base64 encoded blob for cert and key fields) if you want to `carry` all the config in one place.

@ironcladlou  PTAL  and see if this approach seems okay. Thx

### CODE REVIEW

**Comment by ironcladlou** [comment]
Is there a reason we can't annotate the [actual router service](https://github.com/openshift/cluster-ingress-operator/blob/master/assets/router/service-cloud.yaml) to get a cert?

File: pkg/apis/ingress/v1alpha1/types.go [hunk 1]
```diff
@@ -31,6 +31,10 @@ type ClusterIngressSpec struct {
 	// scheduling of ingress implementation pods.
 	NodePlacement *NodePlacement `json:"nodePlacement"`
 
+	// DefaultCertificateSecret is the name of a secret containing the
+	// default certificate configuration (tls.{crt,key}).
+	DefaultCertificateSecret *string `json:"defaultCertificateSecret"`
```
ironcladlou: Would it be possible to look for the name of a well-known secret in the namespace instead of letting the user specify one? Advantage is we could defer introducing new API for the time being while still letting us support the feature.

ramr:
@ironcladlou @Miciah Made changes as per the discussion on the call today. Still unable to launch a cluster on aws to test this. Giving up on that for now and I am going to try testing this locally ...

assets/router/service-internal.yaml
```diff
@@ -0,0 +1,26 @@
...
+    app: router
+spec:
+  type: ClusterIP
+  selector:
+    app: router
+  ports:
+  - name: http
+    port: 80
+    protocol: TCP
+    targetPort: 80
```
ironcladlou: If you replace hard-coded ports with named port references (e.g. `targetPort: http`) we can safely change the target container ports without impacting the service definition.

pkg/apis/ingress/v1alpha1/types.go
ironcladlou: To try and summarize an offline discussion: we keep the field with a default equal to the the name of the internal service's cert secret. This seems more clear (explicit vs. implicitly named reference) and also makes it easier to share a cert between multiple instances.

ironcladlou: Would just `CertificateSecret` make more sense? In our new model, it's *always* defined (by the user or by a default of the internal service secret).

Or does the "Default" refer to its use in the context of a `Route` lacking inline certs? If so, does that inspire any other possible names?

pkg/manifests/manifests.go
```diff
@@ -165,9 +170,44 @@ func (f *Factory) RouterDaemonSet(cr *ingressv1alpha1.ClusterIngress) (*appsv1.D
...
 	return ds, nil
 }
 
+func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*corev1.Service, error) {
+	s, err := NewService(MustAssetReader(RouterServiceInternal))
+	if err != nil {
+		return nil, err
+	}
+
+	name := "router-" + cr.Name
```
ironcladlou: Does this conflict with [the "external" service](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/manifests/manifests.go#L177)?

assets/router/service-internal.yaml
ramr: Yeah, I did do that change (http and https) -- push was pending based on my local testing. Which I just pushed.

pkg/manifests/manifests.go
ramr: Yes it did. I fixed this locally when I was testing - just pushed the commit up.

pkg/apis/ingress/v1alpha1/types.go
ramr: Yeah, I just the put the `Default` to indicate it was the default router certificate. `CertificateSecret` could possibly also work - could just add that it is the default router cert in the comments.

ironcladlou:
We should try and make a test for this. I suspect the feature can't be used with the default ClusterIngress as-is, because the default value of `defaultCertificateSecret` is `nil`, and as soon as the operator picks up the ClusterIngress and creates the DaemonSet, the `nil` will be respected and we don't process any diffs/updates to `defaultCertificateSecret` post-creation.

How could we make an e2e test for this which tests the default and update functionality?

ramr:
@ironcladlou  there's an integration test but yeah an e2e test would be good. 
I was thinking about this a bit ...  currently we don't check for updates on any of the objects.
I can add something to update the daemonset config (or should we recreate it?) if the current vs expected daemonset config (`Spec`) is different. Would that work for you?
Check around: https://github.com/openshift/cluster-ingress-operator/blob/aec3b5481d243b487e6268b9ae35d9b8b7d2e3d7/pkg/stub/handler.go#L211 
(if error is nil)

The e2e test could then check with and without the default certificate (well without is the default starting point).

ramr:
As per an offline discussion, added in checks for default certificate changes to automatically  update the daemonset and added an e2e test for the same. 

@ironcladlou  PTAL and see if that's ok (in a separate commit but will squash if it looks good). Thx

ironcladlou:
This is looking really good — I think the only thing left is to clarify the e2e test location. My understanding was that the feature level testing should live in the [origin extended test suite](https://github.com/openshift/origin/tree/master/test/extended/router) and that the operator e2e tests should contain only things specific to the machinations of the operator itself. I'm pretty sure this particular test falls into the feature test category because of the way it's testing the things an end-user would do, but I want to let @smarterclayton take a look to be sure.

ironcladlou:
Awesome, thanks. Discussed offline; since this functionality doesn't seem to carry much risk of cross-component breakages impacting the feature, seems okay to keep in our repo.

PR #70, opened 2018-11-20
Operator sdk 0.1.x

ironcladlou:
## 🚧 WIP 🚧

Opening this as a public exercise in upgrading the operator-sdk and re-assessing our requirements since we know much more about what we're doing than when we started pre-CVO.

**Key observation**: I believe this PR shows pretty clearly that we're not actually using *any* functionality from `operator-sdk`. Literally all function comes from `controller-runtime`.

The only thing I came close to using from `operator-sdk` was the controller boilerplate, but because of the way the boilerplate abstracts controller/reconciler construction and wiring and violates inversion of control, it was easier and less code to just ditch all of it and do it myself in the entrypoint.

The generated API code remains, but I'm pretty sure that comes from an upstream project.

The `operator-sdk` CLI tools, test scaffolding, or other generated output are of no use for our CVO-managed operator (this is also true of the old version). None of the assumptions the sdk makes around generated assets or operator deployment topology hold true.

Additionally there are some serious problems introduced by `controller-runtime` that stem from our operator's use of multiple namespaces for operator/operated:

1. There's a capabilities regression from the old operator-sdk: you can now only filter watches by namespace *client side* (and we're watching services/daemonsets in another namespace). We can work around this with a custom predicate, but it won't scale.
2. The built-in owner-ref based parent object re-queuing event handling builtins won't work because we can't create cross-namespaces owner references, necessitating the authoring of custom event handlers (and who knows what else we'll discover).

Even if we were to consolidate into a single namespace solve those problems, it would probably be less confusing to link directly to `controller-runtime`. If we can't actually refer to the `operator-sdk` docs when describing how to work with, develop and test the project, and we only depend on it as a way to transitively import `controller-runtime`, what's the point?

**P.S.,** The intent of this analysis/exercise is not to dismiss the operator-sdk generally/for all operators. Different projects have different needs, and it just happens that my team's operators have one way or another become misaligned with the current solution and I'd like to discuss it.

cc @openshift/sig-network-edge @smarterclayton @bparees @derekwaynecarr @deads2k


ironcladlou:
P.S. not sure if this goes without saying, but the intent of this analysis/exercise is not to dismiss the operator-sdk generally/for all operators. Different projects have different needs, and it just happens that my team's operators have one way or another become misaligned with the current solution and I need help navigating us to a better place.

ironcladlou:
My current thinking is that we should refactor to use a single namespace for both our operator and managed components; at least this way we immediately get the event and queue handling control which was one of the reasons we wanted to upgrade in the first place. We also get the functional dynamic status client, removing a dependency on the CVO client.

The operator-sdk dependency itself continues to serve no immediate purpose, but it's also not necessarily a problem to retain it just for the transitive controller-runtime dependency. It'll just get pruned away for now.

bparees:
> My current thinking is that we should refactor to use a single namespace for both our operator and managed components;

the problem w/ that design is if you want to use finalizers on your managed components (where your operator is the finalizer), you'll deadlock when deleting the namespace.  (the operator/finalizer goes away, and thus can't finalize your managed components, so they can't be deleted).

This is why our operator will live in a separate namespace from anything that requires finalization.  (if garbage collection is sufficient for your cleanup needs, then it's not a problem)

ironcladlou:
@bparees 
>the problem w/ that design is if you want to use finalizers on your managed components (where your operator is the finalizer), you'll deadlock when deleting the namespace. (the operator/finalizer goes away, and thus can't finalize your managed components, so they can't be deleted).

>This is why our operator will live in a separate namespace from anything that requires finalization. (if garbage collection is sufficient for your cleanup needs, then it's not a problem)

Hmmm, this is an interesting point. We do in fact rely on finalizers to handle deletion. For cluster-dns-operator I would ask: when would you ever delete the namespace? For cluster-ingress-operator it's maybe a little fuzzier.

Miciah:
I thought we added finalizers because we could not rely on garbage collection because we had two namespaces, which prevents owner refs from necessarily working properly.  Do we need finalizers with a unified namespace?

bparees:
> I thought we added finalizers because we could not rely on garbage collection because we had two namespaces, which prevents owner refs from necessarily working properly. 

correct.

> Do we need finalizers with a unified namespace?

you need finalizers if you have external resources you need to clean up (e.g. external load balancers or other cloud resources?) so your operator is guaranteed to have a chance to clean that stuff up while tearing down the operand.

ironcladlou:
@Miciah 
>I thought we added finalizers because we could not rely on garbage collection because we had two namespaces, which prevents owner refs from necessarily working properly. Do we need finalizers with a unified namespace?

It's more than just the mechanical cleanup actions: the custom finalizer allows our operator a reliable opportunity to run arbitrary cleanup logic prior to deletion (e.g. clean up external resources, reject the deletion outright, etc.)

ironcladlou:
@Miciah 
>Do we really need to see it if garbage collection is reliable?

Not every cleanup action can be reduced to a k8s GC action, especially external dependencies. For example, the relationship between a `ClusterIngress` and its associated wildcard DNS record on AWS.

Miciah:
Good point about DNS.  Long term though, aren't we looking at having another controller handle external DNS? In which case, we probably will be creating some kind of CR that can be owned by the service (assuming this CR will be in the same namespace).

ironcladlou:
Lessons learned applied in https://github.com/openshift/cluster-ingress-operator/pull/82.

PR #72, opened 2018-11-26
Make router image configurable

ironcladlou:
Make the operated router image configurable and add it to image-references.

Also remove unused test stuff.


pkg/manifests/manifests_test.go
```diff
@@ -99,6 +101,9 @@ func TestManifests(t *testing.T) {
 		t.Errorf("RouterDaemonSet has unexpected node selector: %#v",
 			ds.Spec.Template.Spec.NodeSelector)
 	}
+	if e, a := config.RouterImage, ds.Spec.Template.Spec.Containers[0].Image; e != a {
+		t.Errorf("expect router daemoset %q, got %q", e, a)
```
Miciah: s/daemoset/image/

manifests/02-deployment.yaml
```diff
@@ -31,3 +31,5 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: ingress-operator
+            - name: IMAGE
+              value: openshift/origin-v4.0:haproxy-router
```
Miciah: Should be "openshift/origin-haproxy-router:v4.0".

PR #78, opened 2018-12-04
Makefile: don't specify GOARCH

sosiouxme:
ref https://github.com/openshift/cluster-nfd-operator/pull/2#issue-235099502


ironcladlou:
Thanks!

/lgtm

PR #79, opened 2018-12-05
 Modified ingress operator status to use newer "config.openshift.io/v1" API version

pravisankar:


pravisankar:

PR #80, opened 2018-12-05
Assign system-cluster-critical priority to operator and component

ironcladlou:
Apply the `system-cluster-critical` pod priority class to the operator and
router deployments in accordance with the designation of ingress as a core
feature.

Fixes https://github.com/openshift/cluster-ingress-operator/issues/75.


ironcladlou:
@openshift/sig-network-edge PTAL.

PR #81, opened 2018-12-05
Add UserDefinedClusterIngressHA

Miciah:
Add a high-availability strategy that exposes ports 80 and 443 as endpoints on the host network and leaves implementation of high-availability up to the user.

* `pkg/apis/ingress/v1alpha1/clusteringress_types.go`: Define `UserDefinedClusterIngressHA`.
* `pkg/manifests/manifests.go` (`RouterDeployment`): If HA type is `UserDefinedClusterIngressHA`, use host network, and configure probes to use localhost (defaulting to the node IP address could cause problems if it is not routable or is blocked by a firewall).
* `pkg/manifests/manifests_test.go`: Test with `nil`, `CloudClusterIngressHA`, and `UserDefinedClusterIngressHA`.

This PR resolves [NE-116](https://jira.coreos.com/browse/NE-116).

---

@ironcladlou

Should I add handling of updates, so the operator deletes any service LB if one exists and the HA strategy us "UserDefined", or do we want to leave that as part of the outstanding work to handle updates?

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -174,6 +174,19 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	deployment.Spec.Template.Spec.Containers[0].Image = f.config.RouterImage
 
+	if cr.Spec.HighAvailability != nil && cr.Spec.HighAvailability.Type == ingressv1alpha1.UserDefinedClusterIngressHA {
+		// Expose ports 80 and 443 on the host to provide endpoints for
+		// the user's HA solution.
+		deployment.Spec.Template.Spec.HostNetwork = true
+
+		// With container networking, probes default to using the pod IP
+		// address.  With host networking, probes default to using the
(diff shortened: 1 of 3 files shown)


Miciah:
Rebased.

pkg/manifests/manifests.go
```diff
@@ -174,6 +174,19 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	deployment.Spec.Template.Spec.Containers[0].Image = f.config.RouterImage
 
+	if cr.Spec.HighAvailability != nil && cr.Spec.HighAvailability.Type == ingressv1alpha1.UserDefinedClusterIngressHA {
+		// Expose ports 80 and 443 on the host to provide endpoints for
+		// the user's HA solution.
+		deployment.Spec.Template.Spec.HostNetwork = true
```
pravisankar: We use `hostNetwork` instead of `hostPort` for performance reasons (reduce one more network hop)?

PR #82, opened 2018-12-06
Update to operator-sdk 0.1.1

ironcladlou:
Refactor all the operator internals to use controller-runtime. This is necessary to unpin the operator from outdated kube dependencies.

The refactor should be functionally transparent.

Unfortunately, my best advice right now for reasoning through these changes is to dive into the [controller-runtime docs](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager) and [source](https://github.com/kubernetes-sigs/controller-runtime).

The new `OperatorManager` is (I think) the key abstraction which enables our class of operator (multi-namespaces) to make use of controller-runtime. There are [ongoing discussions](https://github.com/operator-framework/operator-sdk/issues/767) and prototypes upstream to try and support the use case — if and when those materialize, we can revisit our architecture.

Replaces https://github.com/openshift/cluster-ingress-operator/pull/70.


cmd/cluster-ingress-operator/main.go
```diff
@@ -4,88 +4,104 @@ import (
...
 	"github.com/openshift/cluster-ingress-operator/pkg/operator"
-	stub "github.com/openshift/cluster-ingress-operator/pkg/stub"
+	operatorconfig "github.com/openshift/cluster-ingress-operator/pkg/operator/config"
 	"github.com/openshift/cluster-ingress-operator/pkg/util"
 
-	"github.com/operator-framework/operator-sdk/pkg/k8sclient"
-	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
-	k8sutil "github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
-	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
```
pravisankar: Looking at the changes, operator-sdk pkg is used only here, mainly to get `WATCH_NAMESPACE` env. 
Why not use `os.LookupEnv(...)` directly and get rid of operator-sdk pkg dependency?

ironcladlou: Good point. I'd actually also consider switching to a Cobra command and consuming this from flags anyway. In either case, seems like a followup?

pravisankar: yes, that works

PR #83, opened 2018-12-07
replace tectonicClusterID tag with openshiftClusterID

staebler:
The tectonicClusterID tag is being replaced with openshiftClusterID in the installer. This change will use the new tag name.

Hold on openshift/installer#817.

/hold

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -55,7 +55,7 @@ type Config struct {
 	// BaseDomain should be the *absolute* name shared by the zones.
 	// Example: 'devcluster.openshift.com.'
 	BaseDomain string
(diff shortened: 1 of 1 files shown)


ironcladlou:
Users are reporting a private zone discovery failure; pretty sure this change has broken wildcard DNS management:

```
time="2018-12-11T16:16:51Z" level=error msg="failed to reconcile: failed to ensure clusteringress openshift-ingress-operator/default: failed to ensure DNS for router service openshift-ingress/router-default: failed to find private zone for cluster"
```

Unfortunately e2e testing of the feature is still in-progress.

ironcladlou:
Still investigating — my latest cluster has the right tags on the zones, so these users could be using out of date installers or resources.

PR #84, opened 2018-12-11
Cloud HA strategy should automatically use PROXY support when available.

ramr:
Ref: https://jira.coreos.com/browse/NE-120

@ironcladlou  @Miciah @pravisankar PTAL thx

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -7,7 +7,6 @@ import (
 
 	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
 	operatorconfig "github.com/openshift/cluster-ingress-operator/pkg/operator/config"
-	"github.com/openshift/cluster-ingress-operator/pkg/util"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -48,12 +47,11 @@ func MustAssetReader(asset string) io.Reader {
 // files. It provides a point of control to mutate the static resources with
 // provided configuration.
 type Factory struct {
-	config        operatorconfig.Config
-	installConfig *util.InstallConfig
+	config operatorconfig.Config
 }
 
-func NewFactory(config operatorconfig.Config, ic *util.InstallConfig) *Factory {
-	return &Factory{config: config, installConfig: ic}
+func NewFactory(config operatorconfig.Config) *Factory {
+	return &Factory{config: config}
 }
 
 func (f *Factory) DefaultClusterIngress() (*ingressv1alpha1.ClusterIngress, error) {
@@ -155,7 +153,7 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 	if cr.Spec.HighAvailability != nil && cr.Spec.HighAvailability.Type == ingressv1alpha1.CloudClusterIngressHA {
 		// For now, check if we are on AWS. This can really be done for
 		// for any external [cloud] LBs that support the proxy protocol.
-		if f.installConfig.Platform.AWS != nil {
+		if f.config.Platform == operatorconfig.AWSPlatform {
 			env = append(env, corev1.EnvVar{Name: "ROUTER_USE_PROXY_PROTOCOL", Value: "true"})
 		}
 	}
@@ -271,7 +269,7 @@ func (f *Factory) RouterServiceCloud(cr *ingressv1alpha1.ClusterIngress) (*corev
 	}
 	s.Spec.Selector["router"] = name
 
-	if f.installConfig.Platform.AWS != nil {
+	if f.config.Platform == operatorconfig.AWSPlatform {
 		if s.Annotations == nil {
 			s.Annotations = map[string]string{}
 		}
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -7,6 +7,7 @@ import (
 
 	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
 	operatorconfig "github.com/openshift/cluster-ingress-operator/pkg/operator/config"
+	"github.com/openshift/cluster-ingress-operator/pkg/util"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -34,6 +35,9 @@ const (
 	// Annotation used to inform the certificate generation service to
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+
+	// Annotation used to enable the proxy protocol on the AWS load balancer.
+	AWSLBProxyProtocolAnnotation = "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"
 )
 
 func MustAssetReader(asset string) io.Reader {
@@ -44,11 +48,12 @@ func MustAssetReader(asset string) io.Reader {
 // files. It provides a point of control to mutate the static resources with
 // provided configuration.
 type Factory struct {
-	config operatorconfig.Config
+	config        operatorconfig.Config
+	installConfig *util.InstallConfig
 }
 
-func NewFactory(config operatorconfig.Config) *Factory {
-	return &Factory{config: config}
+func NewFactory(config operatorconfig.Config, ic *util.InstallConfig) *Factory {
+	return &Factory{config: config, installConfig: ic}
 }
 
 func (f *Factory) DefaultClusterIngress() (*ingressv1alpha1.ClusterIngress, error) {
@@ -145,6 +150,14 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	if cr.Spec.IngressDomain != nil {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_CANONICAL_HOSTNAME", Value: *cr.Spec.IngressDomain})
+	}
+
+	if cr.Spec.HighAvailability != nil && cr.Spec.HighAvailability.Type == ingressv1alpha1.CloudClusterIngressHA {
+		// For now, check if we are on AWS. This can really be done for
+		// for any external [cloud] LBs that support the proxy protocol.
+		if f.installConfig.Platform.AWS != nil {
+			env = append(env, corev1.EnvVar{Name: "ROUTER_USE_PROXY_PROTOCOL", Value: "true"})
+		}
 	}
 
 	if cr.Spec.NodePlacement != nil {
@@ -257,6 +270,13 @@ func (f *Factory) RouterServiceCloud(cr *ingressv1alpha1.ClusterIngress) (*corev
 		s.Spec.Selector = map[string]string{}
 	}
 	s.Spec.Selector["router"] = name
+
+	if f.installConfig.Platform.AWS != nil {
+		if s.Annotations == nil {
+			s.Annotations = map[string]string{}
+		}
+		s.Annotations[AWSLBProxyProtocolAnnotation] = "*"
+	}
 
 	return s, nil
 }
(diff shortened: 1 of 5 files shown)


assets/router/service-cloud.yaml
```diff
@@ -5,6 +5,8 @@ apiVersion: v1
 metadata:
   # Name is set at runtime.
   namespace: openshift-ingress
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
```
pravisankar: This service is created for any cloud provider environments (AWS, GCP, etc.) but the new annotation is specific to AWS (or the annotation name is misleading?)
May be we could inject the annotation only when the platform is AWS?  Similar to what we do here: https://github.com/openshift/cluster-ingress-operator/blob/1c6c23eff519eb40089c59938b09807897149bf4/cmd/cluster-ingress-operator/main.go#L102

ironcladlou: Yeah, the service and the deployment both need configured as a unit when we're using a cloud HA strategy *and* we're on a supported cloud.

ramr:
@pravisankar  addressed your comments. PTAL thx

pkg/manifests/manifests.go
```diff
@@ -222,6 +235,13 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 	}
 	s.Spec.Selector["router"] = name
 
+	if f.installConfig.Platform.AWS != nil {
+		if s.Annotations == nil {
```
pravisankar: This is not needed, `s.Annotations` is already initialized here: https://github.com/openshift/cluster-ingress-operator/blob/1924e280f10d7457468de81112fbaf8952e1faac/pkg/manifests/manifests.go#L228

ramr: D'oh yeah!! Am still zonked with all the meds. Fixing now ... Thanks.

ramr:
hmm, the `/test e2e-aws` is failing because of proxy support enabled.

pravisankar:
/hold
e2e-aws tests need to be fixed

ramr:
It is like a catch-22-thingy. The feature is to automatically enable the proxy protocol on haproxy (and the aws LB) but the tests that run can't with proxy protocol enabled on haproxy I think. Going to retest a few times just to confirm.

ramr:
@ironcladlou  there's a problem with automatically enabling `Proxy Protocol` support on both haproxy and the AWS cloud load balancer. 

The reason is when we enable this - the haproxy router only speaks the `proxy` protocol and doesn't understand `http[s]` - it expects the cloud load balancer to send a proxy protocol request wrapping the `http` request.

Now a consequence of this is that all the tests that use curl or `equivalent http requests in golang code` to talk directly to the haproxy router will fail. As the router's not going to understand that request. It is like a catch-22 on the environments where we run the unit tests and enable this. It makes sense for production env like settings (but for dev/ci it will cause problems). 

My 2 bits would be to make this _not_ be the default in the case tests are run. 
Either let this setting be controlled by a `ClusterIngressSpec` field  ala `EnableLoadBalancerProxyProtocol == true`. Or alternatively use the reverse like a env variable or something to always disable it for the ci/tests environments. Thoughts? Thx

ironcladlou:
Assuming we want PROXY to be the default when supported (we can discuss that assumption separately) — what if during tests we talked to the router through as many supported layers as possible to maximize our test coverage? For example, if on the default AWS install we assert support for AWS ELB in front of the routers, then tests executing in that configuration could speak to the router through its ELB. Thoughts?

P.S., do you already have a list of which tests are trying to talk to the router directly?

ramr:
@ironcladlou  answers below:
1. So my only concern re: PROXY protocol being default or not is that in the case it is the default,
   there is really no way to turn that off in case someone wants to do. Well other than either remove 
   aws creds or change `ClusterIngressSpec.HighAvailability`. We have seen environments that have
  one user-[micro]-service is talking another using the router others where they are actually
  sending health checks for apps using HTTP internally in the cluster. With the PROXY protocol that's 
  not going to work - change needed to use the the external ELB.

2. We could talk thru' the AWS LB for the tests but since they are written at various times/people -- there's a lot of different tools those do end up using and so different fixes neeed.
Example: https://github.com/openshift/origin/blob/master/test/extended/router/router.go#L69  use the url helper but https://github.com/openshift/origin/blob/master/test/extended/router/scoped.go#L275 uses curl directly etc ... not clean! 

3. So as re: list of tests - most of the routers extended tests are in
      https://github.com/openshift/origin/tree/master/test/extended/router  
    but I don't think that would be the entire set. I just ran a simple `grep curl origin/test` as an example
    and showed up a few others ala: https://github.com/openshift/origin/blob/master/test/extended/builds/service.go#L19

Thx

ramr:
Depends on https://github.com/openshift/origin/pull/21680 to fix a few of the broken tests ...

pkg/manifests/manifests.go
```diff
@@ -33,6 +34,9 @@ const (
 	// Annotation used to inform the certificate generation service to
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+
+	// Annotation used to enable the proxy protocol on the AWS load balancer.
+	AWSLBProxyProtocolAnnotation = "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"
```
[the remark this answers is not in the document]
ramr: Just tried to use that but it causes dep errors. See: https://gist.github.com/ramr/1573ecb990fc42f3d3b8c4de36a8c81c

pkg/manifests/manifests.go
```diff
@@ -43,11 +47,12 @@ func MustAssetReader(asset string) io.Reader {
 // files. It provides a point of control to mutate the static resources with
 // provided configuration.
 type Factory struct {
-	config operatorconfig.Config
+	config        operatorconfig.Config
+	installConfig *util.InstallConfig
```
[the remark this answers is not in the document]
ramr: ack - will move later tonight.

test/e2e/operator_test.go
```diff
@@ -87,6 +88,84 @@ func TestDefaultClusterIngressExists(t *testing.T) {
 	}
 }
 
+func TestClusterProxyProtocol(t *testing.T) {
```
ironcladlou: Is this providing much more value than the unit test in this case?

I'm trying to think of how we could make it more black-box either here or in Origin... like, is the feature really "source IP is preserved on supported cloud environments" (PROXY being an implementation detail)?

ironcladlou: I don't want to block on it right now, just food for thought

ramr: That makes sense - have to rebase anyway so will fix the test then. Probably in the next few days (too many tests + needle pricks today to be in a sane frame of mind to do this right now).

ramr: Aah, we need a service to return back the header info. Plus there's already an extended test for router headers (in the origin repo) that checks the remote ip  -- it is disabled currently - but will work to fix on it there.
Leaving this test here for now (we can at least verify that the proxy protocol works once we have all the pieces in place). Thx

Miciah:
Per https://github.com/openshift/api/pull/149 and the discussion at https://github.com/openshift/installer/pull/943#issuecomment-448641312, I think we will want to rename the operator config's `CloudProvider` field and related definitions in `pkg/operator/config/config.go` to "Platform".

Also, I still think we should disable `TestClusterProxyProtocol` on non-AWS platforms, or change the test to expect a platform-dependent value.

Otherwise, these changes look good.

test/e2e/operator_test.go
```diff
@@ -150,6 +152,99 @@ func TestRouterServiceInternalEndpoints(t *testing.T) {
...
+		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: "openshift-ingress", Name: fmt.Sprintf("router-internal-%s", ci.Name)}, internalService); err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to get internal router service: %v", err)
+	}
+
+	// TODO: Wait for interal router service selector bug to be fixed.
```
Miciah: https://github.com/openshift/cluster-ingress-operator/pull/90 merged, so this TODO is stale, right?

ramr: yeah but as per a comment from @ironcladlou previously I was debating on whether to do it here (or do it in origin as there's already a header test that checks for the `*Forwarded*` headers - that seems like a more complete blackbox e2e test).

ironcladlou:
@Miciah @pravisankar this lgtm, please take another look?

ramr:
Had to rebase  due to a conflict in the imports in `operator_test.go` ...

ramr:
/retest

e2e failure: `error: unable to connect to image repository registry.svc.ci.openshift.org/ci-op-z9lqtdxn/release:latest: Get https://registry.svc.ci.openshift.org/v2/: net/http: TLS handshake timeout`

ramr:
/retest

last error: ` aws_vpc.new_vpc: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.`

Miciah:
/lgtm

Looks like you hit https://github.com/openshift/release/issues/2070:

    error: unable to check your credentials - pass --skip-check to bypass this error: Get https://registry.svc.ci.openshift.org/v2/: net/http: TLS handshake timeout


/retest

Miciah:
error: unable to extract layer sha256:2ff694d94f23c41569f248b398f2646a6dc57667fab71ac5b0a5eea593771489 from registry.svc.ci.openshift.org/ci-op-mwi0hk56/stable@sha256:9cc43d03b4c9aec559903dedf39f2554ce9dd4f2037f9eea194af89118f007af: received unexpected HTTP status: 504 Gateway Time-out

/retest

ramr:
/retest

flake: `error: unable to read image registry.svc.ci.openshift.org/ci-op-mwi0hk56/stable@sha256:fb15a3d0a6731251ccc2700179aeb3e928b6b69a1d83067f0a380e3e6f5342f0: received unexpected HTTP status: 504 Gateway Time-out`

Miciah:
error: unable to read image registry.svc.ci.openshift.org/ci-op-mwi0hk56/stable@sha256:ff9e9fc6bd6d4533e754ee3b4b137842c64370550803d65157874a5ba8ac21f0: received unexpected HTTP status: 504 Gateway Time-out

/retest

Miciah:
Gateway timeout again.

/retest

Miciah:
Gateway timeout again.

/retest

Miciah:
error: unable to connect to image repository registry.svc.ci.openshift.org/ci-op-mwi0hk56/stable@sha256:519fe8e4c9996050c6627dce93deb8d59315a56da30ed2840ef602a672f91829: Get https://registry.svc.ci.openshift.org/v2/: dial tcp 35.196.103.194:443: connect: no route to host

/retest

Miciah:
I made https://github.com/openshift/cluster-ingress-operator/pull/101 to replace install config with the cluster infrastructure config.  If this PR slips, it will need to be rebased, else I'll rebase #101 once this PR merges.

ramr:
Had to update this PR == rebase due to conflicts + change to use infrastructure config.

ironcladlou:
Awesome, thanks. Going to tag this since there are no material changes since the last LGTM.

/lgtm

ramr:
/retest

flake: `timeout error - waiting for bootstrap-complete: timed out waiting for the condition`

ironcladlou:
These last test failures don't look like flakes.

Miciah:
Sorry, @ramr, I introduced a conflict with https://github.com/openshift/cluster-ingress-operator/pull/102.

ramr:
Rebased this but I think I still need some help with testing this ... I can't boot up an installer cluster with the HEAD - I can only boot up a cluster with v0.90 but that doesn't have the `Infrastructure changes`.  Plus I think origin router tests may need a change.

ramr:
/retest

Flake: `error: unable to upload the new image manifest: received unexpected HTTP status: 504 Gateway Time-out`

ramr:
FYI, the router tests will fail until they send requests to the ELB rather than the router service (since the router service will now only "speak" the proxy protocol). This needs https://github.com/openshift/origin/pull/21799  to merge.

ironcladlou:
Looks like this is once again good to go.

/hold cancel

ramr:
/retest

last error: `"waiting for openshift-console URL: context deadline exceeded"`

PR #88, opened 2018-12-14
Consume cluster ID from config.openshift.io

ironcladlou:
Consume cluster ID from its stable location in config.openshift.io. The previous
field on the install configmap is deprecated.

Blocks https://github.com/openshift/installer/pull/783.

### DIFF
--- a/cmd/cluster-ingress-operator/main.go
+++ b/cmd/cluster-ingress-operator/main.go
@@ -67,14 +67,21 @@ func main() {
 	ingressConfig := &configv1.Ingress{}
 	err = kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, ingressConfig)
 	if err != nil {
-		logrus.Fatalf("failed to get ingressconfig 'cluster': %v", err)
+		logrus.Fatalf("failed to get ingress 'cluster': %v", err)
 	}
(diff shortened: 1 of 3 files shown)


Miciah:
Would be good to update the reference to the 'openshiftClusterID' tag in https://github.com/openshift/cluster-ingress-operator/blob/548c236294b688fe71bd0c35876ef9394863c05e/pkg/dns/aws/dns.go#L58 but otherwise
/lgtm

Miciah:
The maximum number of VPC endpoints has been reached

/retest

Miciah:
e2e-aws and e2e-aws-operator are both failing because of the VPC limit.

Miciah:
time="2018-12-15T20:42:59Z" level=fatal msg="failed to get clusterversion 'version': clusterversions.config.openshift.io \"version\" is forbidden: User \"system:serviceaccount:openshift-ingress-operator:ingress-operator\" cannot get clusterversions.config.openshift.io at the cluster scope: no RBAC policy matched"

@ironcladlou, you need to add "clusterversions" to the list of resources here: https://github.com/openshift/cluster-ingress-operator/blob/548c236294b688fe71bd0c35876ef9394863c05e/manifests/00-cluster-role.yaml#L47-L52

PR #89, opened 2018-12-16
[WIP] [NE-72] Handle updates to ClusterIngress and update generated ingress…

ramr:
… assets.

Ref: https://jira.coreos.com/browse/NE-72
This is still work in progress - will add tests as well but first wanted to confirm this works 
for you @ironcladlou  PTAL  Thx

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -1,8 +1,9 @@
-package clusteringress
+package controller
 
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/sirupsen/logrus"
 
@@ -16,6 +17,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
@@ -98,6 +100,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 func (r *reconciler) reconcile() (reconcile.Result, error) {
 	// Ensure we have all the necessary scaffolding on which to place router
 	// instances.
+
 	err := r.ensureRouterNamespace()
 	if err != nil {
 		return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/controller.go
```diff
@@ -289,42 +195,68 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress)
...
-func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.Deployment, ci *ingressv1alpha1.ClusterIngress) error {
-	svc, err := r.ManifestFactory.RouterServiceInternal(ci)
-	if err != nil {
-		return fmt.Errorf("failed to build router service: %v", err)
-	}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}, svc)
+// ensureRouterAsset ensures the expected router asset exists and returns the
+// current asset information.
+func (r *reconciler) ensureRouterAsset(key types.NamespacedName, obj runtime.Object) (runtime.Object, error) {
+	err := r.Client.Get(context.TODO(), key, obj)
```
ironcladlou: This is mutating the input which is surprising (no docs). Seems like a generic dynamic client concern, not something scoped to just router assets. Maybe extend the `Client` interface or something?

Also your deep copy additions reminded me that we're probably not being very safe generally in terms of mutating potentially cached items from our client rather than doing copies everywhere...

pkg/operator/controller/controller.go
```diff
@@ -363,6 +295,203 @@ func (r *reconciler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) err
...
+// ensureRouterClusterRole ensures that the router cluster role exists and
+// matches the expected role.
+func (r *reconciler) ensureRouterClusterRole() error {
+	cr, err := r.ManifestFactory.RouterClusterRole()
+	if err != nil {
+		return fmt.Errorf("failed to build router cluster role: %v", err)
+	}
+
+	current := cr.DeepCopy()
+	key := types.NamespacedName{Namespace: cr.Namespace, Name: cr.Name}
```
ironcladlou: Cluster role isn't namespaced

pkg/operator/controller/controller.go
```diff
@@ -148,72 +151,20 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
...
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get router cluster role %s: %v", cr.Name, err)
-		}
-		err = r.Client.Create(context.TODO(), cr)
-		if err == nil {
-			logrus.Infof("created router cluster role %s", cr.Name)
-		} else if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create router cluster role %s: %v", cr.Name, err)
-		}
+	if err := r.ensureRouterClusterRole(); err != nil {
```
ironcladlou: This isn't namespaced, would it make sense to move it out of `ensureRouterNamespace`?

pkg/operator/controller/controller.go
```diff
@@ -363,6 +295,203 @@ func (r *reconciler) ensureRouterDeleted(ci *ingressv1alpha1.ClusterIngress) err
...
+	key := types.NamespacedName{Namespace: sa.Namespace, Name: sa.Name}
+
+	if _, err := r.ensureRouterAsset(key, current); err != nil {
+		return err
+	}
+
+	// TODO: Should we check for Secrets, ImagePullSecrets and
+	//       AutomountServiceAccountToken - we don't really use those.
+	// TODO: should we check for label/annotation changes or ignore them?
+	modified := !reflect.DeepEqual(current.Labels, sa.Labels) || !reflect.DeepEqual(current.Annotations, sa.Annotations)
```
ironcladlou: This is an annoying one since the struct has no Spec to reflectively compare... is there some util in library-go or elsewhere for a generic "has anything mutable changed" comparisons?

ramr:
@ironcladlou  thanks for the comments ... will do the suggested fixes and break this up into 2 PRs as you suggested over the next few days (one for the smaller change to cluster+namespace and the other for the rest). Keeping this around for reference for now.

ramr:
Closing this out in favor of #94 which only updates a subset of the generated resources.

PR #90, opened 2018-12-18
Fix selector for internal router service, so that it is correctly populated

ramr:
Noticed this while fixing up the router tests in origin - the service selector for the internal service adds an extra selector which causes issues. PTAL Thx

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -220,7 +220,7 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 	if s.Spec.Selector == nil {
 		s.Spec.Selector = map[string]string{}
 	}
-	s.Spec.Selector["router"] = name
+	s.Spec.Selector["router"] = "router-" + cr.Name
 
 	return s, nil
 }
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -87,6 +87,69 @@ func TestDefaultClusterIngressExists(t *testing.T) {
 	}
 }
 
+func TestRouterServiceInternalEndpoints(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/manifests/manifests.go
```diff
@@ -217,11 +217,6 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 	}
 	s.Annotations[ServingCertSecretAnnotation] = fmt.Sprintf("router-certs-%s", cr.Name)
 
-	if s.Spec.Selector == nil {
-		s.Spec.Selector = map[string]string{}
-	}
-	s.Spec.Selector["router"] = name
```
pravisankar: What was the exact issue? Service already has 'router' selector and we are overriding the value?
I see that we also set router selector for RouterServiceCloud().

ironcladlou: I'm also missing the issue here... maybe we should add an e2e test in the operator to verify endpoints for the internal service?

Miciah: We still want to select a specific router, right? So before this change, the service selected on "app=router" and "router=router-internal-default", which was wrong because the second key-value pair did not match any router; with this change, the service selects on only "app=router", which selects _any_ router that cluster-ingress-operator deploys; but what we want is that the service select on "app=router" and "router=router-default" in order to select the specific router for this service, right? Otherwise the service will round-robin all routers, and so you could end up getting metrics or health for any router when querying a specific router's internal service.

ramr: D'oh, that's true - will fix up. Thanks

ramr:
gave up waiting a new cluster to test this ... checking with the bot ...

ironcladlou:
Looks good, thanks!

/lgtm

PR #91, opened 2018-12-18
Use cluster config for DNS management setup

ironcladlou:
Use the cluster config API where possible for DNS management setup.

* Get the base domain from the cluster config
* Discover the AWS region from environment or EC2 metadata


ironcladlou:
DNS tested manually. Once https://github.com/openshift/api/pull/145 lands and the installer is updated to use it, we can finally do some e2e tests for this in origin.

ironcladlou:
Or I could introduce something in the operator e2e tests and migrate to origin ASAP?

ironcladlou:
I take it back, adding the test to operator would require adding a bunch of test scaffolding from origin to do things like execute commands from a pods. Not sure if we want to deal with that now when we're so close to being able to write the tests in origin. Anyway, I tested this one myself.

ironcladlou:
Whoops, I nearly forgot to block on https://github.com/openshift/installer/pull/934 (I was testing against a patched installer!)

ironcladlou:
Hello? Bots?

/retest

PR #93, opened 2018-12-20
Optimize private zone discovery

ironcladlou:
Optimize the private zone discovery algorithm to replace the O(n) AWS API calls
with a single O(1) call to find the zone by its tag.

Needs some more testing. Wildcard DNS e2e is still blocked on a couple other things.

/cc @openshift/sig-network-edge @smarterclayton @wking @csrwng


ironcladlou:
Previously, one per startup, the controller would call `ListTagsForResource` in a tight loop for every zone in the account.

ironcladlou:
https://github.com/openshift/origin/pull/21617 passes against this

smarterclayton:
Structure looks fine to me.

smarterclayton:
The hosted zone region lookup thing is a bit wierd

Miciah:
@ironcladlou, although tests passed, the following is in the operator's logs from the last (successful) test run:

    time="2019-01-04T23:33:09Z" level=error msg="failed to reconcile: failed to ensure clusteringress openshift-ingress-operator/default: failed to ensure DNS for router service openshift-ingress/router-default: failed to discover hosted zones: failed to get tagged resources: AccessDeniedException: User: arn:aws:iam::460538899914:user/origin-ci-robot-provision is not authorized to perform: tag:GetResources\n\tstatus code: 400, request id: 184b3616-1079-11e9-af9c-2f2371f9e33a"

I am seeing the same on another PR from today.

So we need to fix IAM policy to allow `GetResource`, and we apparently need better test coverage,

PR #94, opened 2018-12-22
[NE-72] Handle updates to ClusterIngress and update a subset of the generated ingress assets

ramr:
Break out #89  to just handle updates to a subset of the generated assets. The PR only handles updating the router namespace and deployment assets.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -1,8 +1,9 @@
-package clusteringress
+package controller
 
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/sirupsen/logrus"
 
@@ -15,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
@@ -25,6 +27,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
+
+// resourceMatcherFunc checks if the two objects match.
+type resourceMatcherFunc func(old, new runtime.Object) bool
 
 const (
 	// ClusterIngressFinalizer is applied to all ClusterIngresses before they are
@@ -148,40 +153,20 @@ func (r *reconciler) ensureRouterNamespace() error {
 	if err != nil {
 		return fmt.Errorf("failed to build router cluster role: %v", err)
 	}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: cr.Namespace, Name: cr.Name}, cr)
+	err = r.verifyRouterAssetExists(types.NamespacedName{Namespace: cr.Namespace, Name: cr.Name}, cr, nil)
 	if err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get router cluster role %s: %v", cr.Name, err)
-		}
-		err = r.Client.Create(context.TODO(), cr)
-		if err == nil {
-			logrus.Infof("created router cluster role %s", cr.Name)
-		} else if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create router cluster role %s: %v", cr.Name, err)
-		}
+		return err
 	}
 
-	ns, err := r.ManifestFactory.RouterNamespace()
-	if err != nil {
-		return fmt.Errorf("failed to build router namespace: %v", err)
-	}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns)
-	if err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get router namespace %q: %v", ns.Name, err)
-		}
-		err = r.Client.Create(context.TODO(), ns)
-		if err == nil {
-			logrus.Infof("created router namespace %s", ns.Name)
-		} else if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create router namespace %s: %v", ns.Name, err)
-		}
+	if err := r.ensureRouterNamespaceAsset(); err != nil {
+		return err
 	}
 
 	sa, err := r.ManifestFactory.RouterServiceAccount()
 	if err != nil {
 		return fmt.Errorf("failed to build router service account: %v", err)
 	}
+	// TODO: switch this to use verifyRouterAssetExists(key, obj, nil)
 	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: sa.Namespace, Name: sa.Name}, sa)
 	if err != nil {
 		if !errors.IsNotFound(err) {
@@ -199,6 +184,7 @@ func (r *reconciler) ensureRouterNamespace() error {
 	if err != nil {
 		return fmt.Errorf("failed to build router cluster role binding: %v", err)
 	}
+	// TODO: switch this to use verifyRouterAssetExists(key, obj, nil)
 	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: crb.Name}, crb)
 	if err != nil {
 		if !errors.IsNotFound(err) {
@@ -218,33 +204,9 @@ func (r *reconciler) ensureRouterNamespace() error {
 // ensureRouterForIngress ensures all necessary router resources exist for a
 // given clusteringress.
 func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) error {
-	expected, err := r.ManifestFactory.RouterDeployment(ci)
+	deployment, err := r.ensureRouterDeployment(ci)
 	if err != nil {
-		return fmt.Errorf("failed to build router deployment: %v", err)
-	}
-	current := expected.DeepCopy()
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: expected.Namespace, Name: expected.Name}, current)
-	if err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get router deployment %s/%s, %v", expected.Namespace, expected.Name, err)
-		}
-
-		err = r.Client.Create(context.TODO(), current)
-		if err == nil {
-			logrus.Infof("created router deployment %s/%s", current.Namespace, current.Name)
-		} else if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create router deployment %s/%s: %v", current.Namespace, current.Name, err)
-		}
-	}
-
-	if changed, updated := deploymentConfigChanged(current, expected); changed {
-		err = r.Client.Update(context.TODO(), updated)
-		if err == nil {
-			logrus.Infof("updated router deployment %s/%s", updated.Namespace, updated.Name)
-			current = updated
-		} else {
-			return fmt.Errorf("failed to update router deployment %s/%s, %v", updated.Namespace, updated.Name, err)
-		}
+		return err
 	}
 
 	if ci.Spec.HighAvailability != nil {
@@ -264,8 +226,8 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) 
 				deploymentRef := metav1.OwnerReference{
 					APIVersion: "apps/v1",
 					Kind:       "Deployment",
-					Name:       current.Name,
-					UID:        current.UID,
+					Name:       deployment.Name,
+					UID:        deployment.UID,
 					Controller: &trueVar,
 				}
 				service.SetOwnerReferences([]metav1.OwnerReference{deploymentRef})
@@ -285,12 +247,12 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) 
 		}
 	}
 
-	if err := r.ensureInternalRouterServiceForIngress(current, ci); err != nil {
+	if err := r.ensureInternalRouterServiceForIngress(deployment, ci); err != nil {
 		return fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err)
 	}
 
-	if err := r.syncClusterIngressStatus(current, ci); err != nil {
-		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", current.Namespace, current.Name, err)
+	if err := r.syncClusterIngressStatus(deployment, ci); err != nil {
+		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", deployment.Namespace, deployment.Name, err)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


ramr:
forget to cc @openshift/sig-network-edge

ironcladlou:
@Miciah @pravisankar this lgtm, please take a second look?

pkg/operator/controller/controller.go
```diff
@@ -326,6 +313,76 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+
+	// TODO: should we check for label/annotation changes or ignore them?
+	modified := !reflect.DeepEqual(current.Labels, deployment.Labels) || !reflect.DeepEqual(current.Annotations, deployment.Annotations)
+
+	// TODO: this will need changes when unsupported extensions/patches get
+	//       added to the router deployment.
+	modified = modified || !reflect.DeepEqual(current.Spec, deployment.Spec)
+	if modified {
+		if err := r.Client.Update(context.TODO(), deployment); err != nil {
+			return deployment, fmt.Errorf("failed to update router deployment %s, %v", key.String(), err)
```
Miciah: Returning `deployment` instead of `nil` with an error value looks odd.  The caller isn't going to use the deployment value if the error value is non-nil, so why not return `nil` for the deployment value?

pkg/operator/controller/controller.go
```diff
@@ -326,6 +313,76 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+	if _, err := r.ensureRouterAsset(key, current); err != nil {
+		return err
+	}
+
+	// TODO: should we check for all label/annotation changes or just
+	//       specific ones node-selector/run-level?
+	modified := !reflect.DeepEqual(current.Labels, ns.Labels) || !reflect.DeepEqual(current.Annotations, ns.Annotations)
+	modified = modified || !reflect.DeepEqual(current.Spec, ns.Spec)
+	if modified {
+		if err := r.Client.Update(context.TODO(), ns); err != nil {
```
Miciah: The logic seems convoluted between `ensureRouterAsset` (with the get and maybe create) and `ensureRouterNamespaceAsset` (with the maybe update).  How would you feel about adding an optional `func (old, new runtime.Object) (modified bool)` parameter to `ensureRouterAsset` and moving the update logic to the same?

pkg/operator/controller/controller.go
```diff
@@ -326,6 +313,76 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+	key := types.NamespacedName{Name: ns.Name}
+
+	if _, err := r.ensureRouterAsset(key, current); err != nil {
+		return err
+	}
+
+	// TODO: should we check for all label/annotation changes or just
+	//       specific ones node-selector/run-level?
+	modified := !reflect.DeepEqual(current.Labels, ns.Labels) || !reflect.DeepEqual(current.Annotations, ns.Annotations)
+	modified = modified || !reflect.DeepEqual(current.Spec, ns.Spec)
```
Miciah: The only field in `NamespaceSpec` is `Finalizers`; does the operator really need to re-act if someone futzes with the namespace's finalizers?

pkg/operator/controller/controller.go
```diff
@@ -326,6 +313,76 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+
+	if _, err := r.ensureRouterAsset(key, current); err != nil {
+		return nil, err
+	}
+
+	// TODO: should we check for label/annotation changes or ignore them?
+	modified := !reflect.DeepEqual(current.Labels, deployment.Labels) || !reflect.DeepEqual(current.Annotations, deployment.Annotations)
+
+	// TODO: this will need changes when unsupported extensions/patches get
+	//       added to the router deployment.
```
Miciah: I figure we will apply the patches from `UnsupportedExtensions` before we get to this point.

pkg/operator/controller/controller.go
```diff
@@ -285,13 +250,35 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress)
...
+	if err := r.ensureInternalRouterServiceForIngress(deployment, ci); err != nil {
 		return fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err)
 	}
 
 	return nil
 }
 
+// ensureInternalRouterServiceForIngress ensures that an internal service exists
+// ensureRouterAsset ensures the expected router asset exists and returns the
+// current asset information.
```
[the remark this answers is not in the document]
ironcladlou: Or deep copy the input, and return the (potentially mutated) copy to remove the side effects

pkg/operator/controller/controller.go
ironcladlou: this comment made me wonder if we should actually ignore annotations we don't specifically care about; no solid use case in my head at the moment, just a quick thought before I forget to bring it up later

ironcladlou:
> The goal is to verify that the operator reverts the label when the test add it, but if we check too soon (before the update is done), we may get a false positive. By doing other tests in between adding the label and checking that it is removed, we reduce the likelihood of hitting that race condition, but now we are not testing that the operator reverts the label in response to the test's adding it; instead, we are testing that the operator reverts the label in response to that plus some other stuff. 

I wonder if we all agree on what the goal of the test should be?

Here are some possible goals:

1. Ensure that for a given ClusterIngress, some expected cluster resource set *S* eventually exists
2. Ensure that given (1), and given sort of chaos in the system that messes up *S*, the cluster resource set eventually resolves back to *S* from (1)

To me, the TestClusterIngressUpdate in this PR seems to be a blend of something like those two possible goals.

Would love some thoughts on what tests we want here.

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
ramr: Was a remnant of a previous code iteration - we don't need to return the object. Will remove that from the return set.

pkg/operator/controller/controller.go
ramr: ack will do.

pkg/operator/controller/controller.go
ramr: That won't work for all the different types of runtime objects - basically  for the namespace we are just checking `Labels`, `Annotations` and `Spec` here. But as an example for a `ClusterRole` that could be different ala: `Rules` and `AggregationRules` in lieu of `Spec`.

So the reasoning here was `ensureNamespaceAsset` is the main driver here that ensures that a namespace asset exists and matches the expected asset. If it is any different than the expected asset, it gets updated.

The `ensureRouterAsset` is a just a helper function to ensure that an asset exists - we use that to create if an asset is missing. Maybe renaming that name to `createRouterAssetIfNotExists` or something to that effect might help?

pkg/operator/controller/controller.go
ramr: So that does raise a related question - should we care if say someone adds a label or annotation?  

Meaning in the specific case of the router namespace should we just check that the `openshift.io/run-level` label and `openshift.io/node-selector` match and ignore all other changes?

pkg/operator/controller/controller.go
```diff
@@ -285,13 +250,35 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress)
...
 
-	if err := r.ensureInternalRouterServiceForIngress(current, ci); err != nil {
+	if err := r.ensureInternalRouterServiceForIngress(deployment, ci); err != nil {
 		return fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err)
 	}
 
 	return nil
 }
 
+// ensureInternalRouterServiceForIngress ensures that an internal service exists
```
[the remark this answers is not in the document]
ramr: Ack will remove. 

As re: the comment about why not do the cluster role - just picked a subset here (namespace + deployment). The original PR had all of that including the cluster role, service account, cluster role binding, internal + cloud service etc, and that was too many changes so pruned it to just the namespace and deployment ...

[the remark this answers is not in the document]
Miciah: I'm not suggesting checking whether the cluster role changed, but just using `verifyRouterAssetExists` to create it if it does not exist.  (With the change I suggested to `verifyRouterAssetExists`, you could still use it with a nil `equals` function, I think.)

pkg/operator/controller/controller.go
Miciah: Edited previous example to make it more likely to work.

ramr:
@Miciah implemented your review suggestions. PTAL tx
Note this is untested ... still waiting on getting a cluster ... so meanwhile, letting ci also test it.

pkg/operator/controller/controller.go
```diff
@@ -326,6 +316,76 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+
+		// TODO: Should we instead just check that the
+		//       "openshift.io/run-level" label matches?
+		if !reflect.DeepEqual(ns1.Labels, ns2.Labels) {
+			return false
+		}
+
+		// TODO: Should we instead just check that the
+		//       "openshift.io/node-selector" annotation matches?
+		if !reflect.DeepEqual(ns1.Annotations, ns2.Annotations) {
```
pravisankar: I don't think we need to check individual labels/annotations. This will restrict users to add custom labels/annotations on the namespace and I think that is good, it avoids side effects caused by some annotations (for e.g. node selector on namespace will filter the nodes before node selector on pods is applied).
I feel any custom label/annotation should be part of custom resource to ensure consistent behavior.

test/e2e/operator_test.go
```diff
@@ -181,6 +181,30 @@ func TestClusterIngressUpdate(t *testing.T) {
 		t.Fatalf("failed to get default router deployment: %v", err)
 	}
 
+	// update router namespace with a label.
+	routerNS := &corev1.Namespace{}
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
pravisankar: We don't need to poll here. We know router deployment succeeded in openshift-ingress namespace before this call.
cl.Get() should suffice?

test/e2e/operator_test.go
```diff
@@ -216,6 +240,20 @@ func TestClusterIngressUpdate(t *testing.T) {
 		t.Fatalf("failed to get updated router deployment %s/%s: %v", deployment.Namespace, deployment.Name, err)
 	}
 
+	updatedNamespace := &corev1.Namespace{}
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{Name: "openshift-ingress"}, updatedNamespace); err != nil {
+			return false, nil
```
pravisankar: return false, err

test/e2e/operator_test.go
ramr: Think it should. Will change.

pkg/operator/controller/controller.go
```diff
@@ -285,13 +247,41 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress)
...
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to get router %T %s: %v", desired, key.String(), err)
+		}
+
+		if err := r.Client.Create(context.TODO(), current); err == nil {
+			logrus.Infof("created router asset %T %s", current, key.String())
+		} else if !errors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create router asset %T %s: %v", desired, key.String(), err)
+		}
```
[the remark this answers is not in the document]
ramr: yeah, that would be an optimized path here. will change.

pkg/operator/controller/controller.go
ramr: Ok, I will remove those TODOs.

ramr:
some weirdness here - the line #s that the build fails at don't have the referenced variable `current`? I just squashed both commits into one to see if that makes a difference.

ramr:
retrying with a force push (weirdness in the way commits are being picked up? the variable `current` was renamed to use `deployment` here: https://github.com/openshift/cluster-ingress-operator/pull/94/files#diff-c5c8aa376175139b13b0583b01ad56c6R229 in the commit.

pkg/operator/controller/controller.go
```diff
@@ -351,6 +342,68 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+// ensureRouterDeployment ensures that the router deployment exists and matches
+// the expected deployment.
+func (r *reconciler) ensureRouterDeployment(ci *ingressv1alpha1.ClusterIngress) (*appsv1.Deployment, error) {
+	deployment, err := r.ManifestFactory.RouterDeployment(ci)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build router deployment: %v", err)
+	}
+
+	clusterIngressMatcher := func(obj1, obj2 runtime.Object) bool {
+		ci1 := obj1.(*ingressv1alpha1.ClusterIngress)
```
ironcladlou: Isn't `verifyRouterAssetExists` on L400 passing `Deployment`s?

Miciah: You're right, that's the wrong type! If tests weren't flaking, we'd be getting a nice run-time panic.

ramr: line 400 calls `verifyRouterAssetExists` with `deployment` and that's passed back here w/ any current existing instance of a `deployment` object by that name aka calling `clusterIngressMatcher(current, expected)`. 

if you are referring to the failure in the tests : that looks to be from an old push -- because the line #s are way off + the variable name `current` does exist any more in that function. It looks like when the first force push happened, it ending up picking the first commit and not the second.  If not referring to the failure, ignore this paragraph!

ironcladlou: >line 400 calls verifyRouterAssetExists with deployment and that's passed back here w/ any current existing instance of a deployment object by that name aka calling clusterIngressMatcher(current, expected).

Yes, and then doesn't L384 try to cast the deployment to a *ingressv1alpha1.ClusterIngress?

>if you are referring to the failure in the tests 

I'm not

Miciah:
The operator's log file from the last test run has the following:

    level=error msg="failed to reconcile: failed to ensure clusteringress openshift-ingress-operator/default: failed to create router service openshift-ingress/router-default: Service \"router-default\" is invalid: metadata.ownerReferences.uid: Invalid value: \"\": uid must not be empty"

The problem is that `ensureRouterDeployment` is returning the deployment from the factory, and we need the deployment from the API in `ensureRouterForIngress` so that we have the actual UID when we create the service.  This means `verifyRouterAssetExists` needs to return the current deployment after all.

Miciah:
The operator also now needs permission to update the namespace:

    level=error msg="failed to reconcile: namespaces \"openshift-ingress\" is forbidden: User \"system:serviceaccount:openshift-ingress-operator:ingress-operator\" cannot update namespaces in the namespace \"openshift-ingress\": no RBAC policy matched"

ramr:
/retest 

last error was: `"failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"`

ramr:
@Miciah  am remote today, so no "real" cluster access - I just pushed a new commit to allow update permissions for the namespace - we probably need to be made this more restrictive afterwards ... dunno the exact rbac magic incantantion. Thx

Miciah:
@ramr, the RBAC change looks good.  Did you see my comment about the `uid must not be empty` error? Looks like we need `ensureRouterForIngress` (and thus `verifyRouterAssetExists` too) to return the deployment from the API (rather than the one from the manifest factory).

ironcladlou:
Going to put a temporary hold here as an additional guard for `master`...

/hold

ramr:
@Miciah  missed the other comment - looking now ...

ramr:
@Miciah, pushed a fix - PTAL thx

ramr:
/retest

flake: `https://registry.svc.ci.openshift.org/v2/: net/http: TLS handshake timeout`

ramr:
@ironcladlou  fyi - the tests are green now - still want it on hold?

ironcladlou:
Okay, let's try it.

/hold cancel
/lgtm

PR #101, opened 2019-01-12
Use cluster infrastructure config

Miciah:
Replace use of install config configmap with the infrastructure config object.

This is the past piece required to resolve [NE-118](https://jira.coreos.com/browse/NE-118).

* `cmd/cluster-ingress-operator/main.go` (`main`): Get the cluster infrastructure config object instead of install-config configmap.
(`createDNSManager`): Delete the parameter for the install config and add a parameter for the infrastructure config.  Use the infrastructure to determine platform.
* `manifests/00-cluster-role.yaml`: Delete the rule for access to configmaps, and add a rule for getting infrastructures.
* `pkg/operator/operator.go` (`Operator`): Delete `installConfig` field.
(`New`): Delete installConfig parameter.
* `pkg/util/config.go`:
* `pkg/util/config_test.go`: Delete file.


Miciah:
test "release-latest" failed: pod release-latest was already deleted

/retest

PR #108, opened 2019-01-23
Integrated router metrics with openshift-monitoring

pravisankar:
- Stats username and password is created as secret for the given cluster ingress
which is needed to securely scrape haproxy stats and expose prometheus compatible
metrics at '/metrics' path.
- 'openshift.io/cluster-monitoring' label is applied to openshift-ingress namespace
to inform openshift-monitoring to look for ServiceMonitor objects.
- Created ServiceMonitor for router that will tell which endpoint need to be used
by Prometheus server to pull metrics periodically.
- Necessary cluster/namespace scoped roles are assigned to openshift-monitoring
service account to access the metrics endpoint.

Updated dependent packages needed for openshift-monitoring integration
- Also pinned existing 'github.com/openshift/library-go' package
Updated all the operator dependencies to the newer versions

Populated ClusterOperator.Status.Versions and ClusterOperator.Status.RelatedObjects fields
- ClusterOperator.Status.Version field changed to ClusterOperator.Status.Versions
to incorporate versions for operator and correponding operand(s).

Depends on https://github.com/openshift/origin/pull/22002


pkg/manifests/manifests.go
```diff
@@ -141,6 +170,34 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	env := []corev1.EnvVar{
 		{Name: "ROUTER_SERVICE_NAME", Value: cr.Name},
+		{Name: "ROUTER_LISTEN_ADDR", Value: fmt.Sprintf("0.0.0.0:%d", MetricsListenAddressPort)},
+
+		{Name: "STATS_USERNAME", ValueFrom: &corev1.EnvVarSource{
```
ironcladlou: Would the downward API allow you to hard-code this all in the manifest?

pkg/manifests/manifests.go
```diff
@@ -141,6 +170,34 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
...
+			SecretKeyRef: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: name,
+				},
+				Key: "statsPassword",
+			},
+		}},
+	}
+
+	for i := range deployment.Spec.Template.Spec.Containers {
```
ironcladlou: Could this be hard-coded in the manifest?

pkg/manifests/manifests.go
```diff
@@ -261,6 +318,77 @@ func (f *Factory) RouterServiceCloud(cr *ingressv1alpha1.ClusterIngress) (*corev
...
+	if err != nil {
+		return nil, err
+	}
+	sm.Name = "router-" + ci.Name
+
+	for i := range sm.Spec.Endpoints {
+		if sm.Spec.Endpoints[i].Path == "/metrics" {
+			sm.Spec.Endpoints[i].Port = string(MetricsListenAddressPort)
+			sm.Spec.Endpoints[i].TLSConfig.ServerName = fmt.Sprintf("%s.%s.svc", sm.Name, sm.Namespace)
+			sm.Spec.Endpoints[i].BasicAuth = &monv1.BasicAuth{
```
ironcladlou: As previously discussed, there shouldn't be any basic auth for the metrics endpoint exposed by the router controller

pkg/manifests/manifests.go
```diff
@@ -142,8 +169,52 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
...
+			SecretKeyRef: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: statsSecretName,
+				},
+				Key: "statsPassword",
+			},
+		}},
+	}
+
+	if enableMetrics {
```
ironcladlou: We always want metrics, this can all be inlined above and `enableMetrics` can be removed

pkg/operator/controller/controller.go
```diff
@@ -88,10 +88,11 @@ type CACert struct {
...
 type Config struct {
-	Client          client.Client
-	ManifestFactory *manifests.Factory
-	Namespace       string
-	DNSManager      dns.Manager
+	Client              client.Client
+	ManifestFactory     *manifests.Factory
+	Namespace           string
+	DNSManager          dns.Manager
+	EnableRouterMetrics bool
```
ironcladlou: Metrics should always be on and doesn't need to be configurable

pkg/operator/controller/controller.go
```diff
@@ -251,8 +252,32 @@ func (r *reconciler) ensureRouterNamespace() error {
...
+		if err == nil {
+			logrus.Infof("created router stats secret %s/%s", statsSecret.Namespace, statsSecret.Name)
+		} else if !errors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create router stats secret %s/%s: %v", statsSecret.Namespace, statsSecret.Name, err)
+		}
+	}
+
+	expected, err := r.ManifestFactory.RouterDeployment(ci, r.EnableRouterMetrics)
+	if err != nil {
+		r.Client.Delete(context.TODO(), statsSecret)
```
ironcladlou: Yeah, don't worry about this for now — we have a general bug for cleanup and part of that will be refactoring to use something like the SA as the anchor for siblings of the deployment.

pkg/operator/operator.go
```diff
@@ -82,6 +87,8 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
 		Namespace:       config.Namespace,
 		ManifestFactory: mf,
 		DNSManager:      dnsManager,
+		// We could expose this knob in ClusterIngress.Spec if needed
+		EnableRouterMetrics: true,
```
ironcladlou: Let's keep the internals simple — we have no plans to make this configurable

pravisankar:
@ironcladlou updated and squashed related changes, PTAL

ironcladlou:
Thanks... one more thing we nearly forgot. How can we add automated testing?

pkg/manifests/manifests.go
```diff
@@ -142,9 +169,52 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
...
+				},
+				Key: "statsPassword",
+			},
+		}},
+	}
+
+	// Enable prometheus metrics
+	certsSecretName := fmt.Sprintf("router-metrics-certs-%s", cr.Name)
+	certsVolumeName := "metrics-certs"
+	certsVolumeMountPath := fmt.Sprintf("/etc/pki/tls/%s", certsSecretName)
```
Miciah: Do we really want the mount path to have the router's name? If we do, we should use `filepath.Join` to compose the path.  However, it would be simpler to use a constant path name, and I don't see a reason not to.

pkg/manifests/manifests.go
```diff
@@ -142,9 +169,52 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
...
+		MountPath: certsVolumeMountPath,
+		ReadOnly:  true,
+	}
+
+	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, volume)
+	deployment.Spec.Template.Spec.Containers[0].VolumeMounts = append(deployment.Spec.Template.Spec.Containers[0].VolumeMounts, volumeMount)
+
+	env = append(env, corev1.EnvVar{Name: "ROUTER_METRICS_TYPE", Value: "haproxy"})
+	env = append(env, corev1.EnvVar{Name: "ROUTER_METRICS_TLS_CERT_FILE", Value: fmt.Sprintf("%s/tls.crt", certsVolumeMountPath)})
+	env = append(env, corev1.EnvVar{Name: "ROUTER_METRICS_TLS_KEY_FILE", Value: fmt.Sprintf("%s/tls.key", certsVolumeMountPath)})
```
Miciah: Could use `filepath.Join` here.

pkg/manifests/manifests_test.go
```diff
@@ -56,6 +63,30 @@ func TestManifests(t *testing.T) {
 		t.Errorf("invalid RouterClusterRoleBinding: %v", err)
 	}
 
+	if _, err := f.MetricsClusterRole(); err != nil {
+		t.Errorf("invalid MetricsRole: %v", err)
+	}
+
+	if _, err := f.MetricsClusterRoleBinding(); err != nil {
+		t.Errorf("invalid MetricsRoleBinding: %v", err)
+	}
```
Miciah: Copy-and-paste error: The method is named `MetricsClusterRoleBinding`, and the error message says "MetricsRoleBinding".

pkg/manifests/manifests_test.go
```diff
@@ -56,6 +63,30 @@ func TestManifests(t *testing.T) {
 		t.Errorf("invalid RouterClusterRoleBinding: %v", err)
 	}
 
+	if _, err := f.MetricsClusterRole(); err != nil {
+		t.Errorf("invalid MetricsRole: %v", err)
```
Miciah: Copy-and-paste error: The method is named `MetricsClusterRole`, and the error message says "MetricsRole".

ironcladlou:
Can we get e2e coverage on this from the start? The wildcard DNS test fiasco makes me hesitant to put it off.

Miciah:
Looks like we already have some coverage at least:

    [Conformance][Area:Networking][Feature:Router] The HAProxy router should expose prometheus metrics for a route [Suite:openshift/conformance/parallel/minimal]

    fail [github.com/openshift/origin/test/extended/router/metrics.go:116]: Expected error:
        <expfmt.ParseError>: {
            Line: 1,
            Msg: "expected float as value, got \"\"",
        }
        text format parsing error in line 1: expected float as value, got ""
    not to have occurred

Miciah:
@pravisankar, now that it sets it, `syncOperatorStatus` needs to check whether `.status.versions` changed (maybe in a follow-up PR?); right now it only updates if `.status.conditions` changed.  I made a similar change in cluster-dns-operator: https://github.com/openshift/cluster-dns-operator/pull/71/commits/a1140b78ee3f50a1bd292c45c205e36ceb2c1f33

pravisankar:
@Miciah I will update the pr to check for status conditions, versions and relatedobjects changes (I will use your dns-operator pr as reference).

pkg/util/clusteroperator/status.go
```diff
@@ -35,31 +40,52 @@ func SetStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition,
...
-func ConditionsEqual(oldConditions, newConditions []configv1.ClusterOperatorStatusCondition) bool {
-	if len(newConditions) != len(oldConditions) {
-		return false
+// ElementsEqual return true if listX items match listY items ignoring the order of the items.
+// In case of duplicate items, the number of appearances of each item in both lists should match.
+// listX and listY must be an array or slice.
+func ElementsEqual(listX, listY interface{}, opts []cmp.Option) (bool, error) {
+	// Check empty case
+	if listX == nil && listY == nil {
+		return true, nil
```
Miciah: This is going to plotz if exactly one of `listX` and `listY` is nil.  Test case:
```go
		{
			description: "nil and non-nil are inequal",
			expected:    true,
			a: []configv1.OperandVersion{
				{
					Name:    "operator",
					Version: "v1",
				},
			},
		},
```

pkg/util/clusteroperator/status.go
```diff
@@ -35,31 +40,52 @@ func SetStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition,
...
 
-// ConditionsEqual returns true if and only if the provided slices of conditions
-// (ignoring LastTransitionTime) are equal.
-func ConditionsEqual(oldConditions, newConditions []configv1.ClusterOperatorStatusCondition) bool {
-	if len(newConditions) != len(oldConditions) {
-		return false
+// ElementsEqual return true if listX items match listY items ignoring the order of the items.
+// In case of duplicate items, the number of appearances of each item in both lists should match.
+// listX and listY must be an array or slice.
+func ElementsEqual(listX, listY interface{}, opts []cmp.Option) (bool, error) {
```
Miciah: What do you think about putting this in `pkg/util`? I guess tests could go there too.

pkg/util/clusteroperator/status_test.go
```diff
@@ -184,27 +235,58 @@ func TestConditionsEqual(t *testing.T) {
...
+				},
+				{
+					Type:    configv1.OperatorAvailable,
+					Message: "foo",
+				},
+			},
+		},
+		{
+			description: "check duplicate with multiple conditions",
+			expected:    false,
```
Miciah: Looks like this is the one case where the old and new implementations differ.  Not sure it matters, but the new behavior seems reasonable and more consistent.

pkg/util/clusteroperator/status.go
```diff
@@ -35,31 +40,52 @@ func SetStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition,
...
+	// Check empty case
```
Miciah: This and the other comments inside `ElementsEqual` aren't necessary.

assets/router/namespace.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
   labels:
     # set value to avoid depending on kube admission that depends on openshift apis
     openshift.io/run-level: "0"
+    openshift.io/cluster-monitoring: "true"
```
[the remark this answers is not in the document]
sichvoge: Exactly. We are watching for two things: the namespace should start with `openshift-` and if that label is attached to that namespace.

pravisankar:
Depends on https://github.com/openshift/origin/pull/22002

pravisankar:
@openshift/sig-network-edge updated, ready for review. PTAL

pravisankar:
```
Value (ci-op-f0d876y0-43abb-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name"
level=error msg="\tstatus code: 400, request id: 3de6672f-b04d-4953-b816-b69aa2f850da"
```
/retest

ironcladlou:
WOOHOOO!

PR #109, opened 2019-01-25
Use self-signed default router certificate

Miciah:
## Use self-signed default router certificate
Generate a CA certificate and use it to generate default certificates for routers.  Publish the CA certificate in a configmap, which can be incorporated into CA bundles in order to trust the default certificates.

This PR resolves [NE-139](https://jira.coreos.com/browse/NE-139).

* `manifests/00-cluster-role.yaml`: Grant the operator create, get, list, watch, and delete access to configmaps.
* `pkg/manifests/manifests.go` (`RouterServiceInternal`): Change serving-cert secret name so as not to conflict with the secret that the operator creates.
* `pkg/manifests/manifests_test.go` (`TestManifests`): Adjust for the change to `RouterServiceInternal`.
* `pkg/operator/controller/controller.go` (`caCertSecretName`):
(`caCertConfigMapName`): New constants for the name of the secret and configmap for the CA certificate.
(`CACert`): New type to hold a cached copy of the CA certificate and to keep track of the resource versions of the most recently successfully observed secret and configmap for the CA certificate.
(`Config`): Add `CACert` field.
(`ensureRouterForIngress`): Call `ensureDefaultCertificateForIngress`.
(`ensureDefaultCertificateForIngress`): New function to check that a default certificate exists for a given ClusterIngress, and generate a certificate and create a secret for it if none exists.  Use `ensureRouterCACertificate` to get the certificate used to sign the default certificate.
(`ensureRouterCACertificate`): New function to get the CA certificate for signing default certificates, and generate a CA certificate and create a configmap and secret for it if none exists.
* `test/e2e/operator_test.go` (`TestRouterCACertificate`): New test that gets the configmap with the CA certificate and verifies that the router's default certificate is valid using the CA certificate.

---

~~@ironcladlou, this is WIP for now because the implementation currently requires far too broad permissions for the operator.~~

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -8,6 +8,7 @@ rules:
   - ""
   resources:
   - namespaces
+  - secrets
   - serviceaccounts
   - services
   verbs:
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -239,7 +239,7 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 	if s.Annotations == nil {
 		s.Annotations = map[string]string{}
 	}
-	s.Annotations[ServingCertSecretAnnotation] = fmt.Sprintf("router-certs-%s", cr.Name)
+	s.Annotations[ServingCertSecretAnnotation] = fmt.Sprintf("router-metrics-certs-%s", cr.Name)
 
 	if s.Spec.Selector == nil {
 		s.Spec.Selector = map[string]string{}
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -123,11 +123,12 @@ func TestManifests(t *testing.T) {
 			defaultSecretName, deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName)
 	}
 
+	metricsCertSecretName := fmt.Sprintf("router-metrics-certs-%s", ci.Name)
 	if svc, err := f.RouterServiceInternal(ci); err != nil {
 		t.Errorf("invalid RouterServiceInternal: %v", err)
-	} else if svc.Annotations[ServingCertSecretAnnotation] != defaultSecretName {
+	} else if svc.Annotations[ServingCertSecretAnnotation] != metricsCertSecretName {
 		t.Errorf("RouterServiceInternal expected serving secret annotation %s, got %s",
-			defaultSecretName, svc.Annotations[ServingCertSecretAnnotation])
+			metricsCertSecretName, svc.Annotations[ServingCertSecretAnnotation])
 	}
 
 	if deployment.Spec.Template.Spec.HostNetwork != false {
@@ -221,9 +222,9 @@ func TestManifests(t *testing.T) {
 
 	if svc, err := f.RouterServiceInternal(ci); err != nil {
 		t.Errorf("invalid RouterServiceInternal: %v", err)
-	} else if svc.Annotations[ServingCertSecretAnnotation] != defaultSecretName {
+	} else if svc.Annotations[ServingCertSecretAnnotation] != metricsCertSecretName {
 		t.Errorf("RouterServiceInternal expected serving secret annotation %s, got %s",
-			defaultSecretName, svc.Annotations[ServingCertSecretAnnotation])
+			metricsCertSecretName, svc.Annotations[ServingCertSecretAnnotation])
 	}
 
 	if _, err := f.RouterServiceCloud(ci); err != nil {
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -1,8 +1,10 @@
 package clusteringress
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
@@ -11,12 +13,15 @@ import (
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
 
+	"github.com/openshift/library-go/pkg/crypto"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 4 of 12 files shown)


ironcladlou:
What about storing the CA cert itself in `openshift-ingress-operator` and reserving `openshift-ingress` for resources owned by a ClusterIngress? I think this would keep the dependency graph a little cleaner and also allow the CA cert to survive the destruction of the `openshift-ingress` namespace.

Miciah:
> What about storing the CA cert itself in `openshift-ingress-operator` and reserving `openshift-ingress` for resources owned by a ClusterIngress? I think this would keep the dependency graph a little cleaner and also allow the CA cert to survive the destruction of the `openshift-ingress` namespace.

Yeah, that seems reasonable.  We still want to put the secret for the default certificate in `openshift-ingress`, so that doesn't resolve the RBAC issue.

Miciah:
Latest push adds `TestRouterCACertificate` end-to-end test, moves the configmap with the CA certificate to the "openshift-ingress-operator" namespace, and drops configmaps access permissions from the operator's cluster role.

Miciah:
Latest push changes the name of the CA certificate in the `router-ca` configmap from `tls.crt` to `ca-bundle.crt` to match cluster-kube-apiserver-operator's expectations.

manifests/00-cluster-role.yaml
```diff
@@ -8,6 +8,7 @@ rules:
   - ""
   resources:
   - namespaces
+  - secrets
```
ironcladlou: Ouch

knobunc: Yeah, but how can you get around that?  That's the problem with this approach.

At least it's not the router that can read secrets.

ironcladlou: True. Only way I know of to get around it is to let the CVO create the `openshift-ingress` namespace and a role binding in that namespace for the operator SA.

ironcladlou:
LGTM but I'll hold off tagging until @Miciah has had a chance to comment on whether he's satisfied

pkg/operator/controller/controller.go
```diff
@@ -351,6 +388,164 @@ func (r *reconciler) ensureInternalRouterServiceForIngress(deployment *appsv1.De
...
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to get secret %s/%s: %v", secret.Namespace, secret.Name, err)
+		}
+
+		ca, err := r.ensureRouterCACertificate()
+		if err != nil {
+			return fmt.Errorf("failed to get CA certificate: %v", err)
+		}
+		hostnames := sets.NewString(fmt.Sprintf("*.%s", *ci.Spec.IngressDomain))
+		cert, err := ca.MakeServerCert(hostnames, 0)
```
mrogers950: Rather than a single creation of the CA and cert, you should use the library-go certificate rotators. You'll get creation and rotation on both for free.

CA rotator example:
https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/certrotationcontroller/certrotationcontroller.go#L150

Server cert rotator example: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/certrotationcontroller/certrotationcontroller.go#L175

assets/router/deployment.yaml
```diff
@@ -57,8 +61,15 @@ spec:
...
+            readOnly: true
       volumes:
       - name: default-certificate
         secret:
           defaultMode: 420
           # SecretName is set at run-time.
+      - name: metrics-certificate
+        secret:
+          defaultMode: 420
+          # SecretName is set at run-time.
```
pravisankar: These are metrics related changes and we don't have to deal in this pr. 
PR https://github.com/openshift/cluster-ingress-operator/pull/108 injects similar changes in manifests.go instead of deployment.yaml mainly to enable us to turn on or off metrics feature.

pkg/manifests/manifests_test.go
```diff
@@ -211,19 +217,23 @@ func TestManifests(t *testing.T) {
...
-	if deployment.Spec.Template.Spec.Volumes[0].Secret == nil {
-		t.Error("router Deployment has no secret volume")
+	if nVolumes := len(deployment.Spec.Template.Spec.Volumes); nVolumes != 2 {
+		t.Errorf("expected router Deployment to have 2 volumes, found %d", nVolumes)
 	}
 	if deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName != secretName {
 		t.Errorf("expected router Deployment volume with secret %s, got %s",
 			secretName, deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName)
 	}
+	if deployment.Spec.Template.Spec.Volumes[1].Secret.SecretName != metricsSecretName {
```
pravisankar: Refer volume by name and then check the secret?

pkg/operator/controller/controller.go
Miciah: @mrogers950, thanks for the pointer! It looks like certrotationcontroller is designed to handle a single server certificate whereas we want to have one per router.  Does library-go have something we can use to manage multiple server certifcates, or will it only work for the signing certificate and bundle? Anyway, this is out of scope for 4.0, so I'll add a TODO.

manifests/00-cluster-role.yaml
Miciah: After further discussion, we decided to defer this issue for now.

Miciah:
Latest push squashes the RBAC change into the main commit; removes the metrics-certs secret, volume, and environment variables; and adds a TODO to use certrotationcontroller.

pkg/manifests/manifests.go
```diff
@@ -212,11 +212,15 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
...
 	if cr.Spec.DefaultCertificateSecret != nil && len(*cr.Spec.DefaultCertificateSecret) > 0 {
-		secretName = *cr.Spec.DefaultCertificateSecret
+		defaultCertSecretName = *cr.Spec.DefaultCertificateSecret
 	}
-	deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName = secretName
+	deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName = defaultCertSecretName
+
+	// Fill in the metrics certificate secret name.
+	metricsSecretName := fmt.Sprintf("router-metrics-certs-%s", cr.Name)
+	deployment.Spec.Template.Spec.Volumes[1].Secret.SecretName = metricsSecretName
```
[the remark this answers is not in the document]
Miciah: This code was removed per your other comment.

assets/router/deployment.yaml
Miciah: Removed.  Thanks!

pkg/manifests/manifests_test.go
Miciah: This code was removed per your other comment.

pkg/operator/controller/controller.go
mrogers950: Yeah, right now there's one client+server+CA set per controller, so you would need one controller per router.

Miciah:
Latest push changes the annotation on the internal router service so the serving-cert signer will not stomp the secret that the operator creates.

PR #110, opened 2019-01-31
Put router-ca configmap in openshift-config-managed iff needed

Miciah:
## Put `router-ca` configmap in `openshift-config-managed`

* `manifests/00-cluster-role.yaml`: Grant the operator create, get, list, watch, and delete access to configmaps.
* `pkg/operator/controller/controller.go` (`globalMachineSpecifiedConfigNamespace`): New constant for the `openshift-config-managed` namespace.
(`ensureRouterCACertificate`): Create the `router-ca` configmap in `openshift-config-managed`.
* `test/e2e/operator_test.go` (`TestRouterCACertificate`): Get the `router-ca` configmap from the `openshift-config-managed` namespace.

## Avoid shadowing the `errors` package

* `pkg/operator/controller/controller.go` (`reconcile`): Rename `errors` variable to `errs` to avoid shadowing the `errors` package.

## Delete unnecessary caching of the CA certificate

* `pkg/operator/controller/controller.go` (`CACert`): Delete.
(`Config`): Delete CACert.
(`ensureRouterCACertificate`): Delete caching.

## Delete `router-ca` configmap if it is not needed

If there exists no clusteringress that uses the generated default certificate, delete the configmap for the CA certificate.

This change makes it straightforward for other components to incorporate cluster-ingress-operator's CA certificate into their trust bundles exactly when some router has a default certificate signed by the CA certificate: The `router-ca` configmap will exist in the `openshift-config-managed` namespace iff the certificate CA should be trusted.

This is a follow-up to #109.

Related to [NE-139](https://jira.coreos.com/browse/NE-139).

* `pkg/operator/controller/controller.go` (`reconcile`): Use the new `shouldPublishRouterCA` function to check whether the CA certificate should be published in a configmap, and use the new `ensureRouterCAIsPublished` method create or update it and the new `ensureRouterCAIsUnpublished` method to delete it as appropriate.
(`ensureRouterForIngress`): Use the new `ensureDefaultCertificateDeleted` method to delete the operator-generated default certificate if `.spec.defaultCertificateSecret` is set on the clusteringress.
(`ensureDefaultCertificateForIngress`): Delete unnecessary comment.  Replace `ensureRouterCACertificate` with `getRouterCA`.
(`ensureDefaultCertificateDeleted`): New function that deletes the operator-generated default certificate.
(`getRouterCA`): New function to get the CA, or create it if it does not already exist, using `ensureRouterCACertificate`.
(`ensureRouterCACertificate`): Rename from this...
(`ensureRouterCACertificateSecret`): ...to this.  Change from returning the CA itself to returning the secret.  Delete unnecessary comment.  Do not create the configmap, which is now the responsibility of the `reconcile` function.
(`shouldPublishRouterCA`): New function that returns a Boolean value indicating whether there exists any clusteringress that uses the generated default certificate (in which case the CA certificate that was used to generate them should be published).
(`ensureRouterCAIsPublished`): New function that creates or updates the configmap as needed.
(`ensureRouterCAIsUnpublished`): New function that deletes the configmap.
* `pkg/operator/controller/controller_test.go` (`TestShouldPublishRouterCA`): New test for `shouldPublishRouterCA`.
* `test/e2e/operator_test.go` (`TestClusterIngressUpdate`): Make sure that the configmap exists for the default clusteringress, is deleted when the clusteringress has a custom default certificate set, and is recreated when the clusteringress is changed back to using a default certificate generated by the operator.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -74,30 +74,12 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	return c, nil
 }
 
-// CACert holds the CA certificate as well as the resource versions of the most
-// recently observed CA certificate secret and configmap.
-type CACert struct {
-	// CA is the CA certificate that the operator read from the secret
-	// or generated and put in the secret.
-	CA *crypto.CA
-
-	// CACertConfigMapLastObservedResourceVersion is the resource version of
-	// the config map at the time when the operator most recently
-	// successfully read or created it.
-	CACertConfigMapLastObservedResourceVersion string
-	// CACertSecretLastObservedResourceVersion is the resource version of
-	// the secret at the time when the operator most recently successfully
-	// read or created it.
-	CACertSecretLastObservedResourceVersion string
-}
-
 // Config holds all the things necessary for the controller to run.
 type Config struct {
 	Client          client.Client
 	ManifestFactory *manifests.Factory
 	Namespace       string
 	DNSManager      dns.Manager
-	CACert
 }
 
 // reconciler handles the actual ingress reconciliation logic in response to
@@ -482,14 +464,7 @@ func (r *reconciler) ensureRouterCACertificate() (*crypto.CA, error) {
 	}
 	haveConfigMap := !errors.IsNotFound(err)
 
-	if haveSecret && haveConfigMap &&
-		secret.ResourceVersion == r.CACertSecretLastObservedResourceVersion &&
-		cm.ResourceVersion == r.CACertConfigMapLastObservedResourceVersion {
-		return r.CACert.CA, nil
-	}
-
 	var ca *crypto.CA
-
 	if haveSecret {
 		if _, ok := secret.Data["tls.crt"]; !ok {
 			return nil, fmt.Errorf("failed to read %q from secret %s/%s: %3v", "tls.crt", secret.Namespace, secret.Name, err)
@@ -544,10 +519,6 @@ func (r *reconciler) ensureRouterCACertificate() (*crypto.CA, error) {
 		}
 		logrus.Infof("created configmap %s/%s", cm.Namespace, cm.Name)
 	}
-
-	r.CACert.CA = ca
-	r.CACertConfigMapLastObservedResourceVersion = cm.ResourceVersion
-	r.CACertSecretLastObservedResourceVersion = secret.ResourceVersion
 
 	return ca, nil
 }
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -37,6 +37,11 @@ const (
 	// all states. TODO: Make this generic and not tied to the "default" ingress.
 	ClusterIngressFinalizer = "ingress.openshift.io/default-cluster-ingress"
 
+	// GlobalMachineSpecifiedConfigNamespace is the location for global
+	// config.  In particular, the operator will put the configmap with the
+	// CA certificate in this namespace.
+	GlobalMachineSpecifiedConfigNamespace = "openshift-config-managed"
+
 	// caCertSecretName is the name of the secret that holds the CA certificate
 	// that the operator will use to create default certificates for
 	// clusteringresses.
@@ -468,7 +473,7 @@ func (r *reconciler) ensureRouterCACertificate() (*crypto.CA, error) {
 	cm := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      caCertConfigMapName,
-			Namespace: r.Namespace,
+			Namespace: GlobalMachineSpecifiedConfigNamespace,
 		},
 	}
 	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: cm.Namespace, Name: cm.Name}, cm)
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -158,6 +158,19 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
 			errs = append(errs, fmt.Errorf("failed to ensure clusteringress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 		}
 	}
+
+	if len(errs) == 0 {
+		if shouldPublishRouterCA(ingresses.Items) {
+			if err := r.ensureRouterCAIsPublished(); err != nil {
+				errs = append(errs, fmt.Errorf("failed to ensure router CA was published: %v", err))
+			}
+		} else {
+			if err := r.ensureRouterCAIsUnpublished(); err != nil {
+				errs = append(errs, fmt.Errorf("failed to ensure router CA was unpublished: %v", err))
+			}
+		}
+	}
+
 	return reconcile.Result{}, utilerrors.NewAggregate(errs)
 }
 
@@ -309,8 +322,14 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress) 
 		return fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err)
 	}
 
-	if err := r.ensureDefaultCertificateForIngress(current, ci); err != nil {
-		return fmt.Errorf("failed to create default certificate for clusteringress %s: %v", ci.Name, err)
+	if ci.Spec.DefaultCertificateSecret == nil {
+		if err := r.ensureDefaultCertificateForIngress(current, ci); err != nil {
+			return fmt.Errorf("failed to create default certificate for clusteringress %s: %v", ci.Name, err)
+		}
+	} else {
+		if err := r.ensureDefaultCertificateDeleted(current, ci); err != nil {
+			return fmt.Errorf("failed to delete operator-generated default certificate for clusteringress %s: %v", ci.Name, err)
+		}
 	}
 
 	if err := r.syncClusterIngressStatus(current, ci); err != nil {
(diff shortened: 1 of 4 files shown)


Miciah:
We got the following failure:

    === RUN   TestRouterCACertificate
    --- FAIL: TestRouterCACertificate (1.23s)
    	operator_test.go:671: failed to connect to router at abdc9f8d725a111e9860f0e4a9b0e5d5-1964219959.us-east-1.elb.amazonaws.com:443: x509: certificate signed by unknown authority

Reviewing the events and operator logs, I believe the following is the sequence of events that results in the failure:

1. The operator is running, and it has created the default clusteringress and the default router deployment, which has a volume referencing the `router-certs-default` secret with the generated default certificate as the volume source.
2. `TestClusterIngressUpdate` updates the clusteringress by changing `.spec.defaultCertificateSecret` from its initial setting.
3. The operator updates the deployment to use the new secret name as its volume source.  Then the operator deletes the `router-certs-default` secret and the `router-ca` configmap.
4. The deployment controller spawns a new pod based on the updated deployment.
5. `TestClusterIngressUpdate` updates the clusteringress's `.spec.defaultCertificateSecret` back to its initial setting.
6. The operator updates the deployment to use the original secret name, `router-certs-default`, as its volume source.  Then the operator generates a new CA certificate, creates a new `router-ca` configmap, generates a new default certificate, and creates the `router-certs-default` secret containing the new certificate.
6. At this point, the old pod is still present (the deployment controller hasn't deleted it yet), and now the old pod again matches the deployment (it again specifies `router-certs-default` as the volume source), so the deployment controller deletes the new pod and leaves the old pod running.  However, the old pod is still using the old certificate that it read from the _old_ `router-certs-default` secret.
7. `TestRouterCACertificate` reads the `router-ca` configmap.
8. `TestRouterCACertificate` attempts to connect to the router using the CA certificate it reads from `router-ca`.  This fails with "certificate signed by unknown authority" because `TestRouterCACertificate` is using the new CA certificate, but the old router pod is using a certificate generated using the old CA certificate.

Questions:
1. Is the deployment controller's behavior correct?
2. For expediency, should we work around the problem in the test? The problem should only occur when `.spec.defaultCertificateSecret` is rapidly changed to a new value and then reverted back to its previous value.
3. How can we eliminate the race condition?
   a. Put a timestamp or fingerprint in the secret's name to ensure that we don't use the same name for two different certificates? The operator would need to generate the certificate before creating the deployment and update the clusteringress to refer to the generated secret name, and we would need to devise some way to determine during reconciliation whether a default certificate had been generated by the operator or by the administrator.
   b. Put a timestamp or fingerprint in an annotation in the pod spec of the deployment to ensure the deployment controller doesn't re-use an old pod when the certificate changes? Similar problems to the previous idea.
   c. Something elegant?

pkg/operator/controller/controller.go
```diff
@@ -600,3 +576,88 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
...
+func (r *reconciler) ensureRouterCAIsPublished() error {
+	secret, err := r.ensureRouterCACertificateSecret()
+	if err != nil {
+		return fmt.Errorf("failed to get CA secret: %v", err)
+	}
+
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      caCertConfigMapName,
+			Namespace: r.Namespace,
```
ironcladlou: Shouldn't this be in `openshift-config-managed`?

pkg/operator/controller/controller.go
```diff
@@ -168,10 +150,19 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
...
 		err := r.ensureRouterForIngress(&ingress)
 		if err != nil {
-			errors = append(errors, fmt.Errorf("failed to ensure clusteringress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+			errs = append(errs, fmt.Errorf("failed to ensure clusteringress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+		}
+	}
+
+	if len(errs) == 0 {
+		if shouldPublishRouterCA(ingresses.Items) {
+			r.ensureRouterCAIsPublished()
```
ironcladlou: Error is swallowed

pkg/operator/controller/controller.go
```diff
@@ -168,10 +150,19 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
...
+		} else {
+			r.ensureRouterCAIsUnpublished()
```
ironcladlou: Error is swallowed

pkg/operator/controller/controller.go
```diff
@@ -600,3 +576,88 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
...
+	return nil
+}
+
+// ensureRouterCAIsUnpublished ensures the configmap with the CA certificate is
+// deleted.
+func (r *reconciler) ensureRouterCAIsUnpublished() error {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      caCertConfigMapName,
+			Namespace: r.Namespace,
```
ironcladlou: Shouldn't this be in `openshift-config-managed`?

Miciah: Fixed.

pkg/operator/controller/controller.go
Miciah: Fixed.

pkg/operator/controller/controller.go
Miciah: Fixed.  Thanks!

pkg/operator/controller/controller.go
Miciah: Fixed.  Thanks!

test/e2e/operator_test.go
```diff
@@ -283,6 +283,21 @@ func TestClusterIngressUpdate(t *testing.T) {
 		t.Fatalf("failed to get default router deployment: %v", err)
 	}
 
+	// Wait for the CA certificate configmap to exist.
+	configmap := &corev1.ConfigMap{}
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "router-ca"}, configmap); err != nil {
```
ironcladlou: Shouldn't this be `globalMachineSpecifiedConfigNamespace`?

test/e2e/operator_test.go
```diff
@@ -318,12 +333,40 @@ func TestClusterIngressUpdate(t *testing.T) {
 		t.Fatalf("failed to get updated router deployment %s/%s: %v", deployment.Namespace, deployment.Name, err)
 	}
 
+	// Wait for the CA certificate configmap to be deleted.
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "router-ca"}, configmap); err != nil {
```
ironcladlou: `globalMachineSpecifiedConfigNamespace`?

test/e2e/operator_test.go
```diff
@@ -318,12 +333,40 @@ func TestClusterIngressUpdate(t *testing.T) {
...
+
 	ci.Spec.DefaultCertificateSecret = originalSecret
 	err = cl.Update(context.TODO(), ci)
 	if err != nil {
 		t.Errorf("failed to reset ClusterIngress: %v", err)
 	}
 
+	// Wait for the CA certificate configmap to be recreated.
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "router-ca"}, configmap); err != nil {
```
ironcladlou: `globalMachineSpecifiedConfigNamespace`?

test/e2e/operator_test.go
```diff
@@ -578,7 +621,7 @@ func TestRouterCACertificate(t *testing.T) {
 	var certData []byte
 	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
 		cm := &corev1.ConfigMap{}
-		err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "router-ca"}, cm)
+		err := cl.Get(context.TODO(), types.NamespacedName{Namespace: "openshift-config-managed", Name: "router-ca"}, cm)
```
ironcladlou: Use the constant?

test/e2e/operator_test.go
Miciah: Oops! You're right.  Fixed.

test/e2e/operator_test.go
Miciah: Fixed.

test/e2e/operator_test.go
Miciah: Fixed.

test/e2e/operator_test.go
Miciah: Done.

Miciah:
Hm, should I move `GlobalMachineSpecifiedConfigNamespace` to perhaps `pkg/operator/config/config.go`?

ironcladlou:
@Miciah 
>Hm, should I move `GlobalMachineSpecifiedConfigNamespace` to perhaps `pkg/operator/config/config.go`?

What about rendering it from ManifestFactory like all the other assets?

Miciah:
> What about rendering it from ManifestFactory like all the other assets?

We're not using the manifest factory to create the configmap or the namespace, so that seems like a strange place.  Also, manifests currently just exports constants that no other package uses.  It's weird.

PR #113, opened 2019-02-04
Renamed ingress ClusterOperator name from 'openshift-ingress-operator' to 'ingress'

pravisankar:
https://coreos.slack.com/archives/CCH60A77E/p1549234657654400

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -1,4 +1,4 @@
-package clusteringress
+package controller
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


pravisankar:

pkg/operator/controller/status.go
```diff
@@ -21,10 +21,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	IngressClusterOperatorName = "ingress"
```
ironcladlou: I thought the request was to drop the `openshift-` prefix — wouldn't this then become `ingress-operator`?

Miciah: The referenced Slack message says to drop both.

PR #114, opened 2019-02-04
WIP: dns: clean DNS records on clusteringress deletion

ironcladlou:
When a clusteringress is deleted, ensure that any existing DNS records created
for a load balancer service are also deleted.

### DIFF
--- a/hack/uninstall.sh
+++ b/hack/uninstall.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -uo pipefail
 
-WHAT="${WHAT:-all}"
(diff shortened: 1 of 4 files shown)


ironcladlou:
Have done manual testing but need to think about how to do an e2e...

ironcladlou:
Problem here is that we _assume_ that if the ALIAS name and target match then we own the record, but I'm not entirely sure how we can establish a stronger correlation without reinventing external-dns...

ironcladlou:
Replaced by https://github.com/openshift/cluster-ingress-operator/pull/121

/close

PR #117, opened 2019-02-05
dns/aws: update public zone discovery to search parent domains

abhinavdahiya:
The installer needs to move towards [1] where the private zone named as `cluster_name.base_domain` ie. `cluster_domain` because of [2], but the
public zone still remains `base_domain` as that cannot be created by instaler.

This means that cluster-ingress-operator cannot use the public r53 zone with the same name as the `base_domain` from `DNS.config.openshift.io` [3] as it will be set to the
`cluster_domain`.

This changes the public zone discovery to find a public zone which is the nearest parent domain to `cluster_domain`.

[1]: [openshift/installer#1169](https://github.com/openshift/installer/pull/1169)
[2]: [openshift/installer#1136](https://github.com/openshift/installer/issues/1136)
[3]: https://github.com/openshift/api/blob/d67473e7f1907b74d1f27706260eecf0bc9f2a52/config/v1/types_dns.go#L28

/cc @wking @openshift/sig-network-edge

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -115,22 +115,19 @@ func (m *Manager) discoverZones() error {
 	// Find the public zone, which is the non-private zone whose domain matches
 	// the cluster base domain.
 	if len(m.publicZoneID) == 0 {
+		var lastPublicZone *route53.HostedZone
 		f := func(resp *route53.ListHostedZonesOutput, lastPage bool) (shouldContinue bool) {
-			for _, zone := range resp.HostedZones {
-				if aws.StringValue(zone.Name) == m.config.BaseDomain && !aws.BoolValue(zone.Config.PrivateZone) {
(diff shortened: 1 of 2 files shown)


ironcladlou:
/hold

We need to have a discussion about the requirements here

abhinavdahiya:
> /hold

+1

> We need to have a discussion about the requirements here

What kind of requirements? Anything I can include in the commit message for *why* ?

ironcladlou:
@abhinavdahiya 
>What kind of requirements? Anything I can include in the commit message for _why_ ?

Where's the change to the installer which would break our current assumption? Is it now too risky for us to guess hosted zone names to do record CRUD in? Could we get the zone information from cluster config instead? What would that API look like? etc.

ironcladlou:
Replaced by https://github.com/openshift/cluster-ingress-operator/pull/121

/close

PR #119, opened 2019-02-06
manifests: Add ClusterOperator resource

ironcladlou:
This makes the operator conform to:
    
https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md


Miciah:
/hold
We want to let #108 merge and make sure this PR still passes tests before we merge it.

PR #120, opened 2019-02-12
Increase default clusteringress's replicas to 2

Miciah:
## Increase default clusteringress's replicas to 2

* `assets/defaults/cluster-ingress.yaml`: Change `replicas` from 1 to 2.


Miciah:
So far so good—tests all passed.  Let's do another pass to see whether anything shakes out...

/test e2e-aws
/test e2e-aws-operator

Miciah:
e2e-aws and e2e-aws-operators have passed twice now.  The logs show no unexpected errors or hotlooping.  I can see where `TestClusterIngressScale` would in the past have been scaling from 1 to 2 and then back to 1, now it is scaling from 2 to 3 and then back to 2.

@openshift/sig-network-edge, any concerns with this change?

ironcladlou:
Thanks!

/lgtm

PR #122, opened 2019-02-14
Adds info log msg indicating ns/secret used by DNSManager

danehans:
After troubleshooting what ns/secret was being used by the DNS Manager, I thought it would be nice to add an info log msg.


ironcladlou:
This reminds me that we have a lot of work to introduce events throughout... would it be useful to do a sweep of our error handling and logging to add events to cover this case (and more)?

danehans:
Yes, let me take a stab at it. Thank you.

ironcladlou:
I think https://github.com/openshift/cluster-ingress-operator/pull/126 replaces this one

PR #118, opened 2019-02-06
Set pathlen constraint on router-ca

Miciah:
## Move CA configmap code to a separate file

* `pkg/operator/controller/controller.go` (`caCertConfigMapName`, `GlobalMachineSpecifiedConfigNamespace`, `shouldPublishRouterCA`, `ensureRouterCAIsPublished`, `ensureRouterCAIsUnpublished`): Move from here...
* `pkg/operator/controller/controller_ca_configmap.go`: ...to this new file.

## Move CA secret code to a separate file

* `pkg/operator/controller/controller.go` (`caCertSecretName`, `ensureRouterCACertificateSecret`, `getRouterCA`): Move from here...
* `pkg/operator/controller/controller_ca_secret.go`: ...to this new file.

## Move default certificate code to a separate file

* `pkg/operator/controller/controller.go` (`ensureDefaultCertificateForIngress`, `ensureDefaultCertificateDeleted`): Move from here...
* `pkg/operator/controller/controller_default_certificate.go`: ...to this new file.

## Factor configmap publishing out of reconcile loop

* `pkg/operator/controller/controller.go`: Refactor configmap publishing...
* `pkg/operator/controller/controller_ca_configmap.go` (`ensureRouterCAConfigMap`): ...into this new function.

## Ensure router CA configmap even if other steps fail

Always call `ensureRouterCAConfigMap`, even if there were failures during reconciliation.

This new approach could prematurely delete the configmap if updating a clusteringress with a custom certificate fails.  However, the old approach had the more serious problem that it could prevent the configmap from ever getting published if reconciliation had a any kind of persistent failure.

* `pkg/operator/controller/controller.go` (`reconcile`): Unconditionally call `ensureRouterCAConfigMap`.

## Refactor router CA secret handling

Refactor `ensureRouterCACertificateSecret` in terms of reconciling current and desired state.

* `pkg/operator/controller/controller_ca_secret.go` (`routerCASecretName`): New function that returns the namespaced name of the secret.
(`ensureRouterCACertificateSecret`): Get current secret using the new `currentRouterCASecret` method.  If it doesn't exist, create it using the new `desiredRouterCASecret` function and `createRouterCASecret` method.
(`currentRouterCASecret`): New method.
(`desiredRouterCASecret`): New function.
(`createRouterCASecret`): New method.

## Delete `getRouterCA`

[...]

PR #123, opened 2019-02-15
Requeue requests during controller reconciliation if needed

pravisankar:
Currently we requeue the requests in these cases:
 - cluster DNS config not found
 - cluster Infra config not found
 - get or create ServiceMonitor failed due to unknown GVK

Re-queued requests will be retried after 10 seconds by the controller.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -90,6 +91,9 @@ type Config struct {
 // events.
 type reconciler struct {
 	Config
+
+	dnsConfig   *configv1.DNS
+	infraConfig *configv1.Infrastructure
 }
 
 // Reconcile expects request to refer to a clusteringress in the operator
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/controller.go
```diff
@@ -90,6 +91,9 @@ type Config struct {
 // events.
 type reconciler struct {
 	Config
+
+	dnsConfig   *configv1.DNS
+	infraConfig *configv1.Infrastructure
```
ironcladlou: These should be retrieved from the client/cache every reconcile so we can do change detection.

pkg/operator/controller/controller.go
```diff
@@ -200,7 +177,63 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
 		}
 	}
 
-	return reconcile.Result{}, utilerrors.NewAggregate(errs)
+	return result, utilerrors.NewAggregate(errs)
+}
+
+// updateConfig updates DNS and Infra config
+func (r *reconciler) updateConfig(result *reconcile.Result) error {
```
ironcladlou: I think just doing an inline get for these and passing them to functions which need them more clear — this refactor 

1. Adds unnecessary state to the controller, and
2. Obscures the resource dependencies by introducing ad-hoc reference to controller state

pkg/operator/controller/controller.go
```diff
@@ -318,38 +351,45 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress,
...
-		return fmt.Errorf("failed to ensure load balancer service for %s: %v", ci.Name, err)
-	}
-
-	err = r.ensureDNS(ci, dnsConfig, lbService)
+	errs := []error{}
+	lbService, err := r.ensureLoadBalancerService(ci, current)
 	if err != nil {
-		return fmt.Errorf("failed to ensure DNS for %s: %v", ci.Name, err)
+		errs = append(errs, fmt.Errorf("failed to ensure load balancer service for %s: %v", ci.Name, err))
+	} else {
```
ironcladlou: `ensureDNS()` was designed to do the right thing with whatever the current LB state is. There's no semantic error analysis here. Just because an error was returned by `ensureLoadBalancerService` doesn't mean we can't also try ensuring DNS. It could be the LB was created before and now there's some transient error preventing a no-op reconcile.

pkg/operator/controller/controller.go
```diff
@@ -279,7 +312,7 @@ func (r *reconciler) ensureRouterNamespace() error {
 
 // ensureRouterForIngress ensures all necessary router resources exist for a
 // given clusteringress.
-func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) error {
+func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress, result *reconcile.Result) error {
```
ironcladlou: Mutating input is rarely expected by a function caller absent any docs about it. How about just returning`(Result, error)`?

pkg/operator/controller/controller.go

pkg/operator/controller/controller.go
pravisankar: fixed

pkg/operator/controller/controller.go
pravisankar: We need the same result object to be mutated otherwise the caller need to inspect intermediate result object and update itself that can be ignored by passing the original result object.

pravisankar:

pkg/operator/controller/controller.go
Miciah: > We need the same result object to be mutated otherwise the caller need to inspect intermediate result object and update itself that can be ignored by passing the original result object.

Inspecting the intermediate object shouldn't be necessary, although
```go
			// result was declared earlier.
			var err error
			result, err = r.ensureRouterForIngress(ingress, dnsConfig, infraConfig, result)
			// ensureRouterForIngress may have modified `result` or not.
```

pkg/operator/controller/controller.go
```diff
@@ -336,42 +352,49 @@ func (r *reconciler) ensureRouterForIngress(ci *ingressv1alpha1.ClusterIngress,
...
-		return fmt.Errorf("failed to ensure DNS for %s: %v", ci.Name, err)
+	if err = r.ensureDNS(ci, lbService); err != nil {
+		errs = append(errs, fmt.Errorf("failed to ensure DNS for %s: %v", ci.Name, err))
 	}
 
 	internalSvc, err := r.ensureInternalRouterServiceForIngress(ci, deploymentRef)
 	if err != nil {
-		return fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err)
+		errs = append(errs, fmt.Errorf("failed to create internal router service for clusteringress %s: %v", ci.Name, err))
+		return utilerrors.NewAggregate(errs)
```
Miciah: I thought the idea not to return on non-fatal errors.  Can't we append the error value and move on?

Miciah:
I don't like putting `dnsConfig` and `infraConfig` in `reconciler` because we're growing a large state object and introducing dependencies on `reconciler` (for example, adding it as a receiver for `desiredDNSRecords`).  Can we pass `dnsConfig` and `infraConfig` down as parameters?

pkg/operator/controller/controller.go
pravisankar: Internal service creation has no external dependency, so this will fail when api service fails or service IP is already taken, etc. If api service fails then everything should fail so I categorized this as fatal error. Are you considering transient api failures?

Miciah: We want to move in the direction of trying to do as many tasks as we can even as some tasks fail, but we don't necessarily need to do that now.  And you're right, and I get what you're saying—the issue we urgently must fix is that a non-transient error can prevent the operator from performing subsequent tasks, and you are fixing that problem here.

pravisankar:
@Miciah I moved `dnsConfig` and `infraConfig` to reconciler because it has no specific info with respect to current ingress request and to use this info 3 levels down, we are passing these config objects.

Miciah:
> I moved `dnsConfig` and `infraConfig` to reconciler because it has no specific info with respect to current ingress request and to use this info 3 levels down, we are passing these config objects.

Putting them in `reconciler` is just a less explicit way to pass them 3 levels down.  We can put `dnsConfig` and `infraConfig` in `r` and then pass `r` around everywhere as a receiver (à la your changes to `desiredDNSRecords` and `desiredLoadBalancerService`), or we can pass `dnsConfig` and `infraConfig` around as needed using explicit parameters.  Dan and I had a lot of back and forth about a similar issue on #118, where I was making `desiredFoo` methods with `r` receiver, and Dan objected to their having a receiver.

Anyway, if this PR puts out fires, we can merge today and revise in a follow-up.

PR #124, opened 2019-02-15
controller: refactor event triggering

ironcladlou:
Before this change, the ingress controller was receiving edge triggered events
for related resources, requiring us to look up the entire ingress list on every
event because we had no association from the related resources to the ingress.

With this change, related resources which should trigger a clusteringress
reconcile are labelled to refer to their owning clusteringress. This allows the
use of the controller-runtime `Mapper` interface to translate events for related
resources into clusteringress events. Now the ingresss controller can reconcile
a single clusteringress, which is:

1. Easier to understand and refactor, fitting with the way a controller should
be structured
2. Less susceptible to hot loop bugs, as related resource events can now be
coalesced and rate limited at the scope of a single ingress

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -157,6 +157,11 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	deployment.Name = name
 
+	if deployment.Labels == nil {
+		deployment.Labels = map[string]string{}
+	}
+	deployment.Labels["ingress.openshift.io/clusteringress"] = cr.Name
+
 	if deployment.Spec.Template.Labels == nil {
 		deployment.Spec.Template.Labels = map[string]string{}
 	}
@@ -303,6 +308,7 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 		s.Labels = map[string]string{}
 	}
 	s.Labels["router"] = name
+	s.Labels["ingress.openshift.io/clusteringress"] = cr.Name
 
 	if s.Annotations == nil {
 		s.Annotations = map[string]string{}
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -46,7 +46,7 @@ const (
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
 
-	// OwningClusterIngressLabel should be applied to any objects "owned by" to a
+	// OwningClusterIngressLabel should be applied to any objects "owned by" a
 	// clusteringress to aid in selection (especially in cases where an ownerref
 	// can't be established due to namespace boundaries).
 	OwningClusterIngressLabel = "ingress.openshift.io/clusteringress"
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -45,6 +45,11 @@ const (
 	// Annotation used to inform the certificate generation service to
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+
+	// OwningClusterIngressLabel should be applied to any objects "owned by" to a
+	// clusteringress to aid in selection (especially in cases where an ownerref
+	// can't be established due to namespace boundaries).
+	OwningClusterIngressLabel = "ingress.openshift.io/clusteringress"
 )
 
 func MustAssetReader(asset string) io.Reader {
@@ -160,7 +165,7 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 	if deployment.Labels == nil {
 		deployment.Labels = map[string]string{}
 	}
-	deployment.Labels["ingress.openshift.io/clusteringress"] = cr.Name
+	deployment.Labels[OwningClusterIngressLabel] = cr.Name
 
 	if deployment.Spec.Template.Labels == nil {
 		deployment.Spec.Template.Labels = map[string]string{}
@@ -308,7 +313,7 @@ func (f *Factory) RouterServiceInternal(cr *ingressv1alpha1.ClusterIngress) (*co
 		s.Labels = map[string]string{}
 	}
 	s.Labels["router"] = name
-	s.Labels["ingress.openshift.io/clusteringress"] = cr.Name
+	s.Labels[OwningClusterIngressLabel] = cr.Name
 
 	if s.Annotations == nil {
 		s.Annotations = map[string]string{}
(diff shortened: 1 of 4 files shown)


ironcladlou:
This is ready for review.

pkg/operator/controller/controller.go
```diff
@@ -92,16 +92,20 @@ type reconciler struct {
...
-// expressed by ownerrefs).
+// Reconcile expects request to refer to a clusteringress in the operator
+// namespace, and will do all the work to ensure the clusteringress is in the
+// desired state.
 func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	result, err := r.reconcile(request)
+	if err != nil {
+		// TODO: Figure out why the controller-runtime logger impl isn't logging
+		// these.
+		logrus.Errorf("error: %v", err)
```
ironcladlou: I figured this out, but I think it'll need fixed in a followup. We aren't wiring in a controller-runtime logr implementation, so all controller-runtime logs are getting sent to devnull...

pkg/manifests/manifests.go
```diff
@@ -157,6 +157,11 @@ func (f *Factory) RouterDeployment(cr *ingressv1alpha1.ClusterIngress) (*appsv1.
 
 	deployment.Name = name
 
+	if deployment.Labels == nil {
+		deployment.Labels = map[string]string{}
+	}
+	deployment.Labels["ingress.openshift.io/clusteringress"] = cr.Name
```
Miciah: Named const would be better.

pkg/operator/controller/controller.go
```diff
@@ -92,16 +92,20 @@ type reconciler struct {
...
-// checks and such (as various relationships cross boundaries which can't be
-// expressed by ownerrefs).
+// Reconcile expects request to refer to a clusteringress in the operator
+// namespace, and will do all the work to ensure the clusteringress is in the
+// desired state.
 func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	result, err := r.reconcile(request)
+	if err != nil {
+		// TODO: Figure out why the controller-runtime logger impl isn't logging
+		// these.
```
Miciah: Since you figured it out, we can either delete the comment or update it to describe the problem.

pkg/operator/operator.go
```diff
@@ -99,25 +101,45 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
...
 			return nil, fmt.Errorf("failed to create informer for %v: %v", obj, err)
 		}
-		operatorController.Watch(&source.Informer{Informer: informer}, &handler.EnqueueRequestForObject{})
+		operatorController.Watch(&source.Informer{Informer: informer}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+				labels := a.Meta.GetLabels()
+				if ingressName, ok := labels["ingress.openshift.io/clusteringress"]; ok {
+					logrus.Infof("queueing %s for related %s", ingressName, a.Meta.GetSelfLink())
+					return []reconcile.Request{
+						{NamespacedName: types.NamespacedName{
```
Miciah: The formatting is a little weird here.

pkg/operator/operator.go
```diff
@@ -99,25 +101,45 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
...
+	// resource has the expected label associating the resource with a
+	// clusteringress.
+	for _, o := range []runtime.Object{
 		&appsv1.Deployment{},
 		&corev1.Service{},
 	} {
-		informer, err := ingressCache.GetInformer(obj)
+		// TODO: may not be necessary to copy, but erring on the side of caution for
+		// now given we're in a loop.
+		obj := o.DeepCopyObject()
```
Miciah: Since you don't reference `o` inside the `ToRequests` function or any other function that could capture it, I don't see any way using `o` directly could be problematic.

pkg/operator/controller/controller.go
```diff
@@ -203,6 +193,32 @@ func (r *reconciler) reconcile() (reconcile.Result, error) {
...
+		return fmt.Errorf("failed to finalize load balancer service for %s: %v", ingress.Name, err)
+	}
+	logrus.Infof("finalized load balancer service for ingress %s", ingress.Name)
+	err = r.ensureRouterDeleted(ingress)
+	if err != nil {
+		return fmt.Errorf("failed to delete deployment for ingress %s: %v", ingress.Name, err)
+	}
+	logrus.Infof("deleted deployment for ingress %s", ingress.Name)
+	// Clean up the finalizer to allow the clusteringress to be deleted.
+	updated := ingress.DeepCopyObject().(*ingressv1alpha1.ClusterIngress)
```
Miciah: Why use `DeepCopyObject` and a cast instead of using `DeepCopy`?

pkg/operator/controller/controller.go
```diff
@@ -110,85 +114,71 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		}
-
-		// Handle active ingress.
-		err := r.ensureRouterForIngress(&ingress, dnsConfig, infraConfig)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to ensure clusteringress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 		}
 	}
 
+	// TODO: This should be in a different reconciler as it's independnt of an
```
Miciah: Typo: "independnt".

pkg/manifests/manifests.go
ironcladlou: Fixed

pkg/operator/controller/controller.go

pkg/operator/operator.go
ironcladlou: Technically valid but I agree it's weird, tweaked it.

pkg/operator/operator.go
ironcladlou: If you dig deep enough through `GetInformer()` (which does take `obj`) you'll see that `obj` becomes state of at least one struct (`SharedIndexInformer`) which I'm afraid will ultimately be used in some concurrent context... anyway, I haven't done a thorough audit and there's a _ton_ of code `obj` passes through one way or another, so why not just copy it and not have to think about it? There are no performance considerations here...

pkg/operator/controller/controller.go
ironcladlou: Too much time staring at `runtime.Object`... fixed

pkg/operator/controller/controller.go

ironcladlou:
Known metrics flake

/retest

pkg/manifests/manifests.go
```diff
@@ -45,6 +45,11 @@ const (
 	// Annotation used to inform the certificate generation service to
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+
+	// OwningClusterIngressLabel should be applied to any objects "owned by" to a
```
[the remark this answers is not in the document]
ironcladlou: Can fix in a followup, or if this e2e run fails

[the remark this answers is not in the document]
ironcladlou: Fixed

PR #125, opened 2019-02-16
Requeue if status update fails

Miciah:
Tell the controller to requeue if `syncOperatorStatus` fails.  Failing to update the clusteroperator status may block cluster-version-operator from progressing, so it is important to retry the update immediately, rather than waiting for the next event or the resync interval.

* `pkg/operator/controller/controller.go` (`reconcile`): Set `result.Requeue` to `true` if `syncOperatorStatus` fails.

---

@ironcladlou, @pravisankar, not tested, but does the general idea make sense?

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -97,6 +97,7 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
 		err := r.syncOperatorStatus()
 		if err != nil {
 			logrus.Infof("failed to sync operator status: %v", err)
+			result.Requeue = true
 		}
 	}()
 


Miciah:
test "release-latest" failed: pod release-latest was already deleted

/retest

Miciah:

Wondering whether the false `RetrievedUpdates` condition from cluster-version-operator is related:

    Unable to retrieve available updates: unknown version 0.0.1-2019-02-18-161851

/retest

Miciah:
test "release-latest" failed: pod release-latest was already deleted

/retest

Miciah:
The bootstrap user should successfully login with password decoded from kubeadmin secret [Suite:openshift/conformance/parallel]

        Cmd: "oc login --config=/tmp/configfile760398366 --namespace=e2e-test-bootstrap-login-zvkdt -u kubeadmin -p EsiXrz1nEHZsevELxoE0GQYNB-9et1PlF32EwZl5BKA",
        StdErr: "Error from server (InternalError): Internal error occurred: unexpected response: 400",

We published the `router-ca` configmap, and anyway an invalid certificate shouldn't cause an HTTP error, so this failure does not look related to the PR.

/test e2e-aws

pravisankar:
I added hold tag yesterday as we agreed to return errors instead of explicit requeue to do rate limiting retry. 
But I guess we can merge this change and it will get fixed when I land this change: https://github.com/openshift/cluster-ingress-operator/compare/master...pravisankar:fix-retry-controller?expand=1

Miciah:
> I added hold tag yesterday as we agreed to return errors instead of explicit requeue to do rate limiting retry.

I feel like this is a special case because it is out of the normal flow of control (in that it uses a deferred function call), and logically it likely should be a separate controller (so integrating it into the normal flow of control would take us away from where we want to go with this special case).

> But I guess we can merge this change and it will get fixed when I land this change: https://github.com/openshift/cluster-ingress-operator/compare/master...pravisankar:fix-retry-controller?expand=1

Won't what you are doing there modify `errs` after aggregating it to produce the return value?

Miciah:
I don't think this is an urgent fix though (not aware of any flakes caused by it), so we can keep the hold if you feel that is appropriate or will reduce conflicts for other PRs.

Miciah:
Superseded by https://github.com/openshift/cluster-ingress-operator/pull/132.

PR #126, opened 2019-02-18
Refactors Logging to use the zapr implementation of go-logr/logr

danehans:
The current logging implementation (Logrus), does not implement the logr interface used by controller-runtime. This PR refactors logging to use the zapr implementation of logr so cluster-ingress-operator logging is compatible with controller-runtime logging.


danehans:
@ironcladlou thank you for the feedback. I'll split the dep changes to a separate commit.

> Is it okay to use spaces in key names, like `zone id`? Any benefit in starting some convention around  names there (e.g. alpha, -, _ only)?

The interface [doc](https://github.com/go-logr/logr/blob/master/logr.go#L81-L96) states nothing about the use of spaces in key names. The zapr example uses a space in the key name. I also found the controller-runtime uses a space in the following key name:
```
log.Info("Starting workers", "controller", c.Name, "worker count", c.MaxConcurrentReconciles)
```

> I don't see where you've introduced a controller-runtime [SetLogger](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/log) call, which is needed to get the framework logs flowing

I'll update the PR with the controller-runtime `SetLogger` introduced.

pkg/log/log.go
```diff
@@ -0,0 +1,29 @@
...
+		panic(fmt.Sprintf("error building logger: %v", err))
+	}
+	// zapr defines an implementation of the Logger
+	// interface built on top of Zap (go.uber.org/zap).
+	Logger = zapr.NewLogger(zapLog).WithName("cluster-ingress-operator")
+	Logger.Info("started zapr logger")
+
+	// Set a concrete logging implementation for all
+	// controller-runtime deferred Loggers.
+	log.SetLogger(Logger)
```
ironcladlou: Kind of a nit, but `init()` can be difficult to reason about. Alternative is to extract `SetLogger()` out to the entrypoint in `main.go`... as it stands, controller-runtime logging will only be wired if you happen to import our internal logging package, and if you wanted to reconfigure controller-runtime logging in a different context (e.g. tests) you would have to be careful about package loading order relative to your own `SetLogger` call which will compete with package init.

pkg/operator/controller/controller.go
```diff
@@ -81,27 +82,19 @@ type reconciler struct {
...
-	result, err := r.reconcile(request)
-	if err != nil {
-		// TODO: We're not setting up the controller-runtime logger, so errors we
-		// bubble out are not being logged. For now, log them here, but we need to
-		// redo our logging and wire up the controller-runtime logger because who
-		// knows what else is being eaten.
-		logrus.Errorf("error: %v", err)
-	}
-	return result, err
+	return r.reconcile(request)
```
ironcladlou: Now that you've solved this, do we even need the `reconcile` function?

pkg/operator/controller/controller.go
```diff
@@ -81,27 +82,19 @@ type reconciler struct {
...
 }
 
 func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, error) {
 	// TODO: Should this be another controller?
 	defer func() {
 		err := r.syncOperatorStatus()
 		if err != nil {
-			logrus.Infof("failed to sync operator status: %v", err)
+			log.Info("failed to sync operator status", "error", err)
```
ironcladlou: `log.Error`?

pkg/operator/controller/controller.go
```diff
@@ -38,6 +37,13 @@ const (
 	ClusterIngressFinalizer = "ingress.openshift.io/default-cluster-ingress"
 )
 
+var log = logf.Logger.WithName("cluster-ingress-controller")
+
+func init() {
```
ironcladlou: This seems to shift the problem around. Use case: we decide to add a new controller in a sibling package (we actually want to do this for status calc). Either we duplicate the init boilerplate in the new controller package (unnecessary and conflicting calls to `SetRuntimeLogger`), or contexts in which it would make sense to use just one of the controllers (a test) would have to import this other controller's package to enable logging (confusing).

If logging is wired at a layer above the controller-runtime consumers (e.g. controllers), say at the entrypoint, or maybe the operator package, then we can freely add new controller-runtime consumers without having to deal with package-scoped wiring.

danehans: Ah... that makes sense. I thought I would have to call SetLogger in the controller pkg too, but I was wrong.

ironcladlou:
Thanks for making such quick work of this! Can tweak in followups (e.g. the weird filename logging quirk).

/approve
/lgtm

danehans:
@ironcladlou  [here](https://gist.githubusercontent.com/danehans/bdf68641c1dfd00f556d4bcd2ce8b1b7/raw/f7141bd8117d63fdf8c5e31b084dc7a3257cdfe8/cluster_ingress_zapr_logs) are the zapr logs with the calling function's file name and line number removed from messages.

PR #127, opened 2019-02-18
controller: clean up error handling

ironcladlou:
When reconciling a clusteringress, establish the invariant that DNS and
Infrastructure cluster config must exist; this makes it easier to deal with
error handling and there's little we can do without cluster config anyway. Make
it possible to do some additional work when not able to reconcile a
clusteringress. Pass DNS and Infrastructure around where needed for consistency
with the other refactored bits.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -140,7 +140,7 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
 			infraConfig = nil
 		}
 		if dnsConfig == nil || infraConfig == nil {
-			// For now, if the cluster configs are irretrievable, defer reconciliation
+			// For now, if the cluster configs are unavailable, defer reconciliation
 			// because weaving conditionals everywhere to deal with various nil states
 			// is too complicated. It doesn't seem too risky to rely on the invariant
 			// of the cluster config being available.
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -72,9 +72,6 @@ type Config struct {
 // events.
 type reconciler struct {
 	Config
-
-	dnsConfig   *configv1.DNS
-	infraConfig *configv1.Infrastructure
 }
 
 // Reconcile expects request to refer to a clusteringress in the operator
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/controller_dns.go
```diff
@@ -34,7 +34,7 @@ func (r *reconciler) ensureDNS(ci *ingressv1alpha1.ClusterIngress, service *core
 // If service is nil, no records are desired. If an ingress domain is in use,
 // records are desired in every specified zone present in the cluster DNS
 // configuration.
-func (r *reconciler) desiredDNSRecords(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service) ([]*dns.Record, error) {
+func (r *reconciler) desiredDNSRecords(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service, dnsConfig *configv1.DNS) ([]*dns.Record, error) {
```
Miciah: We can drop the receiver too now.

pkg/operator/controller/controller_lb.go
```diff
@@ -66,7 +66,7 @@ func loadBalancerServiceName(ci *ingressv1alpha1.ClusterIngress) types.Namespace
 // clusteringress, or nil if an LB service isn't desired. An LB service is
 // desired if the high availability type is Cloud. An LB service will declare an
 // owner reference to the given deployment.
-func (r *reconciler) desiredLoadBalancerService(ci *ingressv1alpha1.ClusterIngress, deployment *appsv1.Deployment) (*corev1.Service, error) {
+func (r *reconciler) desiredLoadBalancerService(ci *ingressv1alpha1.ClusterIngress, deployment *appsv1.Deployment, infraConfig *configv1.Infrastructure) (*corev1.Service, error) {
```
Miciah: Receiver can be dropped.

pkg/operator/controller/controller.go
```diff
@@ -111,62 +110,60 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
...
-			return result, utilerrors.NewAggregate(errs)
-		}
-
-		if ingress.DeletionTimestamp != nil {
-			// Handle deletion.
-			if err := r.ensureIngressDeleted(ingress); err != nil {
-				errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
-			}
+		if dnsConfig == nil || infraConfig == nil {
+			// For now, if the cluster configs are irretrievable, defer reconciliation
```
Miciah: "irretrievable" implies that we had them before and has a connotation of permanence.  Maybe "missing" or "unavailable"?

pkg/operator/controller/controller_lb.go
ironcladlou: Fixed

pkg/operator/controller/controller_dns.go

PR #130, opened 2019-02-20
Fix ingress domain and high availability defaulting and status reporting

Miciah:
## Factor `deploymentRef` out of `ensureRouterForIngress`

This change will simplify subsequent refactoring.

* `pkg/operator/controller/controller.go` (`ensureRouterForIngress`): Instead of passing the deployment reference, pass the deployment itself to `ensureInternalRouterServiceForIngress` and `ensureMetricsIntegration`.

## Rename `ensureRouterForIngress`

* `pkg/operator/controller/controller.go` (`reconcile`): Adjust to new name.
(`ensureRouterForIngress`): Rename...
(`ensureClusterIngress`): ...to this.

## Factor out new method `ensureRouterDeployment`

* `pkg/operator/controller/controller.go` (`ensureClusterIngress`): Factor router deployment code out..
* `pkg/operator/controller/controller_router_deployment.go` (`ensureRouterDeployment`): ...into this new method.

## Move `ensureRouterDeleted` definition

* `pkg/operator/controller/controller.go` (`ensureRouterDeleted`): Move from here...
* `pkg/operator/controller/controller_router_deployment.go`: ...to here.

## Rename `current` variable in `ensureClusterIngress`

* `pkg/operator/controller/controller.go` (`ensureClusterIngress`): Rename `current` to `deployment`.

## Tolerate errors from `ensureRouterNamespace`

If `ensureRouterDeployment` returns a non-nil error, append it to `errs` and continue to do whatever work we can do without `deployment`.  Make all ensure methods that take the deployment as an input check for a nil value.  (These methods should return a nil error value because `ensureClusterIngress` already has a non-nil error value from `ensureRouterDeployment`, which will cause a retry.)

* `pkg/operator/controller/controller.go` (`ensureClusterIngress`): Append the error value from `ensureRouterDeployment` to `errs`.  Continue even if the error value is non-nil.
(`ensureMetricsIntegration`, `syncClusterIngressStatus`, `ensureInternalRouterServiceForIngress`, `ensureDefaultCertificateForIngress`, `ensureLoadBalancerService`):
* `pkg/operator/controller/controller_service_monitor.go` (`ensureServiceMonitor`): Check for nil deployment argument.

## Move router customization out of manifest factory

Move the logic that customizes the router deployment out of the manifest factory and into the controller.

* `pkg/manifests/manifests.go` (`RouterDeployment`): Rename...
(`RouterDeploymentAsset`): ...to this.  This change is necessary in order to be able to use the name for a function (see next change).
(`RouterDeployment`): Drop the receiver (making this a function) and the error return value.  Panic if unable to read the asset.  Otherwise return the unmodified asset.  Move all customization of the deployment from here...
* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): ...to here.  New function that returns the desired router deployment based on the provided clusteringress, router image name, and infrastructure config.
(`routerDeploymentName`): New function to generate the namespaced name for the router deployment.
(`ensureRouterDeployment`): Add parameter for infrastructure config.  Use `desiredRouterDeployment` to get the expected deployment.
(`ensureRouterDeleted`): Adjust to the factory function's new signature.  Use `routerDeploymentName` to set the deployment's name and namespace.
* `pkg/operator/controller/controller.go` (`ensureClusterIngress`): Pass `infraConfig` to `ensureRouterDeployment`.
* `pkg/manifests/manifests_test.go` (`TestManifests`): Move checks on the router deployment from here...
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDesiredRouterDeployment`): ...to here.

## Fix ingress domain defaulting not to modify spec

Move the ingress domain defaulting from the manifest factory to the controller.  Reflect the actual ingress domain (if any) in the status of the clusteringress rather than setting it in the spec.

This commit resolves [NE-151](https://jira.coreos.com/projects/NE/issues/NE-151).

This commit fixes [BZ#1678723](https://bugzilla.redhat.com/show_bug.cgi?id=1678723).

* `cmd/cluster-ingress-operator/main.go` (`main`): Delete code to get ingress config.  Delete assignment of `operatorConfig.DefaultIngressDomain`.
* `pkg/apis/ingress/v1alpha1/clusteringress_types.go` (`ClusterIngressStatus`): Add `IngressDomain` field.
* `pkg/apis/ingress/v1alpha1/zz_generated.deepcopy.go`: Regenerate.
* `pkg/manifests/manifests.go` (`DefaultClusterIngress`): Delete assignment to `ci.Spec.IngressDomain`.
* `pkg/manifests/manifests_test.go` (`TestManifests`): Delete code related to ingress domain.
* `pkg/operator/config/config.go` (`Config`): Delete `DefaultIngressDomain` field.
* `pkg/operator/controller/controller.go` (`reconcile`): Get the ingress config.  Use `enforceEffectiveIngressDomain` to ensure the clusteringress has the effective ingress domain in its status.
(`enforceEffectiveIngressDomain`): New function to determine the effective ingress domain from a clusteringress and the ingress configuration and publish it to the clusteringress's status.
* `pkg/operator/controller/controller_default_certificate.go` (`desiredRouterDefaultCertificateSecret`): Use the ingress domain from the clusteringress's status instead of from its spec.
* `pkg/operator/controller/controller_dns.go` (`desiredDNSRecords`): Use the ingress domain from the clusteringress's status instead of from its spec.
* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): Use the ingress domain from the clusteringress's status instead of using the clusteringress spec to set the router's canonical hostname.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDesiredRouterDeployment`): Set ingress domain in the clusteringress's status instead of in its spec.  Add checks that the canonical hostname is set, or not, as appropriate.
* `test/e2e/operator_test.go` (`TestRouterCACertificate`): Read ingress domain from clusteringress status.

## Fix high availability defaulting and status

Move the high availability configuration defaulting from the manifest factory to the controller.  Reflect the actual HA configuration in the status of the clusteringress rather than setting it in the spec.

This commit resolves [NE-157](https://jira.coreos.com/browse/NE-157).

* `pkg/apis/ingress/v1alpha1/clusteringress_types.go` (`NoClusterIngressHA`): New constant.  This HA type corresponds to the behavior when `.spec.highAvailability` was nil prior to this commit.
(`ClusterIngressStatus`): Add `HighAvailability` field.
* `pkg/apis/ingress/v1alpha1/zz_generated.deepcopy.go:` Regenerate.
* `pkg/manifests/manifests.go` (`DefaultClusterIngress`): Drop the receiver and the error return value.  Panic if unable to read the asset.  Otherwise return the unmodified asset.
* `pkg/manifests/manifests_test.go` (`TestDefaultClusterIngress`): Delete checks for `.spec.highAvailability`.
* `pkg/operator/controller/controller.go` (`reconcile`): Pass `infraConfig` to `ensureIngressDeleted`.
(`haTypeForInfra`): New function to determine the HA type from the infrastructure config.
(`effectiveHighAvailability`): New function to determine the HA configuration from a clusteringress and infrastructure config.  Determine the current HA type from the clusteringress spec if it is specified there or else using `haTypeForInfra`.
(`ensureIngressDeleted`): Add parameter for infrastructure config.  Use `expectedHighAvailability` to determine the HA configuration.  Pass the HA configuration to `finalizeLoadBalancerService`.
(`ensureClusterIngress`): Use `effectiveHighAvailability` to determine the desired HA configuration.  Pass it to `ensureRouterDeployment`, `ensureLoadBalancerService`, `ensureDNS`, and `syncClusterIngressStatus`.
(`syncClusterIngressStatus`): Add parameter for HA.  Set the current HA status in the clusteringress status.
* `pkg/operator/controller/controller_dns.go` (`ensureDNS`): Add parameter for HA.  Use this parameter to pass the HA type to `desiredDNSRecords`.
(`desiredDNSRecords`): Add parameter for HA type.  Use this parameter instead of using the clusteringress spec to determine whether to manage DNS.
* `pkg/operator/controller/controller_lb.go` (`ensureLoadBalancerService`): Add parameter for HA.  Use this parameter to pass the HA type to `desiredLoadBalancerService`.
(`desiredLoadBalancerService`): Add parameter for HA type.  Use this parameter instead of using the clusteringress spec to determine whether to create an LB service.
(`finalizeLoadBalancerService`): Add parameter for HA.  Use this parameter to pass the HA type to `desiredDNSRecords`.
* `pkg/operator/controller/controller_router_deployment.go` (`ensureRouterDeployment`): Add parameter for HA.  Use this parameter to pass the HA type to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Add parameter for HA type.  Use this parameter instead of using the clusteringress spec to determine whether to use PROXY protocol or host networking.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDesiredRouterDeployment`): Pass appropriate test inputs for the HA type to `desiredRouterDeployment`.
* `pkg/operator/operator.go` (`Start`): Adjust to the new signature of `DefaultClusterIngress`.
* `test/e2e/operator_test.go` (`TestRouterCACertificate`): Get HA type from clusteringress status.

## Refactor router deployment handling

Refactor `ensureRouterDeployment` in terms of reconcling current and desired state.

* `pkg/operator/controller/controller_router_deployment.go` (`ensureRouterDeployment`): Get current deployment using the new `currentRouterDeployment` method.  If no deployment exists, create one using the new `createRouterDeployment` method.  If one does exist, update it to the desired deployment using the new `updateRouterDeployment` method.
(`currentRouterDeployment`, `createRouterDeployment`, `updateRouterDeployment`): New methods.

## Move `deploymentConfigChanged`

Move `deploymentConfigChanged` to be with the rest of the controller code for the router deployment.

* `pkg/operator/controller/controller.go` (`deploymentConfigChanged`): Move
from here...
* `pkg/operator/controller/controller_router_deployment.go:` ...to here.


Miciah:
level=fatal msg="failed to initialize the cluster: Cluster operator monitoring has not yet reported success"

/test e2e-aws-operator

    Unable to retrieve available updates: unknown version 0.0.1-2019-02-20-064155

/test e2e-aws

Miciah:
error: unable to read image registry.svc.ci.openshift.org/ci-op-gmitbmpi/stable@sha256:458c6a0be6dfdbeff28c01adbacaa78ed4d72749593488644307d17a60ceea2b: received unexpected HTTP status: 504 Gateway Time-out

/test e2e-aws

Miciah:
More gateway timeouts.

pkg/operator/controller/controller_dns.go
```diff
@@ -15,8 +15,12 @@ import (
 
 // ensureDNS will create DNS records for the given LB service. If service is
 // nil, nothing is done.
-func (r *reconciler) ensureDNS(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service, dnsConfig *configv1.DNS) error {
-	dnsRecords, err := desiredDNSRecords(ci, service, dnsConfig)
+func (r *reconciler) ensureDNS(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service, dnsConfig *configv1.DNS, ingressDomain *string, ha *ingressv1alpha1.ClusterIngressHighAvailability) error {
+	if ha == nil {
```
ironcladlou: Passing a pointer to this seems really weird... we have a None type and should have come up with the effective value before calling this, so can it ever even be nil? What about checking that ha type is Cloud?

pkg/operator/controller/controller_lb.go
```diff
@@ -34,8 +34,12 @@ const (
 // ensureLoadBalancerService creates an LB service if one is desired but absent.
 // Always returns the current LB service if one exists (whether it already
 // existed or was created during the course of the function).
-func (r *reconciler) ensureLoadBalancerService(ci *ingressv1alpha1.ClusterIngress, deployment *appsv1.Deployment, infraConfig *configv1.Infrastructure) (*corev1.Service, error) {
-	desiredLBService, err := desiredLoadBalancerService(ci, deployment, infraConfig)
+func (r *reconciler) ensureLoadBalancerService(ci *ingressv1alpha1.ClusterIngress, deployment *appsv1.Deployment, infraConfig *configv1.Infrastructure, ha *ingressv1alpha1.ClusterIngressHighAvailability) (*corev1.Service, error) {
+	if deployment == nil || ha == nil {
```
ironcladlou: Remove ha pointer and check whether the type is Cloud?

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -0,0 +1,226 @@
...
+	return types.NamespacedName{
+		Namespace: "openshift-ingress",
+		Name:      "router-" + ci.Name,
+	}
+}
+
+// ensureRouterDeployment ensures the router deployment exists for a given
+// clusteringress.
+func (r *reconciler) ensureRouterDeployment(ci *ingressv1alpha1.ClusterIngress, infraConfig *configv1.Infrastructure, ingressDomain *string, ha *ingressv1alpha1.ClusterIngressHighAvailability) (*appsv1.Deployment, error) {
+	if infraConfig == nil || ha == nil {
```
ironcladlou: I didn't think we called this if infraConfig was nil, and as above, I don't think ha should be a pointer

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, nil
+	}
+
+	expected, err := desiredRouterDeployment(ci, r.Config.RouterImage, ingressDomain, infraConfig, ha.Type)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build router deployment: %v", err)
+	}
+	current := expected.DeepCopy()
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: expected.Namespace, Name: expected.Name}, current)
```
ironcladlou: Might as well introduce `currentRouterDeployment` while you're at it?

pkg/operator/controller/controller.go
```diff
@@ -139,7 +139,12 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
 			logrus.Errorf("failed to get infrastructure 'cluster': %v", err)
 			infraConfig = nil
 		}
-		if dnsConfig == nil || infraConfig == nil {
+		ingressConfig := &configv1.Ingress{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, ingressConfig); err != nil {
+			logrus.Fatalf("failed to get ingress 'cluster': %v", err)
```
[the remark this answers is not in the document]
Miciah: I don't think it'll crash it, but it will give the wrong behavior—`ingressConfig` should be nil, not pointing to a zero-valued `configv1.Ingress`.

pkg/apis/ingress/v1alpha1/clusteringress_types.go
```diff
@@ -83,6 +86,19 @@ type ClusterIngressStatus struct {
...
+	// Otherwise the actual ingress domain will be assigned a default value
+	// based on the cluster ingress config.
+	IngressDomain *string `json:"ingressDomain"`
+
+	// HighAvailability is the actual HA configuration for the cluster
+	// ingress.  If HighAvailability is specified in the spec, it will be
+	// used as the actual HA configuration.  Otherwise the actual HA
+	// configuration will be determined from the cluster infrastructure
+	// config.
+	HighAvailability *ClusterIngressHighAvailability `json:"highAvailability"`
```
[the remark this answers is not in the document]
Miciah: I don't think so.  I can delete the `*`.

pkg/apis/ingress/v1alpha1/clusteringress_types.go
```diff
@@ -83,6 +86,19 @@ type ClusterIngressStatus struct {
 	// corresponding to the ClusterIngress. The number of matching pods
 	// should equal the value of replicas.
 	Selector string `json:"labelSelector"`
+
+	// IngressDomain is the actual ingress domain.  If an ingress domain is
+	// specified in the spec, it will be used as the actual ingress domain.
+	// Otherwise the actual ingress domain will be assigned a default value
+	// based on the cluster ingress config.
+	IngressDomain *string `json:"ingressDomain"`
```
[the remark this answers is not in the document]
Miciah: Not sure.  If nil, it gets its value directly from the ingress config.  I thought we were supposed to use nillable types for values that could be inherited directly from cluster config, but maybe my recollection is wrong.

pkg/operator/controller/controller_dns.go
Miciah: Just wasn't sure we wanted to pass the struct itself rather than a pointer.  I can drop the `*`.

pkg/operator/controller/controller_lb.go
Miciah: You mean lift the check out of `desiredLoadBalancerService`? That would work in this case, but in the general pattern, an ensure method might need to do some clean-up—for example, `ensureDefaultCertificateForIngress` exits immediately if `deployment == nil` because it cannot do anything in that case, but it does not exit immediately if `domain == nil` because that case indicates that the desired secret is nil, meaning the ensure method must delete any current secret.

The pattern I have been following is that the ensure method returns immediately if any prerequisite values are missing, and the desired function will return nil if no resource should exist.  In the case of `ensureLoadBalancerService`, since it only ever creates and never deletes or updates, it happens to be the case that returning nil from the desired function has the same effect as immediately returning nil from the ensure method would have.

But will `ensureLoadBalancerService` handle clean-up in the future? Do we want to follow the pattern demonstrated by `ensureDefaultCertificateForIngress`?

pkg/apis/ingress/v1alpha1/clusteringress_types.go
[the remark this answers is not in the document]
Miciah: Oh, you're talking only about the status, not the spec? An empty string for `.spec.ingressDomain` is invalid (even if we don't validate it yet), right? So yeah, this should never by nil (or empty).

pkg/apis/ingress/v1alpha1/clusteringress_types.go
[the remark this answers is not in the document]
Miciah: On second thought, what value should the status have in a clusteringress that we have not reconciled and synched the status of?

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
Miciah: Fixed.  Thanks!

pkg/operator/controller/controller.go
```diff
@@ -182,11 +187,39 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
...
+	var ingressDomain *string
+	if ingress.Status.IngressDomain != nil {
+		ingressDomain = ingress.Status.IngressDomain
+	} else if ingress.Spec.IngressDomain == nil {
+		ingressDomain = &ingressConfig.Spec.Domain
+	} else if len(*ingress.Spec.IngressDomain) > 0 {
+		ingressDomain = ingress.Spec.IngressDomain
+	}
+
+	var ha ingressv1alpha1.ClusterIngressHighAvailability
```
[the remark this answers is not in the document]
Miciah: Fixed.  Thanks!

pkg/operator/controller/controller_dns.go
Miciah: Fixed.  Thanks!

pkg/operator/controller/controller_lb.go
Miciah: Removed the pointer check.

pkg/apis/ingress/v1alpha1/clusteringress_types.go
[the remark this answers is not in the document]
Miciah: Made it a non-pointer string.  Thanks!

pkg/apis/ingress/v1alpha1/clusteringress_types.go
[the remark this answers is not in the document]
Miciah: Made it a non-pointer value.  Thanks!

pkg/operator/controller/controller.go
```diff
@@ -182,11 +187,39 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
...
+		return ingressv1alpha1.CloudClusterIngressHA
+	}
+	return ingressv1alpha1.UserDefinedClusterIngressHA
+}
+
 // ensureIngressDeleted tries to delete ingress, and if successful, will remove
 // the finalizer.
-func (r *reconciler) ensureIngressDeleted(ingress *ingressv1alpha1.ClusterIngress, dnsConfig *configv1.DNS) error {
+func (r *reconciler) ensureIngressDeleted(ingress *ingressv1alpha1.ClusterIngress, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress) error {
+	var ingressDomain *string
```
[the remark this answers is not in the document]
Miciah: Done.  Thanks!

pkg/operator/controller/controller.go
```diff
@@ -284,90 +317,83 @@ func (r *reconciler) ensureRouterNamespace() error {
...
+func (r *reconciler) ensureClusterIngress(ci *ingressv1alpha1.ClusterIngress, caSecret *corev1.Secret, infraConfig *configv1.Infrastructure, dnsConfig *configv1.DNS, ingressConfig *configv1.Ingress, result *reconcile.Result) error {
+	errs := []error{}
 
-		err = r.Client.Create(context.TODO(), current)
-		if err == nil {
-			logrus.Infof("created router deployment %s/%s", current.Namespace, current.Name)
-		} else if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create router deployment %s/%s: %v", current.Namespace, current.Name, err)
-		}
+	var ingressDomain *string
```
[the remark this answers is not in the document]
Miciah: Fixed.  Thanks!

pkg/operator/controller/controller_router_deployment.go
Miciah: Deleted.

pkg/operator/controller/controller_router_deployment.go
Miciah: Done.

Miciah:
level=error msg="\t* module.bootstrap.aws_security_group.bootstrap (destroy): 1 error occurred:"
    level=error msg="\t* aws_security_group.bootstrap: timeout while waiting for state to become 'success' (timeout: 10m0s)"

/test e2e-aws

Miciah:
Bootstrap failed:

[clusterversion.json](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/130/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/447/artifacts/e2e-aws-operator/clusterversion.json) has the following:

                            "message": "Unable to retrieve available updates: unknown version 0.0.1-2019-02-20-232215",
                            "reason": "RemoteFailed",
                            "status": "False",
                            "type": "RetrievedUpdates"

/test e2e-aws-operator

pkg/operator/controller/controller.go
```diff
@@ -154,14 +160,19 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
 				return result, utilerrors.NewAggregate(errs)
 			}
 
+			ingressDomain, err := effectiveIngressDomain(ingress, ingressConfig)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("failed to find a valid unclaimed ingress domain for clusteringress %s: %v", ingress.Name, err))
```
ironcladlou: hmm, doesn't this mean in the case of an error we'll proceed to use an empty value (which itself has a meaning)? How can you distinguish between a user-specified empty value and the empty fallback we return with the error?

Should this be fatal to the reconciliation?

pkg/operator/controller/controller.go
```diff
@@ -284,90 +332,71 @@ func (r *reconciler) ensureRouterNamespace() error {
...
-	trueVar := true
-	deploymentRef := metav1.OwnerReference{
-		APIVersion: "apps/v1",
-		Kind:       "Deployment",
-		Name:       current.Name,
-		UID:        current.UID,
-		Controller: &trueVar,
-	}
+	deployment, err := r.ensureRouterDeployment(ci, ingressDomain, infraConfig, ha)
+	errs = append(errs, err)
```
ironcladlou: `if err != nil`?

Miciah: `utilerrors.NewAggregate` will ignore nil values, but I can add the check if you prefer.

pkg/operator/controller/controller.go
Miciah: You're right.  We can continue if `ingressDomain` is empty (which means we won't set the router's canonical hostname, publish DNS records, or generate a default certificate), but we must skip the `ensureIngressDelete`/`ensureClusterIngress` if `err` is not nil.  Will fix...

Miciah: Fixed.  Thanks!

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -0,0 +1,245 @@
...
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: statsSecretName,
+				},
+				Key: "statsPassword",
+			},
+		}},
+	}
+
+	// Enable prometheus metrics
+	certsSecretName := fmt.Sprintf("router-metrics-certs-%s", ci.Name)
```
ironcladlou: Just a note for followups: isn't this a hidden dependency on `manifests.RouterServiceInternal`?

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -0,0 +1,245 @@
...
+		// address.  With host networking, probes default to using the
+		// node IP address.  Using localhost avoids potential routing
+		// problems or firewall restrictions.
+		deployment.Spec.Template.Spec.Containers[0].LivenessProbe.Handler.HTTPGet.Host = "localhost"
+		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe.Handler.HTTPGet.Host = "localhost"
+	}
+
+	// Fill in the default certificate secret name.
+	secretName := fmt.Sprintf("router-certs-%s", ci.Name)
+	if ci.Spec.DefaultCertificateSecret != nil && len(*ci.Spec.DefaultCertificateSecret) > 0 {
```
ironcladlou: Hidden dependency on the default cert?

pkg/operator/controller/controller_router_deployment.go
Miciah: Yeah, there's more refactoring to be done around secrets' names.

pkg/operator/controller/controller_router_deployment.go
Miciah: Could use `routerDefaultCertificateSecretName` to get the name—would that be copacetic?

Miciah: Oh, I guess you're saying that `ensureClusterIngress` should pass the secret from `ensureDefaultCertificateForIngress` as an input to `ensureRouterDeployment`?

Miciah: That would be a circular dependency (deployment → secret → deployment), which would necessitate a bit more thought and refactoring to sort out.

ironcladlou: Yeah, not recommending any action with this PR, just wanted to note it stood out at me as something we missed and need to talk more about

pkg/operator/controller/controller.go
ironcladlou: Seems weird to keep a nil entry in the array which is a plain `[]error`, error prone if you were to check the len of the array or feed to a function that has different assumptions... err on the side of consistency?

Miciah: Fixed.

Miciah:
level=fatal msg="waiting for bootstrap-complete: timed out waiting for the condition"

[clusterversion.json](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/130/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/457/artifacts/e2e-aws-operator/clusterversion.json) has

                            "lastTransitionTime": "2019-02-21T18:10:11Z",
                            "message": "Unable to retrieve available updates: unknown version 0.0.1-2019-02-21-175430",
                            "reason": "RemoteFailed",
                            "status": "False",
                            "type": "RetrievedUpdates"

/test e2e-aws-operator

Miciah:
level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Bootstrap Ignition Config\": failed to fetch dependency of \"Common Manifests\": failed to generate asset \"DNS Config\": getting public zone for \"origin-ci-int-aws.dev.rhcloud.com\": listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400, request id: e336d8b7-361b-11e9-9816-4b39b45f47d9"

/test e2e-aws-operator

Miciah:
2019/02/21 22:00:58 Container cli in pod e2e-aws-operator completed successfully
    {"component":"entrypoint","level":"error","msg":"Entrypoint received interrupt: terminated","time":"2019-02-21T22:43:15Z"}
    2019/02/21 22:43:15 error: Process interrupted with signal interrupt, exiting in 2s ...

/test e2e-aws-operator

Miciah:
/test e2e-aws-operator
please?

pkg/operator/controller/controller_default_certificate.go
```diff
@@ -62,11 +66,18 @@ func (r *reconciler) ensureDefaultCertificateForIngress(caSecret *corev1.Secret,
 // desiredRouterDefaultCertificateSecret returns the desired default certificate
 // secret.
 func desiredRouterDefaultCertificateSecret(ca *crypto.CA, deployment *appsv1.Deployment, ci *ingressv1alpha1.ClusterIngress) (*corev1.Secret, error) {
+	// Without an ingress domain, we cannot generate a default certificate.
+	if len(ci.Status.IngressDomain) == 0 {
```
ironcladlou: If in the calling reconciliation function we're enforcing `status.ingressDomain` persistence before we even process the clusteringress further, should this instead return an error indicating we violated our invariant (i.e. a bug)?

pkg/operator/controller/controller_dns.go
```diff
@@ -34,18 +34,22 @@ func (r *reconciler) ensureDNS(ci *ingressv1alpha1.ClusterIngress, service *core
...
 		return records, nil
 	}
 
-	// TODO: This will need revisited when we stop defaulting .spec.ingressDomain
-	// and .spec.highAvailability as both can be nil but used with an effective
-	// system-provided default reported on status.
-	if ci.Spec.HighAvailability == nil || ci.Spec.HighAvailability.Type != ingressv1alpha1.CloudClusterIngressHA || ci.Spec.IngressDomain == nil {
+	// If the clusteringress has no ingress domain, we cannot configure any
+	// DNS records.
+	if len(ci.Status.IngressDomain) == 0 {
```
ironcladlou: As above re: invariant violations

ironcladlou:
Other than continuing to tweak how we handle guard clauses/control flow/invariants I think this is ready to go. What do you think?

pkg/operator/controller/controller_default_certificate.go
Miciah: We could get `len(ci.Status.IngressDomain) == 0` from an explicit empty value, so this is not necessarily an error.

Miciah:
> Other than continuing to tweak how we handle guard clauses/control flow/invariants I think this is ready to go. What do you think?

I'm feeling pretty good about it.  I haven't seen a successful test run though.

Miciah:
Rebased.

PR #131, opened 2019-02-20
Update e2e to test multiple ClusterIngress resources

danehans:
Previously, e2e only tested CRUD operations against the "default" ClusterIngress. This PR adds create/delete tests for multiple user-defined ClusterIngress resources.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -37,6 +37,10 @@ import (
 	"k8s.io/client-go/scale"
 )
 
+const (
+	ingressControllerName = "test"
+)
+
 func getClient() (client.Client, string, error) {
 	namespace, ok := os.LookupEnv("WATCH_NAMESPACE")
 	if !ok {
@@ -52,6 +56,24 @@ func getClient() (client.Client, string, error) {
 		return nil, "", fmt.Errorf("failed to create kube client: %s", err)
 	}
 	return kubeClient, namespace, nil
+}
+
+func newIngressController(name, ns, domain string) *operatorv1.IngressController {
+	repl := int32(1)
+	return &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: ns,
+		},
+		// TODO: Test needs to be infrastructure and platform aware in the very near future.
+		Spec: operatorv1.IngressControllerSpec{
+			Domain:   domain,
+			Replicas: &repl,
+			EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
+				Type: operatorv1.LoadBalancerServiceStrategyType,
+			},
+		},
+	}
 }
 
 func TestOperatorAvailable(t *testing.T) {
@@ -96,6 +118,60 @@ func TestDefaultClusterIngressExists(t *testing.T) {
 	})
 	if err != nil {
 		t.Errorf("failed to get default ClusterIngress: %v", err)
+	}
+}
+
+func TestClusterIngressControllerCreateDelete(t *testing.T) {
+	cl, ns, err := getClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ing := &operatorv1.IngressController{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "default"}, ing); err != nil {
+		t.Fatalf("failed to get default IngressController: %v", err)
+	}
+
+	dnsConfig := &configv1.DNS{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
+		t.Fatalf("failed to get DNS 'cluster': %v", err)
+	}
+
+	ing = newIngressController(ingressControllerName, ns, ingressControllerName+"."+dnsConfig.Spec.BaseDomain)
+	if err := cl.Create(context.TODO(), ing); err != nil {
+		t.Fatalf("failed to create IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
+	}
+
+	// Verify the ingress controller deployment created the specified
+	// number of pod replicas.
+	err = wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{ing.Namespace, ing.Name}, ing); err != nil {
+			return false, nil
+		}
+		if ing.Status.AvailableReplicas != *ing.Spec.Replicas {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to reconcile IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
+	}
+
+	if err := cl.Delete(context.TODO(), ing); err != nil {
+		t.Fatalf("failed to delete IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
+	}
+
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Get(context.TODO(), types.NamespacedName{ing.Namespace, ing.Name}, ing); err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to finalize IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
 	}
 }
 


test/e2e/operator_test.go
```diff
@@ -56,6 +60,33 @@ func getClient() (client.Client, string, error) {
 	return kubeClient, namespace, nil
 }
 
+func newClusterIngress(name, ns, domain string) *ingressv1alpha1.ClusterIngress {
+	ci := &ingressv1alpha1.ClusterIngress{
```
ramr: Could just `return &ingressv1alpha1.ClusterIngress{  ... }`  as `ci` is not really used locally.

test/e2e/operator_test.go
```diff
@@ -101,6 +132,67 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+}
+
+func TestMultipleClusterIngressDelete(t *testing.T) {
+	cl, ns, err := getClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ing0 := &ingressv1alpha1.ClusterIngress{}
+	ing1 := &ingressv1alpha1.ClusterIngress{}
```
ramr: Could probably initialize the name+namespace here in `ClusterIngress` - wee bit more explicit that way and then [re-]use it for `types.NamespaceName` below  in the `cl.Get` call. Not a big deal either way.

test/e2e/operator_test.go
danehans: Good catch. I had a few diff variations of the func and forgot to simply return the ptr on my final iteration. Thx!

ramr:
/retest

Failed with `msg="waiting for bootstrap-complete: timed out waiting for the condition"`

danehans:
@ramr tests are passing.

test/e2e/operator_test.go
```diff
@@ -56,6 +60,31 @@ func getClient() (client.Client, string, error) {
...
+			Kind:       "ClusterIngress",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: ns,
+		},
+		Spec: ingressv1alpha1.ClusterIngressSpec{
+			IngressDomain: &domain,
+			HighAvailability: &ingressv1alpha1.ClusterIngressHighAvailability{
+				Type: ingressv1alpha1.CloudClusterIngressHA,
```
ironcladlou: Might want a TODO here that the test needs to be infrastructure/platform aware in the very near future (maybe when we move it over to the origin extended tests).

test/e2e/operator_test.go
```diff
@@ -56,6 +60,31 @@ func getClient() (client.Client, string, error) {
 	return kubeClient, namespace, nil
 }
 
+func newClusterIngress(name, ns, domain string) *ingressv1alpha1.ClusterIngress {
+	return &ingressv1alpha1.ClusterIngress{
+		TypeMeta: metav1.TypeMeta{
```
ironcladlou: For future ref, you don't need to specify TypeMeta.

test/e2e/operator_test.go
```diff
@@ -56,6 +60,31 @@ func getClient() (client.Client, string, error) {
...
+			Name:      name,
+			Namespace: ns,
+		},
+		Spec: ingressv1alpha1.ClusterIngressSpec{
+			IngressDomain: &domain,
+			HighAvailability: &ingressv1alpha1.ClusterIngressHighAvailability{
+				Type: ingressv1alpha1.CloudClusterIngressHA,
+			},
+			Replicas: int32(2),
+			NodePlacement: &ingressv1alpha1.NodePlacement{
```
ironcladlou: Should probably rely on defaults here unless the test is actually intended to exercise placement

ironcladlou:
Due to flake paranoia, would like to see a few consecutive successful runs before tagging.

/test e2e-aws

danehans:
Is anything else needed for this PR to merge?

ironcladlou:
Is the fact it's passing indicating a false-positive given https://github.com/openshift/cluster-ingress-operator/pull/146? I guess the current assertions only go as far as "if the clusteringress is finalized, we successfully deleted" but there aren't any tests for asserting we actually cleaned up. This seems better than before though. @Miciah @pravisankar PTAL and feel free to tag

test/e2e/operator_test.go
```diff
@@ -101,6 +122,80 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+	ing0 := &ingressv1alpha1.ClusterIngress{}
+	ing1 := &ingressv1alpha1.ClusterIngress{}
+	ingresses := []*ingressv1alpha1.ClusterIngress{ing0, ing1}
+	for i, ing := range ingresses {
+		name := clusterIngressName + strconv.Itoa(i)
+		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: name}, ing); err != nil {
+			t.Errorf("failed to get ClusterIngress %s/%s: %v", ns, name, err)
+		}
+		err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+			if err := cl.Delete(context.TODO(), ing); err != nil {
```
pravisankar: Are we depending on the test case order here i.e TestMultipleClusterIngressCreate() is called before TestMultipleClusterIngressDelete()? If yes, why not combine these tests as a singe test case?

danehans:
@ironcladlou I took @pravisankar advice and consolidated the 2 tests into one that tests create, read and delete for multiple clusteringresses. Now that I think of it, maybe I should add an update test. Before pushing the changes, should this test added to origin or api instead?

Miciah:
> Now that I think of it, maybe I should add an update test.

I think `TestClusterIngressUpdate` covers updates adequately, unless there's a specific interaction between clusteringresses during updates that we need to test.

> Before pushing the changes, should this test added to origin or api instead?

Possibly.  Specifically, [the scoped router test](https://github.com/openshift/origin/blob/c6655ab03317c04abee511e6696c201afb56780c/test/extended/router/scoped.go#L62-L104) could be modified by replacing [the router-scoped fixture](https://github.com/openshift/origin/blob/c6655ab03317c04abee511e6696c201afb56780c/test/extended/testdata/router-scoped.yaml) with the clusteringress/ingresscontroller resource.  However, I believe it would be worth merging this PR now to get the test coverage sooner and then work on adapting the test in origin.

test/e2e/operator_test.go
```diff
@@ -101,6 +122,80 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+	ing0 := newClusterIngress(clusterIngressName+"0", ns, clusterIngressName+"0"+"."+dnsConfig.Spec.BaseDomain)
+	ing1 := newClusterIngress(clusterIngressName+"1", ns, clusterIngressName+"1"+"."+dnsConfig.Spec.BaseDomain)
+	ingresses := []*ingressv1alpha1.ClusterIngress{ing0, ing1}
+	for _, ing := range ingresses {
+		err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+			if err := cl.Create(context.TODO(), ing); err != nil {
+				if errors.IsAlreadyExists(err) {
+					return true, nil
+				} else {
+					return false, nil
```
Miciah: If I may be picky, I'd suggest simplifying as follows:
```go
			if err := cl.Create(context.TODO(), ing); err != nil {
				if errors.IsAlreadyExists(err) {
					return true, nil
				}
				return false, nil
			}
```
And similar for the other `wait.PollImmediate` callbacks.

test/e2e/operator_test.go
```diff
@@ -101,6 +122,72 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+
+	dnsConfig := &configv1.DNS{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
+		t.Fatalf("failed to get DNS 'cluster': %v", err)
+	}
+
+	ing := newIngressController(ingressControllerName, ns, ingressControllerName+"."+dnsConfig.Spec.BaseDomain)
+	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+		if err := cl.Create(context.TODO(), ing); err != nil {
+			if errors.IsAlreadyExists(err) {
```
ironcladlou: Not sure what we gain by polling here; if the instance already exists, we don't know where it came from and shouldn't test it... what about just calling Create and failing if any error is returned?

test/e2e/operator_test.go
```diff
@@ -101,6 +122,72 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to reconcile IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
+	}
+
+	// Verify the ingress controller can be deleted.
+	err = wait.PollImmediate(1*time.Second, 15*time.Second, func() (bool, error) {
+		if err := cl.Delete(context.TODO(), ing); err != nil {
```
pravisankar: Similar to above comment, I don't think polling is needed here

ironcladlou:
There are still multiple: the default, and the user defined one created during the test. Two user defined would make three, and I think the default plus a user defined one is enough to catch the issues this test found in the first place...

For more robust scenarios (like route/namespace selectors) how about we get into the origin extended tests and consider replacing stuff like some of the tests in `scoped.go`?

test/e2e/operator_test.go
```diff
@@ -101,6 +122,72 @@ func TestDefaultClusterIngressExists(t *testing.T) {
...
+		if err := cl.Create(context.TODO(), ing); err != nil {
+			if errors.IsAlreadyExists(err) {
+				return true, nil
+			}
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to create IngressController %s/%s: %v", ing.Namespace, ing.Name, err)
```
Miciah: Should be `Fatalf`.

test/e2e/operator_test.go
Miciah: We need polling to make sure it's gone, but it would make sense to `Delete` outside the `PollImmediate` and do a `Get` inside the `PollImmediate`.

Miciah:
This PR needs to be rebased now that https://github.com/openshift/cluster-ingress-operator/pull/157 has merged.

ironcladlou:
Needs rebased

test/e2e/operator_test.go
danehans: Good point. I'll update.

test/e2e/operator_test.go
danehans: I'll update. I thought we wanted to Errorf on create so we can still proceed to deleting.

test/e2e/operator_test.go
danehans: @pravisankar I thought the poll was needed b/c the delete is not immediate due to the finalizer.
@Miciah We perform a Get at L165 and compare spec<>status replicas. Why is a 2nd get needed?

test/e2e/operator_test.go
danehans: I think your point can be applied to `IsAlreadyExists for the delete operation too.

test/e2e/operator_test.go
Miciah: I mean that as part of the test, we should make sure that the `Delete` call actually eventually deletes the resource.  Does that seem reasonable? If so, we can either repeat the `Delete` call till the resource is gone (in which case we should be checking here for a not-found error value), or we can call `Delete` once and then repeat a `Get` call till the resource is gone.

danehans: @Miciah thanks for the additional explanation. I now understand your perspective. I'll update the `Delete`.

ironcladlou:
@Miciah @pravisankar looks like it's ready for more review

PR #132, opened 2019-02-22
Process as much as possible during reconcile and requeue in case of any errors

pravisankar:
- Do not use Result.Requeue or Result.RequeueAfter when returning errors
as controller will ignore these Requeue/RequeueAfter fields and the current
item is automatically rate limited.

- Some more cleanup:
  * Caller will invoke sub methods only when the prereq args are valid.
  * Tolerate some failures and try to create/update as many related resources
    as possible during reconcile.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -3,7 +3,6 @@ package controller
 import (
 	"context"
 	"fmt"
-	"time"
 
 	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
 	"github.com/openshift/cluster-ingress-operator/pkg/dns"
@@ -17,7 +16,6 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -53,8 +51,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = c.Watch(&source.Kind{Type: &ingressv1alpha1.ClusterIngress{}}, &handler.EnqueueRequestForObject{})
-	if err != nil {
+	if err := c.Watch(&source.Kind{Type: &ingressv1alpha1.ClusterIngress{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return nil, err
 	}
 	return c, nil
@@ -79,74 +76,66 @@ type reconciler struct {
 // namespace, and will do all the work to ensure the clusteringress is in the
 // desired state.
 func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	errs := []error{}
 	result := reconcile.Result{}
-
-	// TODO: Should this be another controller?
-	defer func() {
-		err := r.syncOperatorStatus()
-		if err != nil {
-			log.Error(err, "failed to sync operator status")
-		}
-	}()
 
 	log.Info("reconciling", "request", request)
 
 	// Get the current ingress state.
 	ingress := &ingressv1alpha1.ClusterIngress{}
-	err := r.Client.Get(context.TODO(), request.NamespacedName, ingress)
-	if err != nil {
+	if err := r.Client.Get(context.TODO(), request.NamespacedName, ingress); err != nil {
 		if errors.IsNotFound(err) {
 			// This means the ingress was already deleted/finalized and there are
 			// stale queue entries (or something edge triggering from a related
 			// resource that got deleted async).
 			log.Info("clusteringress not found; reconciliation will be skipped", "request", request)
 		} else {
-			// Try again later.
-			log.Error(err, "failed to get clusteringress", "request", request)
-			result.RequeueAfter = 10 * time.Second
+			errs = append(errs, fmt.Errorf("failed to get clusteringress %q: %v", request, err))
 		}
 		ingress = nil
 	}
 
 	caSecret, err := r.ensureRouterCACertificateSecret()
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure CA secret: %v", err)
+		errs = append(errs, fmt.Errorf("failed to ensure CA secret: %v", err))
+	} else {
+		// TODO: This should be in a different reconciler as it's independent of an
+		// individual ingress. We only really need to trigger this when a
+		// clusteringress is added or deleted...
+		// Find all clusteringresses to compute CA states.
+		ingresses := &ingressv1alpha1.ClusterIngressList{}
+		if err := r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses); err != nil {
+			errs = append(errs, fmt.Errorf("failed to list clusteringresses in namespace %s: %v", r.Namespace, err))
+		} else if err := r.ensureRouterCAConfigMap(caSecret, ingresses.Items); err != nil {
+			errs = append(errs, fmt.Errorf("failed to ensure router ca config map: %v", err))
+		}
 	}
 
-	// Collect errors as we go.
-	errs := []error{}
-
-	// Only reconcile the ingress itself if it exists.
 	if ingress != nil {
 		dnsConfig := &configv1.DNS{}
 		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
-			log.Error(err, "failed to get dns 'cluster'")
+			errs = append(errs, fmt.Errorf("failed to get dns 'cluster': %v", err))
 			dnsConfig = nil
 		}
 		infraConfig := &configv1.Infrastructure{}
 		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
-			log.Error(err, "failed to get infrastructure 'cluster'")
+			errs = append(errs, fmt.Errorf("failed to get infrastructure 'cluster': %v", err))
 			infraConfig = nil
 		}
 		ingressConfig := &configv1.Ingress{}
 		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, ingressConfig); err != nil {
-			log.Error(err, "failed to get ingress 'cluster'")
+			errs = append(errs, fmt.Errorf("failed to get ingress 'cluster': %v", err))
 			ingressConfig = nil
 		}
 
-		if dnsConfig == nil || infraConfig == nil || ingressConfig == nil {
-			// For now, if the cluster configs are unavailable, defer reconciliation
-			// because weaving conditionals everywhere to deal with various nil states
-			// is too complicated. It doesn't seem too risky to rely on the invariant
-			// of the cluster config being available.
-			result.RequeueAfter = 10 * time.Second
-		} else {
-			// Ensure we have all the necessary scaffolding on which to place router
-			// instances.
-			err = r.ensureRouterNamespace()
-			if err != nil {
-				errs = append(errs, err)
-				return result, utilerrors.NewAggregate(errs)
+		// For now, if the cluster configs are unavailable, defer reconciliation
+		// because weaving conditionals everywhere to deal with various nil states
+		// is too complicated. It doesn't seem too risky to rely on the invariant
+		// of the cluster config being available.
+		if dnsConfig != nil && infraConfig != nil && ingressConfig != nil {
+			// Ensure we have all the necessary scaffolding on which to place router instances.
+			if err := r.ensureRouterNamespace(); err != nil {
+				errs = append(errs, fmt.Errorf("failed to ensure router namespace: %v", err))
 			}
 
 			if err := r.enforceEffectiveIngressDomain(ingress, ingressConfig); err != nil {
@@ -160,23 +149,17 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				}
 			} else {
 				// Handle everything else.
-				if err := r.ensureClusterIngress(ingress, caSecret, infraConfig, dnsConfig, &result); err != nil {
+				if err := r.ensureClusterIngress(ingress, caSecret, dnsConfig, infraConfig); err != nil {
 					errs = append(errs, fmt.Errorf("failed to ensure clusteringress: %v", err))
 				}
 			}
 		}
 	}
 
-	// TODO: This should be in a different reconciler as it's independent of an
-	// individual ingress. We only really need to trigger this when a
-	// clusteringress is added or deleted...
-	// Find all clusteringresses to compute CA states.
-	ingresses := &ingressv1alpha1.ClusterIngressList{}
-	err = r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to list clusteringresses in namespace %s: %v", r.Namespace, err)
+	// TODO: Should this be another controller?
+	if err := r.syncOperatorStatus(); err != nil {
+		errs = append(errs, fmt.Errorf("failed to sync operator status: %v", err))
 	}
-	errs = append(errs, r.ensureRouterCAConfigMap(caSecret, ingresses.Items))
 
 	return result, utilerrors.NewAggregate(errs)
 }
@@ -247,14 +230,12 @@ func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.Cluste
 // ensureIngressDeleted tries to delete ingress, and if successful, will remove
 // the finalizer.
 func (r *reconciler) ensureIngressDeleted(ingress *ingressv1alpha1.ClusterIngress, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) error {
-	err := r.finalizeLoadBalancerService(ingress, dnsConfig)
-	if err != nil {
+	if err := r.finalizeLoadBalancerService(ingress, dnsConfig); err != nil {
 		return fmt.Errorf("failed to finalize load balancer service for %s: %v", ingress.Name, err)
 	}
 	log.Info("finalized load balancer service for ingress", "namespace", ingress.Namespace, "name", ingress.Name)
 
-	err = r.ensureRouterDeleted(ingress)
-	if err != nil {
+	if err := r.ensureRouterDeleted(ingress); err != nil {
 		return fmt.Errorf("failed to delete deployment for ingress %s: %v", ingress.Name, err)
 	}
 	log.Info("deleted deployment for ingress", "namespace", ingress.Namespace, "name", ingress.Name)
@@ -263,8 +244,7 @@ func (r *reconciler) ensureIngressDeleted(ingress *ingressv1alpha1.ClusterIngres
 	updated := ingress.DeepCopy()
 	if slice.ContainsString(ingress.Finalizers, ClusterIngressFinalizer) {
 		updated.Finalizers = slice.RemoveString(updated.Finalizers, ClusterIngressFinalizer)
-		err = r.Client.Update(context.TODO(), updated)
-		if err != nil {
+		if err := r.Client.Update(context.TODO(), updated); err != nil {
 			return fmt.Errorf("failed to remove finalizer from clusteringress %s: %v", ingress.Name, err)
 		}
 	}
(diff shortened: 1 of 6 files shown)


pravisankar:

pkg/operator/controller/controller.go
```diff
@@ -112,58 +109,42 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
...
-				}
-			} else {
-				// Handle everything else.
-				if err := r.ensureRouterForIngress(ingress, caSecret, infraConfig, dnsConfig, &result); err != nil {
-					errs = append(errs, fmt.Errorf("failed to ensure clusteringress: %v", err))
-				}
+		// Ensure we have all the necessary scaffolding on which to place router instances.
+		if err = r.ensureRouterNamespace(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to ensure router namespace: %v", err))
+		}
```
Miciah: Neither `ensureIngressDeleted` nor `ensureRouterForIngress` can do any work if the namespace does not exist, so we can skip them if `ensureRouterNamespace` fails.

pkg/operator/controller/controller.go
```diff
@@ -112,58 +109,42 @@ func (r *reconciler) reconcile(request reconcile.Request) (reconcile.Result, err
...
 		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
-			logrus.Errorf("failed to get infrastructure 'cluster': %v", err)
+			errs = append(errs, fmt.Errorf("failed to get infrastructure 'cluster': %v", err))
 			infraConfig = nil
 		}
-		if dnsConfig == nil || infraConfig == nil {
-			// For now, if the cluster configs are unavailable, defer reconciliation
-			// because weaving conditionals everywhere to deal with various nil states
-			// is too complicated. It doesn't seem too risky to rely on the invariant
-			// of the cluster config being available.
```
Miciah: I've been an advocate of doing whatever we can with what we have, but based on Dan's feedback on https://github.com/openshift/cluster-ingress-operator/pull/130, I think Dan wants to keep this invariant for now and avoid doing all those `dnsConfig == nil` and `infraConfig == nil` checks deeper in the controller code.

pkg/operator/controller/controller_ca_configmap.go
```diff
@@ -67,7 +70,7 @@ func (r *reconciler) ensureRouterCAConfigMap(secret *corev1.Secret, ingresses []
 
 // desiredRouterCAConfigMap returns the desired router CA configmap.
 func desiredRouterCAConfigMap(secret *corev1.Secret, ingresses []ingressv1alpha1.ClusterIngress) (*corev1.ConfigMap, error) {
-	if !shouldPublishRouterCA(ingresses) {
+	if secret == nil || !shouldPublishRouterCA(ingresses) {
```
Miciah: I don't think you need to check `secret == nil` both here and in `ensureRouterCAConfigMap`.  The pattern I have been following is that `ensureFoo` is responsible for checking prerequisite values and will return immediately if any are missing, so `desiredFoo` can safely assume the prerequisites are present.  Does that seem reasonable?

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -43,6 +46,10 @@ func serviceMonitorName(ci *ingressv1alpha1.ClusterIngress) types.NamespacedName
 }
 
 func desiredServiceMonitor(ci *ingressv1alpha1.ClusterIngress, svc *corev1.Service, deployment *appsv1.Deployment) *unstructured.Unstructured {
+	if svc == nil || deployment == nil {
+		return nil
+	}
```
Miciah: I don't think you need to check `svc == nil || deployment == nil` both here and in `ensureServiceMonitor`.

ironcladlou:
/hold

Need to merge https://github.com/openshift/cluster-ingress-operator/pull/130 first

pkg/operator/controller/controller_ca_configmap.go
pravisankar: I don't think we want to enforce restrictions that this method is only called by ensureXYZ method. Each method handling its own basic checks will be fool proof during refactors and makes it reusable like  `desiredDNSRecords`.

pkg/operator/controller/controller_service_monitor.go
pravisankar: Same reason as above.

pkg/operator/controller/controller.go
pravisankar: `ensureRouterNamespace` not only creates router namesapce but also creates service account, cluster role and cluster rolebindings. If namespace creation succeeds but fails on other cases, we could definitely perform work during ingress deletion and creation. If namespace itself fails, both `ensureIngressDeleted` and `ensureRouterForIngress` will bailout as expected.

pkg/operator/controller/controller.go
pravisankar: I think we discussed this in our stand up call, when dnsConfig/infraConfig is not present we are able to perform other useful stuff. During my test runs, I have seen dnsConfig not found in the first iterator of reconciliation, we did ensure other resources and the request was queued and in the second iteration, ensureDNS succeeded. 
As Dan pointed out earlier, I have been seeing very aggressive retries by the rate limiter. Probably we need to introduce our custom rate limiter to mitigate this issue.

pravisankar: Looking at the latest changes, ingressConfig is used in too many places and it is getting ugly and hard to reason if these nil checks are done later in the methods when ever it is necessary. So I will revert this to the current scheme of returning early when any of the dns/infra/ingress configs are not present.

pkg/operator/controller/controller.go
```diff
@@ -78,14 +76,15 @@ type reconciler struct {
...
+	errs := []error{}
 
 	// TODO: Should this be another controller?
 	defer func() {
-		err := r.syncOperatorStatus()
-		if err != nil {
-			log.Error(err, "failed to sync operator status")
+		if err := r.syncOperatorStatus(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to sync operator status: %v", err))
+			finalErr = utilerrors.NewAggregate(errs)
```
Miciah: This deferred function is becoming more entangled with the function's main logic.  Now that you have removed all but the final `return` in `Reconcile`, we could move the `syncOperatorStatus` into the main logic flow, which would simplify the error handling.

pkg/operator/controller/controller.go
```diff
@@ -101,64 +100,56 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-			if err != nil {
-				errs = append(errs, err)
-				return result, utilerrors.NewAggregate(errs)
+		// For now, if the cluster configs are unavailable, defer reconciliation
+		// because weaving conditionals everywhere to deal with various nil states
+		// is too complicated. It doesn't seem too risky to rely on the invariant
+		// of the cluster config being available.
+		if dnsConfig != nil && infraConfig != nil && ingressConfig != nil {
+			// Ensure we have all the necessary scaffolding on which to place router instances.
+			if err = r.ensureRouterNamespace(); err != nil {
```
Miciah: It's an extremely minor concern, but I think `:=` would be preferable here because if there is no reason to change the variable in the outer scope, we shouldn't.

pkg/operator/controller/controller.go
```diff
@@ -101,64 +100,56 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				errs = append(errs, fmt.Errorf("failed to ensure router namespace: %v", err))
 			}
 
 			if err := r.enforceEffectiveIngressDomain(ingress, ingressConfig); err != nil {
 				errs = append(errs, fmt.Errorf("failed to enforce the effective ingress domain for clusteringress %s: %v", ingress.Name, err))
-			} else if ingress.DeletionTimestamp != nil {
+			}
```
Miciah: We need this to be an `else if` so that we don't try to ensure anything with `.ingress.Status.IngressDomain` falsely empty.

pkg/operator/controller/controller.go
```diff
@@ -327,21 +321,21 @@ func (r *reconciler) ensureRouterNamespace() error {
...
+// ensureClusterIngress ensures all necessary router resources exist for a given clusteringress.
+func (r *reconciler) ensureClusterIngress(ci *ingressv1alpha1.ClusterIngress, caSecret *corev1.Secret, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) error {
 	errs := []error{}
 
 	ha := effectiveHighAvailability(ci, infraConfig)
 
 	deployment, err := r.ensureRouterDeployment(ci, infraConfig, ha)
 	if err != nil {
-		errs = append(errs, err)
+		errs = append(errs, fmt.Errorf("failed to ensure router deployment for %s: %v", ci.Name, err))
```
Miciah: Set `deployment = nil` here?

pkg/operator/controller/controller.go
```diff
@@ -101,64 +100,56 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			errs = append(errs, fmt.Errorf("failed to get clusteringress %q: %v", request, err))
 		}
 		ingress = nil
 	}
 
 	caSecret, err := r.ensureRouterCACertificateSecret()
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure CA secret: %v", err)
+		errs = append(errs, fmt.Errorf("failed to ensure CA secret: %v", err))
+		caSecret = nil
```
ironcladlou: Throughout: if `err != nil` then we can't rely on the result of the function (`caSecret` in this case).

See https://github.com/openshift/cluster-ingress-operator/pull/136 for an approach which avoids this

ironcladlou: And also `caSecret = nil` is us basically making up a result for the function which returned an error; the actual result is "don't know if caSecret is nil"

Miciah: I don't think of `caSecret == nil` as meaning necessarily that we know there is no secret; rather, it means the input is unavailable.

pravisankar:
@openshift/sig-network-edge ready for review/merge, PTAL

pkg/operator/controller/controller.go
ironcladlou: You literally can't do anything right now without the deployment, so why not just keep returning early? And again, this is using the result of a fn call in the presence of error...

pkg/operator/controller/controller.go
```diff
@@ -156,9 +136,9 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
 					errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
 				}
-			} else {
+			} else if caSecret != nil {
```
ironcladlou: We only care about the CA secret as part of ensureDefaultCertificateForIngress and only then in the default case, so it seems very weird to treat it here as a "global invariant" that would prohibit doing the main reconcile... it seems like the factoring is just wrong here. It could be created lazily for the first clusteringress that needs it down in ensureClusterIngress and then looked up again later as needed during publishing.

pravisankar: Fixed

pkg/operator/controller/controller.go
```diff
@@ -172,9 +152,17 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	ingresses := &ingressv1alpha1.ClusterIngressList{}
 	err = r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to list clusteringresses in namespace %s: %v", r.Namespace, err)
+		errs = append(errs, fmt.Errorf("failed to list clusteringresses in namespace %s: %v", r.Namespace, err))
+	} else if caSecret != nil {
+		if err := r.ensureRouterCAConfigMap(caSecret, ingresses.Items); err != nil {
+			errs = append(errs, fmt.Errorf("failed to ensure router ca config map: %v", err))
```
Miciah: What do you think about moving `ensureRouterCAConfigMap` up with `ensureRouterCACertificateSecret` (on which it depends)? Something like the following:
```go
	caSecret, err := r.ensureRouterCACertificateSecret()
	if err != nil {
		// ...
	} else {
		ingresses := &ingressv1alpha1.ClusterIngressList{}
		if err := r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses); err != nil {
			// ...
		} else if err := r.ensureRouterCAConfigMap(caSecret, ingresses.Items); err != nil {
			// ...
```

pkg/operator/controller/controller.go
```diff
@@ -327,70 +315,57 @@ func (r *reconciler) ensureRouterNamespace() error {
...
+			UID:        deployment.UID,
+			Controller: &trueVar,
 		}
-	}
 
-	if err := r.ensureDefaultCertificateForIngress(caSecret, deployment, ci); err != nil {
-		errs = append(errs, fmt.Errorf("failed to ensure default certificate: %v", err))
-	}
+		lbService, err := r.ensureLoadBalancerService(ci, deploymentRef, infraConfig, ha)
+		if err != nil {
```
Miciah: Might as well use `if lbService, err := ...; err != nil {` form.

pkg/operator/controller/controller.go
```diff
@@ -327,70 +315,57 @@ func (r *reconciler) ensureRouterNamespace() error {
...
+			errs = append(errs, fmt.Errorf("failed to ensure load balancer service for %s: %v", ci.Name, err))
+		} else if err = r.ensureDNS(ci, lbService, dnsConfig, ha); err != nil {
```
Miciah: Could use `:=`  here.

pkg/operator/controller/controller_default_certificate.go
```diff
@@ -86,7 +81,7 @@ func desiredRouterDefaultCertificateSecret(ca *crypto.CA, deployment *appsv1.Dep
 		return nil, fmt.Errorf("failed to encode certificate: %v", err)
 	}
 
-	name := routerDefaultCertificateSecretName(ci, deployment.Namespace)
+	name := routerDefaultCertificateSecretName(ci, namespace)
```
Miciah: Could use `deploymentRef.Namespace` to obviate the separate `namespace` parameter.

pkg/operator/controller/controller.go
```diff
@@ -327,70 +315,55 @@ func (r *reconciler) ensureRouterNamespace() error {
...
-		return utilerrors.NewAggregate(errs)
+		errs = append(errs, fmt.Errorf("failed to ensure router deployment for %s: %v", ci.Name, err))
 	}
 
-	if lbService, err := r.ensureLoadBalancerService(ci, deployment, infraConfig, ha); err != nil {
-		errs = append(errs, fmt.Errorf("failed to ensure load balancer service: %v", err))
-	} else {
-		if err = r.ensureDNS(ci, lbService, dnsConfig, ha); err != nil {
-			errs = append(errs, fmt.Errorf("failed to ensure DNS: %v", err))
+	if deployment != nil {
```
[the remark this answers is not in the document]
Miciah: Or better:
```go
	if deployment, err := r.ensureRouterDeployment(ci, infraConfig, ha); err != nil {
		// ...
	} else {
```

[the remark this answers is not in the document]

pkg/operator/controller/controller.go
pravisankar: fixed

pkg/operator/controller/controller.go

pkg/operator/controller/controller_default_certificate.go
pravisankar: Unfortunately OwnerReference does not contain Namespace field.

Miciah: Whoops! Sorry! Hmm, I think that's why I decided to pass `deployment` before, to avoid needing to pass two arguments (`deploymentRef` and `namespace`).

pkg/operator/controller/controller.go
```diff
@@ -101,52 +92,52 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure CA secret: %v", err)
+		errs = append(errs, fmt.Errorf("failed to ensure CA secret: %v", err))
+	} else {
+		// TODO: This should be in a different reconciler as it's independent of an
+		// individual ingress. We only really need to trigger this when a
+		// clusteringress is added or deleted...
+		// Find all clusteringresses to compute CA states.
+		ingresses := &ingressv1alpha1.ClusterIngressList{}
+		if err = r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses); err != nil {
```
Miciah: Could use `:=` here.

pkg/operator/controller/controller.go
```diff
@@ -478,15 +445,15 @@ func (r *reconciler) ensureMetricsIntegration(ci *ingressv1alpha1.ClusterIngress
 		}
 	}
 
+	if _, err = r.ensureServiceMonitor(ci, svc, deploymentRef); err != nil {
```
Miciah: Could use `:=` here.

ironcladlou:
/hold

I still see problems here

pkg/operator/controller/controller.go
```diff
@@ -79,74 +76,66 @@ type reconciler struct {
...
+		errs = append(errs, fmt.Errorf("failed to ensure CA secret: %v", err))
+	} else {
+		// TODO: This should be in a different reconciler as it's independent of an
+		// individual ingress. We only really need to trigger this when a
+		// clusteringress is added or deleted...
+		// Find all clusteringresses to compute CA states.
+		ingresses := &ingressv1alpha1.ClusterIngressList{}
+		if err := r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, ingresses); err != nil {
+			errs = append(errs, fmt.Errorf("failed to list clusteringresses in namespace %s: %v", r.Namespace, err))
+		} else if err := r.ensureRouterCAConfigMap(caSecret, ingresses.Items); err != nil {
```
ironcladlou: I don't understand why `ensureRouterCAConfigMap` was moved. Doesn't the outcome of the *current ingress reconcile* affect the outcome of `ensureRouterCAConfigMap`?

ironcladlou:
I'm getting concerned about the churn and scope of changes in this one. Can we break out individual fixes into separate PRs which are as narrowly defined as possible?

pkg/operator/controller/controller.go
Miciah: It shouldn't.  We do need to mind the logic around `ensureRouterCAConfigMap` if we change the `.spec.defaultCertificateSecret` defaulting logic, but it shouldn't be an issue otherwise.

The only change in behavior from moving `ensureRouterCAConfigMap` is that we publish the `router-ca` configmap before we reconcile the clusteringress.  With the other changes, we publish the `router-ca` configmap even if the clusteringress reconciliation fails, so this only changes the order in which the objects are created.

pkg/operator/controller/controller_dns.go
```diff
@@ -35,11 +35,6 @@ func (r *reconciler) ensureDNS(ci *ingressv1alpha1.ClusterIngress, service *core
 func desiredDNSRecords(ci *ingressv1alpha1.ClusterIngress, service *corev1.Service, dnsConfig *configv1.DNS) ([]*dns.Record, error) {
 	records := []*dns.Record{}
 
-	// If no service exists, no DNS records should exist.
```
ironcladlou: Rationale for removing this? `ensureLoadBalancerService` still returns `nil` if no LB service should be created, which means `service` here can be `nil`...

ironcladlou:
Thanks for sticking with this one!

/lgtm

pravisankar:
```
level=error msg="Error: Error applying plan:"
level=error
level=error msg="5 errors occurred:"
level=error msg="\t* module.vpc.aws_route_table_association.private_routing[2]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.private_routing.2: timeout while waiting for state to become 'success' (timeout: 5m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_route.to_nat_gw[5]: 1 error occurred:"
level=error msg="\t* aws_route.to_nat_gw.5: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_route.to_nat_gw[2]: 1 error occurred:"
level=error msg="\t* aws_route.to_nat_gw.2: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_route.to_nat_gw[4]: 1 error occurred:"
level=error msg="\t* aws_route.to_nat_gw.4: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"
level=error
level=error
level=error msg="\t* module.vpc.aws_route_table_association.private_routing[5]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.private_routing.5: timeout while waiting for state to become 'success' (timeout: 5m0s)"
```

/retest

PR #133, opened 2019-02-22
Improves zapr logger contextual information

danehans:
PR #126 was merged before implementing final touches to improve contextual information of log messages. This PR prevents reporting the wrapper code as the caller and reduces the verbosity of logger names.

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -14,13 +14,15 @@ var Logger logr.Logger
 
 func init() {
 	// Build a zap development logger.
-	zapLog, err := zap.NewDevelopment()
+	zapLogger, err := zap.NewDevelopment(zap.AddCallerSkip(1))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/log/log.go
```diff
@@ -14,13 +14,15 @@ var Logger logr.Logger
 
 func init() {
 	// Build a zap development logger.
-	zapLog, err := zap.NewDevelopment()
+	zapLogger, err := zap.NewDevelopment(zap.AddCallerSkip(1))
 	if err != nil {
 		panic(fmt.Sprintf("error building logger: %v", err))
 	}
+	defer zapLogger.Sync()
```
ironcladlou: Food for thought: deferred calls aren't honored if `main.main` exits, right? Seems like another reason to manage logging setup/cleanup closer to the entry point and outside `init()` (to increase the odds of successfully flushing prior to a crash)

PR #135, opened 2019-02-22
dns/aws: include operator version in UserAgent for AWS API calls

abhinavdahiya:
As example of this terraform-provider-aws [1]

[1]: https://github.com/terraform-providers/terraform-provider-aws/blob/98f11aecb078d82fd7d9b59e5307874797d8f359/aws/config.go#L420

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -8,11 +8,13 @@ import (
 
 	"github.com/openshift/cluster-ingress-operator/pkg/dns"
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+	"github.com/openshift/cluster-ingress-operator/version"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/elb"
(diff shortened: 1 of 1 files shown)


pkg/dns/aws/dns.go
```diff
@@ -78,6 +80,10 @@ func NewManager(config Config) (*Manager, error) {
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create AWS client session: %v", err)
 	}
+	sess.Handlers.Build.PushBackNamed(request.NamedHandler{
+		Name: "openshift.io/cluster-ingress-operator",
```
ironcladlou: can you make this `openshift.io/ingress-operator`

pkg/dns/aws/dns.go
```diff
@@ -78,6 +80,10 @@ func NewManager(config Config) (*Manager, error) {
...
+		Fn:   request.MakeAddToUserAgentHandler("openshift.io cluster-ingress-operator", version.OperatorVersion),
```
ironcladlou: "openshift.io ingress-operator" please

pkg/dns/aws/dns.go
abhinavdahiya: done with https://github.com/openshift/cluster-ingress-operator/compare/a60a53d976a6f8b9312678fdd5582b4716373a5c..b8dfa91aa6707cd3a3af790e4e6524821973034d

pkg/dns/aws/dns.go
abhinavdahiya: done with https://github.com/openshift/cluster-ingress-operator/compare/a60a53d976a6f8b9312678fdd5582b4716373a5c..b8dfa91aa6707cd3a3af790e4e6524821973034d

ironcladlou:
Thanks!
/lgtm
/retest

PR #136, opened 2019-02-22
controller: try harder to publish status when errors occur

ironcladlou:
Always compute status unless it's absolutely impossible — otherwise transient failures that merely partially degrade service (e.g. metrics integration) will cause the operator to report unavailable status.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -336,47 +336,44 @@ func (r *reconciler) ensureClusterIngress(ci *ingressv1alpha1.ClusterIngress, ca
 
 	deployment, err := r.ensureRouterDeployment(ci, infraConfig, ha)
 	if err != nil {
-		errs = append(errs, err)
-	}
-
-	lbService, err := r.ensureLoadBalancerService(ci, deployment, infraConfig, ha)
-	if err != nil {
-		errs = append(errs, fmt.Errorf("failed to ensure load balancer service for %s: %v", ci.Name, err))
-	}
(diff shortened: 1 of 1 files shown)


ironcladlou:
Pretty sure this solves the bug safely — there does remain one gap in that we won't compute status if we fail to create the deployment, but that can be addressed in a followup (if the deployment is not working we'd report an unavailable status anyway)

ironcladlou:
Welp...

```
[AfterEach] [Conformance][Area:Networking][Feature:Router]                                                                                                                                                        /go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/test/e2e/framework/framework.go:148
Feb 22 22:30:26.020: INFO: Waiting up to 3m0s for all (but 100) nodes to be ready
[AfterEach] [Conformance][Area:Networking][Feature:Router]                                                                                                                                                     
  /go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/router/metrics.go:81
Feb 22 22:30:26.085: INFO: Running AfterSuite actions on all node                                                                                                                                               Feb 22 22:30:26.085: INFO: Running AfterSuite actions on node 1
fail [github.com/openshift/origin/test/extended/router/metrics.go:205]: Expected
    <float64>: 9                                                                                                                                                                                                to be >=           
    <float64>: 10
```

Looks like some other metrics flake.

ironcladlou:
Part of resolving https://github.com/openshift/cluster-ingress-operator/pull/136

PR #137, opened 2019-02-22
Treat HA configuration as immutable

Miciah:
Determine the effective HA configuration and persist it to the clusteringress's status at the beginning of reconciling a clusteringress. Use the HA configuration from status for subsequent reconciliation steps and future reconciling.

* `pkg/operator/controller/controller.go` (`Reconcile`): Use `enforceEffectiveIngressDomain` to ensure the clusteringress's status has the effective HA configuration in its status.
(`effectiveHighAvailability`): Rename...
(`enforceEffectiveHighAvailability`): ... to this.  Publish the effective HA configuration to the clusteringress's status.
(`ensureIngressDeleted`): Delete `ha` variable.  Delete argument to `finalizeLoadBalancerService`.
(`ensureRouterNamespace`): Delete `ha` variable and argument to `ensureRouterNamespace`, `ensureLoadBalancerService`, `ensureDNS`, and `syncClusterIngressStatus`.
(`syncClusterIngressStatus`): Delete `ha` parameter.  Delete setting HA configuration in the clusteringress's status.
* `pkg/operator/controller/controller_dns.go` (`ensureDNS`): Delete `ha` parameter.  Delete argument to `desiredDNSRecords`.
(`desiredDNSRecords`): Delete `haType` parameter.  Instead, use the HA configuration from the clusteringress's status.
* `pkg/operator/controller/controller_lb.go` (`ensureLoadBalancerService`): Delete `ha` parameter.  Delete argument to `desiredLoadBalancerService`.
(`desiredLoadBalancerService`): Delete `haType` parameter.  Instead, use the HA configuration from the clusteringress's status.
(`finalizeLoadBalancerService`): Delete `ha` parameter.  Delete argument to `desiredDNSRecords`.
* `pkg/operator/controller/controller_router_deployment.go` (`ensureRouterDeployment`): Delete `ha` parameter.  Delete argument to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Delete `haType` parameter.  Instead. use the HA configuration from the clusteringress's status.
* `pkg/operator/controller/controller_router_deployment_test.go:` Pass HA configuration test input in the clusteringress's status.  Fix calls to `desiredRouterDeployment`.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -191,15 +191,19 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *ingressv1alpha1.ClusterIn
 		return nil
 	}
 
+	updated := ci.DeepCopy()
 	switch {
 	case ci.Spec.IngressDomain != nil:
-		ci.Status.IngressDomain = *ci.Spec.IngressDomain
+		updated.Status.IngressDomain = *ci.Spec.IngressDomain
 	default:
-		ci.Status.IngressDomain = ingressConfig.Spec.Domain
+		updated.Status.IngressDomain = ingressConfig.Spec.Domain
 	}
 	// TODO Validate and check for conflicting claims.
-	if err := r.Client.Status().Update(context.TODO(), ci); err != nil {
-		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", ci.Namespace, ci.Name, err)
+	if err := r.Client.Status().Update(context.TODO(), updated); err != nil {
+		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
+	}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: updated.Namespace, Name: updated.Name}, ci); err != nil {
+		return fmt.Errorf("failed to get clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
 	}
 	return nil
 }
@@ -224,14 +228,18 @@ func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.Cluste
 		return nil
 	}
 
+	updated := ci.DeepCopy()
 	switch {
 	case ci.Spec.HighAvailability != nil:
-		ci.Status.HighAvailability.Type = ci.Spec.HighAvailability.Type
+		updated.Status.HighAvailability.Type = ci.Spec.HighAvailability.Type
 	default:
-		ci.Status.HighAvailability.Type = haTypeForInfra(infraConfig)
+		updated.Status.HighAvailability.Type = haTypeForInfra(infraConfig)
 	}
-	if err := r.Client.Status().Update(context.TODO(), ci); err != nil {
-		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", ci.Namespace, ci.Name, err)
+	if err := r.Client.Status().Update(context.TODO(), updated); err != nil {
+		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
+	}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: updated.Namespace, Name: updated.Name}, ci); err != nil {
+		return fmt.Errorf("failed to get clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/controller.go
```diff
@@ -212,24 +214,32 @@ func haTypeForInfra(infraConfig *configv1.Infrastructure) ingressv1alpha1.Cluste
...
-// effectiveHighAvailability returns the HA configuration as derived from the
-// given clusteringress and infrastructure config.
-func effectiveHighAvailability(ci *ingressv1alpha1.ClusterIngress, infraConfig *configv1.Infrastructure) (ha ingressv1alpha1.ClusterIngressHighAvailability) {
+// enforceEffectiveHighAvailability determines the effective HA configuration
+// for the given clusteringress and infrastructure config and publishes it to
+// the clusteringress's status.
+func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.ClusterIngress, infraConfig *configv1.Infrastructure) error {
+	// The clusteringress's HA configuration is immutable, so if we have
+	// published an HA type status, we must continue using it.
+	if len(ci.Status.HighAvailability.Type) > 0 {
```
Miciah: Do we need something like `UnknownClusterIngressHAType = ""` that we can use here? I'm not sure whether relying on the underlying type's zero value smells.

Miciah:
I considered combining `enforceEffectiveHighAvailability` and `enforceEffectiveIngressDomain` to save a status update, but it seems cleaner to keep them separate.

pkg/operator/controller/controller.go
```diff
@@ -212,24 +214,32 @@ func haTypeForInfra(infraConfig *configv1.Infrastructure) ingressv1alpha1.Cluste
...
 	switch {
 	case ci.Spec.HighAvailability != nil:
-		ha.Type = ci.Spec.HighAvailability.Type
+		ci.Status.HighAvailability.Type = ci.Spec.HighAvailability.Type
 	default:
-		ha.Type = haTypeForInfra(infraConfig)
+		ci.Status.HighAvailability.Type = haTypeForInfra(infraConfig)
 	}
-	return
+	if err := r.Client.Status().Update(context.TODO(), ci); err != nil {
```
ironcladlou: r.enforceEffectiveIngressDomain called Update on `ci` and now this is calling it on the same instance... in both cases you're mutating something that is almost certainly in an informer cache, which is a bug. I think you need to mutate copies before updating and then fetch the latest each time before proceeding.

pkg/operator/controller/controller.go
ironcladlou: length check looks fine to me for now

ironcladlou:
The cache mutations are a problem, otherwise I think this looks basically good

pkg/operator/controller/controller.go
Miciah: I agree we need to fetch the latest after updating so that we have the correct resource version in subsequent updates.  However, I'm confused about cache mutation.  We supply the empty `ClusterIngress`: https://github.com/openshift/cluster-ingress-operator/blob/dbb0ca8d8620eb761f6d0c48918fcf8b130dd092/pkg/operator/controller/controller.go#L95

So how can mutating it mutate the cache?

Miciah: Looks like the unstructure client copies the result into object that the caller (we) provided:
https://github.com/kubernetes-sigs/controller-runtime/blob/3285ad1e768092947e30e66155a44d3599502811/pkg/client/unstructured_client.go#L104

And actually, it _looks_ like we get back the updated object when we update the status—or at least I don't know what else we're getting back: https://github.com/kubernetes-sigs/controller-runtime/blob/3285ad1e768092947e30e66155a44d3599502811/vendor/k8s.io/client-go/dynamic/simple.go#L109-L117 https://github.com/kubernetes-sigs/controller-runtime/blob/3285ad1e768092947e30e66155a44d3599502811/pkg/client/unstructured_client.go#L143-L147

I wish this were documented...

Miciah: This statement seems to indicate there is no cache to mutate (emphasis added): "New returns a new Client using the provided config and Options. The returned client reads *and* writes directly from the server (**it doesn't use object caches**)." https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#New

ironcladlou: Yes it happens that we are passing around a non-caching client right now, but that's only because we're actively working around the controller-runtime defaults which would normally give us a client that reads from the underlying informer cache. And when reading from the caches (ultimately backed by `cache.Store`), mutating the cache objects violates the immutability contract and breaks thread-safety of the cache.

This is why I suggest we be extremely defensive and work with copies when doing *any mutations*. We're then guaranteed to be safe, and what do we lose? The extra spent cycles and allocation memory are clearly irrelevant in this context.

Miciah: Pushed a fix.

ironcladlou: Since we allocate the struct, if the client fetches from the cache, is our struct populated with references to fields from the cached object or do we get copied/unmarshalled stuff? Gotta did deeper I guess

Miciah: The autogenerated `DeepCopy` follows references and makes copies (e.g., if the original clusteringress's `.Spec.IngressDomain` is non-nil, `DeepCopy` makes a copy of the string and puts a reference to the new string in the copy clusteringresses's `.Spec.IngressDomain`; similar is done for other pointer values and values that contain references, such as label selectors).

pkg/operator/controller/controller.go
```diff
@@ -189,15 +191,19 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *ingressv1alpha1.ClusterIn
 		return nil
 	}
 
+	updated := ci.DeepCopy()
```
ironcladlou: Just a random readability thing to consider next time... would it be more clear to reassign ci like

```go
ci = ci.DeepCopy()
```

ironcladlou: I think you talked me out of the need for using copies anyway, if you want to just revert these changes to reduce the overall diff set... if you want to keep this that's fine too

Miciah: > would it be more clear to reassign ci like
> 
> ```go
> ci = ci.DeepCopy()
> ```

Wouldn't that change the pointer value so that the `r.Client.Get` would no longer be updating the caller's `ci`?

PR #138, opened 2019-02-23
dns/aws: cache load balancer hosted zones

ironcladlou:
Before this change, the DNS manager made unfiltered AWS DescribeLoadBalancer
calls every time a clusteringress was reconciled. After this change, the
pertinent results of the DescribeLoadBalancer call are cached to reduce the
number of calls to once per existing load balancer service during the lifetime
of the operator process.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -189,20 +189,21 @@ func (m *Manager) getLBHostedZone(name string) (string, error) {
 	}
 
 	var id string
-	loadBalancers, err := m.elb.DescribeLoadBalancers(&elb.DescribeLoadBalancersInput{})
+	fn := func(resp *elb.DescribeLoadBalancersOutput, lastPage bool) (shouldContinue bool) {
+		for _, lb := range resp.LoadBalancerDescriptions {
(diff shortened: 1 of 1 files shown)


pkg/dns/aws/dns.go
```diff
@@ -174,6 +178,38 @@ func (m *Manager) getZoneID(zoneConfig configv1.DNSZone) (string, error) {
...
+func (m *Manager) getLBHostedZone(name string) (string, error) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if id, exists := m.lbZones[name]; exists {
+		return id, nil
+	}
+
+	var id string
+	loadBalancers, err := m.elb.DescribeLoadBalancers(&elb.DescribeLoadBalancersInput{})
```
abhinavdahiya: `DescribeLoadBalancers` call seems to be paginated
https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_DescribeLoadBalancers.html

ironcladlou: Don’t have time to fix it right now and this shouldn’t be any worse than before — will leave it to others to decide what they wanna do before Monday

crawford:
Looks like CI needs some cleanup:

> TooManyBuckets: You have attempted to create more buckets than allowed

Let me poke around in the CI account and see if there's anything that can obviously be cleaned up.

crawford:
This is probably going to fail with a complaint about CAdvisor, but let's try anyway.

/retest

PR #139, opened 2019-02-25
Publish global secret with default certificates and keys

Miciah:
Publish a `router-certs` secret in the `openshift-config-managed` namespace that has all routers' configured default certificates and their private keys.

This commit resolves [NE-166](https://jira.coreos.com/browse/NE-166).

* `pkg/operator/controller/names.go` (`routerCertsGlobalSecretName`): New constant.
(`RouterCertsGlobalSecretName`): New function to generate the namespaced name for the secret with the routers' default certificates and their keys.
* `pkg/operator/controller/certificate/publish_certs.go`: New file.
(`ensureRouterCertsGlobalSecret`): New method that creates, updates, or deletes the secret as needed.
(`desiredRouterCertsGlobalSecret`): New function.
(`currentRouterCertsGlobalSecret`, `createRouterCertsGlobalSecret`, `updateRouterCertsGlobalSecret`, `deleteRouterCertsGlobalSecret`): New methods.
(`routerCertsSecretsEqual`): New function.
* `pkg/operator/controller/certificate/publish_certs_test.go`: New file.
(`newSecret`, `newClusterIngress`): New functions.
(`TestDesiredRouterCertsGlobalSecret`): New test for `desiredRouterCertsGlobalSecret`.
* `pkg/operator/controller/certificate/default_cert.go`
(`ensureDefaultCertificateForIngress`): Return a Boolean value indicating whether or not the method created, updated, or deleted the secret for the default certificate.
* `pkg/operator/controller/certificate/controller.go` (`Reconcile`): Use the new return value from `ensureDefaultCertificateForIngress`.  If it is true, list secrets and call the new `ensureRouterCertsGlobalSecret` method.
* `hack/uninstall.sh`: Delete the `router-certs` secret.

### DIFF
--- a/pkg/operator/controller/certificate/publish_certs.go
+++ b/pkg/operator/controller/certificate/publish_certs.go
@@ -1,0 +1,161 @@
+package certificate
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"reflect"
+
+	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ensureRouterCertsGlobalSecret will create, update, or delete the global
+// certificates secret as appropriate.
+func (r *reconciler) ensureRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []ingressv1alpha1.ClusterIngress) error {
+	desired, err := desiredRouterCertsGlobalSecret(secrets, ingresses)
+	if err != nil {
+		return err
+	}
+	current, err := r.currentRouterCertsGlobalSecret()
+	if err != nil {
+		return err
+	}
+	switch {
+	case desired == nil && current == nil:
+		// Nothing to do.
+	case desired == nil && current != nil:
+		if deleted, err := r.deleteRouterCertsGlobalSecret(current); err != nil {
+			return fmt.Errorf("failed to ensure router certificates secret was unpublished: %v", err)
+		} else if deleted {
+			r.recorder.Eventf(current, "Normal", "UnpublishedRouterCertificates", "Unpublished router certificates")
+		}
+	case desired != nil && current == nil:
+		if created, err := r.createRouterCertsGlobalSecret(desired); err != nil {
+			return fmt.Errorf("failed to ensure router certificates secret was published: %v", err)
+		} else if created {
+			r.recorder.Eventf(desired, "Normal", "PublishedRouterCertificates", "Published router certificates")
+		}
+	case desired != nil && current != nil:
+		if updated, err := r.updateRouterCertsGlobalSecret(current, desired); err != nil {
+			return fmt.Errorf("failed to update published router certificates secret: %v", err)
+		} else if updated {
+			r.recorder.Eventf(desired, "Normal", "UpdatedPublishedRouterCertificates", "Updated the published router certificates")
+		}
+	}
+	return nil
+}
+
+// desiredRouterCertsGlobalSecret returns the desired router-certs global
+// secret.
+func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []ingressv1alpha1.ClusterIngress) (*corev1.Secret, error) {
+	if len(ingresses) == 0 || len(secrets) == 0 {
+		return nil, nil
+	}
+
+	nameToSecret := map[string]*corev1.Secret{}
+	for i, certSecret := range secrets {
+		nameToSecret[certSecret.Name] = &secrets[i]
+	}
+
+	ingressToSecret := map[*ingressv1alpha1.ClusterIngress]*corev1.Secret{}
+	for i, ingress := range ingresses {
+		secretName := controller.RouterDefaultCertificateSecretName(&ingress, "")
+		name := secretName.Name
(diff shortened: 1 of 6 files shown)


Miciah:
Rebased for #132.  With this rebase, things get a little messier:
* We need to list clusteringresses before we can `ensureRouterCAConfigMap`.
* We want to keep `ensureRouterCAConfigMap` with `ensureRouterCACertificateSecret` to simplify error handling.
* We need to get `caSecret` with `ensureRouterCACertificateSecret` before we can `ensureRouterCAConfigMap` and `ensureClusterIngress`.
* We want to keep `enforceEffectiveIngressDomain` near `ensureClusterIngress` and the code that gets cluster config objects (I think).
* We need `enforceEffectiveIngressDomain` before we can `ensureGlobalCertificatesSecret`.
* We also would like to `ensureClusterIngress` before we `ensureGlobalCertificatesSecret` (I think) so that we `ensureDefaultCertificateForIngress` first (although things will work out even if we reverse the order, possibly with a little more churn). 
* We need the list of clusteringresses before we can `ensureGlobalCertificatesSecret`.

Since this is a short-term solution, I didn't want it to complicate the rest of the logic, so I added a second `r.Client.List` for clusteringresses.  The best alternative, I think, would be to get the ingress configuration earlier and call `enforceEffectiveIngressDomain` earlier so that we can call `ensureGlobalCertificatesSecret` right after `ensureRouterCACertificateSecret` (if it doesn't fail).  Or perhaps I shouldn't be using `ingress.Status.IngressDomain` in `desiredGlobalCertificatesSecret`?

pkg/operator/controller/certificate/publish_certs.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	nameToSecret := map[string]*corev1.Secret{}
+	for i, certSecret := range secrets {
+		nameToSecret[certSecret.Name] = &secrets[i]
+	}
+
+	ingressToSecret := map[*ingressv1alpha1.ClusterIngress]*corev1.Secret{}
+	for i, ingress := range ingresses {
+		secretName := fmt.Sprintf("router-certs-%s", ingress.Name)
```
ironcladlou: This is duplicated from `desiredRouterDeployment()` — should it be extracted? reminds me of `enforceEffectiveDomain`, `haTypeForInfra`, etc. in `controller.go`.

pkg/operator/controller/certificate/publish_certs.go
```diff
@@ -0,0 +1,160 @@
...
+	return true, nil
+}
+
+// updateRouterCertsGlobalSecret updates the router-certs global secret.
+// Returns true if the secret was updated, false otherwise.
+func (r *reconciler) updateRouterCertsGlobalSecret(current, desired *corev1.Secret) (bool, error) {
+	if routerCertsSecretsEqual(current, desired) {
+		return false, nil
+	}
+	updated := current.DeepCopy()
```
ironcladlou: If the update is successful, should we update into `current` instead of a copy? 😁

ironcladlou: Actually, if we _don't_ update into `current`, haven't we invalidated future updates of `current` (because `current` becomes a stale resource version)?

Miciah: > Actually, if we _don't_ update into `current`, haven't we invalidated future updates of `current` (because `current` becomes a stale resource version)?

In this case, it shouldn't matter because the extent terminates when `ensureRouterCertsGlobalSecret` returns.  However, I do want to set a good example.  I wonder whether we want to establish a pattern of updating into the current value, or whether we want to establish a pattern of functions' returning a Boolean value indicating whether they changed the object, leaving it up to the caller to get the new version if needed.

pkg/operator/controller/certificate/publish_certs.go
Miciah: Are you playing mind games? Looks like you already extracted it into `RouterDefaultCertificateSecretName`.

Miciah: Changed to use `RouterDefaultCertificateSecretName`.  Thanks!

ironcladlou: Hmm, I was actually referring to the [effective value calculation](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/controller_router_deployment.go#L202)

```go
secretName := fmt.Sprintf("router-certs-%s", ci.Name)
if ci.Spec.DefaultCertificateSecret != nil && len(*ci.Spec.DefaultCertificateSecret) > 0 {
  secretName = *ci.Spec.DefaultCertificateSecret
}
```

pkg/operator/controller/certificate/publish_certs.go
ironcladlou: Maybe just doc it for now, or mutate and doc?

ironcladlou: Can address in a followup

pkg/operator/controller/certificate/publish_certs.go
ironcladlou: Can address in a followup

pkg/operator/controller/certificate/publish_certs.go
Miciah: Current godoc says, "Returns true if the secret was updated, false otherwise." We could add "Caller is responsible for getting the updated secret if needed."

pkg/operator/controller/certificate/publish_certs.go
Miciah: I'll get it in a separate PR so as not to delay this feature.

Miciah: https://github.com/openshift/cluster-ingress-operator/pull/149

PR #140, opened 2019-02-26
Introduce certificate controller

ironcladlou:
Incremental refactor towards a certificate controller responsible for:

1. Managing a CA for minting self-signed certs
2. Managing self-signed certificates for any clusteringresses which require them
3. Publishing relevant certificate data to `openshift-config-managed` (we have two use cases for this)

This enables us to:

1. Better manage complexity of the ingress controller by decoupling orthogonal concerns (i.e. certificate management)
2. Carve a path to replace our certificate management strategy with a cluster-provided alternative

This change implements step 1; steps 2 and 3 can be followups.

This change supports NE-166 and https://github.com/openshift/cluster-ingress-operator/pull/139.


pkg/operator/controller/cacert/controller.go
```diff
@@ -11,52 +11,88 @@ import (
...
 	// caCertSecretName is the name of the secret that holds the CA certificate
 	// that the operator will use to create default certificates for
 	// clusteringresses.
 	caCertSecretName = "router-ca"
 )
 
-// routerCASecretName returns the namespaced name for the router CA secret.
-func routerCASecretName(namespace string) types.NamespacedName {
+// CASecretName returns the namespaced name for the router CA secret.
+func CASecretName(operatorNamespace string) types.NamespacedName {
```
ironcladlou: TODO: This seems like a possible recipe for cyclic dependencies, haven't thought much about where the right place to put naming functions might be.

pkg/operator/controller/controller.go
```diff
@@ -95,10 +101,24 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		ingress = nil
 	}
 
-	caSecret, err := r.ensureRouterCACertificateSecret()
-	if err != nil {
-		errs = append(errs, fmt.Errorf("failed to ensure CA secret: %v", err))
-	} else {
+	caSecretName := cacert.CASecretName(r.Namespace)
```
ironcladlou: As outlined in the PR description, everything related to the certs can eventually be stripped out of this controller; trying to get there one small step at a time

pkg/operator/operator.go
```diff
@@ -180,6 +191,20 @@ func (o *Operator) Start(stop <-chan struct{}) error {
 		log.Info("cache synced")
 	}
 
+	// Sync the CA cert every once in a while.
+	ticker := time.NewTicker(24 * time.Hour)
```
ironcladlou: Once should be enough, unless there are errors, and so many errors that we hit the max retry limit, but if that ends up being a real problem we could revisit how else to edge trigger or otherwise deal with the period and/or retry mechanics.

pkg/operator/operator.go
```diff
@@ -180,6 +191,20 @@ func (o *Operator) Start(stop <-chan struct{}) error {
...
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
+			case <-ticker.C:
+				ci := ingressv1alpha1.ClusterIngress{}
```
ironcladlou: Not sure if this particular type makes sense for the event, but the controller doesn't care, and this is convenient

pkg/operator/controller/cacert/controller.go
```diff
@@ -11,52 +11,88 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+var log = logf.Logger.WithName("cacert-ontroller")
```
Miciah: Typo: "ontroller".

pkg/operator/controller/cacert/controller.go
```diff
@@ -11,52 +11,88 @@ import (
...
 }
 
-// ensureRouterCACertificateSecret ensures a CA certificate secret exists that
-// can be used to sign the default certificates for ClusterIngresses.
-func (r *reconciler) ensureRouterCACertificateSecret() (*corev1.Secret, error) {
-	current, err := r.currentRouterCASecret()
+func New(mgr manager.Manager, client client.Client, operatorNamespace string, events <-chan event.GenericEvent) (controller.Controller, error) {
+	reconciler := &reconciler{
+		client:            client,
+		recorder:          mgr.GetRecorder("cacert-controller"),
```
Miciah: Maybe add a `controllerName` constant?

pkg/operator/controller/controller.go
```diff
@@ -95,10 +101,24 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	caSecret := &corev1.Secret{}
+	if err := r.Client.Get(context.TODO(), caSecretName, caSecret); err != nil {
+		if errors.IsNotFound(err) {
+			// Most likely the CA cert controller hasn't created it yet, try again
+			// after a reasonable time.
+			result.RequeueAfter = 5 * time.Second
```
Miciah: I thought there were problems using `currentRouterCASecret`, so we were no longer using it (see #132).

pkg/operator/controller/controller.go
```diff
@@ -95,10 +101,24 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			// TODO: This check won't be necessary if all cert stuff is extracted.
+			if ingress != nil {
+				r.recorder.Event(ingress, "Warning", "DefaultWildcardCACertMissing", "The default wildcard CA certificate is missing")
```
Miciah: This feels like more of a status issue: failing condition is true if the CA is missing.

pkg/operator/operator.go
Miciah: This could be useful for rotation, but otherwise, what problem is it worth fixing in 24 hours that doesn't need to be fixed in <1 hour?

Miciah:
> * [ ]  Does `cacert/controller.go` deserve its own entire package? Or would lumping everything into `operator/controller` be better?

I like having the separate `cacert` package as you've done.  Makes the code more modular and easier to reason about, gives it distinct logger, and lets us keep names simple (`cacert.New(...)`).  One downside I see is that we have a little more duplication (e.g., cannot use one reconciler's `r.currentRouterCASecret(...)` in another reconciler), but maybe it's cleaner that way (or if the methods should be more reusable, they should take a client as a parameter instead of having the reconciler as a receiver).

pkg/operator/controller/cacert/controller.go
ironcladlou: Fixed via constant

pkg/operator/controller/cacert/controller.go
ironcladlou: Fixed

pkg/operator/controller/controller.go
ironcladlou: Error take precedence over requeueAfter when it comes to the result processing, which makes sense — if there's no error, requeueAfter would be effective. Does that seem okay here?

pkg/operator/controller/controller.go
ironcladlou: I think you're right but I didn't want to expand status with this PR

pkg/operator/operator.go
ironcladlou: What value do you recommend? Error retries, and (later) expanding status seem useful here.

pkg/operator/controller/controller.go
Miciah: Yeah, seems reasonable.

ironcladlou:
Could this controller eventually handle all the responsibilities around wildcard certificates?

1. Managing a CA for minting self-signed certs
2. Managing self-signed certificates for any clusteringresses which require them
3. Publishing any used wildcard certificates to `openshift-config-managed

If so, what would be a better package/name? `controller/certs`?

ironcladlou:
Slight tweaking to names and docs to better represent the proposed direction

pkg/operator/operator.go
ironcladlou: I redid this so that we just add a single event on startup to prime the controller. Additional edge triggering can come later.

pkg/operator/controller/controller.go
```diff
@@ -95,10 +101,24 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			}
+		} else {
+			errs = append(errs, fmt.Errorf("failed to get CA secret %v: %v", caSecretName, err))
```
pravisankar: %s: %v ?

ironcladlou:
Not clear from the CI logs what exactly went wrong in the install process...

/retest

PR #141, opened 2019-02-26
controller/certificate: move cert generation to cert controller

ironcladlou:
Relocate default wildcard certificate generation to the certificate controller to more
clearly decouple certificate management from ingress reconcilication.

Change the certificate controller event handling so that reconciliation is
triggered by clusteringress events. This makes the reconciliation code in the
controller easier to reason about and means we only do cert management if
there's a clusteringress around. It also means we can easily react to
clusteringresses being added/deleted.

Depends on https://github.com/openshift/cluster-ingress-operator/pull/140
Supports https://github.com/openshift/cluster-ingress-operator/pull/139

### DIFF
--- a/pkg/operator/controller/certificate/ca.go
+++ b/pkg/operator/controller/certificate/ca.go
@@ -1,0 +1,155 @@
+package certificate
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// caCertSecretName is the name of the secret that holds the CA certificate
+	// that the operator will use to create default certificates for
+	// clusteringresses.
+	caCertSecretName = "router-ca"
+)
+
+// CASecretName returns the namespaced name for the router CA secret.
+func CASecretName(operatorNamespace string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: operatorNamespace,
+		Name:      caCertSecretName,
+	}
+}
+
+func (r *reconciler) ensureRouterCASecret() (*corev1.Secret, error) {
+	current, err := r.currentRouterCASecret()
+	if err != nil {
+		return nil, err
+	}
+	if current != nil {
+		return current, nil
+	}
+	desired, err := desiredRouterCASecret(r.operatorNamespace)
+	if err != nil {
+		return nil, err
+	}
+	if created, err := r.createRouterCASecret(desired); err != nil {
+		return nil, fmt.Errorf("failed to create CA secret: %v", err)
+	} else if created {
+		r.recorder.Event(desired, "Normal", "CreatedWildcardCACert", "Created a default wildcard CA certificate")
+	}
+	return desired, nil
+}
+
+// currentRouterCASecret returns the current router CA secret.
+func (r *reconciler) currentRouterCASecret() (*corev1.Secret, error) {
+	name := CASecretName(r.operatorNamespace)
+	secret := &corev1.Secret{}
+	if err := r.client.Get(context.TODO(), name, secret); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return secret, nil
+}
+
+// generateRouterCA generates and returns a CA certificate and key.
+func generateRouterCA() ([]byte, []byte, error) {
+	signerName := fmt.Sprintf("%s@%d", "cluster-ingress-operator", time.Now().Unix())
+
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate key: %v", err)
+	}
+
+	root := &x509.Certificate{
+		Subject: pkix.Name{CommonName: signerName},
+
+		SignatureAlgorithm: x509.SHA256WithRSA,
+
+		NotBefore:    time.Now().Add(-1 * time.Second),
+		NotAfter:     time.Now().Add(2 * 365 * 24 * time.Hour),
+		SerialNumber: big.NewInt(1),
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+
+		IsCA: true,
+
+		// Don't allow the CA to be used to make another CA.
+		MaxPathLen:     0,
+		MaxPathLenZero: true,
+	}
+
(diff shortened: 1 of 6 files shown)


ironcladlou:
This one's ready for review now

pkg/operator/controller/certificate/ca.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (r *reconciler) ensureRouterCASecret() (*corev1.Secret, error) {
+	current, err := r.currentRouterCASecret()
+	if err != nil {
+		return nil, err
+	}
+	if current != nil {
+		return current, nil
+	}
+	desired, err := desiredRouterCASecret(r.operatorNamespace)
```
pravisankar: `desiredRouterCASecret` will generate new cert every time and current != desired after first iteration.
May be we could just use get + create resource pattern that we use in other places to avoid confusion? or we could rename `desired` to something more accurate.

pkg/operator/controller/certificate/controller.go
```diff
@@ -77,119 +59,56 @@ type reconciler struct {
...
-			Namespace: name.Namespace,
-		},
-		Data: map[string][]byte{
-			"tls.crt": certBytes,
-			"tls.key": keyBytes,
-		},
-		Type: corev1.SecretTypeTLS,
 	}
-	return secret, nil
+	return reconcile.Result{}, utilerrors.NewAggregate(errs)
```
pravisankar: return result, ... (otherwise result.RequeueAfter is not propagated to the caller)

pkg/operator/controller/certificate/controller.go
```diff
@@ -77,119 +59,56 @@ type reconciler struct {
...
 }
 
-// createRouterCASecret creates the router CA secret.
-func (r *reconciler) createRouterCASecret(secret *corev1.Secret) error {
-	if err := r.client.Create(context.TODO(), secret); err != nil {
-		return err
+func routerDeploymentName(ci *ingressv1alpha1.ClusterIngress) types.NamespacedName {
```
pravisankar: I think you suggested moving [Resource]Name to controller/names.go or something on slack. Do we want to do that as a separate pr or is it dropped for some other reason?

pkg/operator/controller/certificate/ca.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := r.createRouterCASecret(desired); err != nil {
+		return nil, fmt.Errorf("failed to create CA secret: %v", err)
+	}
+	r.recorder.Event(desired, "Normal", "CreatedWildcardCACert", "Created a default wildcard CA certificate")
```
Miciah: Seems odd to log in `createRouterCASecret` and emit the event in `ensureRouterCASecret`; do we need to do both, and if so, should we do them in the same place?

pkg/operator/controller/certificate/controller.go
```diff
@@ -77,119 +59,56 @@ type reconciler struct {
...
-		if errors.IsNotFound(err) {
-			return nil, nil
+		return reconcile.Result{}, fmt.Errorf("failed to ensure router CA secret: %v", err)
+	}
+
+	result := reconcile.Result{}
+	errs := []error{}
+	ingress := &ingressv1alpha1.ClusterIngress{}
+	if err := r.client.Get(context.TODO(), request.NamespacedName, ingress); err != nil {
+		if !errors.IsNotFound(err) {
```
Miciah: Looks like `!errors.IsNotFound(err)` should be `errors.IsNotFound(err)`.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -43,12 +43,16 @@ func (r *reconciler) ensureDefaultCertificateForIngress(caSecret *corev1.Secret,
...
 			return fmt.Errorf("failed to delete default certificate: %v", err)
+		} else if deleted {
+			r.recorder.Eventf(ci, "Normal", "DeletedDefaultCertificate", "Deleted default wildcard certificate %q", desired.Name)
 		}
 	case desired != nil && current == nil:
-		if err := r.createRouterDefaultCertificate(desired); err != nil {
+		if created, err := r.createRouterDefaultCertificate(desired); err != nil {
 			return fmt.Errorf("failed to create default certificate: %v", err)
+		} else if created {
+			r.recorder.Eventf(ci, "Normal", "CreatedDefaultCertificate", "Created default wildcard certificate %q", desired.Name)
```
pravisankar: What's the guideline for using events over logging?

pkg/operator/controller/certificate/ca.go
ironcladlou: I missed this refactor with the other event handling; fixed

pkg/operator/controller/certificate/controller.go
ironcladlou: Good catch, fixed

pkg/operator/controller/certificate/controller.go
ironcladlou: Started the cleanup of name handling in https://github.com/openshift/cluster-ingress-operator/pull/142; too much work to do here because of import cycles

pkg/operator/controller/certificate/default_cert.go
ironcladlou: Up to us to decide what's most user friendly AFAIK

pkg/operator/controller/certificate/controller.go
ironcladlou: Good catch, fixed

pkg/operator/controller/certificate/ca.go
ironcladlou: We're not currently diffing them — are you saying there's an error here, or just that it's confusing?

If desiredRouterCASecret isn't deterministic that seems like a problem, and if it _can't be_ deterministic then our diffing logic would have to account for it (if/when we start diffing)

Miciah: `routerCAConfigMapsEqual` compares only `.data`, similar to how `deploymentConfigChanged` compares only select fields.  I think this pattern is reasonable since there will always be fields that should be ignored (e.g., uid).

pravisankar: There is no error here. It was confusing for me when i saw current can be different from desired (issue, expiry times will differ as those are derived from current time). Ignore if you think this does not confuse others.

ironcladlou: Agree it's weird, and we can fix it in a followup as it's no different than today — I just transferred the logic over as close to the same as possible for the restructuring

pravisankar:
@ironcladlou squash your commits?

pkg/operator/controller/certificate/controller.go
```diff
@@ -77,119 +59,56 @@ type reconciler struct {
...
-	if err := r.client.Get(context.TODO(), name, secret); err != nil {
+	result := reconcile.Result{}
+	errs := []error{}
+	ingress := &ingressv1alpha1.ClusterIngress{}
+	if err := r.client.Get(context.TODO(), request.NamespacedName, ingress); err != nil {
 		if errors.IsNotFound(err) {
-			return nil, nil
+			// This means the ingress was already deleted/finalized and there are
+			// stale queue entries (or something edge triggering from a related
+			// resource that got deleted async).
```
Miciah: > (or something edge triggering from a related resource that got deleted async).

This comment does not make sense in this new context at present time.

PR #143, opened 2019-02-27
Remove unnecessary IsAlreadyExists() checks during resource creation.

pravisankar:
- If we are just creating resource without get op, then IsAlreadyExists() check is needed
but when we are using get resource and in case of not found error, create
the resource pattern this check is unnecessary as we know that this error
will not occur.

- Remove stale 'MetricsServiceMonitorAsset' constant, no longer needed


pravisankar:

PR #144, opened 2019-02-27
WIP: Adds service loadbalancer status to clusteringress

danehans:
Previously, you could not tell if the loadbalancer ip/hostname was available for a clusteringress. This PR updates clusteringress api and status sync'ing to support loadbalancer service status.

### DIFF
--- a/pkg/apis/ingress/v1alpha1/clusteringress_types.go
+++ b/pkg/apis/ingress/v1alpha1/clusteringress_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -81,6 +82,9 @@ type ClusterIngressHighAvailability struct {
 type ClusterIngressStatus struct {
 	// Replicas is the actual number of observed router instances.
 	Replicas int32 `json:"replicas"`
+
+	// LoadBalancer is observed service loadbalancer status.
+	LoadBalancer corev1.LoadBalancerStatus `json:"loadBalancer,omitempty"`
 
 	// Selector is a label selector, in string format, for pods
 	// corresponding to the ClusterIngress. The number of matching pods
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -353,9 +353,11 @@ func (r *reconciler) ensureClusterIngress(ci *ingressv1alpha1.ClusterIngress, ca
 			Controller: &trueVar,
 		}
 
-		if lbService, err := r.ensureLoadBalancerService(ci, deploymentRef, infraConfig); err != nil {
+		lbService, err := r.ensureLoadBalancerService(ci, deploymentRef, infraConfig)
+		if err != nil {
 			errs = append(errs, fmt.Errorf("failed to ensure load balancer service for %s: %v", ci.Name, err))
-		} else if lbService != nil {
+		}
+		if len(lbService.Status.LoadBalancer.Ingress) != 0 {
 			if err := r.ensureDNS(ci, lbService, dnsConfig); err != nil {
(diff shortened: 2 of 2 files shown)


ironcladlou:
API is effectively frozen right now we have to be careful...

That said, would be curious to compare this approach to new conditions. Knowing the status of the LB itself doesn't seem to imply much about the functional state of the clusteringress. For example, if we know an LB service is being managed but the LB service is in a pending state, does that imply the clusteringress should report unavailable in the existing condition status item? And would an additional condition alongside it (like LoadBalancerAvailable) provide additional high level context as to the reason for the current availability condition?

danehans:
> API is effectively frozen right now we have to be careful...

That makes sense. Should I close this PR or keep it open for continued discussion and potentially include in the future?

> Knowing the status of the LB itself doesn't seem to imply much about the functional state of the clusteringress.
In my troubleshooting, I found myself looking to see if the svc lb received an `EXTERNAL-IP` quite often. 

Currently, the clusteringress status reports the number of replicas for the router deployment. I find this status to be useful to understand the state of the deployment resource provisioned by clusteringress. I think it would be beneficial to have status visibility into other dependent resources provisioned by clusteringress.

> For example, if we know an LB service is being managed but the LB service is in a pending state, does that imply the clusteringress should report unavailable in the existing condition status item?

IMO, I think so. I don't see how a clusteringress is functional without the LB service `EXTERNAL-IP` and the alias record. Until all dependent resources are created, I think  a `type: Progressing` status is reasonable.

ironcladlou:
>> Knowing the status of the LB itself doesn't seem to imply much about the functional state of the clusteringress.

>Currently, the clusteringress status reports the number of replicas for the router deployment. I find this status to be useful to understand the state of the deployment resource provisioned by clusteringress. I think it would be beneficial to have status visibility into other dependent resources provisioned by clusteringress.

Agree... one difference with deployment is clusteringresses /always/ have deployments, and not all clusteringresses will have LB services associated with them.

> In my troubleshooting, I found myself looking to see if the svc lb received an `EXTERNAL-IP` quite often.

This seems like a good one to talk more about... why were you looking to see the state of the service lb? What was the preceding question you were trying to answer (and can you describe the line of questioning recursively?), and also what led you to even investigate the service lb in particular?

ironcladlou:
>That makes sense. Should I close this PR or keep it open for continued discussion and potentially include in the future?

I'm happy to keep it open even for discussion right now, I'm finding it useful

ironcladlou:
>Agree... one difference with deployment is clusteringresses _always_ have deployments, and not all clusteringresses will have LB services associated with them.

I should have referenced [sig-architecture/api-conventions.md](
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) here to better elaborate on why my statement matters. What I was getting at is that the LB status specifically seems more like a condition than a status field, and my followup line of questioning is intended to explore what questions you (the confounded administrator) were asking which perhaps the right conditions could help answer. For some answers a status field may be more appropriate.

danehans:
> Agree... one difference with deployment is clusteringresses _always_ have deployments, and not all clusteringresses will have LB services associated with them.

Thank you for this information. Is it correct to say every clusteringress will have a service that externally exposes the router?

> This seems like a good one to talk more about... why were you looking to see the state of the service lb? What was the preceding question you were trying to answer (and can you describe the line of questioning recursively?), and also what led you to even investigate the service lb in particular?

I installed my cluster the other day using master. Even though the installer completed successfully, the cluster was not provisioned properly. For example, the route crd and default routes were not created, pods from the router deployment did not reach a running state, and an ELB IP was not getting assigned to a service. I was led to look at the router LB service due to operator log messages stating "failed to update alias in zone" even though the service was created "created load balancer service".

ironcladlou:
We settled on some [new condition API](https://github.com/openshift/api/pull/241) to cover this; let's do a new PR to implement the API.

PR #146, opened 2019-02-28
Updates finalizer to support user-defined clusteringresses

danehans:
This PR refactors the operator finalizer by:

- Adds finalizer support for `clusteringress` and dependent resources, fixing bug: https://bugzilla.redhat.com/show_bug.cgi?id=1682922


pkg/operator/controller/controller.go
```diff
@@ -169,8 +169,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				}
 			} else {
 				// Handle everything else.
+				if err := r.ensureClusterIngressFinalized(ingress); err != nil {
```
ironcladlou: Alternative for consideration: rename to `enforceFinalizer` and add to the `enforceEffectiveIngressDomain` and `enforceEffectiveHighAvailability` "invariant processing chain" above

pkg/operator/controller/controller.go
```diff
@@ -247,6 +249,19 @@ func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.Cluste
 	return nil
 }
 
+// ensureClusterIngressFinalized adds ClusterIngressFinalizer to ingress
+// if it does not exist and does not contain a DeletionTimestamp.
+func (r *reconciler) ensureClusterIngressFinalized(ingress *ingressv1alpha1.ClusterIngress) error {
+	if ingress.ObjectMeta.DeletionTimestamp.IsZero() && !slice.ContainsString(ingress.Finalizers, ClusterIngressFinalizer) || ingress.Finalizers == nil {
```
ironcladlou: If following the `enforce...` pattern above I don't think the deletion check is necessary

pkg/operator/controller/controller.go
```diff
@@ -247,6 +249,19 @@ func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.Cluste
...
+		ingress.Finalizers = append(ingress.Finalizers, ClusterIngressFinalizer)
+		if err := r.Client.Update(context.TODO(), ingress); err != nil {
```
ironcladlou: @Miciah @pravisankar did we decide our "policy" on mutations? As @Miciah previously explained we're pretty safe given the particulars of our client implementation. The policy could just be to note in GoDoc that the parameter will be modified. WDYT?

pkg/operator/controller/controller.go
ironcladlou: Oh wait, we need the other enforces _even when we handle deletes_, nevermind

pkg/operator/controller/controller.go
ironcladlou: Disregard (
https://github.com/openshift/cluster-ingress-operator/pull/146#discussion_r261014348). Only suggestion here now is a potential rename to "enforceFinalizer" for consistency

pkg/operator/controller/controller.go
```diff
@@ -169,8 +169,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				}
 			} else {
 				// Handle everything else.
+				if err := r.ensureClusterIngressFinalized(ingress); err != nil {
+					errs = append(errs, fmt.Errorf("failed to finalize clusteringress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+				}
 				if err := r.ensureClusterIngress(ingress, dnsConfig, infraConfig); err != nil {
```
ironcladlou: Because this isn't in an `else` of the finalizer enforcement, does that leave a gap where we can fail to set the finalizer, reconcile the ingress anyway, get deleted, and then fall through to deletion handling without a finalizer? Can we actually reconcile before we've successfully enforced the finalizer invariant?

pkg/operator/controller/controller.go
Miciah: Using controller-runtime's API, it is clear and natural to use the client as Dane is using it here—that is, we declare the `ingressv1alpha1.ClusterIngress{}` in `Reconcile` and pass around a pointer to it, and we count on `r.Client.Update` to update it for us.  How likely are we to move away from the controller-runtime client?

pkg/operator/controller/controller.go
```diff
@@ -247,6 +249,19 @@ func (r *reconciler) enforceEffectiveHighAvailability(ci *ingressv1alpha1.Cluste
 	return nil
 }
 
+// ensureClusterIngressFinalized adds ClusterIngressFinalizer to ingress
+// if it does not exist and does not contain a DeletionTimestamp.
+func (r *reconciler) ensureClusterIngressFinalized(ingress *ingressv1alpha1.ClusterIngress) error {
```
Miciah: `Finalized` makes it sound like the method finalizes the clusteringress, similar to how `finalizeLoadBalancerService` finalizes the service.  How about `HasFinalizer`?

pkg/operator/controller/controller.go
Miciah: Agree, this should be made an `else if` clause to the preceding `if`.

Miciah:
I'm not sure we want to consolidate finalizer names, and I think it would be safer and clearer if we added each finalizer wherever we created the thing that needs to be finalized.  That is, I would add `ClusterIngressFinalizer` before the call to `createRouterDeployment` in `ensureRouterDeployment`.  Following this logic, we should also be adding the finalizer for DNS right before we update Route 53 in `ensureDNS`.  @ironcladlou, does that seem reasonable?

pkg/operator/controller/controller.go
Miciah: I don't understand how https://github.com/openshift/cluster-ingress-operator/pull/146#discussion_r261014348 means the deletion check is needed.  Anyway, the logic looks wrong and might benefit from reordering, parentheses, or simplification: Do we really want to add the finalizer if `ingress.Finalizers == nil`, even if the ingress is deleted? Do we need the `ingress.Finalizers == nil` check at all (since `slice.ContainsString` treats a nil slice as an empty slice)?

pkg/operator/controller/controller.go
danehans: Thanks for helping me better understand the logic. I'll update.

pkg/operator/controller/controller.go
danehans: I'll update the method based on your feedback.

pkg/operator/controller/controller.go
danehans: I have updated to `enforceFinalizer`.

danehans:
Thanks all for the review. I have implemented all of the requested changes except https://github.com/openshift/cluster-ingress-operator/pull/146#issuecomment-468111794. I'll wait for others to provide input on the topic before resubmitting the PR. I think the points @Miciah makes are valid and I'm open to proceeding in the direction that he suggests.

pkg/operator/controller/controller.go
ironcladlou: Okay, I agree with you, I know you've been trying to hammer this into my head for a while... just making sure. I think we're all on the same page and that we can rely on Update and Create to "refresh" existing references we're passing around.

ironcladlou: It also means we're free to remove the `DeepCopy()` calls floating around, but that's just random cleanup

pkg/operator/controller/controller_lb.go
```diff
@@ -108,6 +106,20 @@ func (r *reconciler) currentLoadBalancerService(ci *ingressv1alpha1.ClusterIngre
 	return service, nil
 }
 
+// enforceLoadBalancerServiceFinalizer adds loadBalancerServiceFinalizer to service
+// if it doesn't exist.
+func (r *reconciler) enforceLoadBalancerServiceFinalizer(service *corev1.Service) error {
+	// TODO: Figure out why DNS updates apply DeletionTimestamp to loadbalancer service.
```
danehans: I added this `TODO` because testing uncovered the following error. The error is attributed to the loadbalancer service being recreated after the service is assigned an `EXTERNAL-IP` and an DNS gets updated. I can see `DeletionTimestamp` is applied to the initial load balancer service.

```
2019-03-01T14:02:23.263-0800	ERROR	operator.init.kubebuilder.controller	controller/controller.go:217	Reconciler error	{"controller": "operator-controller", "request": "openshift-ingress-operator/default", "error": "failed to ensure clusteringress openshift-ingress-operator/default: failed to enforce load balancer service finalizer for ingress openshift-ingress-operator/default: Service \"router-default\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"ingress.openshift.io/load-balancer-service\"}", "errorCauses": [{"error": "failed to ensure clusteringress openshift-ingress-operator/default: failed to enforce load balancer service finalizer for ingress openshift-ingress-operator/default: Service \"router-default\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"ingress.openshift.io/load-balancer-service\"}"}]}
```

pkg/operator/controller/controller.go
```diff
@@ -131,14 +131,20 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			} else if err := r.enforceEffectiveHighAvailability(ingress, infraConfig); err != nil {
 				errs = append(errs, fmt.Errorf("failed to enforce the effective HA configuration for clusteringress %s: %v", ingress.Name, err))
 			} else if ingress.DeletionTimestamp != nil {
-				// Handle deletion.
+				// Handle ingress deletion and finalization.
 				if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
 					errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
 				}
+				if err := r.finalizeClusterIngress(ingress); err != nil {
```
ironcladlou: If `ensureIngressDeleted` returns an error, `finalizeClusterIngress` is still called. Shouldn't the ingress finalizer only be removed once we're certain we successfully cleaned up?

danehans: ack

pkg/operator/controller/controller_lb.go
```diff
@@ -107,6 +105,19 @@ func (r *reconciler) currentLoadBalancerService(ci *ingressv1alpha1.ClusterIngre
 	return service, nil
 }
 
+// enforceLoadBalancerServiceFinalizer adds loadBalancerServiceFinalizer to service
+// if it doesn't exist.
+func (r *reconciler) enforceLoadBalancerServiceFinalizer(service *corev1.Service) error {
```
ironcladlou: I'm not clear why this was extracted out of the 1-liner assignment as part of the create. The reason we need enforcement for the clusteringress is because we can't mutate the incoming resource during admission, so we must update it after the fact. The operator owns the creation of the service, and so can set the finalizer atomically on the service during create. Therefore there's no need to ever check/update again.

ironcladlou:
Would prefer to extract out a focused commit/PR that addresses only 
https://bugzilla.redhat.com/show_bug.cgi?id=1682922

pkg/operator/controller/controller_lb.go
danehans: OK, I can revert. I thought we wanted to "enforce" the finalizer. Previously, the finalizer could be removed by a user.

ironcladlou: I think we can enforce that in validation once the API refactor is done

danehans:
@ironcladlou looks like this PR is failing b/c the openshift e2e tests are running against it.

Miciah:
e2e-aws is always run.  Since e2e-aws-operator passed and the ingress operator logs from the e2e-aws test run do not show anything out of the ordinary, e2e-aws probably hit some test flakes.

/test e2e-aws

PR #147, opened 2019-03-01
Improve and expand checks in deploymentConfigChanged

Miciah:
## Improve volumes handling

Fix assumptions in `deploymentConfigChanged` about the deployment's pod template spec, and make the comparison and update more general.  Specifically, do not assume that the deployment has any volumes, do use `cmp.Equal` to compare the entire slice of volumes rather than only checking the first, and use a `for` loop plus `DeepCopy` to copy the updated volumes.

Before this change, `deploymentConfigChanged` did not react to changes to other volumes and could even panic with an out-of-bounds index if there were no volumes.  Any deployment that the operator creates will always have at least one volume, but we might as well be resilient in case someone else directly modifies the deployment.

* `pkg/operator/controller/controller_router_deployment.go` (`deploymentConfigChanged`): Delete a stale comment.  Use `cmp.Equal` to compare volumes.  Use a `for` loop plus `DeepCopy` to update volumes.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add tests for `deploymentConfigChanged`.

## Check node selector

Make `deploymentConfigChanged` check whether the deployment's pod template spec's node selector changed and set it in the updated deployment.

This commit fixes [bug 1683761](https://bugzilla.redhat.com/show_bug.cgi?id=1683761).

* `pkg/operator/controller/controller_router_deployment.go` (`deploymentConfigChanged`): Compare node selector.  Set the node selector in the updated deployment.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add test case for node selector.

## Check environment variables

Make `deploymentConfigChanged` check whether the deployment's pod template spec's container's environment variables changed and set them in the updated deployment.

This commit fixes [bug 1683762](https://bugzilla.redhat.com/show_bug.cgi?id=1683762) and [bug 1683763](https://bugzilla.redhat.com/show_bug.cgi?id=1683763).

* `pkg/operator/controller/controller_router_deployment.go` (`updateRouterDeployment`): Compare environment variables.  Set the environment variables in the updated deployment.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add test cases for changing `ROUTER_CANONICAL_HOSTNAME` or `ROUTER_USE_PROXY_PROTOCOL`, adding `NAMESPACE_LABELS`, or deleting `ROUTE_LABELS`.

---

This is more targeted than the approach we took in https://github.com/openshift/cluster-ingress-operator/pull/94 and had to revert.  Still, the approach here is broader than it strictly needs to be to resolve the referenced issues.  I'll pay close attention to the cluster-ingress-operator logs from e2e-aws-operator, and I am open to narrowing the added comparisons if need be.

### DIFF
--- a/pkg/operator/controller/controller_router_deployment.go
+++ b/pkg/operator/controller/controller_router_deployment.go
@@ -249,6 +249,7 @@ func (r *reconciler) updateRouterDeployment(current, desired *appsv1.Deployment)
 func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
 	if cmp.Equal(current.Spec.Template.Spec.Volumes, expected.Spec.Template.Spec.Volumes, cmpopts.EquateEmpty()) &&
 		cmp.Equal(current.Spec.Template.Spec.NodeSelector, expected.Spec.Template.Spec.NodeSelector, cmpopts.EquateEmpty()) &&
+		cmp.Equal(current.Spec.Template.Spec.Containers[0].Env, expected.Spec.Template.Spec.Containers[0].Env, cmpopts.EquateEmpty(), cmpopts.SortSlices(func(a, b corev1.EnvVar) bool { return a.Name < b.Name })) &&
 		current.Spec.Replicas != nil &&
 		*current.Spec.Replicas == *expected.Spec.Replicas {
 		return false, nil
@@ -261,6 +262,7 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 	}
 	updated.Spec.Template.Spec.Volumes = volumes
 	updated.Spec.Template.Spec.NodeSelector = expected.Spec.Template.Spec.NodeSelector
+	updated.Spec.Template.Spec.Containers[0].Env = expected.Spec.Template.Spec.Containers[0].Env
 	replicas := int32(1)
 	if expected.Spec.Replicas != nil {
 		replicas = *expected.Spec.Replicas
--- a/pkg/operator/controller/controller_router_deployment.go
+++ b/pkg/operator/controller/controller_router_deployment.go
@@ -248,6 +248,7 @@ func (r *reconciler) updateRouterDeployment(current, desired *appsv1.Deployment)
 // for the cluster ingress deployment and if not returns the updated config.
 func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
 	if cmp.Equal(current.Spec.Template.Spec.Volumes, expected.Spec.Template.Spec.Volumes, cmpopts.EquateEmpty()) &&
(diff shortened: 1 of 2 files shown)


Miciah:
Tests are passing, but I found a nil panic in the logs; fix: https://github.com/openshift/cluster-ingress-operator/pull/152.  Once that fix is merged, I would like to retest this PR and see whether we can get a clean run with no crashes.

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -244,16 +247,22 @@ func (r *reconciler) updateRouterDeployment(current, desired *appsv1.Deployment)
 // deploymentConfigChanged checks if current config matches the expected config
 // for the cluster ingress deployment and if not returns the updated config.
 func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
-	// As per an offline conversation, this checks only the secret name
-	// for now but can be updated to a `reflect.DeepEqual` if needed.
-	if current.Spec.Template.Spec.Volumes[0].Secret.SecretName == expected.Spec.Template.Spec.Volumes[0].Secret.SecretName &&
+	if cmp.Equal(current.Spec.Template.Spec.Volumes, expected.Spec.Template.Spec.Volumes, cmpopts.EquateEmpty()) &&
```
ironcladlou: What would a narrowed version look like?

Comparing the desired route labels, e.g.

`metav1.LabelSelectorAsSelector(ci.Spec.RouteSelector)`

to the current route labels, e.g.

`labels.ParseSelector(env["ROUTE_LABELS"])`

?

Miciah: It could look something like that, for each volume and environment variable we wanted to check.

But should we ignore volumes and variables that are not on some check list? One of the requirements for an operator is to repair its operands.  https://github.com/openshift/cluster-ingress-operator/pull/94 attempted to do so by comparing the deployment's `.metadata.annotations`, `.metadata.labels`, and `.spec` and was reverted because it caused spurious reconciliations.  I believe checking volumes and environments as wholes should be sufficiently narrow to avoid the spurious reconciliations (https://github.com/openshift/cluster-ingress-operator/pull/124 and other recent clean-ups probably help too).  (I further believe it would be worthwhile to determine what in `.metadata.annotations`, `.metadata.labels`, and `.spec` was causing the spurious updates because the more we can check without causing spurious reconciles, the more we can repair or handle updates for.)

The alternative to checking volumes and environments as wholes is to hard-code volumes and a list of environment variables which matter enough that we need to ensure they have the right values, as opposed to volumes and environment variables we want to leave unmanaged.  I cannot imagine what the latter would be, and I cannot see the risk in simply checking all volumes and variables.

ironcladlou: > I cannot imagine what the latter would be, and I cannot see the risk in simply checking all volumes and variables.

Me either, but that's the risk we'll take: discovering in the wild what we're not foreseeing in this regard.

I'm okay to keep what you have here, let's just continue to be diligent with our diff/apply changes.

PR #148, opened 2019-03-01
Simplify ClusterOperatorStatus comparison

Miciah:
Replace `util.ElementsEqual` by using go-cmp's `cmpopts.EquateEmpty` and `cmpopts.SortSlices` options.

* `pkg/operator/controller/status.go` (`syncOperatorStatus`): Move comparison of old and new `ClusterOperatorStatus` values from here...
(`statusesEqual`): ...to this new function.  Replace `util.ElementsEqual` by using `cmp.Equal` with suitable `cmpopts.IgnoreFields`, `cmpopts.EquateEmpty`, and `cmpopts.SortSlices` options.
* `pkg/util/util.go` (`ElementsEqual`): Delete file and function.
* `pkg/util/util_test.go` (`TestElementsEqual`): Move test...
* `pkg/operator/controller/status_test.go` (`TestStatusesEqual`): ...to here.  Test `statusesEqual` instead of `util.ElementsEqual`.

---

@pravisankar, you wrote a lot of the code that is modified here; do you see any problems with this approach? Is this a worthwhile simplification?


pkg/operator/controller/status.go
```diff
@@ -41,10 +40,8 @@ func (r *reconciler) syncOperatorStatus() error {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
 
-	oldConditions := co.Status.Conditions
-	co.Status.Conditions = computeStatusConditions(oldConditions, ns, ingresses, deployments)
-
-	oldRelatedObjects := co.Status.RelatedObjects
+	oldStatus := *co.Status.DeepCopy()
```
pravisankar: do we need explicit dereference here? (*oc).Status.DeepCopy()

pkg/operator/controller/status.go
```diff
@@ -232,3 +209,35 @@ func setStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition,
 
 	return newConditions
 }
+
+// statusesEqual compares to ClusterOperatorStatus values.  Returns true the
```
pravisankar: compares *two* ClusterOperatorStatus... Returns true *if* the...

pkg/operator/controller/status.go
Miciah: `DeepCopy` returns a pointer, so we need to dereference its return value either here or when we pass it to `statusesEqual` further down.

pkg/operator/controller/status.go
Miciah: Whoops! Fixed.  Thanks!

ironcladlou:
Wow, this is cool. Good find

PR #149, opened 2019-03-01
Add function for name of in-use default certificate

Miciah:
* `pkg/operator/controller/names.go`
(`RouterDefaultCertificateSecretName`): Rename...
(`RouterOperatorGeneratedDefaultCertificateSecretName`): ...to this.
(`RouterEffectiveDefaultCertificateSecretName`): New function to get the name of the in-use router default certificate secret.
* `pkg/operator/controller/certificate/default_cert.go` (`desiredRouterDefaultCertificateSecret`, `currentRouterDefaultCertificate`): Use `RouterOperatorGeneratedDefaultCertificateSecretName`.
* `pkg/operator/controller/certificate/publish_certs.go` (`desiredRouterCertsGlobalSecret`):
* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): Use `RouterEffectiveDefaultCertificateSecretName`.

---

This PR addresses https://github.com/openshift/cluster-ingress-operator/pull/139#discussion_r261397620 (@ironcladlou).

### DIFF
--- a/pkg/operator/controller/certificate/publish_certs.go
+++ b/pkg/operator/controller/certificate/publish_certs.go
@@ -64,12 +64,8 @@ func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []ingress
 
 	ingressToSecret := map[*ingressv1alpha1.ClusterIngress]*corev1.Secret{}
 	for i, ingress := range ingresses {
-		secretName := controller.RouterDefaultCertificateSecretName(&ingress, "")
-		name := secretName.Name
-		if ingress.Spec.DefaultCertificateSecret != nil && len(*ingress.Spec.DefaultCertificateSecret) > 0 {
-			name = *ingress.Spec.DefaultCertificateSecret
-		}
-		if secret, ok := nameToSecret[name]; ok {
+		name := controller.RouterEffectiveDefaultCertificateSecretName(&ingress, "")
+		if secret, ok := nameToSecret[name.Name]; ok {
 			ingressToSecret[&ingresses[i]] = secret
 		}
 	}
--- a/pkg/operator/controller/certificate/default_cert.go
+++ b/pkg/operator/controller/certificate/default_cert.go
@@ -80,7 +80,7 @@ func desiredRouterDefaultCertificateSecret(ca *crypto.CA, namespace string, depl
 		return nil, fmt.Errorf("failed to encode certificate: %v", err)
 	}
 
-	name := controller.RouterDefaultCertificateSecretName(ci, namespace)
+	name := controller.RouterOperatorGeneratedDefaultCertificateSecretName(ci, namespace)
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name.Name,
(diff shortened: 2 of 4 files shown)


pkg/operator/controller/certificate/publish_certs.go
```diff
@@ -64,12 +64,8 @@ func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []ingress
 
 	ingressToSecret := map[*ingressv1alpha1.ClusterIngress]*corev1.Secret{}
 	for i, ingress := range ingresses {
-		secretName := controller.RouterDefaultCertificateSecretName(&ingress, "")
-		name := secretName.Name
-		if ingress.Spec.DefaultCertificateSecret != nil && len(*ingress.Spec.DefaultCertificateSecret) > 0 {
-			name = *ingress.Spec.DefaultCertificateSecret
-		}
-		if secret, ok := nameToSecret[name]; ok {
+		name := controller.RouterEffectiveDefaultCertificateSecretName(&ingress, "")
```
ironcladlou: Passing no namespace here is smelly. This is back to the prior convo about whether the naming functions should be aware of the operator/operand namespace rather than relying on callers.

Miciah: That is an important convo, but in the case of `RouterEffectiveDefaultCertificateSecretName`, I wonder whether it would make sense just to return a string with the name rather than a NamespacedName.

ironcladlou: When is the generated default certificate ever _not_ in the operand's namespace?

Miciah: Never.  The secret has to be in the same namespace as the deployment.

ironcladlou: Then to me it doesn't make sense to enforce that invariant independently from every callsite — should the thing that knows the name of the secret also know the _location_?

Miciah: Probably, but we haven't figured out how we're going to handle clusteringresses outside of `openshift-ingress-operator`.  I the clusteringress is outside the operator's namespace, do we create the router in the same namespace, or will the clusteringress have a field indicating where the router should be created?

ironcladlou: People (Clayton) keep talking about that capability but I have no idea what it really means  and I'm sure supporting it would require terrifically invasive refactoring and careful design in the future. Until we have some specific design discussion around what the requirement is and what we need to do to get there, why account for it randomly/inconsistently now?

Miciah: Shall we hard-code "openshift-ingress" as the namespace? In the future, presumably we'll derive the namespace from the clusteringress, somehow (e.g., map clusteringress namespace "openshift-ingress-operator" to operand namespace "openshift-ingress", else use the clusteringress's namespace).

ironcladlou: I feel like that would be an improvement in the short term (maybe even just use constants in `names.go` for now?)

Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/pull/168/files#diff-229ee1cd359a98fd757fa9cf002f7597R190.

PR #150, opened 2019-03-01
Update and report operator release version compliant with the cluster version operator

pravisankar:


cmd/cluster-ingress-operator/main.go
```diff
@@ -73,19 +73,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	operatorConfig := operatorconfig.Config{
+		OperatorReleaseVersion: os.Getenv("RELEASE_VERSION"),
+		Namespace:              operatorNamespace,
+		RouterImage:            routerImage,
+		Platform:               infraConfig.Status.Platform,
```
ironcladlou: Can we go ahead and remove `Platform` from this struct? I don't think it should ever have been there. Anything that needs an `Infrastructure` should be looking it up or taking an argument at this point.

manifests/02-deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           command:
           - cluster-ingress-operator
           env:
+            - name: RELEASE_VERSION
```
ironcladlou: If this isn't mapped somehow in `image-references` how will the value ever change?

pravisankar: As you pointed me to the relevant conversation, good to know that this is the job of the cluster-autoscaler-operator now (https://github.com/openshift/cluster-autoscaler-operator/pull/67/files#diff-ff2a1a1bdac4c0edc4696c1d4bf9d684R29)

cmd/cluster-ingress-operator/main.go
pravisankar: done

pravisankar:
@openshift/sig-network-edge @ironcladlou  updated, ptal

manifests/03-cluster-operator.yaml
```diff
@@ -3,3 +3,7 @@ kind: ClusterOperator
 metadata:
   name: ingress
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
```
ironcladlou: Is this intentional? I don't understand why it's hard-coded here and not empty until updated by the operator

pkg/operator/controller/status.go
```diff
@@ -52,15 +51,23 @@ func (r *reconciler) syncOperatorStatus() error {
...
+		// An available operator resets release version
+		for _, condition := range co.Status.Conditions {
+			if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue {
+				co.Status.Versions = []configv1.OperandVersion{
+					{
+						Name:    "operator",
+						Version: r.OperatorReleaseVersion,
+					},
+					{
+						Name:    "router",
```
ironcladlou: Should we go ahead and start getting in line with https://github.com/openshift/api/pull/219? e.g. router -> ingress-controller?

danehans: +1 on the comment above from @ironcladlou As a newbie, I have find `router` and `clusteringress` confusing. I think `ingress-controller` and `ingress-controller-operator` will be easier for users with a k8s background to understand.

pravisankar: ok, changed operand name to 'ingress-controller'

manifests/03-cluster-operator.yaml
pravisankar: fallout from copy+paste!, these values are no longer hard-coded.

ironcladlou:
Awesome, thanks
/lgtm

PR #151, opened 2019-03-01
Remove unnecessary IsAlreadyExists() checks during resource creation.

pravisankar:
- If we are just creating resource with out get op, then IsAlreadyExists()
check is needed but when we are using get resource and in case of not found
error, creating the resource then this check is unnecessary as we know that
this error will not occur.

- Once we create the resource, we don't need to do get op on the same resource
to get the latest info of the object.

- Remove stale 'MetricsServiceMonitorAsset' constant, no longer needed

### DIFF
--- a/pkg/operator/controller/certificate/default_cert.go
+++ b/pkg/operator/controller/certificate/default_cert.go
@@ -114,9 +114,6 @@ func (r *reconciler) currentRouterDefaultCertificate(ci *ingressv1alpha1.Cluster
 // Returns true if the secret was newly created, otherwise returns false.
 func (r *reconciler) createRouterDefaultCertificate(secret *corev1.Secret) (bool, error) {
 	if err := r.client.Create(context.TODO(), secret); err != nil {
-		if errors.IsAlreadyExists(err) {
-			return false, nil
-		}
 		return false, err
 	}
 	return true, nil
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -34,11 +34,10 @@ const (
 	OperatorRole             = "assets/router/operator-role.yaml"
 	OperatorRoleBinding      = "assets/router/operator-role-binding.yaml"
 
-	MetricsServiceMonitorAsset = "assets/router/metrics/service-monitor.yaml"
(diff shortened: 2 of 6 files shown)


pkg/operator/controller/certificate/default_cert.go
```diff
@@ -114,9 +114,6 @@ func (r *reconciler) currentRouterDefaultCertificate(ci *ingressv1alpha1.Cluster
 // Returns true if the secret was newly created, otherwise returns false.
 func (r *reconciler) createRouterDefaultCertificate(secret *corev1.Secret) (bool, error) {
 	if err := r.client.Create(context.TODO(), secret); err != nil {
-		if errors.IsAlreadyExists(err) {
```
ironcladlou: The caller looks like this:

```go
if created, err := r.createRouterDefaultCertificate(desired); err != nil {
  return false, fmt.Errorf("failed to create default certificate: %v", err)
} else if created {
  r.recorder.Eventf(ci, "Normal", "CreatedDefaultCertificate", "Created default wildcard certificate %q", desired.Name)
  return true, nil
}
```

Doesn't this mean that if the resource already exists, we'll return immediately as if an error occurred? It's not clear to me how the resource already existing is an error. Granted we wouldn't get into this block if the resource already exists according to the preceding fetch, but that isn't a guarantee against cache staleness or timing errors.

Shouldn't all these functions be reentrant?

pravisankar: We are not using informer caches to fetch the resources where there could be staleness or timing errors. controller-runtime is only caching the rest client/mapper handlers but not the items returned for a given resource. So if GET on the resource failed with not found error then it means that the resource is not created and these `IsAlreadyExists(err)` checks are unnecessary because they should not occur. 
I think we need to care about reentrantcy of the reconciliation loop but not for every function that we call as part of reconciliation.
```
Reconciliation iteration 1:  Get some resource X fails and we will create resource X.
Reconciliation iteration 2:  Get resource X will succeed and we will skip create resource.
```

ironcladlou:
/lgtm

Thanks!

PR #153, opened 2019-03-01
Use object from API when recording an event

Miciah:
Avoid "Could not construct reference to: [...] due to: 'selfLink was empty, can't make reference'. Will not report event: [...]" errors from the event recorder by using an object we have gotten from the API, as opposed to one we have constructed ourselves (which is missing the self link).

* `pkg/operator/controller/certificate/ca.go` (`ensureRouterCASecret`): After creating the secret, get the newly created object from the API.  Use the object from the API for the event recorder and for the return value.
* `pkg/operator/controller/certificate/publish_ca.go` (`ensureRouterCAConfigMap`): After creating the configmap, get the newly created object from the API, and use this object for the event recorder. When updating the configmap, use the object we got earlier from the API for the event recorder.
* `pkg/operator/controller/certificate/publish_certs.go` (`ensureRouterCertsGlobalSecret`): After creating the secret, get the newly created object from the API, and use this object for the event recorder. When updating the secret, use the object we got earlier from the API for the event recorder.

---

@ironcladlou, the cited error indicates that I was wrong that `r.client.Update` updates its parameter.  It looks like it mutates the parameter, but evidently not with the updated object.

### DIFF
--- a/pkg/operator/controller/certificate/ca.go
+++ b/pkg/operator/controller/certificate/ca.go
@@ -34,9 +34,15 @@ func (r *reconciler) ensureRouterCASecret() (*corev1.Secret, error) {
 	if created, err := r.createRouterCASecret(desired); err != nil {
 		return nil, fmt.Errorf("failed to create CA secret: %v", err)
 	} else if created {
-		r.recorder.Event(desired, "Normal", "CreatedWildcardCACert", "Created a default wildcard CA certificate")
+		new, err := r.currentRouterCASecret()
+		if err != nil {
+			return nil, err
+		}
+		r.recorder.Event(new, "Normal", "CreatedWildcardCACert", "Created a default wildcard CA certificate")
+		return new, nil
+
 	}
-	return desired, nil
+	return r.currentRouterCASecret()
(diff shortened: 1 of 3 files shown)


Miciah:
The "Could not construct reference" errors are gone from [the operator logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/153/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/556/artifacts/e2e-aws-operator/pods/openshift-ingress-operator_ingress-operator-5c5cc576cb-pnhd4_ingress-operator.log.gz) (which are now 23% to 34% smaller compared to the logs from recent test runs for other PRs).

ironcladlou:
/lgtm

Two things for followup:

1. > the cited error indicates that I was wrong that r.client.Update updates its parameter. It looks like it mutates the parameter, but evidently not with the updated object.

Guess we need to settle this next week once and for all (?)

2. Using `new` for a variable, while legal in this context (weirdly), is something I don't see that often

PR #154, opened 2019-03-04
Bug 1680062: deployment: add ingress controller pod anti-affinity

ironcladlou:
Set anti-affinity on ingress controller pods to enable horizontal scaling.

Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1680062.


assets/router/deployment.yaml
```diff
@@ -15,11 +15,22 @@ spec:
     metadata:
       labels:
         app: router
+        ingress.openshift.io/component: ingress-controller
```
ironcladlou: I realized as part of this work that our `app: router` label is extremely vague. I didn't want to update all the other references as part of this PR, so I did an additive change, and we could do a followup to convert other usages.

assets/router/deployment.yaml
```diff
@@ -15,11 +15,22 @@ spec:
...
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: "kubernetes.io/hostname"
+            labelSelector:
+              matchExpressions:
+              - key: ingress.openshift.io/component
+                operator: In
+                values:
+                - ingress-controller
```
pravisankar: I do not see any issue with the code but do we need to be concerned about performance issues on large clusters?

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/  has this note:
```
Note: Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes.
```

kube-dns adopted and later reverted pod anti-affinity rules:
https://github.com/kubernetes/kubernetes/pull/54166
https://github.com/kubernetes/kubeadm/issues/995

I'm not sure how many nodes do we promise to support as part of OpenShift 4.x installation. Do you think it is safer to add a knob to turn off pod anti-affinity if needed for ingress controller?

ironcladlou: @derekwaynecarr @smarterclayton @pravisankar haven't dug too deeply yet, but looks like the problems currently start at scales at or approaching 5k nodes?

https://github.com/kubernetes/kubernetes/issues/54189

ironcladlou: Decided given the focused optimizations for this particular AntiAffinity rule we're okay with the possible scale implications.

ironcladlou:
We don't have much of a choice but to risk the perf impact right now; we need the ability to predictably spread routers over nodes, and the alternatives would be undesirable things like switch back to daemonset, always use host networking (to indirectly control scheduling), or some yet-to-be-discovered solution.

If we reach scale limitations, we can reassess.

PR #155, opened 2019-03-05
docs: update README

ironcladlou:
Add more detail to the README.

Depends on https://github.com/openshift/api/pull/219.
Will depend on the followup to https://github.com/openshift/api/pull/219 which integrates with the new API.


README.md
```diff
@@ -1,17 +1,121 @@
-# Cluster Ingress Operator
+# OpenShift Ingress Operator
 
-Cluster Ingress Operator deploys and manages [HAProxy](http://www.haproxy.org) to provide highly available network ingress in [OpenShift](https://openshift.io/), enabling both OpenShift [Routes](https://docs.okd.io/latest/rest_api/apis-route.openshift.io/v1.Route.html) and Kubernetes [Ingresses](https://kubernetes.io/docs/concepts/services-networking/ingress/).
 
-The operator tries to be useful out of the box by creating a working default deployment based on the cluster's configuration.
+#### Ingress Operator enables Routes and Kubernetes Ingress on OpenShift.
```
Miciah: This heading seems superfluous and looks kind of strange (it renders in bold-face, so it looks like an emphatic statement rather than a heading).

ironcladlou: Intention is to have a one-sentence tagline/"what is this thing"... recommendations?

ironcladlou:
cc @rjhowe — would love your feedback on whether this is starting to look more useful. Note it references in-flight API changes/docs: https://github.com/openshift/api/pull/219

ironcladlou:
cc @ariordan-redhat

README.md
```diff
@@ -1,17 +1,121 @@
...
+Service](https://kubernetes.io/docs/concepts/services-networking/#loadbalancer)
+and on some platforms offers managed wildcard DNS.
 
-Bugs are tracked in [Bugzilla](https://bugzilla.redhat.com/enter_bug.cgi?product=OpenShift%20Container%20Platform&version=4.0.0&component=Routing).
+![Image of LoadBalancerService](docs/images/endpoint-publishing-loadbalancerservice.png)
+
+#### HostNetwork
+
+The [HostNetwork]() strategy uses host networking to publish the ingress
+controller on node ports where the ingress controller is deployed.
```
Miciah: The reader could interpret "on node ports" to mean "using a nodeport service".  How about "directly on the node host where the ingress controller is deployed"?

README.md
```diff
@@ -1,17 +1,121 @@
...
+
+![Image of HostNetwork](docs/images/endpoint-publishing-hostnetwork.png)
+
+#### Private
+
+The [Private]() strategy does not publish the ingress controller.
+
+![Image of Private](docs/images/endpoint-publishing-private.png)
+
+## Troubleshoot
```
Miciah: "Troubleshooting"

README.md
```diff
@@ -1,17 +1,121 @@
...
+$ oc logs --namespace=openshift-ingress-operator deployments/ingress-operator
+```
+
+To inspect the status of a particular ingress controller:
+
+```shell
+$ oc describe --namespace=openshift-ingress-operator ingresscontroller/<name>
+```
+
+## Contribute
```
Miciah: "Contributing"

README.md
```diff
@@ -1,17 +1,121 @@
...
+
+```shell
+$ oc logs --namespace=openshift-ingress-operator deployments/ingress-operator
+```
+
+To inspect the status of a particular ingress controller:
+
+```shell
+$ oc describe --namespace=openshift-ingress-operator ingresscontroller/<name>
+```
```
Miciah: How about logs for the ingress controller (i.e., router)?
```shell
$ oc logs --namespace=openshift-ingress deployments/router-<name>
```
(Are we going to rename "router" to "ingress-controller"?)

README.md
ironcladlou: Agree... although we might be able to hold off on this strategy entirely for 4.0, need to discuss tomorrow.

README.md
ironcladlou: Waffled on this one... is exposing a potentially volatile naming convention problematic? Is there some other way we could direct users to the logs via the ingresscontroller?  Random ideas:

* Implement `oc logs` for an `ingresscontroller` to abstract discovery of the deployment
* Report the deployment name on status
* Document a stable selector for the deployment to pass to `oc logs --selector`

README.md
```diff
@@ -1,17 +1,121 @@
...
-* On cloud platforms, ingress is exposed externally using Kubernetes [LoadBalancer Services](https://kubernetes.io/docs/concepts/services-networking/#loadbalancer).
-* On Amazon AWS, the operator publishes a wildcard DNS record based on the [cluster ingress domain](https://github.com/openshift/api/blob/master/config/v1/types_ingress.go) and pointing to the service's load-balancer.
+Ingress Operator enables external traffic to reach the services that expose
+[Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) or
+[Route](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html)
+resources.
 
-## How to help
+To provide this functionality, Ingress Operator deploys and manages the
+[OpenShift router](https://github.com/openshift/router) — a
```
Miciah: "a"→"an"

README.md
Miciah: "Ingress Operator enables external access to cluster services by configuring Ingress Controllers, which route traffic as specified by OpenShift [Route](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html)  and Kubernetes [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources."

README.md
```diff
@@ -1,17 +1,121 @@
...
+```
+
+## Manage
+
+Ingress Operator provides the OpenShift [ingresscontroller API](). Create and
+edit `ingresscontroller.operator.openshift.io` resources to manage ingress
+controllers.
+
+Interact with ingress controllers using the `oc` command. Every ingress
+controller lives in the `openshift-ingress-operator` namespace.
```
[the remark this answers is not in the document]
Miciah: For now, we do not handle a `ingresscontroller` it if it is not in `openshift-ingress-operator`, so I think this statement is fine for now... although there is some ambiguity between the `ingresscontroller` resource and the ingress controller (née router) itself.

README.md
```diff
@@ -1,17 +1,121 @@
...
+#### HostNetwork
+
+The [HostNetwork]() strategy uses host networking to publish the ingress
+controller on node ports where the ingress controller is deployed.
+
+![Image of HostNetwork](docs/images/endpoint-publishing-hostnetwork.png)
+
+#### Private
+
+The [Private]() strategy does not publish the ingress controller.
```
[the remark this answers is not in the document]
Miciah: On some clusters, the administrator wants to run an ingress controller that serves only to pods inside the cluster.

README.md
```diff
@@ -1,17 +1,121 @@
...
+
+![Image of Private](docs/images/endpoint-publishing-private.png)
```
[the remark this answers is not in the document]
Miciah: There is the "internal" service, which we create irrespective of the endpoint publishing strategy; might be worth adding to the diagrams.

README.md
```diff
@@ -1,17 +1,121 @@
...
+controllers. To view the default ingress controller, use the `oc` command:
+
+```shell
+$ oc describe --namespace=openshift-ingress-operator ingresscontroller/default
+```
+
+## Manage
+
+Ingress Operator provides the OpenShift [ingresscontroller API](). Create and
+edit `ingresscontroller.operator.openshift.io` resources to manage ingress
```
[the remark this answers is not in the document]
Miciah: Agreed, it would be clearer to say, "Ingress Operator _implements_ the OpenShift ingresscontroller API."

README.md
ironcladlou: Updated... better?

README.md
ironcladlou: Fixed

README.md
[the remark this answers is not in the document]
ironcladlou: Updated, PTAL

README.md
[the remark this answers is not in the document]
ironcladlou: Or as a way to manage it yourself through some unmanaged solution (e.g. wrap it with a NodePort)

README.md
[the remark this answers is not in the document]
ironcladlou: Do admins care about the internal service? Is the internal service useful for anything but communications between the ingress controller and other OpenShift components (e.g. prometheus)? Would the internal service be a candidate for a separate internal architecture design?

README.md
```diff
@@ -1,17 +1,115 @@
...
+Create and edit `ingresscontroller.operator.openshift.io` resources to manage
+ingress controllers.
+
+Interact with ingress controllers using the `oc` command. Every ingress
+controller lives in the `openshift-ingress-operator` namespace.
+
+To scale an ingress controller:
+
+```shell
+$ oc scale --replicas=1 --namespace=openshift-ingress-operator ingresscontroller/<name>
```
Miciah: Link to https://github.com/kubernetes/kubernetes/pull/75210?

README.md
```diff
@@ -1,17 +1,115 @@
...
+out.
+
+Refer to the [ingresscontroller API](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go) for full details on defaults and
+customizing an ingress controller. The most important initial customizations are
+domain and endpoint publishing strategy, as they *cannot currently be changed
+after the ingress controller is created*.
+
+### Endpoint publishing
+
+The `.spec.endpointPublishingStrategy` field is used to publish the ingress
```
Miciah: I'm not sure what "publish the ingress controller endpoints to other networks" means; is that referring to DNS?

pkg/operator/controller/controller.go
```diff
@@ -173,6 +173,12 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *operatorv1.IngressControl
 		updated.Status.Domain = ingressConfig.Spec.Domain
 	}
 	// TODO Validate and check for conflicting claims.
+
+
+	// Get all ingresscontrollers
+	// if one already exists with same ci.spec.domain, set status, log event, etc.
+	// and skip reconciliation
```
Miciah: Oops!

Miciah:
What do you think of reversing the order in which the publish strategies are described, in order to have a progress from simplest to most complex?

README.md
danehans: The name of the operator repo is "cluster-ingress-operator". The title of the doc is "OpenShift Ingress Operator". `ClusterIngress` has now transitioned to `IngressController`. OpenShift includes CR's for `cluster`/`ingresses.config.openshift.io` and supports the Kubernetes `Ingress` resource. I can see how a user would get confused. Should the name of the repo and readme title be more closely align to `IngressController`? For example, repo name "ingress-controller-operator" and title "OpenShift Ingress Controller Operator"?

README.md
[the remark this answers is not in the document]
Miciah: The internal service is the only predictable endpoint by which pods can connect to a private ingresscontroller, without additional configuration on the part of the cluster administrator.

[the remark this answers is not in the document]
ironcladlou: What pods?

README.md
Miciah: Changing the name of the repository is a more onerous endeavor and goes beyond the scope of this PR IMO.  Otherwise, the naming makes sense to me: the ingress operator manages the ingress controller, which works with the ingress (sort of) and (really) route resources.

README.md
ironcladlou: Fixed throughout

README.md
ironcladlou: Fixed

README.md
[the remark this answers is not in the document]
Miciah: Any pods on the cluster, infrastructure or end-user, that need to connect to any routes that are served by the private ingresscontroller.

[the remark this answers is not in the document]
Miciah: It may be that we want to recommend a different approach (other than using the internal service that the operator creates) for this use-case.

README.md
ironcladlou: Ingress Controller Operator vs. Ingress Operator sounds like a good list discussion.

Changing the labels, annotations, and resource names would have a few considerations, mostly self-contained to this project, although we have to be careful about resource names (which I believe are currently being used as an informal contract in some cases and which could potentially be replaced by stable label selectors).

Changing the repo name would have enough ripple effects to warrant its own design document (this isn't meant to be snarky — just a statement of fact for consideration).

README.md
ironcladlou: Now we're showing two ways and making a note about `oc scale`

README.md
ironcladlou: Basically lifted from our new [public API](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L76) docs; do you feel the API docs are similarly confusing? What would you propose here as an alternative, and would you suggest we upstream a change as well?

pkg/operator/controller/controller.go
ironcladlou: 🤦‍♂️

ironcladlou:
>What do you think of reversing the order in which the publishing strategies are described, in order to have a progression from simplest to most complex?

The current way is what I would characterize as "likely the default/most features first" to "things you only see if you customize last".

Do you see the strategies as a progressive series of layered concepts? It's not clear to me that they're layered so much as just "different"

README.md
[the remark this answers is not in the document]
ironcladlou: Is it fair to characterize the internal service as an implementation detail at this time?

Miciah:
> Do you see the strategies as a progressive series of layered concepts? It's not clear to me that they're layered so much as just "different"

I do at a conceptual level, although there are discrepancies in the implementation (host networking versus node ports) and presentation (showing multiple nodes versus leaving it implicit) that diminish this viewpoint.  I don't feel strongly if you want to leave this as it is.

PR #157, opened 2019-03-06
api: refactor to ingresscontroller.operator.openshift.io API

ironcladlou:
Replace the v1alpha1 clusteringress.ingress.openshift.io type with the new
ingresscontroller.operator.openshift.io API from openshift/api.

This commit is intended to be the absolute minimum work to switch to the new
API. Internal references to clusteringresses, routers, and any user-facing
references to things renamed in the new API are not covered by this commit.

Depends on https://github.com/openshift/cluster-ingress-operator/pull/156.


pkg/operator/controller/controller_router_deployment.go
```diff
@@ -172,8 +172,11 @@ func desiredRouterDeployment(ci *ingressv1alpha1.ClusterIngress, routerImage str
 		})
 	}
 
-	replicas := ci.Spec.Replicas
-	deployment.Spec.Replicas = &replicas
+	var desiredReplicas int32 = 2
+	if ci.Spec.Replicas != nil {
+		desiredReplicas = *ci.Spec.Replicas
+	}
+	deployment.Spec.Replicas = &desiredReplicas
```
ironcladlou: Subtle but important difference from prior behavior here, whereby we defaulted `.spec.replicas` (which wasn't a pointer)

pkg/operator/controller/names.go
```diff
@@ -74,10 +74,9 @@ func RouterOperatorGeneratedDefaultCertificateSecretName(ci *ingressv1alpha1.Clu
 
 // RouterEffectiveDefaultCertificateSecretName returns the namespaced name for
 // the in-use router default certificate secret.
-func RouterEffectiveDefaultCertificateSecretName(ci *ingressv1alpha1.ClusterIngress, namespace string) types.NamespacedName {
-	name := RouterOperatorGeneratedDefaultCertificateSecretName(ci, namespace).Name
-	if ci.Spec.DefaultCertificateSecret != nil && len(*ci.Spec.DefaultCertificateSecret) > 0 {
-		name = *ci.Spec.DefaultCertificateSecret
+func RouterEffectiveDefaultCertificateSecretName(ci *operatorv1.IngressController, namespace string) types.NamespacedName {
+	if cert := ci.Spec.DefaultCertificate; cert != nil {
```
ironcladlou: Important difference here, as we decided you can't reference an empty name (which would cause you to use the image's cert)

pkg/operator/controller/controller_router_deployment.go
Miciah: Since the defaulting logic is now in `desiredRouterDeployment`, we should add a check to `TestDesiredRouterDeployment` where `ci.Spec.Replicas` is nil.

Miciah:
Need to update the resource name in the manifests for the CRD, default CR, and clusterrole.

Miciah:
Need to fix `statusReplicasPath` in the CRD.

ironcladlou:
unit and e2e are now passing locally (although I discovered `TestRouterCACertificate` is flaky, possibly because of its reliance on DNS for accessing the ingress controller through the LB service).

Squashed, removed WIP — I think it's ready for review.

ironcladlou:
Aaaand when doing one last-minute check I noticed we're still defaulting `nodePlacement` on spec. So that'll need fixed too.

ironcladlou:
I think the `nodePlacement` defaulting issue is a problem on master as well — would a followup be okay?

pkg/manifests/manifests.go
```diff
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"io"
 
-	ingressv1alpha1 "github.com/openshift/cluster-ingress-operator/pkg/apis/ingress/v1alpha1"
+	operatorv1 "github.com/openshift/api/operator/v1"
```
Miciah: This can be grouped with `routev1` below, also from openshift/api.  Similar goes for many other files.  We can get this class of changes in a follow-up though.

pkg/operator/controller/controller.go
```diff
@@ -184,30 +184,32 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *ingressv1alpha1.ClusterIn
...
+	if ci.Status.EndpointPublishingStrategy != nil {
 		return nil
 	}
 
 	updated := ci.DeepCopy()
 	switch {
-	case ci.Spec.HighAvailability != nil:
-		updated.Status.HighAvailability.Type = ci.Spec.HighAvailability.Type
+	case ci.Spec.EndpointPublishingStrategy != nil:
+		updated.Status.EndpointPublishingStrategy = ci.Spec.EndpointPublishingStrategy
```
Miciah: Would be safer to use `.DeepCopy()` here.

test/e2e/operator_test.go
```diff
@@ -299,10 +297,10 @@ func TestClusterIngressUpdate(t *testing.T) {
 		t.Fatalf("failed to get CA certificate configmap: %v", err)
 	}
 
-	originalSecret := ci.Spec.DefaultCertificateSecret
+	originalSecret := ci.Spec.DefaultCertificate
```
Miciah: Using `DeepCopy()` here would feel more correct.  (`DeepCopy()` will return a nil value if the input is nil, so the logic would still be correct.)

Miciah:
> I think the `nodePlacement` defaulting issue is a problem on master as well — would a followup be okay?

`nodePlacement` defaulting seems orthogonal to this PR.  Fixing it in a separate PR makes sense.

Miciah:
Not sure what's going on with the test failure.  [clusteroperators.json](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/157/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/589/artifacts/e2e-aws-operator/clusteroperators.json) has the expected condition:

```json
        {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-03-09T01:21:50Z",
                "generation": 1,
                "name": "ingress",
                "resourceVersion": "13034",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/ingress",
                "uid": "b6ee84c8-4209-11e9-9ea1-1252ccb292d2"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-03-09T01:21:52Z",
                        "status": "False",
                        "type": "Failing"
                    },
                    {
                        "lastTransitionTime": "2019-03-09T01:21:52Z",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-03-09T01:22:36Z",
                        "status": "True",
                        "type": "Available"
                    }
                ],
                // ...                
```
We might simply need to increase the timeout in `TestOperatorAvailable`.

/test e2e-aws-operator

pkg/manifests/manifests.go
ironcladlou: My intention was to avoid a bunch of reformatting diffs here — followup sounds good

pkg/operator/controller/controller.go
ironcladlou: Good catch, fixed

test/e2e/operator_test.go
ironcladlou: Fixed

ironcladlou:
Feedback addressed, PTAL

PR #158, opened 2019-03-07
Moves default ingress from startup to reconcile

danehans:
Fixes "The ingress operator should recreate the default without user intervention" part of bug:  https://bugzilla.redhat.com/show_bug.cgi?id=1686204

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -64,7 +64,7 @@ func NewFactory(config operatorconfig.Config) *Factory {
 	return &Factory{config: config}
 }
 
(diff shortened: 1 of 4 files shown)


ironcladlou:
Had a discussion offline... creation of the default instance is orthogonal to the reconciliation of an instance —for example, one option is to spin up another goroutine during operator startup which periodically checks to ensure the default exists (and creates it if not).

ironcladlou:
Here's another way: https://github.com/openshift/cluster-ingress-operator/pull/172

danehans:
Closed in favor of https://github.com/openshift/cluster-ingress-operator/pull/172

PR #160, opened 2019-03-08
Bug 1679629: manifests: remove run-level namespace label

ironcladlou:
The operator runs after the apiserver and so doesn't need to skip
admission/SCC checks.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1679629


ironcladlou:
Not sure about those failures, let's see what happens if I
/retest

PR #161, opened 2019-03-08
deps: bump openshift/api

ironcladlou:
Update openshift/api to incorporate https://github.com/openshift/api/pull/238.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -341,14 +341,14 @@
   version = "1.0.1"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


Miciah:
> Update openshift/api to incorporate [openshift/api#238](https://github.com/openshift/api/pull/238).

Doesn't look like it incorporates [openshift/api#238](https://github.com/openshift/api/pull/238).

ironcladlou:
openshift/api@master is at fe0c0ad070ce15d30b0a6dd20adb0ef883efe11a... ingress isn't in vendor because dep prunes unused packages.

Miciah:
Hm, I understand.  Seems a little weird, but all right.

/lgtm

PR #162, opened 2019-03-12
Allow the router to use the hostnetwork SCC

Miciah:
Add RBAC rules to allow routers to use the hostnetwork security context constraint.

Previously the namespace had been labeled with `openshift.io/run-level=0`, but https://github.com/openshift/cluster-ingress-operator/pull/160 removed this label, and so it is necessary to grant the deployment explicit permission to use the SCC.

This change prevents the deployment from failing with the following error when the ingresscontroller is configured to use the HostNetwork endpoint publishing strategy type:

    unable to validate against any security context constraint: [provider restricted: .spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 80: Host ports are not allowed to be used spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 443: Host ports are not allowed to be used spec.containers[0].securityContext.containers[0].hostPort: Invalid value: 1936: Host ports are not allowed to be used]

* `assets/router/cluster-role.yaml`: Add a new rule to allow the router to use the hostnetwork SCC.
* `manifests/00-cluster-role.yaml`: Mirror the new rule from the asset.
* `test/e2e/operator_test.go` (`TestHostNetworkEndpointPublishingStrategy`): New test.


Miciah:
release "release-latest" failed: pod release-latest was already deleted

/retest

PR #163, opened 2019-03-12
Move nodePlacement defaulting to the controller

Miciah:
Instead of having a node selector in the default ingresscontroller asset and the deployment asset, set a default node selector in the controller.

* `assets/defaults/cluster-ingress.yaml`: Delete `nodePlacement` stanza.
* `assets/router/deployment.yaml`: Delete `nodeSelector` stanza.
* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): Set a default node selector on the deployment if the ingresscontroller does not specify an explicit node selector.


pkg/operator/controller/controller_router_deployment.go
```diff
@@ -147,17 +147,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, routerImage strin
...
+			nodeSelector, err = metav1.LabelSelectorAsMap(ci.Spec.NodePlacement.NodeSelector)
 			if err != nil {
 				return nil, fmt.Errorf("clusteringress %q has invalid spec.nodePlacement.nodeSelector: %v",
 					ci.Name, err)
 			}
-
-			deployment.Spec.Template.Spec.NodeSelector = nodeSelector
 		}
 	}
+	deployment.Spec.Template.Spec.NodeSelector = nodeSelector
```
ironcladlou: Just a random thought... I wonder how changing this affects support commitments. For example, we don't support scheduling the ingresscontrollers on masters.

Miciah: > Just a random thought... I wonder how changing this affects support commitments. For example, we don't support scheduling the ingresscontrollers on masters.

This reminds me of our earlier merge-versus-replace discussion.  Should the API even allow disabling the `node-role.kubernetes.io/worker` selector if an ingresscontroller without it is unsupported?

PR #164, opened 2019-03-13
Updates DNS deletion to support not found records

danehans:
Previously, an ingresscontroller delete would hang when the
operation is performed prior to the DNS alias record being
created.

Fixes bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1683515

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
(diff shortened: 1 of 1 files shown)


pkg/dns/aws/dns.go
```diff
@@ -308,6 +309,14 @@ func (m *Manager) updateAlias(domain, zoneID, target, targetHostedZoneID, action
 		},
 	})
 	if err != nil {
+		if action == string(deleteAction) {
+			if aerr, ok := err.(awserr.Error); ok {
+				if strings.Contains(aerr.Message(), "not found") {
```
ironcladlou: Ouch, so there's no specific error type to check for this?

ironcladlou: https://docs.aws.amazon.com/sdk-for-go/api/service/route53/#Route53.ChangeResourceRecordSets

>* ErrCodeNoSuchHostedZone "NoSuchHostedZone"
>No hosted zone exists with the ID that you specified.

Is this the error type we're interested in?

ironcladlou:
/lgtm

Can do followups to improve the logging and any factoring changes that would be useful

PR #166, opened 2019-03-13
Bug 1687640: Fix creation of service monitor to report router metrics

pravisankar:
- Use a new kube client for fetching and creating ServiceMonitor resource
to avoid cached REST scheme/mapper of controller client.

### DIFF
--- a/pkg/operator/client/client.go
+++ b/pkg/operator/client/client.go
@@ -1,0 +1,56 @@
+package client
+
+import (
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+var (
+	// scheme contains all the API types necessary for the operator's dynamic
+	// clients to work. Any new non-core types must be added here.
+	//
+	// NOTE: The discovery mechanism used by the client won't automatically refresh,
+	// so only add types here that are _guaranteed_ to exist before the operator
+	// starts.
+	scheme *runtime.Scheme
+)
+
+func init() {
+	scheme = kscheme.Scheme
+	if err := operatorv1.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+	if err := configv1.Install(scheme); err != nil {
+		panic(err)
+	}
+}
+
+func GetScheme() *runtime.Scheme {
+	return scheme
+}
+
+// NewClient builds an operator-compatible kube client from the given REST config.
+func NewClient(kubeConfig *rest.Config) (client.Client, error) {
+	mapper, err := apiutil.NewDiscoveryRESTMapper(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to discover api rest mapper: %v", err)
+	}
+	kubeClient, err := client.New(kubeConfig, client.Options{
+		Scheme: scheme,
+		Mapper: mapper,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kube client: %v", err)
+	}
+	return kubeClient, nil
+}
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/cluster-ingress-operator/pkg/dns"
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -16,12 +17,14 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 
+	"k8s.io/client-go/rest"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -45,8 +48,13 @@ var log = logf.Logger.WithName("controller")
 // The controller will be pre-configured to watch for ClusterIngress resources
 // in the manager namespace.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
+	kubeClient, err := operatorclient.NewClient(config.KubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kube client: %v", err)
+	}
 	reconciler := &reconciler{
 		Config:   config,
+		client:   kubeClient,
 		recorder: mgr.GetRecorder("operator-controller"),
 	}
 	c, err := controller.New("operator-controller", mgr, controller.Options{Reconciler: reconciler})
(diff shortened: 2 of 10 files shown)


ironcladlou:
Did you notice a perf hit when creating the new client each time? 

Other options:
* Wire in a CachedDiscoveryInterface which is periodically refreshed
* ??

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -8,23 +8,36 @@ import (
...
 
 func (r *reconciler) ensureServiceMonitor(ci *operatorv1.IngressController, svc *corev1.Service, deploymentRef metav1.OwnerReference) (*unstructured.Unstructured, error) {
 	desired := desiredServiceMonitor(ci, svc, deploymentRef)
 
-	current, err := r.currentServiceMonitor(ci)
+	// ServiceMonitor CRD may not be created by the time r.Client is created
+	// and since r.Client will not refresh scheme/mapper fields after creation,
+	// ServiceMonitor GVK may not be recognized.
+	// So we use a new kube client everytime to get or create ServiceMonitor resource.
+	kubeClient, err := operatorclient.NewClient(r.KubeConfig)
```
ironcladlou: from my own local testing:

```
2019-03-15T11:17:10.143-0400    INFO    operator        log/log.go:26   started zapr logger
2019-03-15T11:17:10.154-0400    INFO    operator.entrypoint     cluster-ingress-operator/main.go:45     config.GetConfig()      {"duration": "1.61447ms"}                                                      
2019-03-15T11:17:12.155-0400    INFO    operator.entrypoint     cluster-ingress-operator/main.go:52     operator.Client()       {"duration": "2.000241625s"}
```

My suspicion is that we're introducing a 2s floor time to every call to ensureServiceMonitor

Miciah: Could we add something like `if errors.IsNoMatchError(err) { // recreate r.Client and retry` to the error check in `currentServiceMonitor`?

pravisankar:
/hold cancel
@openshift/sig-network-edge ready for re-review, PTAL

pkg/operator/client/client.go
```diff
@@ -0,0 +1,61 @@
...
+	return scheme
+}
+
+// NewClient builds an operator-compatible kube client from the given REST config.
+func NewClient(kubeConfig *rest.Config) (client.Client, error) {
+	managerOptions := manager.Options{
+		Scheme:         scheme,
+		MapperProvider: apiutil.NewDiscoveryRESTMapper,
+	}
+	mapper, err := managerOptions.MapperProvider(kubeConfig)
```
Miciah: This is just a suggestion for a follow-up: Do we need `managerOptions`, or would the following work?
```go
	mapper, err := apiutil.NewDiscoveryRESTMapper(kubeConfig)
```

pkg/operator/controller/controller.go
```diff
@@ -61,7 +63,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 
 // Config holds all the things necessary for the controller to run.
 type Config struct {
-	Client                 client.Client
+	Client client.Client
+	// ServiceMonitor resource may not be registerd by the time 'Client' is created
+	// and since 'Client' will not refresh scheme/mapper fields after creation,
+	// ServiceMonitor GVK may not be recognized.
+	// Enabling 'Client' refresh needs locking as this field can be accessed in parallel
+	// by multiple go routines. Instead we are creating 'ServiceMonitorClient' for operating
```
Miciah: What does "'Client' refresh" refer to? Do you mean writing to this struct's `Client` field? The struct is not shared with other controllers, so you must be talking about synchronizing access within this controller, right? How does using a separate field avoid the the synchronization problem?

pravisankar: Yes, Client refresh in this context is to modify the fields in `Client` field. 
I was wrongly assumed that in this controller we are trying to ensure default cluster ingress that uses Client to get and create ingress resource if needed. We are running the controller in single threaded mode so we can modify the Client field and don't need to worry about synchronization issue. I will make this change.
Not related to this, but I'm not sure why we are trying to ensure default cluster ingress by looping every minute. May be this can be part of reconciliation (I thought Dane opened a pr?) or can do watch on default ingress to avoid unnecessary looping for this corner case.

Miciah: > Not related to this, but I'm not sure why we are trying to ensure default cluster ingress by looping every minute. May be this can be part of reconciliation (I thought Dane opened a pr?) or can do watch on default ingress to avoid unnecessary looping for this corner case.

That approach has a bootstrap problem: Reconciliation is triggered by an ingresscontroller-created event, so something else needs to create the default ingresscontroller.  We could perhaps use a synthetic event to the controller, or use a watch on a separate channel source, which the operator could use to trigger the first reconciliation, but using a separate goroutine is simple and good enough for now.

ironcladlou:
Since this approach already introduces the restConfig to the controller, would it be easier to have the controller build its own client at construction from the provided rest config, protect it with a lock, and rebuild the single universal client whenever we want (on a period or whenever we detect stale discovery info)?

Drawback would be that the controller client isn't shared (but we're using a non-caching client anyway)

pkg/operator/client/client.go
pravisankar: done

pravisankar:
@Miciah @ironcladlou updated, PTAL

pkg/operator/controller/controller_service_monitor.go
```diff
@@ -85,6 +88,20 @@ func (r *reconciler) currentServiceMonitor(ci *operatorv1.IngressController) (*u
 		Version: "v1",
 	})
 	if err := r.Client.Get(context.TODO(), serviceMonitorName(ci), sm); err != nil {
+		if meta.IsNoMatchError(err) {
+			// Refresh kube client with latest rest scheme/mapper.
+			kClient, err := operatorclient.NewClient(r.KubeConfig)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create kube client: %v", err)
+			}
+			r.Client = kClient
```
ironcladlou: If we're not going to guard, should we at least note that this is not thread safe?

pkg/operator/controller/controller.go
```diff
@@ -61,7 +63,13 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 
 // Config holds all the things necessary for the controller to run.
 type Config struct {
+	// Client is the kube Client and it will refresh scheme/mapper fields if needed
+	// to detect some resources like ServiceMonitor which could get registered after
+	// the client creation.
+	// Since this controller is running in single threaded mode,
+	// we do not need to synchronize when changing rest scheme/mapper fields.
 	Client                 client.Client
```
ironcladlou: Do we need to pass client anymore? I think accepting one implies you can share a given one, but we're going to arbitrarily reassign the reference. Could keep it private and construct one in New.

Miciah:
Need to update `operator_test.go`, which is still using `operator.Client` where `operator` refers to `github.com/openshift/cluster-ingress-operator/pkg/operator`.

Miciah:
Also, the title should have "Bug 1687640:" rather than "Bug 1687640 - "—changelog generation should recognize the former, but I don't know whether it will recognize the latter.

pkg/operator/controller/controller.go

pkg/operator/controller/controller_service_monitor.go
pravisankar: We have noted that this client field is not thread safe in the field declaration.

pravisankar:
@Miciah @ironcladlou ready for another round of review, PTAL

Miciah:
> the title should have "Bug 1687640:"

Sorry, this was ambiguous.  Specifically the title of the PR should follow that format (changelog generation uses the PR title, not the commit title).

Miciah:
Needs rebase, but [the operator logs from the last e2e-aws-operator test run](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/166/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/675/artifacts/e2e-aws-operator/pods/openshift-ingress-operator_ingress-operator-85dd969557-cm2xz_ingress-operator.log.gz) look good: I see 5 "no matches for kind" errors (first one at 2019-03-21T00:54:47.368Z) and then "created servicemonitor" (at 2019-03-21T00:54:57.188Z).

ironcladlou:
lgtm once the rebase is done!

pravisankar:
@Miciah @ironcladlou rebased, ptal

Miciah:
BTW, [the operator logs from the last e2e-aws-operator run](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/166/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/679/artifacts/e2e-aws-operator/pods/openshift-ingress-operator_ingress-operator-66589c546d-m9cmj_ingress-operator.log.gz) show 18 "no matches for kind" errors, starting at  2019-03-21T17:03:06.909Z, and then "created servicemonitor" at 2019-03-21T17:04:15.447Z.  It appears to take a surprisingly long time for the ServiceMonitor API to get registered, but at least on our side we now appear to be retrying correctly and eventually succeeding.

PR #167, opened 2019-03-13
Bug 1687940: deployment: fix scope of pod anti-affinity

ironcladlou:
Before, the anti-affinity rule for ingress controller deployment pods was such
that no controller pods could colocate on a node, even if they were for
different ingress controllers. Now, pods for different ingress controllers can
be colocated on a node.

Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1687940.


ironcladlou:
Not seeing anything suspicious about the router in the logs, the flakes are concerning...

/retest

Miciah:
Need to fix the label selectors in `RouterServiceInternal` and `desiredLoadBalancerService`.

ironcladlou:
Good catch — there might be a gap here in how we construct the selector if it must be shared

ironcladlou:
Well, turns out that the YAML/manifestfactory business was hiding an entanglement of the deployment pod selector (i.e. in addition to the deployment itself, the internal service also needs to match the deployment pods). So I gave the internal service the usual refactoring treatment and extracted the naming and deployment pod selector stuff into `names.go`.

A rotten cluster has stymied my testing, so might have to continue this one on Monday (unless it happens that the CI tests pass and there are no issues during code review).

pkg/operator/controller/controller_internal_service.go
```diff
@@ -0,0 +1,77 @@
...
+		manifests.OwningClusterIngressLabel: ic.Name,
+	}
+
+	s.Annotations = map[string]string{
+		// TODO: remove hard-coded name
+		ServingCertSecretAnnotation: fmt.Sprintf("router-metrics-certs-%s", ic.Name),
+	}
+
+	s.Spec.Selector = map[string]string{
+		s.Spec.Selector[controllerDeploymentLabel]: IngressControllerDeploymentLabel(ic),
```
ironcladlou: I think this is already an improvement, but another possibility I see is to use the selector from the deployment itself here (perhaps by extracting a function into `names.go` which returns the deployment selector.

ironcladlou: I went ahead and did this refactor too

ironcladlou:
So here's my post-hoc justification for increasing the complexity of the PR:

The affinity setup in the manifest YAML is already pretty complex, and at the time I introduced it I thought the value would be static. However, as the bug shows, some of the affinity definition is actually dynamic as a function of the ingress controller. One option would have been to leave the static parts in YAML and mutate just the dynamic piece at runtime, and there's precedent within the codebase for that. However, based on previous conversations, I thought it would be easier to comprehend if we kept the whole affinity piece defined in code closer to the mutation. I also cleaned up the deployment selector names as part of that, which I thought would be trivial.

But, during the course of the refactor, it became clear that the deployment pod selector is used in many contexts, and the as a result I ended up having to touch the related pieces.

I'm hoping the improved clarity of coupling between these parts will justify the expanded scope of the changes. Thoughts?

pkg/operator/controller/controller_internal_service.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+const (
+	// Annotation used to inform the certificate generation service to
+	// generate a cluster-signed certificate and populate the secret.
+	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+)
+
+// ensureInternalRouterServiceForIngress ensures that an internal service exists
+// for a given ClusterIngress.
```
Miciah: "ClusterIngress"→"IngressController"

assets/router/service-internal.yaml
```diff
@@ -5,12 +5,8 @@ apiVersion: v1
 metadata:
   # Name and Annotations are set at runtime.
   namespace: openshift-ingress
```
Miciah: Specifying `namespace` in the asset is redundant now that `desiredInternalIngressControllerService` sets it.

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -78,16 +78,9 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, routerImage strin
...
-			controllerDeploymentLabel: IngressControllerDeploymentLabel(ci),
-		},
-	}
-	deployment.Spec.Template.Labels = map[string]string{
-		// ensure pods are matched to be controlled by the deployment
-		controllerDeploymentLabel: IngressControllerDeploymentLabel(ci),
-	}
+	deployment.Spec.Selector = IngressControllerDeploymentPodSelector(ci)
+
+	deployment.Spec.Template.Labels = IngressControllerDeploymentPodSelector(ci).MatchLabels
```
Miciah: Looks like you could even simplify this to `deployment.Spec.Template.Labels = deployment.Spec.Selector.MatchLabels`.

Miciah:
`test/e2e/operator_test.go` is full of references to the old `router-...` names.

pkg/operator/controller/controller_internal_service.go
ironcladlou: Fixed

assets/router/service-internal.yaml
ironcladlou: Fixed

pkg/operator/controller/controller_router_deployment.go
ironcladlou: Changed

ironcladlou:
Updated `test/e2e/operator_test.go` to use naming functions for the most part (and removed a redundant test).

Ready for review again.

ironcladlou:
Looks like CI account issues.

/retest

test/e2e/operator_test.go
```diff
@@ -175,69 +175,6 @@ func TestClusterIngressControllerCreateDelete(t *testing.T) {
 	}
 }
 
-func TestRouterServiceInternalEndpoints(t *testing.T) {
```
danehans: Curious to know why this test is being removed.

ironcladlou: IMO the test isn't buying us much other than asserting the service exists, which is covered implicitly elsewhere. It did check the endpoints, but that's such a specific test you could cover with a unit (e.g. do the selectors match) and there's origin e2e coverage for metrics which actually uses the service ([currently broken](
https://bugzilla.redhat.com/show_bug.cgi?id=1687640)).

ironcladlou: Spoke offline and decided to do a followup which introduces a new test variant which exercises the metrics endpoint through the service (which will both implicitly test the endpoints and also the actual functional area around the internal service, which is enabling metrics integration)

ironcladlou:
Still trying to get a cluster to try and reproduce any issues. Operator logs and router logs look fine...

/retest

ironcladlou:
Okay, finally reproduced some of these in a local cluster.

ironcladlou:
https://github.com/openshift/origin/blob/master/test/extended/router/router.go#L112 :(

assets/router/service-internal.yaml
```diff
@@ -4,13 +4,8 @@ kind: Service
 apiVersion: v1
 metadata:
   # Name and Annotations are set at runtime.
-  namespace: openshift-ingress
-  labels:
-    app: router
 spec:
   type: ClusterIP
-  selector:
-    app: router
```
pravisankar: Recent changes are hinting us that we don't have much value in creating assets anymore.
Probably we should ditch assets and create objects internally in the code.

pkg/operator/controller/controller_internal_service.go
```diff
@@ -0,0 +1,75 @@
...
+const (
+	// Annotation used to inform the certificate generation service to
+	// generate a cluster-signed certificate and populate the secret.
+	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
+)
+
+// ensureInternalRouterServiceForIngress ensures that an internal service exists
+// for a given IngressController.
+func (r *reconciler) ensureInternalIngressControllerService(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) (*corev1.Service, error) {
+	desired := desiredInternalIngressControllerService(ic, deploymentRef)
```
pravisankar: This line can be moved to line 33 where we use this desired object.

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -73,20 +73,37 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, routerImage strin
 	deployment.Name = name.Name
 	deployment.Namespace = name.Namespace
 
-	if deployment.Labels == nil {
-		deployment.Labels = map[string]string{}
+	deployment.Labels = map[string]string{
+		// associate the deployment with the ingresscontroller
+		manifests.OwningClusterIngressLabel: ci.Name,
```
Miciah: Should this be using `IngressControllerDeploymentLabel`?

pkg/operator/controller/controller_internal_service.go
```diff
@@ -0,0 +1,75 @@
...
+func desiredInternalIngressControllerService(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) *corev1.Service {
+	s := manifests.InternalIngressControllerService()
+
+	name := InternalIngressControllerServiceName(ic)
+
+	s.Namespace = name.Namespace
+	s.Name = name.Name
+
+	s.Labels = map[string]string{
+		manifests.OwningClusterIngressLabel: ic.Name,
```
Miciah: Should this be using `IngressControllerDeploymentLabel`?

assets/router/service-internal.yaml
Miciah: If we keep the assets, maybe we could at least use more abstract names, such as `ClusterIPService`, similar to how we have `LoadBalancerService`.  I wouldn't object to removing the assets entirely though.

pkg/operator/controller/controller_router_deployment.go
ironcladlou: Not sure, this is the same as before,  and deploymentLabel is for ingress.operator.openshift.io/ingress-controller-deployment — we might need to audit our labels generallly

ironcladlou: Meant for this comment to apply to controller_internal_service.go

ironcladlou: And no, this was to associate the deployment with the ingresscontroller, which is the intent of the label... IngressControllerDeploymentLabel was meant to be a way to associate other stuff with the deployment itself. The naming is probably poor

pkg/operator/controller/controller_internal_service.go
ironcladlou: I don't know if it matters since there's an owner ref to the deployment already

pkg/operator/controller/controller_router_deployment.go
Miciah: What is the intended use of `IngressControllerDeploymentLabel`? I was assuming it was, "I have an ingresscontroller and need the label I would use to identify it."

Miciah: Sorry, just saw your last comment (GitHub didn't show it when I submitted mine last comment).  I see now, `IngressControllerDeploymentLabel` identifies the deployment, and we need a label that identifies the ingresscontroller (for which we do not yet have a function).  Please disregard my preceding comments.

Miciah:
level=error msg="4 errors occurred:"
    level=error msg="\t* module.vpc.aws_route_table_association.private_routing[2]: 1 error occurred:"
    level=error msg="\t* aws_route_table_association.private_routing.2: timeout while waiting for state to become 'success' (timeout: 5m0s)"
    level=error
    level=error
    level=error msg="\t* module.vpc.aws_route_table_association.route_net[0]: 1 error occurred:"
    level=error msg="\t* aws_route_table_association.route_net.0: timeout while waiting for state to become 'success' (timeout: 5m0s)"
    level=error
    level=error
    level=error msg="\t* module.vpc.aws_route_table_association.route_net[2]: 1 error occurred:"
    level=error msg="\t* aws_route_table_association.route_net.2: timeout while waiting for state to become 'success' (timeout: 5m0s)"
    level=error
    level=error
    level=error msg="\t* module.vpc.aws_route_table_association.route_net[3]: 1 error occurred:"
    level=error msg="\t* aws_route_table_association.route_net.3: timeout while waiting for state to become 'success' (timeout: 5m0s)"

/test e2e-aws-operator

Miciah:
level=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-qdt49y5z-43abb.origin-ci-int-aws.dev.rhcloud.com:6443 to initialize..."

/test e2e-aws-operator

PR #170, opened 2019-03-14
deps: bump to add awserr and remove k8s autoscale

danehans:
awserr is needed due to pr #164

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -904,6 +904,7 @@
   input-imports = [
     "github.com/aws/aws-sdk-go/aws",
     "github.com/aws/aws-sdk-go/aws/arn",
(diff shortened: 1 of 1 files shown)


Gopkg.lock
```diff
@@ -922,7 +923,6 @@
     "github.com/openshift/library-go/pkg/crypto",
     "go.uber.org/zap",
     "k8s.io/api/apps/v1",
-    "k8s.io/api/autoscaling/v1",
```
Miciah: In case anyone is curious, https://github.com/openshift/cluster-ingress-operator/pull/157 dropped this import because we stopped using the scale client in the `TestClusterIngressScale` E2E test while we sort out an API issue that affects CRDs with the `/scale` subresource.

PR #172, opened 2019-03-15
Bug 1686204: operator: periodically ensure default ingresscontroller

ironcladlou:
The default ingresscontroller is only ensured during operator startup,
necessitating an operator restart to restore defaults when the default
ingresscontroller is deleted.

This change is to periodically ensure the default ingresscontroller exists, so
defaults can be restored without affecting operator availability.

Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1686204.


ironcladlou:
The outcome of the default reconcile could potentially be reflected in operator status/conditions, but my immediate goal is to simply fix primary UX problem. Incorporating status will require more smarts/complexity.

pkg/operator/operator.go
```diff
@@ -201,15 +216,31 @@ func (o *Operator) Start(stop <-chan struct{}) error {
 	}
 }
 
-// ensureDefaultClusterIngress ensures that a default ClusterIngress exists.
-func (o *Operator) ensureDefaultClusterIngress() error {
-	ci := manifests.DefaultClusterIngress()
-	err := o.client.Create(context.TODO(), ci)
-	if err != nil && !errors.IsAlreadyExists(err) {
+// ensureDefaultClusterIngress creates the default ingresscontroller if it
```
Miciah: `ensureDefaultClusterIngress`→`ensureDefaultIngressController`

pkg/operator/operator.go
```diff
@@ -201,15 +216,31 @@ func (o *Operator) Start(stop <-chan struct{}) error {
...
-	} else if err == nil {
-		log.Info("created default clusteringress", "namespace", ci.Namespace, "name", ci.Name)
 	}
+	err = o.client.Create(context.TODO(), ic)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			return nil
+		} else {
+			return err
+		}
```
Miciah: Could simplify slightly:
```go
	if err := o.client.Create(context.TODO(), ic); err != nil {
		if errors.IsAlreadyExists(err) {
			return nil
		}
		return err
```

danehans:
lgtm after addressing the two changes @Miciah suggested.

pkg/operator/operator.go
ironcladlou: Fixed

pkg/operator/operator.go
ironcladlou: Fixed

ironcladlou:
All feedback addressed

PR #174, opened 2019-03-18
Bug 1690333: Implement IngressController Available Status Condition

danehans:
Previously, ingresscontroller status conditions were not implemented.

This commit does the following:

- Implements the ingresscontroller `Available` status condition.
- Refactors ingress operator `Progressing` and `Available` status conditions to be calucalted from the ingresscontroller `Available` condition.
- Replaces deployment logic from ingress operator unit tests with ingresscontroller `Available` conditions.
- Adds ingresscontroller `Available` condition unit tests.

Provides the foundation for fixing bug 1683765 by reporting domain conflicts through status conditions.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -11,7 +11,6 @@ import (
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
 
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
 
@@ -395,8 +394,8 @@ func (r *reconciler) ensureClusterIngress(ci *operatorv1.IngressController, dnsC
 			errs = append(errs, fmt.Errorf("failed to integrate metrics with openshift-monitoring for clusteringress %s: %v", ci.Name, err))
 		}
 
-		if err := r.syncClusterIngressStatus(deployment, ci); err != nil {
-			errs = append(errs, fmt.Errorf("failed to sync status of clusteringress %s/%s: %v", deployment.Namespace, deployment.Name, err))
+		if err := r.syncIngressControllerStatus(deployment, ci); err != nil {
+			errs = append(errs, fmt.Errorf("failed to sync ingresscontroller status: %v", err))
 		}
 	}
 
@@ -479,31 +478,6 @@ func (r *reconciler) ensureMetricsIntegration(ci *operatorv1.IngressController, 
 
 	if _, err := r.ensureServiceMonitor(ci, svc, deploymentRef); err != nil {
 		return fmt.Errorf("failed to ensure servicemonitor for %s: %v", ci.Name, err)
-	}
-
-	return nil
-}
-
-// syncClusterIngressStatus updates the status for a given clusteringress.
-func (r *reconciler) syncClusterIngressStatus(deployment *appsv1.Deployment, ci *operatorv1.IngressController) error {
-	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
-	if err != nil {
-		return fmt.Errorf("router deployment %s/%s has invalid spec.selector: %v", deployment.Namespace, deployment.Name, err)
-	}
-
-	if ci.Status.AvailableReplicas == deployment.Status.AvailableReplicas &&
-		ci.Status.Selector == selector.String() {
-		return nil
-	}
-
-	updated := ci.DeepCopy()
-	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
-	updated.Status.Selector = selector.String()
-	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
-		return fmt.Errorf("failed to update status of clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
-	}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: updated.Namespace, Name: updated.Name}, ci); err != nil {
-		return fmt.Errorf("failed to get clusteringress %s/%s: %v", updated.Namespace, updated.Name, err)
 	}
 
 	return nil
--- a/pkg/operator/controller/ingress_status.go
+++ b/pkg/operator/controller/ingress_status.go
@@ -1,0 +1,100 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// syncIngressControllerStatus computes the current status of ic and
+// updates status upon any changes since last sync.
+func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment, ic *operatorv1.IngressController) error {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
+	}
+
+	updated := ic.DeepCopy()
+	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
+	updated.Status.Selector = selector.String()
+	updated.Status.Conditions = computeIngressStatusConditions(updated.Status.Conditions, deployment)
+	if !ingressStatusesEqual(updated.Status, ic.Status) {
+		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+			return fmt.Errorf("failed to update ingresscontroller status: %v", err)
+		}
+	}
+
+	return nil
+}
+
+// computeIngressStatusConditions computes the ingress controller's current state.
+func computeIngressStatusConditions(conditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment) []operatorv1.OperatorCondition {
+	availableCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if deployment.Status.AvailableReplicas > 0 {
+		availableCondition.Status = operatorv1.ConditionTrue
+	} else {
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "DeploymentUnavailable"
+		availableCondition.Message = "no Deployment replicas available"
+	}
+	conditions = setIngressStatusCondition(conditions, availableCondition)
+
+	return conditions
+}
+
+// setIngressStatusCondition returns the IngressController condition result
+// of setting the specified condition in the given slice of conditions.
+func setIngressStatusCondition(oldConditions []operatorv1.OperatorCondition, condition *operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
+	condition.LastTransitionTime = metav1.Now()
+
+	var newConditions []operatorv1.OperatorCondition
+
+	found := false
+	for _, c := range oldConditions {
+		if condition.Type == c.Type {
+			if condition.Status == c.Status &&
+				condition.Reason == c.Reason &&
+				condition.Message == c.Message {
+				// The ingresscontroller status condition has not changed.
+				return oldConditions
+			}
+
+			found = true
+			newConditions = append(newConditions, *condition)
+		} else {
+			newConditions = append(newConditions, c)
+		}
+	}
+	if !found {
+		newConditions = append(newConditions, *condition)
+	}
+
+	return newConditions
+}
+
+// ingressStatusesEqual compares two IngressControllerStatus values.  Returns true
+// if the provided values should be considered equal for the purpose of determining
+// whether an update is necessary, false otherwise.
+func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
+	conditionCmpOpts := []cmp.Option{
+		cmpopts.IgnoreFields(operatorv1.OperatorCondition{}, "LastTransitionTime"),
+		cmpopts.EquateEmpty(),
+		cmpopts.SortSlices(func(a, b operatorv1.OperatorCondition) bool { return a.Type < b.Type }),
+	}
+	if !cmp.Equal(a.Conditions, b.Conditions, conditionCmpOpts...) || a.AvailableReplicas != b.AvailableReplicas ||
+		a.Selector != b.Selector {
+		return false
+	}
+
+	return true
+}
--- a/pkg/operator/controller/ingress_status_test.go
+++ b/pkg/operator/controller/ingress_status_test.go
@@ -1,0 +1,258 @@
+package controller
+
+import (
+	"fmt"
+	"testing"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestComputeIngressStatusConditions(t *testing.T) {
+	testCases := []struct {
+		description     string
+		availRepl, repl int32
+		condType        string
+		condStatus      operatorv1.ConditionStatus
+	}{
+		{"0/2 deployment replicas available", 0, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionFalse},
+		{"1/2 deployment replicas available", 1, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionTrue},
+		{"2/2 deployment replicas available", 2, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionTrue},
+	}
+
+	for i, tc := range testCases {
+		deploy := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("ingress-controller-%d", i+1),
+			},
+			Status: appsv1.DeploymentStatus{
+				Replicas:          tc.repl,
+				AvailableReplicas: tc.availRepl,
+			},
+		}
+
+		expected := []operatorv1.OperatorCondition{
+			{
+				Type:   tc.condType,
+				Status: tc.condStatus,
+			},
+		}
+		actual := computeIngressStatusConditions([]operatorv1.OperatorCondition{}, deploy)
+		gotExpected := true
+		if len(actual) != len(expected) {
+			gotExpected = false
+		}
+		for _, conditionA := range actual {
+			foundMatchingCondition := false
+
+			for _, conditionB := range expected {
+				if conditionA.Type == conditionB.Type &&
+					conditionA.Status == conditionB.Status {
+					foundMatchingCondition = true
+					break
+				}
+			}
+
+			if !foundMatchingCondition {
+				gotExpected = false
+			}
+		}
+		if !gotExpected {
+			t.Fatalf("%q: expected %#v, got %#v", tc.description,
+				expected, actual)
+		}
+	}
+}
+
+func TestSetIngressStatusCondition(t *testing.T) {
+	testCases := []struct {
+		description   string
+		oldConditions []operatorv1.OperatorCondition
+		newCondition  *operatorv1.OperatorCondition
+		expected      []operatorv1.OperatorCondition
+	}{
+		{
+			description: "new condition",
+			newCondition: &operatorv1.OperatorCondition{
+				Type:   operatorv1.IngressControllerAvailableConditionType,
+				Status: operatorv1.ConditionTrue,
+			},
+			expected: []operatorv1.OperatorCondition{
+				{
+					Type:   operatorv1.IngressControllerAvailableConditionType,
+					Status: operatorv1.ConditionTrue,
+				},
+			},
+		},
+		{
+			description: "existing condition, unchanged",
+			oldConditions: []operatorv1.OperatorCondition{
+				{
+					Type:   operatorv1.IngressControllerAvailableConditionType,
+					Status: operatorv1.ConditionTrue,
+				},
+			},
+			newCondition: &operatorv1.OperatorCondition{
+				Type:   operatorv1.IngressControllerAvailableConditionType,
+				Status: operatorv1.ConditionTrue,
+			},
+			expected: []operatorv1.OperatorCondition{
+				{
+					Type:   operatorv1.IngressControllerAvailableConditionType,
+					Status: operatorv1.ConditionTrue,
+				},
+			},
+		},
+		{
+			description: "existing condition changed",
+			oldConditions: []operatorv1.OperatorCondition{
+				{
+					Type:   operatorv1.IngressControllerAvailableConditionType,
+					Status: operatorv1.ConditionFalse,
+				},
+			},
+			newCondition: &operatorv1.OperatorCondition{
+				Type:   operatorv1.IngressControllerAvailableConditionType,
+				Status: operatorv1.ConditionTrue,
+			},
+			expected: []operatorv1.OperatorCondition{
+				{
+					Type:   operatorv1.IngressControllerAvailableConditionType,
+					Status: operatorv1.ConditionTrue,
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		actual := setIngressStatusCondition(tc.oldConditions, tc.newCondition)
+		a := operatorv1.IngressControllerStatus{Conditions: actual}
+		b := operatorv1.IngressControllerStatus{Conditions: tc.expected}
+		if !ingressStatusesEqual(a, b) {
+			t.Fatalf("%q: expected %v, got %v", tc.description,
+				tc.expected, actual)
+		}
+	}
+}
+
+func TestIngressStatusesEqual(t *testing.T) {
+	testCases := []struct {
+		description string
+		expected    bool
+		a, b        operatorv1.IngressControllerStatus
+	}{
+		{
+			description: "nil and non-nil slices are equal",
+			expected:    true,
+			a: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{},
+			},
+		},
+		{
+			description: "empty slices should be equal",
+			expected:    true,
+			a: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{},
+			},
+			b: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{},
+			},
+		},
+		{
+			description: "condition LastTransitionTime should be ignored",
+			expected:    true,
+			a: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
+					{
+						Type:               operatorv1.IngressControllerAvailableConditionType,
+						Status:             operatorv1.ConditionTrue,
+						LastTransitionTime: metav1.Unix(0, 0),
+					},
+				},
+			},
+			b: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
+					{
+						Type:               operatorv1.IngressControllerAvailableConditionType,
+						Status:             operatorv1.ConditionTrue,
+						LastTransitionTime: metav1.Unix(1, 0),
+					},
+				},
+			},
+		},
+		{
+			description: "check condition reason differs",
+			expected:    false,
+			a: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
+					{
+						Type:   operatorv1.IngressControllerAvailableConditionType,
+						Status: operatorv1.ConditionFalse,
+						Reason: "foo",
+					},
+				},
+			},
+			b: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
+					{
+						Type:   operatorv1.IngressControllerAvailableConditionType,
+						Status: operatorv1.ConditionFalse,
+						Reason: "bar",
+					},
+				},
+			},
+		},
+		{
+			description: "condition status differs",
+			expected:    false,
+			a: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
+					{
+						Type:   operatorv1.IngressControllerAvailableConditionType,
+						Status: operatorv1.ConditionTrue,
+					},
+				},
+			},
+			b: operatorv1.IngressControllerStatus{
+				Conditions: []operatorv1.OperatorCondition{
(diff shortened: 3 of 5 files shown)


pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+	lbReadyCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.LoadBalancerReadyIngressConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if len(services) == 0 {
+		lbReadyCondition.Status = operatorv1.ConditionFalse
+		lbReadyCondition.Reason = "ServiceMissing"
+		lbReadyCondition.Message = "LoadBalancer Service does not exist."
+	} else {
+		for _, s := range services {
```
ironcladlou: The relationship between an ingresscontroller and any desired LB service is 1:1, not 1:M

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+			if s.Spec.Type == corev1.ServiceTypeClusterIP {
+				continue
+			}
+			// TODO: This status does not get updated when the loadbalancer service
+			//       status ip/hostname is updated. Is a service controller needed
+			//       that watches services and updates ingress status?
+			if s.Spec.Type == corev1.ServiceTypeLoadBalancer && len(s.Status.LoadBalancer.Ingress) == 0 ||
+				len(s.Status.LoadBalancer.Ingress[0].Hostname) == 0 {
+				lbReadyCondition.Status = operatorv1.ConditionTrue
+				lbReadyCondition.Reason = "LoadBalancerComplete"
```
ironcladlou: or "LoadBalancerProvisioned"?

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+			}
+		}
+	}
+	conditions = setIngressStatusCondition(conditions, lbReadyCondition)
+
+	dnsManagedCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.DNSManagedIngressConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if ic.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType &&
```
ironcladlou: Something's weird here — the logic which determines whether DNS is required is encapsulated in `controller_dns.go` and duplicating portions of it here seems likely to bite us later.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	conditions = setIngressStatusCondition(conditions, dnsManagedCondition)
+
+	// TODO: Implement DNSReady condition.
+
+	availableCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if ic != nil {
```
ironcladlou: Again, how can `ic` ever be nil?

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+
+// computeIngressStatusConditions computes the ingress controller's current state.
+func computeIngressStatusConditions(conditions []operatorv1.OperatorCondition, ic *operatorv1.IngressController,
+	deployment *appsv1.Deployment, infraConfig *configv1.Infrastructure, services []corev1.Service) []operatorv1.OperatorCondition {
+
+	failingCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeFailing,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if ic != nil {
```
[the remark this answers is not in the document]
danehans: Good point. I'll remove the `Failing` condition based on your feedback of initially implementing a single condition. Let me know if you have any suggestions for implementing an IngressController `Failing` condition.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,273 @@
...
+		failingCondition.Message = "IngressController does not exist."
+	}
+	conditions = setIngressStatusCondition(conditions, failingCondition)
+
+	progressingCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeProgressing,
+		Status: operatorv1.ConditionUnknown,
+	}
+	current := deployment.Status.AvailableReplicas
+	desired := *deployment.Spec.Replicas
```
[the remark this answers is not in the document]
danehans: Where does the function you reference live? I looked through the project repo and controller-runtime pkgs and could not find it.

pkg/operator/controller/ingress_status.go
danehans: My thinking was multiple services are related to an ingresscontroller depending on the endpoint publishing strategy. Therefore, multiple services may be passed to `computeIngressStatusConditions`. This would make it possible to create a condition based on the status of the internal (`ClusterIP`) service. Thoughts?

pkg/operator/controller/ingress_status.go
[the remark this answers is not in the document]
Miciah: We don't have such a function, but we probably should.  The following code should be refactored into a function (and perhaps we should add a named constant for the hard-coded default of 2): https://github.com/openshift/cluster-ingress-operator/blob/ed8733202d60b3406bf84daadc63efa90c8cb285/pkg/operator/controller/controller_router_deployment.go#L179-L182

If you prefer not to clutter this PR with the refactoring, I can make a small PR to do that work.

pkg/operator/controller/ingress_status.go
Miciah: The operator should only be created 0 or 1 LB service for each ingresscontroller.  If the administrator creates an LB service, I don't think we should worry about it.  So I'm, thinking `getIngressState` should be figuring out which LB service, if any, is associated with the ingresscontroller, and `computeIngressStatusConditions` should take exactly one service as input.

Miciah: Since the condition is named `LoadBalancerReady`, I don't think the internal service should be considered in determining its status.

pkg/operator/controller/ingress_status.go
Miciah: The DNS manager's interface could have a `IsManaged` function, which would return `false` for the no-op manager and `true` for the AWS manager.  Alternatively, we could just do a type assertion in `sync`:
```go
	_, isUnmanaged := r.DNSManager.(dns.*NoopManager)
	if isUnManaged {
		// ...
```

Miciah: Never mind my comment; I was confused.  We should refactor or re-use the logic in `desiredDNSRecords` (could even call it directly and then check whether the slice it returns is non-zero length), as Dan said.

ironcladlou:
@pravisankar @Miciah PTAL

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Deployment for IngressController: %s/%s", ic.Namespace, ic.Name)
+	}
+
+	return deployment, nil
+}
+
+// computeIngressStatusConditions computes the ingress controller's current state.
+func computeIngressStatusConditions(conditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment) []operatorv1.OperatorCondition {
+
```
Miciah: Extra blank line.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	availableCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if deployment.Status.AvailableReplicas > 0 {
+		availableCondition.Status = operatorv1.ConditionTrue
+	} else {
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "IngressControllerUnavailable"
+		availableCondition.Message = fmt.Sprintf("Deployment %q contains 0 AvailableReplicas", deployment.Name)
```
Miciah: Suggestion: "contains"→"has".

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if !found {
+		newConditions = append(newConditions, *condition)
+	}
+
+	return newConditions
+}
+
+// ingressStatusesEqual compares two OperatorCondition values.  Returns true
+// if the provided OperatorCondition values should be considered equal for the
```
Miciah: "OperatorCondition"→"IngressControllerStatus" twice.

pkg/operator/controller/ingress_status_test.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+func TestComputeIngressStatusConditions(t *testing.T) {
+	testCases := []struct {
+		description string
+		repl        int
+		available   bool
+	}{
+		{"0/2 deployment replicas available", 0, false},
+		{"1 deployment replicas available", 1, true},
```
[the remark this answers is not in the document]
Miciah: Actually, does the "/2" on the other test cases make sense? The test inputs here only have the number of available.

pkg/operator/controller/status.go
```diff
@@ -88,109 +86,93 @@ func (r *reconciler) syncOperatorStatus() error {
...
-	if len(unavailable) == 0 {
+	if numIngresses == len(ingressesAvailable) {
 		availableCondition.Status = configv1.ConditionTrue
 	} else {
 		availableCondition.Status = configv1.ConditionFalse
 		availableCondition.Reason = "IngressUnavailable"
-		availableCondition.Message = strings.Join(unavailable, "\n")
+		availableCondition.Message = fmt.Sprintf(
+			"Have %d ingress controllers, want %d",
+			len(ingressesAvailable), numIngresses)
```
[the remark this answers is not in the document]
danehans: Should `Available` only be true when availableReplicas for all ingresses are equal to desired replicas?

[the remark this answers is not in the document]
Miciah: It would set `Available` to true iff every ingresscontroller has at least one replica available, and I would set `Progressing` to true iff for any ingresscontroller, the number of available replicas is not equal to the number of desired replicas.  My reasoning is that if any replicas are available, then the ingresscontroller is serving routes, but until available replicas equals desired replicas, it is still progressing towards the desired state.

Another thought though: Will we add a `Progressing` condition to ingresscontroller? If so, we could leave this code for the clusteroperator status as is with a TODO, and once ingresscontrollers have a `Progressing` condition, the clusteroperator could reflect that: if all ingresscontrollers are available, then the clusteroperator is available; and if any ingresscontrollers are progressing, then the clusteroperator is progressing.  What do you think?

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// syncIngressStatus computes the current status of ic and therefrom
+// creates or updates the IngressController resource.
+func (r *reconciler) syncIngressStatus(ic *operatorv1.IngressController) error {
```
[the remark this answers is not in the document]
danehans: It's possible. `syncIngressStatus` is called for a single ingress during a reconciliation. If folded into `syncOperatorStatus()`, `syncIngressStatus` would be called for every ingresscontroller during each reconcile loop. Thoughts?

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+			return fmt.Errorf("failed to update IngressController %s/%s", ic.Namespace, ic.Name)
+		}
+	}
+	return nil
+}
+
+// getIngressState gets and returns the resources necessary to compute the
+// ingress controller's current state.
+func (r *reconciler) getIngressState(ic *operatorv1.IngressController) (*appsv1.Deployment, error) {
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: ic.Namespace, Name: ic.Name}, ic)
```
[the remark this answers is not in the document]
danehans: I thought it would be a good idea to make the call just in case something happened to the ic during reconciliation. Probably overkill, so I'll remove.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get IngressController: %s/%s", ic.Namespace, ic.Name)
+	}
+
+	ns, err := r.ManifestFactory.RouterNamespace()
```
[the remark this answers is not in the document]
danehans: OK. I'll try this func instead: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/names.go#L39

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Namespace for IngressController: %s/%s", ic.Namespace, ic.Name)
```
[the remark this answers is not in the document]
danehans: :-)

Understood, I was trying to convey that we were unable to get the namespace for the provider ingress controller. It will go away if the `RouterDeploymentName` func works as planned.

pkg/operator/controller/ingress_status_test.go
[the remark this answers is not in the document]
danehans: I updated the test with desired and available replicas.

pkg/operator/controller/ingress_status.go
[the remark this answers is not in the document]
Miciah: BTW, here's what I have in mind—the PR has one approach, and its description links to an alternative approach: https://github.com/openshift/cluster-ingress-operator/pull/180

pkg/operator/controller/status.go
```diff
@@ -88,109 +86,93 @@ func (r *reconciler) syncOperatorStatus() error {
...
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace, ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
 	failingCondition := &configv1.ClusterOperatorStatusCondition{
 		Type:   configv1.OperatorFailing,
 		Status: configv1.ConditionUnknown,
 	}
 	if ns == nil {
 		failingCondition.Status = configv1.ConditionTrue
 		failingCondition.Reason = "NoNamespace"
-		failingCondition.Message = "router namespace does not exist"
+		failingCondition.Message = "ingress controller namespace does not exist"
```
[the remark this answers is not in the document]
danehans: I changed to `ingress operator`.

pkg/operator/controller/status.go
```diff
@@ -88,109 +86,93 @@ func (r *reconciler) syncOperatorStatus() error {
...
 		Status: configv1.ConditionUnknown,
 	}
 	numIngresses := len(ingresses)
-	numDeployments := len(deployments)
-	if numIngresses == numDeployments {
+	ingressesAvailable := map[string]bool{}
+	for _, ing := range ingresses {
+		for _, c := range ing.Status.Conditions {
+			if c.Type == operatorv1.IngressControllerAvailableConditionType && c.Status == operatorv1.ConditionTrue {
+				ingressesAvailable[ing.Name] = true
```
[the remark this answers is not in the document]
danehans: Yeah, I over complicated it. Thank for the guidance.

pkg/operator/controller/ingress_status.go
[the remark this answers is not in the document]
Miciah: Not `syncOperatorStatus`, `syncClusterIngressStatus`: https://github.com/openshift/cluster-ingress-operator/blob/e49c483cea90d0360ce653afdc8104e145d67123/pkg/operator/controller/controller.go#L457-L480

pkg/operator/controller/ingress_status.go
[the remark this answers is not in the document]
danehans: I switched to using `RouterDeploymentName` and it did the trick and allowed me to remove the error message.

pkg/operator/controller/status.go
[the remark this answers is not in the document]
danehans: > It would set Available to true iff every ingresscontroller has at least one replica available, and I would set Progressing to true iff for any ingresscontroller, the number of available replicas is not equal to the number of desired replicas.

I believe this is best represented as an `IngressController` status condition. We can then expose the condition through the `ClusterOperator` `Progressing` status, allowing us to decouple `Deployment` logic from `ClusterOperator` status.

> Will we add a Progressing condition to ingresscontroller?

That's my understanding. After talking with @ironcladlou, we thought it made sense to get one status condition right at a time. I'll make a TODO as a reminder of what needs to be done when the `Progressing` `IngressController status condition is implemented.

pkg/operator/controller/ingress_status.go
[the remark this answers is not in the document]
danehans: @Miciah Thanks for sharing this with me. I did not see `syncClusterIngressStatus`. Let me fold the logic from `syncIngressStatus` into it and update my PR.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,104 @@
...
+// updates the syncIngressControllerStatus resource.
+func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment, ic *operatorv1.IngressController) error {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("ingress controller deployment %s/%s has invalid spec.selector: %v", deployment.Namespace, deployment.Name, err)
+	}
+
+	oldStatus := ic.Status.DeepCopy()
+	oldStatus.AvailableReplicas = deployment.Status.AvailableReplicas
+	oldStatus.Selector = selector.String()
```
Miciah: I think you want to assign `ic.Status.AvailableReplicas` and `ic.Status.Selector` here (and then you can delete the assignments to those fields inside the `if` below).

danehans: I have updated the logic. I now create a `DeepCopy`, assign the status fields, compare the copy and the original ingress for equality. If the two are not equal, I update the status through the api using the copy.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,102 @@
...
+func computeIngressStatusConditions(conditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment) []operatorv1.OperatorCondition {
+	availableCondition := &operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	if deployment.Status.AvailableReplicas > 0 {
+		availableCondition.Status = operatorv1.ConditionTrue
+	} else {
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "IngressControllerUnavailable"
```
ironcladlou: Should this be in terms of the deployment? e.g. `DeploymentUnavailable` and `%d deployment replicas available, want %d`?

pkg/operator/controller/controller.go
```diff
@@ -362,8 +361,8 @@ func (r *reconciler) ensureClusterIngress(ci *operatorv1.IngressController, dnsC
 			errs = append(errs, fmt.Errorf("failed to integrate metrics with openshift-monitoring for clusteringress %s: %v", ci.Name, err))
 		}
 
-		if err := r.syncClusterIngressStatus(deployment, ci); err != nil {
-			errs = append(errs, fmt.Errorf("failed to sync status of clusteringress %s/%s: %v", deployment.Namespace, deployment.Name, err))
+		if err := r.syncIngressControllerStatus(deployment, ci); err != nil {
+			errs = append(errs, fmt.Errorf("failed to sync status of ingresscontroller %s/%s: %v", ci.Namespace, ci.Name, err))
```
ironcladlou: Nit (and I don't know everyone agrees with me): I would omit ingress controller details from the error message (as the caller already knows the context)

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("ingress controller deployment %s/%s has invalid spec.selector: %v", deployment.Namespace, deployment.Name, err)
+	}
+
+	updated := ic.DeepCopy()
+	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
+	updated.Status.Selector = selector.String()
+	updated.Status.Conditions = computeIngressStatusConditions(updated.Status.Conditions, deployment)
+	if !ingressStatusesEqual(updated.Status, ic.Status) {
+		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+			return fmt.Errorf("failed to update the status of IngressController %s/%s", ic.Namespace, ic.Name)
```
ironcladlou: As above, I would omit the ingress controller reference

ironcladlou: Also, the actual error is eaten here

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,102 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// syncIngressControllerStatus computes the current status of ic and therefrom
+// updates the syncIngressControllerStatus resource.
+func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment, ic *operatorv1.IngressController) error {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("ingress controller deployment %s/%s has invalid spec.selector: %v", deployment.Namespace, deployment.Name, err)
```
ironcladlou: Random thought for later: we own that field, so if it's invalid, this might be a pretty severe problem unless it's an outside mutation we correct later during a resync

pkg/operator/controller/ingress_status_test.go
```diff
@@ -0,0 +1,262 @@
...
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestComputeIngressStatusConditions(t *testing.T) {
+	testCases := []struct {
+		description     string
+		availRepl, repl int32
```
ironcladlou: You could encode the expected condition into this as well, so your cases would be like

```go
{"0/2 deployment replicas available", 0, 2, OperatorStatusTypeAvailable, ConditionFalse},
{"1/2 deployment replicas available", 1, 2, OperatorStatusTypeAvailable, ConditionTrue},
{"2/2 deployment replicas available", 2, 2, OperatorStatusTypeAvailable, ConditionTrue},
```

And, later, could be expanded to encompass asserting a whole set of conditions.

Perhaps more importantly, another advantage is that you avoid explicitly re-implementing the condition algorithm inside the test code (e.g. as written, L35 would have to eternally mirror the internals of `computeIngressStatusConditions`)

pkg/operator/controller/ingress_status.go
danehans: Good point. Thanks for the insight. When you say "correct later during a resync", it's my understanding that `ingresscontroller.status.selector` would be updated to reflect the mutated  `deployment.Spec.Selector` as apposed to the Operator ensuring `deployment.Spec.Selector` reflects the value of `ingresscontroller.status.selector`. Is my understanding correct?

ironcladlou: Sounds right — however, I believe such a change to the deployment would cause the operator to detect a diff and overwrite the mutated deployment selector field, which would then propagate back to the ingresscontroller status. Could do a quick test to confirm

pkg/operator/controller/ingress_status_test.go
danehans: Thanks for the input. I have updated the test cases as you requested.

pkg/operator/controller/ingress_status.go
danehans: >  I believe such a change to the deployment would cause the operator to detect a diff and overwrite the mutated deployment selector field

@ironcladlou that's not what I observe. If I mutate the `Deployment` selector, a reconciliation is triggered and the mutated selector is observed in the associated `IngressController`.

Miciah: `deploymentConfigChanged` does not compare the deployments' pod selectors, so the operator will not revert modifications to the same, and `syncClusterIngressStatus` will reflect whatever pod selector is in the deployment to the ingresscontroller's status.

danehans: @ironcladlou @Miciah you can review the details of my deployment selector testing [here](https://gist.githubusercontent.com/danehans/09315e07537eca731cd79bf44d3535b2/raw/2e6131d4ac784902f8c365e2dfc15b0a9785456d/gistfile1.txt) that confirm my understanding.

pkg/operator/controller/ingress_status.go
```diff
@@ -0,0 +1,102 @@
...
+		Status: operatorv1.ConditionUnknown,
+	}
+	if deployment.Status.AvailableReplicas > 0 {
+		availableCondition.Status = operatorv1.ConditionTrue
+	} else {
+		availableCondition.Status = operatorv1.ConditionFalse
+		availableCondition.Reason = "DeploymentUnavailable"
+		availableCondition.Message = fmt.Sprintf(
+			"%d Deployment replicas available, want %d",
+			deployment.Status.AvailableReplicas, deployment.Status.Replicas)
```
Miciah: This would be an appropriate message for the `Progressing` condition, but for `Available`, it is false simply because no deployment replicas are available, and the number wanted is irrelevant to the condition.  We can leave this as is though till we add the `Progressing` condition (this is similar to the situation with `computeOperatorStatusConditions`).

pkg/operator/controller/status.go
```diff
@@ -89,102 +87,90 @@ func (r *reconciler) syncOperatorStatus() error {
...
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace, ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
 	failingCondition := &configv1.ClusterOperatorStatusCondition{
 		Type:   configv1.OperatorFailing,
 		Status: configv1.ConditionUnknown,
 	}
 	if ns == nil {
 		failingCondition.Status = configv1.ConditionTrue
 		failingCondition.Reason = "NoNamespace"
-		failingCondition.Message = "router namespace does not exist"
+		failingCondition.Message = "ingress operator namespace does not exist"
```
Miciah: It's the operand namespace (`openshift-ingress`), not the operator namespace (`openshift-ingress-operator`).

pkg/operator/controller/status.go
```diff
@@ -89,102 +87,90 @@ func (r *reconciler) syncOperatorStatus() error {
...
+	conditions = setOperatorStatusCondition(conditions, availableCondition)
 
 	return conditions
 }
 
-// setStatusCondition returns the result of setting the specified condition in
-// the given slice of conditions.
-func setStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition, condition *configv1.ClusterOperatorStatusCondition) []configv1.ClusterOperatorStatusCondition {
+// setOperatorStatusCondition returns an Operator status condition as a
+// result of setting the specified condition in the given slice of conditions.
```
Miciah: It returns a slice of operator status conditions.

pkg/operator/controller/status_test.go
```diff
@@ -25,49 +24,37 @@ func TestComputeStatusConditions(t *testing.T) {
...
 	}{
-		{"no namespace", testInputs{false, 0, 0, 0}, testOutputs{true, false, true}},
-		{"no ingresses, no routers", testInputs{true, 0, 0, 0}, testOutputs{false, false, true}},
-		{"scaling up", testInputs{true, 1, 0, 0}, testOutputs{false, true, false}},
-		{"scaling down", testInputs{true, 0, 1, 0}, testOutputs{false, true, true}},
-		{"0/2 ingresses available", testInputs{true, 2, 0, 2}, testOutputs{false, false, false}},
-		{"1/2 ingresses available", testInputs{true, 2, 1, 1}, testOutputs{false, false, false}},
-		{"2/2 ingresses available", testInputs{true, 2, 2, 0}, testOutputs{false, false, true}},
+		{"no namespace", testInputs{false, 0, 0}, testOutputs{true, false, true}},
+		{"no clusteroperator, no ingresscontroller deployments", testInputs{true, 0, 0}, testOutputs{false, false, true}},
```
Miciah: "no clusteroperator" is not a test input.  Looks like this is simply "no ingresscontrollers".

pkg/operator/controller/ingress_status.go
danehans: @Miciah good point about the IngressController `Available` message.

> (this is similar to the situation with computeOperatorStatusConditions).

I believe a slight difference exists with the Operator `Available` condition. With the IngressController `Available` condition, if 0 Deployments exist then status is `False`. For Operator, each IngressController must have at least 1 Deployment or status is `False`.

Miciah: Right, I mean the situation is similar in that we are conflating `Available` and `Progressing`.  Here, we are only reporting `Available`, but we are reporting a message that is more appropriate for `Progressing` (which does not exist yet).  In the operator status, we are reporting both `Available` and `Progressing`, but we report the same status and message, which are both more appropriate for `Progressing`.

Once we add a `Progressing` condition to `IngressController`, we can sort out the ingresscontroller conditions here, and then it will be possible to sort out the operator conditions too.  I just want to make sure we're on the same page, and I think we are.

pkg/operator/controller/ingress_status.go
Miciah: I expect the deployment controller deleted the old pods (with the `ingress.operator.openshift.io/ingress-controller-deployment=default` label) and create new pods (with the `foo=bar` label) when you modified the deployment, right? And the next time the operator updates the deployment, it will revert the changes, so the pods will all get recreated again.  In the meantime, the anti-affinity policy will not work properly, but I think everything else will be fine.

So if I understand correctly, if an administrator changes the labels, anti-affinity breaks and pods get recreated, but nothing will go horribly broken—which is pretty good considering that outside mutations are not something we recommend or have made efforts to accommodate!

I notice you changed both the deployment's label selector as well as the pod template's labels to match.  If you hadn't done that, the breakage would be more severe.  On the other hand, if the administrator modifies the deployment that _we_ manage, _and_ modifies it to an invalid state, the administrator just asking for problems.

Miciah:
level=info msg="Waiting up to 30m0s for the cluster at https://api.ci-op-cfp602yi-2945d.origin-ci-int-aws.dev.rhcloud.com:6443 to initialize..."
    level=fatal msg="failed to initialize the cluster: Working towards 0.0.1-2019-03-26-194721: 98% complete: timed out waiting for the condition"

/test e2e-aws

PR #175, opened 2019-03-18
Bug 1683765: Ensure IngressController domain is unique

danehans:
Currently, the operator does not prevent duplicate domains among different IngressControllers.

This PR adds a domain conflict check. If an `IngressController` is created with a domain that is equal to an existing `IngressController`, the operator will log the conflict and dependent resources will not be created. When https://github.com/openshift/cluster-ingress-operator/pull/174 merges, the log message and checks for `len(status.domain) == 0` checkes should be transitioned to status condition checks.

### DIFF
--- a/pkg/operator/controller/certificate/controller.go
+++ b/pkg/operator/controller/certificate/controller.go
@@ -76,6 +76,8 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		} else {
 			errs = append(errs, fmt.Errorf("failed to get clusteringress: %v", err))
 		}
+	} else if !controller.IsStatusDomainSet(ingress) {
+		log.Info("ingresscontroller domain not set; reconciliation will be skipped", "request", request)
 	} else {
 		deployment := &appsv1.Deployment{}
 		err = r.client.Get(context.TODO(), controller.RouterDeploymentName(ingress), deployment)
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -143,19 +143,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 			if err := r.enforceEffectiveIngressDomain(ingress, ingressConfig); err != nil {
 				errs = append(errs, fmt.Errorf("failed to enforce the effective ingress domain for clusteringress %s: %v", ingress.Name, err))
-			} else if err := r.enforceEffectiveEndpointPublishingStrategy(ingress, infraConfig); err != nil {
-				errs = append(errs, fmt.Errorf("failed to enforce the effective HA configuration for clusteringress %s: %v", ingress.Name, err))
-			} else if ingress.DeletionTimestamp != nil {
-				// Handle deletion.
-				if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
-					errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
-				}
-			} else if err := r.enforceIngressFinalizer(ingress); err != nil {
-				errs = append(errs, fmt.Errorf("failed to enforce ingress finalizer %s/%s: %v", ingress.Namespace, ingress.Name, err))
-			} else {
-				// Handle everything else.
-				if err := r.ensureClusterIngress(ingress, dnsConfig, infraConfig); err != nil {
-					errs = append(errs, fmt.Errorf("failed to ensure clusteringress: %v", err))
+			} else if IsStatusDomainSet(ingress) {
+				if err := r.enforceEffectiveEndpointPublishingStrategy(ingress, infraConfig); err != nil {
+					errs = append(errs, fmt.Errorf("failed to enforce the effective HA configuration for clusteringress %s: %v", ingress.Name, err))
+				} else if ingress.DeletionTimestamp != nil {
+					// Handle deletion.
+					if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
+						errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
+					}
+				} else if err := r.enforceIngressFinalizer(ingress); err != nil {
+					errs = append(errs, fmt.Errorf("failed to enforce ingress finalizer %s/%s: %v", ingress.Namespace, ingress.Name, err))
+				} else {
+					// Handle everything else.
+					if err := r.ensureClusterIngress(ingress, dnsConfig, infraConfig); err != nil {
+						errs = append(errs, fmt.Errorf("failed to ensure clusteringress: %v", err))
+					}
 				}
 			}
 		}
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/controller.go
```diff
@@ -129,19 +129,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+					if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
+						errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
+					}
+				} else if err := r.enforceIngressFinalizer(ingress); err != nil {
+					errs = append(errs, fmt.Errorf("failed to enforce ingress finalizer %s/%s: %v", ingress.Namespace, ingress.Name, err))
+				} else {
+					// Handle everything else.
+					if err := r.ensureClusterIngress(ingress, dnsConfig, infraConfig); err != nil {
+						errs = append(errs, fmt.Errorf("failed to ensure clusteringress: %v", err))
+					}
```
ironcladlou: @Miciah PTAL, I think you had some ideas about how to tame this control flow — it's getting out of hand IMO

pkg/operator/controller/controller.go
```diff
@@ -182,6 +194,24 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *operatorv1.IngressControl
 	return nil
 }
 
+// ensureUniqueDomain ensures spec.domain from ic is unique across all ingress controllers
+// in the same namespace.
+func (r *reconciler) ensureUniqueDomain(ic *operatorv1.IngressController) (bool, error) {
```
ironcladlou: Function name doesn't seem to quite match the behavior — given the signature I'd have expected something like `isDomainUnique` or something

pkg/operator/controller/controller.go
```diff
@@ -182,6 +194,24 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *operatorv1.IngressControl
...
+	ingresses := operatorv1.IngressControllerList{}
+	if err := r.Client.List(context.TODO(), &client.ListOptions{Namespace: r.Namespace}, &ingresses); err != nil {
+		return false, fmt.Errorf("failed to list clusteringresses: %v", err)
```
ironcladlou: s/clusteringresses/ingresscontrollers

pkg/operator/controller/controller.go
```diff
@@ -168,11 +170,21 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *operatorv1.IngressControl
...
 	switch {
 	case len(ci.Spec.Domain) > 0:
+		unique, err := r.ensureUniqueDomain(updated)
+		if err != nil {
+			return err
+		}
+		if !unique {
+			// TODO: Create DNSConflict status condition when the following PR merges:
+			//       https://github.com/openshift/cluster-ingress-operator/pull/174
+			log.Info("domain conflict, not setting domain in status for clusteringress: ", "namespace", updated.Namespace, "name", updated.Name)
```
ironcladlou: Any concern about log spam? What about an event in this case prior to our use of status?

pkg/operator/controller/certificate/controller.go
```diff
@@ -79,7 +79,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		} else {
 			errs = append(errs, fmt.Errorf("failed to get clusteringress: %v", err))
 		}
-	} else {
+	} else if len(ingress.Status.Domain) > 0 {
```
ironcladlou: something about this feels off... should we have a higher level concept of accepted/rejected abstracted behind a function call or something?

pkg/operator/controller/controller.go
danehans: Good point. I'll update.

pkg/operator/controller/controller.go
danehans: Since the ingresscontroller completes reconciliation, I only see one log message.

pkg/operator/controller/certificate/controller.go
danehans: Will do.

pkg/operator/controller/controller.go
danehans: ack

pkg/operator/controller/controller.go
Miciah: At one time I suggested using a switch, but that in fact would not work (we cannot do `case x := y; y != nil:`).  I think what we need to do is factor out a couple functions, such as `getClusterConfig` to get the cluster config and `reconcileIngressController` to ensure defaults and then create or delete resources as needed.  I can make a separate PR with what I have in mind if that would help.

pkg/operator/controller/controller.go
```diff
@@ -168,11 +170,21 @@ func (r *reconciler) enforceEffectiveIngressDomain(ci *operatorv1.IngressControl
 	updated := ci.DeepCopy()
 	switch {
 	case len(ci.Spec.Domain) > 0:
+		unique, err := r.isDomainUnique(updated)
```
Miciah: `IsDomainUnique` should take a parameter for the domain instead of for the ingresscontroller.

danehans:
> Check if the domain is unique; if not, return an error.

@Miciah it was my understanding that @ironcladlou wanted reconciliation to stop if a domain conflict exists. If `ensureUniqueDomain` only returns an error, then on reconciliation continues to loop for that resource and I get the following log spam.

```
2019-03-21T10:53:16.656-0700	ERROR	operator.init.kubebuilder.controller	controller/controller.go:217	Reconciler error	{"controller": "operator-controller", "request": "openshift-ingress-operator/test0", "error": "failed to enforce the effective ingress domain for clusteringress test0: domain \"apps.danehans.devcluster.openshift.com\" conflicts with ingresscontroller: openshift-ingress-operator/default", "errorCauses": [{"error": "failed to enforce the effective ingress domain for clusteringress test0: domain \"apps.danehans.devcluster.openshift.com\" conflicts with ingresscontroller: openshift-ingress-operator/default"}]}
```

Miciah:
> @Miciah it was my understanding that @ironcladlou wanted reconciliation to stop if a domain conflict exists. If `ensureUniqueDomain` only returns an error, then on reconciliation continues to loop for that resource

Sorry, I was trying to describe the order in which the steps should be performed, and I got the particulars of one of those steps wrong.  You're right, `isDomainUnique` should return both a Boolean value and an error value, as you had before.  Sorry for the confusion!

pkg/operator/controller/controller.go
```diff
@@ -172,30 +174,61 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		updated.Status.Domain = ingressConfig.Spec.Domain
+		domain = ingressConfig.Spec.Domain
 	}
-	// TODO Validate and check for conflicting claims.
+	unique, err := r.isDomainUnique(domain)
+	if err != nil {
+		return err
+	}
+	if !unique {
+		log.Info("domain not unique, not setting status domain for IngressController", "namespace", ic.Namespace, "name", ic.Name)
```
Miciah: You can log and then `return` here—we already know that `ic.Status.Domain` is empty (it's first thing we checked in this method), so at this point we can skip the update below.

Miciah:
This test failure:

    --- FAIL: TestHostNetworkEndpointPublishingStrategy (61.02s)
    	operator_test.go:690: failed to get deployment: timed out waiting for the condition
    	operator_test.go:694: failed to delete the ingresscontroller: timed out waiting for the condition

is caused by the following:

    2019-03-25T16:08:20.487Z        DEBUG   operator.init.kubebuilder.controller    controller/controller.go:236    Successfully Reconciled {"controller": "certificate-controller", "request": "openshift-ingress-operator/hostnetwork"}
    2019-03-25T16:08:20.511Z        INFO    operator.controller     controller/controller.go:224    domain conflicts with existing IngressController        {"domain": "apps.ci-op-w12196xx-43abb.origin-ci-int-aws.dev.rhcloud.com", "namespace": "openshift-ingress-operator", "name": "default"}
    2019-03-25T16:08:20.512Z        INFO    operator.controller     controller/controller.go:197    domain not unique, not setting status domain for IngressController      {"namespace": "openshift-ingress-operator", "name": "hostnetwork"}

[from the operator logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/175/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/701/artifacts/e2e-aws-operator/pods/openshift-ingress-operator_ingress-operator-69c888bd6c-6k9v6_ingress-operator.log.gz).  So it looks like `TestHostNetworkEndpointPublishingStrategy` needs some `Domain` added to the ingresscontroller spec at https://github.com/openshift/cluster-ingress-operator/blob/bbbd0c262bd4b782435abda4169f2df650e52a6b/test/e2e/operator_test.go#L658.

danehans:
@Miciah thanks for sharing. It's weird that I don't see this error running e2e locally.

danehans:
e2e-aws-operator is now passing. Thanks for the heads-up @Miciah.

pkg/operator/controller/controller.go
```diff
@@ -172,30 +174,62 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		return nil
+	} else {
+		updated.Status.Domain = domain
 	}
```
Miciah: The `else` block can be unwrapped:
```suggestion
	if !unique {
		log.Info("domain not unique, not setting status domain for IngressController", "namespace", ic.Namespace, "name", ic.Name)
This conversation was marked as resolved by danehans
 		return nil
	}

	updated.Status.Domain = domain
```

PR #179, opened 2019-03-20
Fix load-balancer finalization

Miciah:
When finalizing a load-balancer service, check whether it has been provisioned, and if not, skip cleaning up its DNS records.

Before this commit, if an ingresscontroller was created with the `LoadBalancerService` endpoint publishing strategy and subsequently deleted before the load balancer had been provisioned, the finalization logic returned an error, which caused the controller to retry and fail to finalize the service repeatedly.

* `pkg/operator/controller/controller_dns.go` (`ensureDNS`): Check whether the load balancer has been provisioned.  Return an error if not.  Pass the host name from the load balancer to `desiredDNSRecords`.
(`desiredDNSRecords`): Take a host name instead of a load-balancer service.
* `pkg/operator/controller/controller_lb.go` (`finalizeLoadBalancerService`): Check whether the load balancer has been provisioned.  Skip DNS clean-up if not.  Pass the host name from the load balancer to `desiredDNSRecords`.

### DIFF
--- a/pkg/operator/controller/controller_dns.go
+++ b/pkg/operator/controller/controller_dns.go
@@ -14,7 +14,14 @@ import (
 // ensureDNS will create DNS records for the given LB service. If service is
 // nil, nothing is done.
 func (r *reconciler) ensureDNS(ci *operatorv1.IngressController, service *corev1.Service, dnsConfig *configv1.DNS) error {
-	dnsRecords, err := desiredDNSRecords(ci, service, dnsConfig)
+	// If no load balancer has been provisioned, we can't do anything with the
+	// configured DNS zones.
+	ingress := service.Status.LoadBalancer.Ingress
+	if len(ingress) == 0 || len(ingress[0].Hostname) == 0 {
+		return fmt.Errorf("no load balancer is assigned to service %s/%s", service.Namespace, service.Name)
+	}
+
+	dnsRecords, err := desiredDNSRecords(ci, ingress[0].Hostname, dnsConfig)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/controller_lb.go
```diff
@@ -118,20 +118,31 @@ func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressControlle
...
-		return err
-	}
-	dnsErrors := []error{}
-	for _, record := range records {
-		if err := r.DNSManager.Delete(record); err != nil {
-			dnsErrors = append(dnsErrors, fmt.Errorf("failed to delete DNS record %v for ingress %s/%s: %v", record, ci.Namespace, ci.Name, err))
-		} else {
-			log.Info("deleted DNS record for ingress", "namespace", ci.Namespace, "name", ci.Name, "record", record)
+	// If no load balancer has been provisioned, we cannot have published
+	// DNS records for it.
```
ironcladlou: Nit: this isn't necessarily true (hence the TODO): if no load balancer is _currently_ reported on status_ it _probably_ means we haven't published DNS

Miciah: Thanks! Updated comment.

pkg/operator/controller/controller_dns.go
```diff
@@ -14,7 +14,14 @@ import (
 // ensureDNS will create DNS records for the given LB service. If service is
 // nil, nothing is done.
 func (r *reconciler) ensureDNS(ci *operatorv1.IngressController, service *corev1.Service, dnsConfig *configv1.DNS) error {
-	dnsRecords, err := desiredDNSRecords(ci, service, dnsConfig)
+	// If no load balancer has been provisioned, we can't do anything with the
+	// configured DNS zones.
+	ingress := service.Status.LoadBalancer.Ingress
+	if len(ingress) == 0 || len(ingress[0].Hostname) == 0 {
```
danehans: Can an ingress IP sneak by causing `desiredDNSRecords` to create an empty record? I know we're only using `Hostname` for now, but I wonder if this can lead to a bug in the future. Does it make sense for the condition to only be `len(ingress[0].Hostname) == 0`?

PR #181, opened 2019-03-22
Add instructions for building remotely on a cluster

Miciah:
Provide a buildconfig, shell scripts, Makefile targets, and instructions for building the operator remotely on a cluster using the buildconfig.

* `HACKING.md`: Add instructions for building using a buildconfig.
* `Makefile`: Add `buildconfig` and `cluster-build` targets.
* `hack/buildconfig.yaml`: New file.
* `hack/create-buildconfig.sh`: New file.
* `hack/start-build.sh`: New file.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -15,7 +15,9 @@ $ make build
 * An [OpenShift cluster](https://github.com/openshift/installer)
 * An admin-scoped `KUBECONFIG` for the cluster.
 
-Uninstall the existing operator and all its managed components:
+#### Building Locally
+
+To build the operator on your local machine and deploy it to the cluster, first uninstall the existing operator and all its managed components:
 
 ```
 $ hack/uninstall.sh
@@ -31,6 +33,41 @@ Follow the instructions to install the operator, e.g.:
 
 ```
 $ oc apply -f /tmp/manifests/path
+```
+
+#### Building Remotely on the Cluster
+
+To build the operator on the remote cluster, first create a buildconfig on the cluster:
+
+```
+$ make buildconfig
+```
+
+The above command will create a buildconfig using the current branch and the URL for the default push remote.  You can also specify an explicit branch or repository URL:
+
+```
+$ make buildconfig GIT_BRANCH=<branch> \
+     GIT_URL=https://github.com/<username>/cluster-ingress-operator.git
+```
+
+Note: If a buildconfig already exists from an earlier `make buildconfig` command, `make buildconfig` will update the existing buildconfig.
+
+Next, start a build from this buildconfig:
+
+```
+$ make cluster-build
+```
+
+Alternatively, if you want to see the logs during the build, specify the `V` flag:
+
+```
+$ make cluster-build V=1
+```
+
+Use the `DEPLOY` flag to start a build and then patch the operator to use the newly built image:
+
+```
+$ make cluster-build DEPLOY=1
 ```
 
 ## Tests
--- a/hack/buildconfig.yaml
+++ b/hack/buildconfig.yaml
@@ -1,0 +1,36 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: ingress-operator-build
+  annotations:
+    description: "Template for building the ingress operator."
+objects:
+- apiVersion: v1
+  kind: ImageStream
+  metadata:
+    name: ingress-operator
+- apiVersion: v1
+  kind: BuildConfig
+  metadata:
+    name: ingress-operator
+  spec:
+    output:
+      to:
+        kind: ImageStreamTag
+        name: ingress-operator:latest
+    source:
+      git:
+        uri: ${GIT_URL}
+        ref: ${GIT_BRANCH}
+      type: Git
+    strategy:
+      dockerStrategy:
+        dockerfilePath: Dockerfile
+      type: Docker
+parameters:
+- description: 'URL for the ingress operator Git repository'
+  name: GIT_URL
+  value: https://github.com/openshift/cluster-ingress-operator
+- description: 'Git branch'
+  name: GIT_BRANCH
+  value: master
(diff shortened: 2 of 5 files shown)


HACKING.md
```diff
@@ -33,6 +35,37 @@ Follow the instructions to install the operator, e.g.:
...
+#### Building Remotely on the Cluster
+
+To build the operator on the remote cluster, first create a buildconfig.  Use the following command, with the appropriate strings substituted for `<username>` and `<branch>`:
+
+```
+$ oc process -f hack/buildconfig.yaml \
+     -p 'GIT_URL=https://github.com/<username>/cluster-ingress-operator.git' \
+     -p 'GIT_BRANCH=<branch>' \
+      | oc -n openshift-ingress-operator create -f -
+```
```
ironcladlou: Could a script discover all this input using `git` so the UX can be something like:

    $ make build-cluster

HACKING.md
```diff
@@ -33,6 +35,37 @@ Follow the instructions to install the operator, e.g.:
...
+
+Next, start a build from this buildconfig:
+
+```
+$ oc -n openshift-ingress-operator start-build cluster-ingress-operator --follow
```
ironcladlou: The `make build-cluster` process could start this as well

HACKING.md
```diff
@@ -33,6 +35,37 @@ Follow the instructions to install the operator, e.g.:
...
+oc -n openshift-ingress-operator patch deploy/ingress-operator --type=json \
+   --patch='
+- op: replace
+  path: /spec/template/spec/containers/0/image
+  value: image-registry.openshift-image-registry.svc:5000/openshift-ingress-operator/cluster-ingress-operator:latest
+- op: replace
+  path: /spec/template/spec/containers/0/imagePullPolicy
+  value: Always
+'
+```
```
ironcladlou: The theoretical `make build-cluster` could handle this, too (if, say, `DEPLOY=true` or something).

HACKING.md
Miciah: We only need to create the buildconfig once, and then we can start arbitrarily many builds using it, so we can either have separate make targets or do some extra work to check whether the buildconfig exists first.  Would separate `make buildconfig-cluster` and `make build-cluster` targets be all right?

HACKING.md
Miciah: Yeah, it could.  I cannot find a good way to find the URL, but `hub browse -u <branch>` can figure it out, so I should be able to too.

Miciah: Done.

HACKING.md
Miciah: Done, using two targets: `buildconfig` and `cluster-build`.

HACKING.md
Miciah: Done (`make cluster-build DEPLOY=1`).

hack/create-buildconfig.sh
```diff
@@ -0,0 +1,29 @@
...
+then
+    # Convert git@host:user/repo to https://host/user/repo.
+    GIT_URL="${GIT_URL/://}"
+    GIT_URL="https://${GIT_URL#git@}"
+fi
+
+oc process -f "$(dirname "$0")/buildconfig.yaml" \
+   -p "GIT_URL=$GIT_URL" \
+   -p "GIT_BRANCH=$GIT_BRANCH" \
+    | oc -n openshift-ingress-operator create -f -
```
ironcladlou: Does allow one to re-deploy changes to `buildconfig.yaml`? Is `oc apply` necessary for that?

hack/buildconfig.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: v1
```
ironcladlou: should cluster-ingress-operator be just ingress-operator throughout?

hack/create-buildconfig.sh
Miciah: Using `oc create`, the user would have to delete the buildconfig before recreating it.  Using `oc apply` is a great idea—fixed, and thanks!

hack/buildconfig.yaml
Miciah: The repository name is "cluster-ingress-operator", and the image name is "openshift/origin-cluster-ingress-operator", so I think it makes sense to keep "cluster-" for the imagestream, and it makes sense for the buildconfig name to match.  Are we trying to move away from using the "cluster-" prefix in general?

ironcladlou: Long ago we started referring to it as "ingress-operator" (e.g. openshift-ingress-operator) and I think the only reason we didn't rename the git repo and image is because it was too much trouble. So, I don't know.

hack/start-build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -euo pipefail
+
+oc -n openshift-ingress-operator start-build ingress-operator \
+   ${V+--follow} --wait
+
+if [[ -n "${DEPLOY+1}" ]]
+then
+    oc -n openshift-ingress-operator patch deploy/ingress-operator \
+       --type=strategic --patch='
```
ironcladlou: Will the CVO allow this? If not, should we be updating the image via the CVO somehow? Would love to get away from disabling the CVO during development

Miciah: CVO will revert the change when it syncs.  Usually I disable CVO if I need it not to revert the image, but I can look into [patching the clusterversion](https://github.com/openshift/cluster-version-operator/blob/3956b767d634e7eab7dc1c322e7393330bd2d641/docs/dev/clusterversion.md#setting-objects-unmanaged) to override the image.  Should that be a separate PR?

ironcladlou: I think disabling CVO is standard practice for us at this point when it comes to any of the hack tooling — was just curious if you had thought more about it. It would be nice to, at some near point in the future, figure out how to work with the CVO rather than against it in these cases.

Miciah:
release "release-latest" failed: pod release-latest was already deleted

/test images

PR #182, opened 2019-03-22
deploymentConfigChanged: Fix volume comparison

Miciah:
Fix several problems in how `deploymentConfigChanged` compares volumes that could cause spurious updates to the router deployment.

#### Add `cmpEnvs` helper function

* `pkg/operator/controller/controller_router_deployment.go` (`cmpEnvs`): Add helper function.
(`deploymentConfigChanged`): Use it.

#### Ignore volumes ordering

Ignore the ordering of volumes in the deployment's pod template spec when comparing deployments.

* `pkg/operator/controller/controller_router_deployment.go` (`deploymentConfigChanged`): Sort volumes on name when comparing volumes.
(`cmpVolumes`): New helper function.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add test case for volumes changing ordering.

#### Fix volume default mode comparison

The default value for a secret volume source's default mode is 420, so treat a nil value as equal to 420.

* `pkg/operator/controller/controller_router_deployment.go` (`cmpSecretVolumeSource`): New helper function.
(`deploymentConfigChanged`): Use `cmpSecretVolumeSource` when comparing volumes.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add test cases where the default mode for the default-certificates volume changes.


ironcladlou:
Nice, thanks
/lgtm

PR #183, opened 2019-03-26
Adds IngressController CRD Validation & Generation

danehans:
PR #185 is required before merging.

Previously, `IngressController` resource fields were not validated. Additionally, the associated CRD was manually managed. This commit adds:

- OpenAPI v3 schema validation to the `IngressController` custom resource and partially fixes Bug [1683766](https://bugzilla.redhat.com/show_bug.cgi?id=1683766).

- Adds `make crd` and `make verify-crd` to generate and verify crd modification respectively.
 
__Note:__ The schema is based on the currently vendored api definitions (853a7faf74b1c63a56e4010d39057559fa13271f) and not the latest from openshift/api.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,18 @@ build:
 	$(GO_BUILD_RECIPE)
 
 .PHONY: generate
-generate:
+generate: crd
 	hack/update-generated-bindata.sh
+
+# Generate ClusterIngress CRD from vendored API spec.
+.PHONY: crd
+crd:
+	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --apis-dir vendor/github.com/openshift/api
+
+# Do not write the ClusterIngress CRD, only compare and return (code 1 if dirty).
+.PHONY: verify-crd
+verify-crd:
+	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --apis-dir vendor/github.com/openshift/api --verify-only
 
 .PHONY: test
 test: verify
@@ -35,7 +45,7 @@ clean:
 	rm -f $(BIN)
 
 .PHONY: verify
-verify:
+verify: verify-crd
 	hack/verify-gofmt.sh
 	hack/verify-generated-bindata.sh
 
--- a/manifests/00-custom-resource-definition.yaml
+++ b/manifests/00-custom-resource-definition.yaml
@@ -17,3 +17,79 @@ spec:
       specReplicasPath: .spec.replicas
       statusReplicasPath: .status.availableReplicas
       labelSelectorPath: .status.labelSelector
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            defaultCertificate:
+              type: object
+            domain:
+              type: string
+            endpointPublishingStrategy:
+              properties:
+                type:
+                  type: string
+              required:
+              - type
+              type: object
+            namespaceSelector:
+              type: object
+            nodePlacement:
+              properties:
+                nodeSelector:
+                  type: object
+              type: object
+            replicas:
+              format: int32
+              type: integer
+            routeSelector:
+              type: object
+          type: object
+        status:
+          properties:
+            availableReplicas:
+              format: int32
+              type: integer
+            conditions:
+              items:
+                properties:
+                  lastTransitionTime:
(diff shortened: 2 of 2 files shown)


manifests/00-custom-resource-definition.yaml
```diff
@@ -17,3 +17,187 @@ spec:
...
+        status:
+          description: status is the most recently observed status of the IngressController.
+          properties:
+            availableReplicas:
+              description: availableReplicas is number of observed available replicas
+                according to the ingress controller deployment.
+              format: int32
+              type: integer
+            conditions:
+              description: 'conditions is a list of conditions and their status.  Available
```
lihongan: I didn't see `status.conditions` when running `oc get ingresscontrollers.operator.openshift.io/default -n openshift-ingress-operator -o yaml`.

danehans: @lihongan status conditions are being implemented in PR #174, which is expected to merge very soon.

lihongan: Get it, thank you @danehans

ironcladlou:
>PR #185 is required before merging.

I think @Miciah at least (maybe also @pravisankar) prefers a single PR for the dep updates and the code which uses the new deps — if so, I'm happy to conform to the majority preference (and then #185 could be brought over into this PR). Up to you guys!

manifests/00-custom-resource-definition.yaml
```diff
@@ -17,3 +17,187 @@ spec:
...
+                False if any of those conditions are unsatisfied.    * LoadBalancerReady   -
+                True if the following conditions are met:     * A load balancer is
+                managed.     * The load balancer is ready.   - False if any of those
+                conditions are unsatisfied.    * DNSManaged   - True if the following
+                conditions are met:     * The endpoint publishing strategy and platform
+                support DNS.     * The ingress controller domain is set.     * dns.config.openshift.io/cluster
+                configures DNS zones.   - False if any of those conditions are unsatisfied.    *
+                DNSReady   - True if the following conditions are met:     * DNS is
+                managed.     * DNS records have been successfully created.   - False
+                if any of those conditions are unsatisfied.'
```
ironcladlou: Are the line breaks really lost in this translation? 🤨

Makefile
```diff
@@ -17,6 +17,16 @@ build:
 generate:
 	hack/update-generated-bindata.sh
 
+# Generate ClusterIngress CRD from vendored API spec.
```
ironcladlou: How do we ensure in CI that the generated validations are in sync w/ the vendored types?

manifests/00-custom-resource-definition.yaml
danehans: Yeah, I noticed the generated formatting of descriptions are a bit skewed from the source spec. We can potentially try to fix this upstream. It seems painful to make changes in the api docs to be more generator friendly or manually manage the diffs.

Makefile
danehans: `make verify-crd` will return `1` if the vendored types differ from the generated validations and vice versa. For example, I added:
```
FooSelector *metav1.LabelSelector `json:"fooSelector,omitempty"`
```
to `NodePlacement` in the vendored `IngressController` type and ran `make verify-crd`:
```
<SNIP>
@@ -99,6 +99,8 @@
                 of the ingress controller.  If unset, defaults are used. See NodePlacement
                 for more details.
               properties:
+                fooSelector:
+                  type: object
                 nodeSelector:
                   description: 'nodeSelector is the node selector applied to ingress
                     controller deployments.  If unset, the default is:    beta.kubernetes.io/os:

verification failed
exit status 1
make: *** [verify-crd] Error 1
```

Should I add an inline doc to the openshift/api section of `Gopkg.toml` saying to run `make verify-crd` after any updates to the dep?

ironcladlou: If you add it to the `verify` target, CI will catch it during unit test execution (which depends on `verify`).

What about adding the generation step in the existing `generate` target and adding the verification to `verify`?

ironcladlou:
Let's get this in to get a foundation established.

/lgtm

danehans:
@ironcladlou I created a branch in my repo and layered [another commit](https://github.com/danehans/cluster-ingress-operator/commit/002dfcef7fb7c7e9930bbe6057582896c2eabae1) on top of this one that adds more granular validation for the `endpointPublishingStrategy` and `domain` fields. Based on the feedback during today's standup, I decided to keep them separate. Let me know if you would like it rolled into this PR.

PR #187, opened 2019-03-28
Adds post admission validation for IngressController

danehans:
Requires PR #186 

Previously, cluster-ingress-operator performed no validation of `IngressController` fields. This commit adds the following post admission validation:

- An object's metadata on creation and that `name` complies with the [rfc 1035](https://tools.ietf.org/html/rfc1035) label format.
- If provided, `domain` complies with [rfc 1123](https://tools.ietf.org/html/rfc1123) domain format.
- If provided, `replicas` is greater than zero.
- If provided , `endpointPublishingStrategy` is one of the three supported types.

Partially fixes bug #1683766


ironcladlou:
We need to have a design convo about static validations 

/hold

ironcladlou:
@deads2k suggests we can implement regex validators in the openapi validation, so let's talk more about our options

danehans:
OK, I was using CVO as a reference implementation.

PR #189, opened 2019-03-28
Remove unnecessary operator config in manifests factory

pravisankar:


pravisankar:

PR #190, opened 2019-03-28
haproxy: set nbthread to 4 by default

ironcladlou:
We want the gains and to soak test for regressions.

/cc @smarterclayton @knobunc @openshift/sig-network-edge @mcurry-rh 

For your consideration.


ironcladlou:
/xref https://coreos.slack.com/archives/CCH60A77E/p1553807594085800

smarterclayton:
/lgtm

Let’s see what happens.  Will watch the queues

PR #191, opened 2019-03-28
Fix error handling in currentRouterDeployment method

pravisankar:


pravisankar:

PR #194, opened 2019-04-01
Bug 1696168: Use 'ingresscontroller.operator.openshift.io' prefix for ingress controller labels

pravisankar:
- Router deployment and service resource name and namespace are set at runtime.
- Remove stale operator role/rolebindings from manifests.
- Rename 'cluster ingress' references to 'ingress controller'.

Most of this is feedback on cluster dns operator https://github.com/openshift/cluster-dns-operator/pull/86
which also holds for ingress operator.

Also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1696168


pravisankar:

Miciah:
Not sure about the implications for upgrades of changing the finalizer.  Otherwise looks good.

pkg/manifests/manifests.go
```diff
@@ -42,7 +40,7 @@ const (
 	// OwningClusterIngressLabel should be applied to any objects "owned by" a
 	// clusteringress to aid in selection (especially in cases where an ownerref
 	// can't be established due to namespace boundaries).
-	OwningClusterIngressLabel = "ingress.openshift.io/clusteringress"
+	OwningClusterIngressLabel = "ingress.operator.openshift.io/clusteringress"
```
ironcladlou: if we're changing this, what about also changing `/clusteringress` to `/ingresscontroller` or something?

pkg/operator/controller/controller.go
```diff
@@ -35,7 +35,7 @@ const (
 	// IngressControllerFinalizer is applied to an IngressController before being
 	// considered for processing; this ensures the operator has a chance to handle
 	// all states.
-	IngressControllerFinalizer = "ingress.openshift.io/ingress-controller"
+	IngressControllerFinalizer = "ingress.operator.openshift.io/ingress-controller"
```
ironcladlou: likewise, `/ingress-controller` to `/ingresscontroller` or just `/controller`?

ironcladlou:
@derekwaynecarr @smarterclayton 
>Not sure about the implications for upgrades of changing the finalizer. Otherwise looks good.

Are we still allowed to make changes which would break upgrades for existing beta clusters?

Miciah:
/lgtm cancel
We have some ongoing discussion about how to name finalizers...

pkg/manifests/manifests.go
```diff
@@ -40,9 +38,9 @@ const (
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
 
 	// OwningClusterIngressLabel should be applied to any objects "owned by" a
-	// clusteringress to aid in selection (especially in cases where an ownerref
+	// ingress controller to aid in selection (especially in cases where an ownerref
 	// can't be established due to namespace boundaries).
-	OwningClusterIngressLabel = "ingress.openshift.io/clusteringress"
+	OwningClusterIngressLabel = "ingress.operator.openshift.io/owning-ingresscontroller"
```
danehans: Due to the name change, I believe `OwningClusterIngressLabel` should be updated to `OwningIngressControllerLabel`.

pkg/manifests/manifests.go
```diff
@@ -40,9 +38,9 @@ const (
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
 
 	// OwningClusterIngressLabel should be applied to any objects "owned by" a
```
danehans: Due to the name change, I believe `OwningClusterIngressLabel` should be updated to `OwningIngressControllerLabel`.

ironcladlou:
/hold cancel

We've got [a request](https://bugzilla.redhat.com/show_bug.cgi?id=1696168) to purge user-facing. This PR can be part of the fix.

pkg/manifests/manifests.go
```diff
@@ -39,10 +37,10 @@ const (
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
 
-	// OwningClusterIngressLabel should be applied to any objects "owned by" a
-	// clusteringress to aid in selection (especially in cases where an ownerref
+	// OwningIngressControllerLabel should be applied to any objects "owned by" a
+	// ingress controller to aid in selection (especially in cases where an ownerref
 	// can't be established due to namespace boundaries).
-	OwningClusterIngressLabel = "ingress.openshift.io/clusteringress"
+	OwningIngressControllerLabel = "ingresscontroller.operator.openshift.io/owning-ingresscontroller"
```
ironcladlou: Isn't this inventing a new nonexistent API Group? compare to `ingress.operator.openshift.io/owner` or something

pkg/operator/controller/controller.go
```diff
@@ -35,7 +35,7 @@ const (
 	// IngressControllerFinalizer is applied to an IngressController before being
 	// considered for processing; this ensures the operator has a chance to handle
 	// all states.
-	IngressControllerFinalizer = "ingress.openshift.io/ingress-controller"
+	IngressControllerFinalizer = "ingresscontroller.operator.openshift.io/finalizer-ingresscontroller"
```
ironcladlou: likewise re: api group

pkg/manifests/manifests.go
Miciah: What do you mean? `operator.openshift.io` is the API group, and it has an `IngressController` kind (no `Ingress` kind).

pravisankar: Looked at 3.x labels and 4.x labels used by other operators and there is no specific convention for owner or finalizer.
Also I think we don't have to be rigid on specific labels/annotations unless we expect some interaction between the operators or some other tools.

Miciah: Agreed, we seem to have flexibility to choose whatever makes sense.  I think `ingresscontroller.operator.openshift.io` makes more sense than either `ingress.openshift.io/clusteringress` or `ingress.operator.openshift.io` because they all look like either `<apigroup>/<kind>` or `<kind>.<apigroup>`, but there is no `Ingress` resource in the `operator.openshift.io` API group, and there is no `ingress.openshift.io` API group or `ClusterIngress` kind anymore whereas there _is_ an `IngressController` kind in the `operator.openshift.io` API group.

pravisankar:
@openshift/sig-network-edge updated, PTAL.

I just want to point out that I have ignored some 'cluster ingress' to 'ingress controller' renaming that could have wider impact and may need some discussion. I will do a follow up pr if we want to make any of these changes:

- Rename 'cluster-ingress-operator' binary name to 'ingress-controller-operator'
- Rename 'github.com/openshift/cluster-ingress-operator' repo to 'github.com/openshift/ingress-controller-operator'. May be this is not feasible at this stage?
- Rename 'clusteroperator/ingress' resource to 'clusteroperator/ingresscontroller'. 'ingresscontroller' may be more appropriate as there is a 'ingress' resource.
- Use 'ingress-controller-operator' as signer name in generateRouterCA() [file: pkg/operator/controller/certificate/ca.go]. This may break existing ca trust chain?

Miciah:
> * Rename 'cluster-ingress-operator' binary name to 'ingress-controller-operator'

How about `ingress-operator`? I think "ingress operator" and "ingress controller" are clear enough, and "ingress controller operator" is wordy.

> * Rename 'github.com/openshift/cluster-ingress-operator' repo to 'github.com/openshift/ingress-controller-operator'. May be this is not feasible at this stage?

Seems like a lot of effort for something that is not user-facing.

> * Rename 'clusteroperator/ingress' resource to 'clusteroperator/ingresscontroller'. 'ingresscontroller' may be more appropriate as there is a 'ingress' resource.

I prefer keeping the name "ingress" for the `ClusterOperator` resource.  Ingress operator manages ingress controllers, which provide ingress as specified by `Ingress` and `Route` resources.

> * Use 'ingress-controller-operator' as signer name in generateRouterCA() [file: pkg/operator/controller/certificate/ca.go]. This may break existing ca trust chain?

Shouldn't be a problem unless we gratuitously recreate the CA and fail to regenerate certificates (or fail to publish the regenerated certificates in the `router-certs` secret), which would be a defect in itself.

pravisankar:
Renamed binary to 'ingress-operator' and updated signer name in generateRouterCA() to 'ingress-operator'.

pravisankar:
@openshift/sig-network-edge @ironcladlou @Miciah updated, PTAL

pkg/operator/controller/certificate/controller.go
```diff
@@ -107,7 +107,7 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	ingresses := &operatorv1.IngressControllerList{}
 	if err := r.client.List(context.TODO(), ingresses, client.InNamespace(r.operatorNamespace)); err != nil {
-		errs = append(errs, fmt.Errorf("failed to list clusteringresses: %v", err))
+		errs = append(errs, fmt.Errorf("failed to list ingresscontrolleres: %v", err))
```
Miciah: Typo: "ingresscontrolleres"→""ingresscontrollers".

PR #195, opened 2019-04-03
deps: integrate with latest openshift version

ironcladlou:
* Upgrade to latest openshift/api and libary-go
* Upgrade to kube 1.13
* Upgrade to latest compatible controller-runtime

Fixes openapi spec generation, which as of the previous version was not picking
up field descriptions.


Gopkg.toml
```diff
@@ -1,10 +1,6 @@
 # Force dep to vendor the code generators, which aren't imported just used at dev time.
 # Picking a subpackage with Go code won't be necessary once https://github.com/golang/dep/pull/1545 is merged.
 required = [
-  "k8s.io/code-generator/cmd/defaulter-gen",
```
danehans: Are these code generators being removed b/c https://github.com/golang/dep/pull/1545 merged?

Gopkg.toml
```diff
@@ -38,24 +34,34 @@ required = [
 
 [[constraint]]
   name = "github.com/openshift/library-go"
-  revision = "fc4b2d5cbcf0ab855422e340c78fe7a53a38bee3"
+  revision = "9067dee609b10991fa6c26b5d51fe095b2331027"
```
danehans: Is there a reason for using this rev instead of latest (`dab26bb3a8dc7fccde7227194af755bbff30ce5d`)?

ironcladlou:
https://github.com/kubernetes-sigs/controller-runtime/pull/383

ironcladlou:
Need a resolution to https://github.com/kubernetes-sigs/controller-runtime/pull/383.

/hold

Gopkg.toml
ironcladlou: Typo on my part, good catch!

Gopkg.toml
ironcladlou: Latest commit cleans up the entire file including the `required` stanza. We still need some stuff in `required` for non-imported code generation, but the subpackage note isn't relevant to us.

ironcladlou:
@openshift/sig-network-edge 

I don't want to wait for this. I'd like to carry the patched `Source` in our tree (e.g. in pkg/util) and duplicate the dependent controller-runtime packages to make it work. When/if upstream merges, we can remove our copy.

If there are no objections, I'll do this tomorrow (April 5).

ironcladlou:
I figured, why wait. I went ahead and pushed the local Source implementation. If upstream patches it'll be trivial to delete our local version.

This is ready for review.

/hold cancel

ironcladlou:
https://github.com/kubernetes-sigs/controller-runtime/pull/383 merged last night so I fixed up the PR to use it!

Miciah:
Typo in the commit message: "libary-go".

ironcladlou:
>Typo in the commit message: "libary-go".

Fixed

pkg/util/source/doc.go
```diff
@@ -0,0 +1,7 @@
+// TODO: this package locally re-implements the following upstream patch:
+//
+// https://github.com/kubernetes-sigs/controller-runtime/pull/383
+//
+// If upstream accepts the patch, this package can be deleted and replaced by
+// upstream code.
+package source
```
Miciah: Looks like your forgot to remove this file.

ironcladlou: I must have failed to isolate those changed in the deleted commit. Thanks!

PR #196, opened 2019-04-03
Do not need to call manifests.RouterDeployment() in ensureRouterDeleted()

pravisankar:


pravisankar:

PR #197, opened 2019-04-05
Adds Watch error handling

danehans:
This commit adds error handling for the `Watch` method.


Miciah:
Well, that's confusing.

Miciah:
I've definitely seen people use `/hold` with a reason, but I guess it doesn't actually do anything.

PR #198, opened 2019-04-08
Sets status conditions for ingress delete

danehans:
Previously, operator 'Progressing' and 'Available' status conditions would not transition when 0 ingress controllers exist. This would cause status conditions to only be recomputed when the new default ingress was auto generated. This commit updates how these two conditions are computed when no ingress controllers exist.

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -131,22 +131,28 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 	}
 	numIngresses := len(ingresses)
 	var ingressesAvailable int
-	for _, ing := range ingresses {
-		for _, c := range ing.Status.Conditions {
-			if c.Type == operatorv1.IngressControllerAvailableConditionType && c.Status == operatorv1.ConditionTrue {
-				ingressesAvailable++
-				break
+	if numIngresses == 0 {
+		progressingCondition.Status = configv1.ConditionTrue
+		progressingCondition.Reason = "NoIngressControllers"
+		progressingCondition.Message = "0 ingress controllers available, want at least 1"
+	} else {
+		for _, ing := range ingresses {
+			for _, c := range ing.Status.Conditions {
+				if c.Type == operatorv1.IngressControllerAvailableConditionType && c.Status == operatorv1.ConditionTrue {
+					ingressesAvailable++
+					break
+				}
 			}
 		}
-	}
-	if numIngresses == ingressesAvailable {
-		progressingCondition.Status = configv1.ConditionFalse
-	} else {
-		progressingCondition.Status = configv1.ConditionTrue
-		progressingCondition.Reason = "Reconciling"
-		progressingCondition.Message = fmt.Sprintf(
-			"%d ingress controllers available, want %d",
-			ingressesAvailable, numIngresses)
+		if numIngresses == ingressesAvailable {
+			progressingCondition.Status = configv1.ConditionFalse
+		} else {
+			progressingCondition.Status = configv1.ConditionTrue
+			progressingCondition.Reason = "Reconciling"
+			progressingCondition.Message = fmt.Sprintf(
+				"%d ingress controllers available, want %d",
+				ingressesAvailable, numIngresses)
+		}
 	}
 	conditions = setOperatorStatusCondition(conditions, progressingCondition)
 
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/controller.go
```diff
@@ -149,6 +149,8 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 					// Handle deletion.
 					if err := r.ensureIngressDeleted(ingress, dnsConfig, infraConfig); err != nil {
 						errs = append(errs, fmt.Errorf("failed to ensure ingress deletion: %v", err))
+					} else if err := r.syncOperatorStatus(); err != nil {
```
ironcladlou: This doesn't seem to match the PR description. If `r.ensureIngressDeleted` on L150 returns an error, isn't `r.syncOperatorStatus` on L168 still called?

pkg/operator/controller/status.go
```diff
@@ -139,6 +139,11 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 			}
 		}
 	}
+	if numIngresses == 0 {
```
ironcladlou: Not sure I understand the intent here. Internally misrepresenting the actual number of ingresses seems weird.

pkg/operator/controller/status_test.go
```diff
@@ -24,8 +24,8 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
 		inputs      testInputs
 		outputs     testOutputs
 	}{
-		{"no operand namespace", testInputs{false, 0, 0}, testOutputs{true, false, true}},
-		{"no ingresscontrollers available", testInputs{true, 0, 0}, testOutputs{false, false, true}},
```
ironcladlou: It's not clear to me why this test case was wrong.

Also, consider using explicit field labels here. It's extremely difficult to understand these test cases because of the requisite context switching to refer back to the `testInput` and `testOutput` structs.

danehans: Thank you for reviewing my PR.  I have pushed a new PR to address your comments.  The previous test case was not wrong.  Previously, it was acceptable to have zero active ingress controllers (from a status conditions perspective). This would cause no transition of the available and progressing status conditions.  With Bug 1695210, having no ingress controllers require the progressing condition to be true and the available condition to be false.

Miciah:
What does ingress delete have to do with upgrades? As I understand the problem, we need to update the `Progressing` condition when `.status.versions[?name=='operator'].version` changes (we could set `.status.conditions[?type=='Progressing'].message` to `fmt.Sprintf("updated to version %q", r.OperatorReleaseVersion)` to force an update so last transition time is updated).  Maybe we need clarification on what an upgrade means to us, or how the operator is supposed to tell an upgrade from a restart when the operand image does not change.

danehans:
@Miciah after our team discussion, I realize that this PR will not solve the upgrade bug. With that said, should the Operator reflect the same failed, progressing and available status conditions whether the desired number of ingress controllers is 0 or greater than 0? IMO, since the Operator requires the default ingress controller, the status conditions should reflect this.

Miciah:
> With that said, should the Operator reflect the same failed, progressing and available status conditions whether the desired number of ingress controllers is 0 or greater than 0?

I'm not sure I understand the question.  During the call, we discussed questions along the lines of the following (among others):
* "What should the operator set for the `Available` condition on the `ClusterOperator` resource if the number of ingress controllers is 0?"
* "What should the operator set for the `Progressing` condition on the `ClusterOperator` resource if the number of ingress controllers is 0?"

Are you asking more or less the same question here?

> IMO, since the Operator requires the default ingress controller, the status conditions should reflect this.

Right, at least in the case of `Progressing`, I think we agreed that `Progressing` should be `True` if there are 0 ingress controllers, because we know that we have a goroutine periodically calling `ensureDefaultIngressController`, which will create the default ingress controller.  (I was a little concerned about the hidden dependency of the status code on this goroutine, but for now we can brush that aside.)

I believe we were also leaning towards a broad definition of `Available`, where we would say that `Available` should be `False` if there are 0 ingress controllers (technically the operator is available and providing the `IngressController` API, but without any ingress controller, we effectively have no ingress into the cluster).

For `Failing`, the answer is less clear, but we did talk about setting `Failing` to `True` if the `Available` condition remains `False`, or if `Progressing` remains `True`, for longer than some specific period of time.  (For example, we could set `Failing` to `False` if `Available` is `False` and the last transition time of `Available` is over 5 minutes in the past.)

danehans:
Superseded by https://github.com/openshift/cluster-ingress-operator/pull/201

PR #200, opened 2019-04-09
Bug 1704982: Fix updating versions for ingress cluster operator

pravisankar:
- Always report 'operator' version irrespective of available condition.

- For upgrades, report old versions until the new version is fully applied.

### DIFF
--- a/manifests/03-cluster-operator.yaml
+++ b/manifests/03-cluster-operator.yaml
@@ -1,7 +1,7 @@
+# This resource is not created by CVO, it is only used as a communication
+# mechanism to tell the CVO which ClusterOperator resource to wait for.
+# IngressController creates and updates this resource.
 apiVersion: config.openshift.io/v1
 kind: ClusterOperator
 metadata:
   name: ingress
-spec: {}
-status:
-  # status is set at runtime
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -20,18 +21,24 @@ import (
 )
 
 const (
-	IngressClusterOperatorName     = "ingress"
-	UnknownReleaseVersionName      = "unknown"
+	IngressClusterOperatorName = "ingress"
+
+	OperatorVersionName          = "operator"
+	IngressControllerVersionName = "ingress-controller"
+	UnknownVersionValue          = "unknown"
+
 	ingressesEqualConditionMessage = "desired and current number of IngressControllers are equal"
-	operatorVersionName            = "operator"
 )
 
 // syncOperatorStatus computes the operator's current status and therefrom
 // creates or updates the ClusterOperator resource for the operator.
 func (r *reconciler) syncOperatorStatus() error {
+	ns := manifests.RouterNamespace()
+
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: IngressClusterOperatorName}}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: co.Name}, co); err != nil {
 		if errors.IsNotFound(err) {
+			initializeClusterOperator(co, ns.Name)
 			if err := r.client.Create(context.TODO(), co); err != nil {
 				return fmt.Errorf("failed to create clusteroperator %s: %v", co.Name, err)
 			}
@@ -40,44 +47,20 @@ func (r *reconciler) syncOperatorStatus() error {
 			return fmt.Errorf("failed to get clusteroperator %s: %v", co.Name, err)
 		}
 	}
+	oldStatus := co.Status.DeepCopy()
 
-	ns, ingresses, err := r.getOperatorState()
+	ingresses, ns, err := r.getOperatorState(ns.Name)
 	if err != nil {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
+	allIngressesAvailable := checkAllIngressesAvailable(ingresses)
 
-	oldStatus := co.Status.DeepCopy()
-	co.Status.Conditions = computeOperatorStatusConditions(oldStatus.Conditions, ns, ingresses)
-	co.Status.RelatedObjects = []configv1.ObjectReference{
-		{
-			Resource: "namespaces",
-			Name:     "openshift-ingress-operator",
-		},
-		{
-			Resource: "namespaces",
-			Name:     ns.Name,
-		},
-	}
-
-	// An available operator resets release version
-	for _, condition := range co.Status.Conditions {
-		if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue {
-			co.Status.Versions = []configv1.OperandVersion{
-				{
-					Name:    operatorVersionName,
-					Version: r.OperatorReleaseVersion,
-				},
-				{
-					Name:    "ingress-controller",
-					Version: r.RouterImage,
-				},
-			}
-		}
-	}
+	co.Status.Versions = r.computeOperatorStatusVersions(oldStatus.Versions, allIngressesAvailable)
+	co.Status.Conditions = r.computeOperatorStatusConditions(oldStatus.Conditions,
+		ns, allIngressesAvailable, co.Status.Versions)
 
 	if !operatorStatusesEqual(*oldStatus, co.Status) {
-		err = r.client.Status().Update(context.TODO(), co)
-		if err != nil {
+		if err := r.client.Status().Update(context.TODO(), co); err != nil {
 			return fmt.Errorf("failed to update clusteroperator %s: %v", co.Name, err)
 		}
 	}
@@ -85,18 +68,53 @@ func (r *reconciler) syncOperatorStatus() error {
 	return nil
 }
 
+// Populate versions and conditions in cluster operator status as CVO expects these fields.
+func initializeClusterOperator(co *configv1.ClusterOperator, nsName string) {
+	co.Status.Versions = []configv1.OperandVersion{
+		{
+			Name:    OperatorVersionName,
+			Version: UnknownVersionValue,
+		},
+		{
+			Name:    IngressControllerVersionName,
+			Version: UnknownVersionValue,
+		},
+	}
+	co.Status.Conditions = []configv1.ClusterOperatorStatusCondition{
+		{
+			Type:   configv1.OperatorDegraded,
+			Status: configv1.ConditionUnknown,
+		},
+		{
+			Type:   configv1.OperatorProgressing,
+			Status: configv1.ConditionUnknown,
+		},
+		{
+			Type:   configv1.OperatorAvailable,
+			Status: configv1.ConditionUnknown,
+		},
+	}
+	co.Status.RelatedObjects = []configv1.ObjectReference{
+		{
+			Resource: "namespaces",
+			Name:     "openshift-ingress-operator",
+		},
+		{
+			Resource: "namespaces",
+			Name:     nsName,
+		},
+	}
+}
+
 // getOperatorState gets and returns the resources necessary to compute the
 // operator's current state.
-func (r *reconciler) getOperatorState() (*corev1.Namespace, []operatorv1.IngressController, error) {
-	ns := manifests.RouterNamespace()
-
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
+func (r *reconciler) getOperatorState(nsName string) ([]operatorv1.IngressController, *corev1.Namespace, error) {
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: nsName}}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: nsName}, ns); err != nil {
 		if errors.IsNotFound(err) {
 			return nil, nil, nil
 		}
-
-		return nil, nil, fmt.Errorf(
-			"error getting Namespace %s: %v", ns.Name, err)
+		return nil, nil, fmt.Errorf("error getting Namespace %s: %v", nsName, err)
 	}
 
 	ingressList := &operatorv1.IngressControllerList{}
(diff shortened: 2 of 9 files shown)


pravisankar:

Miciah:
CVO needs the `versions` stanza in `03-cluster-operator.yaml` to cause it to wait, so that change makes sense.  With that change, does the operator still need to set the `versions` before it is available? The operator should already keep whatever old `versions` is in the `ClusterOperator` resource until the new version is available because `syncOperatorStatus` gets the old status and mutates it to get the new status. Is CVO stripping the `versions` stanza when it instantiates `03-cluster-operator.yaml`?

danehans:
@pravisankar PTAL @ https://github.com/openshift/cluster-ingress-operator/pull/201 which updates the `Progressing` status condition message when configured and running versions are unequal. This triggers an updated `LastTransitionTime`. In this PR, can you update the `Progressing` message when `computeOperatorStatusVersions()` updates any version information? This will cause the `ClusterOperator` resource to properly convey status messages.

danehans:
@pravisankar needs rebase.

pravisankar:
/hold cancel
@openshift/sig-network-edge @danehans ready for review, PTAL

danehans:
`fail [github.com/openshift/origin/test/extended/networking/multicast.go:50]: Expected success, but got an error`

/test e2e-aws

pkg/operator/controller/status.go
```diff
@@ -40,71 +46,153 @@ func (r *reconciler) syncOperatorStatus() error {
...
+		releaseVersion = oldReleaseVersion
+		ingressControllerImage = oldIngressControllerImage
 	}
 
-	ingressList := &operatorv1.IngressControllerList{}
-	if err := r.client.List(context.TODO(), ingressList, client.InNamespace(r.Namespace)); err != nil {
-		return nil, nil, fmt.Errorf("failed to list IngressControllers: %v", err)
+	// Update progressing condition if needed
+	if (releaseVersion != r.OperatorReleaseVersion) || (ingressControllerImage != r.IngressControllerImage) {
+		if !operatorProgressing {
```
danehans: Can you help me understand why progressing is set to true/"Reconciling" when `operatorProgressing` is false? It seems like it would be the opposite.

pkg/operator/controller/status.go
```diff
@@ -40,71 +46,153 @@ func (r *reconciler) syncOperatorStatus() error {
...
+			}
+			progressingCondition.Message += fmt.Sprintf("Moving to release version %q.", r.OperatorReleaseVersion)
+		}
+		if ingressControllerImage != r.IngressControllerImage {
+			if len(progressingCondition.Message) != 0 {
+				progressingCondition.Message += " "
+			}
+			progressingCondition.Message += fmt.Sprintf("Moving to ingress-controller image version %q.", r.IngressControllerImage)
+		}
+		//setOperatorStatusCondition(conditions, progressingCondition)
```
danehans: I think this line was left over.

pkg/operator/controller/status.go
```diff
@@ -40,71 +46,153 @@ func (r *reconciler) syncOperatorStatus() error {
...
-		return nil, nil, fmt.Errorf("failed to list IngressControllers: %v", err)
+	// Update progressing condition if needed
+	if (releaseVersion != r.OperatorReleaseVersion) || (ingressControllerImage != r.IngressControllerImage) {
+		if !operatorProgressing {
+			progressingCondition.Status = configv1.ConditionTrue
+			progressingCondition.Reason = "Reconciling"
+		}
+		if releaseVersion != r.OperatorReleaseVersion {
+			if len(progressingCondition.Message) != 0 {
+				progressingCondition.Message += " "
```
danehans: IMO the progressing message can simply be updated instead of appending the new message to the old message. Thoughts?

pkg/operator/controller/status_test.go
```diff
@@ -434,3 +437,303 @@ func TestOperatorStatusesEqual(t *testing.T) {
...
+			inputs: testInputs{
+				oldReleaseVersion:         "v1",
+				oldIngressControllerImage: "ic-v1",
+				curReleaseVersion:         "v2",
+				curIngressControllerImage: "ic-v1",
+			},
+			outputs: testOutputs{
+				releaseVersion:         "v1",
+				ingressControllerImage: "ic-v1",
+				versionChanged:         true,
```
danehans: I'm not following this test case. The operator is not available and not progressing, the versions in `outputs` does not change from the old versions but `versionChanged: true`?

pkg/operator/controller/status_test.go
```diff
@@ -434,3 +437,303 @@ func TestOperatorStatusesEqual(t *testing.T) {
...
+				oldIngressControllerImage: "ic-v1",
+				curReleaseVersion:         "v2",
+				curIngressControllerImage: "ic-v1",
+				progressingCondition:      true,
+				availableCondition:        true,
+			},
+			outputs: testOutputs{
+				releaseVersion:         "v1",
+				ingressControllerImage: "ic-v1",
+				versionChanged:         true,
```
danehans: Sam as my feedback above. It seems like `versionChanged` should be `false` if the outputs versions do not change.

pkg/operator/controller/status.go
pravisankar: `releaseVersion != r.OperatorReleaseVersion` is true if we did not move to the new version yet. This could happen when the operator condition is not available or is progressing. 
If the operator condition is progressing, progressing condition status and reason is already set and we just need to update the message with version info.
If the operator condition is not progressing (number of ingresses available and desired matched) and not available, then we do need to set progressing condition status to true and the reason as 'reconciling' to reflect the progressing state due to version mismatch.

pkg/operator/controller/status.go
pravisankar: I thought by appending the message, we are providing accurate state noticed by the controller.
Depending on the circumstances, message may have one ore more of these strings:
- X ingress controllers available, want Y
- Moving to release version X
- Moving to ingress-controller image version Y

pkg/operator/controller/status_test.go
pravisankar: `versionChanged` was used to indicate progressing is true and to set corresponding message 'Moving to release version X' but I see the confusion that you stated. I will change the variable name.

pkg/operator/controller/status.go
```diff
@@ -40,71 +46,153 @@ func (r *reconciler) syncOperatorStatus() error {
 			return fmt.Errorf("failed to get clusteroperator %s: %v", co.Name, err)
 		}
 	}
+	oldStatus := co.Status.DeepCopy()
 
-	ns, ingresses, err := r.getOperatorState()
+	ingresses, err := r.getOperatorState(ns.Name)
```
Miciah: `computeOperatorFailingCondition` is expecting `ns` to be nil if the namespace does not exist.  With this change, the `if ns ==n nil` in `computeOperatorFailingCondition` becomes dead code.

pkg/operator/controller/status.go
```diff
@@ -40,71 +46,153 @@ func (r *reconciler) syncOperatorStatus() error {
...
+		ingressControllerImage = oldIngressControllerImage
 	}
 
-	ingressList := &operatorv1.IngressControllerList{}
-	if err := r.client.List(context.TODO(), ingressList, client.InNamespace(r.Namespace)); err != nil {
-		return nil, nil, fmt.Errorf("failed to list IngressControllers: %v", err)
+	// Update progressing condition if needed
+	if (releaseVersion != r.OperatorReleaseVersion) || (ingressControllerImage != r.IngressControllerImage) {
+		if !operatorProgressing {
+			progressingCondition.Status = configv1.ConditionTrue
```
Miciah: Can you be sure `progressingCondition` is not nil?

pkg/operator/controller/status.go
Miciah: What keeps the operator from repeatedly appending the same "Moving to [...]" strings to the condition?

pravisankar: We create progressing condition from scratch in every status iteration and this eliminates appending of same "Moving to" messages (https://github.com/openshift/cluster-ingress-operator/blob/f09e8ee06d8f9bebb0e5aed94cbbe4a41e3eee38/pkg/operator/controller/status.go#L231).

pkg/operator/controller/status.go
pravisankar: Yes, We populate progressing condition during the cluster operator creation so it should not be nil.

Miciah: What if something else created the clusteroperator, or modified it by deleting the condition? Seems like this would fail and crashloop.  Granted, it is an unlikely scenario.

Miciah: I see, we can assume that `computeOperatorStatusConditions` will have set the condition.  Never mind!

pkg/operator/controller/status.go
Miciah: Ah, because `computeOperatorProgressingCondition` resets the message.  Got it.

ironcladlou: For the string building, you could start with a `messages []string`, append to it as you go, and at the end `condition.Messages = strings.Join(messages, "\n")`

pkg/operator/controller/status_test.go
```diff
@@ -434,3 +437,303 @@ func TestOperatorStatusesEqual(t *testing.T) {
 		}
 	}
 }
+
+func TestComputeOperatorStatusVersions(t *testing.T) {
```
ironcladlou: Consider a more compact table layout, e.g.

```go
{
  description: "update operator and ingress controller image, operator is available and progressing",
  inputs: testInputs{
    oldReleaseVersion:         "v1",
    oldIngressControllerImage: "ic-v1",
    curReleaseVersion:         "v2",
    curIngressControllerImage: "ic-v2",
    availableCondition:        true,
    progressingCondition:      true,
  },
  outputs: testOutputs{
    releaseVersion:         "v1",
    ingressControllerImage: "ic-v1",
    versionChanged:         true,
    ingressImageChanged:    true,
  },
},
```

becomes:

```go
{
  desc: "update operator and ingress controller image, operator is available and progressing",
  start: input{fromVersion:"v1", fromImage: "ic-v1", toVersion: "v2", toImage: "ic-v2", available: true, progressing: true},
  expect: output{version: "v1", image: "ic-v1", versionChanged: true, imageChanged: true}
},
```

Just for example

ironcladlou: Or perhaps:

```go
{
  desc: "update operator and ingress controller image, operator is available and progressing",
  from: versions{operator:"v1", operand: "ic-v1"},
  to: versions{operator:"v2", operand: "ic-v2"},
  in: condition{available: true, progressing: true},
  expect: expect{
    versions: versions{operator: "v2", operand: "ic-v2"},
    in: condition{versionChanged: true, imageChanged: true},
  },
},
```

pkg/operator/controller/status.go
pravisankar: fixed

pkg/operator/controller/status.go
pravisankar: done

pkg/operator/controller/status.go
pravisankar: fixed

pkg/operator/controller/status_test.go
pravisankar: Changed the test table to compact layout.

pravisankar:
> It seems weird to me to have `computeOperatorStatusConditions` compute the conditions and then have `computeOperatorStatusVersions` compute revised conditions as well as versions. Would it be feasible to do `computeOperatorStatusVersions` first, and then call `computeOperatorStatusConditions` with old and new versions as inputs so that `computeOperatorProgressingCondition` could compute the `Progressing` condition all in one place?

Unfortunately we can not compute versions before the conditions as it uses the available and progressing conditions. Yes, `computeOperatorStatusVersions` returning versions and conditions seems a bit odd in the earlier version of the code. Anyway I have reorganized the code and I hope it is easy to read now.

pravisankar:
@ironcladlou @Miciah @danehans ready for re-review, PTAL

manifests/03-cluster-operator.yaml
```diff
@@ -1,7 +1,7 @@
+# This resource is not created by CVO, it is only used as a communication
+# mechanism to tell the CVO which ClusterOperator resource to wait for.
+# IngressController creates and updates this resource.
```
ironcladlou: Does this resource's existence in `/manifests` mean that the CVO does in fact create the resource?

Miciah: Nope:

> Also note that the ClusterOperator resource in `/manifests` is only a communication mechanism, to tell the ClusterVersionOperator, which ClusterOperator resource to wait for. The ClusterVersionOperator does not create the ClusterOperator resource, creating and updating it is the responsibility of the respective operator.

https://github.com/openshift/cluster-version-operator/blob/8402d219f36fc79e03edf45918785376113f2cc1/docs/dev/clusteroperator.md#how-clusterversionoperator-handles-clusteroperator-in-release-image

Miciah:
> Unfortunately we can not compute versions before the conditions as it uses the available and progressing conditions. Yes, `computeOperatorStatusVersions` returning versions and conditions seems a bit odd in the earlier version of the code. Anyway I have reorganized the code and I hope it is easy to read now.

Thanks!  The latest version is an improvement, but I think we can simplify it further.  A better solution might come from thinking not in terms of conditions and versions, but rather in terms of available condition, degraded condition, versions, and progressing condition, and thinking about the inputs required to compute each, and what the outputs mean.  The functions compute values to answer the following questions:

| Function                              | Description                                                                                                                                                                                                                    |
| ---                                   | ---                                                                                                                                                                                                                            |
| `computeOperatorAvailableCondition`   | "Are all ingresses available?"                                                                                                                                                                                                 |
| `computeOperatorDegradedCondition`    | "Does the operand namespace exist?"                                                                                                                                                                                            |
| `computeOperatorStatusVersions`       | "What are the versions of the most recently deployed operator and (as a set) operands (i.e., those having `Available` true and `Progressing` false)?"                                                                          |
| `computeOperatorProgressingCondition` | "Are all ingresses available? Is the operator currently upgrading the operator itself or any of its operands (i.e., old and new versions are not equal)? Did the operator just complete an upgrade of itself or its operands?" |

(We do need to reexamine our interpretations of the conditions and change some of these questions, but let's stick with our current interpretations for this PR.)

The inputs and outputs are as follow:

| Function                              | Inputs                                                                                                                        | Outputs                                           |
| ---                                   | ---                                                                                                                           | ---                                               |
| `computeOperatorAvailableCondition`   | `oldStatus.Conditions` and `ingresses`                                                                                        | `conditions` with updated `Available`             |
| `computeOperatorDegradedCondition`    | `conditions` and `ns`                                                                                                         | `conditions` with updated `Degraded`              |
| `computeOperatorStatusVersions`       | `conditions` (for `Available`), `oldStatus.Versions`, `r.OperatorReleaseVersion`, `r.IngressControllerImage`, and `ingresses` | `versions` and "did we just complete an upgrade?" |
| `computeOperatorProgressingCondition` | `conditions`, `ingresses`, `oldStatus.Versions`, `versions`, and "did we just complete an upgrade?"                           | `conditions` with updated `Progressing`           |

Note that `computeOperatorStatusVersions` does not really need the operator `Progressing` condition's state; rather, it needs to know whether both of the following are true: the old and new operand versions are unequal and some ingress has a true `Progressing` condition, and we can answer this question by looking at ingresses without looking at operator conditions.

My primary aim here is to disentangle `computeOperatorStatusVersions` and `computeOperatorProgressingCondition`.  Two more changes I would suggest are that `syncOperatorStatus` should be calling `getOperatorState` rather than `computeOperatorConditionsAndVersions` calling it, and the compute functions should all have explicit inputs rather than having `reconciler` as a receiver.  In this way, I believe we can make the computation of these values more explicit and linear, by which I mean we compute versions and each condition only once, with clear inputs and outputs for each.  WDYT?

pravisankar:
@Miciah can we merge this pr to close the existing bug and handle the refactor that you suggested in a separate pr?

pravisankar:
@Miciah I tried to digest your comments on https://github.com/openshift/cluster-ingress-operator/pull/200#issuecomment-486752470
There are 2 design methodologies related to this in my mind:
(1) Gather everything you need for a specific task and then make a decision
(2) Incremental approach where the current decision is right for the current info it has processed and could change once it tries to process more info related to his.

Both (1) and (2) has its pros and cons. With (1) it is easy to reason but could mean code duplication sometimes and less flexible in where we can handle a particular situation. With (2) it is flexible, less code duplication but might be hard to reason and understand.
What you are suggested here is (1). I don't have a strong opinion but since we can eliminate code duplication in this specific case, for better readability I will vote for (1).

Unless we expect some other component other reconciler to use these `computeBlah` methods, I do not see much benefit of passing reconciler params as inputs instead of using directly as a receiver.

Yes, I agree that we need more clarity on 'Progressing' state. Questions like 'Do we want to populate a message when progressing=false?', etc. need to be answered.

pkg/operator/controller/status.go
```diff
@@ -40,71 +47,165 @@ func (r *reconciler) syncOperatorStatus() error {
...
+// updateOperatorStatusProgressingCondition updates operator's progressing condition with the version change message.
+func updateOperatorStatusProgressingCondition(message string, conditions []configv1.ClusterOperatorStatusCondition) {
+	for i := range conditions {
+		if conditions[i].Type == configv1.OperatorProgressing {
+			conditions[i].Status = configv1.ConditionTrue
+			conditions[i].Reason = "Reconciling"
+			if len(conditions[i].Message) != 0 {
+				conditions[i].Message += "\n"
+			}
+			conditions[i].Message += message
```
Miciah: It would be better to use `setOperatorStatusCondition` to make sure the last transition time is updated.

Miciah:
> There are 2 design methodologies related to this in my mind:
> (1) Gather everything you need for a specific task and then make a decision
> (2) Incremental approach where the current decision is right for the current info it has processed and could change once it tries to process more info related to his.

Yeah, that's one aspect, although see my suggestions more as promoting functional programming over procedural.

> Both (1) and (2) has its pros and cons. With (1) it is easy to reason but could mean code duplication sometimes and less flexible in where we can handle a particular situation. With (2) it is flexible, less code duplication but might be hard to reason and understand.

Right, it's a matter of readability and consistency: it's easier to reason about functions with clear inputs and fewer entanglements, and we have been generally moving in the direction of functional programming (I see an analogue between `syncOperatorStatus` and, e.g., `ensureRouterDeployment`: the former does a `Get` of the old clusteroperator, `getOperatorState`, `computeBlah`, `operatorStatusesEqual`, and finally an update, and the latter does a `currentRouterDeployment`, `desiredRouterDeployment`, `deploymentConfigChanged`, and update; generally, and roughly speaking, the pattern is get current, get desired, compare, and update).

> What you are suggested here is (1). I don't have a strong opinion but since we can eliminate code duplication in this specific case, for better readability I will vote for (1).

All right, I can implement my suggestions in a follow-up PR.

> Yes, I agree that we need more clarity on 'Progressing' state. Questions like 'Do we want to populate a message when progressing=false?', etc. need to be answered.

Yeah, we need to get some answers, and then based on those answers, we can do any necessary work in follow-up PRs.

pravisankar:
@Miciah @ironcladlou @danehans another attempt, PTAL

pkg/operator/controller/status.go
```diff
@@ -40,88 +47,167 @@ func (r *reconciler) syncOperatorStatus() error {
...
+		}
+	}
+
+	releaseVersion := r.OperatorReleaseVersion
+	ingressControllerImage := r.IngressControllerImage
+	// We need to report old version until the operator fully transitions to the new version.
+	// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version-reporting-during-an-upgrade
+	if !allIngressesAvailable {
+		releaseVersion = oldReleaseVersion
+		ingressControllerImage = oldIngressControllerImage
```
Miciah: This can be simplified to `return oldVersions` (and we don't need to compute `oldReleaseVersion`, `oldIngressControllerImage`, `releaseVersion`, or `ingressControllerImage`).

pkg/operator/controller/status.go
```diff
@@ -40,88 +47,167 @@ func (r *reconciler) syncOperatorStatus() error {
...
+			oldAvailableCondition = &oldConditions[i]
+		}
+	}
+
+	conditions := []configv1.ClusterOperatorStatusCondition{}
+	conditions = append(conditions, computeOperatorDegradedCondition(oldDegradedCondition, ns))
+	conditions = append(conditions, r.computeOperatorProgressingCondition(oldProgressingCondition,
+		allIngressesAvailable, curVersions))
+	conditions = append(conditions, computeOperatorAvailableCondition(oldAvailableCondition,
+		allIngressesAvailable))
```
Miciah: What do you think of the following?
```go
	conditions := []configv1.ClusterOperatorStatusCondition{ 
		computeOperatorDegradedCondition(oldDegradedCondition, ns),
		r.computeOperatorProgressingCondition(oldProgressingCondition, allIngressesAvailable, curVersions),
		computeOperatorAvailableCondition(oldAvailableCondition, allIngressesAvailable),
	}
```

pkg/operator/controller/status.go
```diff
@@ -132,100 +218,79 @@ func computeOperatorDegradedCondition(conditions []configv1.ClusterOperatorStatu
...
+func setLastTransitionTime(condition, oldCondition *configv1.ClusterOperatorStatusCondition) configv1.ClusterOperatorStatusCondition {
+	if oldCondition != nil && condition.Status == oldCondition.Status &&
+		condition.Reason == oldCondition.Reason && condition.Message == oldCondition.Message {
+		condition.LastTransitionTime = oldCondition.LastTransitionTime
+	} else {
+		condition.LastTransitionTime = metav1.Now()
 	}
 
-	return newConditions
+	return *condition
```
Miciah: We can get this in follow-up if you prefer, but it is strange that `setLastTransitionTime` mutates `condition` and then returns a value (i.e., a copy).  It would make more sense for `setLastTransitionTime` to either copy and then mutate that or not to return a copy at all.

pravisankar:
@Miciah updated, PTAL

PR #201, opened 2019-04-10
Minor Operator Status Refactor

danehans:
- Checks for `RELEASE_VERSION` env var and sets it to "unknown" if missing.
- Updates status code to stop checking for missing `RELEASE_VERSION`
- Decomposes `computeOperatorStatusConditions()` so individual conditions can be computed.

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -20,7 +20,10 @@ import (
 )
 
 const (
-	IngressClusterOperatorName = "ingress"
+	IngressClusterOperatorName     = "ingress"
+	UnknownReleaseVersionName      = "unknown"
+	ingressesEqualConditionMessage = "desired and current number of IngressControllers are equal"
+	operatorVersionName            = "operator"
 )
 
 // syncOperatorStatus computes the operator's current status and therefrom
@@ -56,20 +59,18 @@ func (r *reconciler) syncOperatorStatus() error {
 		},
 	}
 
-	if len(r.OperatorReleaseVersion) > 0 {
-		// An available operator resets release version
-		for _, condition := range co.Status.Conditions {
-			if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue {
-				co.Status.Versions = []configv1.OperandVersion{
-					{
-						Name:    "operator",
-						Version: r.OperatorReleaseVersion,
-					},
-					{
-						Name:    "ingress-controller",
-						Version: r.RouterImage,
-					},
-				}
+	// An available operator resets release version
+	for _, condition := range co.Status.Conditions {
+		if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue {
+			co.Status.Versions = []configv1.OperandVersion{
+				{
+					Name:    operatorVersionName,
+					Version: r.OperatorReleaseVersion,
+				},
+				{
+					Name:    "ingress-controller",
+					Version: r.RouterImage,
+				},
 			}
 		}
 	}
@@ -107,7 +108,17 @@ func (r *reconciler) getOperatorState() (*corev1.Namespace, []operatorv1.Ingress
 }
 
 // computeOperatorStatusConditions computes the operator's current state.
-func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace, ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace,
+	ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
+	conditions = computeOperatorFailingCondition(conditions, ns)
+	conditions = computeOperatorProgressingCondition(conditions, ingresses)
+	conditions = computeOperatorAvailableCondition(conditions, ingresses)
+
+	return conditions
+}
+
+// computeOperatorFailingCondition computes the operator's current Failing status state.
+func computeOperatorFailingCondition(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace) []configv1.ClusterOperatorStatusCondition {
 	failingCondition := &configv1.ClusterOperatorStatusCondition{
 		Type:   configv1.OperatorFailing,
 		Status: configv1.ConditionUnknown,
@@ -118,9 +129,14 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 		failingCondition.Message = "operand namespace does not exist"
 	} else {
 		failingCondition.Status = configv1.ConditionFalse
+		failingCondition.Message = "operand namespace exists"
 	}
-	conditions = setOperatorStatusCondition(conditions, failingCondition)
 
+	return setOperatorStatusCondition(conditions, failingCondition)
+}
+
+// computeOperatorProgressingCondition computes the operator's current Progressing status state.
+func computeOperatorProgressingCondition(conditions []configv1.ClusterOperatorStatusCondition, ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
 	// TODO: Update progressingCondition when an ingresscontroller
 	//       progressing condition is created. The Operator's condition
 	//       should be derived from the ingresscontroller's condition.
@@ -140,6 +156,7 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 	}
 	if numIngresses == ingressesAvailable {
 		progressingCondition.Status = configv1.ConditionFalse
+		progressingCondition.Message = ingressesEqualConditionMessage
 	} else {
 		progressingCondition.Status = configv1.ConditionTrue
 		progressingCondition.Reason = "Reconciling"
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/status.go
```diff
@@ -148,6 +151,11 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 			"%d ingress controllers available, want %d",
 			ingressesAvailable, numIngresses)
 	}
+	// LastTransitionTime is updated when running and configured
+	// versions are not equal to indicate an upgrade has occurred.
+	if !versionsEqual {
+		progressingCondition.LastTransitionTime = metav1.Now()
```
ironcladlou: Would it make any sense to set a reason (e.g. VersionMismatch) and a message (e.g. "From version X to version Y")?

Miciah: Yeah, better to set a message than to fiddle with last transition time.

danehans: CIO needs to reset `LastTransitionTime` to fix [Bug 1695210](https://bugzilla.redhat.com/show_bug.cgi?id=1695210). After talking with @pravisankar yesterday, we thought this requirement should not trigger a change to the Operator's `Progressing` condition since ingress controllers are not being affected by the Operator upgrade. According to the bug, changing the `Progressing` condition is not required, but updating `LastTransitionTime` is. Thoughts?

Miciah: Updating `Message` using `setOperatorStatusCondition` will update `LastTransitionTime` and should cause `operatorStatusesEqual` to consider the status unequal so `syncOperatorStatus` will update it.

danehans: Ahh... I did not know that. Thanks for the heads-up. I'll update and resubmit the PR.

pkg/operator/controller/status.go
```diff
@@ -148,6 +151,9 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 			"%d ingress controllers available, want %d",
 			ingressesAvailable, numIngresses)
 	}
+	if !versionsEqual {
+		progressingCondition.Message = "configured and current release versions do not match"
```
Miciah: In the case of two successive upgrades, if nothing else updates the `Progressing` condition in between upgrades, it will still have the message from the first upgrade when the second upgrade happens.  We should make sure the message changes between upgrades (simplest way would be to include the version string in the condition's message).

pkg/operator/controller/status_test.go
```diff
@@ -71,14 +151,18 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
...
 			{
 				Type:   configv1.OperatorFailing,
 				Status: failing,
 			},
 			{
-				Type:   configv1.OperatorProgressing,
-				Status: progressing,
+				Type:               configv1.OperatorProgressing,
+				Status:             progressing,
+				LastTransitionTime: transitionTime,
```
Miciah: We don't look at `LastTransitionTime` when computing `gotExpected` below.  I think we need to add something like the following to the condition comparison loop below:
```go
					if conditionA.Type == configv1.OperatorProgressing && tc.outputs.transition && conditionA.LastTransitionTime.IsZero() {
						t.Errorf("%q: expected transition time to be updated", tc.description)
					}
```

danehans: Thanks for catching this. I'll update the test.

pkg/operator/controller/status.go
danehans: Good point. I'll pass the configured version into `computeOperatorStatusConditions()` and include it in the message string.

pkg/operator/controller/status.go
```diff
@@ -148,6 +151,10 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
 			"%d ingress controllers available, want %d",
 			ingressesAvailable, numIngresses)
 	}
+	if !versionsEqual {
+		progressingCondition.Message = fmt.Sprintf(
+			"current release version does not match configured release version: %s", version)
```
Miciah: Consider the context in which this message is read.  As soon as the condition has been updated, this becomes past tense ("did not match"), right? Alternatively, I believe `Upgraded to %s` would be sufficient.

pkg/operator/controller/status.go
```diff
@@ -108,7 +110,8 @@ func (r *reconciler) getOperatorState() (*corev1.Namespace, []operatorv1.Ingress
 }
 
 // computeOperatorStatusConditions computes the operator's current state.
-func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace, ingresses []operatorv1.IngressController) []configv1.ClusterOperatorStatusCondition {
+func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatusCondition, ns *corev1.Namespace,
+	ingresses []operatorv1.IngressController, versionsEqual bool, version string) []configv1.ClusterOperatorStatusCondition {
```
Miciah: What do you think about making the old and new version inputs to `computeOperatorStatusConditions` and letting it do the comparison?

pkg/operator/controller/status_test.go
```diff
@@ -71,14 +158,18 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
...
```
Miciah: Looks like this line can be deleted now.

pkg/operator/controller/status_test.go
Miciah: This might be trickier than I originally thought... If the condition does not exist in the conditions slice that the test passes to `computeOperatorStatusConditions`, then `computeOperatorStatusConditions` will _always_ set it, so it will _always_ have have a non-zero `LastTransitionTime`.  So we need to have at least a `Progressing` condition already set in the input to `computeOperatorStatusConditions`.  I'm starting to think it would be worth adding a separate test function for testing that `Progressing` gets updated on upgrades.  What do you think?

pkg/operator/controller/status.go
danehans: Good point. I'll update.

> Alternatively, I believe Upgraded to %s would be sufficient.

This code checks for a diff between configured and current release versions and updates the `Progressing` message to trigger a new `LastTransitionTime`. As I mentioned in https://github.com/openshift/cluster-ingress-operator/pull/200#issuecomment-482752004, I believe that PR should be updated to include a `Progressing` status condition message whenever a status version is updated.

pkg/operator/controller/status_test.go
danehans: I'm in the process of refactoring the test. Since it's expected that updating a condition message triggers `LastTransitionTime`, I will create a separate test that checks the output message when versions are equal and not equal.

danehans:
In the latest commit I added the following:

- A check for the `RELEASE_VERSION` env variable.
- I split out computation for individual status conditions so https://github.com/openshift/cluster-ingress-operator/pull/200 can compute the `Progressing` condition after computing status version information.

cc @pravisankar

cmd/cluster-ingress-operator/main.go
```diff
@@ -51,8 +52,8 @@ func main() {
 	}
 
 	// Collect operator configuration.
-	operatorNamespace, ok := os.LookupEnv("WATCH_NAMESPACE")
-	if !ok {
+	operatorNamespace := os.Getenv("WATCH_NAMESPACE")
```
danehans: @Miciah I updated ^ to provide consistency with how CIO checks other env vars.

cmd/cluster-ingress-operator/main.go
```diff
@@ -51,8 +52,8 @@ func main() {
...
+	if len(operatorNamespace) == 0 {
```
danehans: @Miciah I updated ^ to provide consistency with how CIO checks other env vars.

pkg/operator/controller/status.go
```diff
@@ -51,20 +58,18 @@ func (r *reconciler) syncOperatorStatus() error {
 		},
 	}
 
-	if len(r.OperatorReleaseVersion) > 0 {
```
danehans: Removed ^ since `r.OperatorReleaseVersion` is now set to "unknown" if env var is not set.

pkg/operator/controller/status.go
```diff
@@ -38,8 +40,14 @@ func (r *reconciler) syncOperatorStatus() error {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
 
+	currentVersion := "unknown"
+	if co.Status.Versions != nil {
+		if len(co.Status.Versions[0].Version) != 0 {
+			currentVersion = co.Status.Versions[0].Version
```
Miciah: It would be better not to assume index position; instead, do a `for version := range co.Status.Versions` and look for one with `version.Name == "operator"`.

pkg/operator/controller/status.go
```diff
@@ -141,31 +162,52 @@ func computeOperatorStatusConditions(conditions []configv1.ClusterOperatorStatus
...
 		progressingCondition.Status = configv1.ConditionTrue
 		progressingCondition.Reason = "Reconciling"
 		progressingCondition.Message = fmt.Sprintf(
 			"%d ingress controllers available, want %d",
 			ingressesAvailable, numIngresses)
 	}
-	conditions = setOperatorStatusCondition(conditions, progressingCondition)
+	if !strings.EqualFold(configVersion, currentVersion) {
+		progressingCondition.Message = fmt.Sprintf(
+			"configurd release version %q doesn't match current version %q",
```
Miciah: Typo: "configurd"→"configured".  I think saying "doesn't match current version" could be confusing, because once we update the clusteroperator status, the "current version" is now the configured version, right? So I think the message will be confusing in the context in which the administrator reads it.

danehans: The intent is for @pravisankar to update https://github.com/openshift/cluster-ingress-operator/pull/200 to call `computeOperatorProgressingCondition()` after updating the operator status version, which would remove the `Progressing` condition message.

Another option is for this PR to remove the Progressing status message related to version inequality and have Ravi's PR set an upgrade-related message when computing the operator status version. Thoughts?

Miciah: > The intent is for @pravisankar to update #200 to call `computeOperatorProgressingCondition()` after updating the operator status version, which would remove the `Progressing` condition message in L175.

Would this all be happening in a single call to `syncOperatorStatus` (i.e., `syncOperatorStatus` would call `computeOperatorProgressingCondition`, then call `computeOperatorStatusVersions`, and then call `computeOperatorProgressingCondition` a second time), or would the first call to `syncOperatorStatus` set `Progressing` with the "doesn't match" message and a subsequent call to `syncOperatorStatus` update `.status.versions` and remove "doesn't match" message?

Are we only updating the message as a way to get `setOperatorStatusCondition` to update the condition's last transition, or will the message ever actually be updated in the API object?

Calling `computeOperatorProgressingCondition` twice in the same call to `syncOperatorStatus` seems non-obvious, and the result is that the message never actually gets published to the API, right?

> Another option is for this PR to remove the Progressing status message related to version inequality and have Ravi's PR set an upgrade-related message when computing the operator status version. Thoughts?

This seems closer to what I thought your intent was here: When we detect that the operator has been updated, immediately update the `Progressing` condition with a message (and, implicitly, a last transition time) that convey that an upgrade has taken place.

pkg/operator/controller/status.go
danehans: Done

pkg/operator/controller/status.go
danehans: > Would this all be happening in a single call to syncOperatorStatus (i.e., syncOperatorStatus would call computeOperatorProgressingCondition, then call computeOperatorStatusVersions, and then call computeOperatorProgressingCondition a second time), or would the first call to syncOperatorStatus set Progressing with the "doesn't match" message and a subsequent call to syncOperatorStatus update .status.versions and remove "doesn't match" message?

The call flow as you describe is correct. `computeOperatorStatusVersions` requires the status condition to be computed before computing the version. If `computeOperatorStatusVersions` makes any changes to status versions, it should then update `Progressing` to indicate "something" happened. I split `computeOperatorStatusConditions` into separate functions to provide a bit more efficiency.

danehans:
@pravisankar I have incorporated your feedback in the latest PR. I will assume PR #200 will address Bug 1695210

PR #202, opened 2019-04-12
Adds Toleration Support

danehans:
Previously, `NodeSelector` was the only supported mechanism for scheduling pods to nodes. This PR adds support for `Tolerations`. When combined with `Taints`, `Tolerations` ensure that pods are not scheduled onto inappropriate nodes.


pkg/operator/controller/controller_router_deployment_test.go
```diff
@@ -201,6 +209,11 @@ func TestDesiredRouterDeployment(t *testing.T) {
 		t.Errorf("router Deployment has unexpected node selector: %#v",
 			deployment.Spec.Template.Spec.NodeSelector)
 	}
+	if len(deployment.Spec.Template.Spec.Tolerations) != 1 ||
+		deployment.Spec.Template.Spec.Tolerations[0] != ci.Spec.NodePlacement.Tolerations[0] {
+		t.Errorf("router Deployment has unexpected toleration: %#v",
+			deployment.Spec.Template.Spec.Tolerations[0])
```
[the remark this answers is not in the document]
danehans: > We should probably have a check earlier, before we set ci.Spec.NodePlacement, to make sure that len(deployment.Spec.Template.Spec.Tolerations) is 0 in that case.

Done
> Inside this if for this check, we know that len(deployment.Spec.Template.Spec.Tolerations) is not 1

I'm not following... `len(deployment.Spec.Template.Spec.Tolerations)` is `1` since it was set earlier in the test: `Tolerations: []corev1.Toleration{toleration},`. Am I missing something in your comment?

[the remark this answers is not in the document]
Miciah: > I'm not following... `len(deployment.Spec.Template.Spec.Tolerations)` is `1` since it was set earlier in the test: `Tolerations: []corev1.Toleration{toleration},`. Am I missing something in your comment?

We set `ci.Spec.NodePlacement.Tolerations`.  Now we are checking what `desiredRouterDeployment` did with that input.  The `if` here is checking that `desiredRouterDeployment` copied `ci.Spec.NodePlacement.Tolerations` to `deployment.Spec.Template.Spec.Tolerations`.  If not, then we need to print an error message, but that error message cannot make assumptions about what `desiredRouterDeployment` actually assigned to `deployment.Spec.Template.Spec.Tolerations`.

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -290,8 +303,9 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 	return true, updated
 }
 
-func cmpEnvs(a, b corev1.EnvVar) bool    { return a.Name < b.Name }
-func cmpVolumes(a, b corev1.Volume) bool { return a.Name < b.Name }
+func cmpEnvs(a, b corev1.EnvVar) bool            { return a.Name < b.Name }
+func cmpTolerations(a, b corev1.Toleration) bool { return a.Key < b.Key }
```
Miciah: I think we need to compare all the fields because you could have two tolerations with the matching `Key` but non-matching other fields.  For example, as far as I can tell, it is possible (even if silly) to have a toleration with key `x` and operator `Exists` and another toleration with the same key but with operator `Equals`, or you could have two tolerations with the same key and operator but different effects.

pkg/operator/controller/controller_router_deployment_test.go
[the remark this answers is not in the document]
danehans: I understand now. Thank you for the detailed explanation.

pkg/operator/controller/controller_router_deployment.go
danehans: Done.

danehans: Done.

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -315,3 +328,25 @@ func cmpSecretVolumeSource(a, b corev1.SecretVolumeSource) bool {
...
+	}
+	if a.Operator != b.Operator {
+		return false
+	}
+	if a.Effect != b.Effect {
+		return false
+	}
+	if a.Effect == corev1.TaintEffectNoExecute {
+		// Field is ignored unless effect is NoExecute.
+		if *a.TolerationSeconds != *b.TolerationSeconds {
```
Miciah: Make sure to check for nil values.

Miciah:
level=error msg="1 error occurred:"
    level=error msg="\t* module.vpc.aws_route.to_nat_gw[2]: 1 error occurred:"
    level=error msg="\t* aws_route.to_nat_gw.2: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"

/test e2e-aws

danehans:
```
Failed to get secret openshift-kube-apiserver/serving-cert: secrets \"serving-cert\" not found
```
/test e2e-aws

pkg/operator/controller/controller_router_deployment.go
```diff
@@ -264,10 +267,17 @@ func (r *reconciler) updateRouterDeployment(current, desired *appsv1.Deployment)
 // deploymentConfigChanged checks if current config matches the expected config
 // for the cluster ingress deployment and if not returns the updated config.
 func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	tolerationsEqual := true
+	if expected.Spec.Template.Spec.Tolerations != nil {
```
Miciah: Do we need this `if`? `cmp.Equal` should handle nil values correctly (and `cmpopts.EquateEmpty()` should make it treat nil and an empty map as equal).

If the `if` can be deleted, we can do it in a follow-up—it's a minor issue, and we need to get this PR merged.

danehans: Let me test without the `cmp.Equal` wrapped in the `if` block. I was having problems with a unit test panicing and I thought this was the answer.

PR #204, opened 2019-04-12
Bug 1699488: Fixes Ingress ClusterOperator Initial Status

danehans:
Previously, `clusteroperator/ingress` was being created without `status`. Even though the resource was being created with status fields properly constructed and included in the `client.create` call, the resource is initially created without `status`. `clusteroperator/ingress` must be created first, then have the status updated.


Miciah:
/lgtm
Same problem exists in cluster-dns-operator; would you like to fix it, or shall I?

danehans:
@Miciah I'll fix in the dns operator too.

danehans:
https://github.com/openshift/cluster-ingress-operator/pull/204 fixes this bug in cluster-dns-operator.

PR #212, opened 2019-04-19
record log into pod status on error

deads2k:
Getting a snippet of the logs is super useful in cases where you don't have logs.  Sure, you should always have logs, but when you don't this is much better than nothing.


smarterclayton:
Approved to help debug (all operator should be using this)

/lgtm
/approve

PR #214, opened 2019-04-19
start-build.sh: Don't patch operator IMAGE env var

Miciah:
Do not override the `IMAGE` environment variable when patching the operator deployment.  `IMAGE` specifies the operand image, and the patch should only change the operator image.

* `hack/start-build.sh`: Delete the `IMAGE` environment variable from the patch to the operator deployment.

### DIFF
--- a/hack/start-build.sh
+++ b/hack/start-build.sh
@@ -16,12 +16,6 @@ then
         "containers": [
           {
             "name": "ingress-operator",
-            "env": [
-              {
-                "name": "IMAGE",
-                "value": "image-registry.openshift-image-registry.svc:5000/openshift-ingress-operator/ingress-operator:latest"
-              }
-            ],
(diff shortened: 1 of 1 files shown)


hack/start-build.sh
```diff
@@ -16,12 +16,6 @@ then
         "containers": [
           {
             "name": "ingress-operator",
-            "env": [
-              {
-                "name": "IMAGE",
-                "value": "image-registry.openshift-image-registry.svc:5000/openshift-ingress-operator/ingress-operator:latest"
-              }
-            ],
```
pravisankar: Don't we need to keep 'IMAGE' env variable to update haproxy version?
I think we may need IMAGE env but the value should point to haproxy instead of ingress-operator image.

Miciah: The buildconfig only rebuilds the operator image, so `start-build.sh` should leave the existing `IMAGE` variable alone.

PR #215, opened 2019-04-22
[WIP] Use CVO to create the default ingress controller

Miciah:
* `manifests/01-ingress-controller.yaml`: New file.
* `pkg/operator/operator.go` (`Start`): Delete goroutine for ensuring the `default` ingress controller exists.
(`ensureDefaultIngressController`): Delete method.

### DIFF
--- a/manifests/01-ingress-controller.yaml
+++ b/manifests/01-ingress-controller.yaml
@@ -1,0 +1,7 @@
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: default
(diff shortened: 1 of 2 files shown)


ironcladlou:
What's the motivation?

If the default is deleted, how can the admin ever restore the default?

Related: should it be possible to delete the default?

ironcladlou:
I get the sense by now we're not going to cede control over this resource to the CVO. Want to close it?

Miciah:
All right.

PR #218, opened 2019-04-26
BUG 1707477: label the openshift-ingress namespace with its name

squeed:
This will let us select for it with network policies.


ironcladlou:
Thanks!
/lgtm

danwinship:
/hold
Would it be better to use a label that more explicitly indicates that this is for selecting via NetworkPolicy?
If not there should at least be a comment there indicating that that label is effectively API

squeed:
This is the way we've done it in 3.x, but, you're right - now's a reasonable chance to change that.

@ironcladlou - thoughts? How does this match with the expected pattern of ingress deployment? Will users be able to deploy the ingress controller in to other namespaces, and, if so, how would that impact this change?

Alternatively, just calling it `SOMETHING.openshift.io/has-router` would be good too.

ironcladlou:
>Would it be better to use a label that more explicitly indicates that this is for selecting via NetworkPolicy?
If not there should at least be a comment there indicating that that label is effectively API

cc @smarterclayton @derekwaynecarr 

>@ironcladlou - thoughts? How does this match with the expected pattern of ingress deployment? Will users be able to deploy the ingress controller in to other namespaces, and, if so, how would that impact this change?

We have no concrete plans to allow ingress controllers in other namespaces, but we've been trying to leave that door open.

>Alternatively, just calling it SOMETHING.openshift.io/has-router would be good too.

Another option... `ingress.openshift.io/has-ingress-controller`

Miciah:
> Another option... `ingress.openshift.io/has-ingress-controller`

We have no `ingress.openshift.io` API group, but rather put this operator's resource kinds in the `operator.openshift.io` API group, so `operator.openshift.io/has-ingress-controller` would make more sense.

danwinship:
That seems kind of indirect.
I don't think there's any rule that says that label names should only use actual existing API groups.
But anyway, since the label exists for networking purposes, it could be `network.openshift.io/policy-group: ingress` or something

squeed:
Policy-group seems reasonable. Or `network.openshift.io/role`.

Are there any other namespaces that would qualify? `role: apiserver`?

danwinship:
anyway, we should commit *something* before freeze...

squeed:
Oh crap, good point. I like policy-group.

squeed:
Updated with `policy-group. I kept name as well, since users may wish to select by it.

ironcladlou:
/lgtm
@squeed @danwinship feel free to remove the hold when you're satisfied with the name

squeed:
aws timeout
/retest

smarterclayton:
Title needed to be “Bug XXXX:”, not “BUG XXXX:”

PR #219, opened 2019-04-26
Bug 1701392: Choose an appropriate default deployment strategy for an endpoint publishing strategy

ironcladlou:
Choose an appropriate default deployment strategy for an endpoint publishing
strategy.

The default is a rolling update with 25% max unavailable and 25% max surge. This
enables 100% availability rollouts at any replica/node count when the publishing
strategy supports colocated ingress controllers.

With HostNetwork, use 25% max unavailable and 0 surge. This is because the
ingress controllers can't be colocated. Usually, replicas will be equal to the
node pool size. Under these conditions, surge requires new nodes to support the
rollout. This means a positive surge can cause the rollout to wedge in the
absence of auto-scaling.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1701392

### DIFF
--- a/pkg/operator/controller/controller_router_deployment.go
+++ b/pkg/operator/controller/controller_router_deployment.go
@@ -16,6 +16,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	configv1 "github.com/openshift/api/config/v1"
 )
@@ -97,6 +98,8 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, routerImage strin
 			},
 		},
 	}
+
+	deployment.Spec.Strategy = deploymentStrategyForPublishingStrategy(ci.Status.EndpointPublishingStrategy)
 
 	statsSecretName := fmt.Sprintf("router-stats-%s", ci.Name)
 	env := []corev1.EnvVar{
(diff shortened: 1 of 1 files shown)


ironcladlou:
Do we want an e2e for this? It would be:

1. Create HostNetwork ingresscontroller w/ replicas 3
2. Modify the ingresscontroller to trigger a rollout (e.g. edit route selector)
3. Ensure new deployment generation becomes 100% available

Miciah:
> With HostNetwork, use 25% max unavailable and 0 surge. This is because the
> ingress controllers can't be colocated. Usually, replicas will be equal to the
> node pool size. Under these conditions, surge requires new nodes to support the
> rollout. This means a positive surge can cause the rollout to wedge in the
> absence of auto-scaling.

Does this solve the problem when the node pool size is less than 4? In that case, 25% of the node pool size will be rounded down to 0, so you have 0 max unavailable and 0 max surge, which is invalid and precludes progress.

ironcladlou:
>Does this solve the problem when the node pool size is less than 4? In that case, 25% of the node pool size will be rounded down to 0, so you have 0 max unavailable and 0 max surge, which is invalid and precludes progress.

Yes.

Miciah:
Ah, you explain in https://bugzilla.redhat.com/show_bug.cgi?id=1701392#c7:

> In 3.x we're setting surge to 0, which triggers this fencepost condition to set a floor of 1 for unavailability even when the spec value is proportional.

Relying on undocumented behavior makes me nervous.  Shall I open a PR on openshift/api to document the behavior for `MaxSurge`?

Miciah:
It would be worth explaining the behavior when `MaxSurge` is 0 and `MaxUnavailable` rounds to 0 in a comment on `deploymentStrategyForPublishingStrategy`.

Miciah:
> Shall I open a PR on openshift/api to document the behavior for `MaxSurge`?

Er, I was looking at the `DeploymentConfig` APIs.  The `Deployment` API has the same documentation issue, but it is not in openshift/api.

PR #220, opened 2019-04-29
Minor: Move service monitor name to pkg/operator/controller/names.go

pravisankar:


Miciah:
/lgtm
/hold
because AIUI we're holding PRs that don't have associated Bugzilla reports.

ironcladlou:
Will release hold when the master is open for 4.2

PR #221, opened 2019-04-30
WIP: Adds e2e for Deployment update strategy

danehans:
Adds an e2e for [this PR](https://github.com/openshift/cluster-ingress-operator/commit/fb9851af20b197f8f64ac68e0f643b397e62d695) from @ironcladlou 

The test is not passing due to this condition:
```
if deployment.Status.UnavailableReplicas > int32(0) {
	return false, nil
}
```

Nothing in the operator logs or watching events jumps out at me as an issue.

This PR does not use @ironcladlou's PR, so a replica count of 1 or 2 should pass and 3 should fail.  I can see the ingress controller and associated deployment get created and updated with the `foo: baz` routeSelector.  During the rolling update, I see `observedGeneration: 2` and `UnavailableReplicas: 1` until the new pods are running, then `UnavailableReplicas` no longer appears when I `$ oc get deploy/router-test -n openshift-ingress -o yaml`. However, the test is failing: 
```
operator_test.go:767: failed to reconcile updated deployment: timed out waiting for the condition
```

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -59,8 +59,8 @@ func getClient() (client.Client, string, error) {
 	return kubeClient, namespace, nil
 }
 
-func newIngressController(name, ns, domain string, epType operatorv1.EndpointPublishingStrategyType) *operatorv1.IngressController {
-	repl := int32(1)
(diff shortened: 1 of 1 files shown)


danehans:
Deployment surge=0  will be used for all endpoint strategy types, so this e2e is no longer needed.

PR #224, opened 2019-05-01
Expands IngressController status conditions

danehans:
- Refactors the `IngressController` `Available` condition to be based on the `Available` status of `IngressController` dependent resources (i.e. `DNS`).
- Adds a `Deployment` status condition.
- Adds unit tests.
- Updates e2e tests.

### DIFF
--- a/pkg/operator/controller/ingress_status.go
+++ b/pkg/operator/controller/ingress_status.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -18,6 +19,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// DeploymentAvailableConditionType is a status condition type indicating
+	// an IngressController Deployment is available.
+	DeploymentAvailableConditionType = "DeploymentAvailable"
+)
+
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
 func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) error {
@@ -31,9 +38,10 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.Selector = selector.String()
 
 	updated.Status.Conditions = []operatorv1.OperatorCondition{}
-	updated.Status.Conditions = append(updated.Status.Conditions, computeIngressStatusConditions(updated.Status.Conditions, deployment)...)
+	updated.Status.Conditions = append(updated.Status.Conditions, computeDeploymentStatus(deployment))
 	updated.Status.Conditions = append(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
 	updated.Status.Conditions = append(updated.Status.Conditions, computeDNSStatus(ic, wildcardRecord, dnsConfig)...)
+	updated.Status.Conditions = append(updated.Status.Conditions, computeIngressStatusConditions(ic, updated.Status.Conditions)...)
 
 	for i := range updated.Status.Conditions {
 		newCondition := &updated.Status.Conditions[i]
@@ -58,29 +66,50 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 }
 
 // computeIngressStatusConditions computes the ingress controller's current state.
-func computeIngressStatusConditions(oldConditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment) []operatorv1.OperatorCondition {
-	oldAvailableCondition := getIngressAvailableCondition(oldConditions)
-
+func computeIngressStatusConditions(ic *operatorv1.IngressController, conditions []operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
 	return []operatorv1.OperatorCondition{
-		computeIngressAvailableCondition(oldAvailableCondition, deployment),
+		computeIngressAvailableCondition(ic, conditions),
 	}
 }
 
 // computeIngressAvailableCondition computes the ingress controller's current Available status state.
-func computeIngressAvailableCondition(oldAvailableCondition *operatorv1.OperatorCondition, deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	availableCondition := operatorv1.OperatorCondition{
+func computeIngressAvailableCondition(ic *operatorv1.IngressController, conditions []operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+	condition := operatorv1.OperatorCondition{
 		Type: operatorv1.IngressControllerAvailableConditionType,
 	}
 
-	if deployment.Status.AvailableReplicas > 0 {
-		availableCondition.Status = operatorv1.ConditionTrue
-	} else {
-		availableCondition.Status = operatorv1.ConditionFalse
-		availableCondition.Reason = "DeploymentUnavailable"
-		availableCondition.Message = "no Deployment replicas available"
+	messages := []string{}
+	for _, c := range conditions {
+		if c.Status == operatorv1.ConditionFalse && (c.Type == DeploymentAvailableConditionType ||
+			c.Type == operatorv1.LoadBalancerReadyIngressConditionType ||
+			c.Type == operatorv1.DNSReadyIngressConditionType) {
+			messages = append(messages, c.Message)
+		}
+	}
+	var message string
+	switch {
+	case len(messages) == 1:
+		message = messages[0]
+	case len(messages) > 1:
+		message = strings.Join(messages, ". ")
 	}
 
-	return availableCondition
+	switch {
+	case len(ic.Status.Domain) == 0:
+		condition.Status = operatorv1.ConditionFalse
+		condition.Reason = "StatusDomainUnset"
+		condition.Message = "The IngressController status domain is not set"
+	case len(messages) != 0:
+		condition.Status = operatorv1.ConditionFalse
+		condition.Reason = "DependantResourceFailure"
+		condition.Message = message
+	default:
+		condition.Status = operatorv1.ConditionTrue
+		condition.Reason = "AsExpected"
+		condition.Message = "All IngressControllers dependant resources are available"
+	}
+
+	return condition
 }
 
 // getIngressAvailableCondition fetches ingress controller's available condition from the given conditions.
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/controller.go
```diff
@@ -195,13 +196,11 @@ func (r *reconciler) enforceEffectiveIngressDomain(ic *operatorv1.IngressControl
...
-		availableCondition := &operatorv1.OperatorCondition{
-			Type:    operatorv1.IngressControllerAvailableConditionType,
-			Status:  operatorv1.ConditionFalse,
-			Reason:  "InvalidDomain",
-			Message: fmt.Sprintf("domain %q is already in use by another IngressController", domain),
+		var conditions []operatorv1.OperatorCondition
+		conditions = computeIngressStatusConditions(conditions, &appsv1.Deployment{}, false)
+		for _, c := range conditions {
+			updated.Status.Conditions = append(updated.Status.Conditions, c)
 		}
```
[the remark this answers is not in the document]
danehans: > Alternatively, would it make sense to call syncIngressControllerStatus instead?

Thanks for the review. I was thinking about using `syncIngressControllerStatus`. However, it performs a status update,  as does `enforceEffectiveIngressDomain` and I thought multiple status update calls would be suboptimal. With the change recommendations that you outline, using ``syncIngressControllerStatus` sounds sensible. I'll start working through the changes after I finish a doc bug that I'm currently working.

Miciah:
/test e2e-aws-operator
I want to see whether [BZ#1705100](https://bugzilla.redhat.com/show_bug.cgi?id=1705100) shows up again.

Miciah:
Still not seeing a new run since yesterday.
/test e2e-aws-operator

ironcladlou:
Need to determine whether these are necessary for 4.1

/hold

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
danehans: @Miciah I implemented some of your suggestions above. I keep hitting e2e failures when I move the domain uniqueness logic from `enforceEffectiveIngressDomain` to `syncIngressControllerStatus`:

```
2019-05-03T11:43:31.721-0700	INFO	operator	log/log.go:26	started zapr logger
=== RUN   TestCreateIngressControllerThenSecret
--- FAIL: TestCreateIngressControllerThenSecret (32.82s)
    certificate_publisher_test.go:69: failed to observe reconciliation of ingresscontroller: timed out waiting for the condition
=== RUN   TestCreateSecretThenIngressController
--- FAIL: TestCreateSecretThenIngressController (62.82s)
    certificate_publisher_test.go:160: failed to observe updated global secret: timed out waiting for the condition
=== RUN   TestOperatorAvailable
--- FAIL: TestOperatorAvailable (12.57s)
    operator_test.go:102: did not get expected available condition: timed out waiting for the condition
=== RUN   TestDefaultIngressControllerExists
--- PASS: TestDefaultIngressControllerExists (2.48s)
=== RUN   TestIngressControllerControllerCreateDelete
--- FAIL: TestIngressControllerControllerCreateDelete (62.80s)
    operator_test.go:159: failed to reconcile IngressController openshift-ingress-operator/test: timed out waiting for the condition
```

pkg/operator/controller/ingress_status.go
```diff
@@ -35,26 +36,123 @@ func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment,
 }
 
 // computeIngressStatusConditions computes the ingress controller's current state.
-func computeIngressStatusConditions(oldConditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment) []operatorv1.OperatorCondition {
+func computeIngressStatusConditions(oldConditions []operatorv1.OperatorCondition, deployment *appsv1.Deployment,
+	uniqueDomain bool) []operatorv1.OperatorCondition {
+	oldDegradedCondition := getIngressDegradedCondition(oldConditions)
+	oldProgressingCondition := getIngressProgressingCondition(oldConditions)
 	oldAvailableCondition := getIngressAvailableCondition(oldConditions)
```
Miciah: I don't know that having separate `getIngressDegradedCondition`, `getIngressProgressingCondition`, and `getIngressAvailableCondition` functions is important for readability, and it means more code and more looping.  What do you think of using a single loop in `computeIngressStatusConditions` to get all three values, similar to how the DNS operator does it? https://github.com/openshift/cluster-dns-operator/blob/540ab8bca50b50880a4eb44feaddee8352f565bf/pkg/operator/controller/status.go#L202-L212

danehans: @Miciah I was going back and forth on whether to have separate functions for each condition or looping through the conditions in a single function. I thought the former would be easier to read and understand. I will update the PR using a single function.

danehans: >  Do these changes seem reasonable, and if so, could we incorporate them into this PR, or would it be better to put them in a separate PR?

They do make sense. I'm working on updating the PR.

Miciah:
Doesn't `syncIngressControllerStatus` need to get the deployment if the status domain was not set and `syncIngressControllerStatus` was able to set it?

danehans:
I don't believe so. If `syncIngressControllerStatus()` was called with an `IngressController` that did not have `status.domain` set, that means all the `ensure/enforce` methods of the main reconcile loop were bypassed and a Deployment does not exist. However, this approach does cause an `IngressController` without a `status.domain` to go through a 2nd reconcile loop. This time the `status.domain` check will pass and the `ensure/enforce` methods will be called. Do you have any suggestions to avoid the 2nd reconcile loop? We can go back to the original `enforceEffectiveIngressDomain` and update status from within the function.

ironcladlou:
Consider condition sets produced by the following:

computeLoadBalancerStatus // returns LoadBalancer* conditions
computeDeploymentStatus // returns Deployment* conditions
computeDNSStatus // returns DNS* conditions

For now, let's say that some subset _S_ of those conditions must be "True" for the ingress controller to be considered available. For example, [LoadBalancerReady=True,DeploymentReady=true,DNSReady=true].

Given a set of the union of those conditions filtered by Status=True indexed by ingress controller, you can compute availability of the ingress controller by checking membership of the set.

New criteria can be later added by introducing new conditions to the set of availability-influencing conditions.

The same methodology could be applied to another meta-condition like Degraded or Progressing.

danehans:
@Miciah I updated the control flow of the main `Reconcile()` loop to remove the need to perform a 2nd reconciliation of an `IngressController` that does not have `status.domain` set. With the latest change, if an `IngressController` that does not have `status.domain` set, `syncIngressControllerStatus` is called. The `ensure/enforce` methods are called if `syncIngressControllerStatus` does not return an error. `syncIngressControllerStatus` is called one last time at the end of the reconciliation.

ironcladlou:
If there's a part of status processing that's mutations for things like domain and publishing, and I think that should be its own reconciler or something.

I think one of the outcome of those mutations would need to be a condition indicating we're "admitting" the ingress controller according to constraints (e.g. it must have a unique domain).

Finally, the main reconciler should be ignoring (and not receiving events for) ingress controllers which haven't been admitted, giving it some bedrock assumptions to stand on. There has to be a trust boundary, and decoupling lets us further disentangle the admission/reconciling/status trifecta.

Miciah:
> I don't believe so. If `syncIngressControllerStatus()` was called with an `IngressController` that did not have `status.domain` set, that means all the `ensure/enforce` methods of the main reconcile loop were bypassed and a Deployment does not exist.

Right, that makes sense.  Might be worth a comment since the logic is a little hairy.

> I updated the control flow of the main `Reconcile()` loop to remove the need to perform a 2nd reconciliation of an `IngressController` that does not have `status.domain` set. With the latest change, if an `IngressController` that does not have `status.domain` set, `syncIngressControllerStatus` is called.

Yeah, this is what I had in mind earlier.

> The `ensure/enforce` methods are called if `syncIngressControllerStatus` does not return an error. `syncIngressControllerStatus` is called one last time at the end of the reconciliation.

Should we call `syncIngressControllerStatus` even if an earlier `ensureFoo` method fails? Maybe check `IsStatusDomainSet` first:
```go
			if !IsStatusDomainSet(ingress) {
				if err := r.syncIngressControllerStatus(ingress, ingressConfig); err != nil {
					// ...
			} else if // ...
			}
			if IsStatusDomainSet(ingress) {
				if err := r.syncIngressControllerStatus(ingress, ingressConfig); err != nil {
					// ...
```

Miciah:
> If there's a part of status processing that's mutations for things like domain and publishing, and I think that should be its own reconciler or something.

Yeah, separating that logic out would make everything a lot more comprehensible.

> I think one of the outcome of those mutations would need to be a condition indicating we're "admitting" the ingress controller according to constraints (e.g. it must have a unique domain).
> 
> Finally, the main reconciler should be ignoring (and not receiving events for) ingress controllers which haven't been admitted, giving it some bedrock assumptions to stand on. There has to be a trust boundary, and decoupling lets us further disentangle the admission/reconciling/status trifecta.

I don't think this PR makes the situation worse; should we tackle this refactoring into separate controllers in a follow-up?

Miciah:
> The `ensure/enforce` methods are called if `syncIngressControllerStatus` does not return an error. 

On second look, I believe the above statement is incorrect because of the `else if`:
```go
			if !IsStatusDomainSet(ingress) {
				if err := r.syncIngressControllerStatus(ingress, ingressConfig); err != nil {
					// ...
			} else if // ...
```
I amend my earlier suggestion as follows:
```go
			if !IsStatusDomainSet(ingress) {
				if err := r.syncIngressControllerStatus(ingress, ingressConfig); err != nil {
					// ...
			}
			if IsStatusDomainSet(ingress) {
				 // ...
				if err := r.enforceEffectiveEndpointPublishingStrategy(ingress, infraConfig); err != nil {
					// ...
				} else if // ...
				}
				if err := r.syncIngressControllerStatus(ingress, ingressConfig); err != nil {
					// ...
```

danehans:
I am updating the PR to address your https://github.com/openshift/cluster-ingress-operator/pull/224#issuecomment-495068119.

> I don't think this PR makes the situation worse; should we tackle this refactoring into separate controllers in a follow-up?

I prefer getting this PR merged after my upcoming changes and refactor into a separate controller in a future PR.

danehans:
I have run the operator e2e tests locally multiple times without `TestIngressControllerUpdate` failure:
```
$ make test-e2e
KUBERNETES_CONFIG="/Users/daneyonhansen/code/go/src/github.com/openshift/installer/dhansen/auth/kubeconfig" WATCH_NAMESPACE=openshift-ingress-operator go test -count 1 -v -tags e2e -run ".*" ./...
<SNIP>
=== RUN   TestIngressControllerUpdate
--- PASS: TestIngressControllerUpdate (5.18s)
<SNIP>
ok  	github.com/openshift/cluster-ingress-operator/test/e2e	94.335s
```
/test e2e-aws-operator

danehans:
`The build "a234567890123456789012345678901234567890123456789012345678-1" status is "Failed"
occurred`

/test e2e-aws
/test e2e-aws-upgrade

danehans:
```
fail [github.com/openshift/origin/test/extended/templates/templateinstance_readiness.go:107]: Unexpected error:
    <*errors.errorString | 0xc0018e3aa0>: {
        s: "Failed to import expected imagestreams",
    }
```
/test e2e-aws

ironcladlou:
I think https://github.com/openshift/cluster-ingress-operator/pull/282 and https://github.com/openshift/cluster-ingress-operator/pull/283 make this one obsolete.

ironcladlou:
I think this one is obsolete now.

/close

PR #227, opened 2019-05-05
Fix hack/release.sh: Skip ClusterOperator resource creation

pravisankar:
- As per the spec[1], ClusterOperator resource must be created by the operator.

[1] https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#how-clusterversionoperator-handles-clusteroperator-in-release-imagehttps://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#how-clusterversionoperator-handles-clusteroperator-in-release-image


pravisankar:

PR #228, opened 2019-05-05
Cleanup: Remove unnecessary Factory struct from manifests

pravisankar:


pravisankar:

Miciah:
/lgtm
/hold
This is a nice clean-up, but we probably should hold it till after 4.1.

ironcladlou:
/hold cancel

what the...

PR #232, opened 2019-05-08
Fixes Bug 1707872: Refactors getLBHostedZone func to request a single load-balancer

danehans:
Previously, `getLBHostedZone()` returned all load-balancers in a zone. This PR updated `getLBHostedZone` to use the elb id (first 32 characters of the elb's dns name) to find its
associated hosted zone id.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -190,31 +190,27 @@ func (m *Manager) getLBHostedZone(name string) (string, error) {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
-	if id, exists := m.lbZones[name]; exists {
-		return id, nil
+	if zoneID, exists := m.lbZones[name]; exists {
+		return zoneID, nil
 	}
 
-	var id string
-	fn := func(resp *elb.DescribeLoadBalancersOutput, lastPage bool) (shouldContinue bool) {
-		for _, lb := range resp.LoadBalancerDescriptions {
-			log.V(0).Info("found load balancer", "name", aws.StringValue(lb.LoadBalancerName), "dns name", aws.StringValue(lb.DNSName), "hosted zone ID", aws.StringValue(lb.CanonicalHostedZoneNameID))
-			if aws.StringValue(lb.CanonicalHostedZoneName) == name {
-				id = aws.StringValue(lb.CanonicalHostedZoneNameID)
-				return false
-			}
-		}
-		return true
-	}
-	err := m.elb.DescribeLoadBalancersPages(&elb.DescribeLoadBalancersInput{}, fn)
+	// An elb name is the first 32 bytes of the elb's dns name.
+	// https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L86-L89
+	lbName := name[:32]
+	input := elb.DescribeLoadBalancersInput{LoadBalancerNames: []*string{aws.String(lbName)}}
+	output, err := m.elb.DescribeLoadBalancers(&input)
 	if err != nil {
 		return "", fmt.Errorf("failed to describe load balancers: %v", err)
 	}
-	if len(id) == 0 {
+	/// only 1 elb should have been created for the lb service.
+	if len(output.LoadBalancerDescriptions) != 1 {
 		return "", fmt.Errorf("couldn't find hosted zone ID of ELB %s", name)
 	}
-	log.Info("associating load balancer with hosted zone", "dns name", name, "zone", id)
-	m.lbZones[name] = id
-	return id, nil
+	zoneID := aws.StringValue(output.LoadBalancerDescriptions[0].CanonicalHostedZoneNameID)
+	log.Info("associating load balancer with hosted zone", "dns name", name, "zone", zoneID)
+	m.lbZones[name] = zoneID
+
+	return zoneID, nil
 }
 
 type action string


pkg/dns/aws/dns.go
```diff
@@ -190,31 +190,25 @@ func (m *Manager) getLBHostedZone(name string) (string, error) {
...
-			if aws.StringValue(lb.CanonicalHostedZoneName) == name {
-				id = aws.StringValue(lb.CanonicalHostedZoneNameID)
-				return false
-			}
-		}
-		return true
-	}
-	err := m.elb.DescribeLoadBalancersPages(&elb.DescribeLoadBalancersInput{}, fn)
+	// An elb id is the first 32 characters of the elb's dns name.
+	lbID := name[0:32]
```
ironcladlou: Interesting... how safe is it to make this assumption? Is it documented somewhere?

danehans: From [elb docs](https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_CreateLoadBalancer.html):

> LoadBalancerName
This name must be unique within your set of load balancers for the region, must have a maximum of 32 characters, must contain only alphanumeric characters or hyphens, and cannot begin or end with a hyphen.

From cloud provider [code](https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L86-L89):
```
	//AWS requires that the name of a load balancer is shorter than 32 bytes.
	if len(ret) > 32 {
		ret = ret[:32]
	}
```

Should we instead call `DefaultLoadBalancerName()` in the cloudprovider pkg to get the lb name? If so, I believe it saves us an AWS API call.

Miciah: Hm, so the AWS cloud provider [looks up the load balancer](https://github.com/kubernetes/cloud-provider-aws/blob/8f6e1e3d2cfa20a0deac088b9c4022d433500369/pkg/cloudprovider/providers/aws/aws.go#L3651) using [a name](https://github.com/kubernetes/cloud-provider-aws/blob/8f6e1e3d2cfa20a0deac088b9c4022d433500369/pkg/cloudprovider/providers/aws/aws.go#L3667) that is ultimately [derived from the Service name](https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L86-L89).  If we can assume that the cloud provider code is correct, we can simplify our code some.  I don't see any implication that the load balancer name must match the hosted zone id though, so I think we still need a `DescribeLoadBalancers` API call to get the hosted zone id though.

danehans: > Hm, so the AWS cloud provider looks up the load balancer using a name that is ultimately derived from the Service name.

That's what I see in the code and when I look at the service id and status.loadbalancer.ingress.hostname.

> If we can assume that the cloud provider code is correct, we can simplify our code some.

OK, let me know if you have any suggestions beyond this PR.

pkg/dns/aws/dns.go
```diff
@@ -190,31 +190,27 @@ func (m *Manager) getLBHostedZone(name string) (string, error) {
...
-				id = aws.StringValue(lb.CanonicalHostedZoneNameID)
-				return false
-			}
-		}
-		return true
-	}
-	err := m.elb.DescribeLoadBalancersPages(&elb.DescribeLoadBalancersInput{}, fn)
+	// An elb name is the first 32 bytes of the elb's dns name.
+	// https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L86-L89
+	lbName := name[:32]
```
Miciah: From the cited code, we know that the name of the load balancer is a function of the UID of the service.  However, `name` is the host name of the service, right? If we had the service's UID, we could go through the steps in the cited code (take the UID, prepend an 'a', elide hyphens, and truncate to 32 characters) to get the LB's name, but we have the host name, not the UID.

danehans: > However, name is the host name of the service, right?

Name is derived from the first 32-bytes of the service's `uid`. For example a service with `uid: 5b039b2f-72b4-11e9-b976-0a20dbb71566`, has an elb name of `a5b039b2f72b411e9b9760a20dbb7156`. The elb's dns name is equivalent to the service's `hostname`.

Miciah: > Name is derived from the first 32-bytes of the service's `uid`.

Correct.

> The elb's dns name is equivalent to the service's `hostname`.

Correct ([source](https://github.com/kubernetes/cloud-provider-aws/blob/8f6e1e3d2cfa20a0deac088b9c4022d433500369/pkg/cloudprovider/providers/aws/aws.go#L3671-L3676)).

The assumption that I want to validate is that the load balancer's `DNSName` is the same as its `LoadBalancerName`.  The most definitive statement I have found so far in the AWS documentation is the following:

> Each Classic Load Balancer receives a default Domain Name System (DNS) name. This DNS name includes the name of the AWS region in which the load balancer is created. For example, if you create a load balancer named my-loadbalancer in the US West (Oregon) region, your load balancer receives a DNS name such as my-loadbalancer-1234567890.us-west-2.elb.amazonaws.com.

https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/using-domain-names-with-elb.html

I'm hoping for a stronger API commitment than a "for example".

ironcladlou: Assumption: only the cloud provider knows the opaque platform-specific load balancer identifier, because the cloud provider creates the load balancer.

Is it thus the responsibility of the cloud provider to surface this sort of information to downstream consumers? I'm not sure what form that would take on a `corev1.Service` (conditions? fields?), but without it we're left essentially reverse engineering potentially volatile internal cloud provider naming functions.

danehans:
@Miciah @ironcladlou based on your feedback and supporting references, it appears the current approach of listing all lb's for a zone needs to stay. I'll close out this PR.

PR #233, opened 2019-05-09
Don't ignore LastTransitionTime when comparing ingress controller status conditions

pravisankar:
- LastTransitionTime field is only changed when there is a change,
so we expect LastTransitionTime to remain the same when there is no change.

- Some refactor related to handling of ingress status conditions and corresponding tests.


pravisankar:

Miciah:
/lgtm
/hold
Holding for 4.2.

PR #235, opened 2019-05-14
Make the ingress operator hostNetwork

danwinship:
We're considering blocking access from the pod network to the AWS metadata IP by default. Even if we don't do it in openshift-sdn, there may be other network plugins that implement this behavior. (The 169.254.0.0/16 range is supposed to be "link-local", so it's technically incorrect to route traffic for 169.254.169.254 from br0 to eth0.) So this makes the ingress operator run as hostNetwork. (As seen in https://github.com/openshift/origin/pull/22826, the ingress operator is currently the only OpenShift pod that tries to reach the metadata IP from the pod network.)

It's possible that it depends on being non-hostNetwork in some way in which case this will totally break things? Let's see what e2e-aws says...

### DIFF
--- a/manifests/02-deployment.yaml
+++ b/manifests/02-deployment.yaml
@@ -32,6 +32,7 @@ spec:
         tolerationSeconds: 120
       serviceAccountName: ingress-operator
       priorityClassName: system-cluster-critical
+      hostNetwork: true
       containers:
         - name: ingress-operator
           terminationMessagePolicy: FallbackToLogsOnError


danwinship:
We may want this to go in for 4.1 so it would be good to know if @openshift/sig-network-edge thinks there are any reasons not to

ironcladlou:
Would really rather get the AWS region from cluster config (https://github.com/openshift/installer/pull/1725) than switch to hostNetwork given the **only** use case is to get the AWS region...

danwinship:
That makes sense

cgwalters:
The other option is for the host to gather this (we already need to query it) and drop it in a file you can mount in and read.

danwinship:
Feel free to close this PR and replace it with an issue / bug / jira card for no longer using the metadata IP.

danwinship:
> Feel free to close this PR and replace it with an issue / bug / jira card for no longer using the metadata IP.

but @knobunc says we definitely want to block access to the metadata API for 4.1, so ingress is going to need *some* fix

ironcladlou:
What if we go back to the deprecated `kube-system/cluster-config-v1` ConfigMap for region pending https://github.com/openshift/installer/pull/1725?

ironcladlou:
See https://github.com/openshift/cluster-ingress-operator/pull/238 for an alternative.

ironcladlou:
Replaced by https://github.com/openshift/cluster-ingress-operator/pull/238 which is less risky.
/close

PR #240, opened 2019-05-17
Bug 1709958: Use graceful shutdown for router

Miciah:
Configure the router deployment to ensure graceful shutdown:

* Set a lifecycle pre-stop handler that runs the haproxy-router image's `shutdown-haproxy` script.
* Set a 90-second termination grace period.
* Create a pod disruption budget with a maximum of 25% unavailable pods.

* `manifests/00-cluster-role.yaml`: Allow the operator to manage `PodDisruptionBudget` resources.
* `pkg/operator/controller/controller.go` (`ensureIngressController`): Call `ensureRouterPodDisruptionBudget`.
* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): Set a pre-stop handler.  Set a grace period of 90 seconds.
(`deploymentConfigChanged`): Compare the current and expected container lifecycle handlers and pod termination grace period, and set them in the updated deployment.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDeploymentConfigChanged`): Add test cases for container lifecycle handlers and pod termination grace period.
* `pkg/operator/controller/names.go` (`RouterPodDisruptionBudgetName`): New function.
* `pkg/operator/controller/router_poddisruptionbudget.go`: New file.
(`ensureRouterPodDisruptionBudget`): New method.  Ensure the appropriate pod disruption budget exists for the given ingress controller.  Get the current one using `currentRouterPodDisruptionBudget`.  If none exists, create one using the `desiredRouterPodDisruptionBudget` function and `createRouterPodDisruptionBudget` method.  If one already exists, update it if necessary using `updateRouterPodDisruptionBudget`.
(`createRouterPodDisruptionBudget`): New method.

---

Depends on https://github.com/openshift/router/pull/28.


pkg/operator/controller/controller_router_deployment_test.go
```diff
@@ -394,11 +394,33 @@ func TestDeploymentConfigChanged(t *testing.T) {
...
+				deployment.Spec.Template.Spec.TerminationGracePeriodSeconds = &ninety
+			},
+			expect: true,
+		},
 	}
 
 	for _, tc := range testCases {
 		nineteen := int32(19)
 		fourTwenty := int32(420) // = 0644 octal.
+		thirty := int64(30)
```
danehans: Is `thirty` expected to match the termination grace period of https://github.com/openshift/router/pull/28? If so, should this be documented here?

Miciah: I only chose it because it's the default value, which means that it is the value that the operator will see on an existing ingress controller's deployment when the operator is upgraded.  It really could be any value (including nil), as long as the "if the deployment terminationGracePeriodSeconds is changed" test case sets a distinct value.

Miciah:
/close
Superseded by #280.

PR #241, opened 2019-05-22
Add LoadBalancer status reporting on IngressController

ironcladlou:
Compute LoadBalancer* status conditions on IngressController resources. Do so by
introducing some new cache/index capabilities.

### DIFF
--- a/pkg/operator/controller/controller.go
+++ b/pkg/operator/controller/controller.go
@@ -47,15 +47,16 @@ var log = logf.Logger.WithName("controller")
 //
 // The controller will be pre-configured to watch for IngressController resources
 // in the manager namespace.
-func New(mgr manager.Manager, config Config) (controller.Controller, error) {
+func New(mgr manager.Manager, statusCache *ingressStatusCache, config Config) (controller.Controller, error) {
 	kubeClient, err := operatorclient.NewClient(config.KubeConfig)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create kube client: %v", err)
 	}
 	reconciler := &reconciler{
-		Config:   config,
-		client:   kubeClient,
-		recorder: mgr.GetEventRecorderFor("operator-controller"),
+		Config:      config,
+		client:      kubeClient,
+		recorder:    mgr.GetEventRecorderFor("operator-controller"),
+		statusCache: statusCache,
 	}
 	c, err := controller.New("operator-controller", mgr, controller.Options{Reconciler: reconciler})
 	if err != nil {
@@ -88,6 +89,8 @@ type reconciler struct {
 	// we do not need to synchronize when changing rest scheme/mapper fields.
 	client   kclient.Client
 	recorder record.EventRecorder
+
+	statusCache *ingressStatusCache
 }
 
 // Reconcile expects request to refer to a ingresscontroller in the operator
--- a/pkg/operator/controller/ingress_status.go
+++ b/pkg/operator/controller/ingress_status.go
@@ -7,10 +7,19 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+
 	operatorv1 "github.com/openshift/api/operator/v1"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // syncIngressControllerStatus computes the current status of ic and
@@ -24,7 +33,24 @@ func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment, 
 	updated := ic.DeepCopy()
 	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
 	updated.Status.Selector = selector.String()
-	updated.Status.Conditions = computeIngressStatusConditions(updated.Status.Conditions, deployment)
+
+	updated.Status.Conditions = []operatorv1.OperatorCondition{}
+	updated.Status.Conditions = append(updated.Status.Conditions, computeIngressStatusConditions(updated.Status.Conditions, deployment)...)
+	updated.Status.Conditions = append(updated.Status.Conditions, r.statusCache.computeLoadBalancerStatus(ic)...)
+
+	for i := range updated.Status.Conditions {
+		newCondition := &updated.Status.Conditions[i]
+		var oldCondition *operatorv1.OperatorCondition
+		for j, possibleOldCondition := range ic.Status.Conditions {
+			if possibleOldCondition.Type == newCondition.Type {
+				old := ic.Status.Conditions[j]
+				oldCondition = &old
+				break
+			}
+		}
+		setIngressLastTransitionTime(newCondition, oldCondition)
+	}
+
 	if !ingressStatusesEqual(updated.Status, ic.Status) {
 		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
 			return fmt.Errorf("failed to update ingresscontroller status: %v", err)
@@ -57,7 +83,6 @@ func computeIngressAvailableCondition(oldAvailableCondition *operatorv1.Operator
 		availableCondition.Message = "no Deployment replicas available"
 	}
 
-	setIngressLastTransitionTime(&availableCondition, oldAvailableCondition)
 	return availableCondition
 }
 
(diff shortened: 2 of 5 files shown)


pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+		controllers: controllers,
+		services:    services,
+	}
+}
+
+func (s *ingressStatus) computeLoadBalancerStatus(ic *operatorv1.IngressController) []operatorv1.OperatorCondition {
+	if ic.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
+		return []operatorv1.OperatorCondition{
+			{
+				Type:    "LoadBalancerManaged",
```
Miciah: `"LoadBalancerManaged"`→`operatorv1.LoadBalancerManagedIngressConditionType`

pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+				Reason:  "UnsupportedEndpointPublishingStrategy",
+				Message: fmt.Sprintf("The %s endpoint publishing strategy does not support a load balancer", ic.Status.EndpointPublishingStrategy.Type),
+			},
+		}
+	}
+
+	conditions := []operatorv1.OperatorCondition{}
+
+	conditions = append(conditions, operatorv1.OperatorCondition{
+		Type:    "LoadBalancerManaged",
```
Miciah: `"LoadBalancerManaged"`→`operatorv1.LoadBalancerManagedIngressConditionType`

pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+		Message: "The LoadBalancer endpoint publishing strategy is specified",
+	})
+
+	pending := indexContains(s.services, "pending-for", ic.Name)
+	provisioned := indexContains(s.services, "provisioned-for", ic.Name)
+
+	switch {
+	case pending:
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    "LoadBalancerAvailable",
```
Miciah: `"LoadBalancerAvailable"`→`operatorv1.LoadBalancerReadyIngressConditionType`

pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "LoadBalancerPending",
+			Message: "The LoadBalancer service is pending",
+		})
+	case provisioned:
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    "LoadBalancerAvailable",
```
Miciah: `"LoadBalancerAvailable"`→`operatorv1.LoadBalancerReadyIngressConditionType`

pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "LoadBalancerProvisioned",
+			Message: "The LoadBalancer service is provisioned",
+		})
+	default:
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    "LoadBalancerAvailable",
```
Miciah: `"LoadBalancerAvailable"`→`operatorv1.LoadBalancerReadyIngressConditionType`

pkg/operator/controller/ingress_status.go
```diff
@@ -7,12 +7,145 @@ import (
...
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "LoadBalancerStatusUnknown",
+			Message: "The LoadBalancer service status is unknown",
```
Miciah: The union of "pending-for" and "provisioned-for" is all LB services that are owned by ingress controllers, right? So if neither index had an entry for the ingress controller, that means the service is non-existent, right? So the status us now unknown, but rather that the LB service does not exist.

pkg/operator/controller/ingress_status.go
ironcladlou: Switched to correct values, deferring upgrade of client-go for now.

pkg/operator/controller/ingress_status.go
ironcladlou: Switched to correct values, deferring upgrade of client-go for now.

pkg/operator/controller/ingress_status.go
ironcladlou: Yes, good point — I changed the default to `LoadBalancerNotFound` and added a test.

pkg/operator/controller/ingress_status.go
Miciah: Why does client-go need to be upgraded? 
 `operatorv1.LoadBalancerManagedIngressConditionType` is a string value.

pkg/operator/operator.go
```diff
@@ -82,29 +82,36 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
...
+	// should be added here.
+	mapper, err := apiutil.NewDiscoveryRESTMapper(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get API Group-Resources")
+	}
+	operandCache, err := cache.New(kubeConfig, cache.Options{Namespace: "openshift-ingress", Scheme: scheme, Mapper: mapper})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create openshift-ingress cache: %v", err)
+	}
+	buildStatusCache := cache.MultiNamespacedCacheBuilder([]string{"openshift-ingress-operator", "openshift-ingress"})
```
ironcladlou: This is a new feature of controller-runtime which will let us work with indexable caches across namespaces.

pkg/operator/controller/ingress_status.go
ironcladlou: You're right, I missed that. Fixed

ironcladlou:
@Miciah @danehans Still need to add some e2e tests, but this seems to be working.

ironcladlou:
There's plenty of room to refactor the cache plumbing, but the gist of it is that given a multi-namespaced `cache.Cache` populated with all objects we care about for status reporting (e.g. ingresscontrollers in the operator namespace, services in the operand(s) namespace, etc.) we can now trust that cache as a source of truth for status and build arbitrary indexes/views on the cache.

So far it's quite verbose to set up the scaffolding for all this, but at the end of the day the actual status algorithms seem reducible to various set operations. Adding new indexes now should be easier.

pkg/operator/controller/ingress_status_test.go
```diff
@@ -6,13 +6,200 @@ import (
...
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Logf("evaluating test %s", test.desc)
+		controllers := toolscache.NewIndexer(toolscache.MetaNamespaceKeyFunc, makeIndexers(controllerIndexers))
+		services := toolscache.NewIndexer(toolscache.MetaNamespaceKeyFunc, makeIndexers(serviceIndexers))
+		status := &ingressStatusCache{
+			contains: func(kind runtime.Object, name, value string) bool {
```
ironcladlou: Can probably find a way to make this more generic, but good enough for now. Can't use the `cache.Cache` interfaces in this context without setting up an apiserver.

pkg/operator/controller/ingress_status.go
```diff
@@ -24,7 +33,25 @@ func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment,
...
+		newCondition := updated.Status.Conditions[i]
+		var oldCondition *operatorv1.OperatorCondition
+		for j, possibleOldCondition := range ic.Status.Conditions {
+			if possibleOldCondition.Type == newCondition.Type {
+				old := ic.Status.Conditions[j]
+				oldCondition = &old
+				break
+			}
+		}
+		lastTransitionTime := getLastTransitionTime(&newCondition, oldCondition)
```
ironcladlou: setLastTransitionTime has an issue w/ modifying the array element pointer in this context, didn't take the time to look closely. Seemed easier to just remove the side effect

pkg/operator/controller/ingress_status.go
```diff
@@ -24,7 +33,25 @@ func (r *reconciler) syncIngressControllerStatus(deployment *appsv1.Deployment,
 	updated := ic.DeepCopy()
 	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
 	updated.Status.Selector = selector.String()
-	updated.Status.Conditions = computeIngressStatusConditions(updated.Status.Conditions, deployment)
+
+	updated.Status.Conditions = []operatorv1.OperatorCondition{}
+	updated.Status.Conditions = append(updated.Status.Conditions, computeIngressStatusConditions(updated.Status.Conditions, deployment)...)
+	updated.Status.Conditions = append(updated.Status.Conditions, r.statusCache.computeLoadBalancerStatus(ic)...)
+
+	for i := range updated.Status.Conditions {
```
ironcladlou: there's probably an elegant functional approach to this one

pkg/operator/controller/ingress_status.go
```diff
@@ -101,3 +139,148 @@ func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
...
+	}
+	return controlledBy(service, isPending)
+}
+
+var serviceIndexers = map[string]client.IndexerFunc{
+	"pending-for":     indexByPending,
+	"provisioned-for": indexByProvisioned,
+}
+var controllerIndexers = map[string]client.IndexerFunc{
+	"publishing":          indexByPublishing,
```
Miciah: Looks like nothing uses this index.

pkg/operator/controller/ingress_status.go
```diff
@@ -101,3 +139,148 @@ func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
...
+	// TODO: after rebase, replace with:
+	// meta.LenList(list) > 0
+	items, _ := meta.ExtractList(list)
+	return len(items) > 0
+}
+
+func NewIngressStatusCache(c cache.Cache) *ingressStatusCache {
+	add := func(cache cache.Cache, kind runtime.Object, indexers map[string]client.IndexerFunc) {
+		for name := range indexers {
+			cache.IndexField(kind, name, indexers[name])
```
Miciah: The error return value is ignored.

pkg/operator/controller/ingress_status.go
ironcladlou: Not anymore it doesn't, I deleted the usage. Isn't the problem completely generic?

Miciah: Yeah.  It is for the conditions that `ingress_status.go` sets.  It looks like it is for the conditions that `status.go` sets as well.  (For `status.go`, we need to be careful to respect the special requirements for updating `LastTransitionTime` of the `Progressing` condition, but we handle that now by assigning a distinct value to its `Message` for the cases in which we need `LastTransitionTime` to be updated.)  So it looks like we can delete `setIngressLastTransitionTime`, the first parameter of `computeIngressStatusConditions`, and the first parameter of `computeIngressAvailableCondition` (and soon `getIngressAvailableCondition` as well), and we could do the same in `ingress.go` too.

pkg/operator/controller/controller.go
```diff
@@ -88,6 +91,11 @@ type reconciler struct {
 	// we do not need to synchronize when changing rest scheme/mapper fields.
 	client   kclient.Client
 	recorder record.EventRecorder
+
+	operatorCache cache.Cache
+	operandCache  cache.Cache
+
+	statusCache *ingressStatusCache
```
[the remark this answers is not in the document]
ironcladlou: I redid this slightly, let me know what you think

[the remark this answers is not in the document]
Miciah: Now `r.statusCache` and `r.StatusCache` are the same, right? Can we drop `statusCache`?

pkg/operator/controller/ingress_status.go
ironcladlou: Yes, that did it! thanks

pkg/operator/controller/ingress_status.go
```diff
@@ -101,3 +139,148 @@ func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
...
+func isProvisioned(service *corev1.Service) bool {
+	ingresses := service.Status.LoadBalancer.Ingress
+	return len(ingresses) > 0 && (len(ingresses[0].Hostname) > 0 || len(ingresses[0].IP) > 0)
+}
+
+func isPending(service *corev1.Service) bool {
+	return !isProvisioned(service)
+}
+
+func controlledBy(service *corev1.Service, matches func(*corev1.Service) bool) []string {
```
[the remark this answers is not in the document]
ironcladlou: Renamed this and all the others to hopefully be more descriptive

pkg/operator/controller/ingress_status.go
ironcladlou: Removed

pkg/operator/controller/ingress_status.go
ironcladlou: If we get an error here I think we should probably panic anyway? It would mean we wired up everything wrong from the start

pkg/operator/controller/controller.go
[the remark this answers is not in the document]
ironcladlou: Whoops, yeah. Latest version removes StatusCache from config entirely, because it's not really config anyway.

ironcladlou:
@Miciah @danehans e2e tests added, this is ready for final review AFAICT.

test/e2e/operator_test.go
```diff
@@ -77,47 +82,87 @@ func newIngressController(name, ns, domain string, epType operatorv1.EndpointPub
...
+
+func operatorConditionMap(conditions ...operatorv1.OperatorCondition) map[string]string {
+	conds := map[string]string{}
+	for _, cond := range conditions {
+		conds[string(cond.Type)] = string(cond.Status)
 	}
+	return conds
+}
+
+func conditionsMatchExpected(expected, actual map[string]string) bool {
```
ironcladlou: Not even sure if this is necessary/desired — should we fail if some condition type we don't expect appears?

ironcladlou: Can address in a followup once we decide

pkg/operator/controller/ingress_status.go
Miciah: Maybe we can implement some of these ideas in a follow-up.

pkg/operator/controller/ingress_status.go
```diff
@@ -101,3 +126,156 @@ func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
...
+		return []string{}
+	}
+	controller, ok := service.Labels[manifests.OwningIngressControllerLabel]
+	if !ok {
+		return []string{}
+	}
+	return []string{controller}
+}
+
+func indexLoadBalancerControllerByName(obj runtime.Object) []string {
```
Miciah: I'm not sure about the name; what do you think of `indexControllerWithLoadBalancerByName`?

ironcladlou: better, but can we do it in a followup so this has some time to soak?

pkg/operator/controller/ingress_status.go
```diff
@@ -101,3 +126,156 @@ func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
...
+	"provisioned-for": indexProvisionedLoadBalancerServiceByOwner,
+}
+
+// controllerIndexers are all the IngressController indexes supported by the
+// cache.
+var controllerIndexers = map[string]client.IndexerFunc{
+	"wants-load-balancer": indexLoadBalancerControllerByName,
+}
+
+// cacheContains is a contains fuction which knows how to query a cache.Cache.
```
Miciah: "fuction"→"function".

The phrase "is a contains function" implies that this function implements some "contains" interface, but its signature doesn't quite match the type of `contains` in `ingressStatusCache`.  Maybe it would be more helpful to say describe the function more abstractly; e.g., "cacheContains returns membership of the given value in the named index for the specified kind using the given cache."

Come to think of it, the `list` parameter is odd—to the caller, its purpose is to convey a resource kind, but the callee uses it to store a list of values of that kind, which forces the parameter to be the type for a list of resources of the kind.  It would be a more obvious API if callers instead could specify the type for the kind.  Alternatively, we could drop the `list` parameter and look up the type from the index using a map.

pkg/operator/controller/ingress_status_test.go
```diff
@@ -6,13 +6,200 @@ import (
...
+
+	toolscache "k8s.io/client-go/tools/cache"
 )
 
+func pendingLBService(owner string) *corev1.Service {
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "router-" + owner,
+			Labels: map[string]string{
+				manifests.OwningIngressControllerLabel: owner,
```
Miciah: Could use `loadBalancerServiceName`.

pkg/operator/controller/ingress_status_test.go
```diff
@@ -6,13 +6,200 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	toolscache "k8s.io/client-go/tools/cache"
 )
 
+func pendingLBService(owner string) *corev1.Service {
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "router-" + owner,
```
Miciah: Could use `loadBalancerServiceName`.

(I added this comment in the wrong place earlier.)

pkg/operator/controller/ingress_status_test.go
```diff
@@ -6,13 +6,200 @@ import (
...
+				},
+			},
+		},
+	}
+}
+
+func clusterIPservice(name string) *corev1.Service {
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "router-" + name + "internal",
```
Miciah: Could use `InternalIngressControllerServiceName`.

pkg/operator/operator.go
```diff
@@ -105,6 +93,28 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
 	if err != nil {
 		return nil, fmt.Errorf("failed to create openshift-ingress cache: %v", err)
 	}
+	// This cache is used to work with objects across all operator-managed
+	// namespaced from a single uniform cache. For now, it's used for status
+	// computation.
+	buildGlobalCache := cache.MultiNamespacedCacheBuilder([]string{"openshift-ingress-operator", "openshift-ingress"})
```
Miciah: `"openshift-ingress-operator"`→`config.Namespace`.

pkg/operator/operator.go
```diff
@@ -154,7 +164,7 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
 
 	return &Operator{
 		manager: operatorManager,
-		caches:  []cache.Cache{operandCache},
+		caches:  []cache.Cache{operandCache, globalCache},
```
Miciah: Does `globalCache` supersede `operandCache`? Maybe we can remove `operandCache` in a follow-up.

test/e2e/operator_test.go
```diff
@@ -77,47 +82,87 @@ func newIngressController(name, ns, domain string, epType operatorv1.EndpointPub
...
-	err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
-		if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: ns, Name: "default"}, ci); err != nil {
-			return false, nil
-		}
-		return true, nil
-	})
+	name := types.NamespacedName{Namespace: ns, Name: "default"}
+	expected := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
+		{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionTrue},
```
Miciah: We do not assume that the test suite is run on a platform where the operator manages load balancers; see the skip rule in `TestClusterProxyProtocol`.

test/e2e/operator_test.go
```diff
@@ -77,47 +82,87 @@ func newIngressController(name, ns, domain string, epType operatorv1.EndpointPub
...
+func TestOperatorAvailable(t *testing.T) {
+	cl, _, err := getClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := []configv1.ClusterOperatorStatusCondition{
+		{Type: configv1.OperatorAvailable, Status: configv1.ConditionTrue},
+	}
+	err = waitForClusterOperatorConditions(cl, expected...)
```
Miciah: Because you made these helpers variadic, building a slice is unnecessary—this code could be simplified as follows:
```go
		expected := configv1.ClusterOperatorStatusCondition{
			Type: configv1.OperatorAvailable,
			Status: configv1.ConditionTrue,
		}
		err = waitForClusterOperatorConditions(cl, expected)
```
Alternatively, if nothing is taking advantage of the variadic parameters, is there any reason not to use slice-valued parameters instead of variadic parameters?

Miciah:
/lgtm
I see no major issues.  We can let this soak and address any outstanding issues with follow-ups.

PR #242, opened 2019-05-30
add selector to servicemonitor

lihongan:
Previously the selector of ServiceMonitor is null, it causes `TLS handshake error from x.x.x.x:xx: remote error: tls: bad certificate` in router pod logs if creating another custom ingresscontroller.
The message is gone after adding proper selector to each ServiceMonitor.
note: the `x.x.x.x` in the log is prometheus pod's IP.


ironcladlou:
Thanks for this!

/lgtm
/approve

lihongan:
I'd like to backport to fix https://bugzilla.redhat.com/show_bug.cgi?id=1724498
/cherrypick release-4.1

PR #243, opened 2019-05-31
Switch to Go modules, Go 1.12, and latest libraries

ironcladlou:
Migrate to Go modules. Switch to Go 1.12. Upgrade to Kube 1.14 and the latest
controller-runtime.

Switch CRD generation away from the openshift fork and to the upstream
controller-tools generator.

Motivations for modules:

1. Modules resolves a transitive controller-tools conflict between
controller-runtime and library-go which `dep` could not accomodate.

2. Kubernetes will soon be using modules, and some related community projects on
which we depend are also migrating. The Kubernetes KEP has a good motivation
section which applies very broadly[1].

There are some important notes for contributors:

1. Vendoring is enabled for simple reproducible builds. It's therefore important
to explicitly enable vendoring in various Go execution contexts if
reproducibility is a concern. All the build tooling has been updated to enable
vendoring. Any end-user scripts or tooling should be audited for vendoring
considerations.[2]

2. If your project lives in a GOPATH, you may have to move your project or
perform extra steps to activate modules.

[1] https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md#motivation
[2] https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away
[3] https://github.com/golang/go/wiki/Modules#how-to-use-modules

Depends on:
- [x] https://github.com/openshift/api/pull/328
- [x] https://github.com/openshift/api/pull/329
- [x] https://github.com/openshift/release/pull/3928


ironcladlou:
@smarterclayton 

I _think_ the unit job is using go 1.10:

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/243/pull-ci-openshift-cluster-ingress-operator-master-unit/777/build-log.txt

Will https://github.com/openshift/release/pull/3928 fix that?

ironcladlou:
/hold

After the dependent PRs merge, I will:

1. Upgrade the openshift/api dependency
2. Regenerate the CRD from new openshift/api metadata
3. Release the hold

manifests/00-custom-resource-definition.yaml
```diff
@@ -1,24 +1,33 @@
...
+  creationTimestamp: null
   name: ingresscontrollers.operator.openshift.io
 spec:
   group: operator.openshift.io
   names:
     kind: IngressController
-    listKind: IngressControllerList
     plural: ingresscontrollers
-    singular: ingresscontroller
-  scope: Namespaced
```
danehans: @ironcladlou why is `IngressController` no longer namespaced?

manifests/00-custom-resource-definition.yaml
```diff
@@ -1,24 +1,33 @@
...
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  creationTimestamp: null
   name: ingresscontrollers.operator.openshift.io
 spec:
   group: operator.openshift.io
   names:
     kind: IngressController
-    listKind: IngressControllerList
```
danehans: @ironcladlou is there a reason why the `listKind` field was removed?

manifests/00-custom-resource-definition.yaml
ironcladlou: Idiosyncrasy (and possible bug with) controller-tools (https://github.com/kubernetes-sigs/controller-tools/issues/214). However, not a problem for us because the scope defaults to namespaced. This explicit configuration may return later depending on how controller-tools addresses the issue.

manifests/00-custom-resource-definition.yaml
ironcladlou: The `listKind` is now inferred by convention relative to `kind`.

manifests/00-custom-resource-definition.yaml
danehans: Thanks for the explanation and issue reference.

ironcladlou:
@Miciah can remove the hold

/hold

hack/update-generated-crd.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -euo pipefail
+
+OUTDIR="${OUTDIR:-}"
+if [ -z "$OUTDIR" ]; then
+  OUTDIR="$(mktemp -d)"
+fi
+
+SKIP_COPY="${SKIP_COPY:-}"
```
Miciah: Should we treat set but empty as false? I have been treating set but empty as true.

hack/update-generated-crd.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -euo pipefail
+
+OUTDIR="${OUTDIR:-}"
+if [ -z "$OUTDIR" ]; then
```
Miciah: FWIW, it's more bashy to use `[[``.

hack/update-generated-crd.sh
```diff
@@ -0,0 +1,23 @@
...
+# to be compatible with the latest metadata expectations. For now, generate the
+# minimal set of type files we know are needed for the ingress types.
+set -x
+GO111MODULE=on GOFLAGS=-mod=vendor go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:trivialVersions=true \
+  paths=./vendor/github.com/openshift/api/operator/v1/doc.go\;./vendor/github.com/openshift/api/operator/v1/types.go\;./vendor/github.com/openshift/api/operator/v1/types_ingress.go \
+  output:crd:dir="$OUTDIR"
+set +x
+
+if [ -z "$SKIP_COPY" ]; then
+  cp $OUTDIR/operator.openshift.io_ingresscontrollers.yaml manifests/00-custom-resource-definition.yaml
```
Miciah: Better to quote to avoid word splitting on `$OUTDIR`.

hack/verify-generated-crd.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+TMP_DIR="$(mktemp -d)"
+
+OUTDIR=$TMP_DIR SKIP_COPY=true ./hack/update-generated-crd.sh
+
+diff -Naup $TMP_DIR/operator.openshift.io_ingresscontrollers.yaml manifests/00-custom-resource-definition.yaml
```
Miciah: Better to quote to avoid word splitting on `$TMP_DIR`.

tools/tools.go
```diff
@@ -0,0 +1,10 @@
+// +build tools
+
+// This package contains import references to packages required only for the
+// build process.
```
Miciah: Is the purpose of these references ensuring that modules recognize that these packages are used?

Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
-WORKDIR /go/src/github.com/openshift/cluster-ingress-operator
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go
 COPY . .
-RUN make build
+RUN make GOPATH="" build
```
Miciah: Why is clearing `GOPATH` needed?

Dockerfile.rhel7
```diff
@@ -1,10 +1,10 @@
...
-WORKDIR /go/src/github.com/openshift/cluster-ingress-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
+WORKDIR /go
 COPY . .
-RUN make build
+RUN make GOPATH="" build
 
-FROM registry.svc.ci.openshift.org/ocp/4.0:base
-COPY --from=builder /go/src/github.com/openshift/cluster-ingress-operator/ingress-operator /usr/bin/
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
```
Miciah: Why does `Dockerfile.rhel7` switch to the Origin image?

Dockerfile
```diff
@@ -1,10 +1,10 @@
...
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
```
Miciah: Maybe tangential to this commit, but why are we still use v4.0?

hack/update-generated-crd.sh
ironcladlou: Fixed

hack/update-generated-crd.sh
ironcladlou: Fixed

hack/verify-generated-crd.sh
ironcladlou: Fixed

hack/update-generated-crd.sh
ironcladlou: Fixed

tools/tools.go
ironcladlou: Added a link to https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module which explains the technique and motivation.

hack/verify-generated-crd.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+TMP_DIR="$(mktemp -d)"
```
Miciah: Should we clean up `$TMP_DIR`?

Dockerfile
ironcladlou: Can't use modules within a GOPATH — however, looking at this again, I found a better way which is to use a working directory not in the image's default GOPATH. I switched the working directory and removed the GOPATH unsetting.

Dockerfile.rhel7
ironcladlou: Copy/paste error, fixed

Dockerfile
ironcladlou: v4.2 not existing is the main reason 😆

hack/verify-generated-crd.sh
ironcladlou: I wanted to see the stuff when debugging... temp is ephemeral and the data generated is tiny, so not sure. Could make it configurable?

Dockerfile
Miciah: Does v4.1 exist?

hack/update-generated-crd.sh
Miciah: Line 5 has the same issue,

Dockerfile
ironcladlou: Nope

hack/update-generated-crd.sh
ironcladlou: Already fixed

ironcladlou:
@Miciah feedback addressed, I did a bit of history cleanup.

hack/update-generated-bindata.sh
```diff
@@ -1,29 +1,10 @@
...
-pushd "${GOPATH}" >/dev/null
-go install github.com/kevinburke/go-bindata/...
-popd >/dev/null
-
 # Using "-modtime 1" to make generate target deterministic. It sets all file
 # time stamps to unix timestamp 1
-${GOPATH}/bin/go-bindata -mode 420 -modtime 1 -pkg manifests -o ${OUTDIR}/pkg/manifests/bindata.go assets/...
+GO111MODULE=on GOFLAGS=-mod=vendor go run github.com/kevinburke/go-bindata/go-bindata -mode 420 -modtime 1 -pkg manifests -o ${OUTDIR}/pkg/manifests/bindata.go assets/...
 
 gofmt -s -w ${OUTDIR}/pkg/manifests/bindata.go
```
Miciah: It would be prudent to quote `${OUTDIR}` on lines 8 and 10.

hack/update-generated-crd.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+OUTDIR="${OUTDIR:-}"
+if [[ -z "$OUTDIR" ]]; then
```
Miciah: Could simplify:
```bash
if [[ -z "${OUTDIR-}" ]]; then
```

hack/update-generated-bindata.sh
ironcladlou: Not a regression, save it for a followup and spare the new round of CI testing?

hack/update-generated-crd.sh
ironcladlou: I like it, but not enough to trigger a new round of CI tests

PR #244, opened 2019-06-04
Bug 1717494: Refactor client and cache handling

ironcladlou:
* Delete custom cache setup, switch exclusively to `MultiNamespacedCache`
* Consolidate client usage and use dynamic discovery (see https://github.com/kubernetes-sigs/controller-runtime/issues/321 — hat tip to @danwinship for https://github.com/openshift/cluster-network-operator/pull/95). Fixes [bz1717494](https://bugzilla.redhat.com/show_bug.cgi?id=1717494).
* Plumb the cache through for indexed lookups

### DIFF
--- a/cmd/ingress-operator/main.go
+++ b/cmd/ingress-operator/main.go
@@ -55,9 +55,10 @@ func main() {
 	// Collect operator configuration.
 	operatorNamespace := os.Getenv("WATCH_NAMESPACE")
 	if len(operatorNamespace) == 0 {
-		log.Error(fmt.Errorf("missing environment variable"), "'WATCH_NAMESPACE' environment variable must be set")
-		os.Exit(1)
+		operatorNamespace = "openshift-ingress-operator"
 	}
+	log.Info("using operator namespace", "namespace", operatorNamespace)
+
 	ingressControllerImage := os.Getenv("IMAGE")
 	if len(ingressControllerImage) == 0 {
 		log.Error(fmt.Errorf("missing environment variable"), "'IMAGE' environment variable must be set")
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -11,6 +11,7 @@ import (
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	operatorconfig "github.com/openshift/cluster-ingress-operator/pkg/operator/config"
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+	operatorutil "github.com/openshift/cluster-ingress-operator/pkg/util"
 	certcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/certificate"
 	certpublishercontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/certificate-publisher"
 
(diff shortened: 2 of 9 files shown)


ironcladlou:
@Miciah @danehans this is ready for review... tomorrow I need to repro https://bugzilla.redhat.com/show_bug.cgi?id=1711373 with the ingress operator and make a new bug to link to this PR. I can do a similar change in dns-operator to fix the existing bug.

cmd/ingress-operator/main.go
```diff
@@ -55,9 +55,10 @@ func main() {
 	// Collect operator configuration.
 	operatorNamespace := os.Getenv("WATCH_NAMESPACE")
 	if len(operatorNamespace) == 0 {
-		log.Error(fmt.Errorf("missing environment variable"), "'WATCH_NAMESPACE' environment variable must be set")
-		os.Exit(1)
+		operatorNamespace = "openshift-ingress-operator"
```
Miciah: This looks like a leftover from development that should be reverted.

pkg/operator/operator.go
```diff
@@ -60,115 +52,65 @@ type Operator struct {
...
-		Namespace: config.Namespace,
-		Scheme:    scheme,
+	mgr, err := manager.New(kubeConfig, manager.Options{
+		Namespace:      config.Namespace,
+		Scheme:         scheme,
+		MapperProvider: operatorutil.NewDynamicRESTMapper,
+		NewCache: cache.MultiNamespacedCacheBuilder([]string{
+			config.Namespace,
+			"openshift-ingress",
+			"openshift-config-managed",
```
Miciah: This should use a named constant.

pkg/operator/operator.go
```diff
@@ -179,34 +121,22 @@ func New(config operatorconfig.Config, dnsManager dns.Manager, kubeConfig *rest.
 func (o *Operator) Start(stop <-chan struct{}) error {
 	// Periodicaly ensure the default controller exists.
 	go wait.Until(func() {
+		if !o.manager.GetCache().WaitForCacheSync(stop) {
+			log.Error(nil, "failed to sync cache before ensuring default ingresscontroller")
+			return
+		}
```
Miciah: Is `ensureDefaultIngressController` actually using the cache?

cmd/ingress-operator/main.go
ironcladlou: It was certainly useful — why not keep it?

pkg/operator/operator.go
ironcladlou: Not right now, but generally it seemed like a good idea to wait until things were up and running

pkg/operator/operator.go
ironcladlou: Fixed

cmd/ingress-operator/main.go
Miciah: Acceptable.

pkg/operator/operator.go
Miciah: Since the `wait.Until` uses a 1-minute period, does this mean that the operator will probably take an extra minute to create the default ingress controller on a new cluster?

Miciah: Hm, I guess in the normal case, `WaitForCacheSync` will block and succeed and delay the creation exactly as long as it takes to sync the cache.

PR #248, opened 2019-06-10
metrics: fix metric registration bug

ironcladlou:
Use a controller name which is always a valid metric name, allowing the
controller-runtime to successfully register all metrics which incorporate
the controller's name.


pkg/operator/controller/controller.go
```diff
@@ -37,6 +37,10 @@ const (
 	IngressControllerFinalizer = "ingresscontroller.operator.openshift.io/finalizer-ingresscontroller"
 )
 
+const (
+	controllerName = "ingress_controller"
+)
```
Miciah: Might as well join the two `const` blocks.

ironcladlou: Done

PR #249, opened 2019-06-10
Bug 1717737: status: add related ingresscontrollers to operator status

ironcladlou:
Add related ingresscontrollers to operator status to ensure ingresscontrollers
are captured by must-gather.

`oc adm must-gather` output:

```
$ tree namespaces/openshift-ingress-operator/
namespaces/openshift-ingress-operator/
└── operator.openshift.io
    └── ingresscontrollers
        └── default.yaml

2 directories, 1 file
```

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1717737

### DIFF
--- a/pkg/operator/controller/status.go
+++ b/pkg/operator/controller/status.go
@@ -38,7 +38,7 @@ func (r *reconciler) syncOperatorStatus() error {
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: IngressClusterOperatorName}}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: co.Name}, co); err != nil {
 		if errors.IsNotFound(err) {
-			initializeClusterOperator(co, ns.Name)
+			initializeClusterOperator(co)
 			if err := r.client.Create(context.TODO(), co); err != nil {
 				return fmt.Errorf("failed to create clusteroperator %s: %v", co.Name, err)
 			}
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/status.go
```diff
@@ -53,6 +53,18 @@ func (r *reconciler) syncOperatorStatus() error {
...
+	related := []configv1.ObjectReference{}
+	for _, ingress := range ingresses {
+		related = append(related, configv1.ObjectReference{
+			Group: operatorv1.GroupName,
+			Resource: "IngressController",
+			Namespace: r.Namespace,
+			Name: ingress.Name,
+		})
+	}
+	co.Status.RelatedObjects = related
```
Miciah: This should append to the `[]configv1.ObjectReference` that is defined in `initializeClusterOperator`.

ironcladlou: I see, good catch

ironcladlou: Fixed

ironcladlou:
@Miciah PTAL

pkg/operator/controller/status.go
```diff
@@ -94,7 +106,11 @@ func initializeClusterOperator(co *configv1.ClusterOperator, nsName string) {
 			Status: configv1.ConditionUnknown,
 		},
 	}
-	co.Status.RelatedObjects = []configv1.ObjectReference{
+	co.Status.RelatedObjects = defaultRelatedObjects(nsName)
```
Miciah: This is probably not needed because IIRC the status is ignored on create.

pkg/operator/controller/status.go
```diff
@@ -53,6 +53,18 @@ func (r *reconciler) syncOperatorStatus() error {
 	if err != nil {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
+
+	related := defaultRelatedObjects(ns.Name)
+	for _, ingress := range ingresses {
+		related = append(related, configv1.ObjectReference{
+			Group:     operatorv1.GroupName,
+			Resource:  "IngressController",
+			Namespace: r.Namespace,
```
Miciah: Semantically it might make more sense to use `ingress.Namespace` here.  Admittedly, it makes no practical difference.

pkg/operator/controller/status.go
ironcladlou: Makes sense — removed/simplified.

pkg/operator/controller/status.go
ironcladlou: Yeah... file this under namespace handling tech debt...

pkg/operator/controller/status.go
```diff
@@ -53,6 +53,27 @@ func (r *reconciler) syncOperatorStatus() error {
 	if err != nil {
 		return fmt.Errorf("failed to get operator state: %v", err)
 	}
+
+	related := []configv1.ObjectReference{
+		{
+			Resource: "namespaces",
+			Name:     "openshift-ingress-operator",
```
Miciah: This should be `r.Namespace`, but we were already hard-coding it before, so it can likewise be filed under "namespace handling tech debt".

PR #251, opened 2019-06-13
dns: introduce CRD DNS architecture and DNS status reporting

ironcladlou:
* Introduce s dnsrecord.ingress.operator.openshift.io resource type to represent
DNS records in a Kube-native API.

* Refactor ingress controller to manage dnsrecords rather than inferring DNS
record requirements indirectly through services.

* Teach the ingresscontroller status computer about dnsrecords so DNS management
status can surface through all operator CRDs in a way consistent with other
resources (e.g. services).

* Introduce a DNS controller which syncs dnsrecords with the cluster's
configured DNS zones, reporting status on the dnsrecord.


ironcladlou:
Here are some status examples:

```yaml
kind: IngressController
metadata:
  name: default
spec: {}
status:
  availableReplicas: 2
  conditions:
  - lastTransitionTime: "2019-06-13T15:26:46Z"
    status: "True"
    type: Available
  - lastTransitionTime: "2019-06-13T15:26:16Z"
    message: The endpoint publishing strategy supports a managed load balancer
    reason: WantedByEndpointPublishingStrategy
    status: "True"
    type: LoadBalancerManaged
  - lastTransitionTime: "2019-06-13T15:26:19Z"
    message: The LoadBalancer service is provisioned
    reason: LoadBalancerProvisioned
    status: "True"
    type: LoadBalancerReady
  - lastTransitionTime: "2019-06-13T15:26:16Z"
    message: DNS management is supported and zones are specified in the cluster DNS
      config.
    reason: Normal
    status: "True"
    type: DNSManaged
  - lastTransitionTime: "2019-06-13T17:29:32Z"
    reason: NoFailedZones
    status: "True"
    type: DNSReady
  domain: apps.dmace.devcluster.openshift.com
  endpointPublishingStrategy:
    type: LoadBalancerService
  selector: ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default

kind: DNSRecord
metadata:
  creationTimestamp: "2019-06-13T16:00:59Z"
  finalizers:
  - operator.openshift.io/ingress
  generation: 1
  labels:
    ingresscontroller.operator.openshift.io/owning-ingresscontroller: default
  name: default-wildcard
  ownerReferences:
  - apiVersion: operator.openshift.io/v1
    controller: true
    kind: IngressController
    name: default
    uid: 953bfbd0-8def-11e9-b5d6-02915d60e384
  resourceVersion: "58271"
  selfLink: /apis/ingress.operator.openshift.io/v1/namespaces/openshift-ingress-operator/dnsrecords/default-wildcard
  uid: 6fdb10e0-8df4-11e9-b5d6-02915d60e384
spec:
  dnsName: '*.apps.dmace.devcluster.openshift.com'
  recordType: ALIAS
  targets:
  - a95dad0368def11e9b5d602915d60e38-93147362.us-east-2.elb.amazonaws.com
status:
  zones:
  - dnsZone:
      tags:
        Name: dmace-rl775-int
        kubernetes.io/cluster/dmace-rl775: owned
  - dnsZone:
      id: Z3URY6TWQ91KVV
```

ironcladlou:
Here's what things look like when dnsrecord status doesn't report any zones:

```yaml
kind: IngressController
metadata:
  creationTimestamp: "2019-06-13T15:26:15Z"
  finalizers:
  - ingresscontroller.operator.openshift.io/finalizer-ingresscontroller
  generation: 1
  name: default
  resourceVersion: "58272"
  selfLink: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/default
  uid: 953bfbd0-8def-11e9-b5d6-02915d60e384
spec: {}
status:
  availableReplicas: 2
  conditions:
  - lastTransitionTime: "2019-06-13T15:26:46Z"
    status: "True"
    type: Available
  - lastTransitionTime: "2019-06-13T15:26:16Z"
    message: The endpoint publishing strategy supports a managed load balancer
    reason: WantedByEndpointPublishingStrategy
    status: "True"
    type: LoadBalancerManaged
  - lastTransitionTime: "2019-06-13T15:26:19Z"
    message: The LoadBalancer service is provisioned
    reason: LoadBalancerProvisioned
    status: "True"
    type: LoadBalancerReady
  - lastTransitionTime: "2019-06-13T15:26:16Z"
    message: DNS management is supported and zones are specified in the cluster DNS
      config.
    reason: Normal
    status: "True"
    type: DNSManaged
  - lastTransitionTime: "2019-06-13T15:26:19Z"
    message: The record isn't present in any zones.
    reason: NoZones
    status: "False"
    type: DNSReady
  domain: apps.dmace.devcluster.openshift.com
  endpointPublishingStrategy:
    type: LoadBalancerService
  selector: ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default

kind: DNSRecord
metadata:
  creationTimestamp: "2019-06-13T16:00:59Z"
  finalizers:
  - operator.openshift.io/ingress
  generation: 1
  labels:
    ingresscontroller.operator.openshift.io/owning-ingresscontroller: default
  name: default-wildcard
  ownerReferences:
  - apiVersion: operator.openshift.io/v1
    controller: true
    kind: IngressController
    name: default
    uid: 953bfbd0-8def-11e9-b5d6-02915d60e384
  resourceVersion: "39161"
  selfLink: /apis/ingress.operator.openshift.io/v1/namespaces/openshift-ingress-operator/dnsrecords/default-wildcard
  uid: 6fdb10e0-8df4-11e9-b5d6-02915d60e384
spec:
  dnsName: '*.apps.dmace.devcluster.openshift.com'
  recordType: ALIAS
  targets:
  - a95dad0368def11e9b5d602915d60e38-93147362.us-east-2.elb.amazonaws.com
status: {}
```

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,157 @@
...
+	if dnsConfig.Spec.PrivateZone != nil {
+		zones = append(zones, *dnsConfig.Spec.PrivateZone)
+	}
+	if dnsConfig.Spec.PublicZone != nil {
+		zones = append(zones, *dnsConfig.Spec.PublicZone)
+	}
+
+	// If the DNS record was deleted, clean up and return.
+	if record.DeletionTimestamp != nil {
+		for i := range zones {
```
ironcladlou: Should we look at record.Status.Zones instead?

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,157 @@
...
+			err := r.dnsProvider.Delete(record, zone)
+			if err != nil {
+				errs = append(errs, err)
+			} else {
+				log.Info("provider deleted record %s in zone %v", record.Name, zone)
+			}
+		}
+		if len(errs) == 0 {
+			updated := record.DeepCopy()
+			if slice.ContainsString(updated.Finalizers, "operator.openshift.io/ingress") {
```
ironcladlou: Use a constant for the annotation

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,157 @@
...
+			})
+			// TODO: return a retry result?
+		}
+		statuses = append(statuses, iov1.DNSZoneStatus{
+			DNSZone:    zone,
+			Conditions: conditions,
+		})
+	}
+
+	// TODO: only update if status changed
```
ironcladlou: Implement this and the AWS client caching can be deleted

pkg/operator/controller/ingress_status.go
```diff
@@ -205,3 +209,80 @@ func getEventsByReason(events []corev1.Event, component, reason string) []corev1
 	}
 	return filtered
 }
+
+func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) []operatorv1.OperatorCondition {
+	if dnsConfig.Spec.PublicZone == nil && dnsConfig.Spec.PrivateZone == nil {
+		return []operatorv1.OperatorCondition{
+			operatorv1.OperatorCondition{
+				Type:    operatorv1.DNSManagedIngressConditionType,
```
ironcladlou: In the e2e test, if this condition is true, wait for DNS on the ingress status to be ready

pkg/operator/controller/dns/controller.go
ironcladlou: Fixed

pkg/operator/controller/dns/controller.go
ironcladlou: Improved the TODO but don't intend to fix now because:

1. It's a bit of work to get right
2. It involves refactoring the DNS provider implementations
3. We're already better off than before because we now only call ensure on creation (so once during startup, and once during ingresscontroller creation), so fixing the last problem here has low ROI for this particular PR (which is already huge)

knobunc:
The CI failure was caused by https://bugzilla.redhat.com/show_bug.cgi?id=1720678.  It should be fixed shortly.

ironcladlou:
I think this is actually ready for review

pkg/dns/azure/dns.go
```diff
@@ -102,7 +105,7 @@ func (m *manager) Delete(record *dns.Record) error {
 		context.TODO(),
 		*targetZone,
 		client.ARecord{
-			Address: record.ARecord.Address,
+			Address: record.Spec.Targets[0],
```
Miciah: As above, this needs a "TODO".

pkg/operator/controller/controller.go
```diff
@@ -312,21 +310,35 @@ func (r *reconciler) enforceIngressFinalizer(ingress *operatorv1.IngressControll
...
+	// Delete the wildcard DNS record, and block ingresscontroller finalization
+	// until the dnsrecord has been finalized.
+	if err := r.deleteWildcardDNSRecord(ingress); err != nil {
+		return fmt.Errorf("failed to delete wildcard dnsrecord: %v", err)
+	}
+	if record, err := r.currentWilcardDNSRecord(ingress); err != nil {
+		return fmt.Errorf("failed to get current wildcard dnsrecord: %v", err)
+	} else {
+		if record != nil {
+			log.V(1).Info("waiting for wildcard dnsrecord to be deleted", "dnsrecord", record.Name)
```
Miciah: Elsewhere we pass `record` in its entirety to `log.Info`.

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,135 @@
...
 
-func newAliasRecord(domain, target string, zone configv1.DNSZone) *dns.Record {
-	return &dns.Record{
-		Zone: zone,
-		Type: dns.ALIASRecord,
-		Alias: &dns.AliasRecord{
-			Domain: domain,
-			Target: target,
-		},
+	desired := desiredWildcardRecord(ic, service)
```
Miciah: Could `service.Status.LoadBalancer.Ingress` change from non-empty to empty, and should we handle that case?  Probably that would only happen in case of an intermittent outage, in which case it leaving the old record might be preferable.  In any case, I'd like us to make a deliberate choice, and document it.

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,135 @@
...
-		Type: dns.ARecordType,
-		ARecord: &dns.ARecord{
-			Domain:  domain,
-			Address: target,
-		},
+	if desired != nil && current == nil {
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return nil, fmt.Errorf("failed to create dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
+		}
+		log.Info("created dnsrecord", "namespace", desired.Namespace, "name", desired.Name)
```
Miciah: Could we log `desired` in its entirety?

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,135 @@
...
-				records = append(records, newARecord(name, ingress.IP, zone))
-			}
+
+	name := WildcardDNSRecordName(ic)
+	domain := fmt.Sprintf("*.%s", ic.Status.Domain)
+	var target string
+	var recordType string
+
+	if len(ingress.Hostname) > 0 {
+		recordType = "ALIAS"
```
Miciah: "ALIAS" is AWS-specific.  Let's document that and think about where the logic to choose between "ALIAS" or "CNAME" belongs.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,167 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	ingressName, ok := record.Labels[manifests.OwningIngressControllerLabel]
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("missing owner label")
+	}
+	ingress := &operatorv1.IngressController{}
+	if err := r.cache.Get(context.TODO(), types.NamespacedName{Namespace: record.Namespace, Name: ingressName}, ingress); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get dnsrecord for record %s: %v", record.Name, err)
```
Miciah: Is it a deliberate design choice not to handle orphaned `DNSRecord` resources?

pkg/api/v1/types.go
```diff
@@ -0,0 +1,71 @@
...
+	Status DNSRecordStatus `json:"status,omitempty"`
+}
+
+type DNSRecordSpec struct {
+	// The hostname of the DNS record
+	DNSName string `json:"dnsName,omitempty"`
+	// The targets the DNS record points to
+	Targets []string `json:"targets,omitempty"`
+	// RecordType type of record, e.g. CNAME, A, SRV, TXT etc
+	RecordType string `json:"recordType,omitempty"`
```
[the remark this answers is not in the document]
ironcladlou: Fixed and switched A to CNAME

pkg/api/v1/types.go
```diff
@@ -0,0 +1,71 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// Record represents a DNS record.
```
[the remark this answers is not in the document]
ironcladlou: Fixed

pkg/operator/controller/controller.go

pkg/operator/controller/controller_dns.go

pkg/operator/controller/dns/controller.go
ironcladlou: I don't know that this state implies the `dnsrecord` itself an orphan. It usually means the resource was deleted and due to timing we have a benign leftover entry to flush from the controller queue.

Also, how would the resource have been deleted from storage without having passed through finalization in `delete()`? So I'm also not sure there's any orphan in the cloud provider, either.

Help me understand the concern?

pkg/operator/controller/controller_dns.go
ironcladlou: Added some comments

pkg/dns/azure/dns.go
```diff
@@ -76,7 +79,7 @@ func (m *manager) Ensure(record *dns.Record) error {
...
```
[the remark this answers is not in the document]
ironcladlou: Fixed

pkg/dns/aws/dns.go
```diff
@@ -223,33 +224,29 @@ const (
...
+func (m *Provider) change(record *iov1.DNSRecord, zone configv1.DNSZone, action action) error {
+	if record.Spec.RecordType != "ALIAS" {
+		return fmt.Errorf("unsupported record type %s", record.Spec.RecordType)
 	}
-	alias := record.Alias
-	if alias == nil {
-		return fmt.Errorf("missing alias record")
-	}
-	domain, target := alias.Domain, alias.Target
+	domain, target := record.Spec.DNSName, record.Spec.Targets[0]
```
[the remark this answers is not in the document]
ironcladlou: Done

pkg/dns/azure/dns.go
ironcladlou: Done

pkg/operator/controller/controller_dns.go
ironcladlou: Added a TODO with these considerations. Not going to be able to solve it here yet.

danehans:
@ironcladlou here is the status of the default wildcard `DNSRecord` when the record already exists:
```
    - conditions:
      - lastTransitionTime: 2019-06-18T16:10:28Z
        message: "The DNS provider failed to ensure the record: failed to update alias
          in zone Z3URY6TWQ91KVV: couldn't update DNS record in zone Z3URY6TWQ91KVV:
          InvalidChangeBatch: [RRSet of type A with DNS name \\052.apps.dhansen.devcluster.openshift.com.
          is not permitted because a conflicting RRSet of type  CNAME with the same
          DNS name already exists in zone devcluster.openshift.com.]\n\tstatus code:
          400, request id: 96a43bad-91e3-11e9-898b-19a2d7b35b17"
        reason: ProviderError
        status: "True"
        type: Failed
```

Do you feel this `DNSRecord` status condition is accurate or should this type of error indicate a `False` status?

danehans:
@ironcladlou I also see the operator continuing to reconcile the default `IngressController`, even after the resource has been successfully reconciled:
```
2019-06-18T09:18:06.963-0700	INFO	operator.controller	controller/controller.go:117	reconciling	{"request": "openshift-ingress-operator/default"}
2019-06-18T09:18:07.958-0700	DEBUG	operator.init.controller-runtime.controller	controller/controller.go:236	Successfully Reconciled	{"controller": "ingress_controller", "request": "openshift-ingress-operator/default"}
2019-06-18T09:18:07.958-0700	INFO	operator.controller	controller/controller.go:117	reconciling	{"request": "openshift-ingress-operator/default"}
2019-06-18T09:18:08.929-0700	DEBUG	operator.init.controller-runtime.controller	controller/controller.go:236	Successfully Reconciled	{"controller": "ingress_controller", "request": "openshift-ingress-operator/default"}
...
```

danehans:
It appears https://github.com/openshift/cluster-ingress-operator/pull/251#issuecomment-503206724 and https://github.com/openshift/cluster-ingress-operator/pull/251#issuecomment-503210069 are related. Reconciliation for the default `IngressController` stopped at `Successfully Reconciled	{"controller": "ingress_controller", "request": "openshift-ingress-operator/default"}` after I deleted the default `IngressController`, manually removed the associated wildcard record from the public hosted zone and restarted the operator.

danehans:
@ironcladlou after deleting the default `IngressController`, the associated `DNSRecord` resource is deleted. However, the wildcard records still exist in the public and private hosted zones. Is this expected behavior?

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (r *reconciler) delete(record *iov1.DNSRecord) error {
+	var errs []error
+	for i := range record.Status.Zones {
+		zone := record.Status.Zones[i].DNSZone
+		err := r.dnsProvider.Delete(record, zone)
+		if err != nil {
+			errs = append(errs, err)
+		} else {
+			log.Info("provider deleted record %s in zone %v", record.Name, zone)
```
danehans: I believe L154 should be something like:
```
log.Info("provider deleted record",  "name", record.Name, "zone", zone)
```

danehans:
@ironcladlou Here is the default-wildcard `DNSRecord`:
```
spec:
  dnsName: '*.apps.dhansen.devcluster.openshift.com'
  recordType: CNAME
  targets:
  - aaf2e0f3691ea11e98be2068844bff81-651064175.us-west-2.elb.amazonaws.com
status:
  zones:
  - dnsZone:
      tags:
        Name: dhansen-z9lgn-int
        kubernetes.io/cluster/dhansen-z9lgn: owned
  - dnsZone:
      id: Z3URY6TWQ91KVV
```

A couple things come to mind. First, the spec indicates a `CNAME` record type but an `Alias` record is created since AWS is the configured provider. This may cause confusion. Secondly, the status provides zone details, but no provider information. IMO it's beneficial to indicate the provider associated with the zones to provide a complete picture. Thoughts?

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
-
+// desiredDNSRecords will return any necessary wildcard DNS records for the
+// ingresscontroller.
+//
+// For now, if the service has more than one .status.loadbalancer.ingress, only
+// the first will be used.
+//
+// TODO: If .status.loadbalancer.ingress is processed once as non-empty and then
+// later becomes empty, what should we do? Currently we'll treat it as an intent
+// to not have a desired record.
```
Miciah: `desiredWildcardRecord` will return nil, but `ensureWildcardDNSRecord` will not delete the current record if `desired` is nil, which is what I was getting at when I suggested adding a comment.

danehans:
@ironcladlou disregard https://github.com/openshift/cluster-ingress-operator/pull/251#issuecomment-503215841. I no longer see the records in Route53 after deleting the default `IngressController`. The behavior in https://github.com/openshift/cluster-ingress-operator/pull/251#issuecomment-503215841 may have been related to the other comments I provided.

pkg/operator/controller/controller_dns.go
ironcladlou: delete is deliberately not handled during ensure in this case, so that part... I think mutations of .status.ingress ingress are a larger topic at this point. Unless we compute desired from a persisted "last observed" state I think there are probably more than one of these edge cases we don't coherently account for...

pkg/operator/controller/dns/controller.go
ironcladlou: Fixed

ironcladlou:
>A couple things come to mind. First, the spec indicates a CNAME record type but an Alias record is created since AWS is the configured provider. This may cause confusion.

Agree — not sure yet how to resolve it in the abstract. The fact we're using an A+Alias in Route53 is a provider detail, and one we aren't exposing any configuration around, so maybe it's a doc gap in the meantime? I wonder if this api should be in an alpha or beta group or something?

>Secondly, the status provides zone details, but no provider information. IMO it's beneficial to indicate the provider associated with the zones to provide a complete picture. Thoughts?

Agree with this, but would like to address it in a followup if that's okay since we don't actually need it yet.

pkg/operator/controller/dns/controller.go
Miciah: A separate problem I failed to notice before: `failed to get dnsrecord` should be `failed to get ingresscontroller`.

Miciah: There is a possible state that we need to consider: a `DNSRecord` exists with no corresponding `IngressController`.  If you want to say that we will do nothing to try to recover from this state because the only ways to get there involve, say, etcd corruption or user tinkering, then that is fine, but I want to make sure we don't simply ignore the case.

ironcladlou: Fixed

ironcladlou: Okay, so I tried to split the difference for now. I handle not found by logging a specific message, added a TODO, and made the retry far less aggressive in this case. For other errors, I retry somewhat less aggressively since those are likely transient.

ironcladlou:
Went ahead and squashed

pkg/api/v1/types.go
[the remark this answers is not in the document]
Miciah: You added a type definition for `DNSRecordType` but continue to use `string` instead of using `DNSRecordType`.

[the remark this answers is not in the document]
ironcladlou: Actually intentional for now due to a random issue I was having with generators that will take some time to run down — I need to do a followup here for deepcopy scripts and cleaning up these types.

Also file under: should this be a "v1" API? It's not yet set in stone.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("missing owner label")
+	}
+	ingress := &operatorv1.IngressController{}
+	if err := r.cache.Get(context.TODO(), types.NamespacedName{Namespace: record.Namespace, Name: ingressName}, ingress); err != nil {
+		if errors.IsNotFound(err) {
+			// TODO: what should we do here? something upstream could detect and delete the orphan? add new conditions?
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, fmt.Errorf("ingresscontroller %s not found for record %s", ingressName, record.Name)
+		} else {
+			return reconcile.Result{RequeueAfter: 15 * time.Second}, fmt.Errorf("failed to get ingresscontroller %s for record %s: %v", ingressName, record.Name, err)
```
Miciah: You need a nil error return value in order for `RequeueAfter` to take effect.

ironcladlou: Phew, good catch! My approach is now to log and retry without returning a reconciler error.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get dns 'cluster': %v", err)
+	}
+
+	record := &iov1.DNSRecord{}
+	if err := r.client.Get(context.TODO(), request.NamespacedName, record); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("dnsrecord not found; reconciliation will be skipped", "request", request)
+			return reconcile.Result{}, nil
+		} else {
```
Miciah: Can promote the `else` block and delete `else` since the `if` block ends with a `return`.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,178 @@
...
+		if err := r.delete(record); err != nil {
+			log.Error(err, "failed to delete dnsrecord; will retry", "dnsrecord", record)
+			return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
+		}
+		return reconcile.Result{}, nil
+	}
+
+	ingressName, ok := record.Labels[manifests.OwningIngressControllerLabel]
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("missing owner label")
```
Miciah: Is it useful to requeue in this case?

pkg/operator/controller/controller_lb.go
```diff
@@ -97,44 +92,24 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
...
 
-// finalizeLoadBalancerService deletes any DNS entries associated with any
-// current LB service associated with the ingresscontroller and then finalizes the
-// service.
-func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressController, dnsConfig *configv1.DNS) error {
+// finalizeLoadBalancerService removes finalizers from any LB service. This used
+// to be to help with DNS cleanup, but now that's no longer necessary, and so we
+// just need to clear the finalizer which might exist on existing resources.
+//
+// TODO: How can we delete this code?
```
Miciah: Call `finalizeLoadBalancerService` on every ingresscontroller when the operator first starts, then delete that logic after a couple releases?

pkg/operator/controller/dns/controller.go
ironcladlou: Fixed

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
-	"github.com/openshift/cluster-ingress-operator/pkg/dns"
 
 	corev1 "k8s.io/api/core/v1"
 
-	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ensureDNS will create DNS records for the given LB service. If service is
```
Miciah: `ensureDNS`→`ensureWildcardDNSRecord`.

pkg/operator/controller/dns/controller.go
ironcladlou: Added some logging and infrequent retries here just in case? Not clear cut what the best behavior is around invalid states in general (should we surface those in a way that's more easily analyzed? Like, in this case we're literally seeing corruption/invalid states, and it seems like we'd want to do more than just log and wait for the problem to fix itself).

Seems like a systemic problem we need to address

pkg/operator/controller/controller_lb.go
Miciah: Even before that, we can delete `enforceIngressFinalizer` now, right?

pkg/operator/controller/controller_dns.go

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
-		Type: dns.ARecordType,
-		ARecord: &dns.ARecord{
-			Domain:  domain,
-			Address: target,
-		},
+	if desired != nil && current == nil {
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return nil, fmt.Errorf("failed to create dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
+		}
+		log.Info("created dnsrecord", "namespace", desired.Namespace, "dnsrecord", desired)
```
Miciah: Does the log message need both `desired.Namespace` and `desired`?

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
 	}
-	if dnsConfig.Spec.PublicZone != nil {
-		zones = append(zones, *dnsConfig.Spec.PublicZone)
+
+	ingress := service.Status.LoadBalancer.Ingress[0]
+
+	// Quick sanity check since we don't know how to handle both being set (is
+	// that even a valid state?)
+	if len(ingress.Hostname) > 0 && len(ingress.IP) > 0 {
+		return nil
```
Miciah: We should log something if we are skipping publishing because of an unsupported state.

pkg/operator/controller/controller_dns.go
ironcladlou: Nope, fixed

pkg/operator/controller/controller_dns.go
ironcladlou: Agree, but I'm not sure this is the place to detect and report the invalid state. Lots of other cases like this too

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
+
+	name := WildcardDNSRecordName(ic)
+	domain := fmt.Sprintf("*.%s", ic.Status.Domain)
+	var target string
+	var recordType string
+
+	if len(ingress.Hostname) > 0 {
+		recordType = iov1.CNAMERecordType
+		target = ingress.Hostname
+	} else {
```
Miciah: Could the service have `.Status.LoadBalancer.Ingress[0]` with empty `.Hostname` and `.IP`?

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
+		},
+		Spec: iov1.DNSRecordSpec{
+			DNSName:    domain,
+			Targets:    []string{target},
+			RecordType: recordType,
+		},
+	}
+}
+
+func (r *reconciler) currentWilcardDNSRecord(ic *operatorv1.IngressController) (*iov1.DNSRecord, error) {
```
Miciah: `currentWilcardDNSRecord`→`currentWildcardDNSRecord`.

pkg/operator/controller/controller_dns.go
ironcladlou: Not sure, there are a lot of unresolved Q's about that status field and our handling of them... but the environments we support are predictable enough to support our current assumptions, so I'm not really focused on solving them in this PR

pkg/operator/controller/controller_dns.go

ironcladlou:
Anything substantial left blocking merge?

Miciah:
> Anything substantial left blocking merge?

Are we possibly going to want to extricate `dns_controller` from the operator, and its reconciliation logic from  `IngressController`?  If so, will need to have a plan for migrating `DNSRecord`, for example to add zones to its spec (which we _should_ be able to achieve with a one-off migration on start-up).

ironcladlou:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/251/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1005

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/251/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1005/artifacts/e2e-aws-operator/nodes.json

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/251/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1005/artifacts/e2e-aws-operator/must-gather/namespaces/openshift-ingress/core/events.yaml

Looks like the ingresscontroller deployment's pods weren't being scheduled, check out the scheduler events complaining about affinity rules... very suspicious

ironcladlou:
>Are we possibly going to want to extricate dns_controller from the operator, and its reconciliation logic from IngressController? If so, will need to have a plan for migrating DNSRecord, for example to add zones to its spec (which we should be able to achieve with a one-off migration on start-up).

Maybe? Pretty far out in any case

ironcladlou:
Going to retest to see if the scheduling issue is a flake — given the unexplained co-location I observed on my new cluster yesterday I'm getting nervous something is wrong there.

Miciah:
> Looks like the ingresscontroller deployment's pods weren't being scheduled, check out the scheduler events complaining about affinity rules... very suspicious

Also, the container is inexplicably killed at 2019-06-18T21:51:02Z.

Miciah:
Actually, it's the operand's container that is getting killed.  I think I'm confused because Prow is doubling the test output and injecting misleading timestamps.

ironcladlou:
Okay, looks like a flake after all. Seems unlikely to be related to anything going on in this PR.

pkg/api/v1/types.go
```diff
@@ -0,0 +1,79 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// DNSRecord represents a DNS record.
```
Miciah: How about a brief statement such as the following? "This resource kind is intended for internal use by the operator only."

pkg/operator/controller/ingress_status.go
```diff
@@ -205,3 +209,80 @@ func getEventsByReason(events []corev1.Event, component, reason string) []corev1
...
+			Type:    operatorv1.DNSReadyIngressConditionType,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "NoZones",
+			Message: "The record isn't present in any zones.",
+		})
+	case len(wildcardRecord.Status.Zones) > 0:
+		failedZones := []configv1.DNSZone{}
+		for _, zone := range wildcardRecord.Status.Zones {
+			for _, cond := range zone.Conditions {
+				if cond.Type == iov1.DNSRecordFailedConditionType {
```
Miciah: Need the following: ` && cond.Status == string(operatorv1.ConditionTrue)`.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	// successfully ensured. Then the DNS provider wouldn't need such aggressive
+	// internal caching.
+	result := reconcile.Result{}
+	statuses := []iov1.DNSZoneStatus{}
+	for i := range zones {
+		zone := zones[i]
+		conditions := []iov1.DNSZoneCondition{}
+		err := r.dnsProvider.Ensure(record, zone)
+		if err != nil {
+			conditions = append(conditions, iov1.DNSZoneCondition{
```
Miciah: Didn't we decide to set `Failed=False` when `err` is nil?

pkg/operator/controller/controller_dns.go
```diff
@@ -1,89 +1,139 @@
...
-			Domain:  domain,
-			Address: target,
-		},
+	if desired != nil && current == nil {
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return nil, fmt.Errorf("failed to create dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
+		}
+		log.Info("created dnsrecord", "dnsrecord", desired)
+		return desired, nil
 	}
```
Miciah: I am not sure we can do without update handling.  I would like to test the following on AWS:

1. Scale the default ingress controller to 1.
2. Look up the AZ of the sole ingress controller pod.
3. Kill ingress controller pods till one appears in a different AZ, which should cause the ELB's DNS name to change.
4. Check whether the wildcard DNS name still resolves.

pkg/operator/controller/controller_dns_test.go
```diff
@@ -3,199 +3,111 @@ package controller
...
-			},
-			expect: []record{
-				{typ: dns.ALIASRecord, name: "*.apps.openshift.example.com", target: "lb.cloud.example.com", zone: publicZone},
-			},
+			expect:      nil,
 		},
 		{
-			description: "private only A",
+			description: "no ingresses",
+			domain:      "",
```
Miciah: This test case should have a non-empty value for `domain` because an empty value also causes a nil result.

pkg/operator/controller/controller_dns_test.go
```diff
@@ -3,199 +3,111 @@ package controller
...
-			},
-			expect: []record{
-				{typ: dns.ARecordType, name: "*.apps.openshift.example.com", target: "192.0.2.1", zone: privateZone},
-			},
+			ingresses:   []corev1.LoadBalancerIngress{},
+			expect:      nil,
 		},
 		{
-			description: "global ALIAS",
+			description: "hostname to ALIAS record",
```
Miciah: `ALIAS`→`CNAME`.

pkg/operator/controller/dns/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	return c, nil
+}
+
+type reconciler struct {
+	client            client.Client
+	cache             cache.Cache
+	recorder          record.EventRecorder
+	operatorNamespace string
+	dnsProvider       dns.Provider
+	dnsConfig         *configv1.DNS
```
Miciah: Half of these fields are not used:

* `recorder` is not used (but maybe it should be).
* `operatorNamespace` is not used and probably should be deleted.
* `dnsConfig` is not used and should be deleted.

pkg/operator/controller/controller_lb.go
ironcladlou: I don't think so? Do you mean removing the finalizer from new LB services we create? The entire cleanup strategy revolves around ingresscontroller finalizers.

pkg/api/v1/types.go
ironcladlou: Would like to run this entire thing by API folks before code freeze. I'm not even sure if a v1 version is appropriate. We can tweak it.

pkg/operator/controller/ingress_status.go
ironcladlou: Fixed

pkg/operator/controller/dns/controller.go
ironcladlou: I don't think we explicitly decided. If we publish True when there's a problem, does that imply we must publish False when there's not a problem? There's no internal reason we need the condition when things are normal, not sure what the conventions are here. We can change in a followup.

pkg/operator/controller/controller_dns_test.go
ironcladlou: Fixed

pkg/operator/controller/controller_dns_test.go
ironcladlou: Fixed

pkg/operator/controller/dns/controller.go
ironcladlou: Fixed

pkg/operator/controller/controller_lb.go
Miciah: Sorry, I mixed up my finalizers.  I mean remove the finalizer from any existing LB service on start-up, and stop adding one in `desiredLoadBalancerService`.

pkg/api/v1/types.go
Miciah: We could add something about how it is unstable too.  The idea is that anyone looking at the resource using `oc explain` or OpenShift Console would at least have some idea of what the thing is and know not to mess with it.

pkg/operator/controller/controller_dns.go
Miciah: I performed the test, and the ELB's DNS name does not change; only the address to which it resolves changes, and because we use an alias on AWS, the wildcard DNS record likewise changes over to the new address without requiring an update by the operator.

DanyC97:
@ironcladlou qq if you don't mind - is this going to work for UPI too on non-AWS env?

ironcladlou:
@DanyC97 
>qq if you don't mind - is this going to work for UPI too on non-AWS env?

The managed wildcard DNS record remains supported today only on AWS and Azure IPI.

This change is a functionally transparent internal refactor and doesn't imply any change in our DNS management capabilities as expressed by the public API.

PR #254, opened 2019-06-18
NE-167: Implement internal load-balancer API

Miciah:
~This PR depends on https://github.com/openshift/cluster-ingress-operator/pull/251.~

#### Bump `openshift/api`

Bump to github.com/openshift/api@2024105dd4c3cc30148d4127545eec7b21a53b65 to get the load-balancer scope API.

#### Honor the load balancer strategy's scope parameter

Respect the "scope" parameter for the ingress controller's load balancer strategy in order to allow choosing between an external (the default) or an internal load balancer.

* `pkg/operator/controller/controller.go` (`enforceEffectiveEndpointPublishingStrategy`): Default `status.endpointPublishingStrategy.loadBalancer.scope` to "External".
* `pkg/operator/controller/controller_lb.go` (`internalLBAnnotations`): New variable.  Map platform type to the annotation for that platform that makes the load balancer internal-only.
(`desiredLoadBalancerService`): Check `status.endpointPublishingStrategy.loadBalancer.scope` and use `internalLBAnnotations` to set the appropriate annotation if the scope is not "External".
* `pkg/operator/controller/dns/controller.go` (`Reconcile`): Omit the public zone if the load balancer is internal-only.
* `test/e2e/operator_test.go` (`TestInternalLoadBalancer`): New test.


Miciah:
`TestInternalLoadBalancer` is getting false positives because of DNS propagation delays.  In a test cluster, when I waited long enough, I eventually got a DNS record that resolved to a public (but non-routable) address.  Writing a robust test may be tricky.

pkg/operator/controller/controller_lb.go
```diff
@@ -19,9 +19,46 @@ import (
...
 	awsLBProxyProtocolAnnotation = "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"
 )
 
+var (
+	// internalLBAnnotation maps platform to the annotation name and value
+	// that tell the cloud provider that is associated with the platform
+	// that the load balancer is internal.
+	//
+	// https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
+	internalLBAnnotation = map[configv1.PlatformType][]string{
```
ironcladlou: I found this structure a little weird compared to:

```go
internalLBAnnotation = map[configv1.PlatformType]map[string]string{
  configv1.AWSPlatformType: {
    "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol": "*",
  },
  // ...
}
```

Which then gives the caller a map to merge with any other label/annotation maps.

But, what you have works, so I just mention it as a matter of style to consider.

pkg/operator/controller/controller_lb.go
```diff
@@ -68,12 +105,22 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
+	if service.Annotations == nil {
+		service.Annotations = map[string]string{}
+	}
 	if infraConfig.Status.Platform == configv1.AWSPlatformType {
-		if service.Annotations == nil {
-			service.Annotations = map[string]string{}
-		}
 		service.Annotations[awsLBProxyProtocolAnnotation] = "*"
 	}
+	if !haveExternalLB {
```
ironcladlou: Random style/readability thing — instead of "if not have external lb then add internal annotation" you could invert to "if internal lb then add internal annotation"

test/e2e/operator_test.go
```diff
@@ -670,3 +672,112 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
...
+
+	// Wait for the load balancer and DNS to be ready.
+	err = waitForIngressControllerCondition(cl, 5*time.Minute, name, defaultAvailableConditions...)
+	if err != nil {
+		t.Errorf("failed to observe expected conditions: %v", err)
+	}
+
+	// Wait for the service to have at least one ingress.
+	lbService := &corev1.Service{}
+	err = wait.PollImmediate(1*time.Second, 2*time.Minute, func() (bool, error) {
```
ironcladlou: Isn't this covered by waitForIngressControllerCondition?

test/e2e/operator_test.go
```diff
@@ -670,3 +672,112 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
...
+		t.Fatalf("failed to get DNS 'cluster': %v", err)
+	}
+
+	name := types.NamespacedName{Namespace: ns, Name: ingressControllerName}
+	domain := name.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newIngressController(name.Name, name.Namespace, domain, operatorv1.LoadBalancerServiceStrategyType)
+	ic.Spec.EndpointPublishingStrategy.LoadBalancer = &operatorv1.LoadBalancerStrategy{
+		Scope: operatorv1.InternalLoadBalancer,
+	}
+	if err := cl.Create(context.TODO(), ic); err != nil {
```
ironcladlou: Move the delete to a defer after this create and you can reduce the branching/nesting level of subsequent checks

test/e2e/operator_test.go
```diff
@@ -670,3 +672,112 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to get LoadBalancer service: %v", err)
+	} else {
+		for _, ingress := range lbService.Status.LoadBalancer.Ingress {
+			if len(ingress.IP) != 0 {
+				if !networkutils.IsPrivateAddress(ingress.IP) {
```
ironcladlou: Is this a safe assumption? I kind of expected you'd assert appropriate k8s metadata (i.e. annotations) since what "private" means is up to the cloud provider.

pkg/operator/controller/controller_lb.go
Miciah: Yeah, I know the string slice is weird.  Using a map seems silly in abstract, but it Go's syntactic sugar makes it look nicer.  Changed.

test/e2e/operator_test.go
Miciah: I need the `lbService` so I can look at its status though.  Are you saying I should omit the polling loop around the `Get`?

pkg/operator/controller/controller_lb.go
Miciah: I went back and forth on this because I am not sure it is always obvious that "internal" implies "not external".  I considered `internalOnly`, but that just moves the negative.  I'll change it to `isInternal` here.  Would you like the same change in `pkg/operator/controller/dns/controller.go` for consistency?

test/e2e/operator_test.go
Miciah: You mean by changing `if  ... { t.Errorf(foo) } else { t.Errorf(bar) }` to `if ... { t.Fatalf(foo) } t.Errorf(bar)`?

test/e2e/operator_test.go
Miciah: I was hoping to test that the thing was in practice internal-only, but that may be impractical, and checking annotations would be a lot easier.  Is this check still useful as an end-to-end test, as opposed to as a unit test?

test/e2e/operator_test.go
ironcladlou: >Are you saying I should omit the polling loop around the Get?

Yeah, sorry, hopefully more clearly: you can just `Get` the service and error if not found, because the preceding higher-layer waitForIngressControllerCondition checks. If the condition check succeeded but the lb service doesn't exist, failing fast seems easy, an we'd want to investigate how that was possible

pkg/operator/controller/controller_lb.go
ironcladlou: Not a big deal, just something that stood out at me, no objection to keeping as-is

test/e2e/operator_test.go
ironcladlou: Yeah, exactly

test/e2e/operator_test.go
ironcladlou: Hmm... my thinking is that our contract with k8s in this regard is that we declare intent for internal by setting known annotations on the service. That's what the docs tell us to do in the API. It doesn't promise that the ingress IP will be internal as defined by IsPrivateAddress (or any other algorithm) or that an IP will be specified instead of a hostname. So I feel like if given our inputs we produce the output "a correctly annotated service" we can call that good enough and trust that we've satisfied our end of the contract with k8s.

If we wanted deeper assertions on the downstream effects of the annotation (e.g. an IP in a certain CIDR when using AWS), those seem platform-specific and less crucial to verifying general correctness here. What do you think?

Miciah: I think that makes sense.  And I suppose the end-to-end test is useful for ensuring the operator sets the right status in `enforceEffectiveEndpointPublishingStrategy` and that `ensureLoadBalancerService` does the right thing with that status.

Miciah:
Rebased.

Miciah:
Timeouts from openshift-sdn and openshift-marketplace health probes and from etcd.

/test e2e-aws-upgrade

test/e2e/operator_test.go
```diff
@@ -671,3 +671,92 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
...
+		annotation["cloud.google.com/load-balancer-type"] = "Internal"
+	case configv1.LibvirtPlatformType:
+		t.Skip("test skipped on libvirt")
+	case configv1.OpenStackPlatformType:
+		annotation["service.beta.kubernetes.io/openstack-internal-load-balancer"] = "true"
+	case configv1.NonePlatformType:
+		t.Skip("test skipped on \"none\" platform type")
+	case configv1.VSpherePlatformType:
+		t.Skip("test skipped on vSphere")
+	}
```
ironcladlou: TODO: Seems like it could be condensed to a lookup from internalLBAnnotations (if that were exposed by the package)

PR #255, opened 2019-06-20
controller: rename some files

ironcladlou:
This just renames some files to look nicer.


ironcladlou:
@Miciah @danehans

Miciah:
I like it.  What do you think of moving these the controller files under a subdirectory of `pkg/operator/controller/`, alongside the other controllers, as follows?
```
pkg/operator/controller/controller.go → pkg/operator/controller/ingress/controller.go
pkg/operator/controller/deployment.go → pkg/operator/controller/ingress/deployment.go
pkg/operator/controller/deployment_test.go → pkg/operator/controller/ingress/deployment_test.go
pkg/operator/controller/dns.go → pkg/operator/controller/ingress/dns.go
pkg/operator/controller/dns_test.go → pkg/operator/controller/ingress/dns_test.go
pkg/operator/controller/internal_service.go → pkg/operator/controller/ingress/internal_service.go
pkg/operator/controller/load_balancer_service.go → pkg/operator/controller/ingress/load_balancer_service.go
pkg/operator/controller/monitoring.go → pkg/operator/controller/ingress/monitoring.go
pkg/operator/controller/ingress_status.go → pkg/operator/controller/ingress/status.go
pkg/operator/controller/ingress_status_test.go → pkg/operator/controller/ingress/status_test.go

pkg/operator/controller
├── certificate
│  ├── ca.go
│  ├── controller.go
│  ├── default_cert.go
│  ├── publish_ca.go
│  └── publish_ca_test.go
├── certificate-publisher
│  ├── controller.go
│  ├── publish_certs.go
│  └── publish_certs_test.go
├── dns
│  └── controller.go
├── ingress
│  ├── controller.go
│  ├── deployment.go
│  ├── deployment_test.go
│  ├── dns.go
│  ├── dns_test.go
│  ├── internal_service.go
│  ├── load_balancer_service.go
│  ├── monitoring.go
│  ├── status.go
│  └── status_test.go
├── names.go
├── status.go
└── status_test.go
```
What I am suggesting would change the package name though and would require fixing a few references and unexported definitions, so it might bee better as a follow-up.

ironcladlou:
>What I am suggesting would change the package name though and would require fixing a few references and unexported definitions, so it might bee better as a follow-up.

That looks pretty good to me — sounds like a followup

Miciah:
> > What I am suggesting would change the package name though and would require fixing a few references and unexported definitions, so it might bee better as a follow-up.
> 
> That looks pretty good to me — sounds like a followup

I took a whack at it with https://github.com/openshift/cluster-ingress-operator/pull/256.  Because the renaming that I suggested puts methods related to status and methods related to ingress in separate packages, they cannot use the same receiver, so I had to split status out into its own controller, which is something we've been wanting to do anyway.

PR #256, opened 2019-06-22
Split ingress and status controllers

Miciah:
Move the ingress controller into its own package, and extract a separate status controller to create and update the `ClusterOperator` resource.

Fix a nil pointer dereference in case the namespace does not exist at the time that the `ClusterOperator` resource's `status` is computed.

Update references to the "operator controller", which is now more properly called the "ingress controller".

* `cmd/ingress-operator/main.go` (`main`): Update references to `UnknownVersionValue`, which is now in the `status` controller package.
* `pkg/operator/controller/certificate/controller.go` (`Reconcile`): Update reference to `IsStatusDomainSet`, which is now in the `ingress` controller package.
* `pkg/operator/controller/controller.go`: Rename...
* `pkg/operator/controller/ingress/controller.go`: ...to this.  Use the correct controller name for logging.
(`Config`): Delete `OperatorReleaseVersion`.
(`Reconcile`): Delete call to `syncOperatorStatus`.
* `pkg/operator/controller/deployment.go`: Rename...
* `pkg/operator/controller/ingress/deployment.go`: ...to this.  Update references to functions in `pkg/operator/controller/names.go`.
* `pkg/operator/controller/deployment_test.go`: Rename...
* `pkg/operator/controller/ingress/deployment_test.go`: ...to this.
* `pkg/operator/controller/dns.go`: Rename...
* `pkg/operator/controller/ingress/dns.go`: ...to this.  Update references to functions in `pkg/operator/controller/names.go`.
* `pkg/operator/controller/dns_test.go`: Rename...
* `pkg/operator/controller/ingress/dns_test.go`: ...to this.
* `pkg/operator/controller/internal_service.go`: Rename...
* `pkg/operator/controller/ingress/internal_service.go`: ...to this.  Update references to functions in `pkg/operator/controller/names.go`.
* `pkg/operator/controller/load_balancer_service.go`: Rename...
* `pkg/operator/controller/ingress/load_balancer_service.go`: ...to this.  Update references to functions in `pkg/operator/controller/names.go`.
* `pkg/operator/controller/monitoring.go`: Rename...
* `pkg/operator/controller/ingress/monitoring.go`: ...to this.  Update references to functions in `pkg/operator/controller/names.go`.
* `pkg/operator/controller/ingress_status.go`: Rename...
* `pkg/operator/controller/ingress/status.go`: ...to this.
* `pkg/operator/controller/ingress_status_test.go`: Rename...
* `pkg/operator/controller/ingress/status_test.go`: ...to this.
* `pkg/operator/controller/names.go` (`caCertSecretName`, `caCertConfigMapName`, `routerCertsGlobalSecretName`): Deleted.
(`controllerDeploymentLabel`): Renamed...
(`ControllerDeploymentLabel`): ...to this.
(`IngressClusterOperatorName`): Moved from `pkg/operator/controller/status.go`. Change type to namespaced name.
(`RouterCASecretName`, `RouterCAConfigMapName`, `RouterCertsGlobalSecretName`): Replace named consts that were only used in these functions with string literals.
* `pkg/operator/controller/status.go`: Rename...
* `pkg/operator/controller/status/controller.go`: ...to this.
(`IngressClusterOperatorName`): Moved to `pkg/operator/controller/names.go`.
(`controllerName`): New constant.
(`log`): New logger.
(`New`): New method.  Create a new status controller, which watches `IngressController` resources and creates and updates `ClusterOperator`.
(`Config`): New type.  Hold configuration for the status controller.
(`reconciler`): New type.  Hold state for the status controller.
(`syncOperatorStatus`): Rename...
(`Reconcile`): ...to this.  Add logging, request parameter, and reconcile result value.  Update reference to `IngressClusterOperatorName`, which is now a function that returns a namespaced name.  Fix a nil pointer dereference if the namespace is not found.
* `pkg/operator/controller/status_test.go`: Rename...
* `pkg/operator/controller/status/controller_test.go`: ...to this.
* `pkg/operator/operator.go`: Update to create both ingress and status controllers.
* `test/e2e/operator_test.go` (`conditionsMatchExpected`): Update reference to `IngressClusterOperatorName`, which is now a function that returns a namespaced name.

### DIFF
--- a/pkg/operator/controller/names.go
+++ b/pkg/operator/controller/names.go
@@ -15,25 +15,18 @@ const (
 	// CA certificate in this namespace.
 	GlobalMachineSpecifiedConfigNamespace = "openshift-config-managed"
 
-	// caCertSecretName is the name of the secret that holds the CA certificate
-	// that the operator will use to create default certificates for
-	// ingresscontrollers.
-	caCertSecretName = "router-ca"
-
-	// caCertConfigMapName is the name of the config map with the public key
-	// for the CA certificate, which the operator publishes for other
-	// operators to use.
-	caCertConfigMapName = "router-ca"
-
-	// routerCertsGlobalSecretName is the name of the secret with the
-	// default certificates and their keys, which the operator publishes for
-	// other operators to use.
-	routerCertsGlobalSecretName = "router-certs"
-
-	// controllerDeploymentLabel identifies a deployment as an ingress controller
+	// ControllerDeploymentLabel identifies a deployment as an ingress controller
 	// deployment, and the value is the name of the owning ingress controller.
-	controllerDeploymentLabel = "ingresscontroller.operator.openshift.io/deployment-ingresscontroller"
+	ControllerDeploymentLabel = "ingresscontroller.operator.openshift.io/deployment-ingresscontroller"
 )
+
+// IngressClusterOperatorName returns the namespaced name of the ClusterOperator
+// resource for the operator.
+func IngressClusterOperatorName() types.NamespacedName {
+	return types.NamespacedName{
+		Name: "ingress",
+	}
+}
 
 // RouterDeploymentName returns the namespaced name for the router deployment.
 func RouterDeploymentName(ci *operatorv1.IngressController) types.NamespacedName {
@@ -44,27 +37,33 @@ func RouterDeploymentName(ci *operatorv1.IngressController) types.NamespacedName
 }
 
 // RouterCASecretName returns the namespaced name for the router CA secret.
+// This secret holds the CA certificate that the operator will use to create
+// default certificates for ingresscontrollers.
 func RouterCASecretName(operatorNamespace string) types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: operatorNamespace,
-		Name:      caCertSecretName,
+		Name:      "router-ca",
 	}
 }
 
 // RouterCAConfigMapName returns the namespaced name for the router CA configmap.
+// The operator uses this configmap to publish the public key for the CA
+// certificate, so that other operators can include it into their trust bundles.
 func RouterCAConfigMapName() types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: GlobalMachineSpecifiedConfigNamespace,
-		Name:      caCertConfigMapName,
+		Name:      "router-ca",
 	}
 }
 
 // RouterCertsGlobalSecretName returns the namespaced name for the router certs
-// secret.
+// secret.  The operator uses this secret to publish the default certificates and
+// their keys, so that the authentication operator can configure the OAuth server
+// to use the same certificates.
 func RouterCertsGlobalSecretName() types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: GlobalMachineSpecifiedConfigNamespace,
-		Name:      routerCertsGlobalSecretName,
+		Name:      "router-certs",
 	}
 }
 
@@ -93,7 +92,7 @@ func IngressControllerDeploymentLabel(ic *operatorv1.IngressController) string {
 func IngressControllerDeploymentPodSelector(ic *operatorv1.IngressController) *metav1.LabelSelector {
 	return &metav1.LabelSelector{
 		MatchLabels: map[string]string{
-			controllerDeploymentLabel: IngressControllerDeploymentLabel(ic),
+			ControllerDeploymentLabel: IngressControllerDeploymentLabel(ic),
 		},
 	}
 }
--- a/cmd/ingress-operator/main.go
+++ b/cmd/ingress-operator/main.go
@@ -15,7 +15,7 @@ import (
 	"github.com/openshift/cluster-ingress-operator/pkg/operator"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	operatorconfig "github.com/openshift/cluster-ingress-operator/pkg/operator/config"
-	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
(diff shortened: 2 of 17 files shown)


ironcladlou:
Took a quick look, and the major thing that stands out at me is this seems to subtly change "when" status is computed in a way that might be worth discussing in depth... today, we recompute operator status only when we reconcile ingresscontrollers, and after we know we've done some sort of processing that might be worth updating status for (or at least, we sort of do, and at least have enough context to inform that decision at the time).

With this change, we would probably update status in parallel with processing the ingresscontroller such that it seems at least possible (not sure how probable) that we'd report status before doing any work that would affect status. Maybe the individual ingresscontroller status update at the end of reconciliation would re-trigger the operator status calc and cause it to catch up?

What I'm wondering is whether operator status calculation specifically should be framed as a reconciliation in response to ingresscontroller events. I'm not sure what a better alternative would be yet, just opening the discussion. For example, maybe the operator status controller could be watching a channel source which other controllers can drop events into when they have done something they know that the operator might be interested in reporting status against.

Miciah:
> With this change, we would probably update status in parallel with processing the ingresscontroller such that it seems at least possible (not sure how probable) that we'd report status before doing any work that would affect status.

Right, that is why I had to fix a nil pointer dereference in case the namespace does not exist at the time of status computation.

> Maybe the individual ingresscontroller status update at the end of reconciliation would re-trigger the operator status calc and cause it to catch up?

That is my understanding.

> What I'm wondering is whether operator status calculation specifically should be framed as a reconciliation in response to ingresscontroller events. I'm not sure what a better alternative would be yet, just opening the discussion. For example, maybe the operator status controller could be watching a channel source which other controllers can drop events into when they have done something they know that the operator might be interested in reporting status against.

With this PR, the operator should be updating the `ClusterOperator` resource's `status` both in response to the creation or update of an `IngressController` and also when the ingress controller updates the `IngressController` resource's `status` field.  The former is new, and means we update at a time an earlier time than we did without this change.  The latter means we still update status at the same time that we did without this change.

This PR also gives us cleaner separation of concerns the flexibility to add watches on other resources if we want to information or events related to those other resources.

The only downside I see is that by updating the status more frequently, we may put more load on the API.  Is this a significant concern?

Miciah:
error: unable to extract layer sha256:2382db1fd05cdadcd290fc4d5d2266f4559675a743fe1ee222f6239a33ed4b3f from registry.svc.ci.openshift.org/ci-op-6tllvscn/stable-initial@sha256:b0111f37d892a5a70d3ab6c0714861ed9e1f1e51166fdda6f6f317d3128462a0: stream error: stream ID 83; INTERNAL_ERROR

/test e2e-aws-upgrade

pkg/operator/controller/names.go
```diff
@@ -15,26 +15,19 @@ const (
...
 	// deployment, and the value is the name of the owning ingress controller.
-	controllerDeploymentLabel = "ingresscontroller.operator.openshift.io/deployment-ingresscontroller"
+	ControllerDeploymentLabel = "ingresscontroller.operator.openshift.io/deployment-ingresscontroller"
 )
 
+// IngressClusterOperatorName returns the namespaced name of the ClusterOperator
+// resource for the operator.
+func IngressClusterOperatorName() types.NamespacedName {
+	return types.NamespacedName{
+		Name: "ingress",
```
danehans: Why use a string instead of a constant?

pkg/operator/controller/names.go
```diff
@@ -44,27 +37,33 @@ func RouterDeploymentName(ci *operatorv1.IngressController) types.NamespacedName
 }
 
 // RouterCASecretName returns the namespaced name for the router CA secret.
+// This secret holds the CA certificate that the operator will use to create
+// default certificates for ingresscontrollers.
 func RouterCASecretName(operatorNamespace string) types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: operatorNamespace,
-		Name:      caCertSecretName,
+		Name:      "router-ca",
```
danehans: Same here.

pkg/operator/controller/names.go
```diff
@@ -44,27 +37,33 @@ func RouterDeploymentName(ci *operatorv1.IngressController) types.NamespacedName
...
 }
 
 // RouterCAConfigMapName returns the namespaced name for the router CA configmap.
+// The operator uses this configmap to publish the public key for the CA
+// certificate, so that other operators can include it into their trust bundles.
 func RouterCAConfigMapName() types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: GlobalMachineSpecifiedConfigNamespace,
-		Name:      caCertConfigMapName,
+		Name:      "router-ca",
```
danehans: and here.

pkg/operator/controller/names.go
```diff
@@ -44,27 +37,33 @@ func RouterDeploymentName(ci *operatorv1.IngressController) types.NamespacedName
...
 // RouterCertsGlobalSecretName returns the namespaced name for the router certs
-// secret.
+// secret.  The operator uses this secret to publish the default certificates and
+// their keys, so that the authentication operator can configure the OAuth server
+// to use the same certificates.
 func RouterCertsGlobalSecretName() types.NamespacedName {
 	return types.NamespacedName{
 		Namespace: GlobalMachineSpecifiedConfigNamespace,
-		Name:      routerCertsGlobalSecretName,
+		Name:      "router-certs",
```
danehans: and here.

pkg/operator/controller/names.go
Miciah: Because it seems redundant to define a named constant that is only used to define (what is effectively) another named constant.

ironcladlou:
This looks like a good iteration to me, let's move forward

/lgtm

PR #257, opened 2019-06-25
Refactors main reconcile loop

danehans:
This PR refactors the main reconciliation loop to check for the presence of `ingress.DeletionTimestamp`. It makes the reconciliation loop easier to understand and more efficient by not making unnecessary api calls (i.e. `configv1.DNS`) for an `IngressController` deletion.


danehans:
This PR was created to reduce the size of https://github.com/openshift/cluster-ingress-operator/pull/224

ironcladlou:
lgtm, @Miciah PTAL and feel free to tag when ready

PR #260, opened 2019-06-26
controller: fix AWS internal loadbalancer annotation

ironcladlou:
Use the correct annotation for AWS internal load balancers. Before this
change, the Internal endpoint publishing strategy scope was not correctly
honored on AWS.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -198,9 +198,11 @@ func (m *Provider) getLBHostedZone(name string) (string, error) {
 	var id string
 	fn := func(resp *elb.DescribeLoadBalancersOutput, lastPage bool) (shouldContinue bool) {
 		for _, lb := range resp.LoadBalancerDescriptions {
-			log.V(2).Info("found load balancer", "name", aws.StringValue(lb.LoadBalancerName), "dns name", aws.StringValue(lb.DNSName), "hosted zone ID", aws.StringValue(lb.CanonicalHostedZoneNameID))
-			if aws.StringValue(lb.CanonicalHostedZoneName) == name {
-				id = aws.StringValue(lb.CanonicalHostedZoneNameID)
+			dnsName := aws.StringValue(lb.DNSName)
+			zoneID := aws.StringValue(lb.CanonicalHostedZoneNameID)
+			log.V(2).Info("found load balancer", "name", aws.StringValue(lb.LoadBalancerName), "dns name", dnsName, "hosted zone ID", zoneID)
+			if dnsName == name {
+				id = zoneID
 				return false
 			}
 		}
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -687,26 +687,17 @@ func TestInternalLoadBalancer(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to get infrastructure config: %v", err)
 	}
+	platform := infraConfig.Status.Platform
 
-	annotation := map[string]string{}
-	switch infraConfig.Status.Platform {
-	case configv1.AWSPlatformType:
-		annotation["service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"] = "*"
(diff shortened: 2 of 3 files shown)


pkg/dns/aws/dns.go
```diff
@@ -198,9 +198,11 @@ func (m *Provider) getLBHostedZone(name string) (string, error) {
 	var id string
 	fn := func(resp *elb.DescribeLoadBalancersOutput, lastPage bool) (shouldContinue bool) {
 		for _, lb := range resp.LoadBalancerDescriptions {
-			log.V(2).Info("found load balancer", "name", aws.StringValue(lb.LoadBalancerName), "dns name", aws.StringValue(lb.DNSName), "hosted zone ID", aws.StringValue(lb.CanonicalHostedZoneNameID))
-			if aws.StringValue(lb.CanonicalHostedZoneName) == name {
-				id = aws.StringValue(lb.CanonicalHostedZoneNameID)
+			dnsName := aws.StringValue(lb.DNSName)
+			zoneID := aws.StringValue(lb.CanonicalHostedZoneNameID)
+			log.V(2).Info("found load balancer", "name", aws.StringValue(lb.LoadBalancerName), "dns name", dnsName, "hosted zone ID", zoneID)
+			if dnsName == name {
```
Miciah: What was the motivation for this change?

test/e2e/operator_test.go
```diff
@@ -687,26 +687,17 @@ func TestInternalLoadBalancer(t *testing.T) {
...
-		t.Skip("test skipped on libvirt")
-	case configv1.OpenStackPlatformType:
-		annotation["service.beta.kubernetes.io/openstack-internal-load-balancer"] = "true"
-	case configv1.NonePlatformType:
-		t.Skip("test skipped on \"none\" platform type")
-	case configv1.VSpherePlatformType:
-		t.Skip("test skipped on vSphere")
+	supportedPlatforms := map[configv1.PlatformType]struct{}{
+		configv1.AWSPlatformType:   {},
+		configv1.AzurePlatformType: {},
```
Miciah: Is there a reason not to use `InternalLBAnnotations` to decide whether the platform should be skipped?

pkg/dns/aws/dns.go
Miciah: Oh, because AWS doesn't set `CanonicalHostedZoneName` on internal LBs.  Never mind.

ironcladlou: I put this change in its own commit (5eee64ddcc678b552e8a4e59145ac73ec89e9752) which explains it

test/e2e/operator_test.go
ironcladlou: A related question is how do we consolidate feature enablement checking generally and in a principled way, here and in the origin conformance tests... using the map here might be slightly better in a vacuum here. Followup?

pkg/dns/aws/dns.go
Miciah: Yeah, the cloud provider code uses `lb.DNSName` for the name that ultimately goes into the `Service` object's `status` ([src](https://github.com/kubernetes/cloud-provider-aws/blob/87ce39a8e225fb0a76cb717de63bedc64cc0af89/pkg/cloudprovider/providers/aws/aws.go#L3675)), so we should be using the same to relate the `Service` back to the LB.

test/e2e/operator_test.go
Miciah: If the test is supposed to define the set of platforms where the feature is expected to work, then it should assert that `annotation` is not nil.  But deferring this matter to a follow-up is fine.

pkg/dns/aws/dns.go
ironcladlou: Something I'd like to discuss upstream is, if the DNS name alone is insufficient to identify the LB (i.e. it must be combined with the hosted zone ID), the cloud provider could perhaps surface the zone ID so that we don't need to do any lookups at all (which is pretty brittle/unsafe)

PR #262, opened 2019-06-28
NE-150, NE-191: Implement TLS profile API

Miciah:
#### Bump `openshift/api` for TLS profile API

Bump to openshift/api@3f47c9a8fa7fe36974d225f0a3ec5ab63e8a6e7a to get the TLS profile API and `IngressControlerStatus`'s `ObservedGeneration` field.

* `go.mod:` Bump.
* `go.sum`:
* `vendor/modules.txt:` Regenerate.

#### [NE-150](https://jira.coreos.com/browse/NE-150): Use `observedGeneration` in admission control

Update `status.observedGeneration` when processing an ingresscontroller for admission, and reprocess an ingresscontroller if its generation differs from its `status.observedGeneration` in order to ensure that it is revalidated if it is mutated.

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Call `needsReadmission`, and reprocess the ingresscontroller if needed.
(`admit`): Update the ingresscontroller's `status.observedGeneration` field when updating its "Admitted" status condition.
(`needsReadmission`): New function.  Return a Boolean value indicating whether the given ingresscontroller needs to be reprocessed for admission, based on its current generation and last observed generation.
* `pkg/operator/controller/ingress/status.go` (`ingressStatusesEqual`): Compare observed generation.

#### [NE-191](https://jira.coreos.com/browse/NE-191): Implement TLS profile API

* `manifests/00-cluster-role.yaml:` Allow the operator to get the apiservers config resource.
* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Get the `cluster` apiserver config, which has the default TLS security profile.  Pass the config to `ensureIngressController`.
(`admit`): Add comment explaining that the TLS security profile does not need a default value set here.
(`tlsProfileSpecForIngressController`): New function.  Return a TLS profile spec based on the provided ingresscontroller and apiserver config, using `hasTLSSecurityProfile` and `tlsProfileSpecForSecurityProfile`.
(`hasTLSSecurityProfile`): New Boolean function.  Determine whether the given ingresscontroller specifies a TLS security profile.
(`tlsProfileSpecForSecurityProfile`): New function.  Return a TLS profile spec based on the provided TLS security profile.
(`validate`): Call `validateTLSSecurityProfile`.
(`validTLSVersions`): New variable.  Map of valid TLS versions.
(`isValidCipher`): New Boolean function.  Determine whether the given string looks like a valid cipher name.
(`validateTLSSecurityProfile`): New function.  Validate the given ingresscontroller's TLS security profile, if it specifies one.
(`ensureIngressController`): Add parameter for the apiserver config.  Pass the config to `ensureRouterDeployment`.
* `pkg/operator/controller/ingress/controller_test.go` (`TestTLSProfileSpecForSecurityProfile`, `TestTLSProfileSpecForIngressController`): New tests.
* `pkg/operator/controller/ingress/deployment.go` (`ensureRouterDeployment`): Add parameter for the apiserver config.  Pass the config to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Add parameter for the apiserver config.  Use the config and `tlsProfileSpecForIngressController` determine the TLS profile spec, and set the `ROUTER_CIPHERS` and `SSL_MIN_VERSION` environment variables in the deployment pod template spec.
(`inferTLSProfileSpecFromDeployment`): New function.  Return a TLS profile spec based on the given deployment's pod template spec.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Define a TLS profile in the test input, and check `ROUTER_CIPHERS` and `SSL_MIN_VERSION` in the test output.
(`TestInferTLSProfileSpecFromDeployment`): New test.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Use `computeIngressTLSProfile` to compute the TLS profile spec for the current deployment, and publish the spec in the ingresscontroller's status.
(`computeIngressTLSProfile`): New function.  Use `inferTLSProfileSpecFromDeployment` to compute the current TLS profile from the deployment pod template spec if the deployment is up to date, otherwise keep the profile that is currently in status.
(`ingressStatusesEqual`): Compare TLS profile specs.
* `test/e2e/operator_test.go` (`TestTLSSecurityProfile`): New test.


Miciah:
Rebased.

Miciah:
Rebased onto https://github.com/openshift/api/pull/350/commits/2063ee76814593234a94ef53ad0d932a3aceaa2e, with a hack to fix the an incorrect type for `IngressControllerSpec`'s `SecurityProfile` field (see https://github.com/openshift/api/pull/350#discussion_r305935357).

Miciah:
Provisioning failures.

/retest

Miciah:
Rebased.

Miciah:
Updated to allow mutation of the TLS security profile.

pkg/operator/controller/ingress/controller.go
```diff
@@ -189,13 +197,13 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 // value of a different type if the ingresscontroller could not be processed.
-func (r *reconciler) admit(current *operatorv1.IngressController, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure) error {
+func (r *reconciler) admit(current *operatorv1.IngressController, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, tlsProfileSpec *configv1.TLSProfileSpec) error {
 	updated := current.DeepCopy()
 
 	setDefaultDomain(updated, ingressConfig)
 	setDefaultPublishingStrategy(updated, infraConfig)
 
-	if err := r.validate(updated); err != nil {
+	if err := r.validate(updated, tlsProfileSpec); err != nil {
```
ironcladlou: It's not clear to me why tlsProfile is being treated like a special case compared to domain and publishing strategy throughout. I was expecting the new argument to this function to be `apiConfig` and a new `setDefaultTLSProfile()` function or something. Was there a reason the pattern couldn't be applied to this field?

ironcladlou: (see also: `validate()` which is now unusually special-cased for tlsProfile)

Miciah: > It's not clear to me why tlsProfile is being treated like a special case compared to domain and publishing strategy throughout. I was expecting the new argument to this function to be `apiConfig` and a new `setDefaultTLSProfile()` function or something. Was there a reason the pattern couldn't be applied to this field?

Because we decided that `status.tlsProfile` should not be set till pods are at level.  What would `setDefaultTLSProfile` do if not set `status.tlsProfile`, or what would `validate` use to obtain the TLS profile if not a parameter value?

ironcladlou: >Because we decided that status.tlsProfile should not be set till pods are at level. What would setDefaultTLSProfile do if not set status.tlsProfile, or what would validate use to obtain the TLS profile if not a parameter value?

Is the important difference here between TLS and domain/publishing that TLS is mutable? That is, it's okay to persist status for domain/publishing in advance of rollout because there's simply no opportunity for divergence between spec/status for immutable fields?

Miciah: > Is the important difference here between TLS and domain/publishing that TLS is mutable?

Basically yeah (although someone could argue that _strictly speaking_ we shouldn't set `status.domain` or `status.endpointPublishingStrategy` till pods are up either).

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -0,0 +1,263 @@
...
+		profile      *configv1.TLSSecurityProfile
+		valid        bool
+		expectedSpec *configv1.TLSProfileSpec
+	}{
+
+		{
+			description:  "default (nil)",
+			profile:      nil,
+			valid:        true,
+			expectedSpec: configv1.TLSProfiles["Intermediate"],
```
ironcladlou: How about a [constant](https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L127) here and below?

pkg/operator/controller/ingress/status.go
```diff
@@ -130,10 +180,16 @@ func computeIngressDegradedCondition(deployment *appsv1.Deployment) operatorv1.O
...
 // whether an update is necessary, false otherwise.
 func ingressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
+	if a.ObservedGeneration != b.ObservedGeneration {
+		return false
+	}
 	if !conditionsEqual(a.Conditions, b.Conditions) || a.AvailableReplicas != b.AvailableReplicas ||
 		a.Selector != b.Selector {
 		return false
 	}
+	if !reflect.DeepEqual(a.TLSProfile, b.TLSProfile) {
```
ironcladlou: Isn't there a `semantic.DeepEqual` around that is supposed to be more appropriate for API types for Reasons?

pkg/operator/controller/ingress/status.go
```diff
@@ -77,6 +79,54 @@ func mergeConditions(conditions []operatorv1.OperatorCondition, updates ...opera
...
+	containers := deployment.Spec.Template.Spec.Containers
+	if len(containers) < 1 || containers[0].Name != "router" {
+		return &configv1.TLSProfileSpec{}
+	}
+
+	var (
+		ciphersString       string
+		minTLSVersionString string
+	)
+	for _, envVar := range containers[0].Env {
```
ironcladlou: This seems to be an abstraction leak from the controller code that knows how to create the router deployment from an ingresscontroller. For example, if we ever changed the way that we configure the router implementation's ciphers, does this become a problem? Let's say we change the env var key. This code now has to be updated and backwards compatible. What if we change to an haproxy API call? Would we change this code to query haproxy?

Miciah: > This seems to be an abstraction leak from the controller code that knows how to create the router deployment from an ingresscontroller.

It's all in the same package.

> For example, if we ever changed the way that we configure the router implementation's ciphers, does this become a problem? Let's say we change the env var key. This code now has to be updated and backwards compatible.

Again, this is all in the same package, same controller.  This change does not couple `pkg/operator/controller/ingress/status.go` any more tightly to HAProxy than `pkg/operator/controller/ingress/deployment.go` is already coupled to HAProxy.  Would moving this logic (inferring a TLS profile spec from a deployment pod template spec) to `pkg/operator/controller/ingress/deployment.go` resolve your concern?

> What if we change to an haproxy API call? Would we change this code to query haproxy?

I hadn't considered that the operator itself might make API call to HAProxy, and I can only imagine what form the communication would take.  I speculate we would define a networkpolicy to grant the operator access to the operand pods, and the operator would attempt to update all HAProxy instances via their podips and then retry on failure or update the status synchronously on success.

pkg/operator/controller/ingress/status.go
Miciah: Grepping the Origin and Kubernetes code, `reflect.DeepEqual` appears to be widely used, `semantic.DeepEqual` not so much.  If I recall correctly, `semantic.DeepEqual` might be useful for some API objects for which certain fields need to be ignored or sorted, but I do not know a reason not to use `reflect.DeepEqual` here.  If anything, I would consider go-cmp (which we use elsewhere in this operator).

pkg/operator/controller/ingress/controller_test.go
Miciah: Will fix here (and in other files) in the next push.

pkg/operator/controller/ingress/status.go
ironcladlou: Okay, thanks.

>Would moving this logic (inferring a TLS profile spec from a deployment pod template spec) to pkg/operator/controller/ingress/deployment.go resolve your concern?

This might be good, but it can be a followup refactoring. I wouldn't block the PR on it.

Miciah:
/test e2e-aws-operator
The job failed because provisioning timed out configuring the load balancer.

ironcladlou:
Thank you very much for this. Let's merge and iterate.

/lgtm

PR #268, opened 2019-07-15
main: support cluster config lookup through platform status

ironcladlou:
Support reading platform config through the platform status embedded in
infrastructure status, which is the default in 4.2. Continue reading AWS
status from the deprecated install config in clusters based on 4.1.


knobunc:
Seems reasonable to me, thanks Dan.

ironcladlou:
/hold

Looks like we may have to keep relying on installconfig for 4.1 clusters (even post-upgrade) — holding while I sort that out.

ironcladlou:
I redid this to be compatible with AWS clusters based on 4.1 which will not have region info in the new platform status config.

/hold cancel

PR #269, opened 2019-07-15
WIP: client: honor cluster proxy config

ironcladlou:
If the cluster proxy config is specified, use the proxy config for all non-apiserver HTTP clients.

### DIFF
--- a/cmd/ingress-operator/main.go
+++ b/cmd/ingress-operator/main.go
@@ -3,7 +3,13 @@ package main
 import (
 	"context"
 	"fmt"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
+	"time"
+
+	"golang.org/x/net/http/httpproxy"
 
 	"github.com/ghodss/yaml"
 
@@ -21,6 +27,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -92,6 +99,23 @@ func main() {
 		os.Exit(1)
 	}
 
+	// If the cluster proxy config specifies proxy info, create an HTTP client using the proxy
+	// which can be used to configure the HTTP client for all non-apiserver connections.
+	proxyConfig := &configv1.Proxy{}
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, proxyConfig); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("cluster proxy config was not found")
+		} else {
+			log.Error(err, "failed to get proxy 'cluster'")
+			os.Exit(1)
+		}
+	}
+	var proxyClient *http.Client
+	if len(proxyConfig.Status.HTTPProxy) > 0 || len(proxyConfig.Status.HTTPSProxy) > 0 {
+		log.Info("using HTTP client with cluster proxy config: %#v", proxyConfig)
+		proxyClient = newProxyClient(proxyConfig)
+	}
+
 	operatorConfig := operatorconfig.Config{
 		OperatorReleaseVersion: releaseVersion,
 		Namespace:              operatorNamespace,
@@ -99,7 +123,7 @@ func main() {
 	}
 
 	// Set up the DNS manager.
-	dnsProvider, err := createDNSProvider(kubeClient, operatorConfig, dnsConfig, platformStatus)
+	dnsProvider, err := createDNSProvider(kubeClient, operatorConfig, dnsConfig, platformStatus, proxyClient)
 	if err != nil {
 		log.Error(err, "failed to create DNS manager")
 		os.Exit(1)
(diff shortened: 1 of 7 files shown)


ironcladlou:
Builds on API bump from https://github.com/openshift/cluster-ingress-operator/pull/268

cmd/ingress-operator/main.go
```diff
@@ -86,18 +91,21 @@ func main() {
...
-	clusterConfig := &corev1.ConfigMap{}
-	err = kubeClient.Get(context.TODO(), types.NamespacedName{Namespace: "kube-system", Name: "cluster-config-v1"}, clusterConfig)
-	if err != nil {
-		log.Error(err, "failed to get configmap 'kube-system/cluster-config-v1'")
-		os.Exit(1)
+	// If the cluster proxy config specifies proxy info, create an HTTP client using the proxy
+	// which can be used to configure the HTTP client for all non-apiserver connections.
+	proxyConfig := &configv1.Proxy{}
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, proxyConfig); err != nil {
+		if errors.IsNotFound(err) {
```
danehans: A configv1.Proxy with nil spec/status should exist even if a proxy is not required for 4.2 clusters. With that said, should `os.Exit(1)` be called here?

cmd/ingress-operator/main.go
```diff
@@ -172,24 +183,30 @@ func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, i
...
+	}
+	proxyFn := cfg.ProxyFunc()
+	proxy := func(req *http.Request) (*url.URL, error) {
+		return proxyFn(req.URL)
+	}
+	return &http.Client{
+		Transport: &http.Transport{
+			Proxy: proxy,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
```
danehans: Can you provide context for the `Timeout`, `KeepAlive` and `DualStack` values?

cmd/ingress-operator/main.go
```diff
@@ -172,24 +183,30 @@ func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, i
...
+				KeepAlive: 30 * time.Second,
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
```
danehans: Same with these values?

danehans:
The WIP PR is looking good. I don't see how the operator can react to `configv1.Proxy` changes. CIO should list/watch `configv1.Proxy` and update its deployment env vars based accordingly. Consider using [FromEnvironment](https://godoc.org/golang.org/x/net/http/httpproxy#FromEnvironment) to create the httpproxy.Config from env var's instead of from configv1.Proxy. [Here's](https://github.com/openshift/console-operator/pull/254) an implementation that may help. Let me know your thoughts.

cmd/ingress-operator/main.go
ironcladlou: Will it exist in 4.1 clusters?

cmd/ingress-operator/main.go
ironcladlou: The struct is copied from [DefaultTransport](https://golang.org/src/net/http/transport.go) in the stdlib. Not sure if there's any better way to get the defaults (unlike default http.Client which has a useful zero-value).

cmd/ingress-operator/main.go
ironcladlou: See https://github.com/openshift/cluster-ingress-operator/pull/269#discussion_r303926838

ironcladlou:
@danehans 
>Consider using FromEnvironment to create the httpproxy.Config from env var's instead of from configv1.Proxy. Here's an implementation that may help. Let me know your thoughts.

I'm deliberately avoiding proxy environment variables because doing so means any code using the `net/http` package default clients will auto-configure to use the proxy in their transports. With the introduction of proxy support, I think we need to move towards maintaining complete control/understanding of each HTTP client in use and how they're configured, and not leave anything to implicit configuration.

Although on its face using environment appears to make everything easy (as we don't need to plumb anything into the cloud provider clients), implicit unexpected configuration will bite us. For example, if we set up environment, our current apiserver clients would end up proxied as they're using the default `net/http` client out of the box. So we'd still have to explicitly wire up _those_ clients, which just shifts the problem around.

Another less important reason we should probably start explicitly managing clients is that we likely want different connection configurations for these different endpoints (timeout values, etc.)

So, I think we should go even further and explicitly configure and wire the HTTP clients for apiserver/controller-runtime as a matter of safety, but I'm not tackling that in this PR.

Hope that helps provide some context...

ironcladlou:
I learned from @sttts that we should be able to rely on NO_PROXY to handle excluding apiserver and any other necessary endpoints from proxying. If we can hitch our wagon to a reliable NO_PROXY, then we can indeed simply this all to environment variables. I'll work up an alternative.

Something else I probably missed here is setting env on the operand, which shouldn't really be necessary (as it communicates only with the apiserver) but probably a good idea anyway for consistency.

danehans:
> Something else I probably missed here is setting env on the operand, which shouldn't really be necessary (as it communicates only with the apiserver) but probably a good idea anyway for consistency.

Wiring-up the operator for proxy is definitely optional if the operator does not need to communicate outside the cluster. You may want to consider using a separate PR for that work.

cmd/ingress-operator/main.go
danehans: It will not exist in 4.1 clusters. Will this version of the ingress-operator be run in 4.1 clusters?

cmd/ingress-operator/main.go
danehans: @ironcladlou thanks for the reference. The clients are currently using the default http.Client, correct?

ironcladlou:
In our case, unless the CVO adds the variables to our operator's deployment, we can't just modify the current process environment and expect the http.DefaultClient to work: https://gist.github.com/ironcladlou/dd1e05301d15534aa2887639c05f03a0

We could change the http.DefaultTransport.Proxy very early as a workaround, which might be easier as long as we provide ample docs...

ironcladlou:
https://github.com/openshift/cluster-ingress-operator/issues/270

ironcladlou:
The underlying requirement has been deemed obsolete (https://github.com/openshift/cluster-ingress-operator/issues/270).

/close

PR #184, opened 2019-03-26
Remove unused method in manifests

pravisankar:
- Remove 'k8s.io/apiextensions-apiserver' pkg dependency
- Retry in case of caching issues

PR #272, opened 2019-07-22
Add a pod disruption budget for ingress controllers

Miciah:
For each ingress controller, create a pod disruption budget with a maximum of 25% unavailable pods.

* `manifests/00-cluster-role.yaml`: Allow the operator to manage `PodDisruptionBudget` resources.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Call `ensureRouterPodDisruptionBudget`.
* `pkg/operator/controller/ingress/poddisruptionbudget.go`: New file.
(`ensureRouterPodDisruptionBudget`): New method.  Ensure the appropriate pod disruption budget exists for the given ingress controller.  Get the current one using `currentRouterPodDisruptionBudget`.  If none exists, create one using the `desiredRouterPodDisruptionBudget` function and `createRouterPodDisruptionBudget` method.  If one already exists, update it if necessary using `updateRouterPodDisruptionBudget`, or delete it using `deleteRouterPodDisruptionBudget` if none is desired.
(`createRouterPodDisruptionBudget`): New method.
(`deleteRouterPodDisruptionBudget`): New method.
* `pkg/operator/controller/names.go` (`RouterPodDisruptionBudgetName`): New function.
* `test/e2e/operator_test.go` (`TestPodDisruptionBudgetExists`): New test.  Verify that a `PodDisruptionBudget` resource exists for the default ingresscontroller.

---

This PR is a spin-off from https://github.com/openshift/cluster-ingress-operator/pull/240.

### DIFF
--- a/pkg/operator/controller/ingress/poddisruptionbudget.go
+++ b/pkg/operator/controller/ingress/poddisruptionbudget.go
@@ -1,0 +1,142 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// ensureRouterPodDisruptionBudget ensures the pod disruption budget exists for
+// a given ingresscontroller.  Returns a Boolean indicating whether the PDB
+// exists, the PDB if it does exist, and an error value.
+func (r *reconciler) ensureRouterPodDisruptionBudget(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) (bool, *policyv1beta1.PodDisruptionBudget, error) {
+	wantPDB, desired, err := desiredRouterPodDisruptionBudget(ic, deploymentRef)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to build pod disruption budget: %v", err)
+	}
+
+	havePDB, current, err := r.currentRouterPodDisruptionBudget(ic)
+	if err != nil {
+		return false, nil, err
+	}
+
+	switch {
+	case !wantPDB && !havePDB:
+		return false, nil, nil
+	case !wantPDB && havePDB:
+		if deleted, err := r.deleteRouterPodDisruptionBudget(current); err != nil {
+			return true, current, fmt.Errorf("failed to delete pod disruption budget: %v", err)
+		} else if deleted {
+			log.Info("deleted pod disruption budget", "poddisruptionbudget", current)
+		}
+	case wantPDB && !havePDB:
+		if created, err := r.createRouterPodDisruptionBudget(desired); err != nil {
+			return false, nil, fmt.Errorf("failed to create pod disruption budget: %v", err)
+		} else if created {
+			log.Info("created pod disruption budget", "poddisruptionbudget", desired)
+		}
+	case wantPDB && havePDB:
+		if updated, err := r.updateRouterPodDisruptionBudget(current, desired); err != nil {
+			return true, nil, fmt.Errorf("failed to update pod disruption budget: %v", err)
+		} else if updated {
+			log.Info("updated pod disruption budget", "poddisruptionbudget", desired)
+		}
+	}
+
+	return r.currentRouterPodDisruptionBudget(ic)
+}
+
+// desiredRouterPodDisruptionBudget returns the desired router pod disruption
+// budget.  Returns a Boolean indicating whether a PDB is desired, as well as
+// the PDB if one is desired.
+func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) (bool, *policyv1beta1.PodDisruptionBudget, error) {
+	name := controller.RouterPodDisruptionBudgetName(ic)
+	pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
+	pdb := policyv1beta1.PodDisruptionBudget{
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/names.go
```diff
@@ -76,6 +76,15 @@ func RouterOperatorGeneratedDefaultCertificateSecretName(ci *operatorv1.IngressC
 	}
 }
 
+// RouterPodDisruptionBudgetName returns the namespaced name for the router
+// deployment's pod disruption budget.
+func RouterPodDisruptionBudgetName(ic *operatorv1.IngressController) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: "openshift-ingress",
+		Name:      "router-" + ic.Name,
```
ironcladlou: using the word "router" in any of our resources was a mistake from the start — could we avoid propagating it further?

ironcladlou:
Any value in trying to devise an e2e test for this?

pkg/operator/controller/names.go
Miciah: We could, but this would be the odd resource out till we renamed other resources, and I am not aware of plans to rename other resources.  What name do you propose? Just use the ingresscontroller's name, or prefix "ingress-" or "ingresscontroller-" or similar to the name?

Miciah:
> Any value in trying to devise an e2e test for this?

Probably.  Should it be anything more sophisticated than checking that a pod disruption budget exists for the default ingress controller?

pkg/operator/controller/names.go
ironcladlou: I like "ingress-", but if you want to just stay consistent with router I'm fine with that too at this point

ironcladlou:
>Probably. Should it be anything more sophisticated than checking that a pod disruption budget exists for the default ingress controller?

Would that be much more useful than a unit test?

At the end of the day I think we can trust the disruption budget feature in k8s itself works, so if we know we're setting the right fields on the deployment, maybe that's enough?

Miciah:
> Would that be much more useful than a unit test?
> 
> At the end of the day I think we can trust the disruption budget feature in k8s itself works, so if we know we're setting the right fields on the deployment, maybe that's enough?

`PodDisuptionBudget` is a discrete API resource, so it makes sense to make sure it is created, but it makes sense to defer to k8s tests to verify that the general disruption budget feature works properly, I agree.

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -0,0 +1,145 @@
...
+		return false, nil, err
+	}
+	return true, pdb, nil
+}
+
+// createRouterPodDisruptionBudget creates a pod disruption budget.  Returns a
+// Boolean indicating whether the PDB was created, and an error value.
+func (r *reconciler) createRouterPodDisruptionBudget(pdb *policyv1beta1.PodDisruptionBudget) (bool, error) {
+	if err := r.client.Create(context.TODO(), pdb); err != nil {
+		return false, err
```
ironcladlou: What if the error is because the resource already exists?

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -0,0 +1,145 @@
...
+// deleteRouterPodDisruptionBudget deletes a pod disruption budget.  Returns a
+// Boolean indicating whether the PDB was deleted, and an error value.
+func (r *reconciler) deleteRouterPodDisruptionBudget(pdb *policyv1beta1.PodDisruptionBudget) (bool, error) {
+	if err := r.client.Delete(context.TODO(), pdb); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	log.Info("created pod disruption budget", "namespace", pdb.Namespace, "name", pdb.Name)
```
ironcladlou: Says created when it means deleted, and the logging is already happening on the calling side anyway

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -0,0 +1,145 @@
...
+func (r *reconciler) updateRouterPodDisruptionBudget(current, desired *policyv1beta1.PodDisruptionBudget) (bool, error) {
+	changed, updated := podDisruptionBudgetChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return false, err
+	}
+	log.Info("updated pod disruption budget", "namespace", updated.Namespace, "name", updated.Name)
```
ironcladlou: Logging is happening on the calling side already

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -0,0 +1,145 @@
...
+	return true, pdb, nil
+}
+
+// createRouterPodDisruptionBudget creates a pod disruption budget.  Returns a
+// Boolean indicating whether the PDB was created, and an error value.
+func (r *reconciler) createRouterPodDisruptionBudget(pdb *policyv1beta1.PodDisruptionBudget) (bool, error) {
+	if err := r.client.Create(context.TODO(), pdb); err != nil {
+		return false, err
+	}
+	log.Info("created pod disruption budget", "namespace", pdb.Namespace, "name", pdb.Name)
```
ironcladlou: Logging is happening on the calling side already

pkg/operator/controller/ingress/poddisruptionbudget.go
Miciah: `ensureRouterPodDisruptionBudget` should call `createRouterPodDisruptionBudget` only if the resource did not already exist.  Morever, returning a non-nil error value should just cause a retry of the reconciliation, which should be reasonably innocuous.  We've been removing similar `IsAlreadyExists` checks in PRs such as #151 and #184.

pkg/operator/controller/ingress/poddisruptionbudget.go
Miciah: Deleted.

pkg/operator/controller/ingress/poddisruptionbudget.go
Miciah: Deleted.

pkg/operator/controller/ingress/poddisruptionbudget.go
Miciah: Deleted.  Thanks!

PR #274, opened 2019-07-26
Add GCP DNS support

ironcladlou:
Introduce GCP DNS support. As with the other implementations, updates are not yet supported.


cmd/ingress-operator/main.go
```diff
@@ -99,7 +100,7 @@ func main() {
 	}
 
 	// Set up the DNS manager.
-	dnsProvider, err := createDNSProvider(kubeClient, operatorConfig, dnsConfig, platformStatus)
+	dnsProvider, err := createDNSProvider(kubeClient, operatorConfig, dnsConfig, platformStatus, releaseVersion)
```
Miciah: Why add a new parameter instead of using `operatorConfig.OperatorReleaseVersion` as the logic for the other handlers does?

cmd/ingress-operator/main.go
```diff
@@ -119,8 +120,10 @@ func main() {
 
 // createDNSManager creates a DNS manager compatible with the given cluster
 // configuration.
-func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (dns.Provider, error) {
+func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus, version string) (dns.Provider, error) {
 	var dnsProvider dns.Provider
+	userAgent := fmt.Sprintf("OpenShift/%s (ingress-operator)", version)
```
Miciah: Can we use this for all handlers for better consistency and code re-use?

pkg/dns/gcp/gcp.go
```diff
@@ -0,0 +1,71 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	gdnsv1 "google.golang.org/api/dns/v1"
+	"google.golang.org/api/option"
+)
+
+var (
+	_   dns.Provider = &Provider{}
+	log              = logf.Logger.WithName("dns")
```
Miciah: Unused.  Presumably we should have some logging though.

pkg/dns/gcp/gcp.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (p *Provider) Ensure(record *iov1.DNSRecord, zone configv1.DNSZone) error {
+	rrs := &gdnsv1.ResourceRecordSet{
+		Kind:    "dns#resourceRecordSet",
+		Name:    record.Spec.DNSName,
+		Rrdatas: record.Spec.Targets,
+		Type:    record.Spec.RecordType,
+	}
+	change := &gdnsv1.Change{Additions: []*gdnsv1.ResourceRecordSet{rrs}}
+	call := p.dnsService.Changes.Create(p.config.Project, zone.ID, change)
```
Miciah: Are additions idempotent?  It wasn't obvious to me on a quick scan through the API documentation.

pkg/dns/gcp/gcp.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (p *Provider) Delete(record *iov1.DNSRecord, zone configv1.DNSZone) error {
+	rrs := &gdnsv1.ResourceRecordSet{
+		Kind:    "dns#resourceRecordSet",
+		Name:    record.Spec.DNSName,
+		Rrdatas: record.Spec.Targets,
+		Type:    record.Spec.RecordType,
+	}
+	change := &gdnsv1.Change{Deletions: []*gdnsv1.ResourceRecordSet{rrs}}
+	call := p.dnsService.Changes.Create(p.config.Project, zone.ID, change)
```
Miciah: Are deletions idempotent, or will deleting a non-existent record return an error?

cmd/ingress-operator/main.go
ironcladlou: Fixed

cmd/ingress-operator/main.go
ironcladlou: Probably, in a followup though

pkg/dns/gcp/gcp.go
ironcladlou: The ensure method is now idempotent

pkg/dns/gcp/gcp.go
ironcladlou: Deletion is also idempotent

ironcladlou:
>We have pkg/dns/aws/dns.go, pkg/dns/azure/dns.go, and now pkg/dns/gcp/gcp.go; should that be pkg/dns/gcp/dns.go?

I say they're all wrong, and renamed the new GCP file `provider.go` 😁

pkg/dns/gcp/gcp.go
ironcladlou: Removed

ironcladlou:
Just finished integration with credentials requests — this is now ready for review.

manifests/00-ingress-credentials-request.yaml
```diff
@@ -37,4 +37,21 @@ spec:
...
+  namespace: openshift-cloud-credential-operator
+spec:
+  secretRef:
+    name: cloud-credentials
+    namespace: openshift-ingress-operator
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/dns.admin
```
ironcladlou: This is subject to change as we move forward — it was the narrowest usable prefab role I could find.

PR #275, opened 2019-07-29
Bug 1734192: getPlatformStatus: Check for nil platformStatus

Miciah:
* `cmd/ingress-operator/main.go` (`getPlatformStatus`): Check whether the cluster infrastructure config's `status.platformStatus` field is nil.


ironcladlou:
Thanks for cleaning up my mess 🙇‍♂️

/lgtm

PR #280, opened 2019-08-05
Bug 1709958: Avoid dropping traffic during upgrade

Miciah:
#### `TestDeploymentConfigChanged`: tolerations ordering

* `pkg/operator/controller/ingress/deployment_test.go` (`TestDeploymentConfigChanged`): Add test case that verifies that the ordering of tolerations is ignored.

#### Avoid dropping traffic during upgrade

Omit the affinity policy when using host network.  The scheduler already prevents colocation of pods that use host networking and request the same ports, so the affinity policy is superfluous in this case.

Change the affinity policy and deployment strategy when using the load-balancer or private endpoint publishing strategy as follows.  First, configure affinity for replicas of different generations of same controller.  Second, configure anti-affinity for replicas of same generation of same controller.  Finally, configure the deployment strategy to surge.

The intention of the changes for the load-balancer and private endpoint publishing strategies is to avoid dropping traffic during rolling updates of an ingress controller's deployment.  If a node loses local endpoints for the deployment, then the service proxy will drop traffic to that node for that ingress controller, and it may take some time for the load balancer to stop sending traffic to that node.  These changes, combined with a change to the ReplicaSet controller, will ensure that a node that has local endpoints for an ingress controller at the start of a rollout will continue to have local endpoints during and at completion of the rollout, thus preventing traffic from being dropped.

* `pkg/operator/controller/controller_router_deployment.go` (`desiredRouterDeployment`): Set the deployment's pod selector to a copy of the pod template's labels, so that subsequently mutating labels does not mutate the selector.  If the ingress controller uses the host network, do not set any affinity policy.  If the ingress controller uses the load-balancer or private endpoint publishing strategy, add a "hash" label to identify the deployment's generation, configure affinity for replicas of different generations of the same ingress controller, configure anti-affinity for the same generation of the same ingress controller, and configure the deployment strategy to surge.
(`deploymentHash`): New function.  Return a stringified hash value for the given deployment, using only the fields that, if changed, should trigger an update.
(`hashableDeployment`): New function.  Return a copy of the given deployment with fields that should be used for computing its hash copied over, fields that are slices sorted, and fields that should be ignored zeroed.
(`deepHashObject`): New function.  Hash the given object.
(`deploymentConfigChanged`): Compare hashes of the current and expected deployments instead of comparing fields directly.  Set labels during an update.
(`cmpEnvs`, `cmpVolumes`, `cmpSecretVolumeSource`, `cmpTolerations`): Deleted.
* `pkg/operator/controller/controller_router_deployment_test.go` (`TestDesiredRouterDeployment`): Verify that the hash label is set when it should be and that it has the right hash value.
(`TestDeploymentConfigChanged`): Add test cases to verify that the hash is ignored in the deployment labels and affinity.  Add a test case to verify that deleting the affinity policy is not ignored.  Add test cases to verify that ordering of the label selector expressions in affinity terms is ignored.
* `pkg/operator/controller/names.go` (`ControllerDeploymentHashLabel`): New constant.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -3,11 +3,13 @@ package ingress
 import (
 	"context"
 	"fmt"
+	"hash"
+	"hash/fnv"
 	"path/filepath"
+	"sort"
 	"strings"
 
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/davecgh/go-spew/spew"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
@@ -21,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	configv1 "github.com/openshift/api/config/v1"
 )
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/names.go
```diff
@@ -18,6 +18,19 @@ const (
...
+	// prevent colocation of ingress controller pod replicas that use the
+	// host network.
+	ControllerDeploymentUsesHostNetworkLabel = "ingresscontroller.operator.openshift.io/uses-hostnetwork"
+
+	// ControllerDeploymentNonceLabel identifies an ingress controller deployment's
+	// generation.  This label is used for affinity, to colocate replicas of
+	// different generations of the same ingress controller, and for
+	// anti-affinity, to prevent colocation of replicas of the same
+	// generation of the same ingress controller.
+	ControllerDeploymentNonceLabel = "ingresscontroller.operator.openshift.io/nonce"
```
ironcladlou: Instead of exposing the host network and generation details through labels, would it be possible for the operator to compute whether a given ingresscontroller is co-locatable and then generically label with something like `ingresscontroller.operator.openshift.io/colocation-disabled`?

Miciah:
I have not been able to diagnose or reproduce the test failures.
/retest

pkg/operator/controller/ingress/deployment.go
```diff
@@ -75,40 +78,102 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
-	// new nodes to support the rollout. This means a positive surge can cause the
-	// rollout to wedge in the absence of auto-scaling.
-	pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
-	deployment.Spec.Strategy = appsv1.DeploymentStrategy{
-		Type: appsv1.RollingUpdateDeploymentStrategyType,
-		RollingUpdate: &appsv1.RollingUpdateDeployment{
-			MaxUnavailable: pointerTo(intstr.FromString("25%")),
-			MaxSurge:       pointerTo(intstr.FromInt(0)),
-		},
+		// Avoid going from one replica to zero replicas on any given node.
```
ironcladlou: Would you mind also talking about the details of deployments that explain how these values accomplish the stated behavior?

Miciah:
Rebased.

Miciah:
Provisioning failure.

Miciah:
Rebased.

Miciah:
Fixed out-of-order imports and updated comments for the deployment strategy and affinity policy.

pkg/operator/controller/ingress/deployment.go
Miciah: I have updated the comments better to tell the tale of Ingress Operator and the Zero-Downtime Rolling Update.

Miciah:
/test e2e-aws-operator
Failed to bootstrap.
```
level=error msg="Attempted to gather ClusterOperator status after installation failure: listing ClusterOperator objects: the server could not find the requested resource (get clusteroperators.config.openshift.io)"
```
Must-gather got a bunch of empty files, so I'm not sure how to diagnose this one.

Miciah:
> Must-gather got a bunch of empty files, so I'm not sure how to diagnose this one.

The artifacts do have bootkube logs:
```

ironcladlou:
This is a big deal, thank you

PR #282, opened 2019-08-06
status: compute ingresscontroller availability from deployment conditions

ironcladlou:
Before this change, ingresscontroller availability was determined based on
deployment available replicas from status. Now, consider the ingresscontroller
available if the deployment has an `Available=True` condition. Available
replicas is a weak proxy for deployment availability. The condition takes into
account the complex availability rules of a deployment and also provides
additional context as to why the condition is in a certain state.

### DIFF
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -182,16 +182,40 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
 	}
 }
 
-func TestComputeIngressStatusConditions(t *testing.T) {
+func TestComputeIngressAvailableCondition(t *testing.T) {
 	testCases := []struct {
-		description     string
-		availRepl, repl int32
-		condType        string
-		condStatus      operatorv1.ConditionStatus
+		description string
+		deployment  []appsv1.DeploymentCondition
+		expect      operatorv1.OperatorCondition
 	}{
-		{"0/2 deployment replicas available", 0, 2, operatorv1.OperatorStatusTypeAvailable, operatorv1.ConditionFalse},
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/status_test.go
```diff
@@ -182,16 +182,40 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
 
-func TestComputeIngressStatusConditions(t *testing.T) {
+func TestComputeIngressAvailableCondition(t *testing.T) {
 	testCases := []struct {
-		description     string
-		availRepl, repl int32
-		condType        string
-		condStatus      operatorv1.ConditionStatus
+		description string
+		deployment  []appsv1.DeploymentCondition
```
Miciah: Maybe `deployment` → `deploymentConditions` or `inputConditions`.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -182,16 +182,40 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+		expect      operatorv1.OperatorCondition
```
Miciah: `expect` could be a Boolean value, with the `operatorv1.OperatorCondition` value still computed in the `for` loop.  Are you anticipating adding test cases that verify that the reason or message is accurately reflected?

pkg/operator/controller/ingress/status_test.go
ironcladlou: Done

pkg/operator/controller/ingress/status_test.go
ironcladlou: Maybe Reason. In any case, I just thought it was more clear to use the type directly in this case.

PR #283, opened 2019-08-07
Bug 1698562: status: introduce ingresscontroller degraded condition

ironcladlou:
Introduce degraded condition computation for ingresscontroller. For now,
degraded only considers failed deployments, which seems like a conservative bare
minimum indicator of a degraded state. Ensure that ingresscontroller deployments
have a useful progressDeadlineSeconds so that degraded deployments are actually
detected in a useful timeframe.

Refacor clusteroperator degraded status to account for ingresscontroller
degraded conditions.


assets/router/deployment.yaml
```diff
@@ -4,6 +4,7 @@ kind: Deployment
 apiVersion: apps/v1
 # name and namespace are set at runtime.
 spec:
+  progressDeadlineSeconds: 120
```
ironcladlou: This might be too long, but 120 seems conservative and we can adjust if necessary.

ironcladlou:
Here's an example of the degraded condition on the ingresscontroller:

```
  - lastTransitionTime: "2019-08-07T15:05:03Z"
    message: 'The deployment failed (reason: ProgressDeadlineExceeded) with message:
      ReplicaSet "router-default-858c66c5c9" has timed out progressing.'
    reason: DeploymentFailed
    status: "True"
    type: Degraded
```

And here's the corresponding degraded condition on the clusteroperator:

```
  - lastTransitionTime: "2019-08-07T15:05:03Z"
    message: 'Some ingresscontrollers are degraded: default'
    reason: IngressControllersDegraded
    status: "True"
    type: Degraded
```

The ingresscontroller and operator are still available in this case because minimum deployment availability is maintained.

pkg/operator/controller/status/controller_test.go
```diff
@@ -136,46 +121,20 @@ func TestComputeOperatorStatusConditions(t *testing.T) {
...
+			Status: tc.expectProgressing,
 		}
 
-		conditions := r.computeOperatorStatusConditions([]configv1.ClusterOperatorStatusCondition{},
-			namespace, tc.allIngressesAvailable, oldVersions, reportedVersions)
+		actual := computeOperatorProgressingCondition(tc.allIngressesAvailable, oldVersions, reportedVersions, tc.curVersions.operator, tc.curVersions.operand)
 		conditionsCmpOpts := []cmp.Option{
 			cmpopts.IgnoreFields(configv1.ClusterOperatorStatusCondition{}, "LastTransitionTime", "Reason", "Message"),
 			cmpopts.EquateEmpty(),
 			cmpopts.SortSlices(func(a, b configv1.ClusterOperatorStatusCondition) bool { return a.Type < b.Type }),
```
Miciah: No longer need `cmpopts.SortSlices` or `cmpopts.EquateEmpty`.

assets/router/deployment.yaml
Miciah: This reminds me that we need to set the readiness probe to use `/healthz/ready`.  I'm nervous about setting the progressing deadline significantly lower than the default, especially if we start using `/healthz/ready`.  Are we confident that the initial sync will finish in time on large clusters?

pkg/operator/controller/status/controller_test.go
```diff
@@ -8,14 +8,10 @@ import (
...
 
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TestComputeOperatorStatusConditions(t *testing.T) {
-	type conditions struct {
-		degraded, progressing, available bool
-	}
+func TestComputeOperatorProgressingCondition(t *testing.T) {
```
Miciah: So we're losing unit-test coverage of `computeOperatorAvailableCondition` and `computeOperatorDegradedCondition`, but I suppose they are sufficiently covered by E2E tests.

pkg/operator/controller/ingress/status.go
```diff
@@ -108,6 +109,24 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "DeploymentFailed",
+				Message: fmt.Sprintf("The deployment failed (reason: %s) with message: %s", cond.Reason, cond.Message),
+			}
+		}
+	}
+	return operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeDegraded,
+		Status: operatorv1.ConditionFalse,
+		Reason: "DeploymentAvailable",
```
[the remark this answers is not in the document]
ironcladlou: Good point, I removed the reason entirely

pkg/operator/controller/status/controller_test.go
ironcladlou: Fixed

assets/router/deployment.yaml
ironcladlou: The default is 600 seconds, which seems too long. Do you agree? Would some e2e test warn us if we chose a value that's too short on average?

pkg/operator/controller/status/controller_test.go
ironcladlou: They currently seem simple enough that unit test coverage would be more code than it's worth given e2e, IMO.

If we start introducing additional inputs to the formulas, unit tests may become useful...

assets/router/deployment.yaml
Miciah: > The default is 600 seconds, which seems too long. Do you agree?

That's what I'm wondering.

> Would some e2e test warn us if we chose a value that's too short on average?

Not if the E2E tests are not representative of production clusters.  Moreover, while we may be fine now, I intend to fix the readiness check to use `/healthz/ready`, which will cause the deployment not to be ready until the router has synched routes, which I could see taking more than 120 seconds on burdened clusters with many routes.

smarterclayton: Progress deadline is usually measured in 10-20m.

smarterclayton: This is way too low.

ironcladlou: We're also not setting readiness endpoints correctly, fixing that and changing back to 10m in a followup.

ironcladlou:
Tests are still going, might as well roll the followups into this one.

/hold

assets/router/deployment.yaml
ironcladlou: Nevermind, fixed here instead

PR #284, opened 2019-08-08
Fix admission to log rejections and not retry

Miciah:
#### Fix admission to log rejections and not retry

When processing an ingresscontroller for admission, distinguish between error values that indicate that admission processing could not be completed and error values that indicate that the processing did complete and the ingresscontroller was rejected.  Do not update the "Admitted" condition in the first case.  Log the rejection and do not retry reconciliation in the latter case.

* `pkg/operator/controller/ingress/controller.go` (`admissionRejected`): New type for admission rejections.
(`Error`): New method for the new `admissionRejected` type.
(`Reconcile`): When admission rejects the ingresscontroller, log the rejection and do not retry.
(`admit`): Only set the "Admitted" condition if validation rejected the ingresscontroller.  Return the error value from `validate` if it is not nil.
(`validate`): Return an `admissionRejected` value if the ingresscontroller is determined to be invalid.

#### Replace admission log messages with events

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Instead of logging admission or rejection, emit events.

----

I noticed in CI logs that the `TestUniqueDomainRejection` test causes the operator to log "admitted ingresscontroller" repeatedly even though the ingresscontroller is being rejected.  It is obvious to me that the log message needs to be changed, but I am not so sure whether or not we should retry admission (in case the conflicting ingresscontroller is deleted in between retries).  @ironcladlou, what do you think?

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -166,13 +166,13 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		if err := r.admit(ingress, ingressConfig, infraConfig); err != nil {
 			switch err := err.(type) {
 			case *admissionRejection:
-				log.Info("rejected ingresscontroller", "ingresscontroller", ingress, "reason", err.Reason)
+				r.recorder.Event(ingress, "Warning", "Rejected", err.Reason)
 				return reconcile.Result{}, nil
 			default:
 				return reconcile.Result{}, fmt.Errorf("failed to admit ingresscontroller: %v", err)
 			}
 		}
-		log.Info("admitted ingresscontroller", "ingresscontroller", ingress)
+		r.recorder.Event(ingress, "Normal", "Admitted", "ingresscontroller passed validation")
 		// Just re-queue for simplicity
 		return reconcile.Result{Requeue: true}, nil
 	}
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -106,6 +106,18 @@ type reconciler struct {
 	recorder record.EventRecorder
 }
 
+// admissionRejection is an error type for ingresscontroller admission
+// rejections.
+type admissionRejection struct {
+	// Reason describes why the ingresscontroller was rejected.
+	Reason string
+}
+
+// Error returns the reason or reasons why an ingresscontroller was rejected.
+func (e *admissionRejection) Error() string {
+	return e.Reason
+}
+
 // Reconcile expects request to refer to a ingresscontroller in the operator
 // namespace, and will do all the work to ensure the ingresscontroller is in the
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -106,6 +106,32 @@ type reconciler struct {
...
+	return e.rejectionReason
+}
+
+// newAdmissionRejectedError returns a new admissionRejectedError with the given
+// rejection reason.
+func newAdmissionRejectedError(reason string) *admissionRejectedError {
+	return &admissionRejectedError{
+		rejectionReason: reason,
+	}
+}
```
ironcladlou: Simpler ([more idiomatic](https://golang.org/doc/effective_go.html#errors)?) to ditch the constructor:

```go
type admissionRejectedError struct {
  Reason string
}

func (e * admissionRejectedError) Error() string {
  return e.Reason
}
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -152,6 +178,10 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// successful, immediately re-queue to refresh state.
 	if !isAdmitted(ingress) {
 		if err := r.admit(ingress, ingressConfig, infraConfig); err != nil {
+			if isAdmissionRejected(err) {
```
ironcladlou: Could instead use a [type switch](https://golang.org/doc/effective_go.html#type_switch)

Miciah: Done.

pkg/operator/controller/ingress/controller.go
Miciah: Done.

Miciah: Should `Reason` be public?

pkg/operator/controller/ingress/controller.go
```diff
@@ -152,7 +164,13 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// successful, immediately re-queue to refresh state.
 	if !isAdmitted(ingress) {
 		if err := r.admit(ingress, ingressConfig, infraConfig); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to admit ingresscontroller: %v", err)
+			switch err := err.(type) {
+			case *admissionRejection:
+				log.Info("rejected ingresscontroller", "ingresscontroller", ingress, "reason", err.Error())
```
ironcladlou: In this context `err` is of type `admissionRejection` and so you can reference `err.Reason`

Miciah: Done.

ironcladlou:
/lgtm

Figuring out better retries can be a followup

PR #286, opened 2019-08-13
dns: always publish records to configured DNS zones

ironcladlou:
Before this change, DNS records were only published to the configured public DNS
if the ingresscontroller load balancer is public. After this change, always
publish DNS records to any configured private or public DNS zone regardless of
the ingresscontroller load balancer scope. The new default seems more generally
useful, especially in the contexts in which internal load balancers are used.
Non-cluster clients on the same VPC don't always resolve DNS through the
internal zone, and so need to resolve the load balancer's internal hostname
through a public zone.

This change also seems more consistent with our published DNS configuration API
which doesn't mention the current behavior with regards to DNS and
internal/external. In fact, it's clear in the DNS configuration docs that the
presence of a public or private zone is the way to configure whether records
should be published in a given zone.

One notable implication of this change is that the public zone DNS record
pointing at the internal load balancer leaks a detail about the internal load
balancer (its hostname or IP).

/cc @smarterclayton @openshift/sig-network-edge @patrickstrick @mwoodson

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -82,13 +82,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, nil
 	}
 
+	// Only process records associated with ingresscontrollers, because this isn't
+	// intended to be a general purpose DNS management system.
 	ingressName, ok := record.Labels[manifests.OwningIngressControllerLabel]
 	if !ok {
 		log.V(2).Info("warning: dnsrecord is missing owner label", "dnsrecord", record, "ingresscontroller", ingressName)
 		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
 	}
-	ingress := &operatorv1.IngressController{}
-	if err := r.cache.Get(context.TODO(), types.NamespacedName{Namespace: record.Namespace, Name: ingressName}, ingress); err != nil {
+	if err := r.cache.Get(context.TODO(), types.NamespacedName{Namespace: record.Namespace, Name: ingressName}, &operatorv1.IngressController{}); err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


Miciah:
You're also deleting the check that the DNSRecord resource has an owning IngressController resource.  We had that orphan check before we added the external-LB check.  Is the orphan check not desirable?

ironcladlou:
>You're also deleting the check that the DNSRecord resource has an owning IngressController resource. We had that orphan check before we added the external-LB check. Is the orphan check not desirable?

I'm no longer sure what value it really adds. Can you think of anything important it would actually guard against?

Miciah:
> Can you think of anything important it would actually guard against?

Administrator-created DNSRecord resources in any of the namespaces that the operator watches—custom DNSRecord resources could be nifty, but it's something we do not intend to expose right now.

abhinavdahiya:
> Non-cluster clients on the same VPC don't always resolve DNS through the
> internal zone, and so need to resolve the load balancer's internal hostname
> through a public zone.

hmm... I don't think this is correct. 
A private zone is created with SOA on the cluster_domain. and therefore when attached to the VPC... all DNS requests from inside the VPC treat that Private Zone as authority to provide resolution for all subdomains on the cluster_domain...

patrickstrick:
The client in question is not an AWS device, but rather the client would be, for example, in the customer's on-premises datacenter connected to the VPC by VPN.

pkg/operator/controller/dns/controller.go
```diff
@@ -82,13 +82,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, nil
 	}
 
+	// Only process records associated with ingresscontrollers, because this isn't
+	// intended to be a general purpose DNS management system.
```
ironcladlou: @Miciah I restored the check and added a comment, PTAL.

PR #287, opened 2019-08-13
Bug 1740441: manifests: make credentialsrequest names consistent

ironcladlou:
Use consistent naming for credentialsrequests.

### DIFF
--- a/manifests/00-ingress-credentials-request.yaml
+++ b/manifests/00-ingress-credentials-request.yaml
@@ -26,7 +26,7 @@ kind: CredentialsRequest
 metadata:
   labels:
     controller-tools.k8s.io: "1.0"
-  name: azure-openshift-ingress
+  name: openshift-ingress-azure
   namespace: openshift-cloud-credential-operator
 spec:
   secretRef:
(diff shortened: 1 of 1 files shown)


ironcladlou:
@csrwng @dgoodwin any upgrade implications here? Looks like the implication could be that in upgraded clusters we would have new and old request resources co-existing and targeting the same secret. Would the new resources simply remain unfulfilled as they conflict? If so that would still be fine from our operators' perspectives as we only consume the resulting secret, not the CSR itself.

dgoodwin:
If the CVO doesn't reconcile deletions and remove the old cred requests, and both versions target the same destination secret, you're probably in line for trouble here. It would depend on the cloud actuator but for AWS, I believe both will get minted, and written to the same target secret, and you could easily be looking at a hotloop. Would need to be tested, should be pretty simple to do in any cluster just create a dupe request pointing to same target and see what happens.

ironcladlou:
@dgoodwin Given that, and since our operator doesn't actually manage these resources, is there any path forward to renaming them? cc @abhinavdahiya

dgoodwin:
You could write to a new target secret and have your operator use that, as well as cleanup the old cred requests.

ironcladlou:
I'm going to leave AWS as-is because it at least has the right prefixing and doesn't seem worth migrating. Renaming Azure and GCP now should be okay since we haven't released 4.2 yet. Agree?

/hold cancel

dgoodwin:
+1

ironcladlou:
@Miciah PTAL

PR #289, opened 2019-08-14
Bug 1740595: ensureServiceMonitor: Update resource if needed

Miciah:
If a servicemonitor already exists but is not the same as the desired servicemonitor, update the current servicemonitor.

https://github.com/openshift/cluster-ingress-operator/pull/242 added a selector to the servicemonitor that `desiredServiceMonitor` returns, so the operator must update any servicemonitors created by older versions of the operator.

* `manifests/00-cluster-role.yaml`: Allow updating servicemonitors.
* `pkg/operator/controller/ingress/monitoring.go` (`ensureServiceMonitor`): Add godoc.  If a servicemonitor is needed and none exists, use `createServiceMonitor` to create it.  If a servicemonitor already exists, update it if needed using `updateServiceMonitor`.  Finally, return a Boolean value indicating whether a servicemonitor exists.
(`desiredServiceMonitor`): Add godoc.  Fix the type of the "endpoints" field, and add a comment why it is important to use the right type.
(`currentServiceMonitor`): Add godoc.  Add a Boolean return value indicating whether a servicemonitor exists.
(`createServiceMonitor`, `updateServiceMonitor`): New methods, used by `ensureServiceMonitor`.
(`serviceMonitorChanged`): New function, used by `updateServiceMonitor`. Compare the specs of the given servicemonitors and return a Boolean value indicating whether an update is needed and, if so, an updated servicemonitor definition.
* `pkg/operator/controller/ingress/metrics.go` (`ensureMetricsIntegration`): Ignore the Boolean return value from `ensureServiceMonitor`.
* `pkg/operator/controller/ingress/monitoring_test.go` (`TestServiceMonitorChanged`): New test.

### DIFF
--- a/pkg/operator/controller/ingress/monitoring.go
+++ b/pkg/operator/controller/ingress/monitoring.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/ingress/monitoring.go
```diff
@@ -75,10 +100,13 @@ func desiredServiceMonitor(ic *operatorv1.IngressController, svc *corev1.Service
 		Version: "v1",
 	})
 	sm.SetOwnerReferences([]metav1.OwnerReference{deploymentRef})
-	return sm
+	return true, sm
```
ironcladlou: If we always want one, why return a boolean at all?

Miciah: Consistency.

ironcladlou: Consistency at what cost? Doesn't it mean that half of the new code in the reconcile logic is actually unreachable?

https://github.com/openshift/cluster-ingress-operator/pull/289/files#diff-b3b3514d6621705093f820e6f9feead0R34-R41

Miciah: Some of the code is unreachable.  For now, we are open-coding the same pattern for every resource, but we want eventually to refactor all the ensure methods with the common logic (the ensure method itself) abstracted away from the resource-specific logic (get current X, get desired X, create, update, delete X).  If this isn't useful, I can delete the delete logic and simplify the switch to 2 cases.

ironcladlou:
Too bad the upgrade test can't exercise this.

pkg/operator/controller/ingress/monitoring.go
ironcladlou: Fair enough. I'll leave it up to you. Remove hold if you want to keep as is.

/hold

ironcladlou: /lgtm

Miciah:
> Too bad the upgrade test can't exercise this.

I'm testing manually and have discovered that `current.Object["Spec"]` and `expected.Object["Spec"]` apparently are not `map[string]interface{}` values.

Miciah:
> I'm testing manually and have discovered that `current.Object["Spec"]` and `expected.Object["Spec"]` apparently are not `map[string]interface{}` values.

...because the field is "spec", not "Spec".  Whoops!

ironcladlou:
Would an e2e test that did a 4.x to 4.x+1 upgrade have caught it?

Miciah:
> Would an e2e test that did a 4.x to 4.x+1 upgrade have caught it?

Only if it did an upgrade and then checked that the servicemonitor were updated.

PR #290, opened 2019-08-19
Bug 1740258: fix prometheus integration for multiple ingresscontrollers

ironcladlou:
Backport of patches which fix prometheus integration when more than one
ingresscontroller exists.

https://github.com/openshift/cluster-ingress-operator/pull/289
https://github.com/openshift/cluster-ingress-operator/pull/242

### DIFF
--- a/pkg/operator/controller/controller_service_monitor.go
+++ b/pkg/operator/controller/controller_service_monitor.go
@@ -3,11 +3,13 @@ package controller
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/controller_service_monitor.go
```diff
@@ -54,7 +64,11 @@ func desiredServiceMonitor(ci *operatorv1.IngressController, svc *corev1.Service
 						"openshift-ingress",
 					},
 				},
-				"selector": map[string]interface{}{},
+				"selector": map[string]interface{}{
+					"matchLabels": map[string]interface{}{
+						manifests.OwningIngressControllerLabel: ci.Name,
+					},
+				},
 				"endpoints": []map[string]interface{}{
```
Miciah: You need this change or else `serviceMonitorChanged` may panic on old resources and return false positives on updated resources:  https://github.com/openshift/cluster-ingress-operator/commit/4c56ec1fcfc538eb1801c914d807eec4d5e33775#diff-b3b3514d6621705093f820e6f9feead0R70-R80

ironcladlou: Fixed, thanks

ironcladlou:
https://github.com/openshift/origin/pull/23630 — if back ported — might be a way to automate testing of this PR.

eparis:
If this needs to be merged in lockstep with another PR please make that VERY clear for us dumb patch managers.

PR #291, opened 2019-08-20
Have HAProxy use modern SSL ciphers

wshearn:
for Jira COMPLY-67. Do not use TLS <1.2

I have tested this on my cluster and it gives the correct results.

/cc @jaybeeunix

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -130,6 +130,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 				Key: "statsPassword",
 			},
 		}},
+		{Name: "ROUTER_CIPHERS", Value: "modern"},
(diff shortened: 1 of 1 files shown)


ironcladlou:
@openshift/sig-network-edge wouldn't this now be a backwards-incompatible change?

Security API is being addressed generally in https://github.com/openshift/api/pull/350.

Miciah:
> @openshift/sig-network-edge wouldn't this now be a backwards-incompatible change?
> 
> Security API is being addressed generally in [openshift/api#350](https://github.com/openshift/api/pull/350).

Good point—the default is "Intermediate" now, and the proposed API codifies that default: https://github.com/openshift/api/pull/350/files#diff-85d84174432a6d2df99d107129e012f8R134

wshearn:
Fixed in the image here https://github.com/openshift/router/pull/30

Issue #285, opened 2019-08-13
Customize router deployment config to disable namespace ownership check


larhauga:
On OpenShift 3 it was possible to customize the default deployment of the router by adding environment variables with the needed customization. I cannot find a way of doing this with the cluster-ingress-operator, as the CRD does not contain any options for mutating the deployment.

The specific configuration that I need is `ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK`: https://github.com/openshift/router/blob/master/pkg/cmd/infra/router/router.go#L73-L92

The specific problem in question is related to solving "HostAlreadyClaimed" when binding to the same hostname (with different paths) in multiple namespaces.

Is there any other way of configuring the router/operator to add options? Or should the CRD be updated with the options for customizing the router?

xavierbaude:
I also using options : NAMESPACE_LABELS and ROUTE_LABELS in order to use private & public router depending need. Would you also add these option too ?

pashok2398:
In my case we need to rewrite the `X-Forwarded-For` header so i using the custom template in our ocp 4.1 installation from this guide:
https://docs.openshift.com/container-platform/3.3/install_config/router/customized_haproxy_router.html#using-configmap-replace-template
I add the `http-request set-header X-Forwarded-For %[src]` to base template and using it as configMap.
Is there any option to use custom template or any option to rewrite the `X-Forwarded-For` header with annotation?

pashok2398:
Hello. Sorry for my perseverance.
Is there any plans to implement this?
Thanks

haoqing0110:
Same question of how to use custom template in OCP 4.2 ? Any update for this issue?

pashok2398:
> Same question of how to use custom template in OCP 4.2 ? Any update for this issue?

+1

larhauga:
Hi, I have totally forgotten this issue, and for that I am sorry.
It took a while, but we managed to create a enhancement proposal for changing the `ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK` env variable on the router deployment.
https://github.com/openshift/enhancements/blob/master/enhancements/ingress/openshift-route-admission-policy.md

We ran with a custom router during the period where this was not possible, by assigning the default router to a specific node, while having a custom deployment of the router. You can then configure the router as you would like, without having the configuration in the Operator.

[...]

PR #292, opened 2019-08-21
disableNamespaceOwnershipCheck: Add optional router configuration

larhauga:
To enable multiple routes in different namespace to share the same domain,
the router will need to be configured to allow admission of multiple
routes from different namespaces.
This is supported in the openshift router and in OpenShift 3, and prevents migration for customers.

This commit fixes issue #285.

https://github.com/openshift/cluster-ingress-operator/issues/285

openshift/api crd change: https://github.com/openshift/api/pull/416

TODO: Update PR with go.mod after #416 is merged.

### DIFF
--- a/manifests/00-custom-resource-definition.yaml
+++ b/manifests/00-custom-resource-definition.yaml
@@ -418,6 +418,15 @@ spec:
                     TODO: Add other useful fields. apiVersion, kind, uid?'
                   type: string
               type: object
+            disableNamespaceOwnershipCheck:
+              description: "disableNamespaceOwnershipCheck disables the namespace
+                ownership checks for a route host with different paths or for overlapping
+                host names in the case of wildcard routes. Please be aware that if
+                namespace ownership checks are disabled, routes in a different namespace
+                can use this mechanism to 'steal' sub-paths for existing domains.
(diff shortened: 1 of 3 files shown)


ironcladlou:
@openshift/sig-network-edge @smarterclayton would it be accurate and/or useful to generalize namespace-boundaries and wildcard support as something like "route admission policies"?

For example, `AllowInterNamespaceClaims`

In any case they are both specific to Route resource.

ironcladlou:
@larhauga 

OpenShift now has an [enhancement proposal process](https://github.com/openshift/enhancements) for changes like this.

We won't be able to accept this PR as-is, because it's proposing API which must first be accepted in [openshift/api](https://github.com/openshift/api) for the operator to consume/implement.

If you're still interested in pursuing the change, I recommend you start with the new process and please tag @openshift/sig-network-edge for assistance. Thanks!

danehans:
This PR needs to be updated to accurately reflect https://github.com/openshift/enhancements/pull/43

Miciah:
https://github.com/openshift/cluster-ingress-operator/pull/345 supersedes this PR.  Thanks for making the first pass at implementing the feature!
/close

PR #294, opened 2019-08-23
Bug 1712525: certs: publish all in-use public cert data

ironcladlou:
Before this change, only the default generated CA tls.crt data was published to
the `router-ca` configmap in the `openshift-config-managed` namespace. Change
the behavior to also concatenate any tls.crt data referenced by an
ingresscontroller's default certificate configuration.

This change is necessary because the ingresscontroller is the only way users can
propagate their CA data to kube-controller-manager, which is required for TLS
Route connectivity from pods.

cc @openshift/sig-network-edge @enj


ironcladlou:
This is still an untested WIP, opening early for discussion.

pkg/operator/controller/certificate/publish_ca.go
```diff
@@ -52,10 +56,32 @@ func (r *reconciler) ensureRouterCAConfigMap(secret *corev1.Secret, ingresses []
...
+//
+// 1. The default generated CA tls.crt data
+// 2. All distinct default certificate secret tls.crt data referenced by ingresscontrollers
+func (r *reconciler) desiredRouterCAConfigMap(ingresses []operatorv1.IngressController) (*corev1.ConfigMap, error) {
+	caName := controller.RouterCASecretName(r.operatorNamespace)
+	secrets := map[string]types.NamespacedName{
+		caName.String(): caName,
+	}
+	for _, ingress := range ingresses {
+		name := controller.RouterEffectiveDefaultCertificateSecretName(&ingress)
```
ironcladlou: @Miciah @enj When, if ever, is it appropriate to publish the default generated certificate?

ironcladlou: In this PR's current implementation, the default generated certificate data is published when it's referenced by an ingresscontroller just like any other reference.

Miciah: > When, if ever, is it appropriate to publish the default generated certificate?

Never.  No consumer of `router-ca` should need the default certificate.  (It is in `router-certs` for consumers who do need it.)

Miciah: > the default generated certificate data is published when it's referenced by an ingresscontroller

It should only exist when it's referenced by an ingresscontroller, and it should be published in `router-certs` but not in `router-ca`.

ironcladlou: Fixed.

pkg/operator/controller/names.go
```diff
@@ -95,11 +95,11 @@ func RsyslogConfigMapName(ic *operatorv1.IngressController) types.NamespacedName
 
 // RouterEffectiveDefaultCertificateSecretName returns the namespaced name for
 // the in-use router default certificate secret.
-func RouterEffectiveDefaultCertificateSecretName(ci *operatorv1.IngressController, namespace string) types.NamespacedName {
-	if cert := ci.Spec.DefaultCertificate; cert != nil {
-		return types.NamespacedName{Namespace: namespace, Name: cert.Name}
+func RouterEffectiveDefaultCertificateSecretName(ic *operatorv1.IngressController) types.NamespacedName {
+	if cert := ic.Spec.DefaultCertificate; cert != nil {
+		return types.NamespacedName{Namespace: ic.Namespace, Name: cert.Name}
```
ironcladlou: Whoops, this is wrong. I think hard-coding the namespaces in this file would actually be more clear at the moment...

ironcladlou: Fixed

pkg/operator/controller/certificate/publish_ca.go
```diff
@@ -52,10 +56,35 @@ func (r *reconciler) ensureRouterCAConfigMap(secret *corev1.Secret, ingresses []
...
+		if cert := ingress.Spec.DefaultCertificate; cert != nil {
+			name := types.NamespacedName{Namespace: "openshift-ingress", Name: cert.Name}
+			secrets[name.String()] = name
+		}
+	}
+
+	var certs [][]byte
+	for _, name := range secrets {
+		secret := &corev1.Secret{}
+		if err := r.client.Get(context.TODO(), name, secret); err != nil {
```
Miciah: Better to list secrets use that list and the list of ingresses to build a map; see how `desiredRouterCertsGlobalSecret` builds `ingressToSecret`: https://github.com/openshift/cluster-ingress-operator/blob/b5d4fb5c3fae2c72f79dbc598f81c2271cd693cd/pkg/operator/controller/certificate-publisher/publish_certs.go#L59-L92

What do you think of keeping the parameter for the CA secret and adding a parameter for the list of secrets so that `desiredRouterCAConfigMap` doesn't need `r`?

ironcladlou:
Further discussion has revealed this fix to be incomplete. We need more design around the input/output expectations to try again.

PR #295, opened 2019-09-03
Bug 1747840: dns: add explicit TTL to all wildcard records

ironcladlou:
Before this commit, wildcard DNS resource record TTLs were inconsistent:

* AWS alias record TTLs are not user-configurable.
* Azure records TTLs were zero (which is essentially undefined and so causes
unpredictable client behavior)
* GCP record TTLs were 120 seconds.

After this commit, all DNS records default to 30 seconds for any provider which
supports setting a TTL for the record type (Azure and GCP at this time). This
brings some consistency, and most importantly, eliminates the possibility of
zero TTLs.

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/client-go/tools/record"
+
 	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
 	"github.com/openshift/cluster-ingress-operator/pkg/dns"
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
@@ -39,6 +41,7 @@ func New(mgr manager.Manager, dnsProvider dns.Provider) (runtimecontroller.Contr
 		client:      mgr.GetClient(),
 		cache:       mgr.GetCache(),
 		dnsProvider: dnsProvider,
+		recorder:    mgr.GetEventRecorderFor(controllerName),
 	}
 	c, err := runtimecontroller.New(controllerName, mgr, runtimecontroller.Options{Reconciler: reconciler})
 	if err != nil {
@@ -54,6 +57,7 @@ type reconciler struct {
 	client      client.Client
 	cache       cache.Cache
 	dnsProvider dns.Provider
+	recorder    record.EventRecorder
 }
 
 func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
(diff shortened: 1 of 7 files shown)


pkg/operator/controller/dns/controller.go
```diff
@@ -89,6 +93,15 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		log.V(2).Info("warning: dnsrecord is missing owner label", "dnsrecord", record, "ingresscontroller", ingressName)
 		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
 	}
+
+	// Existing 4.1 records will have a zero TTL. Instead of making all the client implementations guard against
+	// zero TTLs, simply ignore the record until the TTL is updated by the ingresscontroller controller. Report
+	// this through events so we can detect problems with our migration.
+	if record.Spec.RecordTTL <= 0 {
+		r.recorder.Eventf(record, "Warning", "ZeroTTL", "Record is missing TTL and will be temporarily ignored; the TTL will be automatically updated and the record will be retried.")
+		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
```
Miciah: `RequeueAfter` should be unnecessary—dns_controller's watch on DNSRecord resources should trigger reconciliation when ingress_controller updates the TTL.

pkg/operator/controller/ingress/dns.go
```diff
@@ -28,13 +39,23 @@ func (r *reconciler) ensureWildcardDNSRecord(ic *operatorv1.IngressController, s
...
+	case desired != nil && current == nil:
 		if err := r.client.Create(context.TODO(), desired); err != nil {
 			return nil, fmt.Errorf("failed to create dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
 		}
 		log.Info("created dnsrecord", "dnsrecord", desired)
-		return desired, nil
+		return r.currentWildcardDNSRecord(ic)
+	case desired != nil && current != nil:
+		if updated, err := r.updateDNSRecord(current, desired); err != nil {
+			return nil, fmt.Errorf("failed to update dnsrecord %q: %v", desired.Name, err)
```
Miciah: Could that be `%s/%s` and `desired.Namespace, desired.Name` for consistency?

pkg/operator/controller/ingress/dns.go
```diff
@@ -139,3 +161,29 @@ func (r *reconciler) deleteWildcardDNSRecord(ic *operatorv1.IngressController) e
...
+
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// dnsRecordChanges checks if the current DNSRecord spec matches the expected spec and
+// if not returns an updated one.
+func dnsRecordChanged(current, expected *iov1.DNSRecord) (bool, *iov1.DNSRecord) {
```
Miciah: Godoc does not match function name: `dnsRecordChanges`, `dnsRecordChanged`.

PR #296, opened 2019-09-05
Fix scale subresource's label selector path; set default ingresscontroller's replicas to non-nil

Miciah:
#### Bump `openshift/api`

Bump to github.com/openshift/api@e6c3eeae444f4eea11ef19fc011f63c9dfa61cf1 to get the correct label selector path in the scale subresource.

* `vendor/github.com/openshift/api/operator/v1/types_ingress.go`: Bump.
* `manifests/00-custom-resource-definition.yaml`: Regenerate.

**This PR needs to be rebased once https://github.com/openshift/api/pull/425 has merged.**

#### Add `k8s.io/client-go/scale` dependency

* `vendor/k8s.io/client-go/discovery/`:
* `vendor/k8s.io/client-go/scale/`:
* `vendor/modules.txt`: Import client-go's scale package in order to be able to use it in end-to-end tests.

#### Set default ingresscontroller's replicas to non-nil

Set a non-nil initial value for the `spec.replicas` field of the default ingresscontroller so that the scale client (which is used by the `oc scale` command) will work.

* `pkg/operator/operator.go (ensureDefaultIngressController)`: Set `spec.replicas` to 2.
* `test/e2e/operator_test.go` (`TestIngressControllerScale`): Use the scale client from client-go.
(`getScaleClient`): New function.  Return a scale client.


ironcladlou:
Should there be a bug associated with this (and the API) PR?

Will the CRD be automatically updated/repaired during upgrade?

Miciah:
> Should there be a bug associated with this (and the API) PR?

Probably.  I'll file a Bugzilla report for the incorrect label selector path.

> Will the CRD be automatically updated/repaired during upgrade?

Should be, since it's managed by CVO, but I'll check.

Miciah:
/retitle Bug 1749581: Fix scale subresource's label selector path; set default ingresscontroller's replicas to non-nil

Miciah:
> Will the CRD be automatically updated/repaired during upgrade?

The CVO does periodically synch the CRD.

Miciah:
Rebased.

PR #301, opened 2019-09-24
WIP: dep: bump to kubernetes 1.16

ironcladlou:
Depends on https://github.com/kubernetes-sigs/controller-runtime/pull/588.

TODO
- [ ] Verify correctness of the regenerated CRD manifests


manifests/00-custom-resource-definition.yaml
```diff
@@ -442,22 +88,23 @@ spec:
                 hostNetwork:
                   description: hostNetwork holds parameters for the HostNetwork endpoint
                     publishing strategy. Present only if type is HostNetwork.
+                  nullable: true
```
sttts: this makes the schema weaker. Probably you have a `// +kubebuilder:nullable=true` tag or so in the API types?

manifests/00-custom-resource-definition.yaml
```diff
@@ -442,22 +88,23 @@ spec:
...
                     Present only if type is LoadBalancerService.
+                  nullable: true
                   properties:
                     scope:
                       description: scope indicates the scope at which the load balancer
                         is exposed. Possible values are "External" and "Internal".  The
                         default is "External".
                       type: string
-                  required:
-                  - scope
```
sttts: this looks suspicious.

manifests/00-custom-resource-definition.yaml
```diff
@@ -481,8 +128,6 @@ spec:
                     networking, and is not explicitly published. The user must manually
                     publish the ingress controller."
                   type: string
-              required:
-              - type
```
sttts: this looks suspicious

manifests/00-custom-resource-definition.yaml
```diff
@@ -705,9 +361,6 @@ spec:
                     type: string
                   type:
                     type: string
-                required:
-                - type
-                - status
```
sttts: this looks suspicious.

manifests/00-custom-resource-definition.yaml
Miciah: 4.1 didn't have `scope`, so it cannot be required.

manifests/00-custom-resource-definition.yaml
Miciah: I added `+optional` when I added `+unionDiscriminator` in https://github.com/openshift/api/commit/2024105dd4c3cc30148d4127545eec7b21a53b65#diff-85d84174432a6d2df99d107129e012f8R246, but maybe I shouldn't have.  https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md#go-tags says, "`// +unionDiscriminator` before a field means that this field is the discriminator for the union. Only one field per structure can have this prefix. This field HAS TO be a string, and CAN be optional."

ironcladlou: @Miciah what do you propose we do here?

Miciah: I do not know of a reason why this field should be marked as optional (it came up during review: https://github.com/openshift/api/pull/312#discussion_r293022546, but I do not see a reason given why it should be optional).  `spec.endpointPublishingStrategy` is optional, so I believe `type` does not need to be so.  If it is not too late to do so, I would like to remove `+optional`.  The question that was raised in the discussion about the `+nullable` markers applies here:  Since the old generated CRD does not reflect the markers, is it safe to remove them?

ironcladlou: Like with nullable, how about open a PR and let's get the discussion centralized?

Miciah: https://github.com/openshift/api/pull/440

ironcladlou:
@Miciah updated to incorporate https://github.com/openshift/api/pull/440 and https://github.com/openshift/api/pull/438, PTAL

ironcladlou:
Waiting on https://github.com/openshift/library-go/pull/531 to pick up https://github.com/openshift/api/pull/442.

manifests/00-custom-resource-definition.yaml
ironcladlou: It's required in the 4.2 branch... https://github.com/openshift/cluster-ingress-operator/blob/release-4.2/manifests/00-custom-resource-definition.yaml#L455

manifests/00-custom-resource-definition.yaml
ironcladlou: Fixed

manifests/00-custom-resource-definition.yaml
```diff
@@ -719,22 +372,23 @@ spec:
...
                     Present only if type is LoadBalancerService.
+                  nullable: true
                   properties:
                     scope:
                       description: scope indicates the scope at which the load balancer
                         is exposed. Possible values are "External" and "Internal".  The
                         default is "External".
                       type: string
-                  required:
-                  - scope
```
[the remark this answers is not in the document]
ironcladlou: Status mirrored from spec... should it follow the same validation?

manifests/00-custom-resource-definition.yaml
```diff
@@ -766,12 +416,9 @@ spec:
                 controller pods corresponding to the IngressController. The number
                 of matching pods should equal the value of availableReplicas.
               type: string
-          required:
-          - availableReplicas
-          - selector
-          - domain
```
ironcladlou: @Miciah hmmm

manifests/00-custom-resource-definition.yaml
ironcladlou: Fixed in https://github.com/openshift/api/pull/438

manifests/00-custom-resource-definition.yaml
ironcladlou: cc @Miciah

manifests/00-custom-resource-definition.yaml
Miciah: This is in `status`, so I don't know that we care...?

manifests/00-custom-resource-definition.yaml
ironcladlou: @Miciah https://github.com/openshift/cluster-ingress-operator/blob/release-4.2/manifests/00-custom-resource-definition.yaml#L455 does this mean we need to change 4.2 before release? Is there a problem that has gone unnoticed?

manifests/00-custom-resource-definition.yaml
ironcladlou: @sttts Isn't this from https://github.com/openshift/api/blob/master/operator/v1/types.go#L148? They aren't `omit empty` and aren't explicitly marked required (the package default is optional). Is something wrong with the shared OperatorCondition struct's markers?

manifests/00-custom-resource-definition.yaml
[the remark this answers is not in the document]
ironcladlou: Seems like this should be optional and the requiredness was possibly a mistake. See https://github.com/openshift/cluster-ingress-operator/pull/301#discussion_r328353977.

manifests/00-custom-resource-definition.yaml
Miciah: Seems like this would be a problem.  I can understand how we would have missed this for `spec.endpointPublishingStrategy.scope`: `spec.endpointPublishingStrategy` is optional, and the operator-created default ingresscontroller does not specify it.  However, I wonder why we have not seen problems with `status.endpointPublishingStrategy.scope` during upgrade tests.  I'll launch a 4.1 cluster, set `spec.endpointPublishingStrategy` with null `spec.endpointPublishingStrategy.status`, upgrade the cluster to 4.2, and report back what happens.

manifests/00-custom-resource-definition.yaml
ironcladlou: I don't know how to think about this one. Maybe @deads2k can help or knows who can.

manifests/00-custom-resource-definition.yaml
ironcladlou: @Miciah based on your upgrade testing, seems like this is going to be okay?

manifests/00-custom-resource-definition.yaml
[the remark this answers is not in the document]
ironcladlou: @Miciah based on your upgrade testing, seems like this is going to be okay?

manifests/00-custom-resource-definition.yaml
ironcladlou: Discussion in https://github.com/openshift/api/pull/445

manifests/00-custom-resource-definition.yaml
Miciah: Right, the changing `required` value did not seem to have any effect: I performed the steps I laid out in my previous comment and saw no errors in the operator logs or command output when the upgraded ingress operator reconciled the old IngressController resource that had `spec.endpointPublishingStrategy` but not `spec.endpointPublishingStrategy.scope` set and when I tried to update the resource using `oc annotate`.

Miciah: Facepalm: `endpointPublishingStrategy.loadBalancer` is optional, so `endpointPublishingStrategy.loadBalancer.scope` can be required.  I erroneously wrote `endpointPublishingStrategy.scope` instead of `endpointPublishingStrategy.loadBalancer.scope` earlier.  I think 4.2 is fine to make `scope` required, and we should keep it required in 4.3.

ironcladlou: https://github.com/openshift/api/pull/446 will decide the fate of this field

manifests/00-custom-resource-definition.yaml
ironcladlou: Fixed here.

manifests/00-custom-resource-definition.yaml
ironcladlou: Fixed.

manifests/00-custom-resource-definition.yaml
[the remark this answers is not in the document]
ironcladlou: Fixed.

ironcladlou:
@Miciah I think I got this all cleaned up with the exception of https://github.com/openshift/api/pull/445 which we can address in a followup as necessary. PTAL.

Miciah:
We need to sort out https://github.com/openshift/api/pull/445 as well https://github.com/openshift/cluster-ingress-operator/pull/301/commits/1e17863695a9eb4923b35893a9552c32e6709390#r328396676 (concerning `status.availableReplicas`, `status.selector`, and `status.domain`), but we can get those in a follow-up.
/lgtm

PR #302, opened 2019-09-24
Update OWNERS

frobware:
Add 'frobware'.


Miciah:
Finally!  That leaves https://github.com/openshift/cluster-dns-operator and https://github.com/openshift/router...

PR #307, opened 2019-10-07
Check for Kubernetes 1.15 LB provisioning failure events

miheer:
Update the ingress operator's computeLoadBalancerStatus function's computation of the LoadBalancerReady condition to check for SyncLoadBalancerFailed-type events from the service controller.

https://jira.coreos.com/browse/NE-220

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -197,13 +197,13 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
 		message := "The LoadBalancer service is pending"
 
 		// Try and find a more specific reason for for the pending status.
-		createFailedReason := "CreatingLoadBalancerFailed"
+		createFailedReason := "SyncLoadBalancerFailed"
 		failedLoadBalancerEvents := getEventsByReason(operandEvents, "service-controller", createFailedReason)
 		for _, event := range failedLoadBalancerEvents {
 			involved := event.InvolvedObject
-			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name {
-				reason = "CreatingLoadBalancerFailed"
-				message = fmt.Sprintf("The %s component is reporting CreatingLoadBalancerFailed events like: %s\n%s",
+			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
+				reason = "SyncLoadBalancerFailed"
+				message = fmt.Sprintf("The %s component is reporting SyncLoadBalancerFailed events like: %s\n%s",
 					event.Source.Component, event.Message, "The kube-controller-manager logs may contain more details.")
 				break
 			}
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	types "k8s.io/apimachinery/pkg/types"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 )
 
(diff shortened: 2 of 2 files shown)


ironcladlou:
/approve

Wish we could e2e test this, but I'm still not sure how. If this event actually changed in 1.15, should we file a bug instead of a story to track it?

Miciah:
I figured out a way of testing this change:
1. Create the LoadBalancer service for an ingresscontroller named "test" (the service must be created before the ingresscontroller), with an invalid `service.beta.kubernetes.io/aws-load-balancer-proxy-protocol` annotation:
```shell
oc create -f - <<'EOF'
kind: Service
apiVersion: v1
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: nonsense
  finalizers:
  - ingress.openshift.io/operator
  - service.kubernetes.io/load-balancer-cleanup
  labels:
    app: router
    ingresscontroller.operator.openshift.io/owning-ingresscontroller: test
    router: router-test
  name: router-test
  namespace: openshift-ingress
  labels:
    app: router
spec:
  externalTrafficPolicy: Local
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: http
  - name: https
    port: 443
    protocol: TCP
    targetPort: https
  selector:
    ingresscontroller.operator.openshift.io/deployment-ingresscontroller: test
  type: LoadBalancer
EOF
```
2. Create the "test" ingresscontroller:
```shell
oc create -f - <<'EOF'
kind: IngressController
metadata:
  name: test
spec:
  domain: apps.example.com
EOF
```
3. Check the ingresscontroller's `LoadBalancerReady` status condition:
```shell
oc -n openshift-ingress-operator get ingresscontrollers/test -o yaml
```
```yaml
  - lastTransitionTime: "2019-10-10T00:01:22Z"
    message: |-
      The service-controller component is reporting SyncLoadBalancerFailed events like: Error syncing load balancer: failed to e
nsure load balancer: annotation "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"="nonsense" detected, but the only
value supported currently is '*'
      The kube-controller-manager logs may contain more details.
    reason: SyncLoadBalancerFailed
    status: "False"
    type: LoadBalancerReady
```

pkg/operator/controller/ingress/status.go
```diff
@@ -197,13 +197,13 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
 		// Try and find a more specific reason for for the pending status.
-		createFailedReason := "CreatingLoadBalancerFailed"
+		createFailedReason := "SyncLoadBalancerFailed"
 		failedLoadBalancerEvents := getEventsByReason(operandEvents, "service-controller", createFailedReason)
 		for _, event := range failedLoadBalancerEvents {
 			involved := event.InvolvedObject
-			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name {
-				reason = "CreatingLoadBalancerFailed"
-				message = fmt.Sprintf("The %s component is reporting CreatingLoadBalancerFailed events like: %s\n%s",
+			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
```
Miciah: Please modify the unit tests to verify this change.  I suggest adding a `uid` parameter to `pendingLBService` and to `failedCreateLBEvent`: https://github.com/openshift/cluster-ingress-operator/blob/d7a8e3225007bcb0874f7ce440c607fac4ca9fc0/pkg/operator/controller/ingress/status_test.go#L35 https://github.com/openshift/cluster-ingress-operator/blob/d7a8e3225007bcb0874f7ce440c607fac4ca9fc0/pkg/operator/controller/ingress/status_test.go#L70
Then modify the "no events for current lb" test case: https://github.com/openshift/cluster-ingress-operator/blob/d7a8e3225007bcb0874f7ce440c607fac4ca9fc0/pkg/operator/controller/ingress/status_test.go#L126-L137
Use `service: pendingLBService("default", 1)`, and add `failedCreateLBEvent("default", 2)` to `events` so that the test case verifies that `computeLoadBalancerStatus` ignores events where the name matches but the UID does not match.

Miciah: Is it useful to compare both the name and the uid, or should we delete the name comparison?

pkg/operator/controller/ingress/status_test.go
```diff
@@ -125,10 +128,10 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
 		{
 			name:       "no events for current lb",
 			controller: ingressController("default", operatorv1.LoadBalancerServiceStrategyType),
-			service:    pendingLBService("default"),
+			service:    pendingLBService("default", "1"),
 			events: []corev1.Event{
 				schedulerEvent(),
-				failedCreateLBEvent("secondary"),
```
Miciah: Keep "secondary" (with a distinct UID).

Issue #299, opened 2019-09-19
Enhance cluster-ingress-operator to support libvirt


ahrtr:
Currently cluster-ingress-operator only supports AWS, Azure and GCP.  But in development environment, i.e. CI, libvirt is useful. So it'd better to enhance ingress-operator to support libvirt.

jichenjc:
some to do item: 

[ ] https://github.com/openshift/cluster-ingress-operator/blob/master/cmd/ingress-operator/main.go#L123 not support libvirt, need add a libvirt branch here

PR #308, opened 2019-10-08
Enhance cluster-ingress-operator to support libvirt

gyohuangxin:
Related to #299.
Goal of this PR is enhancing the support to libvirt of cluster-ingress-operator. A new DNS provider for libvirt is created, then dns records are managed by XML through the DNS provider.
However, #299 still **workes in progress**, and the code **doesn't work**, what blocks us is the communication between Ingress and libvirt.
We will try to complete it as soon as possible.


jaypoulz:
Hi @gyohuangxin -
To answer your questions from https://github.com/openshift/cluster-ingress-operator/issues/310

1. The Ensure and Delete calls are referenced in the controller's Reconcile method.
See https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L63.
The path for the Ensure call is:
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L134
The delete call actually goes into a subroutine, but it's still there:
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L82
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L166

So now the question would be, what calls reconcile?
I'm looking into this now.

2. Investigation in progress - I will update this comment.

gyohuangxin:
Sorry for misoperating to close the pr.

@jaypoulz Thanks for your professional answer. It helps me a lot to understand the code, and any process will be updated here.

In addition, I've updated my code. Could you pls take a look?

gyohuangxin:
Reopening it.

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (p *Provider) Delete(record *iov1.DNSRecord, zone configv1.DNSZone) error {
+	return p.Change(record, zone, deleteAction)
+}
+
+// Change methods will perform an action on a record.
+func (p *Provider) Change(record *iov1.DNSRecord, zone configv1.DNSZone, action action) error {
+	// Create a new connections to qemu
+	conn, err := libvirt.NewConnect("qemu:///system")
```
zeenix: The connection URL should not be hardcoded. You'll need to get this somehow from the Installer, just like libvirt provider does: https://github.com/openshift/cluster-api-provider-libvirt/blob/master/pkg/apis/libvirtproviderconfig/v1beta1/libvirtmachineproviderconfig_types.go#L20

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,109 @@
...
+	ips, err := net.LookupIP(p.config.Domain)
+	if err != nil || len(ips) == 0 {
+		log.Error(err, "failed to lookup app domain's IPs")
+	}
+	log.Info("Domain nslookup", "IPs", ips)
+
+	// Find the correct network and update a DNS record
+	for _, network := range networks {
+		name, err := network.GetName()
+		if err == nil && strings.Index(name, p.config.Cluster) != -1 {
```
zeenix: Quite a bit of nesting here and that's undesirable. I'll suggest avoiding that by skipping the loop iteration instead:

```
if err != nil || strings.Index(name, p.config.Cluster) == -1 {
   continue;
```

Also you should be logging the error somewhere so it'll be to `if` statements but still one less level of nesting.

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,109 @@
...
+				// Generate a XML for DNS host
+				v := &Host{Ip: ip.String(), DnsRecord: record.Spec.DNSName}
+				output, err := xml.MarshalIndent(v, "  ", "    ")
+				if err != nil {
+					log.Error(err, "failed to generate a network XML")
+				}
+				log.Info("DNS record is updating", "XML", string(output))
+				// Update the network
+				log.Info("DNS record is updating", "action", action)
+				err2 := network.Update(action, 10, -1, string(output), 0)
```
zeenix: you don't need a separate variable `err2` here. Just re-use `err` here by using `=` instead of `:=`.

jaypoulz:
@gyohuangxin 
Regarding the HostNetworkStrategy:
The only place I see it being used is here: https://github.com/openshift/cluster-ingress-operator/blob/ad258253c7f471b38cca9279bbc560c8bd54777a/pkg/operator/controller/ingress/deployment.go#L228-L239
Here's my interpretation of this comment:
`// DNS is only managed for LB publishing. `
What I think that comment means: the only DNS record that actually gets added as part of the ingress operator is the wildcard DNS address that, if I understand correctly, is pointing to the LoadBalancer.

That said - your concern is exactly right:
The record that needs to get injected gets created here:
https://github.com/openshift/cluster-ingress-operator/blob/370a046016b9ccdbea95f7d1331e6414b14553e4/pkg/operator/controller/ingress/dns.go#L71-L111

Which means it's not getting created for libvirt because we've set the endpoint type to HostNetworkStrategy. I'm looking into what properties a LoadBalancerServiceStrategyType has and where it is referenced. Looks to me like these are the references:
```
pkg/operator/controller/ingress/load_balancer_service.go:87:	if ci.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
pkg/operator/controller/ingress/dns.go:79:	if ic.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
pkg/operator/controller/ingress/deployment.go:170:	if ci.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType {
pkg/operator/controller/ingress/controller.go:259:			strategyType = operatorv1.LoadBalancerServiceStrategyType
pkg/operator/controller/ingress/controller.go:270:	case operatorv1.LoadBalancerServiceStrategyType:
pkg/operator/controller/ingress/status.go:160:		ic.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
pkg/operator/controller/ingress/status.go:254:	if ic.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
```

jaypoulz:
@gyohuangxin I spoke to @crawford and he explained to me that we don't want to overload the LoadBalancingServiceStrategyType with the plain round-robin DNS that dnsmasq supports. He suggested that we make a new Strategy type called `RRDNSServiceStrategyType` that behaves like `HostNetworkStrategyType` except that it also inserts the wildcard DNS record.

That is to say, we will need to add an entry here:
vendor/github.com/openshift/api/operator/v1/types_ingress.go:167

Change this line to be the new strategy:
pkg/operator/controller/ingress/controller.go:261

Ensure these lines also apply to the new strategy:
https://github.com/openshift/cluster-ingress-operator/blob/ad258253c7f471b38cca9279bbc560c8bd54777a/pkg/operator/controller/ingress/deployment.go#L228-L239

And flip this line to include the new strategy:
https://github.com/openshift/cluster-ingress-operator/blob/370a046016b9ccdbea95f7d1331e6414b14553e4/pkg/operator/controller/ingress/dns.go#L79

There may also be changes needed in the controller to ensure that the dns records are being created for the new strategy:
These lines are the likely culprits:
pkg/operator/controller/ingress/controller.go:259
pkg/operator/controller/ingress/controller.go:270

zeenix:
@jaypoulz Thanks for the update.

> He suggested that we make a new Strategy type called `RRDNSServiceStrategyType` that behaves like `HostNetworkStrategyType` except that it also inserts the wildcard DNS record.

I was under the impression that the advantage of patching cluster-ingress-operator was that we don't need a wildcard DNS record. If we want to use wildcards, why not do that from Installer through terraform?

gyohuangxin:
@zeenix Thanks for your code reviewing, and the code is modified as you suggested and pushed again, which I learn a lot from.

cmd/ingress-operator/main.go
```diff
@@ -174,6 +176,23 @@ func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, d
 			return nil, fmt.Errorf("failed to create GCP DNS provider: %v", err)
 		}
 		dnsProvider = provider
+	case configv1.LibvirtPlatformType:
+		cluster := strings.Split(dnsConfig.Spec.BaseDomain, ".")[0]
+		appDomain := strings.Replace(dnsConfig.Spec.BaseDomain, cluster, "apps", 1)
+		// LIBVIRT_URL must be set as 'qemu+ssh://x.x.x.x/system'
+		libvirtUrl := os.Getenv("LIBVIRT_URI")
```
zeenix: No, you need to get this passed by the Installer to ensure you're using the same connection as Installer and other components. As i said before, have a look at how `cluster-api-provider-libvirt` is doing it.

cmd/ingress-operator/main.go
```diff
@@ -174,6 +176,23 @@ func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, d
 			return nil, fmt.Errorf("failed to create GCP DNS provider: %v", err)
 		}
 		dnsProvider = provider
+	case configv1.LibvirtPlatformType:
+		cluster := strings.Split(dnsConfig.Spec.BaseDomain, ".")[0]
+		appDomain := strings.Replace(dnsConfig.Spec.BaseDomain, cluster, "apps", 1)
+		// LIBVIRT_URL must be set as 'qemu+ssh://x.x.x.x/system'
```
zeenix: Also you must not mandate a remote URL here as we've local case too.

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Lookup app domain's ip
+	ips, err := net.LookupIP(p.config.Domain)
+	if err != nil || len(ips) == 0 {
+		log.Error(err, "failed to lookup app domain's IPs")
+	}
+	log.Info("Domain nslookup", "IPs", ips)
+
+	// Find the correct network and update a DNS record
+	for _, network := range networks {
```
zeenix: you don't need to manually find the network. You can use the `LookupNetworkByName` function:  https://github.com/libvirt/libvirt-go/blob/master/connect.go#L1159

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,114 @@
...
+		name, err := network.GetName()
+		if err != nil {
+			log.Error(err, "failed to get network name")
+		}
+		if strings.Index(name, p.config.Cluster) == -1 {
```
zeenix: that's a very strange way to compare strings. Why not just `string1 == string2`?

pkg/dns/libvirt/provider.go
```diff
@@ -0,0 +1,114 @@
...
+			log.Info("find the network but not in use", "network", name)
```
zeenix: Not sure i understand this log message but that's ok since this will go away hopefully when you update the code to make use of `LookupNetworkByName`.

gyohuangxin: Thanks,  we need to find the network which is used by cluster. For example, when cluster's name is "test1", the network is named as "test1-xxx", which we need to compare cluster's name and network's name to find out the network in use. 
LookupNetworkByName function try to lookup a network on the given hypervisor based on its name, so we need to find out the network name  too. BTW, is the network name passed by installer?

gyohuangxin:
> @gyohuangxin: PR needs rebase.

I pushed my code with some wrong commits, I am fixing it.

pkg/dns/libvirt/provider.go
zeenix: The network name is the same as `ClusterID`, which is used for all assets (e.g storage pool and volume) and is based on the cluster name and a random string.

zeenix: And you need to ensure you're touching the network of the specific cluster and hence look for the network that has the same name as the cluster ID.

go.mod
```diff
@@ -17,15 +17,12 @@ require (
...
 	github.com/kevinburke/go-bindata v3.11.0+incompatible
-<<<<<<< HEAD
 	github.com/openshift/api v3.9.1-0.20191015204012-3f47c9a8fa7f+incompatible
 	github.com/openshift/library-go v0.0.0-20190927184318-c355e2019bb3
-=======
 	github.com/libvirt/libvirt-go v5.8.0+incompatible
 	github.com/openshift/api v3.9.1-0.20190905013149-e6c3eeae444f+incompatible
 	github.com/openshift/library-go v0.0.0-20190613200606-e617f832835d
 	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709 // indirect
->>>>>>> Enhance cluster-ingress-operator to support libvirt
```
zeenix: you comitted a merge conflict. Please read up on git merge-conflicts and how to resolve them.

go.mod
```diff
@@ -17,8 +17,15 @@ require (
...
 	github.com/kevinburke/go-bindata v3.11.0+incompatible
+<<<<<<< HEAD
 	github.com/openshift/api v3.9.1-0.20191015204012-3f47c9a8fa7f+incompatible
 	github.com/openshift/library-go v0.0.0-20190927184318-c355e2019bb3
+=======
+	github.com/libvirt/libvirt-go v5.8.0+incompatible
+	github.com/openshift/api v3.9.1-0.20190905013149-e6c3eeae444f+incompatible
+	github.com/openshift/library-go v0.0.0-20190613200606-e617f832835d
+	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709 // indirect
+>>>>>>> Enhance cluster-ingress-operator to support libvirt
```
zeenix: merge conflict, committed here against instead or resolving it.

gyohuangxin: I'm sorry for my carelessness, I will fix it. Thanks.

jichenjc:
@jaypoulz hi, all discussion above seems related to wildcard DNS record, I checked the code and can't find where the other records such as console.apps.test1.aa.testing are created..

here's the workaround we are going to add but after I created the cluster, I didn't see DNSRecord got created.. guess it might be somewhere in libvirt installer?
```
  <hostname>oauth-openshift.apps.test1.aa.testing</hostname>
  <hostname>console-openshift-console.apps.test1.aa.testing</hostname>
  <hostname>downloads-openshift-console.apps.test1.aa.testing</hostname>
  <hostname>alertmanager-main-openshift-monitoring.apps.test1.aa.testing</hostnam
  <hostname>grafana-openshift-monitoring.apps.test1.aa.testing</hostname>
  <hostname>prometheus-k8s-openshift-monitoring.apps.test1.aa.testing</hostname>
```


I think the cluster should have some CRD for dnsrecords? 
I manually created a DNSRecord and I thought the DNSController should watch this 
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L63
and update use it, but because I don't have a label, I should see some warning 
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/dns/controller.go#L93
but not see it in my cluster......
```
kind: DNSRecord
metadata:
  name: dnsrecord1
spec:
  dnsName: "dns.jichen.test.com"
  targets: ["1.2.3.4"]
  recordType: "CNAME"


[core@test1-kmhsf-master-0 ~]$ oc get dnsrecord --all-namespaces
NAMESPACE   NAME         AGE
n1          dnsrecord1   8m58s

```

so overall question is 
1) should we create a DNSRecord for each dns such as oauth-openshift.apps.test1.aa.testing in a successful cluster? if we do, where the DNSRecord is created? 
2) curious why manual create a DNSRecord does not trigger a reconcile
`oc logs ingress-operator-6d9b5bf859-hfrcd  -n openshift-ingress-operator` doesn't show the logs... I will debug here more , any info is helpful

jaypoulz:
@jichenjc 
My understanding is that we only manage DNS for x86 cloud workloads.
https://github.com/openshift/cluster-ingress-operator/blob/370a046016b9ccdbea95f7d1331e6414b14553e4/pkg/operator/controller/ingress/dns.go#L78

Furthermore, those workloads add a single DNS record - the wildcard record.
https://github.com/openshift/cluster-ingress-operator/blob/370a046016b9ccdbea95f7d1331e6414b14553e4/pkg/operator/controller/ingress/dns.go#L111-L137

We have a few options. We can create a wildcard using the passthrough special libvirt dnsmasq namespace, but this has a few drawbacks. Here are some of Xin Huang's findings:

> This feature needs libvirt upgraded to 5.60 (https://libvirt.org/formatnetwork.html#elementsNamespaces). But it doesn't seem to be working, and it requires libvirt's network restarted and the domain must attach the network again, which will result in domain IP changes and network disruptions.

The alternative is to create the records manually according to their routes, as you have described above. This should allows us to insert the records without having to restart the networking.

So to answer your questions directly:
1. Inserting a DNS record per route seems to be the currently explored workaround. I don't know if this is good though, since we will have to maintain the logic that does this to support dynamic routes. If we solve this like this, I imagine this work would have to live in the ingress operator instead of the installer in order to support dynamic routes. Unless the service that creates the routes are calling into the ingress operator to ensure there's a DNS record for that route, we would likely have to add logic there too. @crawford does this track with your understanding?

I think probably a better solution is to add the wild-card record during the installer, were we can set requirements for the libvirt version and do this one-time setup task, without having to worry about changes created by new routes. This way, we can leave libvirt as a HostNetworkStrategy ingress type, and focus on getting the networking done in the installer - where a network restart is less likely to cause disruption for the cluster.

2. My guess is that this is because of https://github.com/openshift/cluster-ingress-operator/blob/370a046016b9ccdbea95f7d1331e6414b14553e4/pkg/operator/controller/ingress/dns.go#L78
Since we aren't using a LoadBalancingStrategyType, the dns controller is never created, and thus doesn't trigger a reconcile.

gyohuangxin:
@jaypoulz Thanks for you suggestion.

We discussed with the strategy type issue before,https://github.com/openshift/cluster-ingress-operator/pull/308#issuecomment-541317797, 
> That said - your concern is exactly right:
> The record that needs to get injected gets created here

and my concern may have been confirmed  after Chen's attempt. I will have a further look.

I agree with you very much, but need to figure out whether **remote issues** need attention at this stage. 
Because the way to inject wildcard DNS record in installer by configure host's dnsmasq doesn't support remote well, it doesn't work among multiple hosts(maybe need special configuration). @zeenix Is there any problem with my understanding?
So we can use Libvirt 5.6's feature to tell installer to inject dnsmasq namespaces into the XML of network which remote nodes attach. @zeenix Is there any problem with my understanding?

Miciah:
> 2\. `oc logs ingress-operator-6d9b5bf859-hfrcd  -n openshift-ingress-operator` doesn't show the logs... I will debug here more , any info is helpful

Try creating it in the `openshift-ingress` namespace.  As you mentioned, you will also need to add a `ingresscontroller.operator.openshift.io/owning-ingresscontroller` label referencing an ingresscontroller.

gyohuangxin:
@jaypoulz @Miciah Another question please, I tried to debug the code again and found something confusing.

I logged here and wanted to know the result of c.
https://github.com/openshift/cluster-ingress-operator/blob/066c8ac03bc25c526b57083b7f37d6653c425ae2/pkg/operator/controller/dns/controller.go#L46
However, the result of c is `"Error": "json: unsupported type: transport.WrapperFunc"` and its type is `*controller.Controller`.

What does it mean and where does the `transport.WrapperFunc` come from?

jichenjc:
@zeenix @jaypoulz 
based on @gyohuangxin 's find and talks above, seems the ingress is tight to LB very much and short term do the update seems hard and complicated, 

>>I think probably a better solution is to add the wild-card record during the installer, were we can set requirements for the libvirt version and do this one-time setup task,

I fully agree this , as we any way will have a SSH access from terraform provider to libvirtd host 
from https://github.com/dmacvicar/terraform-provider-libvirt/pull/661

I think the way for us to proceed can be:

1) add a script to generate from cluster and domain 
such as `prometheus-k8s-openshift-monitoring.apps.test1.aa.testing` 
maybe wildcard is another option..
2) ask CI/CD to ssh to the libvirtd host (the remote one) and put those into /etc/hosts or DNS record, or use virsh net-update command to update
3) then the host will have those info ,just like the workaround we provided in the installer

in the mean time, let's work on ingress operator to make it work by adding those DNS records for long term.. any thoughts?

jaypoulz:
@jichenjc 
I agree with the overall vision. My question is one point 2:
> ask CI/CD to ssh to the libvirtd host (the remote one) and put those into /etc/hosts or DNS record, or use virsh net-update command to update
From the CI/CD pod, we could probably do this. That said, is there any value in adding this workaround to the installer itself for installations of type libvirt? Is the reason we're moving this logic out into the CI/CD code because we can't solve this using a non ssh based work around (e.g. guestfish or something of that nature)?.

gyohuangxin:
> @gyohuangxin
> I can't speak to the error you're seeing. I can point you to the k8s runtime controller constructor:
> https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L63
> 
> Based on the value of `c`, I would have guessed that it was nil, not an error value.
> 
> That said, the only reference I see to transport is:
> https://godoc.org/google.golang.org/api/transport
> https://github.com/golang/go/blob/1a058cd04b7b902e59ff5584c7fda8ddfe388981/src/net/http/transport.go
> 
> If I had to make an educated guess, I was say the the request that was made of the runtimecontroller controller was unsupported - possibly because it's expecting json and not receiving it, or something of that nature. For further hints, I would try logging the values of the underlying objects being passed to that constructor.

@jaypoulz Thanks, as Chen's saying, the Ingress's code is unfriendly with libvirt platform. We will keep our trying, and, at this stage, we will try to inject the wildcard DNS record in host's dnsmaq in installer (implement the workaround in installer). 

> @jichenjc
> I agree with the overall vision. My question is one point 2:
> 
> > ask CI/CD to ssh to the libvirtd host (the remote one) and put those into /etc/hosts or DNS record, or use virsh net-update command to update
> 
> From the CI/CD pod, we could probably do this. That said, is there any value in adding this workaround to the installer itself for installations of type libvirt? Is the reason we're moving this logic out into the CI/CD code because we can't solve this using a non ssh based work around (e.g. guestfish or something of that nature)?
> 
> Also, @crawford - this would be a "manual" step in an otherwise purely automated ipi set-up (as least for the short term). Is this acceptable?

@jaypoulz It's a must to add workaround to the installer, but it doesn't support remote well. What Chen means is that if we have a remote node, we can inject the same wildcard DNS record in its dnsmasq by ssh from CI/CD. @jichenjc Is my understanding wrong?

jichenjc:
yes, @gyohuangxin ,that's exactly what I mean,

ask the node running terraform installer to ssh to the node running libvirt remotely and inject the record to dnsmasq.. thanks

jaypoulz:
Thanks for figuring all of this out. I believe this should be closed in favor of a workaround specific to our test setup. Please see the comments:
https://github.com/openshift/installer/pull/2600#issuecomment-549921268
https://github.com/openshift/installer/pull/2600#issuecomment-549929453

ironcladlou:
https://github.com/openshift/cluster-ingress-operator/pull/308#issuecomment-549932070

/close

PR #309, opened 2019-10-10
Add a render command to facilitate install-time customizations

ironcladlou:
Introduce a `render` command which emits the most essential manifests required
to customize ingress during installation. For now, the command emits the
operator namespace and CRD, which should be enough to allow the creation of
`ingresscontroller.operator.openshift.io` resources before the operator
manifests are created by the CVO.

This enables the installer to create a new `ingresscontroller` before the
ingress operator starts, effectively allowing users to bring their own complete
default `ingresscontroller` resource.


Dockerfile
```diff
@@ -6,10 +6,8 @@ RUN make build
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /ingress-operator/ingress-operator /usr/bin/
 COPY manifests /manifests
-RUN useradd ingress-operator
-USER ingress-operator
```
ironcladlou: This new user should be unnecessary, and by using, the installer (as an image consumer) must override the user when doing things like running `render` with a volume mounted output directory (because `ingress-operator` user doesn't have permissions to write to the volume mount point).

Although I can work around it in the installer, if we have no reason to run as a new user, removing the `useradd` makes things more consistent with the other images used for the same purpose.

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,39 @@
...
+	files := []string{
+		manifests.CustomResourceDefinitionManifest,
+		manifests.NamespaceManifest,
+	}
+
+	if dir == "-" {
+		for _, file := range files {
+			fmt.Printf("# file: %s\n", filepath.Base(file))
+			fmt.Println(manifests.MustAssetString(file))
+			fmt.Println("---")
```
Miciah: Write `---` before, and (optionally) `...` after each file.

cmd/ingress-operator/main.go
```diff
@@ -1,226 +1,51 @@
...
+		},
 	}
 
-	operatorConfig := operatorconfig.Config{
-		OperatorReleaseVersion: releaseVersion,
-		Namespace:              operatorNamespace,
-		IngressControllerImage: ingressControllerImage,
+	var renderOutputDir string
+	var renderPrefix string
+	var cmdRender = &cobra.Command{
```
Miciah: What do you think about defining a `NewRenderCommand` in `cmd/ingress-operator/render.go` so we can move the command and parameter definitions there, and likewise for the `start` command?

cmd/ingress-operator/main.go
```diff
@@ -1,226 +1,51 @@
...
-		os.Exit(1)
-	}
-
-	platformStatus, err := getPlatformStatus(kubeClient, infraConfig)
-	if err != nil {
-		log.Error(err, "failed to get platform status")
-		os.Exit(1)
+	var cmdStart = &cobra.Command{
+		Use:   "start",
+		Short: "Starts the operator",
```
Miciah: "Starts" → "Start".

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		return nil
+	}
+
+	if err := os.MkdirAll(dir, 0666); err != nil {
```
Miciah: Wouldn't `0755` make more sense?

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+)
+
+func render(dir string, prefix string) error {
+	files := []string{
+		manifests.CustomResourceDefinitionManifest,
+		manifests.NamespaceManifest,
```
Miciah: Could we also write out a default ingresscontroller resource so the user does not need to write that from scratch?

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		return nil
+	}
+
+	if err := os.MkdirAll(dir, 0666); err != nil {
+		return fmt.Errorf("failed to create output directory %q: %v", dir, err)
+	}
+
+	for _, file := range files {
+		outputFile := filepath.Join(dir, prefix+filepath.Base(file))
```
Miciah:
```suggestion
		outputFile := filepath.Join(filepath.Join(dir, prefix), filepath.Base(file))
```

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,39 @@
...
+		if err := ioutil.WriteFile(outputFile, manifests.MustAsset(file), 0644); err != nil {
+			return fmt.Errorf("failed to write %s: %v", outputFile, err)
```
Miciah: `%s` → `%q`.

hack/update-generated-bindata.sh
```diff
@@ -5,6 +5,6 @@ OUTDIR="${OUTDIR:-$PWD}"
 
 # Using "-modtime 1" to make generate target deterministic. It sets all file
 # time stamps to unix timestamp 1
-GO111MODULE=on GOFLAGS=-mod=vendor go run github.com/kevinburke/go-bindata/go-bindata -mode 420 -modtime 1 -pkg manifests -o ${OUTDIR}/pkg/manifests/bindata.go assets/...
+GO111MODULE=on GOFLAGS=-mod=vendor go run github.com/kevinburke/go-bindata/go-bindata -mode 420 -modtime 1 -pkg manifests -o ${OUTDIR}/pkg/manifests/bindata.go assets/... manifests/...
```
Miciah: Instead of `manifests/...`, can we build in only the manifests we need?

cmd/ingress-operator/render.go
ironcladlou: Can we have a separate followup discussion about the default handling in that regard? It's not required for this part of the implementation and could have negative consequences to consider.

cmd/ingress-operator/render.go
ironcladlou: I don't think that's the same. The prefix is supposed to be prepended to the file base name, not the output directory.

hack/update-generated-bindata.sh
ironcladlou: go-bindata takes a directory list, not sure there's any real harm here other than binary size inflation?

cmd/ingress-operator/main.go
ironcladlou: Done

cmd/ingress-operator/main.go
ironcladlou: Fixed

cmd/ingress-operator/render.go
ironcladlou: Yes, this actually explains a perm issue I accidentally worked around. Thanks!

cmd/ingress-operator/render.go
ironcladlou: Fixed

cmd/ingress-operator/render.go
ironcladlou: Also fixed the perms of the yaml itself

ironcladlou:
@Miciah latest feedback addressed, PTAL.

cmd/ingress-operator/render.go
Miciah: Ah, right, sorry.  `filepath.Join` would be inappropriate because the prefix is part of the filename, not a separate path component.

hack/update-generated-bindata.sh
Miciah: If specifying files individually is not an option, then another option would be to copy the desired files to a temporary directory and specify that directory instead.  But you're right, this isn't a major problem, just something that jumped out at me when I glanced at the bindata.

cmd/ingress-operator/render.go
Miciah: > Also fixed the perms of the yaml itself

What was wrong with those?  We don't want the executable bit (odd bit) on the yaml file, do we?

ironcladlou: Whoops. Should be 0644, right?

ironcladlou: Re-fixed

ironcladlou:
@Miciah tested with https://github.com/openshift/installer/pull/2523, I think this works.

cmd/ingress-operator/render.go
```diff
@@ -0,0 +1,66 @@
...
+		manifests.CustomResourceDefinitionManifest,
+		manifests.NamespaceManifest,
+	}
+
+	if dir == "-" {
+		for _, file := range files {
+			fmt.Println("---")
+			fmt.Printf("# file: %s\n", filepath.Base(file))
+			fmt.Println(manifests.MustAssetString(file))
+			fmt.Println("...")
```
Miciah: I tried the render command and got the following output (showing the first few lines):
```yaml
---
# file: 00-custom-resource-definition.yaml

---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: null
  name: ingresscontrollers.operator.openshift.io
```
On investigation, I realized `manifests/00-custom-resource-definition.yaml` already has a `---` marker.  We need to delete either the print statements for the markers or the `---` marker in the file.

Miciah: Of course if this loop does not emit markers, that means we also need to add a `---` marker to `manifests/00-namespace.yaml` so that they will be separated when the render command prints them.  I think best would be to remove the marker from `manifests/00-custom-resource-definition.yaml` and keep the print statements as they are.

ironcladlou: We've spent way more time on this particular code branch than is reasonable — it's not actually used for anything but debugging. I went ahead and deleted the entire functionality. Can this now merge?

PR #311, opened 2019-10-12
WIP: NE-150: Use observedGeneration in admission control

Miciah:
Update `status.observedGeneration` when processing an ingresscontroller for
admission, and reprocess an ingresscontroller if its `generation` differs
from its `status.observedGeneration` in order to ensure that it is
revalidated if it is mutated.

https://jira.coreos.com/browse/NE-150

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Call `needsReadmission`, and reprocess the ingresscontroller if needed.
(`admit`): Update the ingresscontroller's `status.observedGeneration` field when updating its "Admitted" status condition.
(`needsReadmission`): New function.  Return a Boolean value indicating whether the given ingresscontroller needs to be reprocessed for admission, based on its current generation and last observed generation.
* `pkg/operator/controller/ingress/status.go` (`ingressStatusesEqual`): Compare observed generation.

### DIFF
--- a/manifests/00-custom-resource-definition.yaml
+++ b/manifests/00-custom-resource-definition.yaml
@@ -414,6 +414,10 @@ spec:
               required:
               - type
               type: object
+            observedGeneration:
+              description: observedGeneration is the most recent generation observed.
+              format: int64
+              type: integer
             selector:
               description: selector is a label selector, in string format, for ingress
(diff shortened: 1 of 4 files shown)


Miciah:
Better to incorporate this into #262 because bumping openshift/api to get `status.observedGeneration` will also get the T:LS profile API.
/close

PR #313, opened 2019-10-23
Propagate authentication operator route health errors as ingress degraded condition

mfojtik:
The goal here is to point people seeing the route degraded condition in authentication operator to ingress operator so they can root cause this faster.
This is not supplementing the health checking mechanism that ingress operator should do regardless of this propagation.

/cc @deads2k 
/cc @ironcladlou

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -140,7 +140,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorAvailableCondition(allIngressesAvailable))
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorProgressingCondition(allIngressesAvailable, oldStatus.Versions, co.Status.Versions, r.OperatorReleaseVersion, r.IngressControllerImage))
-	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorDegradedCondition(state.IngressControllers))
+	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorIngressDegradedCondition(state.IngressControllers))
+
+	authStateObj := &operatorv1.Authentication{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, authStateObj); err != nil {
+		log.Error(err, "unable to get authentication cluster operator")
+	} else {
+		co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorAuthenticationDegradedCondition(authStateObj.Status))
+	}
 
 	if !operatorStatusesEqual(*oldStatus, co.Status) {
 		if err := r.client.Status().Update(context.TODO(), co); err != nil {
@@ -247,8 +254,37 @@ func checkAllIngressesAvailable(ingresses []operatorv1.IngressController) bool {
 	return len(ingresses) != 0
 }
 
-// computeOperatorDegradedCondition computes the operator's current Degraded status state.
-func computeOperatorDegradedCondition(ingresses []operatorv1.IngressController) configv1.ClusterOperatorStatusCondition {
+// computeOperatorAuthenticationDegradedCondition checks the status of authentication operator and propagate the route health degraded
+// condition as AuthenticationRouteDegraded condition for this operator.
+func computeOperatorAuthenticationDegradedCondition(authStatus operatorv1.AuthenticationStatus) configv1.ClusterOperatorStatusCondition {
+	var degradedAuthentication []string
+	for _, cond := range authStatus.Conditions {
+		if cond.Type == operatorv1.OperatorStatusTypeDegraded && cond.Status == operatorv1.ConditionTrue {
+			switch cond.Reason {
+			case "RouteHealthDegradedError":
+				degradedAuthentication = append(degradedAuthentication, cond.Message)
+			}
+		}
+	}
+
+	if len(degradedAuthentication) == 0 {
+		return configv1.ClusterOperatorStatusCondition{
+			Type:   configv1.OperatorDegraded,
+			Status: configv1.ConditionFalse,
+			Reason: "NoAuthenticationRouteDegraded",
+		}
+	}
+
+	return configv1.ClusterOperatorStatusCondition{
+		Type:    configv1.OperatorDegraded,
+		Status:  configv1.ConditionTrue,
+		Reason:  "AuthenticationRouteDegraded",
+		Message: fmt.Sprintf("Authentication route is degraded: %s", strings.Join(degradedAuthentication, ",")),
+	}
+}
+
+// computeOperatorIngressDegradedCondition computes the operator's current Degraded status state.
+func computeOperatorIngressDegradedCondition(ingresses []operatorv1.IngressController) configv1.ClusterOperatorStatusCondition {
 	var degradedIngresses []string
 	for _, ingress := range ingresses {
 		for _, cond := range ingress.Status.Conditions {


ironcladlou:
I'm somewhat ambivalent about the general idea of propagating status from auth in this way (and I will defer to the wider team on that point), but I do ask that we not change our client handling generally, and especially not in the context of this PR. The existing clients available to the controllers should be sufficient without introducing client-go and another client instance, and changing that aspect here seems to introduce unnecessary risk and inconsistency.

ironcladlou:
>The goal here is to point people seeing the route degraded condition in authentication operator to ingress operator so they can root cause this faster.

I don't think this PR does anything towards the stated goal. It doesn't provide us with any new diagnostic information.

ironcladlou:
Well, if the goal is to point the finger to ingress, it succeeds in that, but it doesn't actually provide any new information useful in root cause analysis, which is what would actually accelerate progress.

deads2k:
> Well, if the goal is to point the finger to ingress, it succeeds in that, but it doesn't actually provide any new information useful in root cause analysis, which is what would actually accelerate progress.

It's valuable to be able to direct customers, CEE, insights, PM, management, and developers towards the source of their problems for faster resolution and more accurate allocation of effort toward overall product improvement.  This fix to ingress telemetry will achieve all those things at minimal cost in a way that can be safely backported.

derekwaynecarr:
Is ingress operator always reporting a status?  If the issue was cloud cred operator not minting creds, we know ingress is not reporting status per my chat with Dan yesterday.

pkg/operator/controller/status/controller.go
```diff
@@ -140,7 +140,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorAvailableCondition(allIngressesAvailable))
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorProgressingCondition(allIngressesAvailable, oldStatus.Versions, co.Status.Versions, r.OperatorReleaseVersion, r.IngressControllerImage))
-	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorDegradedCondition(state.IngressControllers))
+	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorIngressDegradedCondition(state.IngressControllers))
+
+	authStateObj := &operatorv1.Authentication{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
```
ironcladlou: Just use an empty struct, the client will marshal into the object and doesn't read the object to construct the get request

pkg/operator/controller/status/controller.go
```diff
@@ -140,7 +140,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "authentication"}, authStateObj); err != nil {
+		log.Error(err, "unable to get authentication cluster operator")
```
ironcladlou: Fail fast/return early on failure

pkg/operator/controller/status/controller.go
```diff
@@ -140,7 +140,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorAvailableCondition(allIngressesAvailable))
 	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorProgressingCondition(allIngressesAvailable, oldStatus.Versions, co.Status.Versions, r.OperatorReleaseVersion, r.IngressControllerImage))
-	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorDegradedCondition(state.IngressControllers))
```
ironcladlou: Why did you delete the call to computeOperatorDegradedCondition?

pkg/operator/controller/status/controller.go
```diff
@@ -140,7 +140,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorIngressDegradedCondition(state.IngressControllers))
+
+	authStateObj := &operatorv1.Authentication{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "authentication"}, authStateObj); err != nil {
+		log.Error(err, "unable to get authentication cluster operator")
+	} else {
+		co.Status.Conditions = mergeConditions(co.Status.Conditions, computeOperatorAuthenticationDegradedCondition(authStateObj.Status))
```
ironcladlou: Doesn't this conflict with the conditions created by computeOperatorDegradedCondition resulting in a last-in-wins scenario?

I guess I would have expected computeOperatorDegradedCondition to be changed to incorporate new criterion (currently just ingress controllers, now ingress controllers and the authentication operator status)

pkg/operator/controller/status/controller.go
mfojtik: it was renamed

pkg/operator/controller/status/controller.go
mfojtik: i would think this is going to be merged to 1 degraded=true with two conditions? at least that is how it works in control plane operator //cc @deads2k

pkg/operator/controller/status/controller.go
mfojtik: my intention was not to affect ingress operator when auth operator object does not exists... iow. do not create unnecessary dependency

pkg/operator/controller/status/controller.go
mfojtik: maybe I don't understand the controller-runtime client, how it will know it should fetch the "cluster" instance of resource?

pkg/operator/controller/status/controller.go
mfojtik: @ironcladlou also the computeOperatorDegradedCondition only return 1 condition, right? it might be we have multiple in this case

pkg/operator/controller/status/controller.go
ironcladlou: I see that now, the diff confused me. Thanks

pkg/operator/controller/status/controller.go
ironcladlou: You address the object to fetch using `types.NamespacedName`:
```go
 r.client.Get(context.TODO(), types.NamespacedName{Namespace: "namespace", Name: "name"}, obj)
```

For a cluster scoped resource, omit the `Namespace` field.

pkg/operator/controller/status/controller.go
ironcladlou: If it's worth logging an error for, should we publish a status that indicates the auth operator is broken in this case? 😁

If it's fine to ignore, why log anything? Maybe just publish a warning event to reduce log spam?

pkg/operator/controller/status/controller.go
ironcladlou: The current merge logic will not aggregate multiple conditions of the same type. What you're saying might be better, but that's not what it does today

pkg/operator/controller/status/controller.go
mfojtik: @ironcladlou i think this is just log error worthy, we don't need event spam when this happen as this is pretty rare, but it is good to have some trace in log if things get weird or this stop working (for debugging)... I can use higher verbosity if you like?

ironcladlou:
Closing due to age — we still need tons of improvement in this area, but it looks like this particular approach won't move forward

/close

PR #314, opened 2019-10-26
Bug 1740374: status: Set Degraded using other status conditions

Miciah:
#### status: Set Degraded using other status conditions

Add a `DeploymentDegraded` status condition for ingresscontrollers.  Use the `Admitted`, `DeploymentDegraded`, `LoadBalancerReady`, and `DNSReady` status conditions to compute the status of the `Degraded` status condition.

* `pkg/api/v1/types.go` (`IngressControllerDeploymentDegradedConditionType`): New constant.
* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Requeue after the duration that `ensureIngressController` returns, if that duration is non-zero.
(`ensureIngressController`): Add a duration return value.  Return the duration value that `syncIngressControllerStatus` returns.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Call `computeDeploymentDegradedCondition`.  Adjust call to `computeIngressDegradedCondition`.  Add a duration return value to indicate if and when the ingresscontroller should be requeued for reconciliation so that the operator will update its status conditions.  Return the duration value from `computeIngressDegradedCondition`.
(`computeDeploymentDegradedCondition`): New function, based on the old definition of `computeIngressDegradedCondition`.  Compute and return the `DeploymentDegraded` status condition.
(`computeIngressDegradedCondition`): Delete the deployment parameter, and add a conditions parameter.  Compute the `Degraded` status condition from the ingresscontroller's other status conditions.  Return the status condition and a duration value to indicate if and when the status conditions should be recomputed.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressDegradedCondition`): New test.

#### status: Compute `DeploymentDegraded` from `Unavailable`

Compute the ingresscontroller's `DeploymentDegraded` status condition from the deployment's desired replicas, maximum unavailable, and current unavailable values.

* `pkg/operator/controller/ingress/status.go` (`computeDeploymentDegradedCondition`): Use current and max unavailable to compute the status condition.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeDeploymentDegradedCondition`): New test.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -9,6 +9,7 @@ import (
 	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+	retryable "github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -185,9 +186,14 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	// The ingresscontroller is safe to process, so ensure it.
 	if err := r.ensureIngressController(ingress, dnsConfig, infraConfig, ingressConfig, apiConfig); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure ingresscontroller: %v", err)
+		switch e := err.(type) {
+		case retryable.Error:
+			log.Error(e, "got retryable error; requeueing", "after", e.After())
+			return reconcile.Result{RequeueAfter: e.After()}, nil
+		default:
+			return reconcile.Result{}, err
+		}
 	}
-
 	return reconcile.Result{}, nil
 }
 
@@ -496,7 +502,10 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 	return nil
 }
 
-// ensureIngressController ensures all necessary router resources exist for a given ingresscontroller.
+// ensureIngressController ensures all necessary router resources exist for a
+// given ingresscontroller.  Any error values are collected into either a
+// retryable.Error value, if any of the error values are retryable, or else an
+// Aggregate error value.
 func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer) error {
 	// Before doing anything at all with the controller, ensure it has a finalizer
 	// so we can clean up later.
@@ -567,7 +576,7 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		errs = append(errs, fmt.Errorf("failed to sync ingresscontroller status: %v", err))
 	}
 
-	return utilerrors.NewAggregate(errs)
+	return retryable.NewMaybeRetryableAggregate(errs)
 }
 
 // IsStatusDomainSet checks whether status.domain of ingress is set.
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -2,13 +2,16 @@ package ingress
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
 	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -31,22 +34,27 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
 	}
 
+	var errs []error
+
 	updated := ic.DeepCopy()
 	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
 	updated.Status.Selector = selector.String()
 	updated.Status.TLSProfile = computeIngressTLSProfile(ic.Status.TLSProfile, deployment)
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(deployment))
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentDegradedCondition(deployment))
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDNSStatus(ic, wildcardRecord, dnsConfig)...)
-	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeIngressDegradedCondition(deployment))
+	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions)
+	errs = append(errs, err)
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, degradedCondition)
 
 	if !ingressStatusesEqual(updated.Status, ic.Status) {
 		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
-			return fmt.Errorf("failed to update ingresscontroller status: %v", err)
+			errs = append(errs, fmt.Errorf("failed to update ingresscontroller status: %v", err))
 		}
 	}
 
-	return nil
+	return retryableerror.NewMaybeRetryableAggregate(errs)
 }
 
 // mergeConditions adds or updates matching conditions, and updates
@@ -123,11 +131,14 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
 	}
 }
 
-func computeIngressDegradedCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+// computeDeploymentDegradedCondition computes the ingresscontroller's
+// "DeploymentDegraded" status condition by examining the status conditions of
+// the deployment.
+func computeDeploymentDegradedCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
 	for _, cond := range deployment.Status.Conditions {
 		if cond.Type == appsv1.DeploymentProgressing && cond.Status == corev1.ConditionFalse && cond.Reason == "ProgressDeadlineExceeded" {
 			return operatorv1.OperatorCondition{
-				Type:    operatorv1.OperatorStatusTypeDegraded,
+				Type:    iov1.IngressControllerDeploymentDegradedConditionType,
 				Status:  operatorv1.ConditionTrue,
 				Reason:  "DeploymentFailed",
 				Message: fmt.Sprintf("The deployment failed (reason: %s) with message: %s", cond.Reason, cond.Message),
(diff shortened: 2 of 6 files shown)


ironcladlou:
Will this resolve https://bugzilla.redhat.com/show_bug.cgi?id=1765282 according to https://bugzilla.redhat.com/show_bug.cgi?id=1765282#c10?

Is Available=False be accounted for here (e.g. when the LB is unready)?

Miciah:
> Will this resolve https://bugzilla.redhat.com/show_bug.cgi?id=1765282 according to https://bugzilla.redhat.com/show_bug.cgi?id=1765282#c10?

The PR solves part of the problem, but another part remains:  The operator only updates status conditions when some event triggers reconciliations.  For example, suppose `LoadBalancerReady` is set to `False`; the operator will not set `Degraded` to `True` until 90 seconds have passed without change to the `LoadBalancerReady` status condition, and unless something triggers reconciliation after that 90 second grace period, the operator will not update the `Degraded` status condition even after that grace period.

> Is Available=False be accounted for here (e.g. when the LB is unready)?

No, this PR in its current state does not change the logic for the `Available` status condition.

Miciah:
Amended to fix status condition predicates logic, improve unit tests, and requeue the ingresscontroller if the `Degraded` status condition is true or is only false by virtue of grace period.

Miciah:
We don't reach `syncIngressControllerStatus` if admission rejects the ingresscontroller; should we call `syncIngressControllerStatus` during admission?

pkg/operator/controller/ingress/status.go
```diff
@@ -123,23 +126,132 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
+	var requeueAfter time.Duration
+	conditionsMap := make(map[string]*operatorv1.OperatorCondition)
+	for i := range conditions {
+		conditionsMap[conditions[i].Type] = &conditions[i]
+	}
+
+	expectedConditions := []struct {
+		condition        string
+		status           operatorv1.ConditionStatus
+		ifConditionsTrue []string
```
ironcladlou: would `predicates` be more succinct?

ironcladlou: eh, maybe not.

pkg/operator/controller/ingress/controller.go
```diff
@@ -497,28 +498,31 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 }
 
 // ensureIngressController ensures all necessary router resources exist for a given ingresscontroller.
-func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer) error {
+// Any error values are collected into an Aggregate error value.  This error value is returned along
+// with a duration value that indicates, if it is non-zero, that the operator should reconcile the
+// ingresscontroller again after that duration has elapsed.
+func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer) (time.Duration, error) {
```
ironcladlou: What about instead using a semantic error pattern? Like returning a RetryableError which embeds the retry context, and on the call site checking the error type

pkg/operator/controller/ingress/status.go
```diff
@@ -123,21 +127,129 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
-				Type:    operatorv1.OperatorStatusTypeDegraded,
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "DeploymentFailed",
-				Message: fmt.Sprintf("The deployment failed (reason: %s) with message: %s", cond.Reason, cond.Message),
+// computeDeploymentDegradedCondition computes the ingresscontroller's
+// "DeploymentDegraded" status condition by examining the status of
+// the deployment.
+func computeDeploymentDegradedCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+	maxUnavailable, _ := intstr.GetValueFromIntOrPercent(deployment.Spec.Strategy.RollingUpdate.MaxUnavailable, int(*(deployment.Spec.Replicas)), false)
+	if int(deployment.Status.UnavailableReplicas) > maxUnavailable {
```
ironcladlou: Is this re-implementing the k8s deployment primitive's condition reporting?

I would have expected a check for this deployment condition: Available=False (MinimumReplicasUnavailable)

pkg/operator/controller/ingress/controller.go
Miciah: If we do that, `ensureIngressController` will need to wrap its aggregate error value with a `RetryableError` with the minimum of the retry period of any `RetryableError` in `errs` (this could be done with some `NewMaybeRetryableAggregate` function), so it might not reduce complexity as much as you are hoping.

pkg/operator/controller/ingress/status.go
Miciah: Yeah, I think we can do that.  (I was inspired by your approach in cluster-dns-operator, but there, you were working with a DemonSet, and here, we have a Deployment, which has an `Available` condition.)

pkg/operator/controller/ingress/controller.go
Miciah: I pushed an update that uses this approach.

pkg/operator/controller/ingress/status.go
Miciah: Done: https://github.com/openshift/cluster-ingress-operator/compare/c6c82534a80ec899646f75c45709f8afd3567104..61f4fccd7b271c95ad91b51341b3c7c08e0a34b4

pkg/operator/controller/ingress/controller.go
```diff
@@ -185,9 +186,12 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	// The ingresscontroller is safe to process, so ensure it.
 	if err := r.ensureIngressController(ingress, dnsConfig, infraConfig, ingressConfig, apiConfig); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure ingresscontroller: %v", err)
+		if retryable, isRetryable := err.(retryable.Error); isRetryable {
+			log.Error(err, "got retryable error; requeueing", "after", retryable.After())
+			return reconcile.Result{RequeueAfter: retryable.After()}, nil
+		}
+		return reconcile.Result{}, err
```
ironcladlou: Bikeshed... go type switch is another style alternative:

```go
switch e := err.(type) {
  case retryable.Error:
    log.Error(e, "got retryable error; requeueing", "after", e.After())
    return reconcile.Result{RequeueAfter: e.After()}, nil
  default:
}
```

pkg/operator/controller/ingress/status.go
```diff
@@ -123,21 +131,143 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
 }
 
-func computeIngressDegradedCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+// computeDeploymentDegradedCondition computes the ingresscontroller's
+// "DeploymentDegraded" status condition by examining the status conditions of
+// the deployment.
+func computeDeploymentDegradedCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
 	for _, cond := range deployment.Status.Conditions {
-		if cond.Type == appsv1.DeploymentProgressing && cond.Status == corev1.ConditionFalse && cond.Reason == "ProgressDeadlineExceeded" {
+		if cond.Type == appsv1.DeploymentAvailable && cond.Status == corev1.ConditionFalse && cond.Reason == "MinimumReplicasUnavailable" {
```
ironcladlou: Wouldn't we be degraded if the deployment were Available=False regardless of reason?

Miciah: That would make sense.  Should we directly reflect the deployment's status, reason, and message then?

pkg/util/retryableerror/retryableerror.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// NewMaybeRetryableAggregate converts a slice of errors, some of which may be
+// RetryableError values, into a RetryableAggregate interface, which is itself
+// an implementation of the error interface.  Nil values will be filtered from
+// the slice.  If the filtered slice is empty, the return value will be nil.  If
+// any values are RetryableError values, the result will be a RetryableError,
+// with RetryAfter equal to the minimum of all the RetryableError values'
+// RetryAfter values.
+func NewMaybeRetryableAggregate(errs []error) error {
```
ironcladlou: Not 100% sure about these semantics... the way controller-runtime works is that if Reconcile() returns an error, the request is unconditionally re-queued with the default rate limiting stuff. Only if a nil err is returned will RetryAfter be honored. So if an aggregate contains a mix of types, we're inverting the usual behavior in response to non-retryable errors... Not sure if it matters, but it seemed weird enough to at least bring up.

pkg/operator/controller/ingress/controller.go
Miciah: That works for me.  I'll update it in the next push.

pkg/operator/controller/ingress/status.go
ironcladlou: Seems like a good start... there's almost no limit to how far the analysis and aggregation can go... mirroring the reason would be way more info than we have now!

pkg/operator/controller/ingress/status.go
ironcladlou: Yeah, DameonSet doesn't give us any conditions to work with ☹️

pkg/operator/controller/ingress/controller.go
ironcladlou: I certainly hadn't considered aggregates... looks like you went for it anyway, so guess we'll figure it out!

pkg/util/retryableerror/retryableerror.go
ironcladlou: On the other hand, this is far more principled an approach than we currently have where literally all errors are equal

Miciah: `NewMaybeRetryableAggregate` could set `retryAfter` to 0 if `errs` contains a non-retryable error.  However, controller-runtime takes `RequeueAfter: 0` to mean "don't requeue"; would it be too weird to have different semantics (where having retryable error type means retry, and if `After()` is 0, then retry immediately)?

Miciah: Maybe it would be less weird just to return a retryable error only if `errs` is all retryable errors.

pkg/operator/controller/ingress/controller.go
Miciah: Done.

pkg/operator/controller/ingress/status.go
Miciah: Done: https://github.com/openshift/cluster-ingress-operator/compare/8b503220967687ec379a5e64d3a443172d2e6b70..a64df5779bcec0aeaf5c4f1e0c4e99a530c05da3

pkg/util/retryableerror/retryableerror.go
Miciah: Done: https://github.com/openshift/cluster-ingress-operator/compare/8b503220967687ec379a5e64d3a443172d2e6b70..a64df5779bcec0aeaf5c4f1e0c4e99a530c05da3

pkg/operator/controller/ingress/status.go
```diff
@@ -123,21 +131,156 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "DeploymentFailed",
-				Message: fmt.Sprintf("The deployment failed (reason: %s) with message: %s", cond.Reason, cond.Message),
+		if cond.Type == appsv1.DeploymentAvailable {
+			switch cond.Status {
+			case corev1.ConditionFalse:
+				return operatorv1.OperatorCondition{
+					Type:    iov1.IngressControllerDeploymentDegradedConditionType,
+					Status:  operatorv1.ConditionTrue,
+					Reason:  "DeploymentAvailableFalse",
```
ironcladlou: DeploymentAvailableFalse → DeploymentUnavailable?

pkg/operator/controller/ingress/status.go
```diff
@@ -123,21 +131,156 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
+					Message: fmt.Sprintf("The deployment has Available status condition set to False (reason: %s) with message: %s", cond.Reason, cond.Message),
+				}
+			case corev1.ConditionTrue:
+				return operatorv1.OperatorCondition{
+					Type:    iov1.IngressControllerDeploymentDegradedConditionType,
+					Status:  operatorv1.ConditionFalse,
+					Reason:  "DeploymentAvailableTrue",
```
ironcladlou: DeploymentAvailableTrue → DeploymentAvailable?

pkg/operator/controller/ingress/status.go
Miciah: Done.

pkg/operator/controller/ingress/status.go
Miciah: Done.  I also changed DeploymentAvailableUnknown → DeploymentAvailabilityUnknown to enhance its prosaic similitude.

ironcladlou:
/retitle WIP: Bug 1740374: status: Set Degraded using other status conditions

ironcladlou:
/lgtm

@danehans @frobware PTAL, would like to get this soaking on master ASAP

pkg/operator/controller/ingress/status_test.go
```diff
@@ -108,6 +111,131 @@ func cond(t string, status operatorv1.ConditionStatus, reason string, lt time.Ti
 	}
 }
 
+func TestComputeIngressDegradedCondition(t *testing.T) {
+	tests := []struct {
+		name           string
+		conditions     []operatorv1.OperatorCondition
+		expectedStatus operatorv1.ConditionStatus
```
frobware: Does this read a little better if it was `degradedStatus`. Might just help clarify what we're asserting.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -108,6 +111,131 @@ func cond(t string, status operatorv1.ConditionStatus, reason string, lt time.Ti
...
+		expectRequeue  bool
+	}{
+		{
+			name: "not admitted",
```
frobware: Add the empty test?

		{
			name:           "no conditions",
			expectedStatus: operatorv1.ConditionFalse,
			expectRequeue:  false,
		},

pkg/operator/controller/ingress/status_test.go
```diff
@@ -108,6 +111,131 @@ func cond(t string, status operatorv1.ConditionStatus, reason string, lt time.Ti
...
+				cond(operatorv1.DNSReadyIngressConditionType, operatorv1.ConditionTrue, "", clock.Now().Add(time.Hour*-1)),
+			},
+			expectedStatus: operatorv1.ConditionFalse,
+			expectRequeue:  false,
+		},
+	}
+
+	for _, test := range tests {
+		actual, err := computeIngressDegradedCondition(test.conditions)
+		if actual.Status != test.expectedStatus {
```
frobware: Do we want to check that `actual` can never be `nil` in the face of a non-`nil` error?

pkg/operator/controller/ingress/status.go
```diff
@@ -123,21 +131,143 @@ func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.
...
+		condition := operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeDegraded,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "DegradedConditions",
+			Message: "One or more other status conditions indicate a degraded state: " + degraded[2:],
+		}
+
+		err := retryableerror.New(errors.New("IngressController is degraded"), requeueAfter)
+
+		return condition, err
```
frobware: Collapse to:

```go
return condition, retryableerror.New(errors.New("IngressController is degraded"), requeueAfter)
```

Miciah: Done.

pkg/operator/controller/ingress/status_test.go
Miciah: I have changed it to `expectIngressDegradedStatus` (and in `TestComputeDeploymentDegradedCondition`, I have changed it to `expectDeploymentDegradedStatus`).

pkg/operator/controller/ingress/status_test.go
Miciah: Done.

pkg/operator/controller/ingress/status_test.go
Miciah: I've added a check for non-retryable errors.

Miciah:
Timeout during provisioning.
/test e2e-aws-operator

ironcladlou:
Feedback looks to be addressed, thanks for the review @frobware!

Let's merge and do followups as necessary.

/lgtm
/retest

Miciah:
The "Pod was active on the node longer than the specified deadline" failure looks like [BZ#1765286](https://bugzilla.redhat.com/show_bug.cgi?id=1765286).

Miciah:
could not copy stable imagestreamtag: Timeout: request did not complete within allowed duration

/retest

Miciah:
could not copy stable imagestreamtag: Timeout: request did not complete within allowed duration
/retest

PR #315, opened 2019-10-28
Bug 1763605: Fixes tls security policy cipher handling

danehans:
Converts `Modern` tls security profiles to `Intermediate`. Remove after haproxy is upgraded to an openssl version that supports tls v1.3.

/assign @Miciah

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -337,9 +337,9 @@ func hasTLSSecurityProfile(ic *operatorv1.IngressController) bool {
 
 // tlsProfileSpecForSecurityProfile returns a TLS profile spec based on the
 // provided security profile, or the "Intermediate" profile if an unknown
-// security profile type is provided.  Note that the return value must not be
-// mutated by the caller; the caller must make a copy if it needs to mutate the
-// value.
+// or "Modern" security profile type is provided.  Note that the return value
+// must not be mutated by the caller; the caller must make a copy if it needs
+// to mutate the value.
 func tlsProfileSpecForSecurityProfile(profile *configv1.TLSSecurityProfile) *configv1.TLSProfileSpec {
 	if profile != nil {
 		if profile.Type == configv1.TLSProfileCustomType {
@@ -348,6 +348,11 @@ func tlsProfileSpecForSecurityProfile(profile *configv1.TLSSecurityProfile) *con
 			}
 			return &configv1.TLSProfileSpec{}
 		} else if spec, ok := configv1.TLSProfiles[profile.Type]; ok {
+			// TODO remove when haproxy is built with an openssl version that supports tls v1.3.
+			if profile.Type == configv1.TLSProfileModernType {
+				log.Info("converted modern tls security profile to intermediate")
+				return configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
+			}
 			return spec
 		}
 	}
@@ -412,7 +417,8 @@ var (
 		configv1.VersionTLS10: {},
 		configv1.VersionTLS11: {},
 		configv1.VersionTLS12: {},
-		configv1.VersionTLS13: {},
+		// TODO: Add VersionTLS13 support after haproxy is built with an openssl
+		//  version that supports tls v1.3.
 	}
 
 	// isValidCipher is a regexp for strings that look like cipher names.
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -23,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -412,7 +417,8 @@ var (
 		configv1.VersionTLS10: {},
 		configv1.VersionTLS11: {},
 		configv1.VersionTLS12: {},
-		configv1.VersionTLS13: {},
+		// TODO: Add VersionTLS13 support after haproxy is built with an openssl
+		//  version that supports tls v1.3.
```
Miciah: If the TLS profile specifies the "Custom" profile type with TLS 1.3 as the minimum TLS version, the ingresscontroller will be rejected.  This could happen, for example, with the default ingresscontroller if the APIServer config specifies such a profile.  We do have a pretty severe warning of the risk of catastrophe from using the "Custom" profile type, so I am not saying that we should not reject an ingresscontroller under these circumstances, but I want to make sure everybody understands and that we have a consensus.  @ironcladlou

ironcladlou: Sounds good to me given the caveats we're documenting for custom, what does @deads2k think?

knobunc:
Is it better to have them convert to intermediate, or to fall-back to TLS 1.2 with the equivalent ciphers to the current modern ones?

pkg/operator/controller/ingress/controller.go
deads2k: I think failing when there is an impossible custom profile is best.  If a custom profile lists tls 1.2 and 1.3, I expect the custom profile to work, but only with serving 1.2 (filtering similar to how we handle others).  If a custom profile only lists 1.3 and it is filtered out, I would expect to see a failure applying the new profile.

The overall change to rejiggle modern into intermediate seems pretty questionable.  We do reserve the right to decide what these values means, but why is it so hard to update the level of openssl?

danehans: > I think failing when there is an impossible custom profile is best.

@Miciah I believe the PR in its current state meets this requirement. A custom profile that specifies `minTLSVersion: VersionTLSv13` is impossible and would not be admitted.

> If a custom profile lists tls 1.2 and 1.3, I expect the custom profile to work, but only with serving 1.2 (filtering similar to how we handle others). If a custom profile only lists 1.3 and it is filtered out, I would expect to see a failure applying the new profile.

@deads2k the `TLSSecurityProfile` API only specifies [MinTLSVersion](https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L160-L166). `MaxTLSVersion` was dropped from the api at the final stages. Since the current haproxy template router does not support TLS v1.3, `VersionTLS13` is impossible.

> why is it so hard to update the level of openssl

@frobware is working on upgrading haproxy, so I'll let him comment on the details.

Miciah: > > I think failing when there is an impossible custom profile is best.
> 
> @Miciah I believe the PR in its current state meets this requirement. A custom profile that specifies `minTLSVersion: VersionTLSv13` is impossible and would not be admitted.

This PR will indeed reject a custom profile that specifies TLS 1.3.  One possible improvement would be also to filter out TLS 1.3 ciphers when validating a custom profile.  `validateTLSSecurityProfile` already rejects a custom profile with an empty ciphers list, so we could add the filtering before the existing emptiness check.

Miciah: > > why is it so hard to update the level of openssl
> 
> @frobware is working on upgrading haproxy, so I'll let him comment on the details.

We don't necessarily need to upgrade HAProxy; it should suffice to compile the current HAProxy version against OpenSSL 1.1.1.  The challenge is that getting OpenSSL 1.1.1 would require building on a RHEL8 base image.

danehans:
TLS 1.3 cipher filtering is causing ingress-operator CPU to run wild.

/hold

danehans:
`e2e-aws` passes when run locally.

/test e2e-aws

pkg/operator/controller/ingress/controller.go
```diff
@@ -348,6 +356,11 @@ func tlsProfileSpecForSecurityProfile(profile *configv1.TLSSecurityProfile) *con
 			}
 			return &configv1.TLSProfileSpec{}
 		} else if spec, ok := configv1.TLSProfiles[profile.Type]; ok {
+			// TODO remove when haproxy is built with an openssl version that supports tls v1.3.
+			if profile.Type == configv1.TLSProfileModernType {
+				log.Info("converted modern tls security profile to intermediate")
```
ironcladlou: This log message doesn't have enough context and isn't actionable, maybe remove it entirely?

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -12,6 +12,10 @@ func TestTLSProfileSpecForSecurityProfile(t *testing.T) {
 	invalidTLSVersion := configv1.TLSProtocolVersion("abc")
 	invalidCiphers := []string{"ECDHE-ECDSA-AES256-GCM-SHA384", "invalid cipher"}
 	validCiphers := []string{"ECDHE-ECDSA-AES256-GCM-SHA384"}
+	tlsVersion13Ciphers := []string{"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256",
```
ironcladlou: Should this be a shared constant?

danehans: Currently `tlsVersion13Ciphers` is only being used by this test. The list of ciphers is represented as a `sets.String` in `/ingress/controller.go` to avoid using a nested `for` loop to filter any tls v1.3 ciphers from a `Custom` profile.

ironcladlou:
This lgtm, but I've feel better if @Miciah or @knobunc gave it a once-over

PR #319, opened 2019-10-29
Refactors IngressController CRD Generation

danehans:
- Removes `ingresscontrollers` type from crd generation in favor of openshift/api crd.
- Updates tooling to sync openshift/api `ingresscontrollers` crd with local copy.
- Updates `00-custom-resource-definition.yaml` based on openshift/api `ingresscontrollers` crd.
- Generates bindata based on updated `00-custom-resource-definition.yaml`.

//assign @ironcladlou @Miciah


hack/sync-ingresscontroller-crd.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+API_CRD="vendor/github.com/openshift/api/operator/v1/0000_50_ingress-operator_00-custom-resource-definition.yaml"
+INGRESS_CRD="manifests/00-custom-resource-definition.yaml"
+COPY_CMD=$(cp $API_CRD $INGRESS_CRD)
```
Miciah: This executes `cp $API_CRD $INGRESS_CRD` and puts its output in `COPY_CMD`.  You probably want a function:
```shell
copy_crd() {
  cp "$API_CRD" "$INGRESS_CRD"
}
```
Then change `$COPY_CMD` to `copy_cmd` below.

hack/sync-ingresscontroller-crd.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if ! stat $INGRESS_CRD >/dev/null; then
+  $COPY_CMD
+elif diff $INGRESS_CRD $API_CRD; then
+  rc=$?
+  if [[ $rc != 0 ]]; then
```
Miciah: The fact that we are in the `elif` branch means that `$?` was non-zero, so `rc` and this `if` are redundant.

Makefile
```diff
@@ -66,7 +70,7 @@ clean:
 	rm -f $(BIN)
 
 .PHONY: verify
-verify: verify-bindata verify-crd verify-gosec
+verify: sync-ingress-crd verify-bindata verify-crd verify-gosec
```
Miciah: Doesn't seem like `make verify` should be modifying anything.

Makefile
```diff
@@ -28,13 +32,13 @@ cluster-build:
 
 # TODO: Add deepcopy generation script/target
 .PHONY: generate
-generate: bindata crd
+generate: sync-ingress-crd bindata crd
 
 .PHONY: bindata
 bindata:
```
Miciah: `bindata` now depends on `sync-ingress-crd`.

Makefile
```diff
@@ -28,13 +32,13 @@ cluster-build:
...
 	hack/update-generated-bindata.sh
 
-# Generate IngressController CRD from vendored API spec.
+# Generate DNSRecord CRD from internal API spec.
 .PHONY: crd
 crd:
 	hack/update-generated-crd.sh
```
ironcladlou: I'm not sure we need any new targets or scripts at all.

What if you add the following line to this `crd` target:

```
cp vendor/github.com/openshift/api/operator/v1/0000_50_ingress-operator_00-custom-resource-definition.yaml manifests/00-custom-resource-definition.yaml
```

And add the following line to the `verify-crd` target:

```
diff vendor/github.com/openshift/api/operator/v1/0000_50_ingress-operator_00-custom-resource-definition.yaml manifests/00-custom-resource-definition.yaml
```

Does that do the job?

hack/update-generated-crd.sh
```diff
@@ -1,26 +1,23 @@
...
 set -x
 GO111MODULE=on GOFLAGS=-mod=vendor go run sigs.k8s.io/controller-tools/cmd/controller-gen crd:trivialVersions=true paths=./pkg/api/v1 output:crd:dir="$OUTDIR"
 set +x
 
 if [[ -z "${SKIP_COPY+1}" ]]; then
-  cp "$OUTDIR/operator.openshift.io_ingresscontrollers.yaml" manifests/00-custom-resource-definition.yaml
   cp "$OUTDIR/ingress.operator.openshift.io_dnsrecords.yaml" manifests/00-custom-resource-definition-internal.yaml
 fi
+
+if ! stat $LOCAL_CRD >/dev/null; then
```
Miciah: Simpler to use `if ! [[ -f "$LOCAL_CRD" ]]`.  Even simpler, you could just run `diff` (or `cmp`) without first checking for the file's existence—`diff` (and `cmp`) will return non-zero if the file doesn't exist.

Makefile
danehans: I refactored with commit ` 71b113b` for the `crd` and `verify-crd` targets to make both internal and vendored api's as suggested by https://github.com/openshift/cluster-ingress-operator/pull/319/files#r341188193.

Miciah: Really, `bindata` probably should depend on `crd`, but we can get that in a follow-up.

PR #320, opened 2019-10-31
Bug 1766141: Ensures LB service finalizer is removed

danehans:
Previously, a service of type `loadBalancer` would still exist after an `ingresscontroller` resource is deleted. Allowing the service to exist after an `ingresscontroller` is deleted can cause it to be reused when the same `ingresscontroller` is recreated.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -64,6 +64,8 @@ func (r *reconciler) ensureRouterDeleted(ci *operatorv1.IngressController) error
 			return err
 		}
 	}
+	log.Info("deleted deployment", "namespace", deployment.Namespace, "name", deployment.Name)
+	r.recorder.Eventf(ci, "Normal", "DeletedDeployment", "Deleted deployment %s/%s", deployment.Namespace, deployment.Name)
 	return nil
 }
 
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -23,6 +23,10 @@ const (
 	//
 	// https://kubernetes.io/docs/concepts/services-networking/service/#proxy-protocol-support-on-aws
 	awsLBProxyProtocolAnnotation = "service.beta.kubernetes.io/aws-load-balancer-proxy-protocol"
+	// lbServiceCleanUpFinalizer is a finalizer attached to any service that has type=LoadBalancer.
+	// Upon deletion of the service, the actual deletion of the resource will be blocked until
+	// this finalizer is removed.
+	lbServiceCleanUpFinalizer = "service.kubernetes.io/load-balancer-cleanup"
 )
 
 var (
@@ -135,10 +139,10 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
 	return service, nil
 }
 
-// finalizeLoadBalancerService removes finalizers from any LB service. This used
-// to be to help with DNS cleanup, but now that's no longer necessary, and so we
-// just need to clear the finalizer which might exist on existing resources.
-//
+// finalizeLoadBalancerService removes the "ingress.openshift.io/operator" finalizer
+// from the load balancer service of ci. This was for helping with DNS cleanup, but
+// that's no longer necessary. We just need to clear the finalizer which might exist
+// on existing resources.
 // TODO: How can we delete this code?
 func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressController) error {
 	service, err := r.currentLoadBalancerService(ci)
@@ -158,4 +162,22 @@ func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressControlle
 		}
 	}
 	return nil
+}
+
+// ensureLoadBalancerCleanupFinalizer ensures the "service.kubernetes.io/load-balancer-cleanup"
+// finalizer does not exist for the load balancer service of ic. For additional background on
+// this finalizer, see:
+// https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190423-service-lb-finalizer.md
+func (r *reconciler) ensureLoadBalancerCleanupFinalizer(ic *operatorv1.IngressController) error {
+	service, err := r.currentLoadBalancerService(ic)
+	if err != nil {
+		return err
+	}
+	if service == nil {
+		return nil
+	}
+	if !slice.ContainsString(service.Finalizers, lbServiceCleanUpFinalizer) {
+		return nil
+	}
+	return fmt.Errorf("finalizer %s exists for service %s/%s", lbServiceCleanUpFinalizer, service.Namespace, service.Name)
 }
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -161,23 +161,11 @@ func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressControlle
 			return fmt.Errorf("failed to remove finalizer from service %s for ingress %s/%s: %v", service.Namespace, service.Name, ci.Name, err)
 		}
 	}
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -159,3 +163,21 @@ func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressControlle
...
+// https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190423-service-lb-finalizer.md
+func (r *reconciler) ensureLoadBalancerCleanupFinalizer(ic *operatorv1.IngressController) error {
+	service, err := r.currentLoadBalancerService(ic)
+	if err != nil {
+		return err
+	}
+	if service == nil {
+		return nil
+	}
+	if slice.ContainsString(service.Finalizers, lbServiceCleanUpFinalizer) {
```
Miciah: Why return nil if the service exists but does not have the finalizer?

danehans: @Miciah I pushed a new commit to address your feedback.

Miciah: I mean, why shouldn't `ensureLoadBalancerCleanupFinalizer` always return an error if the service exists?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -159,3 +163,21 @@ func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressControlle
 	}
 	return nil
 }
+
+// ensureLoadBalancerCleanupFinalizer ensures the "service.kubernetes.io/load-balancer-cleanup"
+// finalizer does not exist for the load balancer service of ic. For additional background on
+// this finalizer, see:
+// https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190423-service-lb-finalizer.md
+func (r *reconciler) ensureLoadBalancerCleanupFinalizer(ic *operatorv1.IngressController) error {
```
ironcladlou: If this logic were moved inline to [finalizeLoadBalancerService()](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/load_balancer_service.go#L151) you could avoid a new function and call to `r.currentLoadBalancerService()` (and I think it makes sense anyway because we're literally delaying finalization of the LB service until some other condition is true)

pkg/operator/controller/ingress/load_balancer_service.go
danehans: I was thinking the func can be used for other purposes in the future where we expect the service to exist but not the finalizer, i.e. bare metal providers.

ironcladlou: Just for clarity, using LB Service publishing is the only time we create a service at all today, and the only new one we've talked about is NodePort.

pkg/operator/controller/ingress/load_balancer_service.go
danehans: @ironcladlou I tried that initially, but then `ensureIngressDeleted()` never proceeds to remove other `ingresscontroller` dependent resources until the cloud infra is cleaned-up. With this approach, the final `ingresscontroller` reconciliation only needs to remove the "ingresscontroller.operator.openshift.io/finalizer-ingresscontroller" finalizer. Which approach do you prefer?

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: If the service exists without a finalizer, why is it better to delete the ingresscontroller immediately rather than waiting till the service has been deleted?

danehans: @Miciah I was not necessarily thinking the func had to be used for a delete operation. I'll update the func based on your initial feedback.

pkg/operator/controller/ingress/load_balancer_service.go
ironcladlou: Hm, is this back to @Miciah's point about failing too soon? To avoid dealing with graphs directly (a path we played with once), the easy thing I've learned to do (from @deads2k) is to make as much progress as possible with parents, aggregate errors, and retry.

Right now we do

```go
finalizeLoadBalancerService() // if fail return early
deleteWildcardDNSRecord() // if fail return early
ensureRouterDeleted() // if fail return early — but finalizeLoadBalancerService will never succeed until this is called!
// finalize ingresscontroller
```

Would this make things better?

```go
finalizeLoadBalancerService() // append errors
deleteWildcardDNSRecord() // append errors
ensureRouterDeleted() // append errors
if errors > 0 return // and retry
// finalize ingresscontroller
```

danehans: @ironcladlou ptal at the latest commit that follows your guidance.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -64,6 +64,7 @@ func (r *reconciler) ensureRouterDeleted(ci *operatorv1.IngressController) error
 			return err
 		}
 	}
+	log.Info("deleted deployment for ingress", "namespace", ci.Namespace, "name", ci.Name)
```
ironcladlou: If it's worth logging at v=0, would an event be better? Not sure I have a consistent set of principles to apply for logging nothing vs. logging v=0, v=1, vs. events in some of these cases. cc @Miciah

ironcladlou:
Just a non-blocking question about logging, lgtm though. Will let @Miciah tag after his review

pkg/operator/controller/ingress/deployment.go
Miciah: Both `log.Info` and emitting an event for "I deleted your router" seem reasonable to me.  The operator's log level is hard-coded to "Debug", which is I guess like `v=6` or higher (because glog and zap have to be contrary for whatever reason).

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: I'm still not getting it.  Maybe it would be useful to take another resource as an example: In the case of the DNSRecord resource, we return before removing finalizer from the IngressController resource if the DNSRecord is still present:  https://github.com/openshift/cluster-ingress-operator/blob/a7f75ef84133e0ae1f8cd71afec2517becfa5099/pkg/operator/controller/ingress/controller.go#L474-L479

This way, we delay the actual deletion of an ingresscontroller as long as a corresponding DNSRecord resource still exists.  This way, the ingresscontroller cannot be deleted (and subsequently recreated) while that old DNSRecord still exists.

It seems to me like we should do the same thing with the Service resource:  Return before removing the finalizer from the IngressController resource until the corresponding service is gone, because we do not want to risk that a new ingresscontroller gets created while the old one's Service resource still exists.  I don't see that it matters what finalizers are on a Service resource—if a service exists, we should keep the finalizer on the IngressController resource so that it cannot be deleted (and subsequently recreated) while the old service still exists.

Is there a situation where we should allow deletion of the ingresscontroller while the service still exists?

ironcladlou: Seems like blockOwnerDeletion=true neatly solves this, if only we had the right "anchor resource" in the openshift-ingress namespace. Then we could constrain all the "waiting" logic to just the anchor object rather than code to wait for every dependent.

Miciah: Our anchor in the `openshift-ingress` namespace is the Deployment resource.

ironcladlou: https://github.com/openshift/cluster-ingress-operator/pull/320/files#r341771387 makes sense to me. Then we don't need to know anything about the upstream finalizer at all.

@danehans I think this would be even simpler...

danehans:
Build error:
`The build "a234567890123456789012345678901234567890123456789012345678-1" status is "Failed"`

`e2e-aws` passes when run locally.

/test e2e-aws

danehans:
https://github.com/openshift/origin/pull/24085 should have fixed the test flake.

/test e2e-aws

PR #321, opened 2019-11-04
Bug 1766851: Add migration for affinity and deployment strategy

Miciah:
When updating a router deployment, check whether the deployment strategy is changing; if it is, update the affinity policy first, and only update the deployment strategy once all pods have the new affinity policy.

This two-step migration is necessary because an old affinity policy may have an anti-affinity rule that prohibits colocation of pods belonging to the same ingresscontroller, the new deployment strategy may require surge, and the presence of pods with the old affinity policy would block a rolling update using the new deployment strategy.

Follow-up to #280.

* `pkg/operator/controller/ingress/deployment.go` (`updateRouterDeployment`): If the deployment strategy is changing, make sure to update the affinity policy first.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -6,6 +6,7 @@ import (
 	"hash"
 	"hash/fnv"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


Miciah:
Provisioning errors.
/test e2e-aws
/test e2e-aws-upgrade

Miciah:
Works in manual testing.

ironcladlou:
Tricky one, thanks!

/lgtm

pkg/operator/controller/ingress/deployment.go
```diff
@@ -646,6 +647,44 @@ func (r *reconciler) updateRouterDeployment(current, desired *appsv1.Deployment)
 		return nil
 	}
 
+	// It may be impossible to transition from the current state to the
+	// desired state if the current state has an affinity policy with an
+	// anti-affinity rule that prohibits colocation of pods belonging to the
```
danehans: Does "anti-affinity rule" cover if the rule is type `nodeAffinity` or `podAffinity`?

Miciah: I think so.  In this case, it is a pod anti-affinity rule.  I can add the word "pod" to the comment.

Miciah: Done.

PR #324, opened 2019-11-07
Bug 1763639: Fixes minTLSVersion for Old profile

danehans:
Sets `minTLSVersion` to 1.1. Note that openshift/api defines `minTLSVersion` as TLS v1.0 for the Old profile, but this version is not supported for ingress. The oldest available TLS version for ingress is 1.1.

/assign @Miciah @ironcladlou


ironcladlou:
Thanks!

/lgtm

PR #328, opened 2019-11-20
Set DNSRecord resources' Failed status conditions to false in absence…

miheer:
Set DNSRecord resources' Failed status conditions to false in absence of failures

If the ingress operator's dns_controller succeeds at publishes a DNS record to a zone, set the Failed status condition for that zone to false.

https://jira.coreos.com/browse/NE-223

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -121,7 +121,18 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if dnsConfig.Spec.PublicZone != nil {
 		zones = append(zones, *dnsConfig.Spec.PublicZone)
 	}
+	statuses, result := r.publishRecordToZones(zones, record)
+	// TODO: only update if status changed
+	updated := record.DeepCopy()
+	updated.Status.Zones = statuses
+	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+		log.Error(err, "failed to update dnsrecord; will retry", "dnsrecord", record)
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+	return result, nil
+}
 
+func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1.DNSRecord) ([]iov1.DNSZoneStatus, reconcile.Result) {
 	// TODO: If we compare desired zones with zones from status (i.e. actual
 	// zones), we can filter to list to those zones which haven't already been
 	// successfully ensured. Then the DNS provider wouldn't need such aggressive
@@ -130,33 +141,28 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	var statuses []iov1.DNSZoneStatus
 	for i := range zones {
 		zone := zones[i]
-		var conditions []iov1.DNSZoneCondition
-		err := r.dnsProvider.Ensure(record, zone)
-		if err != nil {
-			conditions = append(conditions, iov1.DNSZoneCondition{
-				Type:               iov1.DNSRecordFailedConditionType,
-				Status:             string(operatorv1.ConditionTrue),
-				Reason:             "ProviderError",
-				Message:            fmt.Sprintf("The DNS provider failed to ensure the record: %v", err),
-				LastTransitionTime: metav1.Now(),
-			})
+		condition := iov1.DNSZoneCondition{
+			Status:             string(operatorv1.ConditionUnknown),
+			Type:               iov1.DNSRecordFailedConditionType,
+			LastTransitionTime: metav1.Now(),
+		}
+
+		if err := r.dnsProvider.Ensure(record, zone); err != nil {
+			condition.Status = string(operatorv1.ConditionTrue)
+			condition.Reason = "ProviderError"
+			condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
 			result.RequeueAfter = 30 * time.Second
+		} else {
+			condition.Status = string(operatorv1.ConditionFalse)
+			condition.Reason = "ProviderSuccess"
+			condition.Message = "The DNS provider succeeded in ensuring the record"
 		}
 		statuses = append(statuses, iov1.DNSZoneStatus{
 			DNSZone:    zone,
-			Conditions: conditions,
+			Conditions: []iov1.DNSZoneCondition{condition},
 		})
 	}
-
-	// TODO: only update if status changed
-	updated := record.DeepCopy()
-	updated.Status.Zones = statuses
-	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
-		log.Error(err, "failed to update dnsrecord; will retry", "dnsrecord", record)
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-	}
-
-	return result, nil
+	return statuses, result
 }
 
 func (r *reconciler) delete(record *iov1.DNSRecord) error {
--- a/pkg/operator/controller/dns/controller_test.go
+++ b/pkg/operator/controller/dns/controller_test.go
@@ -1,0 +1,62 @@
+package dns
+
+import (
+	"github.com/google/go-cmp/cmp"
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	iov1 "github.com/openshift/api/operatoringress/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/dns"
+	"testing"
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+	var tests = []struct {
+		name   string
+		zones  []configv1.DNSZone
+		expect []string
+	}{
+		{
+			name: "testing for a successful update of 1 record",
+			zones: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			expect: []string{string(operatorv1.ConditionFalse)},
+		},
+		{
+			name:   "when no zones available",
+			zones:  []configv1.DNSZone{},
+			expect: nil,
+		},
+		{
+			name: "when one zone available and one not available",
+			zones: []configv1.DNSZone{
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+				{ID: ""}},
+			expect: []string{string(operatorv1.ConditionFalse), string(operatorv1.ConditionFalse)},
+		},
+	}
+
+	for _, test := range tests {
+		record := &iov1.DNSRecord{
+			Spec: iov1.DNSRecordSpec{
+				DNSName:    "subdomain.dnszone.io.",
+				RecordType: iov1.ARecordType,
+				Targets:    []string{"55.11.22.33"},
+			},
+		}
+		r := &reconciler{
+			//TODO To write a fake provider that can return errors and add more test cases.
+			dnsProvider: &dns.FakeProvider{},
+		}
+		actual, _ := r.publishRecordToZones(test.zones, record)
+		var conditions []string
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
+				conditions = append(conditions, condition.Status)
+			}
+		}
+		if !cmp.Equal(conditions, test.expect) {
+			t.Fatalf("%q: expected:\n%#v\ngot:\n%#v", test.name, test.expect, conditions)
+		}
+	}
+}


miheer:
@Miciah PTAL

pkg/operator/controller/dns/controller.go
```diff
@@ -113,15 +113,27 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 		}
 	}
-
+
```
Miciah: You have some trailing whitespace.

pkg/operator/controller/dns/controller.go
```diff
@@ -141,22 +153,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				LastTransitionTime: metav1.Now(),
 			})
 			result.RequeueAfter = 30 * time.Second
+		} else {
+			conditions = append(conditions, iov1.DNSZoneCondition{
+				Type:               iov1.DNSRecordFailedConditionType,
+				Status:             string(operatorv1.ConditionFalse),
+				Reason:             "ProviderError",
+				Message:            fmt.Sprintf("The DNS provider failed to ensure the record: %v", err),
```
Miciah: Change `Reason` and `Message` to reflect success, or omit them.

pkg/operator/controller/dns/controller.go
```diff
@@ -141,22 +153,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				LastTransitionTime: metav1.Now(),
+			})
+			result.RequeueAfter = 30 * time.Second
```
Miciah: No need to retry on success.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
+	"time"
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+
```
Miciah: Delete the extra blank line.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, test := range tests {
+		r := &reconciler{
+			dnsProvider: &dns.FakeProvider{},
+		}
+		t.Logf("evaluating test %s", test.name)
+		actual, results := r.publishRecordToZones(test.zone, test.record)
+		log.Info("Actual")
+		log.Info(fmt.Sprintf("%#v", actual[0].Conditions[0].Status))
+		log.Info("Result")
+		log.Info(fmt.Sprintf("%#v", results))
```
Miciah: No need for all the logging.  Delete the `t.Logf` and `log.Info` lines, and add `test.name` to the `t.Fatalf` below (prefix `%q: ` to the format string).

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+			zone: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			result: reconcile.Result{RequeueAfter: 10 * time.Second},
+			expect: operatorv1.ConditionFalse,
+		},
+	}
+	for _, test := range tests {
+		r := &reconciler{
+			dnsProvider: &dns.FakeProvider{},
```
Miciah: Maybe add a TODO comment to write a fake provider that can return errors and add more test cases.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+
+	var tests = []struct {
+		name   string
+		record *iov1.DNSRecord
+		zone   []configv1.DNSZone
+		result reconcile.Result
+		expect (operatorv1.ConditionStatus)
```
Miciah: Delete the parentheses.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+
+	var tests = []struct {
+		name   string
+		record *iov1.DNSRecord
+		zone   []configv1.DNSZone
+		result reconcile.Result
```
Miciah: Why is `result` needed?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+		expect (operatorv1.ConditionStatus)
+	}{
+		{
+			name: "check status",
```
Miciah: The name should be more specific.  This is testing for a successful update of 1 record.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+	"time"
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+
+	var tests = []struct {
+		name   string
+		record *iov1.DNSRecord
+		zone   []configv1.DNSZone
```
Miciah: `zone` → `zones`.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,56 @@
...
+					RecordType: iov1.ARecordType,
+					Targets:    []string{"55.11.22.33"},
+				},
+			},
+			zone: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			result: reconcile.Result{RequeueAfter: 10 * time.Second},
+			expect: operatorv1.ConditionFalse,
+		},
```
Miciah: You could add a test case with empty `zones`.  If you do, I suggest changing the type of `expect` to a slice of conditions: `[]operatorv1.ConditionStatus`, and asserting that `r.publishRecordToZones` returns the expected number of conditions with the expected statuses.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{
+			name: "check status",
+			record: &iov1.DNSRecord{
+				Spec: iov1.DNSRecordSpec{
+					DNSName:    "subdomain.dnszone.io.",
+					RecordType: iov1.ARecordType,
+					Targets:    []string{"55.11.22.33"},
+				},
+			},
+			zones:  []configv1.DNSZone{{}},
```
miheer: I tried the test case with no zones but this https://github.com/openshift/cluster-ingress-operator/blob/5b6cf73f3357fd7e2c6681acf1055c940710978d/pkg/operator/controller/dns/controller.go#L148  still gives a nil and the else condition  https://github.com/openshift/cluster-ingress-operator/blob/5b6cf73f3357fd7e2c6681acf1055c940710978d/pkg/operator/controller/dns/controller.go#L159   executes as the 	err := r.dnsProvider.Ensure(record, zone) returns nil . Do we need to define validation for record and zone for the Ensure method ?

miheer:
@Miciah PTAL Added test cases and suggested changes. Thanks in advance!

pkg/operator/controller/dns/controller.go
```diff
@@ -113,15 +113,26 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
 		}
 	}
-
```
Miciah: Please keep this blank line.

pkg/operator/controller/dns/controller.go
```diff
@@ -113,15 +113,26 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
+	statuses, result := r.publishRecordToZones(zones, record)
+	// TODO: only update if status changed
+	updated := record.DeepCopy()
+	updated.Status.Zones = statuses
+	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+		log.Error(err, "failed to update dnsrecord; will retry", "dnsrecord", record)
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
```
Miciah: Just a suggestion: I'd omit the blank line so as to group `return result, nil` with the declaration of `result`.

pkg/operator/controller/dns/controller.go
```diff
@@ -141,22 +152,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-	// TODO: only update if status changed
-	updated := record.DeepCopy()
-	updated.Status.Zones = statuses
-	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
-		log.Error(err, "failed to update dnsrecord; will retry", "dnsrecord", record)
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
-	}
-
-	return result, nil
+	return statuses, result
```
Miciah: What do you think about using `retryableerror.New` in case publishing to one or more of the zones failed?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			expect: operatorv1.ConditionTrue,
+		},
+	}
+	for _, test := range tests {
+		r := &reconciler{
+			dnsProvider: &FakeDNSProvider{},
+		}
+		t.Logf("evaluating test %s", test.name)
```
Miciah: The `t.Logf` statement can be deleted.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+		actual, results := r.publishRecordToZones(test.zones, test.record)
+		_ = results
```
Miciah: What is the purpose of this line?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+					RecordType: iov1.ARecordType,
+					Targets:    []string{"55.11.22.33"},
+				},
+			},
+			zones:  []configv1.DNSZone{{}},
+			expect: operatorv1.ConditionTrue,
+		},
+		{
+			name:   "when no region available",
+			record: nil,
```
Miciah: No need to test nil `record`.  Did you mean to test something different?  What does "region" refer to?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	for _, test := range tests {
+		r := &reconciler{
+			dnsProvider: &FakeDNSProvider{},
+		}
+		t.Logf("evaluating test %s", test.name)
+		actual, results := r.publishRecordToZones(test.zones, test.record)
+		_ = results
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
```
Miciah: Make sure the right number of conditions is returned.  For this unit test, I think it would be acceptable to assume that the number of conditions will equal the number of zones.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+			name: "when no zones available",
+			record: &iov1.DNSRecord{
+				Spec: iov1.DNSRecordSpec{
+					DNSName:    "subdomain.dnszone.io.",
+					RecordType: iov1.ARecordType,
+					Targets:    []string{"55.11.22.33"},
+				},
+			},
+			zones:  []configv1.DNSZone{{}},
+			expect: operatorv1.ConditionTrue,
```
Miciah: Why expect `operatorv1.ConditionTrue`?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+		r := &reconciler{
+			dnsProvider: &FakeDNSProvider{},
+		}
+		t.Logf("evaluating test %s", test.name)
+		actual, results := r.publishRecordToZones(test.zones, test.record)
+		_ = results
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
+				if condition.Status != string(test.expect) {
+					t.Fatalf("expected:\n%#v\ngot:\n%#v", test.expect, condition.Status)
```
Miciah: I think we can replace `test.expect` with hard-coded `operatorv1.ConditionFalse` here for now.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+				}
+			}
+		}
+	}
+}
+
+var _ dns.Provider = &FakeDNSProvider{}
+
+type FakeDNSProvider struct{}
```
Miciah: I meant to use the existing `FakeProvider`.  Will that suffice?

pkg/operator/controller/dns/controller_test.go
miheer: @Miciah actually the function returns two values and one value is results so if we don't use that value the compiler throws an error that results not used

pkg/operator/controller/dns/controller_test.go
miheer: If zones are not available then we shall set the Failed condition status to true right ?

pkg/operator/controller/dns/controller_test.go
miheer: But then I will need condition  in the existing one right [0]   if zone.ID == "" {
		return fmt.Errorf("zone is required")} else {
		return nil
	}  
?
[0] https://github.com/openshift/cluster-ingress-operator/blob/a140e2d5347a4503d263eadf9e0724a87eb73dc6/pkg/dns/dns.go#L22

pkg/operator/controller/dns/controller_test.go
miheer: No sorry I meant record only. I will remove it.

pkg/operator/controller/dns/controller_test.go
miheer: OK I thought it is needed as we can see which tests are running.

pkg/operator/controller/dns/controller.go
miheer: OK

pkg/operator/controller/dns/controller.go
miheer: OK I will remove the blank line

pkg/operator/controller/dns/controller_test.go
Miciah: I see.  Use this instead:
```go
		actual, _ := r.publishRecordToZones(test.zones, test.record)
```

pkg/operator/controller/dns/controller_test.go
Miciah: If `zones` is empty, then `dnsStatus.Conditions` will be empty.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+		{
+			name: "when no zones available",
+			record: &iov1.DNSRecord{
+				Spec: iov1.DNSRecordSpec{
+					DNSName:    "subdomain.dnszone.io.",
+					RecordType: iov1.ARecordType,
+					Targets:    []string{"55.11.22.33"},
+				},
+			},
+			zones:  []configv1.DNSZone{{}},
```
Miciah: This is actually a slice with 1 zone that has all zero (empty) fields.  Use `[]configv1.DNSZone{}`.

pkg/operator/controller/dns/controller_test.go
Miciah: I think this will not be needed once https://github.com/openshift/cluster-ingress-operator/pull/328/files#r351373838 is addressed.

Miciah: Oh, we need this for the "when one zone available and one not available" test case.  Never mind!

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+	operatorv1 "github.com/openshift/api/operator/v1"
+	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
+
+	"testing"
+)
+
+func TestPublishRecordToZones(t *testing.T) {
+	var tests = []struct {
+		name   string
+		record *iov1.DNSRecord
```
Miciah: `record` is always the same; how about omitting it here and defining it in the `for` loop?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+		t.Logf("evaluating test %s", test.name)
+		actual, results := r.publishRecordToZones(test.zones, test.record)
+		_ = results
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
+				conditions = append(conditions, condition.Status)
+			}
+		}
+		for _, test_expect := range test.expect {
+			expect = append(expect, string(test_expect))
```
Miciah: Why do you need `expect`?  Why not use `test.expect` directly below?

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		if strings.Join(conditions, ",") != strings.Join(expect, ",") {
```
Miciah: Use `cmp.Equal`.  See `TestComputeIngressAvailableCondition` in `pkg/operator/controller/ingress/status_test.go` for an example.

pkg/operator/controller/dns/controller_test.go
miheer: @Miciah actually test.expect has type []operatorv1.ConditionStatus  and condition has string data type so to compare I created string arrays for expected and conditions as different data type can't be compared

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: I am now using FakeProvider

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: Sorry I did not get it but now I have tested multiple zones as well and the test cases are working fine. But happy to learn from any feedback given.

miheer:
@Miciah PTAL  Thanks!

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: done

pkg/operator/controller/dns/controller_test.go
Miciah: Could you change the type of `expect` to `[]string` and use `reflect.DeepEqual` for the comparison?

pkg/operator/controller/dns/controller.go
```diff
@@ -141,22 +152,22 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 				LastTransitionTime: metav1.Now(),
 			})
 			result.RequeueAfter = 30 * time.Second
+		} else if err == nil && zone.ID != "" {
```
Miciah: The `if err == nil && zone.ID != ""` can be omitted.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,75 @@
...
+			name: "testing for a successful update of 1 record",
+			zones: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			expect: []operatorv1.ConditionStatus{operatorv1.ConditionFalse},
+		},
+		{
+			name:   "when no zones available",
+			zones:  []configv1.DNSZone{},
+			expect: []operatorv1.ConditionStatus{""},
```
Miciah: I think that should be `[]operatorv1.ConditionStatus{""}` (or `[]string{}` if you change the type).

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,75 @@
...
+	var tests = []struct {
+		name   string
+		zones  []configv1.DNSZone
+		expect []operatorv1.ConditionStatus
+	}{
+		{
+			name: "testing for a successful update of 1 record",
+			zones: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
```
Miciah: The formatting here is strange.  Did you run gofmt?  I would expect the following:
```go
			zones: []configv1.DNSZone{
				{
					ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io",
				},
			},
```

pkg/operator/controller/dns/controller.go
Miciah: If you want to check for an empty zone id in `publishRecordToZones`, you should do so before calling `r.dnsProvider.Ensure`.  You could use something like the following:
```go
		condition := iov1.DNSZoneCondition{
			Status:             string(operatorv1.ConditionUnknown),
			Type:               iov1.DNSRecordFailedConditionType,
			LastTransitionTime: metav1.Now(),
		}
		if len(zone.ID) == 0 {
			condition.Status = string(operatorv1.ConditionTrue)
			condition.Reason = "EmptyZoneId"
			condition.Message = "Zone has an empty id"
		} else if err := r.dnsProvider.Ensure(record, zone); err != nil {
			condition.Status = string(operatorv1.ConditionTrue)
			condition.Reason = "ProviderError"
			condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
			result.RequeueAfter = 30 * time.Second
		} else {
			condition.Status = string(operatorv1.ConditionFalse)
			condition.Reason = "ProviderSuccess"
			condition.Message = fmt.Sprintf("The DNS provider succeeded to ensure the record: %v", err)
		}
		statuses = append(statuses, iov1.DNSZoneStatus{
			DNSZone:    zone,
			Conditions: []iov1.DNSZoneCondition{condition},
		})

````

File: pkg/operator/controller/dns/controller.go [hunk 30]
```diff
@@ -121,7 +121,18 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if dnsConfig.Spec.PublicZone != nil {
 		zones = append(zones, *dnsConfig.Spec.PublicZone)
 	}
+	statuses, result := r.publishRecordToZones(zones, record)
+	// TODO: only update if status changed
+	updated := record.DeepCopy()
+	updated.Status.Zones = statuses
```
Miciah: It occurs to me that we should be using some kind of `mergeConditions` function here, but that is a separate issue, and we can do it in a follow-up PR.

pkg/operator/controller/dns/controller_test.go
miheer: yes not sure why it coming like that

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,75 @@
...
+			expect: []operatorv1.ConditionStatus{operatorv1.ConditionFalse, ""},
+		},
+	}
+
+	var conditions []string
+	var expect []string
+
+	for _, test := range tests {
+		var record *iov1.DNSRecord
+		record = &iov1.DNSRecord{
```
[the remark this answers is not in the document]
miheer: done

pkg/operator/controller/dns/controller.go
miheer: OK

pkg/operator/controller/dns/controller.go
miheer: done

miheer:
@Miciah PTAL added the changes

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+			name: "testing for a successful update of 1 record",
+			zones: []configv1.DNSZone{{
+				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			},
+			expect: []string{string(operatorv1.ConditionFalse)},
+		},
+		{
+			name:   "when no zones available",
+			zones:  []configv1.DNSZone{},
+			expect: []string{""},
```
Miciah:
```suggestion
			expect: nil,
```

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+			dnsProvider: &dns.FakeProvider{},
+		}
+		actual, _ := r.publishRecordToZones(test.zones, record)
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
+				conditions = append(conditions, condition.Status)
+			}
+		}
+		if !cmp.Equal(strings.Join(conditions, ","), strings.Join(test.expect, ",")) {
+			t.Fatalf("expected:\n%#v\ngot:\n%#v", strings.Join(test.expect, ","), strings.Join(conditions, ","))
```
Miciah:
```suggestion
			t.Fatalf("%q: expected:\n%#v\ngot:\n%#v", test.name, test.expect, conditions)
```

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+			//TODO To write a fake provider that can return errors and add more test cases.
+			dnsProvider: &dns.FakeProvider{},
+		}
+		actual, _ := r.publishRecordToZones(test.zones, record)
+		for _, dnsStatus := range actual {
+			for _, condition := range dnsStatus.Conditions {
+				conditions = append(conditions, condition.Status)
+			}
+		}
+		if !cmp.Equal(strings.Join(conditions, ","), strings.Join(test.expect, ",")) {
```
Miciah:
```suggestion
		if !cmp.Equal(conditions, test.expect) {
```
If you make this change without fixing `expect: []string{""}` as I mention in a comment above, you will see that the string comparison is too lax.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+				{ID: ""}},
+			expect: []string{string(operatorv1.ConditionFalse), string(operatorv1.ConditionTrue)},
+		},
+	}
+
+	var conditions []string
+
+	for _, test := range tests {
+		t.Log(test.name)
```
Miciah: Delete the `t.Log` statement and include `test.name` in the `t.Fatalf` format below (see comment below).

pkg/operator/controller/dns/controller_test.go
Miciah: No need to compare lengths if you use `cmp.Equal` as I describe in a comment below.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{
+			name: "when one zone available and one not available",
+			zones: []configv1.DNSZone{
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+				{ID: ""}},
+			expect: []string{string(operatorv1.ConditionFalse), string(operatorv1.ConditionTrue)},
+		},
+	}
+
+	var conditions []string
```
Miciah: Declare `conditions` inside the `for` loop and delete `conditions = nil` inside the loop.

miheer: done

pkg/operator/controller/dns/controller_test.go
miheer: done

miheer:
@Miciah PTAL

pkg/operator/controller/dns/controller_test.go
```diff
@@ -0,0 +1,62 @@
...
+			name: "when one zone available and one not available",
+			zones: []configv1.DNSZone{
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+				{ID: ""}},
+			expect: []string{string(operatorv1.ConditionFalse), string(operatorv1.ConditionTrue)},
+		},
+	}
+
+	for _, test := range tests {
+		var conditions []string
```
Miciah: This can be moved above the `for` loop below that uses it.

miheer: Actually if I do that it will fail like below ->

[miheer@miheer cluster-ingress-operator]$ GO111MODULE=on go test -v -run  TestPublishRecordToZones ./pkg/operator/controller/dns/...
2019-12-21T14:05:41.854+1000	INFO	operator	log/log.go:26	started zapr logger
=== RUN   TestPublishRecordToZones
--- FAIL: TestPublishRecordToZones (0.00s)
    controller_test.go:61: "when no zones available": expected:
        []string(nil)
        got:
        []string{"False"}
FAIL
FAIL	github.com/openshift/cluster-ingress-operator/pkg/operator/controller/dns	0.010s
[miheer@miheer cluster-ingress-operator]$ vi pkg/operator/controller/dns/controller_test.go

Miciah: It works for me.  Were you starting from the correct revision?  This is the change I am suggesting, which I tested locally:
```diff
@@ -37,7 +37,6 @@ func TestPublishRecordToZones(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		var conditions []string
 		record := &iov1.DNSRecord{
 			Spec: iov1.DNSRecordSpec{
 				DNSName:    "subdomain.dnszone.io.",
@@ -50,6 +49,7 @@ func TestPublishRecordToZones(t *testing.T) {
 			dnsProvider: &dns.FakeProvider{},
 		}
 		actual, _ := r.publishRecordToZones(test.zones, record)
+		var conditions []string
 		for _, dnsStatus := range actual {
 			for _, condition := range dnsStatus.Conditions {
 				conditions = append(conditions, condition.Status)
```

miheer:
@Miciah PTAL I did the changes.

ironcladlou:
@Miciah @miheer this is now really stale — are we going to continue?

pkg/operator/controller/dns/controller.go
```diff
@@ -130,33 +141,32 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			condition.Message = "Zone has an empty id"
+		} else if err := r.dnsProvider.Ensure(record, zone); err != nil {
+			condition.Status = string(operatorv1.ConditionTrue)
+			condition.Reason = "ProviderError"
+			condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
 			result.RequeueAfter = 30 * time.Second
+		} else {
+			condition.Status = string(operatorv1.ConditionFalse)
+			condition.Reason = "ProviderSuccess"
+			condition.Message = fmt.Sprintf("The DNS provider succeeded to ensure the record: %v", err)
```
Miciah: `err` is nil here, so omit it from the message.  Also, please change "succeeded to ensure" to "succeeded in ensuring".

Miciah:
I would like to get this finished and merged.

miheer:
@ironcladlou @Miciah I will work on the changes suggested

miheer:
@Miciah PTAL changes done

pkg/operator/controller/dns/controller.go
```diff
@@ -130,33 +141,32 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		condition := iov1.DNSZoneCondition{
+			Status:             string(operatorv1.ConditionUnknown),
+			Type:               iov1.DNSRecordFailedConditionType,
+			LastTransitionTime: metav1.Now(),
+		}
+
+		if len(zone.ID) == 0 {
+			condition.Status = string(operatorv1.ConditionTrue)
+			condition.Reason = "EmptyZoneId"
+			condition.Message = "Zone has an empty id"
```
Miciah, line 153: Actually, it is valid for the zone id to be empty if the zone is specified by tag.  For example, see `getZoneID` in `pkg/dns/aws/dns.go`:
```go
// getZoneID finds the ID of given zoneConfig in Route53. If an ID is already
// known, return that; otherwise, use tags to search for the zone. Returns an
// error if the zone can't be found.
```
And the cluster DNS configuration:
```yaml
  spec:
    baseDomain: ci-op-8xt2f72z-0e76d.origin-ci-int-aws.dev.rhcloud.com
    privateZone:
      tags:
        Name: ci-op-8xt2f72z-0e76d-rvgwz-int
        kubernetes.io/cluster/ci-op-8xt2f72z-0e76d-rvgwz: owned
    publicZone:
      id: Z2GYOLTZHS5VK
```
CI is failing because the operator is not publishing records for the private zone, which is configured with tags, not the zone id:
```
status:
  zones:
  - conditions:
    - lastTransitionTime: "2020-04-01T03:44:21Z"
      message: Zone has an empty id
      reason: EmptyZoneId
      status: "True"
      type: Failed
    dnsZone:
      tags:
        Name: ci-op-8xt2f72z-0e76d-rvgwz-int
        kubernetes.io/cluster/ci-op-8xt2f72z-0e76d-rvgwz: owned
  - conditions:
    - lastTransitionTime: "2020-04-01T03:44:21Z"
      message: The DNS provider succeeded in ensuring the record
      reason: ProviderSuccess
      status: "False"
      type: Failed
    dnsZone:
      id: Z2GYOLTZHS5VK
```
We should delete this check for an empty zone.  (Alternatively, we could report an error if neither id nor tags are specified.)

PR #330, opened 2019-11-20
Updates ensureIngressDeleted() to check for removal of deps

danehans:
Previously, `ensureIngressDeleted()` was not ensuring `ingresscontroller` dependent resources no longer existed. This PR updates `ensureIngressDeleted()` to check for the existence of each `ingresscontroller` dependent resource. If a dependent resource still exists,  `ensureIngressDeleted()` will add an error to the error list. The `ingresscontroller` finalizer is only removed when no errors exist in the error list.

/assign @ironcladlou @Miciah

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -501,31 +501,34 @@ func filterTLS13Ciphers(ciphers []string) []string {
 // the finalizer.
 func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController) error {
 	errs := []error{}
-	if err := r.finalizeLoadBalancerService(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to finalize load balancer service for %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	if svcExists, err := r.finalizeLoadBalancerService(ingress); err != nil {
+		errs = append(errs, fmt.Errorf("failed to finalize load balancer service for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	} else if svcExists {
+		errs = append(errs, fmt.Errorf("load balancer service exists for ingress %s/%s", ingress.Namespace, ingress.Name))
 	}
 
 	// Delete the wildcard DNS record, and block ingresscontroller finalization
 	// until the dnsrecord has been finalized.
 	if err := r.deleteWildcardDNSRecord(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to delete wildcard dnsrecord: %v", err))
+		errs = append(errs, fmt.Errorf("failed to delete wildcard dnsrecord for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 	}
 	if record, err := r.currentWildcardDNSRecord(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to get current wildcard dnsrecord: %v", err))
-	} else {
-		if record != nil {
-			log.V(1).Info("waiting for wildcard dnsrecord to be deleted", "dnsrecord", record)
-			return nil
-		}
+		errs = append(errs, fmt.Errorf("failed to get current wildcard dnsrecord for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	} else if record != nil {
+		errs = append(errs, fmt.Errorf("wildcard dnsrecord exists for ingress %s/%s", ingress.Namespace, ingress.Name))
 	}
 
 	if err := r.ensureRouterDeleted(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to delete deployment for ingress %s: %v", ingress.Name, err))
+		errs = append(errs, fmt.Errorf("failed to delete deployment for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	}
+	if deploy, err := r.currentRouterDeployment(ingress); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -501,31 +501,40 @@ func filterTLS13Ciphers(ciphers []string) []string {
 // the finalizer.
 func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController) error {
 	errs := []error{}
-	if err := r.finalizeLoadBalancerService(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to finalize load balancer service for %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	svc, err := r.finalizeLoadBalancerService(ingress)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to finalize load balancer service for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	}
+	if svc != nil {
```
Miciah: Better to use `else if`.  It should not matter in this case since `finalizeLoadBalancerService` returns a nil service value when it returns a non-nil error value, but generally non-error return values should be ignored in the case of a non-nil error value.

pkg/operator/controller/ingress/controller.go
```diff
@@ -501,31 +501,40 @@ func filterTLS13Ciphers(ciphers []string) []string {
...
-	} else {
-		if record != nil {
-			log.V(1).Info("waiting for wildcard dnsrecord to be deleted", "dnsrecord", record)
-			return nil
-		}
+	record, err := r.currentWildcardDNSRecord(ingress)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to get current wildcard dnsrecord for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	}
+	if record != nil {
```
Miciah: `else if`?

pkg/operator/controller/ingress/controller.go
```diff
@@ -501,31 +501,40 @@ func filterTLS13Ciphers(ciphers []string) []string {
...
 
 	if err := r.ensureRouterDeleted(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to delete deployment for ingress %s: %v", ingress.Name, err))
+		errs = append(errs, fmt.Errorf("failed to delete deployment for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	}
+	deploy, err := r.currentRouterDeployment(ingress)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to get deployment for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	}
+	if deploy != nil {
```
Miciah: `else if`?

pkg/operator/controller/ingress/controller.go
ironcladlou: Shouldn't use this return value if `err` is not nil

pkg/operator/controller/ingress/controller.go
ironcladlou: Shouldn't use this return value if err is not nil

pkg/operator/controller/ingress/controller.go
```diff
@@ -501,31 +501,40 @@ func filterTLS13Ciphers(ciphers []string) []string {
 // the finalizer.
 func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController) error {
 	errs := []error{}
-	if err := r.finalizeLoadBalancerService(ingress); err != nil {
-		errs = append(errs, fmt.Errorf("failed to finalize load balancer service for %s/%s: %v", ingress.Namespace, ingress.Name, err))
+	svc, err := r.finalizeLoadBalancerService(ingress)
```
ironcladlou: Shouldn't read `svc` is `err` is not nil... how about just returning a "service still exist" error and when the service is not found return no error?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -144,28 +140,24 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
 // that's no longer necessary. We just need to clear the finalizer which might exist
 // on existing resources.
 // TODO: How can we delete this code?
-func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressController) error {
+func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressController) (*corev1.Service, error) {
```
Miciah: Instead of a pointer, how about returning a Boolean value to indicate presence of the service?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -144,28 +140,24 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
...
 	service, err := r.currentLoadBalancerService(ci)
 	if err != nil {
```
ironcladlou: if not found, return nil? No reason to return the service

pkg/operator/controller/ingress/controller.go
Miciah: `retryable.New` could be used for such resource-still-exists errors.

danehans:
@ironcladlou @Miciah commit `8f1465d` addresses your initial feedback.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: `currentLoadBalancerService` returns nil if the service does not exist.

Miciah:
/lgtm
but
/hold
in case @ironcladlou (or anyone else) wants to review again.

ironcladlou:
Great, thanks

/hold cancel

PR #331, opened 2019-11-27
publish a router-ca that can be used to verify routes in golang clients

deads2k:
An explanation in code of how to fill in the router-ca.

Immediate mission: allow the default-ingress-ca to always contain a ca-bundle.crt that golang clients can use to validate the ingress wildcard certificate for the default ingresscontroller.  To do this
 1. if you are not `oc get -n openshift-ingress-operator ingresscontroller/default`, do not publish
 2. if the spec.DefaultCertificate.Name is set, use the `.data["tls.crt "]` value in that secret to the `router-ca.data["ca-bundle.crt"]`
 3. if it is not set, use the `ensureRouterCASecret` secret's `.data.["tls.crt"]`

If we can agree to this approach to solve the immediate 4.1, 4.2, and 4.3 problem, then we can update the enhancement.

Future mission: it may be useful to have a ca bundle that can terminate any ingress wildcard.  This seems simple, but it has security repercussions that would effectively be elevating privilege for every ingress-controller.  We have to decide if this ingress-controllers are system trusted or not.  Those that are not may not update this configmap.  I suspect this is already a security hole.

### DIFF
--- a/pkg/operator/controller/certificate/controller.go
+++ b/pkg/operator/controller/certificate/controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 
@@ -109,10 +110,30 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		}
 	}
 
+	// This section creates the legacy router-ca configmap which only ever contains the operator-managed default signing
+	// cert used for signing the default wildcard serving cert
 	ingresses := &operatorv1.IngressControllerList{}
 	if err := r.cache.List(context.TODO(), ingresses, client.InNamespace(r.operatorNamespace)); err != nil {
 		errs = append(errs, fmt.Errorf("failed to list ingresscontrollers: %v", err))
 	} else if err := r.ensureRouterCAConfigMap(ca, ingresses.Items); err != nil {
+		errs = append(errs, fmt.Errorf("failed to publish router CA: %v", err))
+	}
+
+	// We need to construct the CA bundle that can be used to verify the ingress used to serve the console and the oauth-server.
+	// In an operator maintained cluster, this is always `oc get -n openshift-ingress-operator ingresscontroller/default`, skip the rest and return here.
+	// TODO if network-edge wishes to expand the scope of the CA bundle (and you could legitimately see a need/desire to have one CA that verifies all ingress traffic).
+	// TODO this could be accomplished using union logic similar to the kube-apiserver's join of multiple CAs.
+	if ingress == nil || ingress.Namespace != "openshift-ingress-operator" || ingress.Name != "default" {
+		return result, utilerrors.NewAggregate(errs)
+	}
+
+	wildcardServingCertKeySecret := &corev1.Secret{}
+	if err := r.client.Get(context.TODO(), controller.RouterEffectiveDefaultCertificateSecretName(ingress, "openshift-ingress"), wildcardServingCertKeySecret); err != nil {
+		errs = append(errs, fmt.Errorf("failed to lookup wildcard cert: %v", err))
+		return result, utilerrors.NewAggregate(errs)
+	}
+	caBundle := string(wildcardServingCertKeySecret.Data["tls.crt"])
+	if err := r.ensureDefaultIngressCertConfigMap(caBundle); err != nil {
 		errs = append(errs, fmt.Errorf("failed to publish router CA: %v", err))
 	}
 
--- a/pkg/operator/controller/certificate/publish_ca.go
+++ b/pkg/operator/controller/certificate/publish_ca.go
@@ -10,7 +10,23 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
+
+// ensureDefaultIngressCertConfigMap will create or update the configmap containing the public half of the default ingress wildcard certificate
+func (r *reconciler) ensureDefaultIngressCertConfigMap(caBundle string) error {
+	name := controller.DefaultIngressCertConfigMapName()
+	desired := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Data: map[string]string{
+			"ca-bundle.crt": caBundle,
+		},
+	}
+	return r.ensureConfigMap(name, desired)
+}
 
 // ensureRouterCAConfigMap will create, update, or delete the configmap for the
 // router CA as appropriate.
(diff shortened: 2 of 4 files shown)


pkg/operator/controller/certificate/controller.go
```diff
@@ -103,16 +105,31 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+
+	caBundle := string(controllerMaintainedSigningCertKey.Data["tls.crt"])
+	// if we have a custom default certificate, it need to be the one used to very router certificates
+	if len(ingress.Spec.DefaultCertificate.Name) > 0 {
+		secret := &corev1.Secret{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ingress.Namespace, Name: ingress.Spec.DefaultCertificate.Name}, secret); err != nil {
+			errs = append(errs, fmt.Errorf("failed to get custom router cert: %v", err))
+			return result, utilerrors.NewAggregate(errs)
+		}
+		caBundle = string(controllerMaintainedSigningCertKey.Data["tls.crt"])
```
Miciah: This should be `secret.Data["tls.crt"]`, right?

deads2k: fixed

danehans:
> it may be useful to have a ca bundle that can terminate any ingress wildcard. This seems simple, but it has security repercussions that would effectively be elevating privilege for every ingress-controller.

Why is the default ingress controller more secure than any other ingress controller?

Miciah:
> Why is the default ingress controller more secure than any other ingress controller?

On a typical cluster, it is more sensitive because it has the domain that is used for OAuth.

deads2k:
> > it may be useful to have a ca bundle that can terminate any ingress wildcard. This seems simple, but it has security repercussions that would effectively be elevating privilege for every ingress-controller.
> 
> Why is the default ingress controller more secure than any other ingress controller?

It's created and managed by the platform.  The rest are not.

Miciah:
> We have to decide if this ingress-controllers are system trusted or not.

The cluster ingress operator deployment only manages ingresscontrollers that the cluster administrator creates in the `openshift-ingress-operator` namespace.  If in the future we allow unprivileged users to create ingresscontrollers, it will most likely be using unprivileged deployments of the ingress operator.  So in the future we may want to make the operator capable of running unprivileged, but for now, it is safe to say that the ingress operator and any ingresscontrollers it manages are system-trusted.

deads2k:
I'm planning to update this so that instead of changing the meaning of an existing configmap, we also create one named `default-ingress-ca` to properly represent what we're storing. We've gotten away with not including any custom wildcards into `sa-token-secrets/Data[ca.crt]` and I'm inclined to keep it that way.

Miciah:
Thanks!  I have updated https://github.com/openshift/enhancements/pull/126 to describe the approach that David describes above (https://github.com/openshift/cluster-ingress-operator/pull/331#issuecomment-560565825).

@deads2k, we talked about always publishing the default certificate rather than the CA certificate (even in the case that the operator has generated the default certificate and therefore has the CA certificate); if take that approach, do you want to change the name of the new ConfigMap from `default-ingress-ca` to, say, `default-ingress-cert`?  Let me know, and I'll update https://github.com/openshift/enhancements/pull/126 again accordingly.

test/e2e/operator_test.go
```diff
@@ -278,6 +296,20 @@ func TestUpdateDefaultIngressController(t *testing.T) {
...
+		}
+		if defaultIngressCAConfigmap.Data["ca-bundle.crt"] == originalDefaultIngressCAConfigmap.Data["ca-bundle.crt"] {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to observe update of default ingress CA certificate configmap: %v", err)
+	}
+
```
Miciah: Gratuitous blank line.

pkg/operator/controller/certificate/controller.go
```diff
@@ -109,12 +110,33 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 		errs = append(errs, fmt.Errorf("failed to list ingresscontrollers: %v", err))
 	} else if err := r.ensureRouterCAConfigMap(ca, ingresses.Items); err != nil {
 		errs = append(errs, fmt.Errorf("failed to publish router CA: %v", err))
 	}
 
+	// We need to construct the CA bundle that can be used to verify the ingress used to serve the console and the oauth-server.
+	// In an operator maintained cluster, this is always `oc get -n openshift-ingress-operator ingresscontroller/default`, skip the rest and return here.
+	// TODO if network-edge wishes to expand the scope of the CA bundle (and you could legitimately see a need/desire to have one CA that verifies all ingress traffic).
+	// TODO this could be accomplished using union logic similar to the kube-apiserver's join of multiple CAs.
+	if ingress == nil || ingress.Namespace != "openshift-ingress-operator" || ingress.Name != "default" {
```
Miciah: The nil check should be unnecessary as we define `ingress` to be a pointer to an empty `operatorv1.IngressController` and tell the client to fill it in.  As a more general matter, we have been replacing this sort of nil check by using Boolean values (`haveIngress := false` before the`r.client.Get(..., ..., ingress)` call above and `haveIngress = true` if it succeeds).  We can get that in a follow-up if you prefer.

pkg/operator/controller/certificate/controller.go
```diff
@@ -109,12 +110,33 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		return result, utilerrors.NewAggregate(errs)
+	}
+
+	wildcardServingCertKeySecret := &corev1.Secret{}
+	if err := r.client.Get(context.TODO(), controller.RouterEffectiveDefaultCertificateSecretName(ingress, ingress.Namespace), wildcardServingCertKeySecret); err != nil {
```
Miciah: Wrong namespace: The ingresscontroller is in `openshift-ingress-operator`, but the secret is in `openshift-ingress`.

pkg/operator/controller/certificate/publish_ca.go
```diff
@@ -10,16 +10,38 @@ import (
...
+
 // ensureRouterCAConfigMap will create, update, or delete the configmap for the
 // router CA as appropriate.
 func (r *reconciler) ensureRouterCAConfigMap(secret *corev1.Secret, ingresses []operatorv1.IngressController) error {
 	desired, err := desiredRouterCAConfigMap(secret, ingresses)
 	if err != nil {
 		return err
 	}
-	current, err := r.currentRouterCAConfigMap()
+	return r.ensureConfigMap(desired)
```
Miciah: You need to check whether `desired` is nil.  Failure to do so is causing the operator to panic in `currentConfigMap`, called by `ensureConfigMap`, when `desired` is nil.

Miciah:
@deads2k, I see you changed `currentConfigMap` to return nil if it is given a nil input.  I believe the following is what is happening as a result:

1. The certificate controller calls `ensureRouterCAConfigMap`, which calls `desiredRouterCAConfigMap` to get the desired `router-ca` configmap and passes the result to `ensureConfigMap`.  `ensureConfigMap` calls `currentConfigMap`, determines that `router-ca` does not exist but is desired, and creates `router-ca`.
2. The E2E test configures the default ingresscontroller to use a custom default certificate.
3. The certificate controller calls `ensureRouterCAConfigMap`, which calls `desiredRouterCAConfigMap` to get the desired `router-ca` configmap, which is now nil.  `ensureRouterCAConfigMap` passes the nil value to `ensureConfigMap`, which calls `currentConfigMap`, which now reports that `router-ca` does *not* exist because its input was nil.  Because `currentConfigMap` returned nil, `ensureRouterCAConfigMap` does not delete `router-ca`.
4. The E2E test fails because `router-ca` does not get deleted.

Miciah:
/cherrypick release-4.3
The goal is to start publishing `default-ingress-cert` in 4.3.z, move operators to using `default-ingress-cert` instead of using `router-ca` in 4.3.(z+1), and stop publishing `router-ca` in 4.4.0.

Miciah:
/retitle Bug 1788712: publish a router-ca that can be used to verify routes in golang clients

PR #334, opened 2019-12-04
Bug 1765044: Adds proxy support to ingress operator

danehans:
I would like to get feedback on the implementation before proceeding with providers other than aws.


manifests/02-deployment.yaml
```diff
@@ -76,3 +81,10 @@ spec:
       - name: metrics-tls
         secret:
           secretName: metrics-tls
+      - name: trusted-ca
+        configMap:
+          name: trusted-ca
+          optional: false
```
wking: Do we need this?  I'd expect `false` to be the default, although it's not spelled out [in the docs][1].

[1]: https://search.svc.ci.openshift.org/?name=proxy-4.2&search=waiting%20for%20Kubernetes%20API:%20context%20deadline%20exceeded

pkg/dns/aws/filewatcher.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	certs, err := certsutil.CertsFromPEM(caBundle)
+	if err != nil {
+		return err
+	}
+
+	m.route53.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
+	m.elb.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
+	m.tags.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
```
ironcladlou: Should this be made thread-safe?

pkg/dns/aws/filewatcher.go
```diff
@@ -0,0 +1,76 @@
...
+			log.Error(err, "failed to ensure dns client tls transport")
+		}
+	}, 1*time.Minute, stop)
+
+	errChan := make(chan error)
+	go func() {
+		errChan <- m.fileWatcher.Start(stop)
+	}()
+
+	// Wait for the watcher to exit or an explicit stop.
```
wking: this is a bit strange.  Why launch two goroutines and then have a dummy waiter here?  Can we just make `m.fileWatcher.Start` blocking, or is that not something we control.

pkg/dns/aws/filewatcher.go
```diff
@@ -0,0 +1,76 @@
...
+	m.tags.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
+
+	return nil
+}
+
+// caBundlesEqual compares the watched ca bundle with the current
+// ca bundle of FileWatcher, returning false and the current ca
+// bundle if the two are not equal.
+func (m *Provider) caBundlesEqual() (bool, []byte, error) {
+	watchedCAs, err := ioutil.ReadFile(m.fileWatcher.GetFile())
```
wking: This is expensive.  Can we just compare `Stat` mtimes, like I floated in golang/go#35887?  I tested this locally, bumping the ConfigMap content and seeing the mtime change (although you need `Stat` to walk the symlink to see the real mtime in the linked file).  On the other hand, reading the certs once a minute shouldn't be that bad, even if the reads are expensive ;).

wking: And I guess more fundamentally, if we're going to read the whole file off the disk, it seems like we might as well update the CAs without bothering to check if there were any changes.  It's not obvious to me that a bytewise comparison in memory is all that much cheaper than just parsing out the certs into Go CAs.

pkg/dns/aws/filewatcher.go
Miciah: You cannot mutate the clients.

https://github.com/openshift/cluster-ingress-operator/blob/ead9d30aca702ff0a0baf254bd3ae4fe22a22048/vendor/github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi/service.go#L18-L19
https://github.com/openshift/cluster-ingress-operator/blob/ead9d30aca702ff0a0baf254bd3ae4fe22a22048/vendor/github.com/aws/aws-sdk-go/service/elb/service.go#L18-L19
https://github.com/openshift/cluster-ingress-operator/blob/ead9d30aca702ff0a0baf254bd3ae4fe22a22048/vendor/github.com/aws/aws-sdk-go/service/route53/service.go#L18-L19

pkg/dns/aws/filewatcher.go
ironcladlou: Another way might be to forgo fs notifications entirely and wrap access to the underlying file with a struct that implements a TTL for refresh. So like, calling CertFile.Get() would return cached data until its internal TTL (e.g. 30s, 1min) expires, at which point it will lock/refresh/re-cache/reset TTL. Such an approach wouldn't require any goroutines at all.

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,146 @@
...
+	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+	"gopkg.in/fsnotify.v1"
+)
+
+var log = logf.Logger.WithName("filewatcher")
+
+// FileWatcher watches a file for changes.
+type FileWatcher struct {
+	sync.Mutex
+	file        string
```
Miciah: `file`→`fileName` would be clearer.

pkg/util/certs/certs.go
```diff
@@ -0,0 +1,55 @@
...
+
+// MakeTLSTransport returns an HTTP transport with a TLS
+// configuration containing CA certificates from certs.
+func MakeTLSTransport(certs []*x509.Certificate) *http.Transport {
+	cfg := makeTLSConfig(certs)
+	return &http.Transport{TLSClientConfig: cfg}
+}
+
+// CertsFromPEM parses pemCerts into a list of certificates.
+func CertsFromPEM(pemCerts []byte) ([]*x509.Certificate, error) {
```
wking: It's really unfortunate that we need to reproduce this code instead of being able to use the stdlib's [`CertPool.AppendCertsFromPEM`][1].  I guess AWS doesn't take a CertPool for wherever this is being consumed?  Can we bump Go's [`DefaultTransport`][2] `TLSClientConfig` [`RootCAs`][3] instead of talking to the AWS libraries directly?

[1]: https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM
[2]: https://golang.org/pkg/net/http/#RoundTripper
[3]: https://golang.org/pkg/crypto/tls/#Config

pkg/dns/aws/filewatcher.go
ironcladlou: Performance considerations seem pretty inconsequential in this context — we're talking about reading a secret mount, which I believe is memory mapped? We could unconditionally re-read the contents every 15 seconds or something without any diffing at all and I'm not sure it would have any consequential impact on the platform

pkg/util/certs/certs.go
wking: Oh, actually, you can drop `CertsFromPEM` and use `AppendCertsFromPEM` instead in `ensureDNSClientsTLSTransport` to construct a transport.  Then pass that transport in instead of your current 

```golang
	m.route53.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
	m.elb.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
	m.tags.Client.Config.HTTPClient.Transport = certsutil.MakeTLSTransport(certs)
```

which makes three identical transports.

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,146 @@
...
+	currentData []byte
+	watcher     *fsnotify.Watcher
+}
+
+// New returns a new FileWatcher watching the given file.
+func New(file string) (*FileWatcher, error) {
+	var err error
```
Miciah: Why declare `err` here?

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,146 @@
...
+	cw.watcher, err = fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+
+	return cw, nil
+}
+
+// GetFile fetches the file being watched by FileWatcher.
+func (fw *FileWatcher) GetFile() string {
```
Miciah: `GetFile`→`GetFileName`.

manifests/02-deployment.yaml
danehans: @wking I believe `false` is the default. I thought it would be helpful to be explicit. I'll remove.

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,146 @@
...
+	if !(isWrite(event) || isRemove(event) || isCreate(event)) {
+		return
+	}
+
+	log.Info("watched file change", "event", event)
+
+	if isRemove(event) {
+		if err := fw.watcher.Add(event.Name); err != nil {
+			log.Error(err, "error re-watching file %s", fw.file)
+		}
```
Miciah: Return here?  Or do we want to try to read the file after receiving a remove event?

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,146 @@
...
+
+// ReadFile reads the watched file from disk, parses the file,
+// and updates FileWatcher current data.
+func (fw *FileWatcher) ReadFile() error {
+	data, err := ioutil.ReadFile(fw.file)
+	switch {
+	case err != nil:
+		return fmt.Errorf("failed to read file %s: %v", fw.file, err)
+	case len(data) == 0:
+		return fmt.Errorf("file %s contains no currentData", fw.file)
```
Miciah: "currentData"→"data".

pkg/util/certs/certs.go
ironcladlou: Annoying right? The cloud provider libs are idiosyncratic... Azure won't use the stdlib transport the way you would want either, so we have to wire up everything in a way the lib demands

pkg/dns/aws/filewatcher.go
```diff
@@ -0,0 +1,76 @@
...
+
+// startWatcher starts the FileWatcher and periodically ensures DNS clients
+// are using the current ca bundle until a message is received on the stop
+// or error channels.
+func (m *Provider) startWatcher(stop <-chan struct{}) error {
+	go wait.Until(func() {
+		if err := m.ensureDNSClientsTLSTransport(); err != nil {
+			log.Error(err, "failed to ensure dns client tls transport")
+		}
+	}, 1*time.Minute, stop)
```
Miciah: Is this a second, polling-based watcher built on top of the fsnotify-based watcher?  This is very confusing.

pkg/dns/aws/filewatcher.go
danehans: @wking I used the ingress operator's [start()](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/operator.go#L125-L153) as a reference. The first goroutine is periodically ensuring the ca bundle used by dns clients is up-to-date while the second goroutine  starts the file watcher.

manifests/02-deployment.yaml
```diff
@@ -52,6 +54,9 @@ spec:
           resources:
             requests:
               cpu: 10m
+          volumeMounts:
+            - name: trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
```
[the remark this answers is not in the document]
danehans: I'll remove

pkg/operator/watcher/filewatcher.go
danehans: ack

pkg/operator/watcher/filewatcher.go
danehans: ack

pkg/operator/watcher/filewatcher.go
danehans: ack

pkg/dns/aws/filewatcher.go
danehans: @Miciah the fsnotify-based watcher is continuously watching `fileName` and updating `currentData`.  `startWatcher()` starts the fsnotify-based watcher and starts a goroutine that periodically ensures dns clients are using the latest ca bundle by comparing the fsnotify-based watcher's `currentData` with the data from reading `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`.  Let me know if you have any suggestions to improve. Maybe have `handleEvent()` to update the dns clients?

Miciah: What I find confusing is that you essentially have two watchers: a polling-based watcher and an fsnotify-based watcher, where the former is built atop the latter, and half the watcher logic is in the aws provider package and half is in the watcher package.  Is the polling-based watcher logic going to be repeated in every provider implementation?

I would instead amend the fsnotify-based watcher to watch two channels: the fsnotify events channel and a ticker channel.  Dan did something similar in https://github.com/openshift/router/pull/61/commits/4a4f5ab03fcc31d9260d892863dc85ec0ae6a924 (that commit has some extra logic to distinguish between reloads triggered by fsnotify events and reloads triggered by ticks, but otherwise it does essentially what I am am suggesting).  Although others may disagree, I find the single combined fsnotify- and ticker-based watcher easier to understand and reason about.

danehans: @Miciah commit `4c6eb35` updated the PR and my testing against a live cluster produced expected results. However, the commit did not use a bool chan to update elb, tagging and route53 sessions. Commit `97eb21d` implements the chan approach that you suggest. I'm waiting for my cluster to finish installing to test. I'll start testing this commit tomorrow morning PT. Thanks for the guidance!

pkg/operator/watcher/filewatcher.go
danehans: This is what the watcher sees when the trusted ca bundle configmap is updated (i.e. a new ca cert is added):
`2019-12-09T19:55:31.185Z	INFO	operator.filewatcher	watcher/filewatcher.go:113	watched file change	{"event": "\"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem\": REMOVE"}`

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Watch reads events from the watcher's channel and reacts to changes.
+func (fw *FileWatcher) Watch(reload chan bool) {
+	for {
+		select {
+		case event, ok := <-fw.watcher.Events:
+			// Channel is closed.
+			if !ok {
+				return
```
Miciah: Probably should log that the events channel was closed before returning (ditto for the error channel below).

pkg/dns/aws/dns.go
```diff
@@ -75,44 +88,74 @@ type Config struct {
...
-	region := aws.StringValue(sess.Config.Region)
-	if len(region) > 0 {
-		log.Info("using region from shared config", "region name", region)
-	} else {
-		region = config.Region
-		log.Info("using region from operator config", "region name", region)
-	}
-	if len(region) == 0 {
-		return nil, fmt.Errorf("region is required")
-	}
```
Miciah: I'm not sure it is safe to move this validation from the session setup logic here to `createDNSProvider`.  Here, we validate the region from the session configuration, which may be reading from `~/.aws/config` or may be using the region from the cluster config, but `createDNSProvider` only has the cluster config, so it is not validating the same thing.  @ironcladlou, do I have that right?  It should only matter for local development though, so maybe it isn't worth the trouble to check the session config.

cmd/ingress-operator/start.go
```diff
@@ -134,6 +134,9 @@ func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, d
 
 	switch platformStatus.Type {
 	case configv1.AWSPlatformType:
+		if len(platformStatus.AWS.Region) == 0 {
+			return nil, fmt.Errorf("region is required")
+		}
```
ironcladlou: I wonder if this is backwards-compatible with existing storage (https://github.com/openshift/api/commit/dedfb47b1ff14bbd03727fa5720f63d8dc7bea5b)...

See https://github.com/openshift/cluster-ingress-operator/blob/master/cmd/ingress-operator/start.go#L194.

Is this new check necessary?

ironcladlou: Nevermind, I misread — by the time you get here, the platformStatus has been created by getPlatformStatus and should be normalized/valid no matter the source. I think your validation here is correct.

pkg/dns/aws/filewatcher.go
```diff
@@ -0,0 +1,51 @@
...
+		case <-reloadCh:
+			sess, err := NewProviderSession(m.config, operatorReleaseVersion, m.fileWatcher.GetFileData())
+			if err != nil {
+				return fmt.Errorf("failed to create dns provider session: %v", err)
+			}
+			m.lock.Lock()
+			m.elb = elb.New(sess, aws.NewConfig().WithRegion(m.config.Region))
+			m.route53 = route53.New(sess)
+			m.tags = resourcegroupstaggingapi.New(sess, aws.NewConfig().WithRegion("us-east-1"))
+			m.lock.Unlock()
```
ironcladlou: Is this actually thread-safe absent locks near the readers throughout the rest of the code?

cmd/ingress-operator/start.go
danehans: @ironcladlou it's not a new check. I bubbled [this check](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/dns/aws/dns.go#L101-L103) up from the dns/aws pkg.

pkg/dns/aws/dns.go
danehans: @Miciah see https://github.com/openshift/cluster-ingress-operator/pull/334#discussion_r355739759, the check is bubbled up to the aws provider of  `createDNSProvider()`.

pkg/dns/aws/filewatcher.go
danehans: The `FileWatcher` struct uses a mutex to lock/unlock any mutations to the watched file data. The `handleEvent()` method uses a chan to indicate when the watcher updates the current data of the watched file. The `ensureSessionTransport()` updates the aws dns sessions when a value is received on the chan. `StartWatcher()` runs `ensureSessionTransport()` and filewatcher `Start()`, which blocks until the stop channel is closed and listens for fsnotify events and errors on chans. Let me know if you have any suggestions to improve thread safety.

Miciah: The problem is we have code such as the following:
```go
	outerError := m.tags.GetResourcesPages(&resourcegroupstaggingapi.GetResourcesInput{
		ResourceTypeFilters: []*string{aws.String("route53:hostedzone")},
		TagFilters:          tagFilters,
	}, f)
```
To be thread safe, we now need to take a lock before using `m.tags`:
```go
	m.lock.Lock()
	outerError := m.tags.GetResourcesPages(&resourcegroupstaggingapi.GetResourcesInput{
		ResourceTypeFilters: []*string{aws.String("route53:hostedzone")},
		TagFilters:          tagFilters,
	}, f)
	m.lock.Unlock()
```
And ditto for `m.route53` and `m.elb`.

ironcladlou:
@danehans @Miciah @frobware 

The operator already [supports graceful shutdown](https://github.com/openshift/cluster-ingress-operator/blob/master/cmd/ingress-operator/start.go#L126).

What if the operator were to shut down when the file changes just like when it receives a TERM? The operator would be immediately restarted and refresh its state with the latest cert contents. 

The shutdown approach doesn't require a sidecar container and would work for all current and future providers without the need for new state management code (assuming they're wiring up TLS in when constructing their initial clients, which they all should be anyway.)

danehans:
@ironcladlou is it common for customers to monitor operators? If so, do you think it would trigger an alarm and cause cluster admins to wonder why the operator restarted?

ironcladlou:
@danehans 

>@ironcladlou is it common for customers to monitor operators? If so, do you think it would trigger an alarm and cause cluster admins to wonder why the operator restarted?

Restarting the operator during a cert rotation seems like an infrequent, brief, and reasonable outcome to me... what do others think?

ironcladlou:
A restart in this context should take just a couple of seconds — hopefully there are no time-based alerts which would fire if the operator were offline for such a short duration

Miciah:
> Restarting the operator during a cert rotation seems like an infrequent, brief, and reasonable outcome to me... what do others think?

I agree.  Note that we should only need to restart the container (the pod already has the volume and volume mount), not the whole pod.

pkg/dns/aws/filewatcher.go
danehans: The lock/unlock occurs [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/dns/aws/dns.go#L122-L123) for that tagging api call, [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/dns/aws/dns.go#L266-L267) for the route53 api call, and [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/dns/aws/dns.go#L191-L192) for elb api calls.

I'm getting ready to push a commit that will cause the ingress-operator pod to gracefully restart when the ca bundle changes to simplify the implementation and remove the need to worry about runaway goroutines.

pkg/operator/watcher/filewatcher.go
danehans: done

danehans:
@ironcladlou @Miciah commit `fa9cf20` refactors the implementation to gracefully restart when the file being watched by `FileWatcher` changes.

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// handleEvent filters events and gracefully terminates the operator
+// if a write, remove or create event is received for the watched file.
+func (fw *FileWatcher) handleEvent(event fsnotify.Event) {
+	if !(isWrite(event) || isRemove(event) || isCreate(event)) {
+		return
+	}
+	log.Info("watched file changed", "event", event)
+	os.Exit(1)
```
ironcladlou: Should we exit 0 given it's a graceful shutdown?

ironcladlou: err, instead of calling os.Exit, we should be pushing into the stop channel

danehans: @ironcladlou the stop channel is receive only. I can create a send channel that will trigger a `os.Exit()`.

danehans: @ironcladlou commit `9d5ed0b` addresses this issue, ptal.

ironcladlou: https://github.com/openshift/cluster-ingress-operator/blob/master/cmd/ingress-operator/start.go#L126

For example, instead of:

```go
return op.Start(signals.SetupSignalHandler())
```

We could stop on multiple conditions:

```go
stop := make(chan struct{})

// Stop if we receive a signal
go func() {
  stop <- signals.SetupSignalHandler()
}()

// Stop if the cert file changes
go func() {
  // Making up a new function here that blocks until the file changes
  event <- fileWatcher.Watch(cert)
  // Log the event and stop
  stop <- struct{}{}
}()

return op.Start(stop)
```
```

File: pkg/operator/watcher/filewatcher.go [hunk 18]
```diff
@@ -0,0 +1,111 @@
...
+				log.Info("file watch error channel closed")
+				return
+			}
+			log.Error(err, "file watch error")
+		}
+	}
+}
+
+// ReadFile reads the watched file from disk, parses the file,
+// and updates FileWatcher current data.
```
Miciah: This is no longer accurate: "and updates FileWatcher current data".

pkg/operator/watcher/filewatcher.go
```diff
@@ -0,0 +1,111 @@
...
+
+// New returns a new FileWatcher watching the given file.
+func New(file string) (*FileWatcher, error) {
+	fw := &FileWatcher{
+		fileName: file,
+	}
+
+	// Initial read of file.
+	if err := fw.ReadFile(); err != nil {
+		return nil, err
```
Miciah: Is the intent to make the operator container restart as long as the file is absent or empty?

danehans:
From build log:
`error: error pulling image registry.svc.ci.openshift.org/ci/clonerefs@sha256:09762bc02e1aeb2cb2f20faa169194b6096cc85e7cf43fbfcd39b7d3958cbae3: Get https://registry.svc.ci.openshift.org/v2/ci/clonerefs/manifests/sha256:09762bc02e1aeb2cb2f20faa169194b6096cc85e7cf43fbfcd39b7d3958cbae3: dial tcp 35.196.103.194:443: i/o timeout`

/retest

cmd/ingress-operator/start.go
```diff
@@ -118,12 +124,52 @@ func start() error {
...
+		changed := make(chan bool)
+		go func() {
+			orig, err := ioutil.ReadFile(defaultTrustedCABundle)
+			if err != nil {
+				log.Error(err, "failed to read watched file", "filename", defaultTrustedCABundle)
+				os.Exit(1)
+			}
+			for {
+				<-watcher.Events
```
Miciah: We want to watch `watcher.Errors` too.

cmd/ingress-operator/start.go
```diff
@@ -118,12 +124,52 @@ func start() error {
...
+				latest, _ := ioutil.ReadFile(defaultTrustedCABundle)
```
Miciah: We should check the error value.

cmd/ingress-operator/start.go
```diff
@@ -118,12 +124,52 @@ func start() error {
...
+		if err != nil {
+			log.Error(err, "failed to create watcher")
+			os.Exit(1)
+		}
+		defer watcher.Close()
+		if err := watcher.Add(defaultTrustedCABundle); err != nil {
+			log.Error(err, "failed to add file to watcher", "filename", defaultTrustedCABundle)
+			os.Exit(1)
+		}
+		changed := make(chan bool)
```
Miciah: Any reason not to use `chan struct{}`?

danehans: I'll change

cmd/ingress-operator/start.go
danehans: ack. i'll update the pr.

cmd/ingress-operator/start.go
danehans: err... I missed this. I'll update the pr.

cmd/ingress-operator/start.go
```diff
@@ -118,12 +124,68 @@ func start() error {
 		os.Exit(1)
 	}
 
+	// Set up the channels for the watcher and operator.
+	stop := make(chan struct{})
+	signals := signals.SetupSignalHandler()
```
Miciah: I just realized that this variable declaration shadows the package name.  Maybe better use a distinct name for readability?

cmd/ingress-operator/start.go
```diff
@@ -118,12 +124,68 @@ func start() error {
...
+		os.Exit(1)
+	}
+	log.Info("watching file", "filename", defaultTrustedCABundle)
+	go func() {
+		orig, err := ioutil.ReadFile(defaultTrustedCABundle)
+		if err != nil {
+			log.Error(err, "failed to read watched file", "filename", defaultTrustedCABundle)
+			close(stop)
+			return
+		}
```
Miciah: This `ReadFile` and error check might as well go outside the goroutine (then it can use `os.Exit(1)`).

cmd/ingress-operator/start.go
danehans: I'll rename the channel to `signal`.

cmd/ingress-operator/start.go
danehans: ack

ironcladlou:
@danehans thanks for all the work on this one! The approach you and @Miciah landed on is very easy to follow!

danehans:
@kalexand-rh if we have any docs related to manually managing ingress dns records for proxy environments, the limitation can now be removed.

PR #337, opened 2019-12-10
Bug 1781948: Do not wrap errors from syncIngressControllerStatus

Miciah:
An error value from `syncIngressControllerStatus` may represent a failure to update status, in which case `syncIngressControllerStatus` itself adds a clear error message and wrapping it is therefore unnecessary; or the error value may represent a retryable error, in which case wrapping the error value loses the retryable error type and adds an inaccurate error message.

* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Do not wrap the error value from `syncIngressControllerStatus` when adding it to the list of returned error values.
* `pkg/util/retryableerror/retryableerror_test.go` (`TestRetryableError`): Add a test case with a nil error value, and add a nil error value to the "only retryable errors" test case to verify that nil error values are ignored.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -610,9 +610,7 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		errs = append(errs, fmt.Errorf("failed to list events in namespace %q: %v", "openshift-ingress", err))
 	}
 
-	if err := r.syncIngressControllerStatus(ci, deployment, lbService, operandEvents.Items, wildcardRecord, dnsConfig); err != nil {
-		errs = append(errs, fmt.Errorf("failed to sync ingresscontroller status: %v", err))
-	}
+	errs = append(errs, r.syncIngressControllerStatus(ci, deployment, lbService, operandEvents.Items, wildcardRecord, dnsConfig))
 
 	return retryable.NewMaybeRetryableAggregate(errs)
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -611,7 +611,11 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 	}
 
 	if err := r.syncIngressControllerStatus(ci, deployment, lbService, operandEvents.Items, wildcardRecord, dnsConfig); err != nil {
-		errs = append(errs, fmt.Errorf("failed to sync ingresscontroller status: %v", err))
+		syncErr := fmt.Errorf("failed to sync ingresscontroller status: %v", err)
```
ironcladlou: Would using error wrapping remove the need to put this sort of code everywhere?

ironcladlou: Because then we could unwrap the errors once on the outside and check to see if any of the errors in the chain are retryable

Miciah: I think so, but I am going to change this PR not to wrap `err` at all because `syncIngressControllerStatus` decorates its errors sufficient, and we don't want to log "failed to sync ingresscontroller status" if the error is a retryable error (in which case there was no failure to update status).

Miciah:
failed to generate steps from config: step "e2e-aws-upgrade" needs a lease but no lease client provided

/retest

PR #339, opened 2019-12-12
Bug 1783007: Ignore spec.replicas in deployment hash

Miciah:
When computing the hash for a deployment for use in the pod template spec's labels and affinity policy, ignore `spec.replicas` and other fields to which updates should not trigger a rolling update of the deployment.

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Use the new `deploymentTemplateHash` function instead of `deploymentHash`.
(`hashableDeployment`): Add a parameter to indicate whether to hash fields in the deployment spec besides the pod template spec.
(`deploymentHash`): Update call to `hashableDeployment`.
(`TestDesiredRouterDeployment`): Use `deploymentTemplateHash`.
(`otherToleration`): Factor out of `TestDeploymentConfigChanged`.
(`deploymentTemplateHash`): New function.  Use `hashableDeployment`.  Return a hash based on the deployment's metadata and pod template spec.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDeploymentHash`): New test.  Verify that the hashes that `deploymentHash` and `deploymentTemplateHash` return change exactly when expected with respect to mutations to a deployment.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -416,7 +416,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	// compute it now, after all the other fields have been computed, and
 	// inject it into the appropriate fields.
 	if needDeploymentHash {
-		hash := deploymentHash(deployment)
+		hash := deploymentTemplateHash(deployment)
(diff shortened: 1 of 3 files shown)


ironcladlou:
Worth an e2e test?

Miciah:
E2E test added.

ironcladlou:
Awesome, thank you

/lgtm

PR #342, opened 2019-12-17
Configure router for IPv6

Miciah:
Configure the router deployment to enable IPv4, IPv6, or both in accordance with the cluster network config.

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Get the cluster network config and pass it to `ensureIngressController`.
(`ensureIngressController`): Add parameter for network config.  Pass the network config to `ensureRouterDeployment`.
* `pkg/operator/controller/ingress/deployment.go` (`ensureRouterDeployment`): Add parameter for network config.  Pass the network config to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Add parameter for network config.  Use the network config to determine whether the cluster network uses IPv4, IPv6, or both, and set the `ROUTER_IP_V4_V6_MODE` environment variable accordingly.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Define a network config for the test input, and verify that `desiredRouterDeployment` sets `ROUTER_IP_V4_V6_MODE` as appropriate.


pkg/operator/controller/ingress/deployment.go
```diff
@@ -408,6 +409,27 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
 	env = append(env, corev1.EnvVar{Name: "SSL_MIN_VERSION", Value: minTLSVersion})
 
+	usingIPv4 := false
+	usingIPv6 := false
+	for _, clusterNetworkEntry := range networkConfig.Status.ClusterNetwork {
+		addr, _, err := net.ParseCIDR(clusterNetworkEntry.CIDR)
+		if err != nil {
+			continue
+		}
+		if addr.To4() == nil {
```
danehans: I think it's a bit easier to understand if the logic is reversed:
```
		if addr.To4() != nil {
			usingIPv4 = true
		} else {
			usingIPv6 = true
		}
```

Miciah: I have reversed it.  Thanks!

ironcladlou:
/lgtm

Thanks!

ironcladlou:
/lgtm

Let's keep iterating on master.

PR #345, opened 2020-01-07
Implement route admission policy API

ironcladlou:
Implement the route admission policy API[1].

[1] https://github.com/openshift/enhancements/blob/master/enhancements/openshift-route-admission-policy.md


pkg/operator/controller/ingress/controller.go
```diff
@@ -179,18 +179,19 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-			switch err := err.(type) {
-			case *admissionRejection:
-				r.recorder.Event(ingress, "Warning", "Rejected", err.Reason)
-				return reconcile.Result{}, nil
-			default:
-				return reconcile.Result{}, fmt.Errorf("failed to admit ingresscontroller: %v", err)
-			}
+	// Admit the ingresscontroller. Don't process until admission succeeds. If admission results
+	// in an update, immediately re-queue to refresh state.
+	err, updated := r.admit(ingress, ingressConfig, infraConfig)
```
ironcladlou: @Miciah 

After an upgrade, existing admitted ingresscontrollers will have a nil `.spec.routeAdmission` and _should_ have a new generation due to the container image update, triggering re-admission. We have an internal invariant that `.status` will always be populated with defaults, which happens during admission. Consequently, if we ever fail to re-admit upgraded ingresscontrollers for any reason and then process the previously admitted state, we run the risk of violating our invariant and can panic.

Rather than bank on the generation changing, I decided to be even more defensive and start unconditionally admitting ingresscontrollers during reconciliation. This is more expensive in terms of CPU (we will hit our caches for list calls and do some in-memory processing), but the most part, the call should no a persistence no-op, and makes me feel more confident we'll always get things into a valid state before processing.

Is this too paranoid?

danehans: I think the CPU expense is worth preventing the possibility of a panic.

ironcladlou: Oops... this actually doesn't work because `admit()` won't mutate the input, so defaults aren't set and visible to the caller unless (as happens today on master) a persisted update occurs, a re-queue happens, and the state is refreshed during the next reconcile...

This might need re-thought...

ironcladlou: Oops again — the problem is that status _should_ have been persisted. I needed to update `ingressStatusesEqual()` to incorporate the route admission status field, causing a diff and ultimately persistence of status value.

ironcladlou: Nevermind, as of https://github.com/openshift/api/pull/565 we won't be putting the field on status anyway.

ironcladlou:
@Miciah @frobware @danehans this is ready for review (but still needs https://github.com/openshift/api/pull/565 before I can remove the hold)

ironcladlou:
https://github.com/openshift/api/pull/565 merged, this is ready for review

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/google/go-cmp v0.3.0
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/kevinburke/go-bindata v3.11.0+incompatible
-	github.com/openshift/api v3.9.1-0.20191028134408-7e36eed0d19e+incompatible
+	github.com/openshift/api v0.0.0-20200114154634-6f8de8790817
```
frobware: Curious as to why the v-part of the version changed so much? no tag?

ironcladlou: Yeah, no tags

test/e2e/operator_test.go
```diff
@@ -700,6 +701,120 @@ func TestTLSSecurityProfile(t *testing.T) {
...
+	conditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
+		{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+		{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+	}
+	if err := waitForIngressControllerCondition(kclient, 5*time.Minute, icName, conditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Set up a pair of namespaces to in which to create routes
```
frobware: nit: s/to in which/in which/

ironcladlou: Fixed

pkg/operator/controller/ingress/deployment.go
```diff
@@ -430,6 +430,19 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
 
+	routeAdmission := operatorv1.RouteAdmissionPolicy{
+		NamespaceOwnership: operatorv1.StrictNamespaceOwnershipCheck,
+	}
+	if admission := ci.Spec.RouteAdmission; admission != nil && len(admission.NamespaceOwnership) > 0 {
+		routeAdmission.NamespaceOwnership = ci.Spec.RouteAdmission.NamespaceOwnership
+	}
+	switch routeAdmission.NamespaceOwnership {
+	case operatorv1.StrictNamespaceOwnershipCheck:
+		env = append(env, corev1.EnvVar{Name: "ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK", Value: "false"})
```
Miciah: Alternatively, because the default is false, we could omit the environment variable and thereby avoid triggering a rolling update on old clusters.  That's what I did with `ROUTER_IP_V4_V6_MODE`:  https://github.com/openshift/cluster-ingress-operator/blob/a11c8f1d9a7e4cc2334e959700c1257015f5cf23/pkg/operator/controller/ingress/deployment.go#L425-L431
On the other hand, we will probably get a rolling update anyway because we are updating the router image, so maybe it does not matter anyway.

test/e2e/operator_test.go
```diff
@@ -700,6 +701,120 @@ func TestTLSSecurityProfile(t *testing.T) {
 	}
 }
 
+func TestRouteAdmissionPolicy(t *testing.T) {
+	// Set up an ingresscontroller which only selects routes created by this test
```
Miciah: I would have expected this sort of test to be  added in openshift/router (and I'm surprised we don't already have something in openshift/origin) and for the operator's test only to verify that the deployment has the correct configuration, but I suppose adding the test to cluster-ingress-operator is fine (and it is a very fine test indeed!).

pkg/operator/client/client.go
```diff
@@ -38,6 +39,9 @@ func init() {
 	if err := iov1.AddToScheme(scheme); err != nil {
 		panic(err)
 	}
+	if err := routev1.Install(scheme); err != nil {
```
Miciah: Kind of weird that we need to import the route API in `pkg/operator` in order to use it in the test code.  Because the scheme is registered at run-time (even if it is never used), it gets linked into the operator binary, right?

test/e2e/operator_test.go
ironcladlou: Hmmm... good point — I can move the e2e to openshift/origin and replace this one with a unit test that verifies the router config. Maybe I could do next in a followup?

pkg/operator/client/client.go
ironcladlou: I could probably have done it in a test file, but given https://github.com/openshift/cluster-ingress-operator/pull/345#discussion_r366550997 I'll probably just end up deleting this in the e2e-relocation followup if that's the direction we go.

pkg/operator/controller/ingress/deployment.go
ironcladlou: I had the same thoughts, can't decide if adding code to avoid the rollout is worth it.

Miciah: s/adding/omitting/

I'm suggesting we modify `env` only in the `operatorv1.InterNamespaceAllowedOwnershipCheck` case.

ironcladlou: Thought about this some more.

1. As we already noted, if you're running this new operator code, we'll rollout out a new router deployment anyway for the new image version, so that part seems like a wash
2. I think it's safer generally to explicitly set the variable in all cases to true or false to avoid relying on defaulting in the router in the Strict case

So I would suggest what I have here is good. WDYT?

Miciah: Agreed, explicit is good, and we anticipate a new router image, so might as well always add it.

PR #347, opened 2020-01-08
cmd/start: add option to configure shutdown trigger file

ironcladlou:
Add an option to configure the trusted CA bundle location, as the hard-coded
default location is hostile to local development environments.

Update the local run script to allow passing through options to the start
command to make it easier to use the new option.

### DIFF
--- a/cmd/ingress-operator/start.go
+++ b/cmd/ingress-operator/start.go
@@ -38,26 +38,39 @@ const (
 	// operator's namespace that will hold the credentials that the operator
 	// will use to authenticate with the cloud API.
 	cloudCredentialsSecretName = "cloud-credentials"
+
 	// defaultTrustedCABundle is the fully qualified path of the trusted CA bundle
 	// that is mounted from configmap openshift-ingress-operator/trusted-ca.
 	defaultTrustedCABundle = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
 )
 
+type StartOptions struct {
+	// When this file changes, the operator will shut down. This is useful for simple
+	// reloading when things like a certificate changes.
+	ShutdownFile string
+}
+
 func NewStartCommand() *cobra.Command {
-	return &cobra.Command{
+	var options StartOptions
+
+	cmd := &cobra.Command{
 		Use:   "start",
 		Short: "Start the operator",
 		Long:  `starts launches the operator in the foreground.`,
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := start(); err != nil {
+			if err := start(&options); err != nil {
 				log.Error(err, "error starting")
 				os.Exit(1)
 			}
 		},
 	}
+
+	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "file which when modified should trigger a shutdown. If empty, shutdown triggering is disabled.")
+
+	return cmd
 }
 
-func start() error {
+func start(opts *StartOptions) error {
 	metrics.DefaultBindAddress = ":60000"
 
 	// Get a kube client.
@@ -135,15 +148,19 @@ func start() error {
 		os.Exit(1)
 	}
 	defer watcher.Close()
-	if err := watcher.Add(defaultTrustedCABundle); err != nil {
-		log.Error(err, "failed to add file to watcher", "filename", defaultTrustedCABundle)
-		os.Exit(1)
-	}
-	log.Info("watching file", "filename", defaultTrustedCABundle)
-	orig, err := ioutil.ReadFile(defaultTrustedCABundle)
-	if err != nil {
-		log.Error(err, "failed to read watched file", "filename", defaultTrustedCABundle)
-		os.Exit(1)
+
+	var orig []byte
+	if len(opts.ShutdownFile) > 0 {
+		if err := watcher.Add(opts.ShutdownFile); err != nil {
+			log.Error(err, "failed to add file to watcher", "filename", opts.ShutdownFile)
+			os.Exit(1)
+		}
+		log.Info("watching file", "filename", opts.ShutdownFile)
+		orig, err = ioutil.ReadFile(opts.ShutdownFile)
+		if err != nil {
+			log.Error(err, "failed to read watched file", "filename", opts.ShutdownFile)
+			os.Exit(1)
+		}
 	}
 	go func() {
 		for {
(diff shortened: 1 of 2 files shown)


Miciah:
Would a Boolean flag to disable the watch suffice?  Alternatively, should specifying an empty path should disable the watch?

ironcladlou:
>Would a Boolean flag to disable the watch suffice? Alternatively, should specifying an empty path should disable the watch?

Boolean would work but a drawback is you're stuck with a hard-coded path when you do want to test.

Empty path works and also makes testing easier, so gets my vote

ironcladlou:
Oops, my latest commit introduces a race to close the channel.

ironcladlou:
Okay, the latest commit handles this cleanly and makes the mechanism more generic. Operator can take a `--shutdown-file` option which defaults to the trusted CA bundle. If empty, no file is added to the watcher.

ironcladlou:
Gosec doesn't like it because the option variable is passed to io read calls. In https://github.com/openshift/cluster-ingress-operator/pull/349 I propose we just remove gosec entirely.

ironcladlou:
@Miciah @danehans @frobware this is ready for review

cmd/ingress-operator/start.go
```diff
@@ -38,26 +38,35 @@ const (
 	// operator's namespace that will hold the credentials that the operator
 	// will use to authenticate with the cloud API.
 	cloudCredentialsSecretName = "cloud-credentials"
-	// defaultTrustedCABundle is the fully qualified path of the trusted CA bundle
-	// that is mounted from configmap openshift-ingress-operator/trusted-ca.
-	defaultTrustedCABundle = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+
+	// defaultShutdownFile is the fully qualified path of the trusted CA bundle
+	//that is mounted from configmap openshift-ingress-operator/trusted-ca.
+	defaultShutdownFile = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
Miciah: I prefer the old name, `defaultTrustedCABundle`, because it better describes the meaning of the value whereas `defaultShutdownFile` describes one way in which it is used.

cmd/ingress-operator/start.go
```diff
@@ -38,26 +38,35 @@ const (
...
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := start(); err != nil {
+			if err := start(options.ShutdownFile); err != nil {
 				log.Error(err, "error starting")
 				os.Exit(1)
 			}
 		},
 	}
+
+	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultShutdownFile, "file which when modified should trigger a shutdown")
```
Miciah: What do you think about making this a slice?
```go
cmd.Flags().StringSliceVarP(&options.ShutdownFiles, "shutdown-file", "s", []string{defaultTrustedCABundle}, "file which when modified should trigger a shutdown")
```
For your use-case, instead of specifying an empty file name, you would specify an empty list of files to watch, which seems semantically cleaner to me, and if for whatever reason you needed to specify multiple files, you could.

cmd/ingress-operator/start.go
```diff
@@ -157,14 +170,14 @@ func start() error {
 					close(stop)
 					return
 				}
-				latest, err := ioutil.ReadFile(defaultTrustedCABundle)
+				latest, err := ioutil.ReadFile(shutdownFile)
```
Miciah: We could use the event to make the logic more general (which would accommodate my suggestion to take a slice of file names):
```go
			case event, ok := <-watcher.Events:
				// ...
				latest, err := ioutil.ReadFile(event.Name)
```
We would also need to make sure to canonicalize path names, and we would need a map to track multiple files' contents:
```go
	orig := map[string][]byte{}
	for _, path := range shutdownFiles {
		realPath, err := filepath.EvalSymlinks(path)
		if err != nil {
			// ...
		}
		if err := watcher.Add(realPath); err != nil {
			// ...
		}
		orig[realPath], err = ioutil.ReadFile(realPath)
		// ...
	}
	// ...
	go func() {
		for {
			select {
				case event, ok := <-watcher.Events:
				// ...
				realPath, err := filepath.EvalSymlinks(event.Name)
				// ...
				if !bytes.Equal(orig[realPath], latest) {
					// ...
```
So it would be a nicer API IMO but a messier implementation; what do you think?

cmd/ingress-operator/start.go
ironcladlou: >I prefer the old name, defaultTrustedCABundle, because it better describes the meaning of the value whereas defaultShutdownFile describes one way in which it is used.

As far as the program is concerned, the file is just some opaque bytes that trigger shutdown, and nothing about the file conveys any other internal meaning, right? Nothing else in the program interprets this file as a certificate...

cmd/ingress-operator/start.go
ironcladlou: I started with that, but then the logic gets even bigger, and we don't have any use case for multiple, so it didn't seem worth it

ironcladlou: Although that was before I switched to always reading from the watch channel and using the option to control whether to add stuff to the watch, so using a list would probably be only nominally more logic. Still not sure it matters though

cmd/ingress-operator/start.go
ironcladlou: It sounds nice but at the end of the day the only reason I'm making any change here at all is to facilitate local development, so I'm starting to get worried I'm gold plating something that doesn't really matter much at all

frobware: > It sounds nice but at the end of the day the only reason I'm making any change here at all is to facilitate local development, so I'm starting to get worried I'm gold plating something that doesn't really matter much at all

If it's for local dev only would an environment variable suffice (and take precedence)?

cmd/ingress-operator/start.go
Miciah: Nothing in our code does, but ultimately some library code is reading in the file at that path.  My point is that the name should reflect the primary meaning of the file, not just its meaning to our watch code.

If nothing else, I argue "the default shutdown file is the trusted CA bundle file" is clearer than "the default shutdown file is the file of opaque bytes at this path".  But I'm fine with merging as-is if you disagree.

ironcladlou: I think I totally misunderstood what you were saying originally — now I get it and agree the old name should come back

cmd/ingress-operator/start.go
ironcladlou: Options are nicer and basically free for us to use

cmd/ingress-operator/start.go
ironcladlou: Fixed

cmd/ingress-operator/start.go
```diff
@@ -38,26 +38,39 @@ const (
 	// operator's namespace that will hold the credentials that the operator
 	// will use to authenticate with the cloud API.
 	cloudCredentialsSecretName = "cloud-credentials"
+
 	// defaultTrustedCABundle is the fully qualified path of the trusted CA bundle
 	// that is mounted from configmap openshift-ingress-operator/trusted-ca.
 	defaultTrustedCABundle = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
 )
 
+type StartOptions struct {
```
ironcladlou: In a followup I'd like to extract all the environment variables into option parsing into this struct

ironcladlou:
@frobware @Miciah PTAL, this does what I wanted and doesn't seem to regress anything

PR #350, opened 2020-01-10
Bug 1789687: the env ROUTER_IP_V4_V6_MODE should uses v6 instead v6_only

lihongan:
fix Bug https://bugzilla.redhat.com/show_bug.cgi?id=1789687.

The router template uses "v6" to match the env. 
```
    {{- else if eq "v6" $router_ip_v4_v6_mode }}
  bind :::{{env "ROUTER_SERVICE_HTTP_PORT" "80"}} v6only
```
see: https://github.com/openshift/router/blob/8af16cfb1e7bb1a0acf89f96c4e18529d591ec4d/images/router/haproxy/conf/haproxy-config.template#L152

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -425,7 +425,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	if usingIPv6 {
 		mode := "v4v6"
 		if !usingIPv4 {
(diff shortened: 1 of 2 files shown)


ironcladlou:
/approve

Thanks!

@Miciah PTAL?

Miciah:
Thanks!  I must have confused the HAProxy parameter with the `haproxy-config.template` parameter.
/lgtm

PR #351, opened 2020-01-10
Convert more operator configuration to CLI options

ironcladlou:
Convert some environment variable config to CLI options for consistency
and because they're more discoverable and easier to document.

Depends on https://github.com/openshift/cluster-ingress-operator/pull/347.


cmd/ingress-operator/start.go
```diff
@@ -65,76 +73,66 @@ func NewStartCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "file which when modified should trigger a shutdown. If empty, shutdown triggering is disabled.")
+	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "if provided, shut down the operator when this file changes")
+	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", ":6000", "metrics endpoint listen address")
```
Miciah: `6000` → `60000`.

cmd/ingress-operator/start.go
```diff
@@ -48,6 +48,14 @@ type StartOptions struct {
...
 	// reloading when things like a certificate changes.
 	ShutdownFile string
+
+	MetricsListenAddr string
+
+	OperatorNamespace string
+
+	IngressControllerImage string
+
+	ReleaseVersion string
```
Miciah: Godoc?

cmd/ingress-operator/start.go
```diff
@@ -65,76 +73,66 @@ func NewStartCommand() *cobra.Command {
...
-	ingressControllerImage := os.Getenv("IMAGE")
-	if len(ingressControllerImage) == 0 {
-		log.Error(fmt.Errorf("missing environment variable"), "'IMAGE' environment variable must be set")
-		os.Exit(1)
-	}
-	releaseVersion := os.Getenv("RELEASE_VERSION")
-	if len(releaseVersion) == 0 {
-		releaseVersion = statuscontroller.UnknownVersionValue
+	if opts.ReleaseVersion == statuscontroller.UnknownVersionValue {
 		log.Info("RELEASE_VERSION environment variable missing", "release version", statuscontroller.UnknownVersionValue)
```
Miciah: Update the log message not to refer to the environment variable.

cmd/ingress-operator/start.go
```diff
@@ -65,76 +73,66 @@ func NewStartCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "file which when modified should trigger a shutdown. If empty, shutdown triggering is disabled.")
+	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "if provided, shut down the operator when this file changes")
+	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", ":6000", "metrics endpoint listen address")
+	cmd.Flags().StringVarP(&options.OperatorNamespace, "namespace", "n", manifests.DefaultOperatorNamespace, "namespace the operator is deployed to")
+	cmd.Flags().StringVarP(&options.IngressControllerImage, "image", "i", "", "image of the ingress controller the operator will manage (required)")
+	cmd.Flags().StringVarP(&options.ReleaseVersion, "release-version", "", statuscontroller.UnknownVersionValue, "the release version the operator should converge to")
```
Miciah: An alternative approach, which we use in the router, is to read the environment variable to get a default value, and use a hard-coded value as the default if the environment variable is not set.

cmd/ingress-operator/start.go
lihongan: What about the metrics listening by container `kube-rbac-proxy` on port 9393? Will it be replaced by this?

Miciah: Kube-rbac-proxy listens on port 9393 and forwards to 127.0.0.1:60000: https://github.com/openshift/cluster-ingress-operator/blob/a11c8f1d9a7e4cc2334e959700c1257015f5cf23/manifests/02-deployment.yaml#L61-L72

lihongan: so it seems if we change the port by CLI options, we also need to update `--upstream` in the deployment yaml file?

Miciah: Right.  Prior this change, the port was hard-coded to 60000:  https://github.com/openshift/cluster-ingress-operator/blob/a11c8f1d9a7e4cc2334e959700c1257015f5cf23/cmd/ingress-operator/start.go#L74

The deployment yaml file references port 60000, so we should use 60000 as the default value for the command-line option (or update the dpeloyment yaml to use the new port number).

ironcladlou: Fixed

cmd/ingress-operator/start.go
ironcladlou: Fixed

cmd/ingress-operator/start.go
ironcladlou: Fixed

cmd/ingress-operator/start.go
ironcladlou: Isn't this already what's effectively happening, since the env var is used for the CLI options in the deployment YAML?

I also went ahead and made this and most of the other options required with defaults.

Miciah: > Isn't this already what's effectively happening, since the env var is used for the CLI options in the deployment YAML?

Sort of, but if you used `os.Getenv` in the code, then the deployment wouldn't need to translate environment variables into command-line parameters, which could be convenient for people writing their own deployment definitions or running the operator directly (if such people exist).  Anyway, it isn't important; just a suggestion.

ironcladlou:
Something's legit broken in my latest patch, I need to fix it.

/lgtm cancel

ironcladlou:
/test all

@Miciah PTAL, clearing out the PR queue

ironcladlou:
@Miciah please re-tag

cmd/ingress-operator/start.go
frobware: Driving this via env variables would seem natural to me too. But, not important for now.

PR #352, opened 2020-01-13
[release-4.3] Bug 1789124: Configure router for IPv6

Miciah:
#### Configure router for IPv6

Configure the router deployment to enable IPv4, IPv6, or both in accordance with the cluster network config.

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Get the cluster network config and pass it to `ensureIngressController`.
(`ensureIngressController`): Add parameter for network config.  Pass the network config to `ensureRouterDeployment`.
* `pkg/operator/controller/ingress/deployment.go` (`ensureRouterDeployment`): Add parameter for network config.  Pass the network config to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Add parameter for network config.  Use the network config to determine whether the cluster network uses IPv4, IPv6, or both, and set the `ROUTER_IP_V4_V6_MODE` environment variable accordingly.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Define a network config for the test input, and verify that `desiredRouterDeployment` sets `ROUTER_IP_V4_V6_MODE` as appropriate.

#### Bug 1789687: the env `ROUTER_IP_V4_V6_MODE` should uses `v6` instead `v6_only`

---

Supersedes #346.


pkg/operator/controller/ingress/deployment.go
```diff
@@ -408,6 +409,27 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		if addr.To4() != nil {
+			usingIPv4 = true
+		} else {
+			usingIPv6 = true
+		}
+	}
+	if usingIPv6 {
+		mode := "v4v6"
+		if !usingIPv4 {
+			mode = "v6_only"
```
frobware: Didn't we do the following somewhere recently:

```go
mode = "v6"
```

frobware: > ```go
> ```go
> mode = "v6"
> ```
> 
> 
>           
> ```

Oh, this is in the 2nd commit.

Miciah: Yeah, I kept it as two commits as it is on master and as openshift-cherrypick-robot would have done (I don't know how to ask openshift-cherrypick-robot to cherry-pick two PRs at once).  I can squash if that would be preferable.

PR #353, opened 2020-01-22
Bug 1793541: Only initialize DNS providers if DNS config is present

ironcladlou:
If no DNS configuration is provided, don't try to set up provider clients.

This prevents an operator crash if DNS configuration is not provided on
platforms where managed DNS is expected to be supported (e.g. AWS).

In a followup, the provider configuration can be refactored into the provider
implementations themselves, so this part of the code won't need to know anything
about the provider setup beyond picking the right implementation. Then, it would
be safe to always use the appropriate provider for the platform and let the
provider surface configuration errors if DNS records are actually created to
exercise the provider.

### DIFF
--- a/cmd/ingress-operator/start.go
+++ b/cmd/ingress-operator/start.go
@@ -207,6 +207,18 @@ func start(opts *StartOptions) error {
 // createDNSManager creates a DNS manager compatible with the given cluster
 // configuration.
 func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (dns.Provider, error) {
+	// If no DNS configuration is provided, don't try to set up provider clients.
+	// TODO: the provider configuration can be refactored into the provider
+	// implementations themselves, so this part of the code won't need to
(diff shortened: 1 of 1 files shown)


cmd/ingress-operator/start.go
```diff
@@ -207,6 +207,18 @@ func start(opts *StartOptions) error {
 // createDNSManager creates a DNS manager compatible with the given cluster
 // configuration.
 func createDNSProvider(cl client.Client, operatorConfig operatorconfig.Config, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (dns.Provider, error) {
+	// If no DNS configuration is provided, don't try to set up provider clients.
+	// TODO: the provider configuration can be refactored into the provider
```
Miciah: It confused me for a moment to see the TODO added in this context.  The TODO really applies to the logic in the `switch` below, and not specifically to this new short-circuiting logic, right?

ironcladlou: well, the short-circuiting logic would go away in the refactor I'm imagining, so maybe the TODO applies to the entire function?

PR #360, opened 2020-02-12
Bug 1770677: upstream the dnsrecord type

ironcladlou:
Depends on https://github.com/openshift/api/pull/584


ironcladlou:
@Miciah @danehans @frobware PTAL

test/e2e/certificate_publisher_test.go
```diff
@@ -9,8 +9,8 @@ import (
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 
-	iov1 "github.com/openshift/cluster-ingress-operator/pkg/api/v1"
-	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
```
Miciah: Maybe import as `operatorcontroller`?  We are inconsistent in how we import this package, and `operatorcontroller` is the least ambiguous of the options in use.

ironcladlou: Agree, but not sure I agree with another round of CI for it...

PR #361, opened 2020-02-17
Add support for AWS China region route53 Service

wanghaoran1988:


pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
+
+	// TODO: This API will only return hostedzone resources (which are global)
+	// when the region is forced to us-east-1. We don't yet understand why.
+	tagRegion := "us-east-1"
+
+	//If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
+	// 1. hard code route53 api endpoint to https://api.route53.cn, as route53 is not GA in AWS China, and aws sdk didn't have the endpoint.
+	// 2. use the aws china regions to setup tagging api correctly in stead of "use-east-1"
+	if (region == "cn-north-1") || (region == "cn-northwest-1") {
+		endpoint := "https://route53.amazonaws.com.cn"
```
bmeng: The endpoint should be `api.roue53.cn`?

pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
+	awsconfig := &aws.Config{}
+	awsconfig.Region = &region
+
+	// TODO: This API will only return hostedzone resources (which are global)
+	// when the region is forced to us-east-1. We don't yet understand why.
+	tagRegion := "us-east-1"
+
+	//If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
+	// 1. hard code route53 api endpoint to https://api.route53.cn, as route53 is not GA in AWS China, and aws sdk didn't have the endpoint.
+	// 2. use the aws china regions to setup tagging api correctly in stead of "use-east-1"
```
bmeng: Typo `use-east-1`

pkg/dns/aws/dns.go
wanghaoran1988: both works, this one makes more sense I think.

wanghaoran1988: I got this from AWS SA:
```
For the China Route53, there are 2 endpoints,  see below

    1  api.route53.cn

    2 route53.amazonaws.com.cn

Based on the info from service team , both 2 endpoints are workable after GA. But the new endpoint(the 2nd ) is recommended.
```

ironcladlou: @abhinavdahiya @derekwaynecarr I seem to remember there being prior discussion about cloud provider API endpoints being pushed into the cluster config to be consumed by operators — any thoughts on whether we should be pursing a solution there vs. hard-coding endpoint and region rules here?

wanghaoran1988: @ironcladlou I assume you are talking about https://issues.redhat.com/browse/CORS-1271 ? I don't think this will land in short time, could we merge this as a temp solution? as we already hard coded to "us-east-1" for now, it make sense to me we can add china regions here? after that jira is done, maybe we can revisit this?

ironcladlou: @wanghaoran1988 sorry for the delayed response — I think merging a temporary fix is fine here. I do have a suggestion for a clarifying refactor of the implementation:

```go
tagConfig := aws.NewConfig().WithRegion("us-east-1")
if (region == "cn-north-1") || (region == "cn-northwest-1") {
  tagConfig.Endpoint = "https://route53.amazonaws.com.cn"
  tagConfig.Region = region
}

return &Provider{
  // ...
  tags: resourcegroupstaggingapi.New(sess, tagConfig),
  // ...
}
```

Advantage is that it's less code and also makes it clear that the config is for the tagging API specifically.

Thank you for this contribution! We really appreciate it. Thanks again for your patience.

pkg/dns/aws/dns.go
ironcladlou: Also add a space after the `//` on L111.

pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	if len(region) == 0 {
 		return nil, fmt.Errorf("region is required")
 	}
+	awsconfig := &aws.Config{}
+	awsconfig.Region = &region
```
abhinavdahiya: use https://docs.aws.amazon.com/sdk-for-go/api/aws/#String for all string -> *string

pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
+
+	// TODO: This API will only return hostedzone resources (which are global)
+	// when the region is forced to us-east-1. We don't yet understand why.
+	tagRegion := "us-east-1"
+
+	//If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
+	// 1. hard code route53 api endpoint to https://api.route53.cn, as route53 is not GA in AWS China, and aws sdk didn't have the endpoint.
+	// 2. use the aws china regions to setup tagging api correctly in stead of "use-east-1"
+	if (region == "cn-north-1") || (region == "cn-northwest-1") {
```
abhinavdahiya:
```
select region {
case endpoints.CnNorth1RegionID, endpoints.CnNorthwest1RegionID:
....
}
```
try and use the const from https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants

pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	if len(region) == 0 {
 		return nil, fmt.Errorf("region is required")
 	}
+	awsconfig := &aws.Config{}
```
abhinavdahiya: i think this makes it seem like awsconfig for all clients, when this is for r53 client only, So let's name it so.

pkg/dns/aws/dns.go
```diff
@@ -101,13 +101,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
+	// when the region is forced to us-east-1. We don't yet understand why.
+	tagRegion := "us-east-1"
+
+	//If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
+	// 1. hard code route53 api endpoint to https://api.route53.cn, as route53 is not GA in AWS China, and aws sdk didn't have the endpoint.
+	// 2. use the aws china regions to setup tagging api correctly in stead of "use-east-1"
+	if (region == "cn-north-1") || (region == "cn-northwest-1") {
+		endpoint := "https://route53.amazonaws.com.cn"
+		awsconfig.Endpoint = &endpoint
+		tagRegion = region
```
abhinavdahiya: for all public regions we use `us-east-1`

can you verify that the resourcetaggingapi returns results for hosted zone for both regions north-1 and northwest-1 or do we have to pick one.

if you have access to aws china maybe add the aws cli invocation results?

wanghaoran1988: Good catch, just confirmed with AWS SA, this service hosted in cn-northwest-1 region:
```
Failed when using cn-north-1:
aws route53  list-hosted-zones  --region cn-north-1 --endpoint-url https://route53.amazonaws.com.cn 
aws route53   list-tags-for-resources  --region cn-north-1 --endpoint-url https://route53.amazonaws.com.cn --resource-id <id> --resource-type hostedzone
Succeed with cn-northwest-1:
aws route53  list-hosted-zones  --region cn-northwest-1 --endpoint-url https://route53.amazonaws.com.cn --> succeed
aws route53   list-tags-for-resources  --region cn-northwest-1 --endpoint-url https://route53.amazonaws.com.cn --resource-id <id> --resource-type hostedzone
```

pkg/dns/aws/dns.go
wanghaoran1988: Are you talking using "switch"?

wanghaoran1988:
@iranzo @abhinavdahiya Comments addressed, could you take another look?

pkg/dns/aws/dns.go
```diff
@@ -102,12 +103,26 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 		return nil, fmt.Errorf("region is required")
 	}
 
+	// TODO: This API will only return hostedzone resources (which are global)
+	// when the region is forced to us-east-1. We don't yet understand why.
```
abhinavdahiya: so this comment was for resourcetagging api https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75L108-L110 and the current setup makes it look like it's for r53Config...

abhinavdahiya:
> @iranzo @abhinavdahiya Comments addressed, could you take another look?

there is a inconsistency in how we are generating aws config.

// places that directly create or set aws.Config struct
https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R108
https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R118-R119

// places that use functions to create or update aws.Config
https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R109
https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R117

it just seems odd, any reason why it's that and why we cannot just use one way?

pkg/dns/aws/dns.go
wanghaoran1988: done

wanghaoran1988:
> > @iranzo @abhinavdahiya Comments addressed, could you take another look?
> 
> there is a inconsistency in how we are generating aws config.
> 
> // places that directly create or set aws.Config struct
> https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R108
> https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R118-R119
> 
> // places that use functions to create or update aws.Config
> https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R109
> https://github.com/openshift/cluster-ingress-operator/pull/361/files#diff-e30b09d9b7476806ee0e9a9de47a7d75R117
> 
> it just seems odd, any reason why it's that and why we cannot just use one way?

All updated to use functions

wanghaoran1988:
/retest 
@abhinavdahiya Could you take another look?

abhinavdahiya:
LGTM

ironcladlou:
Thank you all for working through this.

/lgtm

PR #335, opened 2019-12-08
Add support for IBM cloud platform load balancers

AndreaM12345:
Need to get approval for https://github.com/openshift/api/pull/536 and then can update the glide to pull in the new Platform

PR #362, opened 2020-02-19
Bug 1800338: [release-4.3] update to setDefaultPublishingStrategy() to add IBMCloudPlatform

AndreaM12345:


Miciah:
The history relating to this change is confusing and took some effort to sort out, so I would like to document it in case anyone else is confused.

The change that this PR makes was merged into master with #335.  However, it does not appear in any normal commit on master: the change was apparently accidentally incorporated into a _merge_ commit, and the _normal_ commit with the change was evidently dropped before #335 was merged.

After #335 was merged, #355 was made to cherry-pick the changes in #335 to release-4.3.  However, the cherry-pick omitted merge commits (which is normal), and so it also omitted this change.

This PR extracts the change from the merge commit into a sort of "synthetic" normal commit.  As a result, the commit in this backport does not exist on master.

We cannot fix master's history at this point, and after this backport, release-4.3's history will show a commit that does not exist on master, which is unfortunate but unavoidable now.  However, going forward, I think we should avoid having any changes other than actual conflict resolutions in merge commits; substantive changes should _always_ be in normal commits.  Unfortunately, GitHub makes it easy to miss this sort of thing if one only reviews files changed and not individual commits.

bparees:
skipping this one for approval this week due to bug severity and the number of other changes already accepted.

PR #363, opened 2020-02-21
Tune AWS load balancers to be consistent with other platforms

smarterclayton:
GCP and Azure both hardcode LB timeouts - set AWS to match the lower bound of Azure (2 failures @ 5s interval, 2 healthy). By default AWS is 6 failures @ 10s which is a very long timeout for the router.

In the future, if these attributes become customizable on GCP and Azure, they should be consistent with our AWS settings.

If we see flakiness at this lower check interval, we may want to consider bumping to 3 successive failures. However, it's generally best to start at the low end and then tune up to be as responsive
as possible to rollouts.

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -40,6 +40,7 @@ var (
 			"service.beta.kubernetes.io/aws-load-balancer-internal": "0.0.0.0/0",
 		},
 		configv1.AzurePlatformType: {
+			// Azure load balancers are not customizable and are set to (2 fail @ 5s interval, 2 healthy)
 			"service.beta.kubernetes.io/azure-load-balancer-internal": "true",
 		},
 		// There are no required annotations for this platform as of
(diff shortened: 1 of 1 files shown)


smarterclayton:
@sttts when we picked numbers for kube-apiserver, what were our bounds?  Any considerations I might be missing?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -38,15 +38,22 @@ var (
 	InternalLBAnnotations = map[configv1.PlatformType]map[string]string{
```
ironcladlou: These annotations are only for setting up internal LBs — to apply annotations to all LB service resources, gotta add them [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/load_balancer_service.go#L117)

sttts:
> @sttts when we picked numbers for kube-apiserver, what were our bounds? Any considerations I might be missing?

We started out with 10 sec interval and threshold 3. We never questioned those numbers. The 70 sec minimal termination delay was just derived from that, working well in practice. So no strong analysis behind all of that.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -40,6 +40,7 @@ var (
 			"service.beta.kubernetes.io/aws-load-balancer-internal": "0.0.0.0/0",
 		},
 		configv1.AzurePlatformType: {
+			// Azure load balancers are not customizable and are set to (2 fail @ 5s interval, 2 healthy)
```
ironcladlou: Looks like a leftover that needs deleted

Miciah:
/retitle Bug 1809665: Tune AWS load balancers to be consistent with other platforms
/cherry-pick release-4.4
/cherry-pick release-4.3

PR #366, opened 2020-03-02
Router should deploy with a very long grace period

smarterclayton:
In order to allow routers to be tuned to shutdown when a variable
length of connections have completed, the grace period needs to be
set fairly high. For now, set it to 1 hour until we determine what
tunable limits we should allow. In general few transactions will
take 1 hour to complete.


smarterclayton:
4.4 bug is 1809667 and 4.3 bug is 1809668

PR #372, opened 2020-03-09
update externalTrafficPolicy to cluster for IBMCloud platform

AndreaM12345:


ironcladlou:
At a glance, this would appear to break client source IP preservation — can you explain what problem this patch solves?

AndreaM12345:
I think its just some quirks we have with the LB that IBM deploys.  By having clusterIP we have more flexibility for where the routers and LB pods land.

ironcladlou:
Won't this change result in traffic being allowed to flow from the LB to all nodes regardless of whether they're hosting ingress controller pods? And when that happens, the traffic can then hop from node to node, losing the client source IP... am I missing something?

I think we should clearly articulate what exactly is broken and how this change fixes it.

Miciah:
Does `service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type` enable PROXY protocol?  We define a constant with that value and the identifier `iksLBProxyProtocolAnnotations`:
https://github.com/openshift/cluster-ingress-operator/blob/cb601461711f1675acc72cba590f1a87a921ed5a/pkg/operator/controller/ingress/load_balancer_service.go#L27
However, the [documentation](https://www.ibm.com/support/knowledgecenter/bluemix_stage/containers/cs_loadbalancer.html) for that annotation describes it as, "Annotation to specify a private or public load balancer."  So it is not clear what that annotation really does.

Setting aside the matter of source preservation, the hop also has a performance implication, which is why we do use "Local" external traffic policy on AWS (where we do enable PROXY protocol).

AndreaM12345:
Yep source IP will be lost which we accept as a known limitation.  However we want traffic to be  allowed to flow from the LB to all nodes regardless of whether they're hosting ingress controller pods.  The reason for this is so if the scheduler needs to but the router on a different node than the LB it can be done.  Also in our LB case by choosing `Local` for the `ExternalTrafficPolicy` then we only send traffic to one router pod instead of the multiple router pods.  Having `Cluster` allows us to send to both router pods

AndreaM12345:
The iksLBProxyProtocolAnnotations = "service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type" 
annotation tells the LB where to look for IPs to assign to the LB..basically there is a CM in the cluster that indicates which IPs go with the public network and which go with the private network.

Miciah:
> Yep source IP will be lost which we accept as a known limitation. However we want traffic to be allowed to flow from the LB to all nodes regardless of whether they're hosting ingress controller pods. The reason for this is so if the scheduler needs to but the router on a different node than the LB it can be done. Also in our LB case by choosing `Local` for the `ExternalTrafficPolicy` then we only send traffic to one router pod instead of the multiple router pods. Having `Cluster` allows us to send to both router pods

The LB can only target one node at a time (i.e., it is really a proxy)?

> The iksLBProxyProtocolAnnotations = "service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type"
> annotation tells the LB where to look for IPs to assign to the LB..basically there is a CM in the cluster that indicates which IPs go with the public network and which go with the private network.

Thanks!  I made https://github.com/openshift/cluster-ingress-operator/pull/373 to clean up the annotation declarations.

AndreaM12345:
Not sure I understand the question about:
"The LB can only target one node at a time (i.e., it is really a proxy)?"

The LB can talk to more than one router but only when the ExternalTrafficPolicy is set to `Cluster` I guess not sure what the implications are if its a proxy or not a proxy

Miciah:
I mean, it seems like you are saying the LB points to a node (LB → node 1), and you want to set `ExternalTrafficPolicy` to `Cluster` so that the node forwards to other nodes (LB → node 1 → {node 2, node 3, ..., node _n_}), and I am wondering why you cannot configure the LB to point directly to the nodes (LB → {node 1, node 2, node 3, ..., node _n_}).  That is how a load balancer usually works (it balances across multiple endpoints), but it seems like you are saying the LB can only send traffic to one node (which is more like a proxy than a load balancer).  Am I misunderstanding?

AndreaM12345:
For the IBM default LB when `externalTrafficPolicy: Local` is set just puts the LB VIP on one of the cluster's worker nodes that hosts the router (or more generally whatever node hosts a backend pod).  So this is why the traffic would then just go to one router (and why we want the default now for IKS to be externalTrafficPolicy: Cluster until we get a better default LB implementation

AndreaM12345:
wondering if there were further questions or if this PR could be moved to testing?

AndreaM12345:
looks like the tests look good...let me know what other questions need to be answered

AndreaM12345:
@Miciah @ironcladlou Wondering what info is needed to move to approve?

ironcladlou:
@AndreaM12345 what I'd like to see is a detailed comment explaining what's going on here, as it's inconsistent with the way every other LB implementation we work with behaves. And also updating the commit message to explain the requirement. Fair?

AndreaM12345:
For just the IBMCloud Platform would like to change the LoadBalancer to be `Cluster` instead of `Local` for the `ExternalTrafficPolicy`  The reason is that the LoadBalancer for classic clusters is created within the cluster.  Understand that by setting the `ExternalTrafficPolicy` to `Cluster` will loose the source IP.  However loosing the source IP is an acceptable loss. The LB implementation puts the VIP right on one of the worker nodes.  It uses keepalived to maintain the VIP and ensure redundancy. From there if externalTrafficPolicy: Local it will only send traffic to pods on the local node (if there are multiple pods on that node it will balance between them). if externalTrafficPolicy: Cluster, then it balances traffic to all the pods in the cluster.

As far as how it is implemented, the LB just relies on the iptables rules kube-proxy puts in to send traffic from the VIP node to pod IPs (not targeting any nodes directly).

ironcladlou:
@AndreaM12345 okay, thanks — can you please add that info to a commend on the code change and to the commit message?

/approve

danehans:
@AndreaM12345 when you have a moment, can you please address https://github.com/openshift/cluster-ingress-operator/pull/372#issuecomment-604974410. When 4.6 opens, I'll tag.

PR #374, opened 2020-03-11
Implement logging API

Miciah:
#### Add helpers for deployment unit tests

* `pkg/operator/controller/ingress/deployment_test.go` (`checkDeploymentHasEnvVar`, `checkDeploymentHasContainer`): New helpers.
(`TestDesiredRouterDeployment`): Use the new helpers.

#### Bump openshift/api for logging API

Bump to [github.com/openshift/api@8cfd79130f9c03777ae5db16e2c55e1039f977aa](https://github.com/openshift/api/commit/8cfd79130f9c03777ae5db16e2c55e1039f977aa) to get the logging API.

* `go.mod`: Bump.
* `hack/update-generated-crd.sh`:
* `hack/verify-generated-crd.sh`: Update path to the vendored ingresscontroller CRD file.
* `go.sum`:
* `pkg/manifests/bindata.go`:

#### Implement logging API

This commit resolves [NE-285](https://issues.redhat.com/browse/NE-285).

* `pkg/operator/controller/ingress/deployment.go` (`RouterLogLevelEnvName`, `RouterSyslogAddressEnvName`, `RouterSyslogFormatEnvName`, `RouterSyslogFacilityEnvName`): New constants.
(`desiredRouterDeployment`): Check whether `spec.logging.access` is specified on the ingresscontroller, and configure logging accordingly.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Verify that the `unsupported-logging` annotation overrides `spec.logging.access`.  Verify that `spec.logging.access` has the desired effect.
(`TestInferTLSProfileSpecFromDeployment`): Use "logs" as the name of the syslog sidecar container in order to be consistent with the logging API.
* `test/e2e/operator_test.go` (`TestSyslog`): New test.  Create a pod that runs an rsyslog instance and an ingresscontroller that logs to that rsyslog instance and verify that the rsyslog instance logs an HTTP request.


pkg/operator/controller/ingress/deployment.go
```diff
@@ -325,7 +325,34 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	secretName := controller.RouterEffectiveDefaultCertificateSecretName(ci, deployment.Namespace)
 	deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName = secretName.Name
 
+	var (
+		useSyslogSidecar  bool
+		useSyslogEndpoint bool
```
ironcladlou, line 330: These bools are mutually exclusive and seem to be proxies for `Spec.Logging.Access.Destination.Type`... what if instead you declare up front a LoggingDestinationType pointer which defaults to `nil` and is set to either Container or Syslog depending on annotations or logging config, and then finally switch on the destination rather than the booleans?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -325,7 +325,34 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		accessLogLevel    string
+		accessLogFormat   string
+		syslogEndpoint    string
+		syslogFacility    string
```
ironcladlou, line 334: format/endpoint/facility seem like extracted fields from SyslogLoggingDestinationParameters — what if you use the actual struct instead?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -191,7 +191,7 @@ func TestDesiredRouterDeployment(t *testing.T) {
 	}
 
 	for _, container := range deployment.Spec.Template.Spec.Containers {
-		if container.Name == "syslog" {
+		if container.Name == "logs" {
```
ironcladlou, line 194: Latest api revision provides a ContainerLoggingSidecarContainerName constant

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -417,13 +467,48 @@ func TestDesiredRouterDeployment(t *testing.T) {
 
 	foundSyslogContainer = false
 	for _, container := range deployment.Spec.Template.Spec.Containers {
-		if container.Name == "syslog" {
+		if container.Name == "logs" {
```
ironcladlou, line 470: ContainerLoggingSidecarContainerName

pkg/operator/controller/ingress/deployment.go
```diff
@@ -385,11 +412,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		if len(syslogFacility) > 0 {
+			env = append(env, corev1.EnvVar{Name: "ROUTER_LOG_FACILITY", Value: syslogFacility})
+		}
+		env = append(env,
+			corev1.EnvVar{Name: "ROUTER_LOG_LEVEL", Value: accessLogLevel},
+			corev1.EnvVar{Name: "ROUTER_SYSLOG_ADDRESS", Value: syslogEndpoint},
+		)
+		if len(accessLogFormat) > 0 {
+			env = append(env, corev1.EnvVar{Name: "ROUTER_SYSLOG_FORMAT", Value: accessLogFormat})
+		}
```
ironcladlou, line 430: Time to introduce some constants for the environment variables? Worth making a file just for them?

Miciah: I introduced constants for the new variables in `deployment.go`.  I'm not sure it would be helpful to put them in a separate file—they don't make much sense to use in other files.

pkg/operator/controller/ingress/deployment_test.go
Miciah: Done.

pkg/operator/controller/ingress/deployment_test.go
Miciah: Updated.

pkg/operator/controller/ingress/deployment.go
Miciah: Done (using the `AccessLogging` type).

pkg/operator/controller/ingress/deployment.go
Miciah: Latest push does something along those lines, using `AccessLogging` to store the type and parameters (ensuring that the specified type's parameters field is non-nil).

Miciah:
Rebased.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -32,6 +32,46 @@ var otherToleration = corev1.Toleration{
...
+			}
+			t.Errorf("router Deployment has unexpected %q container", name)
+		}
+	}
+	if expect {
+		t.Errorf("router Deployment has no %q container", name)
+	}
+}
+
 func checkRollingUpdateParams(t *testing.T, deployment *appsv1.Deployment, maxUnavailable intstr.IntOrString, maxSurge intstr.IntOrString) {
```
ironcladlou, line 75: Nice little refactor here, thanks — one more thing it made me think of (not required, just a sidebar discussion) is you can now express your expectations with a struct and make a table test out of the env assertions, like

```
{
  env{name: WildcardRouteAdmissionPolicy, equals: "false"},
  env{name: "ROUTER_USE_PROXY_PROTOCOL", equals: nil},
}
```

And feed them into your assertion function. Likewise with container checks. And then keep adding sugar until we reinvent Ginkgo 😁

pkg/operator/controller/ingress/deployment.go
```diff
@@ -336,7 +341,49 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	secretName := controller.RouterEffectiveDefaultCertificateSecretName(ci, deployment.Namespace)
 	deployment.Spec.Template.Spec.Volumes[0].Secret.SecretName = secretName.Name
 
+	var (
+		accessLogging  *operatorv1.AccessLogging
+		accessLogLevel = "debug"
+	)
 	if enabled, level := ExtraLoggingEnabled(ci, ingressConfig); enabled {
```
ironcladlou, line 348: What if instead we delete that annotation and all the code to implement it? It was never supported.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -336,7 +341,49 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		case operatorv1.ContainerLoggingDestinationType:
+			accessLogging = &operatorv1.AccessLogging{
+				Destination: operatorv1.LoggingDestination{
+					Type:      operatorv1.ContainerLoggingDestinationType,
+					Container: &operatorv1.ContainerLoggingDestinationParameters{},
+				},
+				HttpLogFormat: ci.Spec.Logging.Access.HttpLogFormat,
+			}
+		case operatorv1.SyslogLoggingDestinationType:
+			if ci.Spec.Logging.Access.Destination.Syslog != nil {
```
ironcladlou, line 367: Would a `nil` field be a valid state in this case? It looks like if type is Syslog and `Spec.Logging.Access.Destination.Syslog` is nil, we do nothing and also don't reject the IngressController as invalid? Or does openapi handle this case somehow via the union type?

test/e2e/operator_test.go
```diff
@@ -869,6 +871,158 @@ func TestRouteAdmissionPolicy(t *testing.T) {
...
+	syslogPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-ingress",
+			Name:      "syslog",
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "syslog",
+					Image: "openshift/origin-haproxy-router:v4.0",
```
ironcladlou, line 885: Recycle the operand image used by the payload being tested?

test/e2e/operator_test.go
```diff
@@ -869,6 +871,158 @@ func TestRouteAdmissionPolicy(t *testing.T) {
...
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to observe syslog pod IP address: %v", err)
+	}
+
+	// Create an ingresscontroller that logs to the endpoint.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "syslog"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
```
ironcladlou, line 962: Do we need something even more lightweight than newPrivateController? For example, if you used NodePort, I think the test would be just as good but wouldn't consume a costly LB, DNS entries, etc

pkg/operator/controller/ingress/deployment.go
Miciah: All right, I can do that.

pkg/operator/controller/ingress/deployment.go
Miciah: I'll check whether specifying a nil value passes validation.  If it does, then yeah, the user can specify `spec.logging.access.destination.type: Syslog` and a nil value for `spec.logging.access.destination.syslog`, and the result will be that logging will not be enabled.  Without `spec.logging.access.destination.syslog`, we don't have an endpoint.

test/e2e/operator_test.go
Miciah: You mean first get the ingress controller deployment, look up the image that it is using, and use it here?  All right, makes sense.

test/e2e/operator_test.go
Miciah: `newPrivateController` uses the "Private" endpoint publishing strategy type, so it doesn't do LB or DNS.

ironcladlou: Whoops, confused it with internal. Nevermind.

test/e2e/operator_test.go
ironcladlou: Yeah, although it might actually be in cluster config API somewhere

pkg/operator/controller/ingress/deployment.go
Miciah: Yup, I can create an ingresscontroller with `spec.logging.access.destination.type: Syslog` and without `spec.logging.access.destination.syslog`.

pkg/operator/controller/ingress/deployment.go
Miciah: Deleted.  This led me to notice another problem:  The logic to create the configmap for rsyslog only looked at the annotation and not at `spec.logging.access`.  I have fixed this problem as well now.

test/e2e/operator_test.go
Miciah: Done.

pkg/operator/controller/ingress/deployment.go
Miciah: I pushed another changes that further tweaks that logic (it should only create the configmap if a container logging is used) and adds an end-to-end test that creates an ingress controller that uses container logging (if no configmap is created, the ingress controller pod should fail to start).

ironcladlou:
lgtm, will let @danehans do the honors after he gets a chance to look through it

pkg/operator/controller/ingress/deployment.go
```diff
@@ -488,6 +514,41 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	return deployment, nil
 }
 
+// accessLoggingForIngressController returns an AccessLogging value for the
```
danehans, line 517: @Miciah accessLoggingForIngressController returns a pointer not a value, right?

Miciah: Right, it's a pointer.  We can fix the comment in a follow-up.

PR #376, opened 2020-03-14
Delete pod disruption budget logic

Miciah:
#272 in 4.2 added logic to create pod disruption budgets.  We subsequently decided to remove pod disruption budgets because they caused problems for draining and for small clusters, and so #323 in 4.3 changed the logic to remove any existing pod disruption budget.  The logic is consequently unnecessary in 4.4 and later.

* `manifests/00-cluster-role.yaml`: Delete rule to allow the operator to manage the `poddisruptionbudgets` resource.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Delete call to `ensureRouterPodDisruptionBudget`.
* `pkg/operator/controller/ingress/poddisruptionbudget.go`: Delete file.
(`ensureRouterPodDisruptionBudget`, `desiredRouterPodDisruptionBudget`, `currentRouterPodDisruptionBudget`, `createRouterPodDisruptionBudget`, `deleteRouterPodDisruptionBudget`, `updateRouterPodDisruptionBudget`, `podDisruptionBudgetChanged`): Delete.
* `pkg/operator/controller/names.go` (`RouterPodDisruptionBudgetName`): Delete function.
* `test/e2e/operator_test.go` (`TestPodDisruptionBudgetDoesNotExist`): Delete test.


manifests/00-cluster-role.yaml
```diff
@@ -27,13 +27,6 @@ rules:
   verbs:
   - "*"
 
-- apiGroups:
-  - policy
-  resources:
-  - poddisruptionbudgets
-  verbs:
-  - "*"
-
```
ironcladlou: Will there be some window where, once this is applied, existing operator process reconciliations could fail to apply the PDB due to RBAC denials? Even if so I'm not sure it matters

PR #385, opened 2020-03-30
Add a dependency verification script

ironcladlou:
Add a `verify` target and decouple from `test` because verification is only required as a presubmit check once https://github.com/openshift/release/pull/8016 merges.


ironcladlou:
/hold

Waiting on https://github.com/openshift/release/pull/8016

PR #384, opened 2020-03-30
Bumps k8s and openshift/api deps

danehans:
In support of the [Wildcard Admission Policy](https://github.com/openshift/enhancements/pull/252) feature.

/assign @ironcladlou @knobunc 
/cc @Miciah @frobware

PR #386, opened 2020-03-31
Add a dependency verification script

ironcladlou:
Add a `verify` target and decouple from `test` because verification is only required as a
presubmit check once https://github.com/openshift/release/pull/8016 merges.

### DIFF
--- a/hack/verify-deps.sh
+++ b/hack/verify-deps.sh
@@ -2,11 +2,16 @@
 set -euo pipefail
 
 function print_failure {
-  echo "There appear to be uncommitted changes to go.mod. Please run 'go tidy' and 'go vendor'."
+  echo "There are unexpected changes to the vendor tree following 'go mod vendor' and 'go mod tidy'. Please"
+  echo "run these commands locally and double-check the Git repository for unexpected changes which may"
+  echo "need to be committed."
   exit 1
 }
 
-go mod vendor
-go mod tidy
+if [ "${OPENSHIFT_CI:-false}" = true ]; then
+  go mod vendor
+  go mod tidy
 
-test -z "$(git status --porcelain)" || print_failure
+  test -z "$(git status --porcelain)" || print_failure
+  echo "verified Go modules"
+fi
--- a/hack/verify-deps.sh
+++ b/hack/verify-deps.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+function print_failure {
+  echo "There appear to be uncommitted changes to go.mod. Please run 'go tidy' and 'go vendor'."
+  exit 1
+}
+
+go mod vendor
+go mod tidy
+
+test -z "$(git status --porcelain)" || print_failure
--- a/Makefile
+++ b/Makefile
@@ -39,17 +39,8 @@ bindata:
 crd:
 	hack/update-generated-crd.sh
 
-.PHONY: verify-bindata
-verify-bindata:
-	hack/verify-generated-bindata.sh
-
-# Do not write the CRDs, only compare and return (code 1 if dirty).
-.PHONY: verify-crd
-verify-crd:
-	hack/verify-generated-crd.sh
-
 .PHONY: test
-test: verify
+test:
 	$(GO) test ./...
 
 .PHONY: release-local
@@ -66,8 +57,11 @@ clean:
 	rm -f $(BIN)
 
 .PHONY: verify
-verify: verify-crd verify-bindata
+verify:
 	hack/verify-gofmt.sh
+	hack/verify-generated-crd.sh
+	hack/verify-generated-bindata.sh
+	hack/verify-deps.sh
 
 .PHONY: uninstall
 uninstall:


hack/verify-deps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+function print_failure {
+  echo "There appear to be uncommitted changes to go.mod. Please run 'go tidy' and 'go vendor'."
+  exit 1
+}
+
+go mod vendor
+go mod tidy
```
Miciah: Does this only make sense to do in the CI context?

ironcladlou: If you were to run it locally, you could catch the problem _before_ you pushed... is there a reason you wouldn't want it to run locally?

hack/verify-deps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+function print_failure {
+  echo "There appear to be uncommitted changes to go.mod. Please run 'go tidy' and 'go vendor'."
```
frobware: Instructions should be `go _mod_ {tidy,vendor}`

hack/verify-deps.sh
Miciah: The script runs `go mod vendor` and `go mod tidy` and then tells the user to run `go mod vendor` and `go mod tidy`.  What would running the same commands a second time accomplish?

hack/verify-deps.sh
frobware: And, arguably, listed as `vendor`, then `tidy`.

hack/verify-deps.sh
ironcladlou: The instructions are wrong — should it instead say to review the diffs?

Miciah: In the CI context, it makes sense to tell the use to run the command and commit the changes.  In the local-development context, it makes sense to tell the user to review the diff and commit.  (It would be nice if `go mod` had a `--dry-run` flag.)

ironcladlou: I changed the script to only verify deps when running in a CI environment (see https://steps.svc.ci.openshift.org/help#env)

hack/verify-deps.sh
ironcladlou: Fixed

hack/verify-deps.sh
```diff
@@ -2,11 +2,16 @@
 set -euo pipefail
 
 function print_failure {
-  echo "There appear to be uncommitted changes to go.mod. Please run 'go tidy' and 'go vendor'."
+  echo "There are unexpected changes to the vendor tree following `go mod vendor` and `go mod tidy`. Please"
```
Miciah: Use single-quotes to prevent command expansion of ``` `go mod vendor` ``` and ``` `go mod tidy` ```.

ironcladlou:
Unfortunately nowhere in https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/386/pull-ci-openshift-cluster-ingress-operator-master-verify/12/ can I see the output of the `make verify` execution to make sure it worked...

ironcladlou:
going to do a negative test with a junk change to go.mod in this PR

/hold

hack/verify-deps.sh
```diff
@@ -2,11 +2,16 @@
...
+  echo "run these commands locally and double-check the Git repository for unexpected changes which may"
+  echo "need to be committed."
   exit 1
 }
 
-go mod vendor
-go mod tidy
+if [ "${OPENSHIFT_CI:-false}" = true ]; then
```
Miciah: So now the entire verification step is skipped outside of CI context?  I suppose that is fine.  It might be better to add a separate make target for CI-specific verification, but maybe that would be more effort than it would be worth.

ironcladlou:
Negative test looks okay other than the expansion issue

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/386/pull-ci-openshift-cluster-ingress-operator-master-verify/14

hack/verify-deps.sh
ironcladlou: Fixed

hack/verify-deps.sh
ironcladlou: Yes. For now just run the script directly and set the env var.

Miciah:
/hold
The verify job is now failing:
```
There are unexpected changes to the vendor tree following 'go mod vendor' and 'go mod tidy'. Please
run these commands locally and double-check the Git repository for unexpected changes which may
need to be committed.
```
This PR needs to be rebased to resolve the issue.

Miciah:
Actually, I think the job is fine, but #384 put the tree in a state that fails verification.  #388 should resolve the issue.

ironcladlou:
Looks like the script did its job? That's good news!

/hold cancel

PR #387, opened 2020-03-31
Bug 1809665: Re-add pod disruption budget for ingress controllers

Miciah:
#### Add a pod disruption budget for ingress controllers

For each ingress controller, create a pod disruption budget with a maximum of 25% unavailable pods.

* `manifests/00-cluster-role.yaml`: Allow the operator to manage the `poddisruptionbudgets` resource.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Call `ensureRouterPodDisruptionBudget`.
* `pkg/operator/controller/ingress/poddisruptionbudget.go`: New file.
(`ensureRouterPodDisruptionBudget`): New method.  Ensure the appropriate pod disruption budget exists for the given ingress controller.  Get the current one using `currentRouterPodDisruptionBudget`.  If none exists, create one using the `desiredRouterPodDisruptionBudget`.  If one already exists, update it if necessary using `updateRouterPodDisruptionBudget`, or delete it if none is desired.
* pkg/operator/controller/names.go (`RouterPodDisruptionBudgetName`): New function.
* test/e2e/operator_test.go (`TestPodDisruptionBudgetExists`): New test. Verify that a pod disruption budget exists for the default ingress controller.

#### Use max unavailable 25%, max surge 25%

Change the parameters for an ingress controller deployment's rolling update strategy when the ingress controller's endpoint publishing strategy type is `Private`, `LoadBalancerService`, or `NodePortService`: Set max unavailable to 25% to prevent the deployment from immediately reporting negative availability when it loses a replica, and set max surge to 25% to allow rollouts to progress more quickly.

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Configure the rolling update strategy with max unavailable set to 25% and max surge set to 25% if the endpoint publishing strategy type is `Private`, `LoadBalancerService`, or `NodePortService`.

----

@smarterclayton, @ironcladlou


pkg/operator/controller/ingress/deployment.go
```diff
@@ -128,8 +128,8 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		deployment.Spec.Strategy = appsv1.DeploymentStrategy{
 			Type: appsv1.RollingUpdateDeploymentStrategyType,
 			RollingUpdate: &appsv1.RollingUpdateDeployment{
-				MaxUnavailable: pointerTo(intstr.FromInt(0)),
-				MaxSurge:       pointerTo(intstr.FromInt(1)),
+				MaxUnavailable: pointerTo(intstr.FromString("25%")),
```
danehans, line 131: inline docs should be updated.

Miciah: Fixed.  Thanks!

Miciah:
Latest push updates the comment on the modified rolling update strategy parameters and increases max unavailable to 50%.  Max unavailable is rounded down, so if replicas is set to 2, a max unavailable value of 25% would round to 0 whereas a max unavailable value of 50% rounds to 1.

ironcladlou:
/hold

How can we automate disruption testing to prove we've got these scenarios right? And, by extension, to prove the current behavior is wrong?

danehans:
@ironcladlou what about something like https://github.com/openshift/origin/pull/24471 for routes?

ironcladlou:
I think for these scenarios, @smarterclayton summed it up:

>An easy one is to trigger three drains at the same time.  That would be a product level disruption suite.

(Is that something we can hope to achieve in the operator level e2e?)

danehans:
> (Is that something we can hope to achieve in the operator level e2e?)

@smarterclayton [suggests](https://coreos.slack.com/archives/G0113MXMESJ/p1586305332408600?thread_ts=1586304754.407300&cid=G0113MXMESJ) that the test be global.

Miciah:
/title Bug 1809665: Re-add pod disruption budget for ingress controllers

Miciah:
/retitle Bug 1809665: Re-add pod disruption budget for ingress controllers

ironcladlou:
Let's soak test this.

/lgtm

PR #389, opened 2020-04-03
DO NOT MERGE: Disable service traffic policy local for testing

smarterclayton:
The local policy may be incompatible with proper graceful rollout
operation.


danehans:
@smarterclayton can this PR be closed?

PR #390, opened 2020-04-12
Bug 1809354: dns: Avoid unnecessary updates

Miciah:
#### dns: Avoid spuriously updating a `DNSRecord`'s status

* `pkg/operator/controller/dns/controller.go` (`Reconcile`): Use the new `dnsZoneStatusSlicesEqual` function to check whether an update is needed, and skip the update if it not.
(`publishRecordToZones`): Use the new `mergeStatuses` function to merge the new zone statuses into the existing ones.
(`mergeStatuses`): New function.  Merge a slice of updated zone statuses into a slice of old zone statuses, using the new `mergeConditions` function.
(`clock`): New variable.
(`mergeConditions`): New function.  Merge a slice of updated zone status conditions into an slice of old status conditions, using `conditionChanged`.
(`conditionChanged`): New function.  Return a Boolean value indicating whether two zone status conditions should be considered equal.
(`dnsZoneStatusSlicesEqual`): New function.  Return a Boolean value indicating whether two slices of zone statuses should be considered equal for the purpose of determining whether a status update is needed.
* `pkg/operator/controller/dns/controller_test.go` (`TestPublishRecordToZonesMergesStatus`): New test.  Verify that `publishRecordToZones` correctly merges status updates.
(`TestDnsZoneStatusSlicesEqual`): New test.  Verify that `dnsZoneStatusSlicesEqual` behaves correctly.

#### dns: Log success or failure to publish record

* `pkg/operator/controller/dns/controller.go` (`publishRecordToZones`): Log success or failure to publish the DNS record to a given DNS zone.

#### Bump openshift/api for DNSRecord observedGeneration

Bump to github.com/openshift/api@17ada6e4245b4cc7a1f81ec04228e9093ec4f89e to get `DNSRecordStatus`'s `ObservedGeneration` field.

* `go.mod`: Bump.
* `go.sum`:
* `manifests/00-custom-resource-definition-internal.yaml`:
* `pkg/manifests/bindata.go`:
* `vendor/*`: Regenerate.

#### dns: Avoid publishing already published DNSRecords

When publishing a `DNSRecord` to a zone, check its status to determine whether the `DNSRecord` is already published to that zone, and if it is, skip publishing it to that zone.

* `pkg/operator/controller/dns/controller.go` (`Reconcile`): Update the `DNSRecord`'s observed generation.
(`publishRecordToZones`): Use the observed generation and the new `recordIsAlreadyPublishedToZone` function to skip publishing the `DNSRecord` to a zone if the `DNSRecord` has previously been observed and its status conditions indicate that it is already published to the zone.
(`recordIsAlreadyPublishedToZone`): New function.  Use the provided `DNSRecord`'s status to determine whether the `DNSRecord` is already published to the provided zone.
* `pkg/operator/controller/dns/controller_test.go` (`TestRecordIsAlreadyPublishedToZone`): New test.

#### dns/aws: Delete record caching

Delete caching in the DNS provider implementation for AWS now that the DNS controller uses the `DNSRecord`'s status to avoid performing unnecessary updates.

* `pkg/dns/aws/dns.go` (`Provider`): Delete the `updatedRecords` field.
(`NewProvider`): Delete initialization of `updatedRecords`.
(`change`): Delete updating of `updatedRecords`.


ironcladlou:
As discussed offline: this doesn't address a perhaps bigger problem which is that we're still unnecessarily updating the cloud provider every time because we're not comparing spec to status to identify diffs. If we did, not only would we only update the cloud provider when we think there's a change to be made, we could remove the AWS provider's internal caching logic.

We also discussed how when Ensure() fails, we should only update the condition last transition time if the condition has a material change, and in the face of repeated failures, we should log events so that a timeline of failure can be established.

Miciah:
The test failure should be resolved by #396.

/test e2e-aws-operator

This change should also fix https://bugzilla.redhat.com/show_bug.cgi?id=1837324.

/retitle Bug 1809354, 1837324: dns: Avoid unnecessary updates

Miciah:
Apparently support for linking multiple bugs was never merged.

/retitle Bug 1809354: dns: Avoid unnecessary updates

Miciah:
I forgot that the reason that this was marked WIP was that it was blocked on https://github.com/openshift/api/pull/644.

/retitle WIP: Bug 1809354: dns: Avoid unnecessary updates

pkg/operator/controller/dns/controller.go
```diff
@@ -145,6 +147,13 @@ func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1
 	result := reconcile.Result{}
 	var statuses []iov1.DNSZoneStatus
 	for i := range zones {
+		// Only publish the record if the DNSRecord has been modified
```
danehans, line 150: @Miciah can the TODO in L143-146 be removed as part of this change?

Miciah: Yup!  Fixed.  Thanks!

danehans:
@Miciah regarding https://github.com/openshift/cluster-ingress-operator/pull/390#issuecomment-630823224, yes that's my understanding. Then the go.mod/go.sum in this PR needs to be updated to use openshift/api instead of your modified version.

Miciah:
Rebased onto github.com/openshift/api@17ada6e4245b4cc7a1f81ec04228e9093ec4f89e.

Miciah:
Rebased to resolve a conflict in  `bindata.go` from #401.

danehans:
`error: unable to parse image registry.svc.ci.openshift.org/ci-op-qxhzdvvc/stable@sha256:bc679512a4ddcf2d7b433ed99b0ef3014ec167cddfcba95079b38aae5e28d0ef: cannot retrieve image configuration for manifest sha256:bc679512a4ddcf2d7b433ed99b0ef3014ec167cddfcba95079b38aae5e28d0ef: received unexpected HTTP status: 503 Service Unavailable
`

/retest

Miciah:
I added a test case to `TestPublishRecordToZonesMergesStatus` to make sure we have complete coverage of `mergeConditions` (thanks @sgreene570!).

PR #391, opened 2020-04-16
Bug 1834989: hashableDeployment: Fix liveness/readiness probes

Miciah:
#### `updateRouterDeployment`: Add diff to log message

* `pkg/operator/controller/ingress/deployment.go` (`updateRouterDeployment`): Add the diff of the previous and updated deployments to the "updated router deployment" log message.

#### `hashableDeployment`: Fix liveness/readiness probes

When comparing deployments to determine whether an update is required, treat implicit and explicit default values as equal.

Before this commit, the update logic would keep trying to revert the default values that the API set for probe parameters.

* `pkg/operator/controller/ingress/deployment.go` (`hashableDeployment`): Use the new `hashableProbe` function so that two deployments have the same hash if the only difference between them is that one does not specify liveness or readiness probe parameters and the other specifies the default parameter values.
(`hashableProbe`): New function.  Return a copy of the given probe, copying exactly the fields that should be used for computing a deployment's hash, and zeroing out any fields where the values are equal to their defaults.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -665,9 +665,8 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 			Image:           container.Image,
 			ImagePullPolicy: container.ImagePullPolicy,
 			Name:            container.Name,
-			LivenessProbe:   container.LivenessProbe,
-			ReadinessProbe:  container.ReadinessProbe,
-			VolumeMounts:    container.VolumeMounts,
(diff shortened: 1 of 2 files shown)


danehans:
`e2e-aws-operator` failed due to an e2e polling timeout fixed by  https://github.com/openshift/cluster-ingress-operator/commit/17c6350cc719bf90da9b3e0667515f1a8c9870f2.

/test e2e-aws-operator

danehans:
Same as https://github.com/openshift/cluster-ingress-operator/pull/391#issuecomment-628329615

/test e2e-aws-operator

PR #392, opened 2020-04-20
Bug 1826113: status: Improve LoadBalancerManaged=false wording

Miciah:
Reword the `LoadBalancerManaged` status condition's reason and message to avoid using phrases such as "unsupported" or "does not support" when the endpoint publishing strategy does not include managing a load balancer.

* `pkg/operator/controller/ingress/status.go` (`computeLoadBalancerStatus`): Change reason and message strings for the `LoadBalanacerManaged` status condition.


danehans:
`level=error msg="Error: error updating tags: error tagging resource (dopt-07c6ca6efdd171711): InvalidDhcpOptionsID.NotFound: The dhcpOptions ID 'dopt-07c6ca6efdd171711' does not exist"
`

/test e2e-aws

ironcladlou:
I wonder how this is a valid gh label state:

bugzilla/unspecified
bugzilla/valid-bug

/lgtm

PR #394, opened 2020-04-27
Support changing ingresscontroller load balancer scope

ironcladlou:
Add support for changing ingresscontroller load balancer scope. Handle this by
deleting the existing load balancer service and re-creating the service using
the desired scope.

/cc @openshift/sig-network-edge @mwoodson

### DIFF
--- a/hack/run-local.sh
+++ b/hack/run-local.sh
@@ -11,4 +11,4 @@ RELEASE_VERSION=$(oc get clusterversion/version -o json | jq -r '.status.desired
 echo "Image: ${IMAGE}"
 echo "Release version: ${RELEASE_VERSION}"
 
-IMAGE="${IMAGE}" RELEASE_VERSION="${RELEASE_VERSION}" ${DELVE:-} ./ingress-operator start $@
+IMAGE="${IMAGE}" RELEASE_VERSION="${RELEASE_VERSION}" ${DELVE:-} ./ingress-operator start --image $IMAGE --release-version $RELEASE_VERSION --namespace openshift-ingress-operator --shutdown-file "" $@
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -333,6 +333,12 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig 
 		ic.Status.EndpointPublishingStrategy = effectiveStrategy
 		return true
 	}
+	// Detect changes to LB scope, which is something we can safely roll out.
+	specLB := ic.Status.EndpointPublishingStrategy.LoadBalancer
+	statusLB := effectiveStrategy.LoadBalancer
+	if specLB != nil && statusLB != nil && specLB.Scope != statusLB.Scope {
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.Scope = effectiveStrategy.LoadBalancer.Scope
+	}
 	return false
 }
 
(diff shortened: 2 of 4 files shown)


hack/run-local.sh
```diff
@@ -11,4 +11,4 @@ RELEASE_VERSION=$(oc get clusterversion/version -o json | jq -r '.status.desired
 echo "Image: ${IMAGE}"
 echo "Release version: ${RELEASE_VERSION}"
 
-IMAGE="${IMAGE}" RELEASE_VERSION="${RELEASE_VERSION}" ${DELVE:-} ./ingress-operator start $@
+IMAGE="${IMAGE}" RELEASE_VERSION="${RELEASE_VERSION}" ${DELVE:-} ./ingress-operator start --image $IMAGE --release-version $RELEASE_VERSION --namespace openshift-ingress-operator --shutdown-file "" $@
```
frobware, line 14: Quote `$@` as `"$@"`.

pkg/operator/controller/ingress/controller.go
```diff
@@ -333,6 +333,12 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
 		ic.Status.EndpointPublishingStrategy = effectiveStrategy
 		return true
 	}
+	// Detect changes to LB scope, which is something we can safely roll out.
+	specLB := ic.Status.EndpointPublishingStrategy.LoadBalancer
+	statusLB := effectiveStrategy.LoadBalancer
```
danehans, line 338: I believe you want to swap  specLB and statusLB var names.

test/e2e/operator_test.go
```diff
@@ -580,6 +580,109 @@ func TestInternalLoadBalancer(t *testing.T) {
 	}
 }
 
+func TestScopeChange(t *testing.T) {
+	platform := infraConfig.Status.Platform
+
+	supportedPlatforms := map[configv1.PlatformType]struct{}{
```
danehans, line 586: @ironcladlou is there a reason why GCP is not included?

danehans:
@Miciah other than a few nits, LGTM. I think the nits can be addressed as a follow-up, but I'll let you tag.

gregsheremeta:
Hi! What's the status of this? Is it blocked, or can it be merged?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -66,23 +69,58 @@ var (
...
+	case desired != nil && current == nil:
+		if err := r.createLoadBalancerService(desired); err != nil {
+			return nil, err
 		}
-		log.Info("created load balancer service", "namespace", desiredLBService.Namespace, "name", desiredLBService.Name)
-		return desiredLBService, nil
+	case desired != nil && current != nil:
+		// If switching from internal/external, delete and recreate the service.
+		if IsServiceInternal(desired) != IsServiceInternal(current) {
+			if _, err := r.finalizeLoadBalancerService(ci); err != nil {
```
Miciah, line 89: The cloud provider implementations for Azure and GCE apparently can handle changing between internal and external internally:

https://github.com/kubernetes/kubernetes/blob/0a14265b7e4c9d207ec4ebd1be687e029e7180d4/staging/src/k8s.io/legacy-cloud-providers/azure/azure_loadbalancer.go#L161-L197
https://github.com/kubernetes/kubernetes/blob/678d3f2841e8bc03582825a36e47a530d1fa16a4/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer.go#L129-L153

The cloud provider implementation for AWS cannot:

https://github.com/kubernetes/kubernetes/blob/0a14265b7e4c9d207ec4ebd1be687e029e7180d4/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L193

So it seems like we could skip this logic for Azure and GCE and instead just update the annotation on the service.

Miciah:
/close
Superseded by https://github.com/openshift/cluster-ingress-operator/pull/472.

PR #399, opened 2020-05-14
Bug 1813894: Add configmap for service CA bundle

Miciah:
Create a configmap that is annotated to request injection of the service CA bundle, and configure ingress controller deployments to mount this configmap and use it instead of using the service account token to get the service CA.

* `assets/router/deployment.yaml`: Add a volume and volume mount for the new configmap, and set the `DEFAULT_DESTINATION_CA_PATH` environment variable to configure openshift-router to use the bundle from the configmap.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Call `ensureServiceCAConfigMap` once the `openshift-ingress` namespace has been created and before creating the deployment.
* `pkg/operator/controller/ingress/serviceca_configmap.go`: New file.
(`ensureServiceCAConfigMap`): New method.  Ensure the configmap for the service CA bundle exists.
(`desiredServiceCAConfigMap`): New function.  Return the desired configmap for the service CA bundle.
(`currentServiceCAConfigMap`): New method.  Return the current configmap for the service CA bundle.
(`updateServiceCAConfigMap`): New method.  Update the service CA configmap bundle if the expected annotation is missing.
* `pkg/operator/controller/names.go` (`ServiceCAConfigMapName`): New function. Return the name of the configmap for the service CA bundle.


Miciah:
/test e2e-aws-operator

/retitle Bug 1813894: Add configmap for service CA bundle

@openshift/openshift-team-network-edge, this is ready for review.

pkg/operator/controller/ingress/serviceca_configmap.go
```diff
@@ -0,0 +1,104 @@
...
+		return false, nil, err
+	}
+	return true, cm, nil
+}
+
+// updateServiceCAConfigMap updates the configmap for the service CA bundle.
+// Returns a Boolean indicating whether the configmap was updated, and an error
+// value.
+func (r *reconciler) updateServiceCAConfigMap(current, desired *corev1.ConfigMap) (bool, error) {
+	if current.Annotations["service.beta.openshift.io/inject-cabundle"] == "true" {
```
frobware, line 92: This acts as a gate? Once present we never update - is that the implied semantics?  If so, maybe the check should occur by callers of `updateServiceCAConfigMap` so that update always means update.

Miciah: That is pretty much right: cluster-ingress-operator manages the configmap's lifecycle and annotation, and the serving cert signer manages the data.  If the configmap's annotation were deleted, cluster-ingress-operator would restore it, but otherwise it should not update the configmap.  Would it be best to fold `updateServiceCAConfigMap` into `ensureServiceCAConfigMap`?

frobware: For me this was mostly about the semantics. A function called `update...` doesn't always update, but only if you peek at its implementation. If it is only called from one place them maybe folding would help.

Miciah: Here, "updated" means "update if needed", and in this particular case, "if needed" means "if the critical annotation is not set properly".  It is kind of a degenerate case of a pattern we use elsewhere in the operator, where `ensureFoo` calls `createFoo`, `deleteFoo`, and `updateFoo`, and `updateFoo` determines whether update is needed.

frobware: OK, fine for me. Thanks for explaining the pattern.

PR #396, opened 2020-05-13
Bug 1835025: TestRouteAdmissionPolicy: Wait for rolling update

Miciah:
#### `TestRouteAdmissionPolicy`: Log error if `Get` fails

Fix test code to output the error value if a client `Get` call fails.

* `test/e2e/operator_test.go` (`TestRouteAdmissionPolicy`): Log errors from `Get`.

#### `TestRouteAdmissionPolicy`: Wait for rolling update

Poll the ingress controller's deployment after changing the route admission policy to make sure that the deployment is updated and rolled out before checking that the new policy is in effect.

* `test/e2e/operator_test.go` (`TestRouteAdmissionPolicy`): Add polling to make sure that the ingress controller deployment is updated and rolled out.

---

@danehans

PR #400, opened 2020-05-19
Bug 1835025: TestRouteAdmissionPolicy: Fix wait for deployment

Miciah:
Follow-up to #396.

* `test/e2e/operator_test.go` (`TestRouteAdmissionPolicy`): Use `waitForDeploymentComplete` instead of `waitForIngressControllerCondition`.
(`waitForDeploymentComplete`): New function.  Wait for the given deployment to complete its rollout.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -777,8 +777,8 @@ func TestRouteAdmissionPolicy(t *testing.T) {
 	// The updated ingresscontroller deployment may take a few minutes to
 	// roll out, so make sure that it is updated, and then make sure that it
 	// has finished rolling out before checking the route.
+	deployment := &appsv1.Deployment{}
 	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
-		deployment := &appsv1.Deployment{}
 		if err := kclient.Get(context.TODO(), controller.RouterDeploymentName(ic), deployment); err != nil {
 			return false, err
 		}
@@ -792,7 +792,7 @@ func TestRouteAdmissionPolicy(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to observe ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK=true: %v", err)
 	}
-	if err := waitForIngressControllerCondition(kclient, 3*time.Minute, icName, conditions...); err != nil {
+	if err := waitForDeploymentComplete(t, kclient, deployment, 3*time.Minute); err != nil {
 		t.Fatalf("failed to observe expected conditions: %v", err)
 	}
 
@@ -841,7 +841,6 @@ func TestRouteAdmissionPolicy(t *testing.T) {
 		t.Fatalf("failed to update ingresscontroller: %v", err)
 	}
 	err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
-		deployment := &appsv1.Deployment{}
 		if err := kclient.Get(context.TODO(), controller.RouterDeploymentName(ic), deployment); err != nil {
 			return false, err
 		}
@@ -855,7 +854,7 @@ func TestRouteAdmissionPolicy(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to observe ROUTER_ALLOW_WILDCARD_ROUTES=true: %v", err)
 	}
-	if err := waitForIngressControllerCondition(kclient, 3*time.Minute, icName, conditions...); err != nil {
+	if err := waitForDeploymentComplete(t, kclient, deployment, 3*time.Minute); err != nil {
 		t.Fatalf("failed to observe expected conditions: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


danehans:
`e2e-aws-operator` fails due to:

```
--- FAIL: TestInternalLoadBalancer (91.04s)
    operator_test.go:566: failed to observe expected conditions: Get https://api.ci-op-7sg17nf6-43abb.origin-ci-int-aws.dev.rhcloud.com:6443/apis/operator.openshift.io/v1/namespaces/openshift-ingress
```

/test e2e-aws-operator

Miciah:
`TestRouteHTTP2EnableAndDisableIngressController` timed out because the default ingresscontroller never became available because the Route 53 API was throttled:

```
The DNS provider failed to ensure the record: failed to update alias in zone Z09357351CVW2O185N0TP: couldn't update DNS record in zone Z09357351CVW2O185N0TP: Throttling: Rate exceeded
```

/test e2e-aws-operator

Miciah:
* could not run steps: step e2e-aws-operator failed: failed to acquire lease: resources not found

/test e2e-aws-operator

Miciah:
Failed to provision the cluster:

    Error: Unable to find matching route for Route Table (rtb-0182f841594046c3b) and destination CIDR block (0.0.0.0/0).

/test e2e-aws-operator

Miciah:
The test passed.  Let's see whether it passes consistently.

/test e2e-aws-operator

Miciah:
Passed again.

/test e2e-aws-operator

Miciah:
@danehans, the test has passed 3 times in a row now, and the earlier e2e-aws-operator failures since the last push were caused by other tests or provisioning failures, so I am feeling fairly confident that it resolves the problem.  Could you take another look please?

test/e2e/operator_test.go
```diff
@@ -1163,6 +1162,38 @@ func waitForAvailableReplicas(cl client.Client, ic *operatorv1.IngressController
...
 }
 
+// Wait for the provided deployment to complete its rollout.
+func waitForDeploymentComplete(cl client.Client, deployment *appsv1.Deployment, timeout time.Duration) error {
+	replicas := int32(1)
+	name := types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name}
+	deployment = &appsv1.Deployment{}
+	err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+		if err := cl.Get(context.TODO(), name, deployment); err != nil {
+			return false, nil
```
sgreene570, line 1172: Should `false, err` be returned here, instead of `false, nil`?

Miciah: Probably, yeah.  Generally, I think we should either `t.Logf(..., err)` or `return false, err`.  In this case, the deployment should already exist, and we're only watching for a mutation, so using `return false, err` would make sense; on the other hand, doing so would make the test more likely to break if there are transient network or API issues.  Would it make most sense in general to `t.Logf(..., err)` and `return false, nil` to decrease the likelihood that tests fail due to networking or API failures unrelated to the test?

Miciah: https://bugzilla.redhat.com/show_bug.cgi?id=1828618 is related to the general question (maybe it prompted your comment here).

sgreene570: Yup, that BZ definitely prompted my comment. 

I definitely agree that we should either log the error responsibly and return `false, nil` *or* just flat out fail by returning return `false, err` in these situations.

To counter your point about transient network or API issues, if any of these issues errors were to persist for a length of time close to the timeout, returning `false, err` and failing the test immediately would be beneficial. Additionally, parts of the ingress operator's e2e test already fail on unsuccessful API calls (for example, https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L193). So I'm wondering how much we would really be decreasing the likelihood of the tests failing from transient issues by using the responsible logging and return `false, nil` strategy. For this reason, I am leaning towards turning `false, err` and outright failing. Thoughts?

Miciah: > To counter your point about transient network or API issues, if any of these issues errors were to persist for a length of time close to the timeout, returning `false, err` and failing the test immediately would be beneficial.

I disagree—if networking or API is broken, other tests should report that breakage.  Having this component's tests fail due to other components' breakage makes it more difficult to determine which component is at fault and leads to misdirected Bugzilla reports.

> Additionally, parts of the ingress operator's e2e test already fail on unsuccessful API calls (for example, https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L193). So I'm wondering how much we would really be decreasing the likelihood of the tests failing from transient issues by using the responsible logging and return `false, nil` strategy. For this reason, I am leaning towards turning `false, err` and outright failing. Thoughts?

That is why I am speaking generally—I believe this component's tests should not fail on errors that are caused by other components, so I am suggesting that as a general principle, we should log and retry on errors that are not caused by our component (and transient errors are very unlikely to be caused by our component).  Does that make sense?

sgreene570: That does make sense, and I think you've convinced me. I agree that logging and retrying transient errors, when applicable, makes the most sense. Let's include this conversation in https://bugzilla.redhat.com/show_bug.cgi?id=1828618 and continue the conversation there. Would like to here the rest of the teams thoughts if thats OK too.

Miciah:
Failed to launch the cluster.

/test e2e-aws-operator

danehans:
```
fail [github.com/openshift/origin/test/e2e/upgrade/upgrade.go:138]: during upgrade to registry.svc.ci.openshift.org/ci-op-3ht8lmy3/release@sha256:aed90ddf207d20929117ab1d13151e712d0281e899631bc703c656cc44f1fdad
Unexpected error:
    <*errors.errorString | 0xc002f35780>: {
        s: "Cluster did not complete upgrade: timed out waiting for the condition: Cluster operator openshift-apiserver is reporting a failure: APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable for apiserver.openshift-apiserver",
    }
    Cluster did not complete upgrade: timed out waiting for the condition: Cluster operator openshift-apiserver is reporting a failure: APIServerDeploymentDegraded: 1 of 3 requested instances are unavailable for apiserver.openshift-apiserver
occurred
```
I've seen this with other PRs recently.

/test e2e-aws-upgrade

PR #401, opened 2020-05-20
Bug 1837251: Add unsupported http/2 kill switch

frobware:
Adding:

  ingress.operator.openshift.io/unsupported-disable-http2: "true"

to either an ingresscontroller or the ingress configuration will
disable HTTP/2 support in the router deployment.

Corresponding router change: https://github.com/openshift/router/pull/133


frobware:
The failures might be genuine to this PR.... 

/retest

Miciah:
The operator's log output from the CI run repeats the following error:

    E0520 17:24:22.257635       1 reflector.go:178] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to list *v1.Ingress: ingresses.config.openshift.io is forbidden: User "system:serviceaccount:openshift-ingress-operator:ingress-operator" cannot list resource "ingresses" in API group "config.openshift.io" at the cluster scope

The operator has permission to get the resource:  https://github.com/openshift/cluster-ingress-operator/blob/8a8beb18fcd70d581a1375cac0978c1eb6688fad/manifests/00-cluster-role.yaml#L75-L84

However I think you need permission to list and watch:

```yaml
- apiGroups:
  - config.openshift.io
  resources:
  - ingresses
  verbs:
  - list
  - watch
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -83,9 +85,37 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &iov1.DNSRecord{}}, &handler.EnqueueRequestForOwner{OwnerType: &operatorv1.IngressController{}}); err != nil {
 		return nil, err
 	}
+	if err := c.Watch(&source.Kind{Type: &configv1.Ingress{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(reconciler.ingressConfigToIngressController)}, predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return true },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return true },
+		GenericFunc: func(e event.GenericEvent) bool { return true },
```
Miciah, line 92: The predicates can be omitted entirely if they always return true.

pkg/operator/controller/ingress/controller.go
```diff
@@ -83,9 +85,37 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+	}); err != nil {
+		return nil, err
+	}
 	return c, nil
 }
 
+func (r *reconciler) ingressConfigToIngressController(o handler.MapObject) []reconcile.Request {
+	var requests []reconcile.Request
+	controllers := &operatorv1.IngressControllerList{}
+	if err := r.cache.List(context.Background(), controllers, &client.ListOptions{Namespace: o.Meta.GetNamespace()}); err != nil {
```
Miciah, line 102: Use `client.InNamespace(r.Namespace)` instead of `&client.ListOptions{Namespace: o.Meta.GetNamespace()}`.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -498,6 +514,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: WildcardRouteAdmissionPolicy, Value: "false"})
 	}
 
+	if HTTP2IsDisabledByAnnotation(ci.Annotations) || HTTP2IsDisabledByAnnotation(ingressConfig.Annotations) {
```
Miciah, line 517: What if the cluster ingress config has `ingress.operator.openshift.io/unsupported-disable-http2=true` and the ingresscontroller has `ingress.operator.openshift.io/unsupported-disable-http2=false`?  Seems like the ingresscontroller's annotation should override the ingress config's.  Granted, this is an unsupported annotation, so I wouldn't worry too much about this if it makes the logic hairy.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -498,6 +514,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		log.Info("Disabling HTTP/2 for router deployment", "namespace", deployment.Namespace, "name", deployment.Name)
```
Miciah, line 518: This will get logged every time the operator reconciles the ingresscontroller, which seems excessive.  I'd just as soon drop the log statement.

test/e2e/operator_test.go
```diff
@@ -1076,6 +1077,162 @@ func TestContainerLogging(t *testing.T) {
...
+		}
+		if len(d.Spec.Template.Spec.Containers) < 1 {
+			t.Fatal("expected routerDeployment to have at least one container")
+		}
+		if http2IsDisabledInEnv(d.Spec.Template.Spec.Containers[0].Env) {
+			t.Fatal("expected routerDeployment to have http/2 enabled by default")
+		}
+	}
+
+	refreshObjects := func() (*operatorv1.IngressController, *configv1.Ingress, *appsv1.Deployment) {
```
Miciah, line 1101: Maybe `t.Helper()` here too?

Also, are you defining these functions inside `TestRouteHTTP2EnableAndDisable` in order to group them, or is there some other reason?  You could pass in `t` (as you already do with `assertPreConditions`), and I don't see any other free variables.  That said, I'm not opposed to defining the functions this way for grouping if that is the style you prefer.

All that said, it might be worth moving the test and its helpers to a separate file as `operator_test.go` has grown rather large.

test/e2e/operator_test.go
```diff
@@ -1076,6 +1077,162 @@ func TestContainerLogging(t *testing.T) {
...
+	if err := waitForIngressControllerCondition(kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+	if err := waitForRouterDeploymentHTTP2Disabled(kclient, ic, routerDeployTimeout, true); err != nil {
+		t.Fatalf("expected router routerDeployment to have http/2 disabled: %v", err)
+	}
+
+	// Revert the previous change to the default ingresscontroller
+	if err := disableHTTP2ForIngressController(false, ic); err != nil {
+		t.Errorf("failed to update infrastructure config: %v", err)
```
Miciah, line 1140: "infrastructure config" → "ingresscontroller".

test/e2e/operator_test.go
```diff
@@ -1076,6 +1077,162 @@ func TestContainerLogging(t *testing.T) {
...
+	if err := waitForIngressControllerCondition(kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	ic, ingressConfig, routerDeployment := refreshObjects()
+	assertPreConditions(t, ic, ingressConfig, routerDeployment)
+
+	// Disable http/2 on the default ingresscontroller
+	if err := disableHTTP2ForIngressController(true, ic); err != nil {
+		t.Errorf("failed to update infrastructure config: %v", err)
```
Miciah, line 1129: "infrastructure config" → "ingresscontroller".

test/e2e/operator_test.go
frobware: The purpose wasn't necessarily to group them but to contain their scope and their semantics to just the function under test. They call fatal (now with the helper) which may be undesriable as free standing functions. I can move refreshObjects() out but assertPreConditions() really is for that test function alone.

t.Helper - done.

pkg/operator/controller/ingress/deployment.go
frobware: > Seems like the ingresscontroller's annotation should override the ingress config's.

Interesting. I interpreted the ingress config to have the higher priority. Rather than annotating all the ingresscontrollers you could just add it to the ingress config to disable it everywhere.

frobware:
@Miciah I addressed all your feedback with the exception of changing the logic for who wins when the config is "true/false" vis-a-vis when the ingresscontroller disables with "true/false". How much complexity do we want here? If either is set, the ingress controller will have http/2 disabled.

I also moved the new e2e test into its own file and factored out the functions from the core test logic.

PTAL

frobware:
I did some manual testing with https://github.com/openshift/router/pull/133 and using the http/2 tests in origin I was able to see that backend would report "HTTP 1.1" when the kill switch was in effect. And testing against the grpc-interop-reencrypt route also failed.

pkg/operator/controller/ingress/deployment.go
Miciah: My reasoning is that if the user explicitly added an annotation to both the ingress config as well as to an individual ingresscontroller, then the user is expressing a general preference with the former annotation and an overriding preference for the specific ingresscontroller with the latter annotation.

Miciah:
You got a test failure:

    === RUN   TestRouteHTTP2EnableAndDisable
    --- FAIL: TestRouteHTTP2EnableAndDisable (1.27s)
        http2_disable_test.go:58: failed to update ingresscontroller: Operation cannot be fulfilled on ingresscontrollers.operator.openshift.io "default": the object has been modified; please apply your changes to the latest version and try again

You probably need to do a `Get` in `disableHTTP2ForIngressController` so that you have the latest revision when you do the `Update`.  That change probably would be sufficient, but if you wanted to be extra careful, you could use `wait.Poll` and retry the update.  (You could do the same in `disableHTTP2ForIngressConfig`, but it is unlikely that the ingress config will be updated during the test.)

frobware:
> That change probably would be sufficient, but if you wanted to be extra careful, you could use `wait.Poll` and retry the update

Going to do both. Admittedly I've only been testing the single http/2 e2e locally, relying on CI for the full gamut.

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	//
+	// Now assert that changing the "cluster" ingress
+	// configuration will also disable/enable http/2 for all
+	// ingress controllers.
+
+	if err := waitForIngressControllerCondition(kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
```
Miciah, line 74: This check is redundant (but harmless).

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+	ic, ingressConfig, routerDeployment, err := http2RefreshTestObjects(t, kclient, 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get test objects: %v", err)
+	}
+	if err := checkHTTP2IsEnabled(ic, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// Disable http/2 on the default ingresscontroller
+	if err := disableHTTP2ForIngressController(t, kclient, 1*time.Minute, true, ic); err != nil {
```
Miciah, line 46: As a mild suggestion, this would read better if it were something like `setHTTP2DisabledForIngressController(..., true, ...)` (and similar for `disableHTTP2ForIngressConfig`).

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+		if v.Name == ingresscontroller.RouterDisableHTTP2EnvName {
+			if val, err := strconv.ParseBool(v.Value); err == nil {
+				return val
+			}
+		}
+	}
+	return false
+}
+
+func http2RefreshTestObjects(t *testing.T, client client.Client, timeout time.Duration) (*operatorv1.IngressController, *configv1.Ingress, *appsv1.Deployment, error) {
```
Miciah, line 131: Looks like `t.Helper()` got lost.

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+	ic := operatorv1.IngressController{}
+	if err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+		if err := client.Get(context.TODO(), defaultName, &ic); err != nil {
+			t.Logf("Get %q failed: %v, retrying...", defaultName, err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get %q", err)
```
Miciah, line 140: `"failed to get %q"` → `"failed to get default ingresscontroller: %v"`.

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil, nil, nil, fmt.Errorf("failed to get ingress configuration: %v", err)
+	}
+
+	return &ic, &ingressConfig, &deployment, nil
+}
+
+func checkHTTP2IsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.Ingress, d *appsv1.Deployment) error {
+	icName := path.Join(ic.Namespace, ic.Name)
+	deploymentName := path.Join(d.Namespace, d.Name)
+	ingressConfigName := path.Join(ingressConfig.Namespace, ingressConfig.Name)
```
Miciah, line 172: Using `path.Join` here is a little odd—any reason to use that over using a namespaced name (which gets stringified to "namespace/name"), `fmt.Sprintf`, or a direct string concatenation?

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if ingress.HTTP2IsDisabledByAnnotation(ic.Annotations) {
+		return fmt.Errorf("expected ingress controller %q to have HTTP/2 enabled by default", icName)
+	}
+	if ingress.HTTP2IsDisabledByAnnotation(ingressConfig.Annotations) {
+		return fmt.Errorf("expected infrastructure config %q to have HTTP/2 enabled by default", ingressConfigName)
```
Miciah, line 178: "infrastructure config" → "ingress config".

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+		deployment := &appsv1.Deployment{}
+		if err := client.Get(context.TODO(), controller.RouterDeploymentName(c), deployment); err != nil {
+			return false, nil
+		}
+		return http2IsDisabledInEnv(deployment.Spec.Template.Spec.Containers[0].Env) == expectedDisabledStatus, nil
+	})
+}
+
+func disableHTTP2ForIngressConfig(t *testing.T, client client.Client, timeout time.Duration, disableHTTP2 bool, c *configv1.Ingress) error {
+	name := types.NamespacedName{Namespace: c.Namespace, Name: c.Name}
```
Miciah, line 200: `t.Helper()`.

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		t.Fatalf("failed to get test objects: %v", err)
+	}
+	if err := checkHTTP2IsEnabled(ic, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// Disable http/2 on the ingress config "cluster"
+	if err := disableHTTP2ForIngressConfig(t, kclient, 1*time.Minute, true, ingressConfig); err != nil {
+		t.Fatalf("failed to update infrastructure config: %v", err)
```
Miciah, line 86: "ingress config".

test/e2e/http2_disable_test.go
```diff
@@ -0,0 +1,245 @@
...
+		t.Fatalf("expected router deployment to have http/2 disabled: %v", err)
+	}
+
+	// Revert the previous change to the ingress config
+	ic, ingressConfig, routerDeployment, err = http2RefreshTestObjects(t, kclient, 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get test objects: %v", err)
+	}
+	if err := disableHTTP2ForIngressConfig(t, kclient, 1*time.Minute, false, ingressConfig); err != nil {
+		t.Fatalf("failed to update infrastructure config: %v", err)
```
Miciah, line 101: "ingress config".

test/e2e/http2_disable_test.go
frobware: These functions no longer call `Fatal` - they return a wrapped error. We carry `t` for logging through `t.Logf()`.

test/e2e/http2_disable_test.go
frobware: path.Join yields the same as a fmt.Sprintf and/or direct string concat would. Dunno. Can change if desired.

test/e2e/http2_disable_test.go
frobware: These functions no longer call Fatal - they return a wrapped error. We carry t for logging through t.Logf().

PR #404, opened 2020-05-21
pkg/operator/controller/status: AsExpected reason for Available and Progressing

wking:
We've used a message without a reason for `Progressing=False` since 7b6339ed6b (#229) and for Availabe=True since 28d343900d (#201).  But if we feel like we have a message we want to set to help humans understand the condition, we should be setting a reason string for machines too.  `AsExpected` follows [the library-go precedent][1].

[1]: https://github.com/openshift/library-go/blob/94c59dec54be25c8527e51e8c0a885712aeb01b5/pkg/operator/status/condition.go#L67


Miciah:
/lgtm
Let me know if you want this in 4.5 (we'll need to link a Bugzilla report if you do); otherwise it will merge in 4.6.

PR #406, opened 2020-05-28
Add run-local steps to hacking doc

sgreene570:
Adds brief directions for the `make run-local` (`hack/run-local.sh`) command to the `HACKING.md` document.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -15,7 +15,7 @@ $ make build
 * An [OpenShift cluster](https://github.com/openshift/installer)
 * An admin-scoped `KUBECONFIG` for the cluster.
 
-#### Building Locally
+#### Building Locally & Deploying to the Cluster
 
 To build the operator on your local machine and deploy it to the cluster, first uninstall the existing operator and all its managed components:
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -35,6 +35,28 @@ Follow the instructions to install the operator, e.g.:
...
+To run the operator binary in the cluster from your local machine (as opposed to on the cluster in a pod):
+
+```
+$ make run-local
+```
+
+Note, to re-scale the operator on the cluster after local testing is complete, run:
+
+```
+$ oc scale --replicas 1 -n openshift-ingress-operator deployments ingress-operator
```
Miciah, line 57: `make run-local` also scales CVO to 0 replicas; should we suggest `oc scale --replicas 1 -n openshift-cluster-version deployments/cluster-version-operator` too, or recommend it instead (once it is running, CVO will scale the ingress operator up too)?

sgreene570: I think providing only the command to scale the CVO back up would be sufficient. On that note, should the `Building Locally & Deploying to the Cluster` section provide a similar suggestion for scaling the CVO, since `make uninstall` scales the CVO to 0 replicas as well?

Miciah: > On that note, should the `Building Locally & Deploying to the Cluster` section provide a similar suggestion for scaling the CVO, since `make uninstall` scales the CVO to 0 replicas as well?

Yeah, that would be reasonable.

sgreene570: Fixed, thanks!!

PR #407, opened 2020-06-02
Bug 1842742: nodePortServiceChanged: Fix for session affinity

Miciah:
When comparing services to determine whether an update is required, treat implicit and explicit default values as equal.

Before this commit, the update logic would keep trying to revert the default value that the API set for the service's session affinity.

* `pkg/operator/controller/ingress/nodeport_service.go` (`nodePortServiceChanged`): Use the new `cmpServiceAffinity` function so that two services are considered equal if the only difference between them is that one does not specify the session affinity and the other service has the default value.
(`cmpServiceAffinity`): New function.
* `pkg/operator/controller/ingress/nodeport_service_test.go` (`TestNodePortServiceChanged`): Verify that `nodePortServiceChanged` returns false if the only mutation is that the session affinity has been updated from empty to the default value.


sgreene570:
looks good to me. :ship: 
/lgtm

PR #408, opened 2020-06-10
controller: Make ensure conditions consistent

sgreene570:
Switches all ensure<thing> functions in pkg/operator/controller/ingress
to use a have<thing> boolean rather than manually checking if
<thing> != nil for consistency. 

Also makes sure that failed update branches in each `ensure<thing>` function return the current object, if available, as mentioned in https://github.com/openshift/cluster-dns-operator/pull/176.


Miciah:
/lgtm
Thanks!

Miciah:
e2e-aws-operator failed on `TestRouteAdmissionPolicy`; the failure is probably not caused by the changes in this PR (see #400).

PR #410, opened 2020-06-10
Implement HTTP Forwarded header policy API

Miciah:
This commit resolves [NE-318](https://issues.redhat.com/browse/NE-318).

* `pkg/operator/controller/ingress/deployment.go` (`RouterForwardedHeadersPolicy`): New constant with the name of the related environment variable.
(`desiredRouterDeployment`): Set the `ROUTER_SET_FORWARDED_HEADERS` environment variable as appropriate.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Verify that `spec.httpHeaders.forwardedHeaderPolicy` has the expected effect.
* `test/e2e/forwarded_header_policy_test.go`: New file.
(`buildEchoPod`, `buildEchoService`, `buildRoute`, `testRouteHeaders`): New helper functions.
(`testPodCount`): New helper variable for `testRouteHeaders`.
(`TestForwardedHeaderPolicyAppend`, `TestForwardedHeaderPolicyReplace`, `TestForwardedHeaderPolicyNever`, `TestForwardedHeaderPolicyIfNone`): New tests.

---

Implements the following enhancement: https://github.com/openshift/enhancements/pull/371

Corresponding router changes: https://github.com/openshift/router/pull/134

Corresponding API changes: https://github.com/openshift/api/pull/688

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1279,6 +1279,30 @@ func waitForDeploymentComplete(t *testing.T, cl client.Client, deployment *appsv
 	return nil
 }
 
+// Wait for the provided deployment to have the specified environment variable set.
+func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.Deployment, timeout time.Duration, name, value string) error {
+	t.Helper()
+	deploymentName := types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name}
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		deployment := &appsv1.Deployment{}
+		if err := kclient.Get(context.TODO(), deploymentName, deployment); err != nil {
+			t.Logf("error getting deployment %s: %v", name, err)
+			return false, nil
+		}
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if container.Name == "router" {
+				for _, v := range container.Env {
+					if v.Name == name {
+						return v.Value == value, nil
+					}
+				}
+			}
+		}
+		return false, nil
+	})
+	return err
+}
+
 func clusterOperatorConditionMap(conditions ...configv1.ClusterOperatorStatusCondition) map[string]string {
 	conds := map[string]string{}
 	for _, cond := range conditions {
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -543,6 +543,7 @@ func TestInternalLoadBalancer(t *testing.T) {
 	supportedPlatforms := map[configv1.PlatformType]struct{}{
 		configv1.AWSPlatformType:      {},
 		configv1.AzurePlatformType:    {},
+		configv1.GCPPlatformType:      {},
 		configv1.IBMCloudPlatformType: {},
 	}
 	if _, supported := supportedPlatforms[platform]; !supported {
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e:
-	$(GO) test -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	$(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
 
 .PHONY: clean
 clean:
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -35,6 +35,8 @@ import (
 const (
 	WildcardRouteAdmissionPolicy = "ROUTER_ALLOW_WILDCARD_ROUTES"
 
+	RouterForwardedHeadersPolicy = "ROUTER_SET_FORWARDED_HEADERS"
+
 	RouterLogLevelEnvName       = "ROUTER_LOG_LEVEL"
 	RouterSyslogAddressEnvName  = "ROUTER_SYSLOG_ADDRESS"
 	RouterSyslogFormatEnvName   = "ROUTER_SYSLOG_FORMAT"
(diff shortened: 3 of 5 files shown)


Miciah:
Failed on `TestForwardedHeaderPolicy`, which is expected till https://github.com/openshift/router/pull/134 merges.

test/e2e/operator_test.go
```diff
@@ -1076,6 +1077,303 @@ func TestContainerLogging(t *testing.T) {
...
+	if err := waitForDeploymentComplete(kclient, deployment, 3*time.Minute); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+	testRoute(0, nil)
+	testRoute(2, []string{"x-forwarded-for:foo", "x-forwarded-for:bar"})
+
+	// Verify that we get the expected behavior if we set the policy to
+	// "if-none".  We should always receive at least 1 X-Forwarded-For
+	// header, and if the request specifies more than 1, we should receive
+	// the number that are in the request.
```
sgreene570, line 1354:
```suggestion
	// the number of headers specified in the request.
```

Does that make more sense, or am I missing something?

Miciah: I've updated the wording to be more explicit.

test/e2e/operator_test.go
```diff
@@ -1163,6 +1461,60 @@ func waitForAvailableReplicas(cl client.Client, ic *operatorv1.IngressController
...
+				}
+			}
+		}
+		return false, nil
+	})
+	return err
+}
+
+// Wait for the provided deployment to complete its rollout.
+func waitForDeploymentComplete(cl client.Client, deployment *appsv1.Deployment, timeout time.Duration) error {
```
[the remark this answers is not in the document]
Miciah: #400 merged, and I've removed this definition from this PR.

Miciah:
Rebased.

test/e2e/operator_test.go
```diff
@@ -1075,6 +1076,295 @@ func TestContainerLogging(t *testing.T) {
...
+				line := scanner.Text()
+				if strings.HasPrefix(line, "x-forwarded-for:") {
+					numMatches++
+					t.Logf("found match %d of %d expected: %s", numMatches, expectedMatches, line)
+					if numMatches > expectedMatches {
+						return false, fmt.Errorf("got %d x-forwarded-for headers, expected %d", numMatches, expectedMatches)
+					}
+				}
+			}
+			return numMatches == expectedMatches, nil
```
sgreene570, line 1268: Line 1263 checks for the case where `numMatches > expectedMatches`. Perhaps before the return on line 1268, we should check for `numMatches > 0 && numMatches < expectedMatches` and return an error if that condition is true. This way, for example, the polling loop would half if only 1 x-forwarded-for header was found in the curl response when 2 were expected, and the test failure message would be more verbose. Thoughts? I may be overthinking this, but since you're calling `testRoute` several times I figured it would be worth mentioning.

sgreene570:
This PR depends on https://github.com/openshift/router/pull/134 for CI to pass, correct?

Miciah:
> This PR depends on [openshift/router#134](https://github.com/openshift/router/pull/134) for CI to pass, correct?

You're right; I mixed up my PRs.  I have one PR that depends on an unmerged API change and another PR that depends on an unmerged router change...

test/e2e/operator_test.go
Miciah: That makes sense.  It is unlikely that the logs will have partial headers, so returning an error if 0 &lt; `numMatches` &lt; `expectedMatches` should safe to do.

Miciah:
https://github.com/openshift/router/pull/134 merged, so tests now should pass.
/test e2e-aws-operator

Miciah:
`TestForwardedHeaderPolicy` is timing out in some CI runs, logging "waiting to start: ContainerCreating" as it polls the test client pod container.  However, I launched a cluster and manually ran the test a dozen times without getting any failures.  The test client pod uses the router image, so the image is already pulled, and I previously raised poll loop's timeout to 2 minutes.  I believe it is unreasonable that a simple test pod that uses a pre-pulled image should take over 2 minutes to start, so I am hoping that the failing CI runs were flakes.

The previous run of e2e-aws-operator did succeed, so let's see if we can get another successful run in CI.
/test e2e-aws-operator

Miciah:
On second thought, the router image is only necessarily pre-pulled on the 2 worker nodes where the router is scheduled, and it may be possible for the test client pod to be scheduled to the 3rd worker node, which may not have previously pulled the image.  Maybe it would make sense to increase the timeout from 2 minutes, but we cannot raise it much before we hit the 10-minute test timeout.

Miciah:
Hm, now `TestForwardedHeaderPolicy` is timing out (again).

/test e2e-aws-operator

Miciah:
Split `TestForwardedHeaderPolicy` into multiple tests to avoid hitting the 10-minute timeout.

Miciah:
Provisioning failure.
/test e2e-aws-operator

Miciah:
Both `TestForwardedHeaderPolicyAppend` and `TestRouteHTTP2EnableAndDisableIngressController` timed out.  I'll increase the timeout for `TestForwardedHeaderPolicyAppend` and related tests, but it looks like the cluster was not in good order.

Miciah:
No build log output?!
/test e2e-aws-operator

Miciah:
Last e2e-aws failure looks related to [BZ1857928 – [sig-operator] an end user use OLM can subscribe to the cockroachdb operator](https://bugzilla.redhat.com/show_bug.cgi?id=1857928).

e2e-aws-operator is reaching the 10-minute timeout.  I'll try pushing a change to increase that timeout value.

Miciah:
Seeing "an end user can use OLM can subscribe to the operator" fail again in e2e-aws, and seeing numerous timeouts in e2e-aws-operator that look like a generally unstable cluster.
/test e2e-aws
/test e2e-aws-operator

Miciah:
Test container setup failed on an API failure.
/test e2e-aws

PR #412, opened 2020-06-15
Fix ensureWildcardDNSRecord & ensureLoadBalancerService

sgreene570:
Add a `wantFoo` boolean to `ensureWildcardDNSRecord` &
`ensureLoadBalancerService` for consistency. Fixes a bug
accidentally introduced in #408.

Also fixes a small comment typo for `desiredWildcardDNSRecord`, as well as updates the necessary unit tests.

### DIFF
--- a/pkg/operator/controller/ingress/dns.go
+++ b/pkg/operator/controller/ingress/dns.go
@@ -34,20 +34,20 @@ func (r *reconciler) ensureWildcardDNSRecord(ic *operatorv1.IngressController, s
 		return false, nil, nil
 	}
 
-	desired := desiredWildcardRecord(ic, service)
+	wantWC, desired := desiredWildcardDNSRecord(ic, service)
 	haveWC, current, err := r.currentWildcardDNSRecord(ic)
 	if err != nil {
 		return false, nil, err
 	}
 
 	switch {
-	case !haveWC:
+	case wantWC && !haveWC:
 		if err := r.client.Create(context.TODO(), desired); err != nil {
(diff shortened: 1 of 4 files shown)


sgreene570:
Shrinks the `vendor` dir by ~.5 mb, if anyone was curious.

pkg/operator/controller/ingress/dns.go
```diff
@@ -56,10 +56,10 @@ func (r *reconciler) ensureWildcardDNSRecord(ic *operatorv1.IngressController, s
 		}
 	}
 
-	return true, current, nil
+	return haveWC, current, nil
 }
 
-// ensureWildcardDNSRecord will return any necessary wildcard DNS records for the
+// desiredWildcardDNSRecord will return any necessary wildcard DNS records for the
```
Miciah, line 62: `desiredWildcardDNSRecord` → `desiredWildcardRecord`.

sgreene570: Whoops. This typo brings up an inconsistency, however.
<bikeshed>
Should `desiredWildCardRecord` actually be called `desiredWildCardDNSRecord` since the other functions in `dns.go` follow that pattern?
</bikeshed>

Miciah: Ah, good point.  I wouldn't object to renaming `desiredWildcardRecord` to `desiredWildcardDNSRecord`.  Bonus: That would save you the trouble of changing the godoc again!

sgreene570: fixed! :smile:

PR #414, opened 2020-06-15
controller/ingress: Remove excess API calls for ingress

sgreene570:
Removes extra API get calls in pkg/operator/controller/ingress
when an Ingress resource already exists and no update is performed.

### DIFF
--- a/pkg/operator/controller/ingress/monitoring.go
+++ b/pkg/operator/controller/ingress/monitoring.go
@@ -30,20 +30,21 @@ func (r *reconciler) ensureServiceMonitor(ic *operatorv1.IngressController, svc 
 
 	switch {
 	case !haveSM:
-		if created, err := r.createServiceMonitor(desired); err != nil {
+		if err := r.client.Create(context.TODO(), desired); err != nil {
 			return false, nil, fmt.Errorf("failed to create servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
-		} else if created {
-			log.Info("created servicemonitor", "namespace", desired.GetNamespace(), "name", desired.GetName())
 		}
+		log.Info("created servicemonitor", "namespace", desired.GetNamespace(), "name", desired.GetName())
+		return r.currentServiceMonitor(ic)
 	case haveSM:
 		if updated, err := r.updateServiceMonitor(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
 		} else if updated {
 			log.Info("updated servicemonitor", "namespace", desired.GetNamespace(), "name", desired.GetName())
+			return r.currentServiceMonitor(ic)
 		}
 	}
 
-	return r.currentServiceMonitor(ic)
+	return true, current, nil
 }
 
 // desiredServiceMonitor returns the desired servicemonitor for the given
@@ -118,15 +119,6 @@ func (r *reconciler) currentServiceMonitor(ic *operatorv1.IngressController) (bo
 		return false, nil, err
 	}
 	return true, sm, nil
-}
-
-// createServiceMonitor creates a servicemonitor.  Returns a Boolean indicating
-// whether the servicemonitor was created, and an error value.
-func (r *reconciler) createServiceMonitor(sm *unstructured.Unstructured) (bool, error) {
-	if err := r.client.Create(context.TODO(), sm); err != nil {
-		return false, err
-	}
-	return true, nil
 }
 
 // updateServiceMonitor updates a servicemonitor.  Returns a Boolean indicating
--- a/pkg/operator/controller/ingress/nodeport_service.go
+++ b/pkg/operator/controller/ingress/nodeport_service.go
@@ -36,26 +36,29 @@ func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, dep
 		return false, nil, nil
 	case !wantService && haveService:
 		if err := r.client.Delete(context.TODO(), current); err != nil {
-			if errors.IsNotFound(err) {
-				return false, nil, nil
+			if !errors.IsNotFound(err) {
+				return true, current, fmt.Errorf("failed to delete NodePort service: %v", err)
 			}
-			return true, current, fmt.Errorf("failed to delete NodePort service: %v", err)
+		} else {
+			log.Info("deleted NodePort service", "service", current)
 		}
-		log.Info("deleted NodePort service", "service", current)
+		return false, nil, nil
 	case wantService && !haveService:
 		if err := r.client.Create(context.TODO(), desired); err != nil {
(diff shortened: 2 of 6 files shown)


pkg/operator/controller/ingress/nodeport_service.go
```diff
@@ -47,15 +47,17 @@ func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, dep
...
 		if updated, err := r.updateNodePortService(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update NodePort service: %v", err)
 		} else if updated {
 			log.Info("updated NodePort service", "service", desired)
+			return r.currentNodePortService(ic)
 		}
 	}
 
-	return r.currentNodePortService(ic)
+	return true, current, nil
```
Miciah, line 60: We should still `return r.currentNodePortService(ic)` or possibly `return false, nil, nil` after line 44 under the `!wantService && haveService` case where we have just deleted the service.

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -47,15 +47,17 @@ func (r *reconciler) ensureRouterPodDisruptionBudget(ic *operatorv1.IngressContr
...
 		if updated, err := r.updateRouterPodDisruptionBudget(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update pod disruption budget: %v", err)
 		} else if updated {
 			log.Info("updated pod disruption budget", "poddisruptionbudget", desired)
+			return r.currentRouterPodDisruptionBudget(ic)
 		}
 	}
 
-	return r.currentRouterPodDisruptionBudget(ic)
+	return true, current, nil
```
Miciah, line 60: The `!wantPDB && havePDB` case needs a return statement.

pkg/operator/controller/ingress/rsyslog_configmap.go
```diff
@@ -52,16 +52,18 @@ func (r *reconciler) ensureRsyslogConfigMap(ic *operatorv1.IngressController, de
...
 		if updated, err := r.updateRsyslogConfigMap(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update configmap: %v", err)
 		} else if updated {
 			log.Info("updated configmap", "configmap", desired)
+			return r.currentRsyslogConfigMap(ic)
 		}
 	}
 
-	return r.currentRsyslogConfigMap(ic)
+	return true, current, nil
```
Miciah, line 66: The `!wantCM && haveCM` case needs a return statement.

pkg/operator/controller/ingress/serviceca_configmap.go
```diff
@@ -41,15 +41,17 @@ func (r *reconciler) ensureServiceCAConfigMap() (bool, *corev1.ConfigMap, error)
...
 		if updated, err := r.updateServiceCAConfigMap(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update configmap: %v", err)
 		} else if updated {
 			log.Info("updated configmap", "configmap", desired)
+			return r.currentServiceCAConfigMap()
 		}
 	}
 
-	return r.currentServiceCAConfigMap()
+	return true, current, nil
```
Miciah, line 54: The `!wantCM && haveCM` case needs a return statement.

pkg/operator/controller/ingress/nodeport_service.go
sgreene570: whoops, totally missed that. Thanks!

pkg/operator/controller/ingress/rsyslog_configmap.go
```diff
@@ -43,25 +43,30 @@ func (r *reconciler) ensureRsyslogConfigMap(ic *operatorv1.IngressController, de
 		return false, nil, nil
 	case !wantCM && haveCM:
 		if deleted, err := r.deleteRsyslogConfigMap(current); err != nil {
-			return true, current, fmt.Errorf("failed to delete configmap: %v", err)
+			if !errors.IsNotFound(err) {
```
Miciah, line 46: `deleteRsyslogConfigMap` returns `false, nil` if `errors.IsNotFound(err)`.  It might be best to get rid of `deleteRsyslogConfigMap` and inline its logic here for consistency.  (The reason that the deletion logic is inconsistent is that we—mostly I—have waffled back and forth on using a separate method for `deleteFoo` versus inlining the logic.)

sgreene570: Sure. I agree that inline deletion makes more sense here.

pkg/operator/controller/ingress/monitoring.go
```diff
@@ -34,16 +34,18 @@ func (r *reconciler) ensureServiceMonitor(ic *operatorv1.IngressController, svc
 			return false, nil, fmt.Errorf("failed to create servicemonitor %s/%s: %v", desired.GetNamespace(), desired.GetName(), err)
 		} else if created {
 			log.Info("created servicemonitor", "namespace", desired.GetNamespace(), "name", desired.GetName())
+			return r.currentServiceMonitor(ic)
 		}
```
Miciah, line 38: If `err == nil && !created`, then control continues on to `return true, current, nil` at the end of the method, and `current` is still nil.  Simplest fix would be to move the return statement on line 37 down a line (outside the `else if`).  I wouldn't object if you decided to delete `createServiceMonitor` and inline its logic here.

pkg/operator/controller/ingress/rsyslog_configmap.go
```diff
@@ -42,26 +42,31 @@ func (r *reconciler) ensureRsyslogConfigMap(ic *operatorv1.IngressController, de
...
 			log.Info("deleted configmap", "configmap", current)
 		}
+		return false, nil, nil
 	case wantCM && !haveCM:
 		if created, err := r.createRsyslogConfigMap(desired); err != nil {
 			return false, nil, fmt.Errorf("failed to create configmap: %v", err)
 		} else if created {
 			log.Info("created configmap", "configmap", desired)
+			return r.currentRsyslogConfigMap(ic)
 		}
```
Miciah, line 59: Similarly to the case I commented on in `ensureServiceMonitor`, if `err == nil && !created`, control flows down to the `return true, current, nil` statement below, and `current` is nil.

pkg/operator/controller/ingress/monitoring.go
sgreene570: Fixed, went with inline logic to save some lines.

sgreene570: Just to be clear, previously `createServiceMonitor` was only capable of returning `false, err` or `true, nil`, so the case of `err == nil && !created` would not have been possible. However, that is in no way obvious, so using inline logic is definitely the way to go. Thanks!

Miciah:
Thanks!
/lgtm

PR #419, opened 2020-06-25
Bug 1850813: Delete Router Deployment Only Upon DNSRecord Deletion

danehans:
`pkg/operator/controller/ingress/controller.go`: Updates `ensureIngressDeleted()` to delete the router deployment only when the `dnsrecord` object has been deleted. The router deployment is the owning object of the LB Service which is used by the AWS DNS provider to find the hosted zoned ID for record deletion (when the record is not cached).

/assign @Miciah @knobunc

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -547,19 +547,24 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 	if err := r.deleteWildcardDNSRecord(ingress); err != nil {
 		errs = append(errs, fmt.Errorf("failed to delete wildcard dnsrecord for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 	}
-	if haveRec, _, err := r.currentWildcardDNSRecord(ingress); err != nil {
+	haveRec, _, err := r.currentWildcardDNSRecord(ingress)
+	switch {
+	case err != nil:
 		errs = append(errs, fmt.Errorf("failed to get current wildcard dnsrecord for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
-	} else if haveRec {
+	case haveRec:
 		errs = append(errs, fmt.Errorf("wildcard dnsrecord exists for ingress %s/%s", ingress.Namespace, ingress.Name))
-	}
-
-	if err := r.ensureRouterDeleted(ingress); err != nil {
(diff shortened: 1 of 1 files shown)


danehans:
@Miciah this is a quick and simple fix but not very efficient since an additional reconcile loop will be triggered if the record is not immediately deleted. A more elegant solution is to update `dnsrecords.status` to include an optional hosted zone ID field that can be populated by the AWS provider during record creation. This will allow the operator to use this status field for a record delete operation instead of doing a hosted zone lookup using the router's LB service. Thoughts?

Miciah:
Let's see if I have this right:

* The `IngressController` owns the `DNSRecord`.
* The `Deployment` owns the `Service`, which in a sense owns the ELB.
* Currently, the `DNSRecord` cannot be deleted unless the ELB is present.
* So until the `DNSRecord` is deleted, we need to block deletion of the `Deployment` in order to block deletion of the `Service` in order to block deletion of the ELB in order to be able to delete the `DNSRecord`.

(That's the short-term fix; as you said, the long-term fix is to enrich the `DNSRecord` status with the information that the AWS DNS provider currently gets from the ELB.)

Does that sum it up?

danehans:
@Miciah your https://github.com/openshift/cluster-ingress-operator/pull/419#issuecomment-649164699 perfectly articulates the situation.

danehans:
`level=error msg="Error: Unable to find matching route for Route Table (rtb-03fc8fbd9bd9d4f9e) and destination CIDR block (0.0.0.0/0)."`

/test e2e-aws-upgrade

danehans:
```
error: some steps failed:
  * could not run steps: step [release:initial] failed: release "release-initial" failed: pod was deleted while ci-operator step was waiting for it
time="2020-06-25T00:38:48Z" level=info msg="Reporting job state \"failed\" with reason \"executing_graph:step_failed:assembling_release:creating_release:running_pod\""
```
/test e2e-aws

Miciah:
Thanks!
/lgtm

danehans:
```
fail [github.com/openshift/origin/test/e2e/upgrade/upgrade.go:138]: during upgrade to registry.svc.ci.openshift.org/ci-op-q0n8374d/release@sha256:921898453084708fb71eb58c467098f4c8324f37491badfbe5b6f4e7324ecdad
Unexpected error:
    <*errors.errorString | 0xc0016a4130>: {
        s: "Cluster did not acknowledge request to upgrade in a reasonable time: timed out waiting for the condition",
    }
    Cluster did not acknowledge request to upgrade in a reasonable time: timed out waiting for the condition
occurred
```
/test e2e-aws-upgrade

PR #420, opened 2020-06-25
Bug 1830271: status: Replace "DeploymentDegraded" condition

Miciah:
Delete the `DeploymentDegraded` status condition, add `DeploymentAvailable`, `DeploymentReplicasMinAvailable`, and `DeploymentReplicasAllAvailable` status conditions, and use the new conditions to compute the ingresscontroller's `Degraded` status condition.

The old `DeploymentDegraded` status condition was based on the deployment's `Available` status condition, but the deployment may report `Available=True` before the deployment's minimum availability is met, so relying on the deployment's `Available` condition alone is insufficient to determine if the ingresscontroller's deployment is degraded.

* `pkg/operator/controller/ingress/controller.go` (`IngressControllerDeploymentDegradedConditionType`): Delete the `DeploymentDegraded` condition type.
(`IngressControllerDeploymentAvailableConditionType`, `IngressControllerDeploymentReplicasMinAvailableConditionType`, `IngressControllerDeploymentReplicasAllAvailableConditionType`): New constants for the `DeploymentAvailable`, `DeploymentReplicasMinAvailable`, and `DeploymentReplicasAllAvailable` condition types.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Delete use of `computeDeploymentDegradedCondition`.  Add calls to `computeDeploymentAvailableCondition`, `computeDeploymentReplicasMinAvailableCondition`, and `computeDeploymentReplicasAllAvailableCondition`.
(`computeDeploymentDegradedCondition`): Rename...
(`computeDeploymentAvailableCondition`): ...to this.  Add a note that the deployment may report `Available=True` without meeting minimum availability. Replace the `DeploymentDegraded` condition type with `DeploymentAvailable`, and invert the condition status.
(`computeDeploymentReplicasMinAvailableCondition`): New function.  Compute the `DeploymentReplicasMinAvailable` condition from the deployment's reported available replicas and maximum unavailable replicas.
(`computeDeploymentReplicasAllAvailableCondition`): New function.  Compute the `DeploymentReplicasAllAvailable` condition from the deployment's reported available replicas.
(`computeIngressDegradedCondition`): Replace `DeploymentDegraded` with `DeploymentAvailable` with the expected status inverted.  Add `DeploymentReplicasMinAvailable` and `DeploymentReplicasAllAvailable` with status `True` to the set of expected conditions.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressDegradedCondition`): Update to check `DeploymentAvailable`, `DeploymentReplicasMinAvailable`, and `DeploymentReplicasAllAvailable` instead of `DeploymentDegraded`.
(`TestComputeDeploymentDegradedCondition`): Rename...
(`TestComputeDeploymentAvailableCondition`): ...to this.  Replace `DeploymentDegraded` with `DeploymentAvailable` with the expected status inverted.
(`TestComputeDeploymentReplicasMinAvailableCondition`, `TestComputeDeploymentReplicasAllAvailableCondition`): New tests.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -21,6 +21,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // clock is to enable unit testing
@@ -41,7 +42,9 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.Selector = selector.String()
 	updated.Status.TLSProfile = computeIngressTLSProfile(ic.Status.TLSProfile, deployment)
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(deployment))
-	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentDegradedCondition(deployment))
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDNSStatus(ic, wildcardRecord, dnsConfig)...)
 	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions)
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -188,10 +219,15 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition)
...
-			condition:   IngressControllerDeploymentDegradedConditionType,
-			status:      operatorv1.ConditionFalse,
+			condition:   IngressControllerDeploymentAvailableConditionType,
+			status:      operatorv1.ConditionTrue,
 			gracePeriod: time.Second * 30,
 		},
+		{
+			condition:   IngressControllerDeploymentReplicasAvailableConditionType,
+			status:      operatorv1.ConditionTrue,
+			gracePeriod: time.Second * 60,
```
Miciah, line 229: I'm contemplating adding another condition, `DeploymentMinReplicasAvailable`, which would be `True` iff `deployment.Status.AvailableReplicas < *deployment.Spec.Replicas - maxUnavailable`, where `maxUnavailable` would be computed using `deployment.Spec.Strategy.RollingUpdate.MaxUnavailable` (i.e., the ingresscontroller's `DeploymentMinReplicasAvailable` condition would report what the deployment's `Available` condition is supposed to report).  `DeploymentMinReplicasAvailable` would be included here in `expectedConditions` with a short grace period because its status should remain `True` during upgrades.  The grace period for `DeploymentReplicasAvailable` should be longer to allow it to become temporarily `False` during upgrades without causing the ingresscontroller to set `Degraded=True`.

sgreene570: Would `DeploymentMinReplicasAvailable` be true if `deployment.Status.AvailableReplicas` were `0` and `*deployment.Spec.Replicas - maxUnavailable` was > 0, or is that not possible?

Miciah: > Would `DeploymentMinReplicasAvailable` be true if `deployment.Status.AvailableReplicas` were `0` and `*deployment.Spec.Replicas - maxUnavailable` was > 0, or is that not possible?

In that case, `DeploymentMinReplicasAvailable` would be `False`, as would be `DeploymentAvailable` and `DeploymentAvailable`.

Miciah: It might make sense to rename `DeploymentReplicasAvailable` to `DeploymentAllReplicasAvailable` as well.

sgreene570: > It might make sense to rename `DeploymentReplicasAvailable` to `DeploymentAllReplicasAvailable` as well.

I am in favor of this, since it helps remove some of the ambiguity.

PR #417, opened 2020-06-23
Move DNS provider initialization to DNS controller

Miciah:
* `cmd/ingress-operator/start.go` (`NewStartCommand`): Delete initialization of the DNS provider.
(`cloudCredentialsSecretName`, `createDNSProvider`, `getPlatformStatus`): Move from here...
* `pkg/operator/controller/dns/controller.go:` ...to here.
(`New`): Delete parameter for DNS provider; the controller now initializes it.  Add config parameter.  Add watches on the `DNS` config and `Infrastructure` config objects.  Use `ToDNSRecords` mapper function to reconcile all `DNSRecord` objects when a config object changes.
(`Config`): New type to hold the namespace and operator release version, which are needed for initializing the DNS provider.
(`reconciler`): Add `Config` and `infraConfig` fields.
(`Reconcile`): Get the cluster `Infrastructure` config object, and initialize or re-initialize the DNS provider if it was not already initialized or if the `Infrastructure` config has changed.
(`ToDNSRecords`): New method.  Return reconcilation requests for all `DNSRecords`.
* `pkg/operator/operator.go` (`New`): Update initialization of the DNS controller to pass the controller's config object instead of the DNS provider object.
* `manifests/00-cluster-role.yaml`: Allow the operator to list and watch the `dnses` and `infrastructures` config resources.
* `pkg/manifests/bindata.go`: Regenerate.

PR #425, opened 2020-07-09
Bug 1854383: dns: Reread cloud credentials secret if it changes

Miciah:
Watch the cloud credentials secret and recreate the DNS provider if the credentials change.

* `pkg/operator/controller/dns/controller.go` (`New`): Watch the cloud credentials secret and trigger reconciliation if it changes.
(`reconciler`): Add `cloudCredentials` field.
(`Reconcile`): Refactor DNS provider create logic from here...
(`createDNSProviderIfNeeded`): ...to here.  Check if the cloud credentials secret data changed in addition to checking the infrastructure status to determine whether the DNS provider needs to be recreated.

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -37,8 +37,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	runtimecontroller "sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/dns/controller.go
```diff
@@ -191,6 +193,53 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	platformStatus, err := getPlatformStatus(r.client, infraConfig)
+	if err != nil {
+		return fmt.Errorf("failed to get platform status: %v", err)
+	}
+
+	creds := &corev1.Secret{}
+	switch platformStatus.Type {
+	case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.GCPPlatformType:
+		name := types.NamespacedName{Namespace: r.Config.Namespace, Name: cloudCredentialsSecretName}
+		if err := r.cache.Get(context.TODO(), name, creds); err != nil {
```
sgreene570, line 218: I noticed before these changes `creds` was being populated by a call to `r.client.Get` instead of `r.cache.Get`. Just wanted to verify that this is intentional and won't cause any defects.

Miciah: Yeah, it is intentional.  Want to avoid incurring the API call on every reconciliation, and it may avoid cache consistency issues between the watch on the infrastructure resource that triggers the reconciliation and the reconciliation logic here that reads the infrastructure object.  It might have made sense to make this change in #417, which moved this logic from the operator startup (where we don't yet have the cache) to the reconciliation logic (where we do).  It definitely makes sense to use the cache if we're using the value on every reconciliation.

PR #431, opened 2020-07-29
Omit ELB EP in TestIngressControllerCustomEndpoints

Miciah:
Don't set the endpoint for the AWS ELB API in `TestIngressControllerCustomEndpoints`.  Unlike the Route 53 and tagging APIs, the ELB API needs a region-specific endpoint, and trying to use the us-west-2 endpoint when running in the us-east-1 region fails.

* `test/e2e/operator_test.go` (`TestIngressControllerCustomEndpoints`): Don't set a custom endpoint for the AWS ELB API.

----

@danehans


danehans:
@Miciah I need to look into this issue in more detail. I'll tag in the meantime so we can get CI unstuck.

/lgtm

PR #435, opened 2020-08-01
Changing default router deployment number to 3

rhdedgar:
This PR is in response to OSD-3918, which outlines increasing the router deployment to match the default number of infra nodes. It also provides an an increase in capacity and stability, and a slightly more intuitive ELB Status from the AWS console.

e.g. By having only 2 router pods and 3 infra nodes, one infra node continuously fails health-checks behind the *.app.clustername.openshiftapps.com ELB.

Please let me know if there's a more appropriate place to manage this configuration.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -157,7 +157,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	volumes := deployment.Spec.Template.Spec.Volumes
 	routerVolumeMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
 
-	var desiredReplicas int32 = 2
+	var desiredReplicas int32 = 3
 	if ci.Spec.Replicas != nil {
 		desiredReplicas = *ci.Spec.Replicas
 	}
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -157,7 +157,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	volumes := deployment.Spec.Template.Spec.Volumes
 	routerVolumeMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
 
-	var desiredReplicas int32 = 2
+	var desiredReplicas int32 = 3
```
Miciah, line 160: `ensureDefaultIngressController` specifies `spec.replicas` when it creates the default ingresscontroller, so changing the default here is going to be ineffective as far as the default ingresscontroller is concerned.

`ensureDefaultIngressController` really choose a value for `spec.replicas` dynamically. The documented minimum number of worker nodes is 2, so we cannot necessarily go above that, and on clusters with more than 3 workers, it might be reasonable to scale up to, say, max(min(number of worker nodes, 5), floor(20% of worker nodes)).

Miciah:
This change probably warrants some design discussion, so I've created https://issues.redhat.com/browse/NE-361 to track computing the default ingresscontroller's `spec.replicas` value dynamically, and we'll discuss it at a future planning meeting.

abhinavdahiya:
One thing to note is that we define minimum compute footprint to be 2 replicas. That means that on that minimum we will have one not scheduled replica which might cause notavialable clusteroperator

rhdedgar:
Closing this pull request at least until https://issues.redhat.com/browse/NE-361 has been addressed.

PR #436, opened 2020-08-03
Bug 1862892: Reconcile openshift-ingress-router clusterrole on upgrades

frobware:
On upgrades we have never needed to handle changes to the
openshift-ingress-router cluster role. This broke 4.5->4.6 upgrades
because the router switched to watching endpointslices in 4.6 and, as
this permission was not granted to the role, it failed to start.

This change now ensures that the cluster role gets updated/reconciled
with the (bindata) values from assets/router/cluster-role.yaml. In
this particular upgrade case it will now add the endpointslices RBAC
rules allowing the router to start.

### DIFF
--- a/pkg/operator/controller/ingress/cluster-role.go
+++ b/pkg/operator/controller/ingress/cluster-role.go
@@ -1,0 +1,87 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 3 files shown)


Miciah:
Thanks!
/lgtm

PR #437, opened 2020-08-05
Bug 1866568: Removes RestMapper since DynamicRestMapper is now the pkg default

danehans:
https://github.com/kubernetes-sigs/controller-runtime/pull/554 merged to fix https://github.com/kubernetes-sigs/controller-runtime/issues/321. This PR updates the operator to use the controller-runtime `DynamicRestMapper` (default option) instead of using the internal implementation. The controller-runtime `DynamicRestMapper` was tested by:

1. Running the ingress operator locally.
2. Removing the ingresscontroller CRD and default ingresscontroller CR from the api server.
3. Recreating the the ingresscontroller CRD and the operator successfully created the default ingresscontroller CR.

/assign @Miciah @knobunc 
/cc @frobware @sgreene570


Miciah:
This is great!  Can we simplify further by deleting `pkg/util/restmapper.go` and using `apiutil.NewDynamicRESTMapper(cfg)` directly in `pkg/operator/operator.go`?

danehans:
> This is great! Can we simplify further by deleting pkg/util/restmapper.go and using apiutil.NewDynamicRESTMapper directly in pkg/operator/operator.go?

@Miciah done in commit `826c995`.

Miciah:
> > This is great! Can we simplify further by deleting pkg/util/restmapper.go and using apiutil.NewDynamicRESTMapper directly in pkg/operator/operator.go?
> 
> @Miciah done in commit `826c995`.

Thanks!  I was hoping we could collapse it down to the following:
```go
		MapperProvider: apiutil.NewDynamicRESTMapper,
```
But it looks like `apiutil.NewDynamicRESTMapper`'s variadic `opts` parameter causes a type mismatch (`cannot use apiutil.NewDynamicRESTMapper (value of type func(cfg *rest.Config, opts ...apiutil.DynamicRESTMapperOption) (meta.RESTMapper, error)) as func(c *rest.Config) (meta.RESTMapper, error) value in struct literal`).

However, it looks like the dynamic REST mapper is now the default mapper:  https://github.com/openshift/cluster-ingress-operator/blob/8aa1ce2f3fc2384f7c1688b8cc16d599f9ac89ea/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go#L384-L388

Any reason not to delete the designated initializer for the `MapperProvider` field entirely so we get the default?

danehans:
> Any reason not to delete the designated initializer for the MapperProvider field entirely so we get the default?

@Miciah I wasn't aware that the DynamicRestMapper is now the default. Thanks for looking into this. Commit `179dcbc` implements your feedback.

Miciah:
I love seeing all that code deleted!
/lgtm
Can we do the same cleanup in cluster-dns-operator?  It's using controller-runtime 0.6.0, which appears to have the fix.

danehans:
@Miciah thanks for making me aware that dns operator uses the same internal RestMapper implementation. https://github.com/openshift/cluster-dns-operator/pull/189 is the same PR as this for dns operator.

danehans:
```
/bin/bash: line 129: /tmp/artifacts/junit/junit_install_status.xml: No such file or directory
error: failed to execute wrapped command: exit status 1
```

^ is unrelated to this PR.

/test e2e-aws

danehans:
```
error: some steps failed:
  * could not run steps: step e2e-aws-upgrade failed: template e2e-aws-upgrade has required parameter BRANCH which is not defined
```

/test e2e-aws-upgrade

PR #438, opened 2020-08-06
Bug 1866575: Escape unique-id format

Miciah:
#### desiredRouterDeployment: Escape unique-id format

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Escape the unique-id header format string.

#### Add HTTP Unique-Id header E2E test

* `test/e2e/operator_test.go` (`TestUniqueIdHeader`): New test.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1462,6 +1462,128 @@ func TestNetworkLoadBalancer(t *testing.T) {
 	}
 }
 
+func TestUniqueIdHeader(t *testing.T) {
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "uniqueid"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	ic.Spec.HTTPHeaders = &operatorv1.IngressControllerHTTPHeaders{
+		UniqueId: operatorv1.IngressControllerHTTPUniqueIdHeaderPolicy{Name: "x-unique-id"},
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+	conditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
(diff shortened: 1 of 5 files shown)


test/e2e/operator_test.go
```diff
@@ -1462,6 +1463,192 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+	echoPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				"app": "unique-id-echo",
+			},
+			Name:      "unique-id-echo",
+			Namespace: deployment.Namespace,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
```
sgreene570, line 1505: I've noticed more than one test in this repo at this point make use of this "echo" container spec.
I believe the same could be say about the curl client container.
Since the e2e test files in this repo keep getting longer and longer, I think it would be wise to try and re-use common test container specs (and any other common functionality, really) within the repo. Thoughts? Definitely doesn't need to happen in this PR, so perhaps we should open a Jira ticket to track this improvement?

test/e2e/operator_test.go
```diff
@@ -1462,6 +1463,192 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+		}
+		scanner := bufio.NewScanner(readCloser)
+		defer func() {
+			if err := readCloser.Close(); err != nil {
+				t.Errorf("failed to close reader for pod %s: %v", clientPod.Name, err)
+			}
+		}()
+		for scanner.Scan() {
+			line := scanner.Text()
+			if strings.HasPrefix(line, "x-unique-id:") {
```
sgreene570, line 1640: Would it be worth enhancing this test to check 2 different requests so the uniqueness of the unique-id header could be verified? I may be overthinking this, but it seems that there would be some benefit in verifying the default unique-id header format string works as intended (as far as the unique part goes). Thoughts?

test/e2e/operator_test.go
Miciah: Latest push does some refactoring to reduce duplicate (still more could be done though).

test/e2e/operator_test.go
Miciah: Done.  Thanks for the suggestion!

test/e2e/operator_test.go
sgreene570: Great, thanks for that!

sgreene570:
Gave this another read. Looks great!
/lgtm

PR #439, opened 2020-08-17
Bug 1869387: Adds CNAME support to AWS DNS

danehans:
Based on guidance from AWS [docs](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-r53.html), update AWS DNS to use the CNAME record type for the AWS GovCloud partition. All other regions continue to use the Alias record type.

/assign @knobunc @Miciah 
/cc @frobware @sgreene570 

xref: https://github.com/openshift/installer/pull/4043


sgreene570:
Looks great!
/lgtm

PR #450, opened 2020-09-01
Implement client certificate parameters

Miciah:
Implement APIs for configuration related to client certificates, which can be used to enable mutual TLS.

* `pkg/operator/controller/crl/controller.go`: New file.
(`controllerName`, `clientCAConfigmapIndexFieldName`): New constants.
(`log`): New variable.
(`reconciler`): New type.  Store the state of the controller.
(`New`): New function.  Return a controller that manages configmaps with certificate revocation lists.  Create an index to map client CA configmap names to ingresscontrollers.  Set up watches on ingresscontrollers and configmaps.  Enqueue a reconciliation request for an ingresscontroller when it has a client CA configmap and the ingresscontroller or the configmap is created, updated, or deleted.  Use the `configmapToIngressController` method to map configmaps to ingresscontrollers, and use the `configmapIsInUse`, `hasConfigmap`, and `configmapReferenceChanged` method to filter watch events.
(`ingressControllersWithConfigmap`): New method.  Use the index to generate a list of the ingresscontrollers that reference the given client CA configmap, and return the list.
(`configmapToIngressController`): New method.  Use `ingressControllersWithConfigmap` to generate a list of reconciliation requests, one for each ingresscontroller that references the given configmap, and return this list.
(`configmapIsInUse`): New method.  Return a Boolean value indicating if the given client CA configmap is in use by some ingresscontroller, using the `ingressControllersWithConfigmap` method.
(`hasConfigmap`): New method.  Return a Boolean value indicating if a client CA configmap exists for the given ingresscontroller.
(`configmapReferenceChanged`): New method.  Return a Boolean value indicating if the client CA configmap reference for the given ingresscontroller has changed.
(`Reconcile`): New method.  Reconcile an ingresscontroller by ensuring that it has an associated client CA CRL configmap if the ingresscontroller has any client CA certificates that specify CRL distribution points.
* `pkg/operator/controller/crl/crl_configmap.go`: New file.
(`authorityKeyIdentifier`): New type.  Store the authority key identifier for a certificate.
(`authorityKeyIdentifierOID`): New variable with the ASN.1 object identifier for the authority key identifier X509v3 extension.
(`ensureCRLConfigmap`): New method.  Ensure the client CA CRL configmap exists for the given ingresscontroller if appropriate, using the client CA configmap, the existing CRL configmap if it exists, the `currentCRLConfigMap` and `updateCRLConfigMap` methods, and the `buildCRLMap` and `desiredCRLConfigMap` functions.
(`buildCRLMap`): New function.  Build a map of key id to CRL.
(`desiredCRLConfigMap`): New function.  Return the desired CRL configmap for the given ingresscontroller, using the given client CA data and CRLs and the `getCRL` function.
(`getCRL`): Get a CRL using the provided distribution points from the CA certificate.
(`currentCRLConfigMap`): New method.  Get the current CRL configmap.
(`updateCRLConfigMap`): New method.  Update the CRL configmap.
(`CRLConfigmapsEqual`): New function.  Compare CRL configmaps.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Get the current client CA configmap if it exists, and pass it to `ensureRouterDeployment`.
* `pkg/operator/controller/ingress/deployment.go` (`RouterClientAuthPolicy`, `RouterClientAuthCA`, `RouterClientAuthCRL`, `RouterClientAuthFilter`): New constants for the router environment variable names.
(`ensureRouterDeployment`): Add parameters for the client CA configmap.  Pass the parameters to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Configure the client certificate policy.  If a client CA configmap is provided, configure a volume mount for it.  If any client CA specifies any CRL distribution points, configure a volume mount for the CRL configmap.  If a list of allowed subject patterns is specified, configure it in the router deployment.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Adjust calls to `desiredRouterDeployment`.
* `pkg/operator/controller/names.go` (`ClientCAConfigMapName`): New function.  Return the name of the client CA configmap.
(`CRLConfigMapName`): New function.  Return the name of the CRL configmap.
* `pkg/operator/operator.go` (`New`): Start the crl controller.
* `test/e2e/client_tls_test.go`: New file.
(`TestClientTLS`): New test.  Verify that the operator properly configures client TLS policy on the router, and that the router enforces it.
(`generateClientCA`, `generateClientCertificate`, `encodeCert`, `encodeKey`): New helper functions for `TestClientTLS`.
* `test/e2e/util.go` (`buildExecPod`): New helper function for `TestClientTLS`.
* `go.sum`:
* `vendor/*`: Regenerate.

----

Related to https://github.com/openshift/enhancements/pull/462.

Depends on https://github.com/openshift/api/pull/731.

Related to https://github.com/openshift/router/pull/175.


Miciah:
Rebased.

Miciah:
Added E2E test.

pkg/operator/controller/crl/controller.go
```diff
@@ -0,0 +1,226 @@
...
+			return []string{}
+		}
+		return []string{ic.Spec.ClientTLS.ClientCA.Name}
+	})); err != nil {
+		return nil, fmt.Errorf("failed to create index for ingresscontroller: %v", err)
+	}
+
+	configmapsInformer, err := operatorCache.GetInformer(context.Background(), &corev1.ConfigMap{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create informer for configmaps: %v", err)
```
sgreene570, line 74:
```suggestion
		return nil, fmt.Errorf("failed to create informer for configmaps: %w", err)
```

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+		return false, nil, fmt.Errorf("failed to build configmap: %v", err)
+	}
+
+	switch {
+	case !wantCM && !haveCM:
+		return false, nil, nil
+	case !wantCM && haveCM:
+		if err := r.client.Delete(context.TODO(), current); err != nil {
+			if !errors.IsNotFound(err) {
+				return true, current, fmt.Errorf("failed to delete configmap: %v", err)
```
sgreene570, line 80:
```suggestion
				return true, current, fmt.Errorf("failed to delete configmap: %w", err)
```

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+		// (cannot use TLS without the revocation list, and cannot get
+		// the revocation list without using TLS).
+		//
+		// TODO Support ldap.
+		if !strings.HasPrefix(distributionPoint, "http:") {
+			errs = append(errs, fmt.Errorf("unsupported distribution point type: %s", distributionPoint))
+			continue
+		}
+		log.Info("retrieving CRL distribution point", "distribution point", distributionPoint)
+		resp, err := http.Get(distributionPoint)
```
sgreene570, line 233: Do we need to take into account cluster-wide proxy in this case? (I ask because im really not sure).

pkg/operator/controller/ingress/controller.go
```diff
@@ -705,7 +705,21 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		errs = append(errs, err)
 	}
 
-	haveDepl, deployment, err := r.ensureRouterDeployment(ci, infraConfig, ingressConfig, apiConfig, networkConfig)
+	var haveClientCAConfigmap bool
+	clientCAConfigmap := &corev1.ConfigMap{}
+	if len(ci.Spec.ClientTLS.ClientCA.Name) != 0 {
+		clientCAConfigmapName := types.NamespacedName{
+			Namespace: "openshift-ingress",
```
sgreene570, line 712: Can we use a constant here instead of `"openshift-ingress"`?

test/e2e/client_tls_test.go
```diff
@@ -0,0 +1,468 @@
...
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+	cl, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		t.Fatalf("failed to create kube client: %v", err)
+	}
+
+	// curl execs a Curl command in the test client pod and returns an error
+	// value.  The Curl command uses the specified certificate from the
+	// client certificates configmap, sends a request for the canary route
```
sgreene570, line 247: Ah, right, the canary route doesn't necessarily have to be served by the default ingress controller. Clever! :grin:

pkg/operator/controller/names.go
```diff
@@ -106,6 +106,25 @@ func RouterOperatorGeneratedDefaultCertificateSecretName(ci *operatorv1.IngressC
 	}
 }
 
+// ClientCAConfigMapName returns the namespaced name for the client CA
+// configmap.
+func ClientCAConfigMapName(ic *operatorv1.IngressController) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: "openshift-ingress",
+		// TODO Is this user-specified or operator-managed?
```
sgreene570, line 114: Is this an open question, or is this a TODO saying more logic needs to be added here?

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+	var errs []error
+	for _, distributionPoint := range distributionPoints {
+		// The distribution point is typically a URL with the "http"
+		// scheme.  "https" is generally not used because the
+		// certificate list is signed, and because using TLS to get the
+		// certificate list could introduce a circular dependency
+		// (cannot use TLS without the revocation list, and cannot get
+		// the revocation list without using TLS).
+		//
+		// TODO Support ldap.
```
sgreene570, line 227: Is LDAP support expected in 4.9?

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+	return true, &crlConfigmap, nil
+}
+
+// getCRL gets a certificate revocation list using the provided distribution
+// points and returns the certificate list.
+func getCRL(distributionPoints []string) (*pkix.CertificateList, error) {
+	var errs []error
+	for _, distributionPoint := range distributionPoints {
+		// The distribution point is typically a URL with the "http"
+		// scheme.  "https" is generally not used because the
```
sgreene570, line 221: Thank you for providing the https explanantion!

pkg/operator/controller/crl/crl_configmap.go
sgreene570: Should we consider tweaking any of the default settings on the default client (e.g. timeouts)?

pkg/operator/controller/crl/crl_configmap.go
sgreene570: This comment applies to a few other spots in this switch statement, I think.

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// authorityKeyIdentifier is a certificate's authority key identifier.
+type authorityKeyIdentifier struct {
```
sgreene570, line 28: Wondering if theres any go libraries we can use to get stuff like this for free.

sgreene570:
Myriad of strange e2e-operator failures


/test e2e-aws-operator

pkg/operator/controller/ingress/deployment.go
```diff
@@ -791,6 +798,120 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+					routerVolumeMounts = append(routerVolumeMounts, clientCACRLVolumeMount)
+
+					clientAuthCRLPath := filepath.Join(clientCACRLVolumeMount.MountPath, clientCACRLFilename)
+					env = append(env, corev1.EnvVar{Name: RouterClientAuthCRL, Value: clientAuthCRLPath})
+				}
+			}
+
+			if len(ci.Spec.ClientTLS.AllowedSubjectPatterns) != 0 {
+				pattern := "(?:" + strings.Join(ci.Spec.ClientTLS.AllowedSubjectPatterns, "|") + ")"
+				env = append(env, corev1.EnvVar{Name: RouterClientAuthFilter, Value: pattern})
```
sgreene570, line 909: This may be a silly question, but is there any chance a user could provide a regex that would break the HAProxy template (e.g. unmatched quotes)? I may be imagining a problem that doesn't exist, but I figured it was worth asking.

pkg/operator/controller/crl/controller.go
Miciah: Fixed.

pkg/operator/controller/crl/crl_configmap.go
Miciah: Fixed.  Thanks!

pkg/operator/controller/crl/crl_configmap.go
Miciah: Good questions.  I guess there's no reason this shouldn't use the proxy, right?  If it's an internal server, the user could just add it to `noProxy`.

pkg/operator/controller/ingress/controller.go
Miciah: I updated this to use the namespace from `ensureRouterNamespace`.

pkg/operator/controller/names.go
Miciah: I'm not sure what I was thinking when I wrote this comment, but I realize now, the configmap should be in the `openshift-config` namespace, and I should add a controller to copy the configmap from the `openshift-config` namespace to the `openshift-ingress` namespace.

pkg/operator/controller/crl/crl_configmap.go
Miciah: No, LDAP is out of scope.

pkg/operator/controller/crl/crl_configmap.go
Miciah: I don't think there's anything in the standard libraries (although they were useful as a reference to write all this code).

pkg/operator/controller/crl/crl_configmap.go
sgreene570: Makes sense to me.

pkg/operator/controller/names.go
sgreene570: Right, that makes sense.

pkg/operator/controller/ingress/deployment.go
Miciah: The user probably could do some configuration injection tricks this way.  I'm not too worried about the cluster admin pwning the cluster admin's own cluster.  Maybe it would be worth adding some validation to prevent accidents though?  We'd need to import a library that could parse PCRE though.

sgreene570: Hm, yea, might not be worth fixing. Just food for thought.

pkg/operator/controller/clientca-configmap/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	ic := &operatorv1.IngressController{}
+	if err := r.client.Get(ctx, request.NamespacedName, ic); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("ingresscontroller not found; reconciliation will be skipped", "request", request)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get ingresscontroller %q: %w", request.NamespacedName, err)
+	}
+
+	const finalizer = "ingresscontroller.operator.openshift.io/finalizer-clientca-configmap"
```
sgreene570, line 197: Consider moving this definition to `pkg/manifests/manifests.go` alongside `IngressControllerFinalizer`

pkg/operator/controller/clientca-configmap/controller.go
```diff
@@ -0,0 +1,231 @@
+package clientcaconfigmap
```
sgreene570, line 1: The Ingress Operator's controller count is growing! At some point it may be beneficial to add some high level docs describing what each controller does (and where they live in the codebase).

pkg/operator/controller/crl/crl_configmap.go
Miciah: https://github.com/openshift/cluster-ingress-operator/pull/334/commits/ead9d30aca702ff0a0baf254bd3ae4fe22a22048 already added the `config.openshift.io/inject-proxy: ingress-operator` annotation to the operator deployment, and this is an HTTP (not TLS) client, so it should already pick up any cluster proxy configuration.

Miciah:
Rebased on https://github.com/openshift/api/pull/731.

Miciah:
Rebased.

pkg/operator/controller/clientca-configmap/clientca_configmap.go
```diff
@@ -0,0 +1,137 @@
...
+// updateClientCAConfigMap updates a configmap.  Returns a Boolean indicating
+// whether the configmap was updated, and an error value.
+func (r *reconciler) updateClientCAConfigMap(ctx context.Context, current, desired *corev1.ConfigMap) (bool, error) {
+	if clientCAConfigmapsEqual(current, desired) {
+		return false, nil
+	}
+	updated := current.DeepCopy()
+	updated.Data = desired.Data
+	if err := r.client.Update(ctx, updated); err != nil {
+		if errors.IsAlreadyExists(err) {
```
frobware, line 120: Do you have an example when this would hold true? Is something else also updating the same map?

pkg/operator/controller/clientca-configmap/clientca_configmap.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	log.Info("updated configmap", "namespace", updated.Namespace, "name", updated.Name)
+	return true, nil
+}
+
+// clientCAConfigmapsEqual compares two client CA configmaps.  Returns true if
+// the configmaps should be considered equal for the purpose of determining
+// whether an update is necessary, false otherwise.
+func clientCAConfigmapsEqual(a, b *corev1.ConfigMap) bool {
+	if !reflect.DeepEqual(a.Data, b.Data) {
```
frobware, line 133: nit: the function could just return the evaluation of `DeepEqual`.

pkg/operator/controller/clientca-configmap/controller.go
```diff
@@ -0,0 +1,231 @@
...
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldIC := e.ObjectOld.(*operatorv1.IngressController)
+				newIC := e.ObjectNew.(*operatorv1.IngressController)
+				oldName := oldIC.Spec.ClientTLS.ClientCA.Name
+				newName := newIC.Spec.ClientTLS.ClientCA.Name
+				if oldName != newName {
+					return true
+				}
+				if oldIC.DeletionTimestamp != newIC.DeletionTimestamp {
```
frobware, line 73: nit: return the result of this expression.

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+		}
+
+	}
+
+	var clientCAData []byte
+	if haveClientCA {
+		clientCABundleFilename := "ca-bundle.pem"
+		if data, ok := clientCAConfigmap.Data[clientCABundleFilename]; !ok {
+			return haveCM, current, fmt.Errorf("client CA configmap %s/%s is missing %q", clientCAConfigmap.Namespace, clientCAConfigmap.Name, clientCABundleFilename)
```
frobware, line 63: What is the ongoing runtime behaviour when this happens - do we just continue to reconcile?

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+			return false, nil, fmt.Errorf("failed to create configmap: %w", err)
+		}
+		log.Info("created configmap", "namespace", desired.Namespace, "name", desired.Name)
+		return r.currentCRLConfigMap(ctx, ic)
+	case wantCM && haveCM:
+		if updated, err := r.updateCRLConfigMap(ctx, current, desired); err != nil {
+			return true, current, fmt.Errorf("failed to update configmap: %w", err)
+		} else if updated {
+			log.Info("updated configmap", "namespace", desired.Namespace, "name", desired.Name)
+			return r.currentCRLConfigMap(ctx, ic)
```
frobware, line 97: Could `updateCRLConfigMap` also return the updated value which would prevent the additional get here?

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -0,0 +1,295 @@
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("error getting %q: %w", distributionPoint, err))
+			continue
+		}
+		bytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("error reading response from %q: %w", distributionPoint, err))
+			continue
+		}
+		crl, err := x509.ParseCRL(bytes)
```
frobware, line 243: Close the response body; `defer resp.Body.Close()`.

test/e2e/client_tls_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// via the router's internal service, and returns an error if the Curl
+	// command fails or the HTTP response status code indicates an error.
+	curl := func(cert string) error {
+		req := cl.CoreV1().RESTClient().Post().Resource("pods").
+			Namespace(clientPod.Namespace).Name(clientPod.Name).
+			SubResource("exec").
+			Param("container", clientPod.Spec.Containers[0].Name)
+		cmd := []string{
+			"/bin/curl", "-k", "-v",
```
frobware, line 256: Is it practical to make a programmable HTTP request without needing to curl from within a pod?

test/e2e/client_tls_test.go
```diff
@@ -0,0 +1,468 @@
...
+			strings.Join(cmd, " "), stdoutStr, stderr.String())
+		if err != nil {
+			return err
+		}
+		httpStatusCode := stdoutStr[len(stdoutStr)-3:]
+		switch string(httpStatusCode[0]) {
+		case "0", "4", "5":
+			return fmt.Errorf("got HTTP %s status code", httpStatusCode)
+		}
+		return nil
```
frobware, line 297: Make this the `default` in the switch.

pkg/operator/controller/clientca-configmap/clientca_configmap.go
Miciah: I'm not sure why an update could return an already-exists response.  We're inconsistent in checking for already-exists responses when doing updates.  I'll remove the check here, and we can remove similar checks elsewhere in a follow-up.

pkg/operator/controller/clientca-configmap/clientca_configmap.go
Miciah: Done.

pkg/operator/controller/clientca-configmap/controller.go
Miciah: Done.

pkg/operator/controller/crl/crl_configmap.go
Miciah: It could, but this is the pattern we follow in other `ensure` functions.  I'd rather be consistent here, and we can refactor this and other code in a follow-up.

pkg/operator/controller/crl/crl_configmap.go
Miciah: That is very important!  Fixed.  Thanks!

Because this is in a loop, I am avoiding use of `defer` and instead doing this:
```go
		bytes, err := ioutil.ReadAll(resp.Body)
		// Close the body immediately rather than using defer because
		// this is inside a loop.
		err2 := resp.Body.Close()
		if err2 != nil {
			errs = append(errs, fmt.Errorf("error closing response body from %q: %w", distributionPoint, err2))
		}
		// Now check the error value from reading the body.
		if err != nil {
```
If you prefer, I could wrap the whole thing in a `defer func() { ... }()`, or refactor the loop body into a separate function, so that the deferred close happens before the next URL is tried.  Or maybe I'm overthinking this.  What do you think?

test/e2e/client_tls_test.go
Miciah: Done.  Thanks!

pkg/operator/controller/crl/crl_configmap.go
frobware: You can't call `Close()` without first checking for an error because `Body` will be `nil` if there was an error. If the loop terminates quickly I'd wrap it in a defer. If not, then I'd explicitly close it after a successful read. (Not sure I'd error on close failing.)

test/e2e/client_tls_test.go
frobware: I should have been more explicit; something that just does:

```go
resp, err := http.Get("http://example.com/")
```

but in the context of this test I don't know how practical that would be. Given that this works end-2-end I'd leave it, at least for now.

pkg/operator/controller/crl/crl_configmap.go
Miciah: > You can't call `Close()` without first checking for an error because `Body` will be `nil` if there was an error.

I thought I could.  Wouldn't using `defer resp.Body.Close()` cause the function to call `resp.Body.Close()` irrespective of whether there were an error on the read?

Miciah: FWIW, here's an example from https://pkg.go.dev/net/http:
```go
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
body, err := io.ReadAll(resp.Body)
```
We have similar code in the canary controller.  The deferred function will be called even if `io.ReadAll` returns an error.

frobware: Ah, sorry. I was incorrectly focusing on the read and not the get.

Miciah: I went ahead and refactored the HTTP code into a separate function.  The overall structure is nicer now too, with a `switch` statement for the different distribution point types (although currently only HTTP is supported).

Miciah:
I've never seen this before:

        configurable_route_test.go:356: Number of roleBindings in list != 1
    --- FAIL: TestConfigurableRouteNoConsumingUserNoRBAC (122.52s)

I'm not finding any similar failures on search.ci either.  Unfortunately the CI artifacts don't include operator logs from the time of the failure.  

/test e2e-aws-operator

Miciah:
failed to load configuration: failed to make request to configresolver: Get "https://config.ci.openshift.org/config?branch=master&org=openshift&repo=cluster-ingress-operator": connection error: PROTOCOL_ERROR

/test e2e-aws-operator

Miciah:
Known flake:

    [sig-auth][Feature:SCC][Early] should not have pod creation failures during install [Suite:openshift/conformance/parallel]

/test e2e-upgrade

Miciah:
Looks like a sick cluster:

    + oc new-project e2e-test-capabilities-check
    Error from server (Timeout): Timeout: request did not complete within requested timeout context deadline exceeded

/test e2e-upgrade

Miciah:
fail [github.com/openshift/origin/test/extended/util/disruption/disruption.go:190]: Jul 22 00:03:36.814: Unexpected alerts fired or pending during the upgrade:
    
    alert TargetDown fired for 241 seconds with labels: {job="promtail", namespace="loki", service="promtail", severity="warning"}
    alert TargetDown fired for 61 seconds with labels: {job="cco-metrics", namespace="openshift-cloud-credential-operator", service="cco-metrics", severity="warning"}

/test e2e-upgrade

PR #451, opened 2020-09-01
NE-408: Allow configuring ELB connection idle timeout

Miciah:
#### Bump openshift/api for ELB connection idle timeout

Bump to github.com/openshift/api@b25f69a603a76ccc809f986c9f5811f0825febbb to get the AWS ELB connection idle timeout API.

* `go.mod`: Update.
* `go.sum`:
* `pkg/manifests/bindata.go`:

#### `test/e2e`: import API errors package as `apierrors`

Import the apimachinery errors package as "apierrors".

* `test/e2e/operator_test.go`: Import the "k8s.io/apimachinery/pkg/api/errors" package as "apierrors" so as not to conflict with the standard "errors" package.

#### `desiredLoadBalancerService`: Simplify with "lb" var

* `pkg/operator/controller/ingress/load_balancer_service.go` (`desiredLoadBalancerService`): Introduce an "lb" variable to shorten some long lines.

#### `desiredLoadBalancerService`: Check for nil LB status

* `pkg/operator/controller/ingress/load_balancer_service.go` (`desiredLoadBalancerService`): Add a nil check just in case `status.endpointPublishingStrategy.LoadBalancer` is nil somehow.

#### setDefaultPublishingStrategy: Rework GCP logic

Rework the GCP load-balancer provider parameters defaulting logic in preparation for the next change.  Besides simplifying the logic, this change also changes the defaulting logic to ignore unknown provider parameters and to ignore provider parameters for platforms other than the actual platform.  These changes should avoid surprises when more provider parameters are added to the API later on as well as prevent weird behavior when the user sets GCP provider parameters on non-GCP clusters.

* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Rework defaulting logic for GCP load-balancer provider parameters.

#### Allow configuring ELB connection idle timeout

* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Handle changes to the connection idle timeout for an AWS ELB.
* `pkg/operator/controller/ingress/controller_test.go` (`TestSetDefaultPublishingStrategyHandlesUpdates`): Add test cases for changing the ELB connection idle timeout.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`awsELBConnectionIdleTimeoutAnnotation`): New constant.
(`managedLoadBalancerServiceAnnotations`): Add `awsELBConnectionIdleTimeoutAnnotation`.
(`desiredLoadBalancerService`): Set the connection idle timeout annotation if the ingresscontroller specifies a non-nil `connectionIdleTimeout` value.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestDesiredLoadBalancerServiceAWSIdleTimeout`): New test.
* `test/e2e/operator_test.go` (`TestAWSELBConnectionIdleTimeout`): New test.
* `test/e2e/util.go` (`buildSlowHTTPDPod`): New helper for `TestAWSELBConnectionIdleTimeout`.

----

Related to https://github.com/openshift/enhancements/pull/461.


Miciah:
In [the last e2e-aws-operator run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/451/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1512083107149254656), `TestConfigurableRouteRBAC`, `TestConfigurableRouteNoSecretNoRBAC`, `TestConfigurableRouteNoConsumingUserNoRBAC`, `TestHostNetworkEndpointPublishingStrategy`, and `TestAWSELBConnectionIdleTimeout` failed.  The last failed with the following error message:
```
     operator_test.go:2208: failed to get ingresscontroller: Get "https://api.ci-op-1yr8trkg-43abb.origin-ci-int-aws.dev.rhcloud.com:6443/apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/test-idle-timeout": http2: client connection lost 
```
I could add some retry logic around the `Get` call that failed, but the E2E tests have a lot of `Get` calls without retry logic, and the other failures in the test run lead me to think that the cluster had some other issue unrelated to this PR.  

/retest

Miciah:
Latest push adds a unit test that I had forgotten to commit.

Miciah:
Latest push adds a test case that I had also forgotten to commit.

Miciah:
In [the last e2e-aws-operator run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/451/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1512204867488190464), `TestConfigurableRouteNoSecretNoRBAC`, `TestConfigurableRouteNoSecretNoRBAC`, `TestConfigurableRouteNoConsumingUserNoRBAC`, and `TestHostNetworkEndpointPublishingStrategy` failed.  The last is showing up a lot [in search.ci](https://search.ci.openshift.org/?search=FAIL%3A+TestHostNetworkEndpointPublishingStrategy&maxAge=336h&context=2&type=build-log&name=cluster-ingress-operator&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job).

/retest

pkg/operator/controller/ingress/controller.go
```diff
@@ -412,19 +412,31 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
...
+			var lbType operatorv1.LoadBalancerProviderType
+			if specLB.ProviderParameters != nil {
+				lbType = specLB.ProviderParameters.Type
 			}
-
-			if specLB.ProviderParameters != nil && statusLB.ProviderParameters != nil &&
-				specLB.ProviderParameters.GCP != statusLB.ProviderParameters.GCP {
-				ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP = specLB.ProviderParameters.GCP
-				changed = true
+			switch lbType {
```
candita, line 419: Can you add a comment here, like the comment in the previous revision?

candita: This comment is meant for commit f68e0326feb3cf713f0

pkg/operator/controller/ingress/controller.go
```diff
@@ -412,19 +412,63 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
 				changed = true
 			}
```
candita, line 414: Is this change fixing a bug? Can you add a comment here, to replace the one you removed?

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -337,6 +353,18 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
 			expectedResult: false,
 			expectedIC:     makeIC(spec(elb()), status(nlb())),
 		},
+		{
+			name:           "loadbalancer ELB connection idle timeout changed from unset to 2m",
+			ic:             makeIC(spec(elbWithIdleTimeout(&metav1.Duration{Duration: 2 * time.Minute})), status(elb())),
```
candita, line 358: Why is the test case initially `status(elb())` and not `status(elbWithIdleTimeout(nil))`?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -338,10 +346,10 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 
 			// Set the GCP Global Access annotation for internal load balancers on GCP only
 			if platform.Type == configv1.GCPPlatformType {
-				if ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
-					ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.Type == operatorv1.GCPLoadBalancerProvider &&
-					ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP != nil {
-					globalAccessEnabled := ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP.ClientAccess == operatorv1.GCPGlobalAccess
+				if lb.ProviderParameters != nil &&
```
candita, line 349: Could `lb` be nil?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -353,16 +361,23 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
-			} else {
-				service.Annotations[awsLBHealthCheckIntervalAnnotation] = awsLBHealthCheckIntervalDefault
+			service.Annotations[awsLBHealthCheckIntervalAnnotation] = awsLBHealthCheckIntervalDefault
+			if lb != nil && lb.ProviderParameters != nil {
+				if aws := lb.ProviderParameters.AWS; aws != nil && lb.ProviderParameters.Type == operatorv1.AWSLoadBalancerProvider {
+					switch aws.Type {
+					case operatorv1.AWSNetworkLoadBalancer:
+						service.Annotations[AWSLBTypeAnnotation] = AWSNLBAnnotation
+						// NLBs require a different health check interval than CLBs.
+						// See <https://bugzilla.redhat.com/show_bug.cgi?id=1908758>.
```
candita, line 371: Scattering bug numbers around feels like it's going to be a problem once Bugzilla is gone.

pkg/operator/controller/ingress/controller.go
```diff
@@ -412,19 +412,63 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
...
+					if v := specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout; v != nil {
+						specIdleTimeout = *v
+					}
+				}
+				if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
+					if v := statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout; v != nil {
+						statusIdleTimeout = *v
+					}
+				}
+				if specIdleTimeout != statusIdleTimeout {
```
candita, line 433: Should this also do some small validation on the timeout value to make sure a good value doesn't get changed into a bad value? I know for durations, there is no validation in the api.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -337,6 +353,18 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
...
+			expectedResult: true,
+			expectedIC:     makeIC(spec(elbWithIdleTimeout(&metav1.Duration{Duration: 2 * time.Minute})), status(elbWithIdleTimeout(&metav1.Duration{Duration: 2 * time.Minute}))),
+		},
+		{
+			name:           "loadbalancer ELB connection idle timeout changed from 2m to unset",
+			ic:             makeIC(spec(elb()), status(elbWithIdleTimeout(&metav1.Duration{Duration: 2 * time.Minute}))),
+			expectedResult: true,
+			expectedIC:     makeIC(spec(elb()), status(elbWithIdleTimeout(nil))),
+		},
```
candita, line 367: Maybe add a test for a bad value to see what happens.

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -463,6 +464,148 @@ func TestDesiredLoadBalancerService(t *testing.T) {
...
+					AWS: &operatorv1.AWSLoadBalancerParameters{
+						Type:                          operatorv1.AWSClassicLoadBalancer,
+						ClassicLoadBalancerParameters: nil,
+					},
+				},
+			},
+			expectAnnotationPresent: false,
+		},
+		{
+			name: "nil providerParameters",
```
candita, line 514:
```suggestion
		name: "nil providerParameters.aws.classicLoadBalancerParameters.connectionIdleTimeout",
```

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -463,6 +464,148 @@ func TestDesiredLoadBalancerService(t *testing.T) {
...
+						Type: operatorv1.AWSClassicLoadBalancer,
+						ClassicLoadBalancerParameters: &operatorv1.AWSClassicLoadBalancerParameters{
+							ConnectionIdleTimeout: &metav1.Duration{Duration: time.Hour},
+						},
+					},
+				},
+			},
+			expectAnnotationPresent: true,
+			expectedAnnotationValue: "3600",
+		},
```
candita, line 559: Maybe one more test with an invalid `connectionIdleTimeout`?

test/e2e/operator_test.go
```diff
@@ -2057,6 +2059,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
 	}
 }
 
+// TestAWSELBConnectionIdleTimeout verifies that the AWS ELB connection-idle
```
candita, line 2062: Do you also need to add/change coverage for the GCP changes you made?

test/e2e/operator_test.go
```diff
@@ -2057,6 +2059,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+			// unlikely to cause false negatives or false positives.
+			if elapsed.Seconds() > float64(15) {
+				t.Logf("got expected EOF error after unexpectedly long elapsed time %v", elapsed)
+				return false, nil
+			}
+
+			t.Logf("got expected EOF after elapsed time %v", elapsed)
+			return true, nil
+		} else {
+			defer response.Body.Close()
```
candita, line 2189: Do you need to close the body before you exit the polling loop?  It's just a test, but it still looks like it's in the wrong place to me.

test/e2e/operator_test.go
```diff
@@ -2057,6 +2059,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+			body, err := ioutil.ReadAll(response.Body)
+			if err != nil {
+				t.Log(err)
+				return false, nil
+			}
+
+			elapsed := time.Now().Sub(start)
```
candita, line 2196: Do you need to check this elapsed time to make sure it is within the limits?

test/e2e/util.go
```diff
@@ -135,6 +135,38 @@ func buildExecPod(name, namespace, image string) *corev1.Pod {
 	}
 }
 
+// buildSlowHTTPDPod returns a pod that responds to HTTP requests slowly.
+func buildSlowHTTPDPod(name, namespace string) *corev1.Pod {
```
candita, line 139: Nice.

pkg/operator/controller/ingress/controller.go
Miciah: The intention wasn't to fix bugs, but it probably does prevent some weird behavior and could prevent future bugs.  I've added some comments and rewritten the commit message to elaborate on this point.

pkg/operator/controller/ingress/controller.go
Miciah: Good point, we might as well check for invalid values here.  I've changed the check below from `!= 0` to `> 0`.  Let me know if you had further validations in mind.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Many other projects use Bugzilla outside of OpenShift.  I doubt Bugzilla will ever go away (and I would be very disappointed if it did).

pkg/operator/controller/ingress/load_balancer_service_test.go
Miciah: Thanks!  Fixed.

pkg/operator/controller/ingress/load_balancer_service_test.go
Miciah: Done.  I also added a check for negative values in `desiredLoadBalancerService` so the new test would pass.

test/e2e/operator_test.go
Miciah: I think `TestInternalLoadBalancerGlobalAccessGCP` already provides sufficient coverage.

test/e2e/operator_test.go
Miciah: We don't need to close the body if the client got an error.  The deferred call to `response.Body.Close()` will run when the polling loop function returns (note the `func()` in `wait.PollImmediate(..., ..., func() (bool, error) {`), so I think this logic is correct and will close the body every time the polling loop runs and makes a successful client request.

test/e2e/operator_test.go
Miciah: I don't think it matters here; this is the branch where the client request succeeded.  The other branch, which checks for timeout values, is the one where we really care that the timeout happened and was within the expected bound.

pkg/operator/controller/ingress/controller_test.go
Miciah: I'll add a test case for `status(elbWithIdleTimeout(nil))`.

pkg/operator/controller/ingress/controller_test.go
Miciah: Is the "loadbalancer ELB connection idle timeout changed from unset to -1s" test what you had in mind?

pkg/operator/controller/ingress/controller.go
```diff
@@ -412,19 +412,70 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
...
+						}
+					}
+					if statusLB.ProviderParameters.AWS == nil {
+						statusLB.ProviderParameters.AWS = &operatorv1.AWSLoadBalancerParameters{}
+					}
+					if statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil {
+						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters = &operatorv1.AWSClassicLoadBalancerParameters{}
+					}
+					var v *metav1.Duration
+					if specIdleTimeout.Duration > 0 {
```
candita, line 451: can `specIdleTimeout` be nil?

Miciah: No, it isn't a pointer type—it's declared as  `var specIdleTimeout, statusIdleTimeout metav1.Duration`.

(BTW, the line number that you are asking about is also the number of this PR!)

Miciah:
Rebased.

Miciah:
Fixed the bump commit's message.

Miciah:
I had forgotten to `git add` a file under `vendor/`.

Miciah:
Updated E2E test.

Miciah:
`TestAWSELBConnectionIdleTimeout` passed which shows that the new feature works.  `TestHTTPCookieCapture` failed, which is most likely a flake.
/retitle NE-408: Allow configuring ELB connection idle timeout
/test e2e-aws-operator

pkg/operator/controller/ingress/controller.go
frobware: Does [commit "rework GCP logic"](b45521cb8bc9d9b6f04cc6e6c7237e6b68bdada7) belong to this PR? Can it go in the upcoming commit?

Miciah: It does.  Are you suggesting I merge the commits?  I thought it would be much simpler to understand the changes if I first refactored to simplify the existing logic and then added the new logic.

test/e2e/operator_test.go
```diff
@@ -2165,6 +2167,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+	} else if v != expected {
+		t.Fatalf("expected %s=%s, found %s=%s", key, expected, key, v)
+	}
+
+	// Make sure we can resolve the route's host name.  It may take some
+	// time for the ingresscontroller's wildcard DNS record to propagate.
+	if err := wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
+		_, err := net.LookupIP(route.Spec.Host)
+		if err != nil {
+			t.Log(err)
```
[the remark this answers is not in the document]
Miciah: It's test output, it should be under 300 lines, and it's hidden unless a test fails.  Is that too chatty?  I could add some logic to suppress the log message if it's identical to the previous one.

test/e2e/operator_test.go
```diff
@@ -2165,6 +2167,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+			return false, nil
+		}
+
+		return true, nil
+	}); err != nil {
+		t.Errorf("failed to observe expected condition: %v", err)
```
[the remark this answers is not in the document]
Miciah: I suppose it wouldn't hurt to make this `Fatal`.

test/e2e/operator_test.go
```diff
@@ -2165,6 +2167,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+				return false, nil
+			}
+
+			elapsed := time.Now().Sub(start)
+			t.Logf("got response after elapsed time %v: %v", elapsed, string(body))
+		}
+
+		return false, nil
+	}); err != nil {
+		t.Errorf("failed to observe expected condition: %v", err)
```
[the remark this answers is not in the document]
Miciah: Isn't it potentially useful to know whether a large value causes the expected behavior when diagnosing why a low value does not?

test/e2e/operator_test.go
```diff
@@ -2165,6 +2167,226 @@ func TestNetworkLoadBalancer(t *testing.T) {
...
+			t.Log(err)
+			return false, nil
+		}
+
+		elapsed := time.Now().Sub(start)
+		t.Logf("got expected response after elapsed time %v: %v", elapsed, string(body))
+
+		return true, nil
+	}); err != nil {
+		t.Errorf("failed to observe expected condition: %v", err)
```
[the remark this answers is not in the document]
Miciah: You're suggesting making every `Error` into a `Fatal`.  What's an example of where you would advise using `Error`?  

We should use `Fatal` if any subsequent testing can only produce garbage.  However, I think in some of these instances where I used `Error` in this test, there could be value in continuing the test.  For example, even if setting a 3-second timeout didn't behave as expected, we can still try a 120-second timeout to see whether it behaves as expected, and the result could be helpful in diagnosing why the 3-second timeout didn't behave as expected.

Miciah:
I've changed `Error` to `Fatal` in the E2E test.

test/e2e/operator_test.go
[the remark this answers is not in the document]
frobware: > You're suggesting making every `Error` into a `Fatal`. What's an example of where you would advise using `Error`?

Table-driven unit tests.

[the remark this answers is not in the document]
frobware: > We should use `Fatal` if any subsequent testing can only produce garbage

Isn't that the case for all these e2e tests; we try to setup up the cluster/objects/resources/state in a very particular way and if that doesn't happen is it worth generating cascading failures messages if the test was to continue? If you were to debug a failing test you're likely to start with the first error message. Would further error messages help?

frobware (approved):

PR #457, opened 2020-09-12
Bug 1866299: Refactors AWS Provider Validation

danehans:
Previously, an AWS provider would be used without validating that clients could communicate with associated API endpoints. This PR adds validation by ensuring each AWS provider client can make a list/describe API call.

/assign @Miciah @knobunc 
/cc @frobware @sgreene570


danehans:
`level=error msg="Cluster operator machine-config Degraded is True with MachineConfigControllerFailed: Unable to apply 4.6.0-0.ci.test-2020-09-12-235633-ci-op-kk4xnx01: timed out waiting for the condition during waitForControllerConfigToBeCompleted: controllerconfig is not completed: status for ControllerConfig machine-config-controller is being reported for 0, expecting it for 1"
`

/test e2e-aws

pkg/dns/aws/dns.go
```diff
@@ -208,48 +195,41 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
-				return true
-			}
-		default:
-			if uri == standardRoute53Endpoint || strings.Contains(uri, endpoints.UsEast1RegionID) {
-				return true
-			}
+// validateServiceEndpoints validates that provider clients can communicate
+// with associated API endpoints by each making a list/describe/get call.
+func validateServiceEndpoints(provider *Provider) error {
+	if provider.route53 != nil {
```
Miciah, line 208: Can `provider.route53`, `provider.elb`, `provider.elbv2`, or `provider.tags` ever be nil?

pkg/dns/aws/dns.go
```diff
@@ -208,48 +195,41 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
 		}
 	}
-	return false
+	if provider.tags != nil {
+		input := resourcegroupstaggingapi.GetResourcesInput{TagsPerPage: aws.Int64(int64(100))}
+		if _, err := provider.tags.GetResources(&input); err != nil {
+			return fmt.Errorf("failed to get group tagging resources: %v", err)
+		}
+	}
+	return nil
```
Miciah, line 232: What do you think of validating all endpoints, accumulating errors, and returning the result at the end?  Something like this:
```go
 	var errs []error
	if _, err := ...; err != nil {
		errs = append(errs, fmt.Errorf(...))
	}
	if _, err := ...; err != nil {
		errs = append(errs, fmt.Errorf(...))
	}
	if _, err := ...; err != nil {
		errs = append(errs, fmt.Errorf(...))
	}
	if _, err := ...; err != nil {
		errs = append(errs, fmt.Errorf(...))
	}
	return kerrors.NewAggregate(errs)
```
(If `errs` is empty, `kerrors.NewAggregate(errs)` returns nil.)

Miciah (commented):
Just a couple minor suggestions.  This looks like a good approach.

pkg/dns/aws/dns.go
danehans: Not the way it's being called from `NewProvider()` today. I'll update to remove the nil check.

pkg/dns/aws/dns.go
danehans: Ah, yes I like that approach. Let me update.

danehans:
@Miciah commit `ddc335f` implements your feedback. Thanks for the review!

danehans:
```
2020/09/14 17:37:09 pod didn't start running within 10 minutes: 
* Container artifacts is not ready with reason PodInitializing and message 
* Container test is not ready with reason PodInitializing and message 
```

/test e2e-aws

PR #343, opened 2020-01-04
Add NodePortService endpoint publishing strategy

Miciah:
This commit resolves [NE-260](https://issues.redhat.com/browse/NE-260).

* `README.md`: Document `NodePortService`.
* `docs/images/endpoint-publishing-nodeportservice.png`: New file.
* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Add `NodePortServiceStrategyType`.
(`ensureIngressController`): Call `ensureNodePortService`.
* `pkg/operator/controller/ingress/nodeport_service.go`: New file.
(`ensureNodePortService`): New method.  Ensure the appropriate NodePort service exists for the given ingress controller, if one is desired.  Get the current one using `currentNodePortService`.  If none exists and one is needed, create it using the `desiredNodePortService` function.  If one already exists, update it if necessary using `updateNodePortService`, or delete it if none is desired.
(`desiredNodePortService`): New function.
(`currentNodePortService`): New method.
(`updateNodePortService`): New method.  Use `nodePortServiceChanged` to determine whether an update is needed.
(`nodePortServiceChanged`): New function.
* `pkg/operator/controller/ingress/nodeport_service_test.go`: New file.
(`TestDesiredNodePortService`): New test.
(`TestNodePortServiceChanged`): New test.
* `pkg/operator/controller/names.go` (`NodePortServiceName`): New function.
* `test/e2e/operator_test.go` (`TestNodePortServiceChanged`): New test.
(`newNodePortController`): New function.

PR #465, opened 2020-09-22
Bug 1881155: Improve status reporting when pods cannot be scheduled

Miciah:
#### status: Report reason for degraded status

* `pkg/operator/controller/status/controller.go` (`computeOperatorDegradedCondition`): For each ingresscontroller that is reporting `Degraded=True`, report the status condition's reason and message.

#### ingress: Add `PodsScheduled` status condition

* `pkg/operator/controller/ingress/controller.go` (`IngressControllerPodsScheduledConditionType`): New constant.
(`ensureIngressController`): Get a list pods in the `openshift-ingress` namespace and pass the list to `syncIngressControllerStatus`.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Add `pods` parameter.  Call `computeDeploymentPodsScheduledCondition`, passing in the deployment and pods.
(`computeDeploymentPodsScheduledCondition`): New function.  Use the deployment and pods to identify unscheduled pods, and compute and return the `PodsScheduled` status condition.
(`computeIngressDegradedCondition`): Add `PodsScheduled` with status `True` to the set of expected conditions.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputePodsScheduledCondition`): New test.
(`TestComputeIngressDegradedCondition`): Update to check `PodsScheduled`.

#### ingress: Make `DeploymentDegraded` more verbose

* `pkg/operator/controller/ingress/status.go` (`computeIngressDegradedCondition`): When building the `DeploymentDegraded` status condition's message, include the reason and message of each status condition that does not have the expected status.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -436,7 +436,7 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition) 
 
 		var degraded string
 		for _, cond := range degradedConditions {
-			degraded = degraded + fmt.Sprintf(", %s=%s", cond.Type, cond.Status)
+			degraded = degraded + fmt.Sprintf(", %s=%s (%s: %s)", cond.Type, cond.Status, cond.Reason, cond.Message)
 		}
 		degraded = degraded[2:]
 
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
@@ -15,11 +16,13 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	oputil "github.com/openshift/cluster-ingress-operator/pkg/util"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
@@ -29,7 +32,7 @@ var clock utilclock.Clock = utilclock.RealClock{}
 
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
-func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) error {
+func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) error {
 	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
 	if err != nil {
 		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
@@ -41,6 +44,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.AvailableReplicas = deployment.Status.AvailableReplicas
 	updated.Status.Selector = selector.String()
 	updated.Status.TLSProfile = computeIngressTLSProfile(ic.Status.TLSProfile, deployment)
+	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentPodsScheduledCondition(deployment, pods))
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(deployment))
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
 	updated.Status.Conditions = mergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -102,6 +105,73 @@ func computeIngressTLSProfile(oldProfile *configv1.TLSProfileSpec, deployment *a
...
 }
 
+// computeDeploymentPodsScheduledCondition computes the ingress controller's
+// current PodsScheduled status condition state by inspecting the PodScheduled
+// conditions of the pods associated with the deployment.
+func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods []corev1.Pod) operatorv1.OperatorCondition {
+	unscheduled := make(map[*corev1.Pod]corev1.PodCondition)
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil || selector.Empty() {
+		return operatorv1.OperatorCondition{
```
sgreene570, line 115: Could you add a unit test that reaches this condition?

pkg/operator/controller/ingress/status.go
```diff
@@ -102,6 +105,73 @@ func computeIngressTLSProfile(oldProfile *configv1.TLSProfileSpec, deployment *a
...
+			if cond.Status == corev1.ConditionTrue {
+				continue
+			}
+			unscheduled[&pods[i]] = pod.Status.Conditions[j]
+		}
+	}
+	if len(unscheduled) != 0 {
+		var haveUnschedulable bool
+		message := "Some pods are not scheduled:"
+		// Sort keys so that the result is deterministic.
```
sgreene570, line 139: Could you help me understand why sorting `keys` here is necessary?

Miciah: Without sorting, the ordering is nondeterministic.  Suppose you have asked for 2 replicas and have 3 master nodes and 0 worker nodes.  The `PodsSchedulable` status condition reports the following:

      - lastTransitionTime: "2020-09-22T00:51:15Z"
        message: 'Some pods are not scheduled: Pod "router-default-889b9ccdc-5x5z8" cannot be scheduled: 0/3 nodes are available: 3 node(s) didn''t match node selector. Pod "router-default-889b9ccdc-zww95" cannot be scheduled: 0/3 nodes are available: 3 node(s) didn''t match node selector. Make sure you have sufficient worker nodes.'
        reason: PodsNotScheduled
        status: "False"
        type: PodsSchedulable

A moment later, the operator's reconciliation loop runs again and recomputes the status conditions.  This time, the `PodsSchedulable` status condition reports the following:

      - lastTransitionTime: "2020-09-22T00:52:15Z"
        message: 'Some pods are not scheduled: Pod "router-default-889b9ccdc-zww95" cannot be scheduled: 0/3 nodes are available: 3 node(s) didn''t match node selector. Pod "router-default-889b9ccdc-5x5z8" cannot be scheduled: 0/3 nodes are available: 3 node(s) didn''t match node selector. Make sure you have sufficient worker nodes.'
        reason: PodsNotScheduled
        status: "False"
        type: PodsSchedulable

We had bad luck, and the ordering changed.  As a consequence, the message changed, and therefore the status condition's last transition time changed as well.  This means that the last transition time is unreliable: it does not necessarily tell us when the status really changed.

`computeIngressDegradedCondition` scans several status conditions, including (as of this PR) `PodsSchedulable`, to determine whether it should report `Degraded=True`.  Rather than immediately reporting `Degraded=True` the instant it sees something wrong, `computeIngressDegradedCondition` allots each status condition a grace period and checks the status condition's last transition time to determine whether that period has expired.  Therefore `computeIngressDegradedCondition` needs the last transition time to be accurate so that it reports `Degraded=True` once the status condition's grace period has expired; the nondeterministic ordering may prevent it from ever reporting `Degraded=True` when `PodsSchedulable=False`.

pkg/operator/controller/ingress/status.go
Miciah: Done.

pkg/operator/controller/ingress/status.go
sgreene570: I see. That makes sense! Thanks for the details.

pkg/operator/controller/status/controller.go
```diff
@@ -264,11 +265,27 @@ func computeOperatorDegradedCondition(ingresses []operatorv1.IngressController)
 			Reason: "NoIngressControllersDegraded",
 		}
 	}
+	message := "Some ingresscontrollers are degraded:"
+	// Sort keys so that the result is deterministic.
+	keys := make([]*operatorv1.IngressController, 0, len(degradedIngresses))
+	for ingress := range degradedIngresses {
+		keys = append(keys, ingress)
+	}
+	sort.Slice(keys, func(i, j int) bool {
```
sgreene570, line 274: This slice sort function is the same sort used in this PR's changes to `pkg/operator/controller/ingress/status.go`, right? 
Would it be worth it to add this sort function to `pkg/util/slice`, or somehow refactor this PR to encourage code re-use?

Miciah: The two anonymous functions are deceptively similar in appearance, but in one case, `keys` is a `[]*operatorv1.IngressController`, and in the other case, `keys` is a `[]*corev1.Pod`.  I could add a helper in something like `pkg/util/sort.go`:
```go
// ObjectLess compares two ObjectMeta values and returns true iff the first one
// is less than the second one as determined by their creation timestamps, using
// their UIDs as a tie breaker.
func ObjectLess(x, y *metav1.ObjectMeta) bool {
	if x.CreationTimestamp.Equal(&y.CreationTimestamp) {
		return x.UID < y.UID
	}
	return x.CreationTimestamp.Before(&y.CreationTimestamp)
}
```
and then use it with `sort.Slice`:
```go
	sort.Slice(keys, func(i, j int) bool {
		return oputil.ObjectLess(&keys[i].ObjectMeta, &keys[j].ObjectMeta)
	})
```
Would you prefer that to what I have now?  Is there a simpler approach that I'm not seeing?

sgreene570: > Would you prefer that to what I have now? Is there a simpler approach that I'm not seeing?

The helper `ObjectLess` function is exactly what i had in mind.

Miciah: All right, latest push adds this change.

pkg/operator/controller/ingress/status.go
wking: > As a consequence, the message changed, and therefore the status condition's last transition time changed as well.

[API docs][1] say that `lastTransitionTime` is the time of the last `status` change regardless of `reason` or `message` changes.  For example, see [this library-go code][2].

[1]: https://github.com/openshift/api/blob/2a18526802e31a4ea3b0ff6ec807b27f1630074a/config/v1/types_cluster_operator.go#L123
[2]: https://github.com/openshift/library-go/blob/9e6c27e916a0a2aa8eef019683b1c05caecfaa61/pkg/config/clusteroperator/v1helpers/status.go#L29-L31

wking: But `lastTransitionTime` aside, I still like constructing the message deterministically to avoid potentially distracting churn on meaningless ordering changes.

Miciah: Isn't this technically a breaking API change?  * grin *

https://github.com/openshift/api/pull/501/commits/40c55f8085093f8dc7b93d70f431011ba74e6fdc#diff-b279e5574ea47c9158c793cb4ea7bf4aL123-R123

We picked up the change in 4.4:

https://github.com/openshift/cluster-ingress-operator/pull/343/commits/5454b6886fdba160253dfe01d459defbe2b8520e#diff-a882102033bf49e063e42db2ed09b51fL123-R123

So this is a long-standing issue, maybe best resolved separately.

PR #466, opened 2020-09-23
Bug 1881210: desiredNodePortService: Add port 1936

Miciah:
#### `desiredRouterDeployment`: Fix comment on host ports

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Mention that the deployment exposes port 1936 in addition to ports 80 and 443.

#### `desiredNodePortService`: Add port 1936

When creating the `NodePort`-type service for an ingresscontroller that specifies the `NodePortService` endpoint publishing strategy type, include a "metrics" port with port 1936 so that external load balancers can use the `/healthz/ready` endpoint.

* `pkg/operator/controller/ingress/nodeport_service.go` (`ensureNodePortService`): Pass an argument to `desiredNodePortService` to tell it not to add a "metrics" port if the service already exists and omits the port.
(`desiredNodePortService`): Add a `wantMetricsPort` parameter.  Add port 1936 to the service's ports spec if `wantMetricsPort` is true.
* `pkg/operator/controller/ingress/nodeport_service_test.go` (`TestDesiredNodePortService`): Add a test case where `wantMetricsPort` is true.  Verify that `desiredNodePortService` returns the expected service. (`TestNodePortServiceChanged`): Update the service used for tests to be consistent with what `desiredNodePortService` returns.
* `test/e2e/operator_test.go` (`TestNodePortServiceEndpointPublishingStrategy`): Verify that the operator creates the `NodePort`-type service with the expected ports.  Verify that the operator does not add the "metrics" port if it is is removed from the service.

### DIFF
--- a/pkg/operator/controller/ingress/nodeport_service.go
+++ b/pkg/operator/controller/ingress/nodeport_service.go
@@ -24,12 +24,29 @@ import (
 // indicating whether the NodePort service exists, the current NodePort service
 // if it does exist, and an error value.
 func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) (bool, *corev1.Service, error) {
-	wantService, desired := desiredNodePortService(ic, deploymentRef)
-
 	haveService, current, err := r.currentNodePortService(ic)
 	if err != nil {
 		return false, nil, err
 	}
+
+	// For compatibility, omit the "metrics" port iff the service already
+	// exists and doesn't have a "metrics" port.  This serves two purposes:
+	// (1) It avoids exhausting the nodeport range on upgrades if the
+	// cluster has many nodeport services and few available nodeports.
+	// (2) It enables the cluster administrator to remove the metrics port
+	// from an existing nodeport service to avoid exposing the port.
+	wantMetricsPort := false
+	if !haveService {
+		wantMetricsPort = true
+	} else {
+		for _, port := range current.Spec.Ports {
+			if port.Name == "metrics" {
+				wantMetricsPort = true
+			}
+		}
+	}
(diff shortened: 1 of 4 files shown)


sgreene570:
@Miciah do we want to revive this PR?

pkg/operator/controller/ingress/nodeport_service.go
```diff
@@ -24,13 +24,26 @@ import (
...
 func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference) (bool, *corev1.Service, error) {
-	wantService, desired := desiredNodePortService(ic, deploymentRef)
-
 	haveService, current, err := r.currentNodePortService(ic)
 	if err != nil {
 		return false, nil, err
 	}
 
+	// For compatibility, omit the "metrics" port iff the service already
+	// exists and doesn't have a "metrics" port.
```
sgreene570, line 33: Could you add a comment here mentioning that this logic allows a user to remove the metrics port from the node port service (as discussed offline) ?

sgreene570:
> * `test/e2e/operator_test.go` (`TestNodePortServiceEndpointPublishingStrategy`): Verify that the operator creates the `NodePort`-type service with the expected ports.  Verify that the operator does not add the "metrics" port of it is is removed from the service.

Small commit message typo. "of it is removed" -> "if it is removed".

pkg/operator/controller/ingress/nodeport_service.go
Miciah: Added.  Thanks!

Miciah:
Updated per Steve's comments.

PR #468, opened 2020-09-25
Organize dockerfiles

sgreene570:
Appropriately moves `Dockerfile` (the dockerfile currently used by ART) to `Dockerfile.rhel`. Moves` Dockerfile.rhel7` (not used by ART, not a RHEL-8 base) to `Dockerfile`. Basically following along with other components Dockerfile naming conventions, such as the [Cluster Version Operator](https://github.com/openshift/cluster-version-operator/blob/master/Dockerfile.rhel).

This will have repercussions in CI/ART and will most likely require follow up PRs in `openshift/release`, etc. We can safely merge this once 4.7 branching is open.

Slack context w/ Clayton: https://coreos.slack.com/archives/CCH60A77E/p1599076992180700?thread_ts=1598992667.131100&cid=CCH60A77E

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.15 AS builder
 WORKDIR /ingress-operator
 COPY . .
 RUN make build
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.15 AS builder
 WORKDIR /ingress-operator
 COPY . .
 RUN make build
```
sgreene570, line 4: I am unsure if `Dockerfile` should use `registry.svc.ci.openshift.org/ocp/4.0:base` or `registry.svc.ci.openshift.org/ocp/4.6:base` in the next build stage (technically no diff in this file, but it would be a diff from `Dockerfile.rhel7`).

PR #469, opened 2020-09-29
Bug 1883701: Bump openshift/api

dmage:
This PR vendors new openshift/api to get new CRDs.


manifests/00-custom-resource-definition.yaml
```diff
@@ -10,1028 +11,1024 @@ spec:
     listKind: IngressControllerList
     plural: ingresscontrollers
     singular: ingresscontroller
-  preserveUnknownFields: false
-  scope: ""
-  subresources:
-    scale:
-      labelSelectorPath: .status.selector
-      specReplicasPath: .spec.replicas
-      statusReplicasPath: .status.availableReplicas
```
Miciah, line 19: Where did this go?

Miciah:
> [openshift/api#736](https://github.com/openshift/api/pull/736) added restrictions for logLevel values. This PR vendors new openshift/api to get new CRDs.

The `IngressController` API doesn't have any `LogLevel` field, so this PR is not needed on that account.  That said, it would be good to bump to get the v1 CRD.  We need to address https://github.com/openshift/cluster-ingress-operator/pull/469/files#r496931106 though.

manifests/00-custom-resource-definition.yaml
dmage: It seems it was removed at https://github.com/openshift/api/pull/720

sttts: tagged the PR.

dmage: bumped openshift/api again

Miciah:
Thanks!
/approve

PR #474, opened 2020-10-12
TestForwardedHeaderPolicy*: Make case-insensitive

Miciah:
#### `TestForwardedHeaderPolicy*`: Show pod logs if failed

* `test/e2e/forwarded_header_policy_test.go` (`testRouteHeaders`): If the polling loop fails, output the client pod's spec and logs as part of the failure message.

#### `TestForwardedHeaderPolicy*`: Make case-insensitive

Ignore case when comparing the expected response and the actual response in the Forwarded header policy E2E tests. https://github.com/openshift/router/pull/194 turns off HAProxy's HTX option when HTTP/2 is not enabled.  As a result, HAProxy does not down-case HTTP header names.

* `test/e2e/forwarded_header_policy_test.go` (`testRouteHeaders`): Ignore case when comparing the expected response and the actual response.

### DIFF
--- a/test/e2e/forwarded_header_policy_test.go
+++ b/test/e2e/forwarded_header_policy_test.go
@@ -91,7 +91,24 @@ func testRouteHeaders(t *testing.T, image string, route *routev1.Route, address 
 		return numMatches == expectedMatches, nil
 	})
 	if err != nil {
-		t.Fatalf("failed to observe the expected output: %v", err)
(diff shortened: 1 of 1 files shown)


Miciah:
Retconning a BZ to facilitate a backport to 4.6.
/retitle Bug 1887934: TestForwardedHeaderPolicy*: Make case-insensitive

PR #476, opened 2020-10-15
NE-199 Phase 1: Add initial Canary Controller and Canary Resources

sgreene570:
This PR adds an initial canary controller in `pkg/operator/controller/canary/` that manages Ingress canary resources in the `openshift-ingress-canary` namespace. The canary controller reconciles a daemonset, service, and route to expose the `hello-openshift` container. Phase 2 of [NE-199](https://issues.redhat.com/browse/NE-199) will implement the network checks and metrics/status reporting in the canary controller.


sgreene570:
```
 unable to parse image registry.build01.ci.openshift.org/ci-op-qzf3i2tn/stable@sha256:ac7e21e580c9b7b53a7c93c407967898e2e7420f4b41c2a350e896d860804983: cannot retrieve image configuration for manifest sha256:ac7e21e580c9b7b53a7c93c407967898e2e7420f4b41c2a350e896d860804983: received unexpected HTTP status: 500 Internal Server Error
```

Not sure if there's registry issues again, or if those test failures were one offs
/retest

sgreene570:
rebased.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		Config: config,
+		client: mgr.GetClient(),
+		cache:  mgr.GetCache(),
+	}
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+
+	if err := c.Watch(&source.Kind{Type: &operatorv1.IngressController{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
sgreene570, line 54: It's probably a little excessive to watch and enqueue reconciles for IngressController events. This might be something that I revisit in a follow up PR (unless there's an obvious way to trigger the canary controllers first reconcile that i'm missing :grin:).

sgreene570:
```
level=error msg=Error: Error launching source instance: InvalidParameterValue: Value (ci-op-kc8b4rtg-2945d-cxxvb-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name

```
/retest

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if haveRoute {
+		return true, current, nil
+	} else {
+		err := r.createCanaryRoute(desired)
+		if err != nil {
+			return false, nil, err
+		}
+	}
```
Miciah, line 34:
```suggestion
	if haveRoute {
		return true, current, nil
	}
	if err := r.createCanaryRoute(desired); err != nil {
		return false, nil, err
	}
```

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	Config
+
+	client client.Client
+	cache  cache.Cache
+}
+
+// Switch the current RoutePort that the route points to.
+// Use this function to periodically update the canary route endpoint
+// to verify if the router has wedged.
+func (r *reconciler) rotateRouteEndpoint(service *corev1.Service, current *routev1.Route) (*routev1.Route, error) {
```
Miciah, line 186: Looks like nothing calls this right now.  Is that going to be in phase 2?

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,126 @@
...
+	log.Info("deleted canary route", "namespace", route.Namespace, "name", route.Name)
+	return true, nil
+}
+
+// canaryRouteChanged returns true if current and expected differ in Spec
+func canaryRouteChanged(current, expected *routev1.Route) (bool, *routev1.Route) {
+	changed := false
+	updated := current.DeepCopy()
+
+	if !cmp.Equal(current.Spec, expected.Spec, cmpopts.EquateEmpty()) {
```
Miciah, line 91: This will cause false positives, for example, when the API defaults `spec.host`.

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,126 @@
...
+// ensureCanaryRoute ensures the ingress canary route exists
+func (r *reconciler) ensureCanaryRoute(service *corev1.Service) (bool, *routev1.Route, error) {
+	desired := desiredCanaryRoute(service)
+	haveRoute, current, err := r.currentCanaryRoute()
+	if err != nil {
+		return false, nil, err
+	}
+
+	if haveRoute {
+		return true, current, nil
```
Miciah, line 28: Any reason not to call `updateCanaryRoute` here?

pkg/operator/controller/canary/route_test.go
```diff
@@ -0,0 +1,96 @@
...
+	routeTarget := route.Spec.Port.TargetPort
+	validTarget := false
+	for _, port := range service.Spec.Ports {
+		if cmp.Equal(routeTarget, port.TargetPort) {
+			validTarget = true
+		}
+	}
+
+	if !validTarget {
+		t.Errorf("Expected route targetPort to be a port in the parameter service. Route targetPort not in service targetPort list")
```
Miciah, line 55: You can use `t.Error` here... or you could print `route.Spec.Port.TargetPort` and `service.Spec.Ports` in the error message.

pkg/operator/controller/canary/service.go
```diff
@@ -0,0 +1,74 @@
...
+		return false, nil, err
+	}
+	if haveService {
+		return true, current, nil
+	} else {
+		err := r.createCanaryService(desired)
+		if err != nil {
+			return false, nil, err
+		}
+	}
```
Miciah, line 29:
```suggestion
	if haveService {
		return true, current, nil
	}
	if err := r.createCanaryService(desired); err != nil {
		return false, nil, err
	}
```

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if !changed {
+		return false, nil
+	}
+	return true, updated
+}
+
+// desiredCanaryRoute returns the desired canary route read in
+// from manifests
+func desiredCanaryRoute(service *corev1.Service) *routev1.Route {
```
Miciah, line 104: It seems to me that both the service and the route should be owned by the deployment.  Any reason why the service should be but the route should not be?

pkg/operator/controller/canary/service_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ownerRefs := service.OwnerReferences
+
+	if len(ownerRefs) != 1 {
+		t.Errorf("Expected service owner references to be of length 1, got %d", len(ownerRefs))
+	}
+
+	if !cmp.Equal(ownerRefs[0], deploymentRef) {
+		t.Errorf("Expected service owner reference name %q, but got %q", deploymentRef.Name, ownerRefs[0].Name)
+	}
```
Miciah, line 54:
```suggestion
	expectedOwnerRefs := []metav1.OwnerReference{deploymentRef}
	if !cmp.Equal(service.OwnerReferences, expectedOwnerRefs) {
		t.Errorf("Expected service owner references %#v, but got %#v", expectedOwnerRefs, service.OwnerReferences)
	}
```

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+	updated, err := chooseRandomServicePort(service, current)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to rotate route port: %v", err)
+	}
+
+	// update route resource here
```
Miciah, line 193: The method name already says as much as this comment.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+type reconciler struct {
+	Config
+
+	client client.Client
+	cache  cache.Cache
+}
+
+// Switch the current RoutePort that the route points to.
+// Use this function to periodically update the canary route endpoint
+// to verify if the router has wedged.
```
Miciah, line 185: BTW, this is pretty clever.

assets/canary/deployment.yaml
```diff
@@ -0,0 +1,19 @@
+# Hello Openshift Ingress Canary deployment
+# Specific values are set at runtime
+kind: Deployment
```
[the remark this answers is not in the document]
sgreene570: Let's go with a daemonset so we can cover connectivity to all schedulable nodes (rather than just 1+ w/ a deployment).

assets/canary/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: openshift-ingress-canary
+  annotations:
+    openshift.io/node-selector: ""
```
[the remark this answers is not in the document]
sgreene570: I believe I copied the router's namespace asset to make this one, so no, I don't think this is necessary at the moment.

manifests/00-cluster-role.yaml
```diff
@@ -136,13 +136,13 @@ rules:
...
 # Mirrored from assets/router/cluster-role.yaml
+# and expanded for assets/canary/route.yaml
 - apiGroups:
   - route.openshift.io
   resources:
   - routes
   verbs:
-  - list
-  - watch
+  - '*'
```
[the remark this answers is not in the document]
sgreene570: I might be missing something, but how would the operator be able to take advantage of a new role + role binding in the `openshift-ingress-canary` namespace? The canary controller (which resides within the ingress operator) needs to be able to perform all (well, most at least, so maybe the "*" isn't necessary) actions on routes.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+	availablePorts := []intstr.IntOrString{}
+	// Filter currentPort from servicePorts
+	for _, port := range servicePorts {
+		if !cmp.Equal(port.TargetPort, currentPort.TargetPort) {
+			availablePorts = append(availablePorts, port.TargetPort)
+		}
+	}
+
+	updated.Spec.Port = &routev1.RoutePort{
+		TargetPort: availablePorts[rand.Intn(len(availablePorts))],
```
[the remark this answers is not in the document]
sgreene570: My thinking was we might end up expanding a canary service to have > 2 ports. But since that isn't going to happen for now, ill convert `chooseRandomServicePort` to `cycleServicePort` as you suggested.

pkg/manifests/manifests.go
```diff
@@ -57,6 +66,10 @@ const (
 	DefaultOperatorNamespace = "openshift-ingress-operator"
 	DefaultOperandNamespace  = "openshift-ingress"
 
+	// DefaultCanaryNamespace is the default namespace for
+	// the ingress canary check resources.
+	DefaultCanaryNamespace = "openshift-ingress-canary"
```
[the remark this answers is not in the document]
sgreene570: I was going off of `DefaulfOperatorNamespace` and `DefaultOperandNamespace` for consistency, so let's address this is in a follow up as you mentioned.

manifests/00-cluster-role.yaml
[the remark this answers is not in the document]
Miciah: The role would be in the `openshift-ingress-canary` namespace and allow full control over routes, and the role binding bind that role to the operator's service account.  Because it would be a role (not a clusterrole) in the `openshift-ingress-canary` namespace, the role binding should only grant access to routes within that namespace, if I understand correctly.  Something like the following should work (but I have not verified that it does):
```yaml
---
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: ingress-operator-canary-controller
  namespace: openshift-ingress-canary
  annotations:
rules:
- apiGroups:
  - route.openshift.io
  resources:
  - routes
  verbs:
  - *
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: ingress-operator-canary-controller
  namespace: openshift-ingress-canary
  annotations:
subjects:
- kind: ServiceAccount
  name: ingress-operator
roleRef:
  kind: Role
  apiGroup: rbac.authorization.k8s.io
  name: ingress-operator-canary-controller
```

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+var (
+	log = logf.Logger.WithName(controllerName)
+)
+
+// New creates the ingress canary controller.
+//
+// The canary controller will watch the Default IngressController, as well as
+// the canary service, deployment, and route resources.
```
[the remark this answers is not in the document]
sgreene570: My thinking was the canary controller should watch the canary resources for any changes, and trigger a reconcile if someone (or something?) were to modify a canary resource. But sounds that won't be necessary since the canary resources are within their own namespace, and I could see someone making the case for cluster admins having control over the resources anyways without fighting the operator.

For phase 2, my thoughts were to run the route test on a regular interval (~1 minutes, perhaps) via a go routine. For now ill drop the watches on all resources except the default ingress controller.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+		errors = append(errors, err)
+		return result, utilerrors.NewAggregate(errors)
+	}
+	if !haveService {
+		errors = append(errors, fmt.Errorf("failed to get canary service"))
+		return result, utilerrors.NewAggregate(errors)
+	}
+
+	if haveRoute, _, err := r.ensureCanaryRoute(service); err != nil {
+		if err != nil {
```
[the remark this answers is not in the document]
sgreene570: whoops, thanks!

pkg/operator/controller/canary/controller.go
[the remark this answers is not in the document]
Miciah: Ah, the watch is so the canary controller can revert external changes to the resources— that makes perfect sense, and I don't know why I didn't realize it before.  * grin *.  Feel free to keep the watches, or drop them if you prefer (maybe it makes sense to allow modification of the canary resources for troubleshooting purposes?).

pkg/operator/controller/canary/controller.go
sgreene570: yes, `rotateRouteEndpoints` and `chooseRandomServicePort` are for phase 2.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,231 @@
...
+// Config holds all the things necessary for the controller to run.
+type Config struct {
+	Namespace   string
+	CanaryImage string
+}
+
+// reconciler handles the actual canary reconciliation logic in response to
+// events.
+type reconciler struct {
+	Config
```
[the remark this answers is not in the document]
sgreene570: Ill address this now while im pushing a large batch of fixups.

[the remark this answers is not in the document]
sgreene570: Ah, I see, `Config` was public so that the operator could create the Config structs in pkg/operator/operator.go (and pass them to individual controllers when calling `New(...)`). With this in mind, would you still want to switch each controller's `Config` from public to private in a follow up fix?

pkg/operator/controller/canary/controller.go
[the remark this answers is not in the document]
sgreene570: > maybe it makes sense to allow modification of the canary resources for troubleshooting purposes?

Let's go with this suspicion for now. :grin:

manifests/00-cluster-role.yaml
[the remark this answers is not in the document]
sgreene570: Ill include these resources in the manifests in my next push and verify that they work. Thanks!

pkg/operator/controller/canary/controller.go
sgreene570: Thanks! I hope this trick helps cluster admins know precisely when the router is wedged.

pkg/operator/controller/canary/route.go
sgreene570: Fixed!

manifests/00-cluster-role.yaml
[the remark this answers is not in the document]
sgreene570: @Miciah the CVO does not (and in my opinion, probably should not) create the `openshift-ingress-canary`, so the CVO cannot create a Role in the `openshift-ingress-canary` namespace. Additionally, the operator would also need similar route RBAC permissions to be able to create a role in the `openshift-ingress-canary` namespace for routes.

So I see 2 solutions:
1) The CVO creates the `openshift-ingress-canary` namespace, and then creates the ingress canary controller role.
(This seems like an awkward thing to do especially if the canary controller isnt a mandatory component of the operator)
2) We fall back to just giving the operator '*' permissions on routes in it's `cluster-role`.

Thoughts?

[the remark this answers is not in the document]
sgreene570: Note, ive reverted this PR to use `*` permissions for routes in the operator's cluster role

sgreene570:
commit `https://github.com/openshift/cluster-ingress-operator/commit/89d3bca6284134967c94781403936a2b2e85d11f` adds a predicate filter to the canary controller so reconciles are only triggered on events relating to the default ingress controller.

sgreene570:
`fail [github.com/openshift/origin/test/extended/operators/operators.go:94]: Nov  4 16:03:01.486: Some cluster operators are not ready: marketplace (missing: Degraded)`
Ingress operator looks fine.
/retest

sgreene570:
```
Container artifacts is not ready with reason ImagePullBackOff and message Back-off pulling image "busybox"
```
:thinking: 
/retest

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		errors = append(errors, err)
+		return result, utilerrors.NewAggregate(errors)
+	}
+	if !haveService {
+		errors = append(errors, fmt.Errorf("failed to get canary service"))
+		return result, utilerrors.NewAggregate(errors)
+	}
+
+	if haveRoute, _, err := r.ensureCanaryRoute(service); err != nil {
```
danehans, line 103: Change the returned values of `ensureCanaryRoute()` to the following if the *route is not being used:
```
(bool, error)
```

pkg/operator/controller/canary/controller.go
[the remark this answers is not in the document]
danehans: > With this in mind, would you still want to switch each controller's Config from public to private in a follow up fix?

I say yes, create a BZ or Jira card to track the issue and xref this convo.

pkg/operator/controller/canary/controller.go
danehans: Why not remove the unused funcs and store them in a WIP phase 2 branch instead of including them in this PR?

pkg/manifests/manifests.go
```diff
@@ -174,6 +187,38 @@ func MetricsRoleBinding() *rbacv1.RoleBinding {
 	return rb
 }
 
+func CanaryNamespace() *corev1.Namespace {
+	ns, err := NewNamespace(MustAssetReader(CanaryNamespaceAsset))
+	if err != nil {
+		panic(err)
```
frobware, line 193: Adding panics makes things a little harder to test. Can these not return error?

pkg/operator/controller/canary/daemonset_test.go
```diff
@@ -0,0 +1,60 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestDesiredCanaryDaemonSet(t *testing.T) {
+	canaryImage := "openshift/hello-openshift:latest"
+
+	daemonset := desiredCanaryDaemonSet(canaryImage)
+
+	expectedDaemonSetName := types.NamespacedName{
```
danehans, line 17: Is this the same ns/name as `controller.CanaryDaemonSetName()`? If so, I suggest calling the function instead.

pkg/operator/controller/canary/daemonset_test.go
```diff
@@ -0,0 +1,60 @@
...
+	if !cmp.Equal(daemonset.Name, expectedDaemonSetName.Name) {
+		t.Errorf("Expected daemonset name to be %s, but got %s", expectedDaemonSetName.Name, daemonset.Name)
+	}
+
+	if !cmp.Equal(daemonset.Namespace, expectedDaemonSetName.Namespace) {
+		t.Errorf("Expected daemonset namespace to be %s, but got %s", expectedDaemonSetName.Namespace, daemonset.Namespace)
+	}
+
+	expectedLabels := map[string]string{
+		"ingress.openshift.io/canary": "canary_controller",
```
danehans, line 31: If this label is used elsewhere, I suggest creating a const for the strings.

manifests/02-deployment.yaml
```diff
@@ -58,6 +60,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: IMAGE
               value: openshift/origin-haproxy-router:v4.0
+            - name: CANARY_IMAGE
+              value: openshift/hello-openshift:latest
```
frobware, line 64: Should we be specifying latest? It seems less deterministic.

pkg/operator/controller/canary/daemonset_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if !cmp.Equal(daemonset.Labels, expectedLabels) {
+		t.Errorf("Expected daemonset labels to be %q, but got %q", expectedLabels, daemonset.Labels)
+	}
+
+	labelSelector := &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app": "ingress-canary",
```
danehans, line 40: Should we follow the [k8s label recommendations](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) instead of "app: ingress-canary"?

manifests/02-deployment.yaml
sgreene570: The CVO will override this value anyways, so I think `latest` is fine.

pkg/manifests/manifests.go
sgreene570: These new functions follow along with all of the other asset creation functions. IMO, we should panic here since being unable to read any of the operator's resources in via `assets` is very concerning.

pkg/operator/controller/canary/namespace.go
```diff
@@ -0,0 +1,27 @@
...
+
+// ensureCanaryNamespace ensures that the ingress-canary namespace exists
+func (r *reconciler) ensureCanaryNamespace() error {
+	ns := manifests.CanaryNamespace()
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to get ingress canary namespace %q: %v", ns.Name, err)
+		}
+		if err := r.client.Create(context.TODO(), ns); err != nil {
+			return fmt.Errorf("failed to create ingress canary namespace %q: %v", ns.Name, err)
```
danehans, line 21: I think the func is easier to read and streamlined when:
```
	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
		if errors.IsNotFound(err) {
			return r.client.Create(context.TODO(), ns)
		}
                return fmt.Errorf("failed to get ingress canary namespace %q: %v", ns.Name, err)
```

The same approach can be applied to all ensure funcs.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Reconcile ensures that the canary controller's resources
+// are in the desired state.
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	result := reconcile.Result{}
+	errors := []error{}
+
+	if err := r.ensureCanaryNamespace(); err != nil {
+		errors = append(errors, err)
+		return result, utilerrors.NewAggregate(errors)
```
frobware, line 71: The returns (generally) could become just:

    return result, utilerrors.NewAggregate(append(errors, err))

pkg/operator/controller/canary/controller.go
sgreene570: Is this suggestion really necessary? A follow up PR might need to use a field from the canary route resource. Additionally, this pattern is used for the ingress controller logic https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/controller.go#L596.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	client client.Client
+}
+
+// Switch the current RoutePort that the route points to.
+// Use this function to periodically update the canary route endpoint
+// to verify if the router has wedged.
+func (r *reconciler) rotateRouteEndpoint(service *corev1.Service, current *routev1.Route) (*routev1.Route, error) {
+	updated, err := cycleServicePort(service, current)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to rotate route port: %v", err)
```
frobware, line 134: What are our rules (or existing behaviour) for capitalising error message? (Do we even have any?)

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	_, err = r.updateCanaryRoute(current, updated)
+	if err != nil {
+		return current, err
+	}
+
+	return updated, nil
+}
+
+// cycleServicePort returns a route resource with Spec.Port set to a port
+// selected from service.Spec.Ports, excluding route.Spec.Port to gaurantee
```
frobware, line 146: typo: guarantee

pkg/operator/controller/canary/controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for _, tc := range testCases {
+		route, err := cycleServicePort(tc.service, tc.route)
+		if tc.success {
+			if err != nil {
+				t.Errorf("Expected test case %s to not return an err, but got err %v", tc.description, err)
+			}
+
+			if cmp.Equal(tc.route.Spec.Port, route.Spec.Port) {
+				t.Errorf("Expected route to have a new port assignment. Still has port %s", route.Spec.Port)
```
frobware, line 92: Log the expected value as well.

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,140 @@
...
+		// associate the route with the canary controller
+		manifests.OwningIngressCanaryCheckLabel: controllerName,
+	}
+
+	route.Spec.To.Name = controller.CanaryServiceName().Name
+
+	// Set spec.port.targetPort to the first port available in the canary service.
+	// The canary controller will toggle which targetPort the router targets
+	// to test > 1 endpoint, so it does not matter which port is selected as long
+	// as the canary service has > 1 ports available.
```
frobware, line 134: Are there any implications to document if this doesn't hold true?

pkg/operator/controller/canary/route.go
```diff
@@ -0,0 +1,140 @@
...
+	case haveRoute:
+		if updated, err := r.updateCanaryRoute(current, desired); err != nil {
+			return true, current, err
+		} else if updated {
+			return r.currentCanaryRoute()
+		}
+	}
+
+	return true, current, nil
+}
```
danehans, line 42: I always find this control flow difficult to understand. Can this be simplified as follows:

```
func (r *reconciler) ensureCanaryRoute(service *corev1.Service) (bool, *routev1.Route, error) {
	desired := desiredCanaryRoute(service)

	current, err := r.currentCanaryRoute()
	if err != nil {
		if errors.IsNotFound(err) {
			return r.createCanaryRoute()
		}
		return err
	}

        // updateRouteIfNeeded performs the comparison and updates the object in
        // the API server if they differ.
	if err := r.updateRouteIfNeeded(current, desired); err != nil {
		return err
	}

	return nil
}
```

A BZ or Jira card can be created as tech debt to address the control flow in other areas of ingress/dns operators.

danehans:
Is an e2e test being added to this PR or a follow-on PR?

sgreene570:
> Is an e2e test being added to this PR or a follow-on PR?

Follow up PR will add an e2e test to verify that the canary daemonset is reachable via routes.

pkg/operator/controller/canary/daemonset_test.go
sgreene570: fixed. Thanks!

pkg/operator/controller/canary/daemonset_test.go
sgreene570: Fixed. Thanks!

pkg/operator/controller/canary/daemonset_test.go
sgreene570: fixed.

pkg/operator/controller/canary/namespace.go
sgreene570: Interesting. My main concern with this approach is that the responsibility to log successful resource creation is pushed back to the caller. Thoughts?

pkg/operator/controller/canary/controller.go
sgreene570: Not sure if it's formally documented anywhere, but I'll be sure to stick to lowercase.

pkg/operator/controller/canary/controller.go
sgreene570: Good point.

pkg/operator/controller/canary/controller.go
sgreene570: fixed.

pkg/operator/controller/canary/route.go
sgreene570: Updated the comment

pkg/operator/controller/canary/controller_test.go
sgreene570: updated the test.

manifests/02-deployment.yaml
sgreene570: For local testing, the latest tag is fine.

pkg/operator/controller/canary/controller.go
sgreene570: fixed, thanks!

pkg/operator/controller/canary/route.go
sgreene570: This example flow leaves out the call to `r.currentCanaryRoute()` _after_ the call to `r.updateCanaryRoute`. From my understanding, it is important that we return the newest version of the resource after the update has been performed. 

Additionally, the current switch statement format is useful for situations where there is both a `haveFoo` and `wantFoo` boolean (this is the case with the ingress operator's [ensureLoadBalancer func](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/load_balancer_service.go#L161-L197)).

sgreene570: I guess most of this is subjective. I definitely prefer the boolean based control flow, and I believe we have been trying to work towards standardizing on that format across both of our operators, but I could be wrong. @Miciah might have more context to that end. @danehans if you feel strongly about changing all of our `ensureFoo` functions to follow a new control flow, would you mind opening the BZ/jira ticket and bringing it up with the team? I'm not entirely sure if I follow what it is you have in mind.

pkg/operator/controller/canary/controller.go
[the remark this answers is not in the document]
sgreene570: Done
https://issues.redhat.com/browse/NE-448

pkg/operator/controller/canary/controller.go
sgreene570: In retrospect I probably should not have committed functions that aren't called by anything.

That being said, the functions have already been reviewed, and will be utilized soon.
@danehans would it be OK if we merge them now with the understanding that they will be used as a part of NE-199 phase 2, which is coming shortly?

manifests/00-cluster-role.yaml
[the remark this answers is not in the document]
sgreene570: @Miciah https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping is the solution we are looking for! I amended this PR so that the ingress operator is granted `bind` and `escalate` permissions on `roles` and `rolebindings` in it's cluster role. I also added the `role.yaml` and `role-binding.yaml` assets for the ingress canary. With these new changes, the ingress operator's route permissions are extended to `*` for the `openshift-ingress-canary` namespace _only_. PTAL when you can.

manifests/00-cluster-role.yaml
```diff
@@ -59,6 +60,14 @@ rules:
   - watch
   - update
 
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  verbs:
+  - escalate
+  - bind
```
sgreene570, line 69: See https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-escalation-prevention-and-bootstrapping

assets/canary/role-binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: canary-role
+  namespace: openshift-ingress-canary
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: canary-role
+subjects:
+- kind: ServiceAccount
+  name: ingress-operator
```
danehans, line 12: Does the canary daemonset need to be configured to use this service account?

hack/run-local.sh
```diff
@@ -14,5 +15,5 @@ echo "Image: ${IMAGE}"
 echo "Release version: ${RELEASE_VERSION}"
 echo "Namespace: ${NAMESPACE}"
 
-${DELVE:-} ./ingress-operator start --image "${IMAGE}" --release-version "${RELEASE_VERSION}" \
+${DELVE:-} ./ingress-operator start --image "${IMAGE}" --canary-image "${CANARY_IMAGE}" --release-version "${RELEASE_VERSION}" \
```
danehans, line 18: According to [this](https://github.com/openshift/cluster-ingress-operator/pull/476/files#diff-24ce9f24fd9e7e47b7ac69fa0b597fe0470c1ba7b8257d0fa6537c9caf1b5deeR64), `canary-image` is optional. If this is the case, the flag should be set optionally in the `hack/*` scripts.

manifests/00-cluster-role.yaml
```diff
@@ -59,6 +60,17 @@ rules:
   - watch
   - update
 
+# Add escalate and bind verbs for role/rolebindings.
+# This way, the ingress canary controller can safely posses wider route permissions than the ingress operator.
```
danehans, line 64: Is `posses` a typo?

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Reconcile ensures that the canary controller's resources
+// are in the desired state.
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	result := reconcile.Result{}
+	errors := []error{}
+
+	if err := r.ensureCanaryNamespace(); err != nil {
+		return result, utilerrors.NewAggregate(append(errors, err))
```
danehans, line 69: Instead of returning early after `ensureCanaryNamespace()`, I would expect `err` to be appended to `errors` so additional `ensureX()` functions are called. For example:

```
	if err := r.ensureCanaryNamespace(); err != nil {
                errors = append(errors, fmt.Errorf("my error msg: %v", err))
        }
...
       return utilerrors.NewAggregate(errs)
```

pkg/operator/controller/canary/controller.go
danehans: It's not required, but in this case and the case you reference can cause confusion. If a returned value is not being used, then it should not be returned.

pkg/operator/controller/canary/controller.go
danehans: I prefer that they're removed, but I won't block the PR if they stay. I suggest at least adding a TODO to any unimplemented functions in case phase 2 gets pushed out for some reason.

pkg/operator/controller/canary/daemonset.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+// ensureCanaryDaemonSet ensures the canary daemonset exists
+func (r *reconciler) ensureCanaryDaemonSet() (bool, *appsv1.DaemonSet, error) {
+	desired := desiredCanaryDaemonSet(r.Config.CanaryImage)
+	haveDepl, current, err := r.currentCanaryDaemonSet()
```
danehans, line 17: `haveDepl` can be a misleading name for a DaemonSet, I suggest using `have` or `haveDs`.

pkg/operator/controller/canary/namespace.go
danehans: Would this be acceptable:
```
if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
    if errors.IsNotFound(err) {
        // The namespace doesn't exist, so create it.
        if err := r.client.Create(context.TODO(), ns); err != nil {
            return fmt.Errorf("failed to create ingress canary namespace %q: %v", ns.Name, err)
    }
    log.Info("created ingress canary namespace", "name", ns.Name)
    return fmt.Errorf("failed to get ingress canary namespace %q: %v", ns.Name, err)
}
```

sgreene570: In your above example, `log.Info` is called to log creation success even when creation has failed. Maybe you mean something like this?
```go
if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
    if errors.IsNotFound(err) {
        // The namespace doesn't exist, so create it.
        if err := r.client.Create(context.TODO(), ns); err != nil {
            log.Info("created ingress canary namespace", "name", ns.Name)
            return fmt.Errorf("failed to create ingress canary namespace %q: %v", ns.Name, err)
    } else {
            return fmt.Errorf("failed to get ingress canary namespace %q: %v", ns.Name, err)
    }
}
```

assets/canary/role-binding.yaml
sgreene570: At the moment, no. The canary-role gives the ingress-operator service account (and thus, the operator) extended permissions for routes in the `openshift-ingress-canary` namespace only. The operand (hello-openshift in this case) does not need any RBAC permissions.

pkg/operator/controller/canary/daemonset.go
sgreene570: Whoops. Fixed.

pkg/operator/controller/canary/controller.go
sgreene570: Good catch. Fixed.

manifests/00-cluster-role.yaml
sgreene570: Reworded the comment.

hack/run-local.sh
sgreene570: Good point. I've updated the `hack/run-local.sh` script accordingly.

sgreene570:
Thanks for the reviews @danehans . I've updated this PR to address your comments. 

Would like to get @Miciah's input on the route RBAC changes before an lgtm (since Miciah posed the original question on route permissions for the canary controller).

pkg/operator/controller/canary/controller.go
Miciah: The most formal documentation I can find is https://github.com/golang/go/wiki/CodeReviewComments#error-strings.  Lower-case for errors and upper-case for logs is the prevalent convention in the Go world.

pkg/operator/controller/canary/route.go
Miciah: I find using a `switch` more readable because there are 2 Boolean variables (have X and want X) → 4 cases → 4 actions (no-op, create, delete, or update).  We have been trying to standardize this convention with the notion that maybe someday we can factor out the "ensure" logic into a dependency tree of resources, where each resource defines its "desired' function and CRUD operators.  We haven't done that refactoring, but IMO the consistency is a win.

pkg/operator/controller/canary/controller.go
Miciah: You cannot create a resource in the namespace if the namespace does not exist, so it makes sense to return early if `ensureCanaryNamespace` returns an error.

assets/canary/daemonset.yaml
```diff
@@ -0,0 +1,24 @@
+# Hello Openshift Ingress Canary daemonset
+# Specific values are set at runtime
+kind: DaemonSet
+apiVersion: apps/v1
+# name and namespace are set at runtime.
+spec:
+  progressDeadlineSeconds: 600
+  template:
+    spec:
+      priorityClassName: system-cluster-critical
```
Miciah, line 10: `priorityClassName: system-cluster-critical` is debatable.  Is this a deliberate choice, or copy and paste?

hack/run-local.sh
```diff
@@ -14,5 +14,11 @@ echo "Image: ${IMAGE}"
 echo "Release version: ${RELEASE_VERSION}"
 echo "Namespace: ${NAMESPACE}"
 
-${DELVE:-} ./ingress-operator start --image "${IMAGE}" --release-version "${RELEASE_VERSION}" \
+if [[ -v "ENABLE_CANARY" ]]; then
```
Miciah, line 17: TIL:

> f.  test/[/[[ have a new -v variable unary operator, which returns success if
>    `variable' has been set.

https://tiswww.case.edu/php/chet/bash/CHANGES

`[[ -v ]]` is new in Bash 4.2, so this will break on macOS.

hack/run-local.sh
```diff
@@ -14,5 +14,11 @@ echo "Image: ${IMAGE}"
...
+    CANARY_ARG="--canary-image"
+    CANARY_IMAGE=$(oc get -n openshift-ingress-operator deployments/ingress-operator -o json | jq -r '.spec.template.spec.containers[0].env[] | select(.name=="CANARY_IMAGE").value')
+    echo "Canary Image: ${CANARY_IMAGE}"
+fi
+
+${DELVE:-} ./ingress-operator start --image "${IMAGE}" ${CANARY_ARG:-} ${CANARY_IMAGE:-} --release-version "${RELEASE_VERSION}" \
```
Miciah, line 23: This logic breaks if the user specifies `CANARY_IMAGE` directly.  You could use `${CANARY_IMAGE:+--canary-image=${CANARY_IMAGE}}`, or even simpler, `--canary-image=${CANARY_IMAGE:-}` (the equals sign ensures the subsequent parameter is not used as the value, so the empty string is used as the parameter value for `--canary-image`).  

I don't personally use this shell script though, so don't waste time on this if you are happy with it as you have it.

pkg/operator/controller/canary/controller.go
sgreene570: Fixed.

manifests/00-cluster-role.yaml
```diff
@@ -59,6 +60,18 @@ rules:
   - watch
   - update
 
+# Add escalate and bind verbs for role/rolebindings.
+# This allows the operator to grant broad route permissions to itself in the ingress-canary namespace
+# without holding the same permissions in the operator namespace.
```
Miciah, line 65: Does this mean that the operator now has the ability to create an arbitrary clusterrole and then bind its own serviceaccount to that clusterrole?  If so, this effectively gives the operator unbound permissions.

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if err := r.ensureCanaryRole(); err != nil {
+		errors = append(errors, fmt.Errorf("failed to ensure canary role: %v", err))
+	}
+
+	haveDs, daemonset, err := r.ensureCanaryDaemonSet()
+	if err != nil {
+		errors = append(errors, fmt.Errorf("failed to ensure canary daemonset: %v", err))
+	}
+	if !haveDs {
```
Miciah, line 82:
```suggestion
	} else if !haveDs {
```

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,174 @@
...
+		Name:       daemonset.Name,
+		UID:        daemonset.UID,
+		Controller: &trueVar,
+	}
+
+	haveService, service, err := r.ensureCanaryService(daemonsetRef)
+	if err != nil {
+		errors = append(errors, fmt.Errorf("failed to ensure canary service: %v", err))
+	}
+	if !haveService {
```
Miciah, line 99:
```suggestion
	} else if !haveService {
```

pkg/operator/controller/canary/controller.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	return updated, nil
+}
+
+// cycleServicePort returns a route resource with Spec.Port set to the
+// next available port in service.Spec.Ports that is not the current route.Spec.Port.
+func cycleServicePort(service *corev1.Service, route *routev1.Route) (*routev1.Route, error) {
+	servicePorts := service.Spec.Ports
+	currentPort := route.Spec.Port
```
Miciah, line 150: Add a null check so the operator doesn't dereference a null pointer if someone deletes the port on the route.

assets/canary/daemonset.yaml
sgreene570: copy paste artifact. Ill drop the `priorityClassName` field for now.

manifests/00-cluster-role.yaml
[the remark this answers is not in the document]
sgreene570: Going with route '*' permissions after a discussion with miciah. While these permissions are broad, this is the "least bad" approach. The above mentioned `escalate`/`bind` approach introduces far more concerning privilege escalation issues to the operator.

pkg/operator/controller/canary/controller.go
sgreene570: Fixed, thanks!!

hack/run-local.sh
sgreene570: I have updated the shell script based off of your comments.

pkg/operator/controller/canary/controller.go
sgreene570: fxied.

pkg/operator/controller/canary/controller.go
sgreene570: fixed.

sgreene570:
PR has been updated based on all of the feedback from @danehans and @Miciah . The canary role has been dropped from this PR (the operator will just have `*` permissions on routes instead, as @Miciah and I deemed this to be the "least bad" approach).

Miciah:
Thanks!  Sorry for holding this up for so long.  Let's merge it and iterate.  
/lgtm

sgreene570:
Not sure where the retest bot went.

/retest

sgreene570:
Whoops, accidentally clicked the "close" button instead of "comment" :cry:

sgreene570:
Not sure why the retest bot is absent still
/refresh
/skip

PR #478, opened 2020-10-23
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#414 and doesn't change existing behavior


sgreene570:
@guillaumerose could you help me understand precisely what adding the `include.release.openshift.io/self-managed-high-availability: "true"` annotation does to the ingress operator's manifests?

Also, I think you need to run `make update` and commit the bindata changes so this PR passes the `verify` job.

deads2k:
This matches the name in https://github.com/openshift/enhancements/blob/master/enhancements/update/cluster-profiles.md#proposal .

@sgreene570 The CVO is going to transition from opt-out cluster profiles to opt-in cluster profiles.  This means that the @openshift/sig-network-edge team will get to opt-in for all future cluster profiles instead of discovering that your manifests are suddenly part of a new payload.

The first transition requires that the current profile `include.release.openshift.io/self-managed-high-availability: "true"` get included in every manifest we currently ship.  It's essentially, "keep working as it does today" and future annotations for `include.release.openshift.io/***` mean that you are going to be supporting a new payload in a potentially new topology.

/lgtm

/assign @sgreene570 
for approval ack

PR #472, opened 2020-10-11
Bug 1891625: Support changing ingresscontroller load balancer scope

Miciah:
Add support for changing ingresscontroller load balancer scope.  On AWS and IBM cloud, this requires deleting the existing load balancer service and recreating it with the desired scope.   On Azure and GCP, it suffices to update the existing service's annotations.

This change also adds logic to remove the service finalizer, which is no longer needed to clean up DNS records since the DNSRecord CRD was added.

* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Update scope if needed.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`externalLBAnnotations`): New variable.  Map platform type to the annotation for that platform that makes the load balancer external, if the platform requires an explicit annotation.
(`ensureLoadBalancerService`): Add delete and update logic.  Use the new `deleteLoadBalancerServiceFinalizer` method to delete any finalizer on any existing service.  Use the new `createLoadBalancerService`, `deleteLoadBalancerService`, and `updateLoadBalancerService` methods to create, update, or delete the service as needed.
(`desiredLoadBalancerService`): Delete logic to add a finalizer to the service.  Use new externalLBAnnotations variable to simplify logic.
(`finalizeLoadBalancerService`): Refactor to use the new `deleteLoadBalancerServiceFinalizer` method.
(`createLoadBalancerService`, `deleteLoadBalancerService`): New methods.
(`updateLoadBalancerService`): New method.  Update the LoadBalancer service, using the new `loadBalancerServiceChanged` function.
(`loadBalancerServiceScopeChanged`): New function.  Check if the load balancer's scope changed.
(`loadBalancerServiceChanged`): New function.  Check if the current service needs to be updated, and if so, whether it needs to be modified or deleted and recreated.
(`deleteLoadBalancerServiceFinalizer`): New method.  Delete any finalizer from the service.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestLoadBalancerServiceChanged`): New test.
(`TestLoadBalancerServiceChangedScopeNeedsRecreate`): New test.
* `test/e2e/operator_test.go` (`isServiceInternal`): New function.  Return a Boolean value indicating whether the provided service is annotated to request an internal load-balancer.
(`TestScopeChange`): New test.  Verify that mutating scope from the default external scope to internal scope and from internal back to external succeeds on AWS, Azure, GCP, and IBM Cloud.

----

Based on #394.

PR #480, opened 2020-10-28
WIP: Set publishNotReadyAddresses to true on LB services

Miciah:
Set `spec.publishNotReadyAddresses` to true on the LoadBalancer-type service that the operator creates for ingress controllers on cloud platforms.

Setting `publishNotReadyAddresses` causes the endpointslice controller to publish endpoints of terminating pods.  Such endpoints are published with `endpoints[].conditions.terminating` set to `false.`  This causes the service proxy to continue to handle traffic as long as the endpoint exists but report negative responses to health probes if there is no other endpoint, which all facilitates graceful termination of router pods.

* `assets/router/service-cloud.yaml`: Add `publishNotReadyAddresses: true`.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestLoadBalancerServiceChanged`): Add test case to verify that `loadBalancerServiceChanged` checks and updates `publishNotReadyAddresses` as needed.

----

Depends on https://github.com/openshift/cluster-ingress-operator/pull/472, https://github.com/openshift/sdn/pull/131, and https://github.com/openshift/kubernetes/pull/431.


Miciah:
`spec.publishNotReadyAddresses` is insufficient.  The `EndpointSliceTerminatingCondition` feature gate is necessary and sufficient.  
/close

PR #481, opened 2020-10-28
Bug 1892338: Fix HAProxyReloadFail alert

sgreene570:
https://github.com/openshift/router/pull/209 reworks the HAProxy reload fails metric so that the HAProxyReloadFail alert can
be improved. The new template_router_reload_failure metric in router #209 that replaces the template_router_reload_fails metric is a simple boolean gauge metric, which allows the HAProxyReloadFail alert to fire for the duration of the HAProxy reload outage. Previously, the HAProxyReloadFail alert would fire for ~5 minutes regardless of whether or not reloads were still continuing to fail on the router.


sgreene570:
Depends on https://github.com/openshift/router/pull/209

/hold

wking (approved):

sgreene570:
https://github.com/openshift/router/pull/209 has merged, so this PR is ready to go

PR #484, opened 2020-11-02
dns: Don't hotloop on updates to DNSRecord status

Miciah:
Add a predicate to the DNS controller's watch on `DNSRecords` so that only updates a `DNSRecord's` spec triggers reconciliation (and updates to its status do not).

Before this change, one reconciliation could update the `DNSRecord's` status and trigger a second reconciliation.  In the case of a successful update, the second reconciliation would have no work to do, so it would complete without triggering a third reconciliation.  In the case of a failed update, the second reconciliation would retry the update.  If retrying the update then failed, the second reconciliation could update the `DNSRecord's` status to record the new failure, triggering a third reconciliation, which could similarly trigger a fourth reconciliation, and so on ad infinitum.

* `pkg/operator/controller/dns/controller.go` (`New`): Add a predicate to the watch on `DNSRecords` to ignore updates to status.

----

@ironcladlou

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -67,7 +67,7 @@ func New(mgr manager.Manager, config Config) (runtimecontroller.Controller, erro
 	if err != nil {
 		return nil, err
 	}
-	if err := c.Watch(&source.Kind{Type: &iov1.DNSRecord{}}, &handler.EnqueueRequestForObject{}); err != nil {
+	if err := c.Watch(&source.Kind{Type: &iov1.DNSRecord{}}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{}); err != nil {
 		return nil, err
 	}
 	if err := c.Watch(&source.Kind{Type: &configv1.DNS{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(reconciler.ToDNSRecords)}); err != nil {


Miciah:
Seems like the predicate filtered out delete events?!  At least I'm not seeing any "deleted dnsrecord" log messages in https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/484/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1323382581768491008/artifacts/e2e-aws-operator/pods/openshift-ingress-operator_ingress-operator-687f7b4fbb-q5d94_ingress-operator.log.

Miciah:
Latest push uses `GenerationChangedPredicate`, which returns `true` for create, delete, and generic events, and returns false for updates if the generation has not changed.  Maybe that will fix handling of delete events.

ironcladlou:
Thanks, lgtm although there are some assumptions baked into that predicate to consider:

```
// * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
// E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
// For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
//
// * With this predicate, any update events with writes only to the status field will not be reconciled.
// So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.
```

hopefully regression testing would reveal any possible sync issue here.

PR #485, opened 2020-11-09
aws: use shared credentials file instead of static only

abhinavdahiya:
aws: use shared credentials file instead of static only

In order to support various styles of authenitcation, mostly STSAsumeRoleUsingWebhookIdentity, the operator should support using the AWS SDK's default
handling of the shared credential file. So previously the operator only supported static credentials from specific keys in the AWS secret, but now the operator
supports reading `credentials` key to load the shared credential file and pass it on as-is to AWS SDK to authenticate based on the contents of that file.

To make the code easier, the operator now converts the previous access_key, secret_key case into a shared credential file using those values and then using
that like if it was provided.

see https://issues.redhat.com/browse/CO-1256 for more details.

 manifests: add inert serviceaccount token for later use by STS on AWS

Add a projected service account token to the ingress-operator container in the ingress-operator pod. We will use this token to get STS creds when that is wired in.
This is inert until then. see https://issues.redhat.com/browse/CO-1256

The audience `openshift` chosen is arbitrary, but platform-neutral choice. also see openshift/machine-api-operator#743 for similar consistent change.


manifests/02-deployment.yaml
```diff
@@ -95,3 +98,9 @@ spec:
           items:
             - key: ca-bundle.crt
               path: tls-ca-bundle.pem
+      - name: bound-sa-token
```
sgreene570, line 101: Could you add a comment / TODO explaining that this service account token is inert for now (as you mentioned in your commit message)?

abhinavdahiya: done with 9fec421 , added a comment on the usage of the definition of the volume.

pkg/util/aws/shared_credentials_file.go
```diff
@@ -0,0 +1,57 @@
...
+		)
+
+	default:
+		return "", errors.New("invalid secret for aws credentials")
+	}
+
+	f, err := ioutil.TempFile("", "aws-shared-credentials")
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create file for shared credentials")
+	}
```
joelddiaz, line 44: might want to add a `defer f.Close()`

pkg/util/aws/shared_credentials_file_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}, {
+		data: map[string]string{
+			"credentials": `[default]
+assume_role = this_role_for_ingress_operator
+web_identity_token = /path/to/some/file
+`,
+		},
+
+		sharedConfig: `[default]
+assume_role = this_role_for_ingress_operator
```
joelddiaz, line 66: not that it matters for the specific test cases here, but the actual AWS config field names are `role_arn` and `web_identity_token_file`

manifests/02-deployment.yaml
sgreene570: Looks like bindata needs to be updated and committed from the changes to `9fec421`

pkg/util/aws/shared_credentials_file.go
```diff
@@ -0,0 +1,57 @@
...
+		data = newConfigForStaticCreds(
+			string(secret.Data["aws_access_key_id"]),
+			string(secret.Data["aws_secret_access_key"]),
+		)
+
+	default:
+		return "", errors.New("invalid secret for aws credentials")
+	}
+
+	f, err := ioutil.TempFile("", "aws-shared-credentials")
```
Miciah, line 41: What cleans up old temporary files when the token changes?

Miciah:
`createDNSProviderIfNeeded` checks if the secret has updated, but the idea here is that the secret does not change, but rather it references a file (`/var/run/secrets/openshift/serviceaccount/token`), and the contents of that file change, right?  So `createDNSProviderIfNeeded` needs to be checking if the contents of that file changes, or else it will keep using the token even after it expires.

Miciah:
Oh, we're passing the path to the token in the AWS client config; does the AWS client automatically reread the token?

joelddiaz:
> Oh, we're passing the path to the token in the AWS client config; does the AWS client automatically reread the token?

Yes, the AWS client auto-refreshing the token as-needed is precisely what happens.

pkg/util/aws/shared_credentials_file.go
abhinavdahiya: done with b5ede51

pkg/util/aws/shared_credentials_file.go
abhinavdahiya: see https://github.com/openshift/cluster-ingress-operator/pull/485/files#diff-cfa88f3094d390a95d1d9b230dadae57aefa04e78041f7697f941ca110f7bca4R450-R452 for when cleanup happens.

PR #486, opened 2020-11-11
[release-4.6] Bug 1896167: Back port HAProxy reload failures alert fix

sgreene570:
This is the 4.6 backport of https://github.com/openshift/cluster-ingress-operator/pull/481

---

https://github.com/openshift/router/pull/209 reworks the HAProxy reload fails metric so that the HAProxyReloadFail alert can
be improved. The new template_router_reload_failure metric in router #209 that replaces the template_router_reload_fails metric is a simple boolean gauge metric, which allows the HAProxyReloadFail alert to fire for the duration of the HAProxy reload outage. Previously, the HAProxyReloadFail alert would fire for ~5 minutes regardless of whether or not reloads were still continuing to fail on the router.


Miciah:
/lgtm
Should we `/hold` for https://github.com/openshift/router/pull/215?

sgreene570:
> /lgtm
> Should we `/hold` for [openshift/router#215](https://github.com/openshift/router/pull/215)?

Yes!
/hold

PR #490, opened 2020-11-12
Bug 1896923: Configure operator metrics handler to use localhost

Miciah:
Configure the operator's metrics handler to listen on 127.0.0.1 only.

Before this change, metrics were exposed on port 60000, without authentication.  Although the pod spec does not advertise port 60000, the port was still accessible outside the pod.  After this change, port 60000 is inaccessible outside the pod, so the only way to get metrics is by connecting to port 9393 and authenticating with kube-rbac-proxy.

* `cmd/ingress-operator/start.go` (`NewStartCommand`): Configure metrics to listen on 127.0.0.1:60000.


sgreene570:
Thanks @Miciah ! 
/lgtm

PR #495, opened 2020-11-23
Bug 1870373: Ingress Operator reports available when DNS fails to provision

candita:
BZ-1870373: Ingress Operator reports available when DNS fails to provision

- Check DNSReady and LoadBalancerReady when determining Availability in computeIngressAvailableCondition
- Call computeIngressAvailableCondition after computing other conditions in syncIngressControllerStatus
- Change unit tests for computeIngressAvailableCondition
- Refactor the checks for conditions into checkConditions, to share between computeIngressAvailable and computeIngressDegraded

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -31,6 +31,17 @@ import (
 // clock is to enable unit testing
 var clock utilclock.Clock = utilclock.RealClock{}
 
+// expectedCondition contains a condition that is expected to be checked when
+// determining Available or Degraded status of the ingress controller
+type expectedCondition struct {
+	condition string
+	status    operatorv1.ConditionStatus
+	// ifConditionsTrue is a list of prerequisite conditions that should be true
+	// or else the condition is not checked.
+	ifConditionsTrue []string
+	gracePeriod      time.Duration
+}
+
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
 func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) error {
@@ -46,12 +57,12 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.Selector = selector.String()
 	updated.Status.TLSProfile = computeIngressTLSProfile(ic.Status.TLSProfile, deployment)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentPodsScheduledCondition(deployment, pods))
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDNSStatus(ic, wildcardRecord, dnsConfig)...)
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(updated.Status.Conditions))
 	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions, updated.Name)
 	errs = append(errs, err)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
@@ -172,35 +183,101 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
 }
 
 // computeIngressAvailableCondition computes the ingress controller's current Available status state
-// by inspecting the Available condition of deployment. The ingresscontroller is only available if
-// the deployment is also available.
-func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	for _, cond := range deployment.Status.Conditions {
-		if cond.Type != appsv1.DeploymentAvailable {
-			continue
-		}
-		switch cond.Status {
-		case corev1.ConditionTrue:
-			return operatorv1.OperatorCondition{
-				Type:   operatorv1.IngressControllerAvailableConditionType,
-				Status: operatorv1.ConditionTrue,
-			}
-		case corev1.ConditionFalse:
-			return operatorv1.OperatorCondition{
-				Type:    operatorv1.IngressControllerAvailableConditionType,
-				Status:  operatorv1.ConditionFalse,
-				Reason:  cond.Reason,
-				Message: "The deployment is unavailable: " + cond.Message,
-			}
-		}
+// by inspecting the following:
+// 1) the Available condition of Deployment,
+// 2) the DNSReady condition of the IngressController, and
+// 3) the LoadBalancerReady condition of the IngressController.
+// The ingresscontroller is judged Available only if all 3 conditions are true
+func computeIngressAvailableCondition(conditions []operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+	expected := []expectedCondition{
+		{
+			condition: IngressControllerDeploymentAvailableConditionType,
+			status:    operatorv1.ConditionTrue,
+		},
+		{
+			condition: operatorv1.DNSReadyIngressConditionType,
+			status:    operatorv1.ConditionTrue,
+			ifConditionsTrue: []string{
+				operatorv1.LoadBalancerManagedIngressConditionType,
+				operatorv1.LoadBalancerReadyIngressConditionType,
+				operatorv1.DNSManagedIngressConditionType,
+			},
+		},
+		{
+			condition:        operatorv1.LoadBalancerReadyIngressConditionType,
+			status:           operatorv1.ConditionTrue,
+			ifConditionsTrue: []string{operatorv1.LoadBalancerManagedIngressConditionType},
+		},
 	}
 
-	return operatorv1.OperatorCondition{
-		Type:    operatorv1.IngressControllerAvailableConditionType,
-		Status:  operatorv1.ConditionFalse,
-		Reason:  "DeploymentAvailabilityUnknown",
-		Message: "The deployment's Available condition couldn't be interpreted",
+	// Cover the rare case of no conditions
+	if len(conditions) == 0 {
+		return operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}
 	}
+	_, unavailableConditions, _ := checkConditions(expected, conditions)
+	if len(unavailableConditions) != 0 {
+		degraded := formatConditions(unavailableConditions)
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.IngressControllerAvailableConditionType,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "IngressControllerUnavailable",
+			Message: "One or more status conditions indicate unavailable: " + degraded,
+		}
+	}
+	return operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionTrue,
+	}
+}
+
+// checkConditions compares expected operator conditions to existing operator
+// conditions and returns a list of graceConditions, degradedconditions, and a
+// requeueing wait time.
+func checkConditions(expectedConds []expectedCondition, conditions []operatorv1.OperatorCondition) ([]*operatorv1.OperatorCondition, []*operatorv1.OperatorCondition, time.Duration) {
+	var graceConditions, degradedConditions []*operatorv1.OperatorCondition
+	var requeueAfter time.Duration
+	conditionsMap := make(map[string]*operatorv1.OperatorCondition)
+
+	for i := range conditions {
+		conditionsMap[conditions[i].Type] = &conditions[i]
+	}
+	now := clock.Now()
+	for _, expected := range expectedConds {
+		condition, haveCondition := conditionsMap[expected.condition]
+		if !haveCondition {
+			continue
+		}
+		if condition.Status == expected.status {
+			continue
+		}
+		failedPredicates := false
+		for _, ifCond := range expected.ifConditionsTrue {
+			predicate, havePredicate := conditionsMap[ifCond]
+			if !havePredicate || predicate.Status != operatorv1.ConditionTrue {
+				failedPredicates = true
+				break
+			}
+		}
+		if failedPredicates {
+			continue
+		}
+		if expected.gracePeriod != 0 {
+			t1 := now.Add(-expected.gracePeriod)
+			t2 := condition.LastTransitionTime
+			if t2.After(t1) {
+				d := t2.Sub(t1)
+				if len(graceConditions) == 0 || d < requeueAfter {
+					// Recompute status conditions again
+					// after the grace period has elapsed.
+					requeueAfter = d
+				}
+				graceConditions = append(graceConditions, condition)
+				continue
+			}
+		}
+		degradedConditions = append(degradedConditions, condition)
+	}
+	return graceConditions, degradedConditions, requeueAfter
 }
 
 // computeDeploymentAvailableCondition computes the ingresscontroller's
@@ -339,18 +416,7 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
 // reconcile the ingresscontroller again after that period to update its status
 // conditions.
 func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, icName string) (operatorv1.OperatorCondition, error) {
-	var requeueAfter time.Duration
-	conditionsMap := make(map[string]*operatorv1.OperatorCondition)
-	for i := range conditions {
-		conditionsMap[conditions[i].Type] = &conditions[i]
-	}
-
-	expectedConditions := []struct {
-		condition        string
-		status           operatorv1.ConditionStatus
-		ifConditionsTrue []string
-		gracePeriod      time.Duration
-	}{
+	expectedConditions := []expectedCondition{
 		{
 			condition: IngressControllerAdmittedConditionType,
 			status:    operatorv1.ConditionTrue,
@@ -410,54 +476,16 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, 
 		expectedConditions = append(expectedConditions, canaryCond)
 	}
 
-	var graceConditions, degradedConditions []*operatorv1.OperatorCondition
-	now := clock.Now()
-	for _, expected := range expectedConditions {
-		condition, haveCondition := conditionsMap[expected.condition]
-		if !haveCondition {
-			continue
-		}
-		if condition.Status == expected.status {
-			continue
-		}
-		failedPredicates := false
-		for _, ifCond := range expected.ifConditionsTrue {
-			predicate, havePredicate := conditionsMap[ifCond]
-			if !havePredicate || predicate.Status != operatorv1.ConditionTrue {
-				failedPredicates = true
-				break
-			}
-		}
-		if failedPredicates {
-			continue
-		}
-		if expected.gracePeriod != 0 {
-			t1 := now.Add(-expected.gracePeriod)
-			t2 := condition.LastTransitionTime
-			if t2.After(t1) {
-				d := t2.Sub(t1)
-				if len(graceConditions) == 0 || d < requeueAfter {
-					// Recompute status conditions again
-					// after the grace period has elapsed.
-					requeueAfter = d
-				}
-				graceConditions = append(graceConditions, condition)
-				continue
-			}
-		}
-		degradedConditions = append(degradedConditions, condition)
+	// Cover the rare case of no conditions
+	if len(conditions) == 0 {
+		return operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeDegraded, Status: operatorv1.ConditionFalse}, nil
 	}
-
+	graceConditions, degradedConditions, requeueAfter := checkConditions(expectedConditions, conditions)
 	if len(degradedConditions) != 0 {
 		// Keep checking conditions every minute while degraded.
-		requeueAfter = time.Minute
+		retryAfter := time.Minute
 
-		var degraded string
-		for _, cond := range degradedConditions {
-			degraded = degraded + fmt.Sprintf(", %s=%s (%s: %s)", cond.Type, cond.Status, cond.Reason, cond.Message)
-		}
-		degraded = degraded[2:]
-
+		degraded := formatConditions(degradedConditions)
 		condition := operatorv1.OperatorCondition{
 			Type:    operatorv1.OperatorStatusTypeDegraded,
 			Status:  operatorv1.ConditionTrue,
@@ -465,14 +493,12 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, 
 			Message: "One or more other status conditions indicate a degraded state: " + degraded,
 		}
 
-		return condition, retryableerror.New(errors.New("IngressController is degraded: "+degraded), requeueAfter)
+		return condition, retryableerror.New(errors.New("IngressController is degraded: "+degraded), retryAfter)
 	}
-
 	condition := operatorv1.OperatorCondition{
 		Type:   operatorv1.OperatorStatusTypeDegraded,
 		Status: operatorv1.ConditionFalse,
 	}
-
 	var err error
 	if len(graceConditions) != 0 {
 		var grace string
@@ -483,8 +509,19 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, 
 
 		err = retryableerror.New(errors.New("IngressController may become degraded soon: "+grace), requeueAfter)
 	}
-
 	return condition, err
+}
+
+func formatConditions(conditions []*operatorv1.OperatorCondition) string {
+	var formatted string
+	if len(conditions) == 0 {
+		return ""
+	}
+	for _, cond := range conditions {
+		formatted = formatted + fmt.Sprintf(", %s=%s (%s: %s)", cond.Type, cond.Status, cond.Reason, cond.Message)
+	}
+	formatted = formatted[2:]
+	return formatted
 }
 
 // IngressStatusesEqual compares two IngressControllerStatus values.  Returns true
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -1,7 +1,6 @@
 package ingress
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
@@ -236,12 +235,22 @@ func TestComputePodsScheduledCondition(t *testing.T) {
 }
 
 func TestComputeIngressDegradedCondition(t *testing.T) {
+	// Inject a fake clock and don't forget to reset it
+	fakeClock := utilclock.NewFakeClock(time.Time{})
+	clock = fakeClock
+	defer func() {
+		clock = utilclock.RealClock{}
+	}()
+
 	tests := []struct {
 		name                        string
 		icName                      string
 		conditions                  []operatorv1.OperatorCondition
 		expectIngressDegradedStatus operatorv1.ConditionStatus
 		expectRequeue               bool
+		// A degraded condition will give a 1 minute retry duration
+		// unless there is a grace period expected
+		expectAfter time.Duration
 	}{
 		{
 			name:                        "no conditions set",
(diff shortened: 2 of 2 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -30,6 +30,16 @@ import (
 // clock is to enable unit testing
 var clock utilclock.Clock = utilclock.RealClock{}
 
+// expectedCondition contains a condition that is expected to be checked when
+// determining Available or Degraded status of the ingress controller
+type expectedCondition struct {
+	condition        string
+	status           operatorv1.ConditionStatus
+	ifConditionsTrue []string // ifConditionsTrue is a list of dependent conditions that should be be true as well
```
Miciah, line 38:
```suggestion
	// ifConditionsTrue is a list of prerequisite conditions that should be be true
	// or else the condition is not checked.  
	ifConditionsTrue []string
```

pkg/operator/controller/ingress/status.go
```diff
@@ -30,6 +30,16 @@ import (
...
+	gracePeriod      time.Duration
+	mustHave         bool // mustHave determines whether this expectedCondition must be there
```
Miciah, line 40:
```suggestion
	// mustHave specifies whether this expectedCondition must be there.
	mustHave         bool
```
Is it true that `computeIngressAvailableCondition` will specify `mustHave: true` for all its expected conditions whereas `computeIngressDegradedCondition` will specify `mustHave: false` for all its expected conditions?  If so, would it make more sense for `checkAnnotatedConditions` to have an `ignoreMissingConditions` parameter?

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
-// by inspecting the Available condition of deployment. The ingresscontroller is only available if
-// the deployment is also available.
-func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	for _, cond := range deployment.Status.Conditions {
-		if cond.Type != appsv1.DeploymentAvailable {
-			continue
+// by inspecting the following:
+// 1) the Available condition of deployment
+// 2) the DNSReady condition
+// 3) the LoadBalancerReady condition
```
Miciah, line 187:
```suggestion
// 1) the Available condition of Deployment,
// 2) the DNSReady condition of the IngressController, and
// 3) the LoadBalancerReady condition of the IngressController.
```

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+// The ingresscontroller is judged Available only if all 3 conditions are true
+func computeIngressAvailableCondition(conditions []operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
+	conditionsMap := make(map[string]*operatorv1.OperatorCondition)
+	for i := range conditions {
+		conditionsMap[conditions[i].Type] = &conditions[i]
+	}
```
Miciah, line 193: Any reason not to pull this into `checkAnnotatedConditions`?

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+	expected := []expectedCondition{
+		{
+			condition: string(appsv1.DeploymentAvailable),
```
Miciah, line 196: You're checking the IngressController's status conditions, not the Deployment's, so you need to use the IngressController condition that mirrors the Deployment's, right?  
```suggestion
			condition: IngressControllerDeploymentAvailableConditionType,
```

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
-func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	for _, cond := range deployment.Status.Conditions {
-		if cond.Type != appsv1.DeploymentAvailable {
-			continue
+// by inspecting the following:
+// 1) the Available condition of deployment
+// 2) the DNSReady condition
+// 3) the LoadBalancerReady condition
+// The ingresscontroller is judged Available only if all 3 conditions are true
+func computeIngressAvailableCondition(conditions []operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
```
Miciah, line 189: Why is the return value a slice?

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
 		}
 	}
+}
 
-	return operatorv1.OperatorCondition{
-		Type:    operatorv1.IngressControllerAvailableConditionType,
-		Status:  operatorv1.ConditionFalse,
-		Reason:  "DeploymentAvailabilityUnknown",
-		Message: "The deployment's Available condition couldn't be interpreted",
+func checkAnnotatedConditions(expecteds []expectedCondition, conditionsMap map[string]*operatorv1.OperatorCondition) ([]*operatorv1.OperatorCondition, []*operatorv1.OperatorCondition) {
```
Miciah, line 243: Could this be named simply `checkAnnotatedConditions`?

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +181,109 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+					// after the grace period has elapsed.
+					requeueAfter = d
+				}
+				graceConditions = append(graceConditions, condition)
+				continue
+			}
+		}
+		degradedConditions = append(degradedConditions, condition)
 	}
+	return graceConditions, degradedConditions
```
Miciah, line 286: Need to return `requeueAfter` for `computeIngressDegradedCondition`.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+		conditions  []operatorv1.OperatorCondition
+		expect      []operatorv1.OperatorCondition
 	}{
 		{
-			description: "deployment available",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
+			description: "deployment, dns, and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
```
Miciah, line 830:
```suggestion
				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionTrue},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
-			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionTrue},
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionTrue}},
 		},
 		{
-			description: "deployment not available",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionFalse},
+			description: "deployment not available, but dns and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionFalse},
```
Miciah, line 840:
```suggestion
				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionFalse},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
-			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse},
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}},
 		},
 		{
-			description: "deployment availability unknown",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionUnknown},
+			description: "dns not available, but deployment and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
```
Miciah, line 850:
```suggestion
				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionTrue},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
-			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse},
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}},
 		},
 		{
-			description: "deployment availability not present",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionUnknown},
+			description: "lb not available, but dns and deployment available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
```
Miciah, line 862:
```suggestion
				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionTrue},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
-		deploy := &appsv1.Deployment{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: fmt.Sprintf("ingress-controller-%d", i+1),
-			},
-			Status: appsv1.DeploymentStatus{
-				Conditions: tc.deploymentConditions,
+		{
+			description: "all availability unknown",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionUnknown},
```
Miciah, line 872:
```suggestion
				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionUnknown},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+		expect      []operatorv1.OperatorCondition
 	}{
 		{
-			description: "deployment available",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
+			description: "deployment, dns, and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
```
Miciah, line 831:
```suggestion
				{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionTrue},
				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionTrue}},
 		},
 		{
-			description: "deployment not available",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionFalse},
+			description: "deployment not available, but dns and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionFalse},
+				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
```
Miciah, line 841: Need to add `operatorv1.DNSManagedIngressConditionType`.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionUnknown},
+			description: "dns not available, but deployment and lb available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionFalse},
+				{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionTrue},
```
Miciah, line 855: You have `operatorv1.LoadBalancerReadyIngressConditionType` twice in this test case.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}},
 		},
 		{
-			description: "deployment availability not present",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionUnknown},
+			description: "lb not available, but dns and deployment available",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: string(appsv1.DeploymentAvailable), Status: operatorv1.ConditionTrue},
+				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
```
Miciah, line 863: Need to add `operatorv1.DNSManagedIngressConditionType`.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -821,51 +820,72 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+				{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionTrue},
 			},
-			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse},
+			expect: []operatorv1.OperatorCondition{{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}},
 		},
 		{
-			description: "deployment availability not present",
-			deploymentConditions: []appsv1.DeploymentCondition{
-				{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionUnknown},
+			description: "lb not available, but dns and deployment available",
```
Miciah, line 860: This should never happen... but it doesn't hurt to cover it.

pkg/operator/controller/ingress/status.go
candita: For now that is true, but I could envision a future use where an individual condition of a set might be a "mustHave".  I thought it best to leave it individually specified, but don't mind changing that if needed.

pkg/operator/controller/ingress/status.go
candita: Will change that.

pkg/operator/controller/ingress/status.go
candita: Not sure what you mean?  It is already named checkAnnotatedConditions...

pkg/operator/controller/ingress/status.go
candita: Ouch, that would have been bad.

pkg/operator/controller/ingress/status.go
Miciah: Sorry, I meant to suggest `checkConditions`.

sgreene570:
/retitle [WIP] Bug 1870373: Ingress Operator reports available when DNS fails to provision

candita:
/retitle [WIP] Bug 1870373: Ingress Operator reports available when DNS fails to provision

pkg/operator/controller/ingress/status.go
candita: Ok, I will change it

candita: Done.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,45 +184,113 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
 }
 
 // computeDeploymentAvailableCondition computes the ingresscontroller's
 // "DeploymentAvailable" status condition by examining the status conditions of
 // the deployment.  The "DeploymentAvailable" condition is true if the
 // deployment's "Available" condition is true.
-//
-// Note: Due to a defect in the deployment controller, the deployment reports
-// Available=True before minimum availability requirements are met (see
-// <https://bugzilla.redhat.com/show_bug.cgi?id=1830271#c5>).
```
Miciah, line 212: I think this comment is still accurate.

pkg/operator/controller/ingress/status.go
```diff
@@ -339,17 +420,8 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
 // conditions.
 func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition) (operatorv1.OperatorCondition, error) {
 	var requeueAfter time.Duration
```
Miciah, line 422: This declaration is redundant now and can be removed.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,45 +184,113 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+			},
+			gracePeriod: time.Second * 30,
+			mustHave:    true,
+		},
+		{
+			condition:        operatorv1.LoadBalancerReadyIngressConditionType,
+			status:           operatorv1.ConditionTrue,
+			ifConditionsTrue: []string{operatorv1.LoadBalancerManagedIngressConditionType},
+			gracePeriod:      time.Second * 90,
+			mustHave:         true,
```
Miciah, line 216: Specifying `mustHave: true` for `operatorv1.DNSReadyIngressConditionType` and `operatorv1.LoadBalancerReadyIngressConditionType` is not going to work properly when the endpoint publishing strategy type is not `LoadBalancerService` (or if DNS management is turned off by clearing the zones in the cluster DNS config object) because `computeLoadBalancerStatus` and `computeDNSStatus` do not set the `LoadBalancerReady` or `DNSReady` status conditions if the operator is not managing an LB or DNS.

pkg/operator/controller/ingress/status.go
candita: Even though the BZ was closed?

pkg/operator/controller/ingress/status.go
candita: You're right, I noticed it had an adverse affect on the end to end testing.  I have removed this construct and replaced with a check for empty conditions.

pkg/operator/controller/ingress/status.go
candita: Done.

pkg/operator/controller/ingress/status.go
candita: Ok thanks, I will put it back.

pkg/operator/controller/ingress/status.go
candita: This has been removed.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
-func computeIngressAvailableCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	for _, cond := range deployment.Status.Conditions {
-		if cond.Type != appsv1.DeploymentAvailable {
+// by inspecting the following:
+// 1) the Available condition of Deployment,
+// 2) the DNSReady condition of the IngressController, and
+// 3) the LoadBalancerReady condition of the IngressController.
+// The ingresscontroller is judged Available only if all 3 conditions are true
+func computeIngressAvailableCondition(conditions []operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+
```
Miciah, line 191: The blank line can be deleted.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+		},
+		{
+			condition: operatorv1.DNSReadyIngressConditionType,
+			status:    operatorv1.ConditionTrue,
+			ifConditionsTrue: []string{
+				operatorv1.LoadBalancerManagedIngressConditionType,
+				operatorv1.LoadBalancerReadyIngressConditionType,
+				operatorv1.DNSManagedIngressConditionType,
+			},
+			gracePeriod: time.Second * 30,
```
Miciah, line 205: Did we decide to specify grace periods in `computeIngressAvailableCondition`?  I thought doing so wasn't necessary for `Available` (it's only necessary for `Degraded` to prevent flapping).

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+			status:           operatorv1.ConditionTrue,
+			ifConditionsTrue: []string{operatorv1.LoadBalancerManagedIngressConditionType},
+			gracePeriod:      time.Second * 90,
+		},
+	}
+
+	// Cover the rare case of no conditions
+	if len(conditions) == 0 {
+		return operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeAvailable, Status: operatorv1.ConditionFalse}
+	}
```
Miciah, line 218: `conditions` should never be empty (`syncIngressControllerStatus` sets a whole bunch of status conditions before it calls `computeIngressAvailableCondition`).  Is this check only needed to make the "all availability not present" unit test case pass, or is the intention to check if none of the conditions mentioned in `expected` is present?

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "IngressControllerUnavailable",
+			Message: "One or more status conditions indicate unavailable: " + degraded,
+		}
+	} else {
+		return operatorv1.OperatorCondition{
+			Type:   operatorv1.IngressControllerAvailableConditionType,
+			Status: operatorv1.ConditionTrue,
+		}
+	}
```
Miciah, line 234:
```suggestion
	}
	return operatorv1.OperatorCondition{
		Type:   operatorv1.IngressControllerAvailableConditionType,
		Status: operatorv1.ConditionTrue,
	}
```

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+}
+
+// checkConditions compares expected operator conditions to existing operator
+// conditions and returns a list of graceConditions, degradedconditions, and a
+// requeueing wait time.
+func checkConditions(expecteds []expectedCondition, conditions []operatorv1.OperatorCondition) ([]*operatorv1.OperatorCondition, []*operatorv1.OperatorCondition, time.Duration) {
+
```
Miciah, line 241: The blank line can be deleted.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,108 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+	var graceConditions, degradedConditions []*operatorv1.OperatorCondition
+	conditionsMap := make(map[string]*operatorv1.OperatorCondition)
+
+	for i := range conditions {
+		conditionsMap[conditions[i].Type] = &conditions[i]
+	}
+	now := clock.Now()
+	// Keep checking conditions every minute if needed
+	requeueAfter := time.Minute
```
Miciah, line 250: `requeueAfter` should start at 0 and only be changed to a positive value if any grace conditions are found.  
```suggestion
	var requeueAfter time.Duration
```
(Hm, seems like we have a gap in test coverage.)

pkg/operator/controller/ingress/status.go
candita: I will remove.

pkg/operator/controller/ingress/status.go
candita: Yes, it was added to make the unit test case pass.  But also, conditions don't need to be checked if there are none, even if that is extremely unlikely.

pkg/operator/controller/ingress/status.go
candita: Thanks Miciah - I updated the unit test to check for the correct duration to be returned.  PTAL and let me know if there are any other concerns for this PR.

pkg/operator/controller/ingress/status.go
```diff
@@ -171,35 +182,101 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
...
+	return operatorv1.OperatorCondition{
+		Type:   operatorv1.IngressControllerAvailableConditionType,
+		Status: operatorv1.ConditionTrue,
+	}
+}
+
+// checkConditions compares expected operator conditions to existing operator
+// conditions and returns a list of graceConditions, degradedconditions, and a
+// requeueing wait time.
+func checkConditions(expecteds []expectedCondition, conditions []operatorv1.OperatorCondition) ([]*operatorv1.OperatorCondition, []*operatorv1.OperatorCondition, time.Duration) {
```
sgreene570, line 235: thoughts on naming this parameter `expectedConds` instead of `expecteds` (or something along those lines)?

pkg/operator/controller/ingress/status.go
```diff
@@ -392,69 +458,29 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition)
...
 	}
-
 	condition := operatorv1.OperatorCondition{
 		Type:   operatorv1.OperatorStatusTypeDegraded,
 		Status: operatorv1.ConditionFalse,
 	}
-
 	var err error
 	if len(graceConditions) != 0 {
 		var grace string
```
Miciah, line 486: We could simplify this as `grace := formatConditions(graceConditions)`.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -254,15 +256,18 @@ func TestComputeIngressDegradedCondition(t *testing.T) {
...
 			name: "pods not scheduled for <10m",
 			conditions: []operatorv1.OperatorCondition{
-				cond(
-					IngressControllerPodsScheduledConditionType, operatorv1.ConditionFalse, "", clock.Now().Add(time.Minute*-9)),
+				cond(IngressControllerPodsScheduledConditionType, operatorv1.ConditionFalse, "", clock.Now().Add(time.Minute*-9)),
 			},
 			expectIngressDegradedStatus: operatorv1.ConditionFalse,
 			expectRequeue:               true,
+			// Grace period is 10 minutes, subtract the 9 minute spoofed last transition time
+			expectAfter: time.Minute,
```
Miciah, line 270: This is racy because it assumes that 0 time passes between the point in time when this test case is defined with the time value `clock.Now().Add(time.Minute*-9)` and the point in time when `computeIngressDegradedCondition`/`checkConditions` calls `clock.Now()` to compute the remaining grace time for the test case.  

I see you round the time value in the test runner (`if test.expectAfter.Seconds() != e.After().Round(time.Second).Seconds()`), which should reduce the likelihood of hitting the race condition, but I think it still could happen.  

To force the race condition, add `time.Sleep(time.Second)` right before the `for` loop, and then run the tests:
```
go test ./pkg/operator/controller/ingress
--- FAIL: TestComputeIngressDegradedCondition (1.00s)
    status_test.go:450: "pods not scheduled for <10m": expected requeue after 1m0s, got 58.999745214s
    status_test.go:450: "deployment unavailable for <30s": expected requeue after 10s, got 8.999650475s
    status_test.go:450: "deployment minimum replicas unavailable for <60s": expected requeue after 40s, got 38.999604168s
    status_test.go:450: "deployment not all replicas available for <60m": expected requeue after 40m0s, got 39m58.999553896s
    status_test.go:450: "LB provisioning failing <90s": expected requeue after 30s, got 28.999502123s
    status_test.go:450: "DNS failing <30s": expected requeue after 15s, got 13.999467914s
```

To fix the race condition, please make the test use a fake clock by adding the following lines at the start of `TestComputeIngressDegradedCondition`: 
```go
	clock = utilclock.NewFakeClock(time.Time{})
	defer func() {
		clock = utilclock.RealClock{}
	}()
```
This way, we control the passage of time (the hands on the fake clock don't move unless we move them).  Then the raciness is thereby eliminated, and the rounding can be deleted.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -341,6 +360,8 @@ func TestComputeIngressDegradedCondition(t *testing.T) {
 			},
 			expectIngressDegradedStatus: operatorv1.ConditionFalse,
 			expectRequeue:               true,
+			// Minimum grace period of combined conditions (DNSReadyIngressConditionType) is 30 seconds
```
Miciah, line 363:
```suggestion
			// Minimum grace period of combined conditions (LoadBalancerReadyIngressConditionType) is 30 seconds
```

pkg/operator/controller/ingress/status.go
candita: The format is different for reporting graceConditions than reporting degradedConditions, so I did not use the method here.
grace: fmt.Sprintf(", %s=%s", cond.Type, cond.Status)
degraded: fmt.Sprintf(", %s=%s (%s: %s)", cond.Type, cond.Status, cond.Reason, cond.Message)

pkg/operator/controller/ingress/status_test.go
candita: DNSReadyIngressConditionType grace period is the minimum at 30 seconds.  LoadBalancerManagedIngressConditionType grace period is 90 seconds.

pkg/operator/controller/ingress/status_test.go
candita: Done!

pkg/operator/controller/ingress/status_test.go
Miciah: Sorry, I must have been confused.  You're right; this test case looks fine.

Miciah:
The fix looks correct, and you've added some really nice clean-ups and test coverage improvements as well.  Thanks!  
/lgtm

sgreene570:
Rebase looks good.
/lgtm
/honk

PR #498, opened 2020-11-27
Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-ingress-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.


cfergeau:
Probably best to wait until https://github.com/openshift/api/pull/792 is merged as this will make it easier to fix CI.

sgreene570:
Interesting that https://github.com/openshift/cluster-ingress-operator/pull/494 merged without a bump for `openshift/api`. 
Either way, this PR will correct the issue on `master`. Thanks! 
/lgtm

PR #499, opened 2020-11-30
hack: verify-generated-crd.sh should fail on diffs

sgreene570:
Add an exit code to the `hack/verify-generated-crd.sh`
script so discovered diffs return an exit code of 1,
rather than 0 (which fakes out the verify CI job).


sgreene570:
/hold for https://github.com/openshift/cluster-ingress-operator/pull/498
Since this PR _will_  fail on the current master branch, intentionally.

PR #502, opened 2020-12-04
Bug 1904582: Assume ingresscontroller is external absent status

Miciah:
If an ingresscontroller's `status.endpointPublishingStrategy.type` field is `LoadBalancerService` but `status.endpointPublishingStrategy.loadBalancer` is `nil`, assume the ingresscontroller should have an external load balancer.

#254 added support for internal scope for load balancers.  In doing so, it changed the operator logic to initialize `status.endpointPublishingStrategy.loadBalancer` to indicate the scope when an ingresscontroller was created, using the corresponding fields in the ingresscontroller's `spec` to determine the scope.  The same change also changed operator logic to assume that the ingresscontroller has internal scope if `status.endpointPublishingStrategy.loadBalancer` is `nil`. However, if an ingresscontroller already exists, the operator does not update `status.endpointPublishingStrategy.loadBalancer`.  This change was made in OpenShift 4.2, which means that if a cluster was upgraded from 4.1, any ingresscontrollers that were created prior to the upgrade would not have `status.endpointPublishingStrategy.loadBalancer` set.

Subsequently, #472 made the ingresscontroller's scope mutable, meaning if an ingresscontroller has an external load balancer and `status.endpointPublishingStrategy.loadBalancer` indicates that the ingresscontroller should have an internal load balancer, then the operator changes the load balancer from external to internal.

In combination, those two changes cause the operator to change an ingresscontroller's load balancer from external to internal if the cluster has been upgraded from 4.1 → 4.2 → ... → 4.6 and the ingresscontroller was originally created on OpenShift 4.1.

This PR rectifies the situation by amending the logic that was added in #254 to assume that a `nil` value for `status.endpointPublishingStrategy.loadBalancer` means that the load balancer should be external. This assumption is valid because `status.endpointPublishingStrategy.loadBalancer` is `nil` exactly when the ingresscontroller was created on an OpenShift 4.1 cluster, and external scope was the only supported option on OpenShift 4.1.

* `pkg/operator/controller/ingress/load_balancer_service.go` (`desiredLoadBalancerService`): Assume `nil` `status.endpointPublishingStrategy.loadBalancer` means external scope.

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -222,7 +222,7 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef 
 
 	service.Spec.Selector = controller.IngressControllerDeploymentPodSelector(ci).MatchLabels
 
-	isInternal := ci.Status.EndpointPublishingStrategy.LoadBalancer == nil || ci.Status.EndpointPublishingStrategy.LoadBalancer.Scope == operatorv1.InternalLoadBalancer
+	isInternal := ci.Status.EndpointPublishingStrategy.LoadBalancer != nil && ci.Status.EndpointPublishingStrategy.LoadBalancer.Scope == operatorv1.InternalLoadBalancer
 
 	if service.Annotations == nil {
 		service.Annotations = map[string]string{}


Miciah:
We'll need this backported to 4.6 (which made scope mutable) but no further.  
/cherry-pick release-4.6

Miciah:
/retitle Bug 1904582: Assume ingresscontroller is external absent status

sgreene570:
LGTM once the bugzilla is linked

bbrowning:
Does this open the door for the potential reverse problem? If someone manually changed a 4.1.z cluster's ingresscontroller to have internal load balancer after the cluster was initially installed, will this inadvertently change that back to an external load balancer?

Or would any manual change to point the apps domain at an internal load balancer have been reverted by the ingress operator anyway, before this change?

Miciah:
> Does this open the door for the potential reverse problem? If someone manually changed a 4.1.z cluster's ingresscontroller to have internal load balancer after the cluster was initially installed, will this inadvertently change that back to an external load balancer?

No, if the administrator changed the scope, then that means there is an explicit value for `spec.endpointPublishingStrategy.loadBalancer.scope`.  

> Or would any manual change to point the apps domain at an internal load balancer have been reverted by the ingress operator anyway, before this change?

What kind of manual changes do you mean?  Directly modifying the `Service` or `DNSRecord` objects that the operator creates is unsupported.

Miciah:
> No, if the administrator changed the scope, then that means there is an explicit value for `spec.endpointPublishingStrategy.loadBalancer.scope`.

And to complete that thought, #472 means that changes to `spec.endpointPublishingStrategy.loadBalancer.scope` are reflected to `status.endpointPublishingStrategy.loadBalancer.scope`: https://github.com/openshift/cluster-ingress-operator/blob/f2d7665356506f82e2397d47200cc3ccbb8c0eae/pkg/operator/controller/ingress/controller.go#L365-L367

bbrowning:
> What kind of manual changes do you mean? Directly modifying the `Service` or `DNSRecord` objects that the operator creates is unsupported.

I was referring to a user modifying the cloud resources directly for started-in-4.1 clusters where users could not choose internal versus external load balancers via the API. Back then if you wanted an internal only load balancer, your only option was to modify the cloud resources directly, right? And I don't think we reconciled manual changes away? I know of no way to determine what, if any, clusters would have done that and if it was never a suggested or supported option for older clusters than my concern is probably irrelevant.

Miciah:
> > What kind of manual changes do you mean? Directly modifying the `Service` or `DNSRecord` objects that the operator creates is unsupported.
> 
> I was referring to a user modifying the cloud resources directly for started-in-4.1 clusters where users could not choose internal versus external load balancers via the API. Back then if you wanted an internal only load balancer, your only option was to modify the cloud resources directly, right? And I don't think we reconciled manual changes away? I know of no way to determine what, if any, clusters would have done that and if it was never a suggested or supported option for older clusters than my concern is probably irrelevant.

That's true, we _didn't_ reconcile away user-added annotations on the service, but as you say, we have never suggested or supported administrators doing that.  In general, directly modifying operator-managed resources is unsupported.

Miciah:
There have been cases where we suggested setting `spec.endpointPublishingStrategy.type: Private` and creating a user-managed service.  If a user has done that, that's fine; the operator will not mess with user-created resources.

PR #505, opened 2020-12-07
Bug 1887441: Conditionally delete generated default cert

sgreene570:
certificate: Conditionally delete generated default cert 

pkg/operator/controller/certificate/default_cert.go: 

If an Ingress Controller is modified so that the Spec.DefaultCertificate field is now set, only delete the prior operator generated default certificate if the secret referenced by Spec.DefaultCertificate exists. This is to correct the bug identified in BZ 1887441 in which modifying an Ingress Controller to reference a secret that does not exist via Spec.DefaultCertificate breaks the authentication operator (which relies on the router-certs secret). This commit also cleans up the default_cert.go file to align with recent improvements in the ingress-operator codebase. Namely, the `currentRouterDefaultCertificate` and `desiredRouterDefaultCertificate` functions are enhanced so that they return an indicator boolean (ie haveCert and wantCert) in addition to the values previously returned.

certificate-publisher: Check for ic.Spec.DefaultCertificate:

pkg/operator/controller/certificate-publisher/publish_certs.go: Change desiredRouterCertsGlobalSecret to check if an ingress controller's Spec.DefaultCertificate secret exists in the parameter secrets slice. If it does not, fall back to the operator generated default certificate. This makes sense since if Spec.DefaultCertificate is updated to point to a non-exists secret, the certificate controller will not delete the operator generated default certificate. This commit essentially inlines a more in-depth version of controller.RouterEffectiveDefaultCertificateSecretName(...) into desiredRouterCertsGlobalSecret, and is a part of the fix for BZ 1887441, in which the router certs global secret is clobbered by the ingress operator when an ingress controller specifies a default certificate secret that does not exist (which could be the result of a simple typo or user error).

### DIFF
--- a/pkg/operator/controller/certificate-publisher/publish_certs.go
+++ b/pkg/operator/controller/certificate-publisher/publish_certs.go
@@ -68,7 +68,19 @@ func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []operato
 
 	ingressToSecret := map[*operatorv1.IngressController]*corev1.Secret{}
 	for i, ingress := range ingresses {
-		name := controller.RouterEffectiveDefaultCertificateSecretName(&ingress, operandNamespace)
+		// Check if ingress.Spec.DefaultCertificate is an available secret
+		// in the secrets slice. If it is not, attempt to fall back to the
+		// operator generated default certificate. If ingress.Spec.DefaultCertificate
+		// is updated to point to a non-existent secret, the certificate controller
+		// will not delete the operator generated default certificate.
+		// See https://bugzilla.redhat.com/show_bug.cgi?id=1887441
+		if defaultCert := ingress.Spec.DefaultCertificate; defaultCert != nil {
+			if secret, ok := nameToSecret[defaultCert.Name]; ok {
+				ingressToSecret[&ingresses[i]] = secret
+				continue
+			}
+		}
+		name := controller.RouterOperatorGeneratedDefaultCertificateSecretName(&ingress, operandNamespace)
 		if secret, ok := nameToSecret[name.Name]; ok {
 			ingressToSecret[&ingresses[i]] = secret
 		}
--- a/pkg/operator/controller/certificate/default_cert.go
+++ b/pkg/operator/controller/certificate/default_cert.go
@@ -17,67 +17,77 @@ import (
 
 // ensureDefaultCertificateForIngress creates or deletes an operator-generated
 // default certificate for a given IngressController as appropriate.  Returns true
-// if it creates, updates, or deletes the secret for the certificate, false
-// otherwise.
+// if it the secret exists, or false if it does not, as well as any errors.
 func (r *reconciler) ensureDefaultCertificateForIngress(caSecret *corev1.Secret, namespace string, deploymentRef metav1.OwnerReference, ci *operatorv1.IngressController) (bool, error) {
 	ca, err := crypto.GetCAFromBytes(caSecret.Data["tls.crt"], caSecret.Data["tls.key"])
 	if err != nil {
 		return false, fmt.Errorf("failed to get CA from secret %s/%s: %v", caSecret.Namespace, caSecret.Name, err)
 	}
-	desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
+	wantCert, desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
 	if err != nil {
 		return false, err
 	}
-	current, err := r.currentRouterDefaultCertificate(ci, namespace)
+	if !wantCert {
+		// If the operator generated certificate is not being used, ensure that the ingress controller's
+		// Spec.DefaultCertificate secret exists before deleting the operator generated secret.
+		// See https://bugzilla.redhat.com/show_bug.cgi?id=1887441
+		err := r.lookupUserSpecifiedRouterDefaultCertificate(ci, namespace)
+		if err != nil {
+			return false, fmt.Errorf("failed to lookup user specified default certificate: %v", err)
+		}
+	}
+
+	haveCert, current, err := r.currentRouterDefaultCertificate(ci, namespace)
 	if err != nil {
 		return false, err
 	}
 	switch {
-	case desired == nil && current == nil:
+	case !wantCert && !haveCert:
 		// Nothing to do.
-	case desired == nil && current != nil:
+	case !wantCert && haveCert:
 		if deleted, err := r.deleteRouterDefaultCertificate(current); err != nil {
-			return false, fmt.Errorf("failed to delete default certificate: %v", err)
+			return true, fmt.Errorf("failed to delete default certificate: %v", err)
 		} else if deleted {
 			r.recorder.Eventf(ci, "Normal", "DeletedDefaultCertificate", "Deleted default wildcard certificate %q", current.Name)
-			return true, nil
+			return false, nil
 		}
-	case desired != nil && current == nil:
+	case wantCert && !haveCert:
 		if created, err := r.createRouterDefaultCertificate(desired); err != nil {
 			return false, fmt.Errorf("failed to create default certificate: %v", err)
 		} else if created {
 			r.recorder.Eventf(ci, "Normal", "CreatedDefaultCertificate", "Created default wildcard certificate %q", desired.Name)
 			return true, nil
 		}
-	case desired != nil && current != nil:
+	case wantCert && haveCert:
 		// TODO Update if CA certificate changed.
+		return true, nil
 	}
 	return false, nil
 }
 
 // desiredRouterDefaultCertificateSecret returns the desired default certificate
 // secret.
-func desiredRouterDefaultCertificateSecret(ca *crypto.CA, namespace string, deploymentRef metav1.OwnerReference, ci *operatorv1.IngressController) (*corev1.Secret, error) {
+func desiredRouterDefaultCertificateSecret(ca *crypto.CA, namespace string, deploymentRef metav1.OwnerReference, ci *operatorv1.IngressController) (bool, *corev1.Secret, error) {
 	// Without an ingress domain, we cannot generate a default certificate.
 	if len(ci.Status.Domain) == 0 {
-		return nil, nil
+		return false, nil, nil
 	}
 
 	// If the ingresscontroller specifies a default certificate secret, the
 	// operator does not need to generate a certificate.
 	if ci.Spec.DefaultCertificate != nil {
-		return nil, nil
+		return false, nil, nil
 	}
 
 	hostnames := sets.NewString(fmt.Sprintf("*.%s", ci.Status.Domain))
 	cert, err := ca.MakeServerCert(hostnames, 0)
 	if err != nil {
-		return nil, fmt.Errorf("failed to make certificate: %v", err)
+		return false, nil, fmt.Errorf("failed to make certificate: %v", err)
 	}
 
 	certBytes, keyBytes, err := cert.GetPEMBytes()
 	if err != nil {
-		return nil, fmt.Errorf("failed to encode certificate: %v", err)
+		return false, nil, fmt.Errorf("failed to encode certificate: %v", err)
 	}
 
 	name := controller.RouterOperatorGeneratedDefaultCertificateSecretName(ci, namespace)
@@ -93,21 +103,21 @@ func desiredRouterDefaultCertificateSecret(ca *crypto.CA, namespace string, depl
 		},
 	}
 	secret.SetOwnerReferences([]metav1.OwnerReference{deploymentRef})
-	return secret, nil
+	return true, secret, nil
 }
 
 // currentRouterDefaultCertificate returns the current router default
 // certificate secret.
-func (r *reconciler) currentRouterDefaultCertificate(ci *operatorv1.IngressController, namespace string) (*corev1.Secret, error) {
+func (r *reconciler) currentRouterDefaultCertificate(ci *operatorv1.IngressController, namespace string) (bool, *corev1.Secret, error) {
 	name := controller.RouterOperatorGeneratedDefaultCertificateSecretName(ci, namespace)
 	secret := &corev1.Secret{}
 	if err := r.client.Get(context.TODO(), name, secret); err != nil {
 		if errors.IsNotFound(err) {
-			return nil, nil
+			return false, nil, nil
 		}
-		return nil, err
+		return false, nil, err
 	}
-	return secret, nil
+	return true, secret, nil
 }
 
 // createRouterDefaultCertificate creates a router default certificate secret.
@@ -129,4 +139,16 @@ func (r *reconciler) deleteRouterDefaultCertificate(secret *corev1.Secret) (bool
 		return false, err
 	}
 	return true, nil
+}
+
+// lookupUserSpecifiedRouterDefaultCertificate checks to see if the given ingress controller's
+// Spec.DefaultCertificate field corresponds to an existing secret. This function assumes that
+// ci.Spec.DefaultCertificate is not nil.
+func (r *reconciler) lookupUserSpecifiedRouterDefaultCertificate(ci *operatorv1.IngressController, namespace string) error {
+	secret := &corev1.Secret{}
+	name := controller.RouterEffectiveDefaultCertificateSecretName(ci, namespace)
+	if err := r.client.Get(context.TODO(), name, secret); err != nil {
+		return err
+	}
+	return nil
 }


pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,73 @@ import (
...
-	case desired == nil && current == nil:
+	case !wantCert && !haveCert:
 		// Nothing to do.
-	case desired == nil && current != nil:
+	case !wantCert && haveCert:
 		if deleted, err := r.deleteRouterDefaultCertificate(current); err != nil {
 			return false, fmt.Errorf("failed to delete default certificate: %v", err)
 		} else if deleted {
 			r.recorder.Eventf(ci, "Normal", "DeletedDefaultCertificate", "Deleted default wildcard certificate %q", current.Name)
 			return true, nil
```
Miciah, line 51: Hm, this should be `return false, nil`.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,73 @@ import (
 
 // ensureDefaultCertificateForIngress creates or deletes an operator-generated
 // default certificate for a given IngressController as appropriate.  Returns true
-// if it creates, updates, or deletes the secret for the certificate, false
+// if it creates or deletes the secret for the certificate, false
 // otherwise.
```
Miciah, line 21: For other `ensure` functions, the Boolean indicates whether the object exists, not whether it was just created.  Any reason `ensureDefaultCertificateForIngress` should be different?  (This question concerns the logic as it was prior to this PR, so feel free to defer on this point in this PR.)

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,73 @@ import (
...
 	if err != nil {
 		return false, fmt.Errorf("failed to get CA from secret %s/%s: %v", caSecret.Namespace, caSecret.Name, err)
 	}
-	desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
+	wantCert, desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
 	if err != nil {
 		return false, err
+	} else if !wantCert {
+		// If the opeartor generated certificate is not being used, ensure that the ingress controller's
+		// Spec.DefaultCertificate secret exists before deleting the operator generated secret.
```
Miciah, line 32:
```suggestion
		// If the operator-generated certificate is not being used, ensure that the ingress controller's
		// Spec.DefaultCertificate secret exists before deleting the operator generated secret.
```
Maybe link to the Bugzilla report in the comment.

`ensureDefaultCertificateForIngress` manages the secret in the `openshift-ingress` namespace.  In the scenario described in the Bugzilla report, the auth operator breaks because the certificate is removed from the secret the `openshift-config-managed` namespace, right?  The latter secret is managed by `ensureRouterCertsGlobalSecret` in the certificate-publisher controller, and it currently doesn't publish the secret unless it is referenced by an ingresscontroller, so I think changes will be required to that logic as well.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,73 @@ import (
...
-	case desired != nil && current == nil:
+	case wantCert && !haveCert:
 		if created, err := r.createRouterDefaultCertificate(desired); err != nil {
 			return false, fmt.Errorf("failed to create default certificate: %v", err)
 		} else if created {
 			r.recorder.Eventf(ci, "Normal", "CreatedDefaultCertificate", "Created default wildcard certificate %q", desired.Name)
 			return true, nil
 		}
-	case desired != nil && current != nil:
-		// TODO Update if CA certificate changed.
```
Miciah, line 53: From the PR description:

> Finally, this commit removes a stale TODO comment requesting that update logic be added to
> default_cert.go. Comparing the results of `desiredRouterDefaultCertificateSecret` to
> `currentRouterDefaultCertificate` would be pointless since `desiredRouterDefaultCertificateSecret`
> always returns a uniquely generated certificate.

It is true that comparing the results of `desiredRouterDefaultCertificateSecret` to `currentRouterDefaultCertificate` would be pointless, but we could still check if the CA certificate matches or is newer than the serving certificate. .

Anyway, I think we should return `true, nil` here.

pkg/operator/controller/certificate/default_cert.go
sgreene570: Ah, I naively though that the `router-certs-default` secret being deleted was the only crux of the bug. I will make the changes in the certificate-publisher as well.

sgreene570:
Tagging WIP for now while I work out the code changes for the certificate-publisher controller.

pkg/operator/controller/certificate/default_cert.go
sgreene570: Ah, right. 
For now I'll leave the TODO and return true, nil.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,73 @@ import (
...
 func (r *reconciler) ensureDefaultCertificateForIngress(caSecret *corev1.Secret, namespace string, deploymentRef metav1.OwnerReference, ci *operatorv1.IngressController) (bool, error) {
 	ca, err := crypto.GetCAFromBytes(caSecret.Data["tls.crt"], caSecret.Data["tls.key"])
 	if err != nil {
 		return false, fmt.Errorf("failed to get CA from secret %s/%s: %v", caSecret.Namespace, caSecret.Name, err)
 	}
-	desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
+	wantCert, desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
 	if err != nil {
 		return false, err
+	} else if !wantCert {
```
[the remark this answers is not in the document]
sgreene570: fixed. Thanks!

pkg/operator/controller/certificate/default_cert.go
sgreene570: yup, fixed.

pkg/operator/controller/certificate/default_cert.go
sgreene570: No wonder this file was confusing to work with. I've adjusted the return values to match our other `ensure` functions, as you suggested.

pkg/operator/controller/certificate/default_cert.go
sgreene570: I've updated the certificate publisher controller. @Miciah PTAL when you get the chance.

pkg/operator/controller/certificate-publisher/controller.go
```diff
@@ -86,9 +86,11 @@ func New(mgr manager.Manager, operatorNamespace, operandNamespace string) (runti
 	}
 
 	if err := c.Watch(&source.Kind{Type: &operatorv1.IngressController{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
-		CreateFunc:  func(e event.CreateEvent) bool { return reconciler.hasSecret(e.Meta, e.Object) },
-		DeleteFunc:  func(e event.DeleteEvent) bool { return reconciler.hasSecret(e.Meta, e.Object) },
-		UpdateFunc:  func(e event.UpdateEvent) bool { return reconciler.secretChanged(e.ObjectOld, e.ObjectNew) },
+		CreateFunc: func(e event.CreateEvent) bool { return reconciler.hasSecret(e.Meta, e.Object) },
+		DeleteFunc: func(e event.DeleteEvent) bool { return reconciler.hasSecret(e.Meta, e.Object) },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return reconciler.hasSecret(e.MetaNew, e.ObjectNew) && reconciler.secretChanged(e.ObjectOld, e.ObjectNew)
```
Miciah, line 92: What happens if ingresscontroller A and ingresscontroller B each has a valid secret reference, then A is updated with a broken reference, and then B is updated to a new but still valid reference?  Doesn't that break A?  

Looks like your intention is to prevent updates to the global secret if any ingresscontroller has an invalid secret reference, right?  If so, it probably would work better to check for invalid references in `ensureRouterCertsGlobalSecret`, `desiredRouterCertsGlobalSecret`, or `Reconcile`.  

Maybe though it would be better not to make an invalid reference in ingresscontroller A block updates related to ingresscontroller B.  Instead, we could to change `ensureRouterCertsGlobalSecret` to get the current list (if it exists) and pass it to `desiredRouterCertsGlobalSecret`, and change `desiredRouterCertsGlobalSecret` to use the current list to fill in the PEM data for ingresscontrollers that don't have secrets.

sgreene570:
`FAIL: TestHeaderNameCaseAdjustment`
/test e2e-aws-operator

pkg/operator/controller/certificate-publisher/controller.go
sgreene570: See latest changes. I modified `desiredRouterCertsGlobalSecret` along the lines of your final suggestion.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,77 @@ import (
...
+		}
+	}
+
+	haveCert, current, err := r.currentRouterDefaultCertificate(ci, namespace)
 	if err != nil {
 		return false, err
 	}
 	switch {
-	case desired == nil && current == nil:
+	case !wantCert && !haveCert:
```
candita, line 45: I like the clearer variable names here, but maybe you can just add two lines to set wantCert if desired != nil and haveCert if current != nil, and keep the signatures/returns the same on the two methods you changed?

sgreene570: Granted we've been slow to adopt the `haveFoo` boolean pattern across every `ensureFoo` function in our components, but we have been moving towards using it to essentially avoid lots of `desired != nil` checks. The functions in which I've made signature changes to belong to this file and are private, so I don't think changing them is a big deal. Personally, having the boolean return value is simpler for me to understand, especially in the more complicate `ensureFoo` functions, such as `ensureLoadBalancerService` (https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/load_balancer_service.go#L142)

pkg/operator/controller/certificate-publisher/publish_certs.go
```diff
@@ -68,7 +68,19 @@ func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []operato
 
 	ingressToSecret := map[*operatorv1.IngressController]*corev1.Secret{}
 	for i, ingress := range ingresses {
-		name := controller.RouterEffectiveDefaultCertificateSecretName(&ingress, operandNamespace)
+		// Check if ingress.Spec.DefaultCertificate is an available secret
+		// in the secrets slice. If it is not, attempt to fall back to the
+		// operator generated default certificate. If ingress.Spec.DefaultCertificate
+		// is updated to point to a non-existent secret, the certificate controller
+		// will not delete the operator generated default certificate.
+		// See https://bugzilla.redhat.com/show_bug.cgi?id=1887441
```
candita, line 76: It would be a good idea to log the issue with non-existent secret, because if there was a legitimate reason to change it and it couldn't be changed due to a typo for example, the user would want to know that.

sgreene570: The issue gets logged by the default certificate controller's reconcile loop here: https://github.com/openshift/cluster-ingress-operator/pull/505/files#diff-16d9c01d5eac10bd8a3012d74ef5feb542287746e676beca44479eb80cbaa124R36

Does that seem sufficient in this case?

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,77 @@ import (
...
 		// Nothing to do.
-	case desired == nil && current != nil:
+	case !wantCert && haveCert:
 		if deleted, err := r.deleteRouterDefaultCertificate(current); err != nil {
 			return false, fmt.Errorf("failed to delete default certificate: %v", err)
 		} else if deleted {
 			r.recorder.Eventf(ci, "Normal", "DeletedDefaultCertificate", "Deleted default wildcard certificate %q", current.Name)
-			return true, nil
+			return false, nil
```
candita, line 52: Why is this now considered false?

sgreene570: If deletion fails, we technically still have the resource, so we should return true there. If deletion succeeds, however, we should return false since we no long have the resource. Fixed, Thanks!

sgreene570: https://github.com/openshift/cluster-ingress-operator/pull/505#discussion_r537829578 explains the inversion here btw.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,77 @@ import (
...
+		// If the operator generated certificate is not being used, ensure that the ingress controller's
+		// Spec.DefaultCertificate secret exists before deleting the operator generated secret.
+		// See https://bugzilla.redhat.com/show_bug.cgi?id=1887441
+		err := r.lookupUserSpecifiedRouterDefaultCertificate(ci, namespace)
+		if err != nil {
+			return false, fmt.Errorf("failed to lookup user specified default certificate: %v", err)
+		}
+	}
+
+	haveCert, current, err := r.currentRouterDefaultCertificate(ci, namespace)
```
candita, line 40: Is it important to verify that the current cert secret also exists?  What if the `currentRouterDefaultCertificate` doesn't exist, was accidentally (or purposely) deleted?

sgreene570: `currentRouterDefaultCertificate` already does this and we handle any error/false condition from it, right?

sgreene570:
`level=error msg=Error: Error waiting for instance (i-0dce7576fc496afbf) to become ready: Failed to reach target state. Reason: Server.InternalError: Internal error on launch`
/test e2e-aws-operator

pkg/operator/controller/certificate/default_cert.go
candita: Looks like at the moment the reconciler doesn't even use the boolean!  https://github.com/openshift/cluster-ingress-operator/blob/175d7dfe2f97d0db2196706ea180b2a0e7c4da05/pkg/operator/controller/certificate/controller.go#L105

sgreene570: I noticed that as well. But it might in the future!

pkg/operator/controller/certificate/default_cert.go
candita: Looks like if it's not found then it returns false `haveCert` but nil `err`:
https://github.com/openshift/cluster-ingress-operator/pull/505/files#diff-16d9c01d5eac10bd8a3012d74ef5feb542287746e676beca44479eb80cbaa124R116

We don't handle/report any error on the inability to locate the current default certificate.  Then it just overwrites it with the existing default.  If I understand this properly.

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -17,67 +17,77 @@ import (
...
+	wantCert, desired, err := desiredRouterDefaultCertificateSecret(ca, namespace, deploymentRef, ci)
 	if err != nil {
 		return false, err
 	}
-	current, err := r.currentRouterDefaultCertificate(ci, namespace)
+	if !wantCert {
+		// If the operator generated certificate is not being used, ensure that the ingress controller's
+		// Spec.DefaultCertificate secret exists before deleting the operator generated secret.
+		// See https://bugzilla.redhat.com/show_bug.cgi?id=1887441
+		err := r.lookupUserSpecifiedRouterDefaultCertificate(ci, namespace)
```
candita, line 34: It might be better to move the call to `lookupUserSpecifiedRouterDefaultCertificate` to `desiredRouteDefaultCertificateSecret` https://github.com/openshift/cluster-ingress-operator/pull/505/files#diff-16d9c01d5eac10bd8a3012d74ef5feb542287746e676beca44479eb80cbaa124R79

pkg/operator/controller/certificate/default_cert.go
sgreene570: If we are unable to locate the current operator generated default certificate, but we want the operator to generate one, then we create it. In the case where `haveCert` and `wantCert` are true, we no-op, basically.

pkg/operator/controller/certificate/default_cert.go
sgreene570: Leaving the call to `r.lookupUserSpecifiedRouterDefaultCertificate` here would save us some API calls, so I think I would prefer to not refactor `desiredRouterDefaultCertificateSecret` in this case.

sgreene570: I think I would prefer not to conflate `desiredRouterDefaultCertificateSecret` with logic that checks if the user specified secret exists.

candita: I may have misunderstood the comments in https://github.com/openshift/cluster-ingress-operator/pull/505/files#diff-16d9c01d5eac10bd8a3012d74ef5feb542287746e676beca44479eb80cbaa124R144-R146

I was supposing that if you're checking `ci.Spec.DefaultCertificate != nil` in `desiredRouteDefaultCertificateSecret` then you would also want to perform `lookupUserSpecifiedRouterDefaultCertificate` there too (or instead).

pkg/operator/controller/certificate/default_cert.go
```diff
@@ -130,3 +140,15 @@ func (r *reconciler) deleteRouterDefaultCertificate(secret *corev1.Secret) (bool
...
 	return true, nil
 }
+
+// lookupUserSpecifiedRouterDefaultCertificate checks to see if the given ingress controller's
+// Spec.DefaultCertificate field corresponds to an existing secret. This function assumes that
+// ci.Spec.DefaultCertificate is not nil.
+func (r *reconciler) lookupUserSpecifiedRouterDefaultCertificate(ci *operatorv1.IngressController, namespace string) error {
+	secret := &corev1.Secret{}
+	name := controller.RouterEffectiveDefaultCertificateSecretName(ci, namespace)
+	if err := r.client.Get(context.TODO(), name, secret); err != nil {
```
candita, line 150: I think it might be best to check if the error is due to a system issue, like a timeout from the api.  If there was a timeout looking up the secret, we wouldn't want to generate a new one.

pkg/operator/controller/certificate/default_cert.go
candita: I think there's a special case if we're unable to find the secret due to a timeout on the api.   Wdyt?

pkg/operator/controller/certificate/default_cert.go
sgreene570: >  If there was a timeout looking up the secret, we wouldn't want to generate a new one.

I think this is the current behavior. If `lookupUserSpecifiedRouterDefaultCertificate` returns an error for any reason (API timeout, the secret does not exist, etc.), then `ensureDefaultCertificateForIngress` returns that error, and the reconcile loop will retry the process again.

pkg/operator/controller/certificate/default_cert.go
sgreene570: > I think there's a special case if we're unable to find the secret due to a timeout on the api. Wdyt?

In the case, `err != nil` on the next line would evaluate to true, `ensureDefaultCertificateForIngress` would return the error early, and the reconcile loop would retry the ensure again.

pkg/operator/controller/certificate/default_cert.go
sgreene570: > Maybe it would be a good idea to add a unit test?
Whoops, didn't see this comment before.
We generally don't have unit tests for reconciler methods. Perhaps there should be an e2e test to cover this case?

quarterpin:
/lgtm

Verified via pre-merge verification workflow, more reference related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1887441#c5

PR #506, opened 2020-12-08
add loadBalancerSourceRanges as a preserved field

matt-simons:
Case 02819675

As an OCP customer
I want to be able to set loadBalancerSourceRanges in the ingress service
so that I can restrict internet ingress to trusted sources

After upgrading from 4.5.x to 4.6.x we found that `loadBalancerSourceRanges` is no longer preserved when we patch our `router-default` Service. This regression is blocking our ability to upgrade to 4.6 in production as we can not restrict internet access to our cluster ingress to trusted sources.

This was a previously recommended solution https://access.redhat.com/solutions/5158751 

It seems during the refactor https://github.com/openshift/cluster-ingress-operator/compare/release-4.5...openshift:release-4.6 only `ClusterIP`, `ExternalIPs` and `HealthCheckNodePort` were considered as fields that other controllers or users could modify. This PR would add `LoadBalancerSourceRanges` to that list for the existing use case above.

Would it be possible to accept this PR and backport it to 4.6?

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -381,6 +381,7 @@ func loadBalancerServiceChanged(current, expected *corev1.Service, platform *con
 	// modified.
 	updated.Spec.ClusterIP = current.Spec.ClusterIP
 	updated.Spec.ExternalIPs = current.Spec.ExternalIPs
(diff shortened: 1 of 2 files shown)


Miciah:
#507 reverted the change that caused the operator to set `loadBalancerSourceRanges`, so we can close this PR.  
/close

PR #507, opened 2020-12-08
Bug 1905490: Revert "Support changing ingresscontroller load balancer scope"

Miciah:
This reverts #482.

PR #508, opened 2020-12-09
WIP: status: Report Upgradeable status condition

Miciah:
Add an "Upgradeable" status condition for ingresscontrollers and for the ingress clusteroperator.  Compute the clusteroperator's status condition from the individual ingresscontrollers'.  Compute each ingresscontroller's "Upgradeable" status by checking its load balance service (if the ingresscontroller has one).  If someone or something other than the operator has modified the service, mark the ingresscontroller as not upgradeable.  This ensures that the administrator cannot modify the service and then upgrade to a version of the operator that reverts the administrator's modification.

* `pkg/operator/controller/ingress/controller.go` (`Reconcile`): Compute the platform status from the infrastructure config using `GetPlatformStatus`.  Pass the platform status to `ensureIngressController`.
(`ensureIngressController`): Add a parameter for the platform status.  Pass the platform status instead of the infrastructure config to `ensureLoadBalancerService`.  Pass the deployment reference and platform status to `syncIngressControllerStatus`.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`externalLBAnnotations`): New variable.  Map platform type to the annotation for that platform that makes the load balancer external, if the platform requires an explicit annotation.
(`ensureLoadBalancerService`): Delete the parameter for the infrastructure config.  Add a parameter for the platform status.  Delete redundant computation of platform status (since it is now a parameter).  Delete computation of the PROXY protocol flag (`desiredLoadBalancerService` can compute it).
(`desiredLoadBalancerService`): Delete PROXY protocol parameter.  Instead, compute the value using `IsProxyProtocolNeeded`.
(`loadBalancerServiceScopeChanged`): New function.  Check if the load balancer scope's changed.  This function is only added for consistency with the main branch.
(`loadBalancerServiceChanged`): New function.  Check if the given load balancer services differ.
(`loadBalancerServiceIsUpgradeable`): New function.  Check if the given load balancer service differs from the expected service for the given ingresscontroller, deployment, and platform status.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestDesiredLoadBalancerService`): Refactor to be more table-driven.  Change test cases to specify which annotations are expected to be present, and which are expected to be absent, in order to simplify the test logic.
(`checkServiceHasAnnotation`): Fix logic for when `expectValue` is false.
(`TestLoadBalancerServiceChanged`): New test.
(`TestLoadBalancerServiceChangedScopeNeedsRecreate`): New test.
(`TestLoadBalancerServiceIsUpgradeable`): New test.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Add parameters for the deployment reference and platform status.  Use these parameters and the new `computeIngressUpgradeableCondition` function to compute the ingresscontroller's "Upgradeable" status condition.
(`computeIngressUpgradeableCondition`): New function.  Compute the ingresscontroller's "Upgradeable" status condition.
* `pkg/operator/controller/status/controller.go` (`Reconcile`): Use the new `computeOperatorUpgradeableCondition` function to compute the clusteroperator's "Upgradeable" status condition.
(`computeOperatorUpgradeableCondition`): New function.  Compute the clusteroperator's "Upgradeable" status condition from each ingresscontrollers' "Upgradeable" status condition.

----

Related to #507.


Miciah:
One of the control plane nodes failed to become ready.
/test e2e-aws-operator

sgreene570:
This depends on https://github.com/openshift/cluster-ingress-operator/pull/507, right?

Miciah:
> This depends on #507, right?

Yeah.  Also, I need to change the logic not to set `Upgradeable=False` when annotations have been modified.  But before all that, I need to make a PR to add similar logic on the master branch.  My plan is that 4.7 will set `Upgradeable=False` if the operator detects any external modifications to the service, and 4.6 will set `Upgradeable=False` if the operator detects any external modifications to the service other than to its annotations.

Miciah:
Looks like tests passed but must-gather failed.
/test e2e-aws-operator

Miciah:
must-gather failed again.
/test e2e-aws-operator

assets/router/service-cloud.yaml
```diff
@@ -23,3 +23,4 @@ spec:
     protocol: TCP
     port: 443
     targetPort: https
+  sessionAffinity: None
```
sgreene570, line 26: I'm curious to find out if any customers are currently setting `sessionAffinity` to a non-default value :)

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -404,3 +397,93 @@ func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev
...
+	}
+
+	if !want {
+		return nil
+	}
+
+	changed, updated := loadBalancerServiceExternallyModified(current, desired, platform)
+	if changed {
+		diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+		return fmt.Errorf("load balancer service has been modified; changes must be reverted before upgrading: %s", diff)
```
sgreene570, line 485: Great idea to log the diff here!

PR #510, opened 2020-12-09
WIP: Implement LB allowed source ranges API

Miciah:
* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Update allowed source ranges if needed.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`desiredLoadBalancerService`): Copy the value of the ingresscontroller's `spec.status.endpointPublishingStrategy.loadBalancerSourceRanges.allowedSourceRanges` field to the service's `spec.loadBalancerSourceRanges` field.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestLoadBalancerServiceChanged`): Add test case for allowed source ranges.
* `test/e2e/operator_test.go` (`TestAllowedSourceRanges`): New test.  Verify that the operator correctly configures the load balancer service and updates the service when the allowed source ranges are changed.

----

Depends on the following PR:

- [ ] https://github.com/openshift/api/pull/822


Miciah:
/retitle WIP: Bug 1906560: Implement LB allowed source ranges API

test/e2e/operator_test.go
```diff
@@ -758,6 +759,118 @@ func TestScopeChange(t *testing.T) {
...
+			return "", err
+		}
+
+		return response.Status, nil
+	}
+
+	// The initial source range of 127.0.0.0/8 should block all traffic.
+	// Poll for a minute and verify that we do not get an HTTP response.
+	gotResponse := false
+	err := wait.PollImmediate(10*time.Second, 1*time.Minute, func() (bool, error) {
```
sgreene570, line 835: nit: We _want_ a timeout error in this case, so maybe we should replace `err` with `_`.

test/e2e/operator_test.go
```diff
@@ -758,6 +759,118 @@ func TestScopeChange(t *testing.T) {
...
+		status, err := checkRoute()
+		if err != nil {
+			t.Logf("Got error from LB: %v", err)
```
sgreene570, line 838: Should we change this log message to indicate that an error from the LB is an expected thing? Wouldn't want to confuse anyone reading our test logs who doesn't know what this test does.

test/e2e/operator_test.go
```diff
@@ -758,6 +759,118 @@ func TestScopeChange(t *testing.T) {
...
+	t.Log("Clearing allowed source ranges")
+	if err := kclient.Get(context.TODO(), name, ic); err != nil {
+		t.Fatalf("failed to get ingresscontroller %s: %v", name, err)
+	}
+	ic.Spec.EndpointPublishingStrategy.LoadBalancer.AllowedSourceRanges = []operatorv1.CIDR{}
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatal(err)
+	}
+
+	err = wait.PollImmediate(10*time.Second, 2*time.Minute, func() (bool, error) {
```
sgreene570, line 860: Any particular reason that this loop has a timeout of 2 minutes, but the prior loop has a timeout of only 1 minute?

test/e2e/operator_test.go
Miciah: Fair point.  I suppose `err` is redundant since we have `gotResponse`.

test/e2e/operator_test.go
Miciah: Yeah, I wasn't sure how to word this.  The phrase "expected error" doesn't seem quite right because it implies that a specific error was expected.  I'll add "as expected" to the message.

test/e2e/operator_test.go
Miciah: I have a couple reasons, but I'm not sure how sound they are.  First, the first timeout is _expected_ to be hit (the request should always fail), so it's a question of how long we want the test to sit around making sure that the thing that should never happen never happens.  One minute seems like a reasonable amount of time to be confident that the LB really isn't accepting connections, without increasing the overall test time in the expected case too much.  Second, the second timeout is expected _not_ to be hit, and if we only hit it because the configuration change takes longer than expected, that's a test flake, so the high timeout usually won't increase the overall test time, and when it does, it reduces flakiness, which is worth waiting an extra minute (maybe even a couple extra minutes).  Does that sound reasonable?

sgreene570: > Does that sound reasonable?

Yup. We can always readjust these values in the future need be.

PR #515, opened 2020-12-15
Bug 1905324: Request appropriate memory for ingress-operator container

rfredette:
Cluster components are expected to request cpu & memory resources. ingress-operator already requested an amount of cpu time, but didn't include a memory request.

Following the guidelines [here](https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits), I ran `make test-e2e` against a cluster on AWS running a 4.7 nightly build. I queried prometheus for the metric `container_memory_usage_bytes{container="ingress-operator"}` every 15 seconds between the start and finish of the e2e run, then I sorted the data points by memory usage to find the 90th percentile, which in my run was 53,706,752 bytes, or about 51.2MiB.

Adding 10% overhead to that number produced 56.34MiB, which I rounded to 56MiB


sgreene570:
56 MiB sounds good to me (unless anyone can think of a reason we would want to round that to the nearest multiple of 5 or 10).

Miciah:
`TestNetworkLoadBalancer` failed, but it should pass now that #518 has merged.  
/test e2e-aws-operator

Miciah:
Must-gather failed.
/test e2e-aws-operator

Miciah:
/lgtm
Must-gather failed again.  
/test e2e-aws-operator

PR #519, opened 2021-01-04
NE-508 Support cluster HA mode

Miciah:
#### Bump openshift/api for cluster HA mode API

Bump to https://github.com/openshift/api/commit/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e to get the infrastructure topology API.

* `go.mod`: Bump.
* `go.sum`:

#### Set default spec.replicas to 1 if cluster is non-HA

* `pkg/operator/operator.go` (`ensureDefaultIngressController`): Check if `status.infrastructureTopology` is set to "SingleReplica" in the cluster infrastructure config.  If it is, specify only 1 replica in the default ingresscontroller.

---

Depends on the following PR:

- [X] https://github.com/openshift/api/pull/827

See also https://github.com/openshift/enhancements/pull/555 and https://github.com/openshift/enhancements/pull/560.


Miciah:
So many failures, but none obviously related to ingress.
/retest

pkg/operator/operator.go
```diff
@@ -168,31 +169,37 @@ func (o *Operator) Start(stop <-chan struct{}) error {
...
 	// Set the replicas field to a non-nil value because otherwise its
 	// persisted value will be nil, which causes GETs on the /scale
 	// subresource to fail, which breaks the scaling client.  See also:
 	// https://github.com/kubernetes/kubernetes/pull/75210
 	//
 	// TODO: Set the replicas value to the number of workers.
-	two := int32(2)
-	ic := &operatorv1.IngressController{
+	replicas := int32(2)
+	if infraConfig.Status.HighAvailabilityMode == configv1.NoneHighAvailabilityMode {
```
sgreene570, line 190: Note that https://github.com/openshift/api/pull/827 currently has the `SingleReplicaTopologyMode` constant instead of `NoneHighAvailabilityMode`. We can pick that up when 827 merges.

PR #522, opened 2021-01-08
Bug 1905100: Add "ingress.operator.openshift.io/hard-stop-after" annotation

frobware:
Annotating either an ingresscontroller or the ingress config with this
new annoation will redeploy the router and that will configure haproxy
to emit the haproxy "hard-stop-after" global option.

An ingresscontroller with a valid annotation set will override
ingresses.config/cluster (if set).

Examples:

Annotating the ingress config:

    $ oc annotate ingresses.config/cluster ingress.operator.openshift.io/hard-stop-after=1h

Annotating the "default" ingresscontroller:

    $ oc -n openshift-ingress-operator annotate ingresscontrollers/default ingress.operator.openshift.io/hard-stop-after=30m

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -54,6 +54,9 @@ const (
 
 	RouterDisableHTTP2EnvName          = "ROUTER_DISABLE_HTTP2"
 	RouterDefaultEnableHTTP2Annotation = "ingress.operator.openshift.io/default-enable-http2"
+
+	RouterHardStopAfterEnvName    = "ROUTER_HARD_STOP_AFTER"
+	RouterHardStopAfterAnnotation = "ingress.operator.openshift.io/hard-stop-after"
 )
 
 // ensureRouterDeployment ensures the router deployment exists for a given
@@ -134,6 +137,28 @@ func HTTP2IsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.In
 	}
 
 	return configHasHTTP2Enabled
+}
+
+// HardStopAfterIsEnabledByAnnotation returns true if the map m has
+// the key RouterHardStopAfterEnvName and its value is not the empty
+// string ("").
+func HardStopAfterIsEnabledByAnnotation(m map[string]string) (bool, string) {
+	if val, ok := m[RouterHardStopAfterAnnotation]; ok && len(val) > 0 {
+		return true, val
+	}
+	return false, ""
+}
+
+// HardStopAfterIsEnabled returns true if either the ingress
+// controller or the ingress config has the "hard-stop-after"
+// annotation. The presence of the annotation on the ingress
+// controller, irrespective of its value, always overrides any setting
+// on the ingress config.
+func HardStopAfterIsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.Ingress) (bool, string) {
+	if controllerAnnotation, controllerValue := HardStopAfterIsEnabledByAnnotation(ic.Annotations); controllerAnnotation {
+		return controllerAnnotation, controllerValue
+	}
+	return HardStopAfterIsEnabledByAnnotation(ingressConfig.Annotations)
 }
 
 // desiredRouterDeployment returns the desired router deployment.
@@ -621,6 +646,10 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: RouterDisableHTTP2EnvName, Value: "false"})
 	} else {
 		env = append(env, corev1.EnvVar{Name: RouterDisableHTTP2EnvName, Value: "true"})
+	}
+
+	if enabled, value := HardStopAfterIsEnabled(ci, ingressConfig); enabled {
+		env = append(env, corev1.EnvVar{Name: RouterHardStopAfterEnvName, Value: value})
 	}
 
 	deployment.Spec.Template.Spec.Volumes = volumes
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
@@ -140,11 +141,16 @@ func HTTP2IsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.In
 }
 
 // HardStopAfterIsEnabledByAnnotation returns true if the map m has
-// the key RouterHardStopAfterEnvName and its value is not the empty
-// string ("").
+// the key RouterHardStopAfterEnvName and its value is a valid HAProxy
+// time duration.
 func HardStopAfterIsEnabledByAnnotation(m map[string]string) (bool, string) {
 	if val, ok := m[RouterHardStopAfterAnnotation]; ok && len(val) > 0 {
-		return true, val
+		if clippedVal, err := clipHAProxyTimeoutValue(val); err != nil {
+			log.Error(err, "invalid HAProxy time value", "annotation", RouterHardStopAfterAnnotation, "value", val)
+			return false, ""
+		} else {
+			return true, clippedVal
+		}
 	}
 	return false, ""
 }
@@ -1070,4 +1076,47 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 	}
 	updated.Spec.Replicas = &replicas
 	return true, updated
+}
+
+// clipHAProxyTimeoutValue prevents the HAProxy config file from using
+// timeout values that exceed the maximum value allowed by HAProxy.
+// Returns an error in the event that a timeout string value is not
+// parsable as a valid time duration, or the clipped time duration
+// otherwise.
+//
+// TODO: this is a modified copy from openshift/router but returns ""
+// if there's any error.
+//
+// Ideally we need to share this utility function via:
+// https://github.com/openshift/library-go/blob/master/pkg/route/routeapihelpers
+func clipHAProxyTimeoutValue(val string) (string, error) {
+	const haproxyMaxTimeout = "2147483647ms" // max timeout allowable by HAProxy
+
+	endIndex := len(val) - 1
+	maxTimeout, err := time.ParseDuration(haproxyMaxTimeout)
+	if err != nil {
+		return "", err
+	}
+	// time.ParseDuration doesn't work with days
+	// despite HAProxy accepting timeouts that specify day units
+	if val[endIndex] == 'd' {
+		days, err := strconv.Atoi(val[:endIndex])
+		if err != nil {
+			return "", err
+		}
+		if maxTimeout.Hours() < float64(days*24) {
+			log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
+			return haproxyMaxTimeout, nil
+		}
+	} else {
+		duration, err := time.ParseDuration(val)
+		if err != nil {
+			return "", err
+		}
+		if maxTimeout.Milliseconds() < duration.Milliseconds() {
+			log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
+			return haproxyMaxTimeout, nil
+		}
+	}
+	return val, nil
 }
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -522,6 +522,7 @@ func TestDesiredRouterDeployment(t *testing.T) {
 	checkDeploymentHasEnvVar(t, deployment, "ROUTER_UNIQUE_ID_FORMAT", true, `"foo"`)
 
 	checkDeploymentHasEnvVar(t, deployment, "ROUTER_H1_CASE_ADJUST", false, "")
+	checkDeploymentHasEnvVar(t, deployment, RouterHardStopAfterEnvName, false, "")
 }
 
 func TestInferTLSProfileSpecFromDeployment(t *testing.T) {
--- a/test/e2e/hard_stop_after_test.go
+++ b/test/e2e/hard_stop_after_test.go
@@ -1,0 +1,404 @@
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
+	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// hardStopAfterRetryTimeout duration for retrying operations.
+	hardStopAfterRetryTimeout = 5 * time.Minute
+)
+
+// NOTE: This test will mutate the default ingresscontroller and the
+// "cluster" ingress configuration.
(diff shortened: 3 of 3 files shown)


frobware:
/hold

Will push some e2e tests.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -136,6 +139,28 @@ func HTTP2IsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.In
 	return configHasHTTP2Enabled
 }
 
+// HardStopAfterIsEnabledByAnnotation returns true if the map m has
+// the key RouterHardStopAfterEnvName and its value is not the empty
+// string ("").
+func HardStopAfterIsEnabledByAnnotation(m map[string]string) (bool, string) {
```
Miciah, line 145: No need to export this, is there?  Same for `HardStopAfterIsEnabled`.  (But `HTTP2IsEnabledByAnnotation` is already public, and anyway it's a minor thing, so I won't object if you prefer to keep this as-is.)
```suggestion
// hardStopAfterIsEnabledByAnnotation returns true if the map m has
// the key RouterHardStopAfterEnvName and its value is not the empty
// string ("").
func hardStopAfterIsEnabledByAnnotation(m map[string]string) (bool, string) {
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -136,6 +139,28 @@ func HTTP2IsEnabled(ic *operatorv1.IngressController, ingressConfig *configv1.In
...
+	if val, ok := m[RouterHardStopAfterAnnotation]; ok && len(val) > 0 {
+		return true, val
+	}
+	return false, ""
+}
+
+// HardStopAfterIsEnabled returns true if either the ingress
+// controller or the ingress config has the "hard-stop-after"
+// annotation. The presence of the annoation on the ingress
```
Miciah, line 154:
```suggestion
// annotation. The presence of the annotation on the ingress
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -623,6 +648,10 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: RouterDisableHTTP2EnvName, Value: "true"})
 	}
 
+	if enabled, value := HardStopAfterIsEnabled(ci, ingressConfig); enabled && len(value) > 0 {
```
Miciah, line 651: `HardStopAfterIsEnabled` already returns `false` if `len(value) == 0`.
```suggestion
	if enabled, value := HardStopAfterIsEnabled(ci, ingressConfig); enabled {
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -522,6 +522,7 @@ func TestDesiredRouterDeployment(t *testing.T) {
 	checkDeploymentHasEnvVar(t, deployment, "ROUTER_UNIQUE_ID_FORMAT", true, `"foo"`)
 
 	checkDeploymentHasEnvVar(t, deployment, "ROUTER_H1_CASE_ADJUST", false, "")
+	checkDeploymentHasEnvVar(t, deployment, RouterHardStopAfterEnvName, false, "")
```
Miciah, line 525: We should add test coverage somewhere in this unit test for when the annotation is set.

Miciah (commented):
Just some very minor suggestions Overall, it looks good.

pkg/operator/controller/ingress/deployment_test.go
frobware: > We should add test coverage somewhere in this unit test for when the annotation is set.

In progress.

pkg/operator/controller/ingress/deployment.go
frobware: This was belt-and-braces in lieu of any future change. Emitting `hard-stop-after` (i.e., no value) would mean haproxy fails to parse the config.  I can drop it, just wary of crash looping haproxy.

pkg/operator/controller/ingress/deployment.go
frobware: > No need to export this, is there? Same for `HardStopAfterIsEnabled`. (But `HTTP2IsEnabledByAnnotation` is already public, and anyway it's a minor thing, so I won't object if you prefer to keep this as-is.)

Currently used by e2e tests (not pushed).  If I don't need it then I'll change the scope.

pkg/operator/controller/ingress/deployment.go
Miciah: The unit test can verify that this code does the right thing if the annotation is set but has an empty value.

frobware: Dropped.

frobware:
/hold cancel

I pushed some tests. This is ready for a review.

test/e2e/hard_stop_after_test.go
```diff
@@ -0,0 +1,368 @@
...
+	if err := waitForIngressConfigHardStopAfterToBeUnset(client, timeout, ingressConfig); err != nil {
+		return err
+	}
+	if err := waitForIngressControllerHardStopAfterToBeUnset(client, timeout, ic); err != nil {
+		return err
+	}
+	return waitForRouterDeploymentHardStopAfterToBeUnset(client, timeout, routerDeployment)
+}
+
+func waitForIngressConfigHardStopAfterToBeSet(client client.Client, timeout time.Duration, c *configv1.Ingress, expectedDuration time.Duration) error {
```
sgreene570, line 180: could `waitForIngressConfigHardStopAfterToBeSet`, `waitForIngressConfigHardStopAfterToBeUnSet`, `waitForIngressControllerHardStopAfterToBeSet`, & `waitForIngressControllerHardStopAfterToBeUnSet` all be collapsed into one function with additional parameters for object (ingressconfig or ingresscontroller) and a set/unset bool?

frobware: Funnily enough I did that first time; second time I found it easier to read with the discrete functions.

test/e2e/hard_stop_after_test.go
```diff
@@ -0,0 +1,368 @@
...
+		t.Fatalf("failed to clear hard-stop-after on ingresscontroller: %v", err)
+	}
+	if err := WaitForHardStopAfterUnsetInAllComponents(kclient, hardStopAfterRetryTimeout, ic, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("some component still has hard-stop-after set: %v", err)
+	}
+}
+
+// NOTE: This test will mutate the default ingresscontroller and the
+// "cluster" ingress configuration.
+func TestRouteHardStopAfterEnableOnIngressControllerHasPriorityOverIngressConfig(t *testing.T) {
```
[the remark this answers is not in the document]
frobware: I think it's important to semantically to test this at runtime. They don't take a huge amount of time to run. The unit test coverage isn't huge - we just test that by default hard-stop-after is not set, which is not particularly extensive.  There's also some precedence here with the the http/2 enablement.

test/e2e/hard_stop_after_test.go
```diff
@@ -0,0 +1,368 @@
...
+	if err := WaitForHardStopAfterUnsetInAllComponents(kclient, hardStopAfterRetryTimeout, ic, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// First set hard-stop-after on the ingress config
+	if err := hardStopAfterTestIngressConfig(t, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// Secondly, oveerride setting by adding hard-stop-after on the ingresscontroller.
```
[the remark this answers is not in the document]
frobware: Fixed.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -623,6 +649,12 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: RouterDisableHTTP2EnvName, Value: "true"})
 	}
 
+	if enabled, value := HardStopAfterIsEnabled(ci, ingressConfig); enabled {
+		if _, err := time.ParseDuration(value); err != nil {
+			env = append(env, corev1.EnvVar{Name: RouterHardStopAfterEnvName, Value: value})
```
sgreene570, line 654: `time.ParseDuration` can't parse things that specify days (ie `1d`). We don't expect users to be setting values that high right?

sgreene570: Also if we fail to parse the annotation should we log that?

Miciah: I think the logic is inverted here:  We should use `value` only if `time.ParseDuration(value)` returns a nil error value.

Miciah: Perhaps `HardStopAfterIsEnabledByAnnotation` should perform the validation.  What should happen if `ingresses.config/cluster` has a valid annotation and then the user adds an invalid annotation to the ingresscontroller?  I think the first annotation should remain in effect, but with the current logic, the invalid annotation on the ingresscontroller causes the valid annotation on the ingress.config to be ignored.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1076,3 +1076,45 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 	updated.Spec.Replicas = &replicas
 	return true, updated
 }
+
+// clipHAProxyTimeoutValue prevents the HAProxy config file from using
+// timeout values that exceed the maximum value allowed by HAProxy.
+// Return the parameter string instead of an err in the event that a
+// timeout string value is not parsable as a valid time duration.
```
Miciah, line 1083: But it does return the error value too.

pkg/operator/controller/ingress/deployment.go
frobware: Done and added a test; though I will add an additional test that explicitly sets "" as a timeout value - just need to rearrange the test a bit now.

pkg/operator/controller/ingress/deployment.go
frobware: Updated.

test/e2e/hard_stop_after_test.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		t.Fatalf("failed to get test objects: %v", err)
+	}
+
+	if err := hardStopAfterTestIngressConfig(t, ingressConfig, routerDeployment); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// Cleanup
+	if err := clearHardStopAfterDurationForIngressConfig(t, kclient, hardStopAfterRetryTimeout, ingressConfig); err != nil {
```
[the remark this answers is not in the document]
frobware: This doesn't hold true in terms of moving them above any t.Fatal calls. It's only at the places that say `// cleanup` that you could make them into a defer block so I don't know what we would gain. I will remove the comments. If you want these to be rock solid and have no effect on any other test then the better thing to do would be to test against their own controller, config, deployment - but they will take longer to run.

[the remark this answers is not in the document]
frobware: /retest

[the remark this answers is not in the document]
sgreene570: So if any test assertions fail, then we don't need to revert the default ingress controller back to it's original non-annotated state?

test/e2e/hard_stop_after_test.go
```diff
@@ -0,0 +1,404 @@
...
+	if err := hardStopAfterTestIngressController(t, kclient, ic, routerDeployment, (500 * time.Minute).String()); err != nil {
+		t.Fatalf("test assertions failed: %v", err)
+	}
+
+	// Remove the controller setting.
+	if err := setHardStopAfterDurationForIngressController(t, kclient, hardStopAfterRetryTimeout, "", ic); err != nil {
+		t.Fatalf("failed to clear hard-stop-after on ingresscontroller: %v", err)
+	}
+
+	// Deployment should reveet back to config value
```
sgreene570, line 93: nit: reveet -> revert

test/e2e/hard_stop_after_test.go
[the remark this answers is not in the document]
sgreene570: http2 tests seem to follow a similar pattern to this PR. So maybe we should double check any possible defer improvements in a follow up for all e2e tests in general.

sgreene570:
e2e-aws-operator passed
/lgtm in the interest of time.

test/e2e/hard_stop_after_test.go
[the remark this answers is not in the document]
frobware: > defer improvements i

It's not clear to me what a defer improvement would look like.  The test has failed. CI will eventually fail, so there's nothing that would aid getting a green run. Mutating the same objects in multiple tests doesn't help things.

Should defer look like: 

```go
	defer func() {
		if err := setHardStopAfterDurationForIngressController(t, kclient, hardStopAfterRetryTimeout, "", ic); err != nil {
			t.Fatalf("failed to clear hard-stop-after on ingresscontroller: %v", err)
		}
		if err := waitForHardStopAfterUnsetInAllComponents(kclient, hardStopAfterRetryTimeout, ic, ingressConfig, routerDeployment); err != nil {
			t.Fatalf("some component still has hard-stop-after set: %v", err)
		}
	}()

	if err := hardStopAfterTestIngressController(t, kclient, ic, routerDeployment, (300 * time.Minute).String()); err != nil {
		t.Fatalf("test assertions failed: %v", err)
	}
```

Or?

```go
	defer func() {
		if err := waitForHardStopAfterUnsetInAllComponents(kclient, hardStopAfterRetryTimeout, ic, ingressConfig, routerDeployment); err != nil {
			t.Fatalf("some component still has hard-stop-after set: %v", err)
		}
	}()

	defer func() {
		if err := setHardStopAfterDurationForIngressController(t, kclient, hardStopAfterRetryTimeout, "", ic); err != nil {
			t.Fatalf("failed to clear hard-stop-after on ingresscontroller: %v", err)
		}
	}()

	if err := hardStopAfterTestIngressController(t, kclient, ic, routerDeployment, (300 * time.Minute).String()); err != nil {
		t.Fatalf("test assertions failed: %v", err)
	}
```

given that defer runs in LIFO.

frobware:
/hold

Need to redo `TestRouteHardStopAfterTestZeroLengthDuration` as it doesn't exercise setting a `""` value, it just clears the annotation  by virtue of calling `setHardStopAfterDurationForIngressController` which is not really the same thing.

test/e2e/hard_stop_after_test.go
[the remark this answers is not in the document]
frobware: FTR, TIL that fatal arranges for all defers to be called. Thank you @sgreene570 .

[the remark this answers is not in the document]
sgreene570: I think the former (grouping all cleanup code into one defer block) would make the most sense (if we decide we want to add any defers at all).

frobware:
/hold cancel

@sgreene570 I addressed https://github.com/openshift/cluster-ingress-operator/pull/522#issuecomment-758725438. PTAL.

sgreene570:
> /hold cancel
> 
> @sgreene570 I addressed [#522 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/522#issuecomment-758725438). PTAL.

Looks good. 
/lgtm

sgreene570:
`operator_test.go:402: failed waiting deployment openshift-ingress-operator/default to scale to 3: failed to achieve expected replicas, last observed: 2`
/test e2e-aws-operator

frobware:
> @frobware: #522 failed to apply on top of branch "release-4.6":
> 
> ```
> Applying: Add "ingress.operator.openshift.io/hard-stop-after" annotation
> Using index info to reconstruct a base tree...
> M	pkg/operator/controller/ingress/deployment.go
> M	pkg/operator/controller/ingress/deployment_test.go
> Falling back to patching base and 3-way merge...
> Auto-merging pkg/operator/controller/ingress/deployment_test.go
> CONFLICT (content): Merge conflict in pkg/operator/controller/ingress/deployment_test.go
> Auto-merging pkg/operator/controller/ingress/deployment.go
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> Patch failed at 0001 Add "ingress.operator.openshift.io/hard-stop-after" annotation
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> To restore the original branch and stop patching, run "git am --abort".
> ```

Manual backport here: https://github.com/openshift/cluster-ingress-operator/pull/535

PR #525, opened 2021-01-11
Bug 1915079: Disable canary route churn by default

sgreene570:
**canary: Disable canary route rotation by default**

Forcing the router to reload every 5 minutes for canary testing has a
heavy performance impact for larger clusters. By default, disable the
canary route rotation logic. Canary route rotation can be enabled for
internal use or debugging by setting the new
"ingress.operator.openshift.io/rotate-canary-route" annotation to
"true" on the default ingress controller.

pkg/operator/controller/canary/controller.go:
Add a mutex to the canary controller's reconciler type
to make the new reconciler `enableCanaryRouteRotation` field
go-routine safe. Set this field to true when the default ingress
controller has the new  canary route rotation annotation
set to true. Use this field in the canary check polling loop
to determine if the canary route endpoint should be rotated.

---

**test/e2e: Add canary route rotation annotation test**

test/e2e/canary_test.go: Add an e2e test to verify
that annotating the default ingress controller with the
canary route rotation annotation works as intended.

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -3,6 +3,7 @@ package canary
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
@@ -44,6 +45,14 @@ const (
 	// canaryCheckFailureCount is how many successive failing canary checks should
 	// be observed before the default ingress controller goes degraded.
 	canaryCheckFailureCount = 5
+
+	// CanaryRouteRotationAnnotation is an annotation on the default ingress controller
+	// that specifies whether or not the canary check loop should periodically rotate
+	// the endpoints of the canary route. Canary route rotation is disabled by default
+	// to prevent router reloads from impacting ingress performance periodically.
+	// Canary route rotation is enabled when the canary route rotation annotation has
+	// a value of "true" (disabled otherwise).
+	CanaryRouteRotationAnnotation = "ingress.operator.openshift.io/rotate-canary-route"
 )
 
 var (
@@ -57,8 +66,9 @@ var (
 // the canary service, daemonset, and route resources.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	reconciler := &reconciler{
-		config: config,
-		client: mgr.GetClient(),
+		config:                    config,
+		client:                    mgr.GetClient(),
+		enableCanaryRouteRotation: false,
 	}
 	c, err := controller.New(canaryControllerName, mgr, controller.Options{Reconciler: reconciler})
 	if err != nil {
@@ -166,6 +176,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return result, utilerrors.NewAggregate(errors)
 	}
 
+	// Get the canary route rotation annotation value
+	// from the default ingress controller.
+	ic := &operatorv1.IngressController{}
+	if err := r.client.Get(context.TODO(), request.NamespacedName, ic); err != nil {
+		errors = append(errors, fmt.Errorf("failed to get ingress controller %s: %v", request.NamespacedName.Name, err))
+		return result, utilerrors.NewAggregate(errors)
+	}
+
+	if val, ok := ic.Annotations[CanaryRouteRotationAnnotation]; ok {
+		v, _ := strconv.ParseBool(val)
+		r.mu.Lock()
+		r.enableCanaryRouteRotation = v
+		r.mu.Unlock()
+	}
+
 	// Start probing the canary route once the canary route
 	// has been admitted.
 	if checkRouteAdmitted(route) {
@@ -190,11 +215,23 @@ type reconciler struct {
 	config Config
 
 	client client.Client
+
+	// Use a mutex so enableCanaryRotation is
+	// go-routine safe.
+	mu                        sync.Mutex
+	enableCanaryRouteRotation bool
+}
+
+func (r *reconciler) isCanaryRouteRotationEnabled() bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.enableCanaryRouteRotation
 }
 
 func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	// Keep track of how many canary checks have passed
-	// so the route endpoint can be periodically cycled.
+	// so the route endpoint can be periodically cycled
+	// (when canary route rotation is enabled).
 	checkCount := 0
 
 	// Keep track of successive canary check failures
@@ -211,8 +248,13 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			log.Info("canary check route does not exist")
 			return
 		}
-		// Periodically rotate the canary route endpoint.
-		if checkCount > canaryCheckCycleCount {
+
+		// Check if canary route rotations are enabled every iteration.
+		rotationEnabled := r.isCanaryRouteRotationEnabled()
+
+		// Periodically rotate the canary route endpoint if
+		// rotationEnabled is true.
+		if rotationEnabled && checkCount > canaryCheckCycleCount {
 			haveService, service, err := r.currentCanaryService()
 			if err != nil {
 				log.Error(err, "failed to get canary service")
@@ -250,7 +292,12 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			log.Error(err, "error updating canary status condition")
 		}
 		successiveFail = 0
-		checkCount++
+		// Only increment checkCount if periodic canary route
+		// endpoint rotation is enabled to prevent unbounded
+		// integer growth.
+		if rotationEnabled {
+			checkCount++
+		}
 	}, canaryCheckFrequency, stop)
 
 	return nil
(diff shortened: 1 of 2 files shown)


sgreene570:
must-gather failed
/test e2e-aws-operator

test/e2e/canary_test.go
```diff
@@ -142,3 +147,75 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+// canary controller respect the canary route rotation
+// annotation for the default ingress controller.
+func TestCanaryRouteRotationAnnotation(t *testing.T) {
+	// Set the CanaryRouteRotation annotation to true
+	// on the default ingress controller.
+	if err := setDefaultIngressControllerRotationAnnotation(t, true); err != nil {
+		t.Fatalf("failed to set canary route rotation annotation: %v", err)
+	}
+
+	defer setDefaultIngressControllerRotationAnnotation(t, false)
```
frobware, line 161: How does the test fail if this fails?

sgreene570: fixed so the error is caught and we fail on that need be.

Miciah (commented):
/lgtm
/hold
I made a few small suggestions, but feel free to release the hold and let it merge as is (assuming CI starts passing).

test/e2e/canary_test.go
```diff
@@ -142,3 +147,83 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+		return true, nil
+	})
+
+	if err != nil {
+		t.Fatalf("failed to observe canary route rotation: %v", err)
+	}
+}
+
+func setDefaultIngressControllerRotationAnnotation(t *testing.T, val bool) error {
+	ic := &operatorv1.IngressController{}
```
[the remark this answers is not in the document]
sgreene570: fixed.

test/e2e/canary_test.go
```diff
@@ -142,3 +147,83 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
 		},
 	}
 }
+
+// TestCanaryRouteRotationAnnotation verifies that the
+// canary controller respect the canary route rotation
```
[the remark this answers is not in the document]
sgreene570: fixed

pkg/operator/controller/canary/controller.go
```diff
@@ -211,8 +251,9 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			log.Info("canary check route does not exist")
 			return
 		}
-		// Periodically rotate the canary route endpoint.
-		if checkCount > canaryCheckCycleCount {
+		// Periodically rotate the canary route endpoint if
+		// r.isCanaryRouteRotationEnabled() is true.
+		if r.isCanaryRouteRotationEnabled() && checkCount > canaryCheckCycleCount {
```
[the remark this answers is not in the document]
sgreene570: right, I noticed this as well. Fixed.

pkg/operator/controller/canary/controller.go
```diff
@@ -166,6 +174,26 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+	// Get the canary route rotation annotation value
+	// from the default ingress controller.
+	ic := &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      manifests.DefaultIngressControllerName,
+			Namespace: r.config.Namespace,
+		},
+	}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ic.Namespace, Name: ic.Name}, ic); err != nil {
+		errors = append(errors, fmt.Errorf("failed to get ingress controller %s: %v", ic.Name, err))
```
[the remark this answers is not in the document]
sgreene570: Right, might as well take advantage of the reconicle request param! fixed.

pkg/operator/controller/canary/controller.go
```diff
@@ -166,6 +174,26 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		return result, utilerrors.NewAggregate(errors)
+	}
+
+	if val, ok := ic.Annotations[CanaryRouteRotationAnnotation]; ok && len(val) > 0 {
```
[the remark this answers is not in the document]
sgreene570: right, fixed.

pkg/operator/controller/canary/controller.go
```diff
@@ -44,6 +45,12 @@ const (
 	// canaryCheckFailureCount is how many successive failing canary checks should
 	// be observed before the default ingress controller goes degraded.
 	canaryCheckFailureCount = 5
+
+	// canaryRouteRotationAnnotation is an annotation on the default ingress controller
+	// that determines whether or not the canary check loop should periodically rotate
+	// the endpoints of the canary route. Canary route rotation is disable by default
+	// to prevent router reloads from impacting ingress performance periodically.
```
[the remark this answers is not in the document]
sgreene570: go-docs matter!
fixed.

sgreene570:
@Miciah I pushed fixes for your suggestions despite CI being green since recently merged changes to this repo would have triggered retests.

PTAL when you can.

pkg/operator/controller/canary/controller.go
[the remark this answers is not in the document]
candita: +1

pkg/operator/controller/canary/controller.go
```diff
@@ -44,6 +45,12 @@ const (
 	// canaryCheckFailureCount is how many successive failing canary checks should
 	// be observed before the default ingress controller goes degraded.
 	canaryCheckFailureCount = 5
+
```
candita, line 48: If this is a user-facing annotation, should this annoation be documented elsewhere?   Also, please add that the user must set the value as true or false, i.e. the annotation implies false if its value is not set to true.

sgreene570: I enhanced the go-doc per your request. 
We don't expect users to want to opt-in to this functionality given the performance impact on the default cluster routes, so I think the documentation here is sufficient. This annotation would be useful for debugging a customer case via bugzilla or our support engineers, or for internal debugging / testing.

pkg/operator/controller/canary/controller.go
```diff
@@ -166,6 +176,21 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return result, utilerrors.NewAggregate(errors)
 	}
 
+	// Get the canary route rotation annotation value
+	// from the default ingress controller.
+	ic := &operatorv1.IngressController{}
+	if err := r.client.Get(context.TODO(), request.NamespacedName, ic); err != nil {
+		errors = append(errors, fmt.Errorf("failed to get ingress controller %s: %v", ic.Name, err))
```
Miciah, line 183: Is `ic.Name` necessarily set if the get failed?  (Doesn't really matter much given that there's only one ingresscontroller that it could be.)

test/e2e/canary_test.go
```diff
@@ -142,3 +147,84 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+		return true, nil
+	})
+
+	if err != nil {
+		t.Fatalf("failed to observe canary route rotation: %v", err)
+	}
+}
+
+func setDefaultIngressControllerRotationAnnotation(t *testing.T, val bool) error {
+	t.Helper()
```
frobware, line 209: Is this necessary given we only appear to access `t` for logging?

test/e2e/canary_test.go
```diff
@@ -142,3 +147,84 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+	// canary checks. Canary checks happen once a minute.
+	updatedCanaryRoute := &routev1.Route{}
+	err = wait.PollImmediate(5*time.Second, 8*time.Minute, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), name, updatedCanaryRoute); err != nil {
+			t.Logf("failed to get canary route %s: %v", name, err)
+			return false, nil
+		}
+		// If the canaryRoute and updatedCanaryRoute do not have the same targetPort,
+		// then the canary route rotation annotation is working.
+		if cmp.Equal(canaryRoute.Spec.Port.TargetPort, updatedCanaryRoute.Spec.Port.TargetPort) {
```
frobware, line 196: nit: collapse the multiple returns to the result of the `Equal` expression.

pkg/operator/controller/canary/controller.go
sgreene570: Im going to assume it does not get set......fixed. 
Thanks!

Miciah:
Latest changes look good.  Thanks!
/lgtm
/hold
in case @frobware wants to review.

frobware (approved):
LGTM. Minor nits/observations but nothing to hold LGTM for.

test/e2e/canary_test.go
sgreene570: Miciah suggested adding this:
```
You can call t.Helper() at the start of this function to cause any log output that this function generates to have the caller's line number.
```

frobware: I thought this was only useful when you want to fail or error. No big deal.

Miciah: The log messages will have line numbers, right?  The line numbers of the call sites are more specific than line numbers in the helper.

sgreene570:
hitting VPC limits

PR #526, opened 2021-01-13
Bug 1905778: Fix inconsistent ingress operator status after upgrade

rfredette:
`status.DeploymentDegraded` was removed in favor of several more specific status fields in 4.6, but if the cluster was upgraded from 4.5 or earlier, the existing `DeploymentDegraded` status will continue to be reported

If the cluster was upgraded from 4.1 or earlier, `status.EndpointPublishingStrategy.LoadBalancer` (introduced in 4.2?) may not be present when `status.EndpointPublishingStrategy.Type` is set to `LoadBalancerService`. Internally, the assumption is made that this is equivalent to `status.EndpointPublishingStrategy.LoadBalancer` being set to `.LoadBalancer.scope: External`, but this PR makes that implicit assumption explicit.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -230,6 +230,18 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	if ingress.Status.EndpointPublishingStrategy != nil && ingress.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType && ingress.Status.EndpointPublishingStrategy.LoadBalancer == nil {
+		log.Info("EndpointPublishingStrategy missing LoadBalancer")
+		// set loadbalancer to external
+		ingress.Status.EndpointPublishingStrategy.LoadBalancer = &operatorv1.LoadBalancerStrategy{
+			Scope: operatorv1.ExternalLoadBalancer,
+		}
+		if err := r.client.Status().Update(context.TODO(), ingress); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update status: %v", err)
+		}
+		return reconcile.Result{Requeue: true}, nil
+	}
+
 	// The ingresscontroller is safe to process, so ensure it.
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -105,6 +106,18 @@ func MergeConditions(conditions []operatorv1.OperatorCondition, updates ...opera
 	return conditions
 }
 
+// PruneConditions removes any conditions that are not currently supported.
+// Returns the updated condition array.
+func PruneConditions(conditions []operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
+	for i, condition := range conditions {
+		switch condition.Type {
+		case "DeploymentDegraded":
```
sgreene570, line 114: We stopped using `DeploymentDegraded` in 4.6, right? Might be worth dropping a comment above indicating that this condition exists for clusters < 4.6, to avoid any confusion.

rfredette: That's a good idea. I'll include a comment about it to clarify

pkg/operator/controller/ingress/controller.go
```diff
@@ -230,6 +230,18 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	if ingress.Status.EndpointPublishingStrategy != nil && ingress.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType && ingress.Status.EndpointPublishingStrategy.LoadBalancer == nil {
+		log.Info("EndpointPublishingStrategy missing LoadBalancer")
+		// set loadbalancer to external
```
Miciah, line 235: The log message could be a little more explicit, and the comment is superfluous.
```suggestion
		log.Info("Setting default value for empty status.endpointPublishingStrategy.loadBalancer field", "ingresscontroller", ingress)
```

Miciah (commented):
Please add to-do comments to remove this fix-up logic in 4.8.  Once the operator has done the fix-ups, it shouldn't be necessary to perform the same fix-ups again, so we only need the fix-ups in one release (4.7) and don't need them in subsequent releases (4.8 and later).

Miciah:
Thanks!
/lgtm

PR #529, opened 2021-01-14
Bug 1916401: DNS: Skip deleting records that were not published.

sgreene570:
**pkg/operator/controller/dns/controller.go:**

When deleting DNS records from the provider, only delete records that
have been successfully published by the provider. This resolves
BZ#1916401, in which an ingress controller created with a domain
that refers to a non-existent zone cannot be deleted. If the provider
failed to publish the DNS record, it will also fail to delete it.
It is safe to delete a DNSRecord resource in the cluster when the record
does not exist in the provider, since no upstream record needs to be
deleted first.

---


sgreene570:
`failed to delete ingresscontroller test: timed out waiting for ingresscontroller to be deleted`
/test e2e-aws-operator

PR #530, opened 2021-01-15
Bug 1917614: get hostedzone tags from route53 on AWS C2S

staebler:
When running in an AWS C2S region, the ingress operator has two issues
* It needs to use a custom CA bundle to access the AWS API.
* It cannot use the resourcegrouptagging API since that is not available.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -86,6 +86,8 @@ type Config struct {
 	// ServiceEndpoints is the list of AWS API endpoints to use for
 	// Provider clients.
 	ServiceEndpoints []ServiceEndpoint
+	// CustomCABundle is a custom CA bundle to use when accessing the AWS API
+	CustomCABundle string
 }
 
 // ServiceEndpoint stores the configuration of a custom url to
@@ -101,10 +103,14 @@ type ServiceEndpoint struct {
 }
 
 func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error) {
-	sess, err := session.NewSessionWithOptions(session.Options{
+	sessionOpts := session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 		SharedConfigFiles: []string{config.SharedCredentialFile},
-	})
+	}
+	if config.CustomCABundle != "" {
+		sessionOpts.CustomCABundle = strings.NewReader(config.CustomCABundle)
+	}
+	sess, err := session.NewSessionWithOptions(sessionOpts)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create AWS client session: %v", err)
 	}
@@ -146,6 +152,10 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 		// in the same region as the Route53 client to find the hosted zone
 		// of managed records.
 		tagConfig = tagConfig.WithRegion(endpoints.UsGovWest1RegionID)
+	case endpoints.UsIsoEast1RegionID:
+		// Do not override the region for the C2S region
+		tagConfig = tagConfig.WithRegion(region)
+		r53Config = r53Config.WithRegion(region)
 	default:
 		// Since Route 53 is not a regionalized service, the Tagging API will
 		// only return hosted zone resources when the region is "us-east-1".
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -46,6 +47,8 @@ const (
 var (
 	_   dns.Provider = &Provider{}
 	log              = logf.Logger.WithName("dns")
+
+	hostedZoneIDRegex = regexp.MustCompile("^/?hostedzone/([^/]+)$")
 )
 
 // Provider is a dns.Provider for AWS Route53. It only supports DNSRecords of
@@ -153,8 +156,9 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 		// of managed records.
 		tagConfig = tagConfig.WithRegion(endpoints.UsGovWest1RegionID)
 	case endpoints.UsIsoEast1RegionID:
-		// Do not override the region for the C2S region
-		tagConfig = tagConfig.WithRegion(region)
+		// The resourcetagging API is not available in C2S
+		tagConfig = nil
+		// Do not override the region in C2s
 		r53Config = r53Config.WithRegion(region)
 	default:
 		// Since Route 53 is not a regionalized service, the Tagging API will
@@ -177,6 +181,10 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 				r53Config = r53Config.WithEndpoint(ep.URL)
 				log.Info("using route53 custom endpoint", "url", ep.URL)
 			case ep.Name == TaggingService:
+				if tagConfig == nil {
+					log.Info("found resourcegroupstaggingapi custom endpoint which will be ignored since the %s region does not support that API", region)
+					continue
+				}
 				tagFound = true
 				url := ep.URL
 				// route53 for govcloud is based out of us-gov-west-1,
@@ -193,13 +201,17 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 			}
 		}
 	}
+	var tags *resourcegroupstaggingapi.ResourceGroupsTaggingAPI
+	if tagConfig != nil {
+		tags = resourcegroupstaggingapi.New(sess, tagConfig)
+	}
 	p := &Provider{
 		elb: elb.New(sess, elbConfig),
 		// TODO: Add custom endpoint support for elbv2. See the following for details:
 		// https://docs.aws.amazon.com/general/latest/gr/elb.html
 		elbv2:     elbv2.New(sess, aws.NewConfig().WithRegion(region)),
 		route53:   route53.New(sess, r53Config),
-		tags:      resourcegroupstaggingapi.New(sess, tagConfig),
+		tags:      tags,
 		config:    config,
 		idsToTags: map[string]map[string]string{},
 		lbZones:   map[string]string{},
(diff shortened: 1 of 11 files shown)


pkg/dns/aws/dns.go
```diff
@@ -244,7 +267,25 @@ func (m *Provider) getZoneID(zoneConfig configv1.DNSZone) (string, error) {
...
+	} else {
+		id, err = m.lookupZoneIDWithoutResourceTagging(zoneConfig)
+	}
+	if err != nil {
+		return id, err
+	}
+
+	// Update the cache
+	m.idsToTags[id] = zoneConfig.Tags
+	log.Info("found hosted zone using tags", "zone id", id, "tags", zoneConfig.Tags)
```
sgreene570, line 282: This log statement should be updated, since we may have gotten here with or without tags, right?

pkg/operator/controller/dns/controller.go
```diff
@@ -54,6 +54,13 @@ const (
 	// operator's namespace that will hold the credentials that the operator
 	// will use to authenticate with the cloud API.
 	cloudCredentialsSecretName = "cloud-credentials"
+
+	// kubeCloudConfigNamespace is the namespace where the kube cloud config ConfigMap is located
+	kubeCloudConfigNamespace = "openshift-config-managed"
```
sgreene570, line 59: This namespace constant is already available here: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/names.go#L16

Would you mind using the existing definition in place of a new constant?

pkg/dns/aws/dns.go
```diff
@@ -101,10 +106,14 @@ type ServiceEndpoint struct {
...
 
 func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error) {
-	sess, err := session.NewSessionWithOptions(session.Options{
+	sessionOpts := session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 		SharedConfigFiles: []string{config.SharedCredentialFile},
-	})
+	}
+	if config.CustomCABundle != "" {
+		sessionOpts.CustomCABundle = strings.NewReader(config.CustomCABundle)
```
sgreene570, line 114: In the event that `config.CustomCABundle` is empty when it should not be (ie on a c2s cluster), will `NewSessionWithOptions` return an error? Or where would we error out in that case?

pkg/operator/controller/dns/controller.go
staebler: Yep. I will make that change.

pkg/dns/aws/dns.go
staebler: If the custom CA bundle is not present but required, the `NewSessionWithOptions` call will still succeed. That call does not actually make any calls to the AWS API. The errors will come in the AWS API calls (eg, the `ListHostedZonesPages` call). The errors will be x509 errors and look something like the following.
```
RequestError: send request failed
caused by: Post \"https:/route53.c2s.ic.gov/\": x509: certificate signed by unknown authority" 
```

staebler:
970c5d2c...bd796a6c
* Use `GlobalMachineSpecifiedConfigNamespace` constant.

pkg/dns/aws/dns.go
staebler: Ultimately, you will see this as a failure to create the DNS provider, with an error from the call to `validateServiceEndpoints`.

staebler:
bd796a6c...ecc9ce16
* This time **actually** use `GlobalMachineSpecifiedConfigNamespace` and not some other random constant.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -569,3 +572,59 @@ func TestRecordIsAlreadyPublishedToZone(t *testing.T) {
...
+			corev1.AddToScheme(scheme)
+			resources := []runtime.Object{}
+			if tc.cm != nil {
+				resources = append(resources, tc.cm)
+			}
+			client := fake.NewFakeClientWithScheme(scheme, resources...)
+			r := reconciler{client: client}
+			actualCABundle, err := r.customCABundle()
+			if err != nil {
+				t.Fatalf("unexpected error from useCustomCABundle: %v", err)
```
Miciah, line 623:
```suggestion
				t.Fatalf("unexpected error from r.customCABundle: %v", err)
```

pkg/dns/aws/dns.go
```diff
@@ -46,6 +47,8 @@ const (
 var (
 	_   dns.Provider = &Provider{}
 	log              = logf.Logger.WithName("dns")
+
+	hostedZoneIDRegex = regexp.MustCompile("^/?hostedzone/(.+)$")
```
Miciah, line 51: Can the hosted zone id have a slash?  I'm asking because the old logic did not permit slashes in hosted zone ids, but this regexp does, and I'm just curious if the change was deliberate.

pkg/dns/aws/dns.go
```diff
@@ -167,6 +181,9 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 				r53Config = r53Config.WithEndpoint(ep.URL)
 				log.Info("using route53 custom endpoint", "url", ep.URL)
 			case ep.Name == TaggingService:
+				if tagConfig == nil {
+					continue
+				}
```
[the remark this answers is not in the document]
staebler: I am fine with logging a warning if you like. The user should never specify a custom service endpoint for the resourcegrouptagging API when running in a C2S region, however. I contend that warning about the service endpoint is not terribly helpful since the ingress operator will not be accessing that endpoint regardless.

As an aside, the logic here for the custom endpoints should be taking into consideration the region specified for the custom endpoints. For example, the custom endpoints could contain a specification for the endpoint to use when accessing us-west-1, which should not be considered if the region installed to is actually us-east-1. Before using a custom endpoint, there should be a check that the region of the service endpoint matches the region of the installation.

pkg/dns/aws/dns.go
staebler: Yes, the change is deliberate.
The ID in the output from `ListHostedZones` starts with a slash.
```
$ aws route53 list-hosted-zones --max-items=1 | jq '.HostedZones[].Id'
"/hostedzone/Z0799914V7ZYCRPWCWYK"
```
The resource ARN in the output from `GetResources` does not have a slash before "hostedzone".
```
$ aws resourcegroupstaggingapi get-resources --resource-type-filters route53:hostedzone --max-items=1 | jq ".ResourceTagMappingList[].ResourceARN"
"arn:aws:route53:::hostedzone/Z03898872UPH0MB9MUBXE"
```

staebler:
ecc9ce16...989bc712
* Fix name of function called used in test failure message.
* Log when provided tagging custom endpoint is ignored.

pkg/dns/aws/dns.go
sgreene570: (I could be wrong) but I think @Miciah was referring to how this regex permits slashes _in_ the hosted zone ID (the part after `/hostedzone/`), ie `/hostedzone/abc/123`. From my understanding, the prior code splits the entire string on `/`'s, and only allows 2 results https://github.com/openshift/cluster-ingress-operator/pull/530/files#diff-1a81cbdf88bfc6111c38c87b8d2c9d360f6e44f4b76b1f5c11435a3fd3c47cfaL261.

staebler: Ah, sorry. I misunderstood. I can add restrictions that will error on extra slashes.

staebler:
989bc712...5660b43d
* Fail zone ID lookup if there are extra slashes after "hostedzone/" in the resource ID.

staebler:
The e2e-aws-operator tests are passing. The failure is coming from capturing the must-gather logs afterwards.
```
error: some steps failed:
  * could not run steps: step e2e-aws-operator failed: "e2e-aws-operator" post steps failed: "e2e-aws-operator" pod "e2e-aws-operator-gather-must-gather" failed: the pod ci-op-gbw3m91q/e2e-aws-operator-gather-must-gather failed after 11m34s (failed containers: test): ContainerFailed one or more containers exited

Container test exited with code 1, reason Error
---
Running must-gather...
error: gather never finished for pod must-gather-48946: timed out waiting for the condition
```
Looking over test history, this looks to be common flake and not related to the changes in this PR.

pkg/dns/aws/dns.go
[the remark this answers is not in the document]
Miciah: > The user should never specify a custom service endpoint for the resourcegrouptagging API when running in a C2S region, however. I contend that warning about the service endpoint is not terribly helpful since the ingress operator will not be accessing that endpoint regardless.

Maybe the user should never, but it looks like the user can.  The relevant API godoc just says, "serviceEndpoints list contains custom endpoints which will override default service endpoint of AWS Services."  If the user doesn't specify custom endpoints, the operator figures out which endpoints to use based on the region, but if the user _does_ specify custom endpoints, it seems like operators must use those custom endpoints to comply with the API.  That said, I did not write the API, I am only trying to interpret it, and I'll defer your judgment here.  

> As an aside, the logic here for the custom endpoints should be taking into consideration the region specified for the custom endpoints. For example, the custom endpoints could contain a specification for the endpoint to use when accessing us-west-1, which should not be considered if the region installed to is actually us-east-1. Before using a custom endpoint, there should be a check that the region of the service endpoint matches the region of the installation.

How can the user specify a region for a custom endpoint?  The API has no field to specify an endpoint's region: 
```go
// AWSServiceEndpoint store the configuration of a custom url to
// override existing defaults of AWS Services.
type AWSServiceEndpoint struct {
	// name is the name of the AWS service.
	// The list of all the service names can be found at https://docs.aws.amazon.com/general/latest/gr/aws-service-information.html
	// This must be provided and cannot be empty.
	//
	// +kubebuilder:validation:Pattern=`^[a-z0-9-]+$`
	Name string `json:"name"`

	// url is fully qualified URI with scheme https, that overrides the default generated
	// endpoint for a client.
	// This must be provided and cannot be empty.
	//
	// +kubebuilder:validation:Pattern=`^https://`
	URL string `json:"url"`
}
```
https://github.com/openshift/api/blob/bb6857301efec6d005852debef3030781c26dd78/config/v1/types_infrastructure.go#L239-L255

Miciah:
Thanks!
/lgtm

pkg/dns/aws/dns.go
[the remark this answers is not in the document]
staebler: Regarding the region, I don't know the history here, but I would have expected the operator to read the service endpoints from the cloud config rather than from the Infrastructure resource. The service endpoints in the Infrastructure resource are inputs that the cluster-config-opreator uses to generate the cloud config. In the cloud config, each service endpoint is specific to a region.

https://github.com/kubernetes/kubernetes/blob/35bf00acb1f5f2be22cdcb180c5e1227a72934fc/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L626-L646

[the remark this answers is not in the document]
staebler: > if the user does specify custom endpoints, it seems like operators must use those custom endpoints to comply with the API

I think that we are compliant with that. We use the custom endpoint specified for every call to the resourcegroupstagging API. In the case of C2S, we do not make any calls to the resourcegroupstagging API.

PR #531, opened 2021-01-15
Bug 1916706: go.mod: Bump go.uber.org/zap to v1.16.0

sgreene570:
This PR attempts to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1916706 by updating the zap logging dependency to v1.16.0, which includes https://github.com/uber-go/zap/pull/854.


Miciah:
`encodeStringer` already had logic to recover panics.  It looks like https://github.com/uber-go/zap/pull/854 only changes what the recovery logic prints.  Why didn't the recovery logic prevent the panic before, and how does this change prevent the panic?

sgreene570:
> `encodeStringer` already had logic to recover panics. It looks like [uber-go/zap#854](https://github.com/uber-go/zap/pull/854) only changes what the recovery logic prints. Why didn't the recovery logic prevent the panic before, and how does this change prevent the panic?

I don't think this change would prevent the panic, after further investigating. /close for now (@frobware feel free to /reopen if you disagree).

PR #533, opened 2021-01-19
Bug 1917942: canary: Check canary service for nil elements

sgreene570:
**pkg/operator/controller/canary/route.go:** 
Check that the canary service is not nil and that the canary service has a non-empty spec.ports field when building the canary route via `desiredCanaryRoute` to avoid nil dereferences.

**pkg/operator/controller/canary/route_test.go:** 
Ensure `desiredCanaryRoute` does not return any errors.

---

These changes resolve [BZ 1917942](https://bugzilla.redhat.com/show_bug.cgi?id=1917942), in which canary controller panic's were observed in OpenShift CI.

### DIFF
--- a/pkg/operator/controller/canary/route_test.go
+++ b/pkg/operator/controller/canary/route_test.go
@@ -19,7 +19,11 @@ func TestDesiredCanaryRoute(t *testing.T) {
 		Name: "test",
 	}
 	service := desiredCanaryService(daemonsetRef)
-	route := desiredCanaryRoute(service)
+	route, err := desiredCanaryRoute(service)
+
+	if err != nil {
+		t.Fatalf("desiredCanaryService returned an error: %v", err)
+	}
 
 	expectedRouteName := types.NamespacedName{
 		Namespace: "openshift-ingress-canary",
@@ -106,7 +110,10 @@ func TestCanaryRouteChanged(t *testing.T) {
 	service := desiredCanaryService(daemonsetRef)
 
 	for _, tc := range testCases {
-		original := desiredCanaryRoute(service)
+		original, err := desiredCanaryRoute(service)
+		if err != nil {
+			t.Fatalf("desiredCanaryService returned an error: %v", err)
+		}
 		mutated := original.DeepCopy()
 		tc.mutate(mutated)
 		if changed, updated := canaryRouteChanged(original, mutated); changed != tc.expect {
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/canary/route_test.go
```diff
@@ -19,7 +19,11 @@ func TestDesiredCanaryRoute(t *testing.T) {
 		Name: "test",
 	}
 	service := desiredCanaryService(daemonsetRef)
-	route := desiredCanaryRoute(service)
+	route, err := desiredCanaryRoute(service)
+
+	if err != nil {
+		t.Errorf("desiredCanaryService returned an error: %v", err)
```
Miciah, line 25: No sense in continuing with the unit test if we don't have a service, right?  
```suggestion
		t.Fatalf("desiredCanaryService returned an error: %v", err)
```

pkg/operator/controller/canary/route_test.go
```diff
@@ -106,7 +110,10 @@ func TestCanaryRouteChanged(t *testing.T) {
 	service := desiredCanaryService(daemonsetRef)
 
 	for _, tc := range testCases {
-		original := desiredCanaryRoute(service)
+		original, err := desiredCanaryRoute(service)
+		if err != nil {
+			t.Errorf("desiredCanaryService returned an error: %v", err)
```
Miciah, line 115:
```suggestion
			t.Fatalf("desiredCanaryService returned an error: %v", err)
```

pkg/operator/controller/canary/route_test.go
```diff
@@ -106,7 +110,10 @@ func TestCanaryRouteChanged(t *testing.T) {
 	service := desiredCanaryService(daemonsetRef)
 
 	for _, tc := range testCases {
-		original := desiredCanaryRoute(service)
+		original, err := desiredCanaryRoute(service)
+		if err != nil {
```
candita, line 114: Consider adding a test case that produces an error, and checking tc.expect to make sure the error is expected.

sgreene570: I thought about adding these, but didn't in the interest of time. Maybe that would be a good thing to add in a future cleanup? I could add these test cases to my current cleanup PR, https://github.com/openshift/cluster-ingress-operator/pull/520.

candita: It should be quick to add a new function to exercise the validation in desiredCanaryRoute, and check that an error was produced for invalid input.  It doesn't have to be a part of this function.

sgreene570: See https://github.com/openshift/cluster-ingress-operator/pull/520/commits/d6462c265642771af4be6c992af662c7db342a0c

PR #537, opened 2021-01-25
Bug 1919151: Azure: Log when using domain not in zone

sgreene570:
pkg/dns/azure/dns.go: Log when using domain not in zone

Modify getARecordName so that the DNS controller logs a message when
DNS record with domain strings that do not contain the given zone name
are updated.

### DIFF
--- a/pkg/dns/azure/dns.go
+++ b/pkg/dns/azure/dns.go
@@ -130,7 +130,13 @@ func (m *provider) Replace(record *iov1.DNSRecord, zone configv1.DNSZone) error 
 }
 
 // getARecordName extracts the ARecord subdomain name from the full domain string.
-// azure defines the ARecord Name as the subdomain name only.
+// Azure defines the ARecord Name as the subdomain name only.
+// This function logs a message if recordDomain is not a subdomain of zoneName.
 func getARecordName(recordDomain string, zoneName string) (string, error) {
-	return strings.TrimSuffix(strings.TrimSuffix(recordDomain, "."), "."+zoneName), nil
+	trimmedDomain := strings.TrimSuffix(recordDomain, ".")
+	if !strings.HasSuffix(trimmedDomain, "."+zoneName) {
+		log.Info("domain is not a subdomain of zone. The DNS provider may still succeed in updating the record, "+
+			"which might be unexpected", "domain", recordDomain, "zone", zoneName)
+	}
+	return strings.TrimSuffix(trimmedDomain, "."+zoneName), nil
 }


pkg/dns/azure/dns.go
```diff
@@ -132,5 +132,14 @@ func (m *provider) Replace(record *iov1.DNSRecord, zone configv1.DNSZone) error
 // getARecordName extracts the ARecord subdomain name from the full domain string.
 // azure defines the ARecord Name as the subdomain name only.
 func getARecordName(recordDomain string, zoneName string) (string, error) {
-	return strings.TrimSuffix(strings.TrimSuffix(recordDomain, "."), "."+zoneName), nil
+	trimmed := strings.TrimSuffix(strings.TrimSuffix(recordDomain, "."), "."+zoneName)
```
candita, line 135: This looks like it might have been okay for someone to specify a subdomain ending with a trailing ".", and not necessarily ending in the zoneName.  Do we still need to honor that API contract?

sgreene570: I think the original intention of this code was to trim any trailing `.`s off of a full domain name. I don't believe we support creating an ingress controller with `spec.domain` to set to a subdomain.

sgreene570: Note that this PR mirrors the current functionality of the ingress-operator's DNS management on GCP and AWS.

sgreene570:
`27ac428` >>> `3107e22` 
* Added more test cases for `getARecordName` to exercise full-domains ending in `.`

sgreene570:
> /test e2e-azure-operator
> /test e2e-azure

both were green on first try! :grin:

pkg/dns/azure/dns.go
```diff
@@ -132,5 +132,14 @@ func (m *provider) Replace(record *iov1.DNSRecord, zone configv1.DNSZone) error
...
+
+	// strings.TrimSuffix returns the original string if no suffix was trimmed.
+	// Return an error in this case so that bad domains (ie "test.example.com") won't be
+	// published as a sub domain of the given zone (ie "test.example.com.<cluster-domain>").
+	if trimmed == strings.TrimSuffix(recordDomain, ".") {
+		return trimmed, fmt.Errorf("error extracting subdomain from full domain string %s zoneName suffix %s not found", recordDomain, zoneName)
+	}
+
+	return trimmed, nil
```
Miciah, line 144: How about this?
```go
	trimmedDomain := strings.TrimSuffix(recordDomain, ".")
	if !strings.HasSuffix(trimmedDomain, "."+zoneName) {
		return "", fmt.Errorf("domain %q is not a subdomain of zone %q", recordDomain, zoneName)
	}
	return strings.TrimSuffix(trimmedDomain, "."+zoneName), nil
```

pkg/dns/azure/dns.go
Miciah: Theoretically though, could someone have an ingresscontroller with `spec.domain: <domain>` for which the operator publishes a record for `<domain>.<zone>`?  And then, with this change, if the user tried to delete the ingresscontroller, deletion would fail, right?

sgreene570: Ah, yea, I didn't think about the deletion flow. To mitigate this, would it make sense to have an alternative version of `getARecordName` to handle deletions safely?

pkg/dns/azure/dns.go
sgreene570: fixed. Thanks!

pkg/dns/azure/dns.go
sgreene570: I've updated this PR to take deletion safety into account. PTAL @Miciah @candita

sgreene570:
/hold 
We've decided it would make sense to just log a message here instead of blocking users from publishing domains outside of the given zone.

sgreene570:
rebased

/test e2e-azure

sgreene570:
not sure where the retest bot went :thinking: 

/refresh
/retest
/honk

PR #540, opened 2021-01-27
Bug 1917942: canary: Fail on all errors in `Reconcile(...)`

sgreene570:
**pkg/opeator/controller/canary/controller.go:**
This commit forces the canary controller to hard-fail on any errors
encountered during the canary controller's reconcile function since none
of these errors are safe to continue on. This fix is to avoid nil
derefernce panics when the controller fails to ensure a resource.

---

This is a follow-up PR to https://github.com/openshift/cluster-ingress-operator/pull/533. I observed these canary controller panics when scrolling through recent CI failures via https://search.ci.openshift.org/?search=ingress-operator_previous.log.gz&maxAge=336h&context=1&type=junit&name=&maxMatches=5&maxBytes=20971520&groupBy=job .

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -22,7 +22,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


candita:
/lgtm
This looks fine to me, as long as the ordering of the checks is optimal. DS first, followed by service and route.

sgreene570:
> /lgtm
> This looks fine to me, as long as the ordering of the checks is optimal. DS first, followed by service and route.

yup, the service depends on the dameonset, and the route depends on the service.

PR #541, opened 2021-01-27
[release-4.6] Bug 1921193: Fix inconsistent ingress operator status after upgrade

rfredette:
4.6 backport of #526 

`status.DeploymentDegraded` was removed in favor of several more specific status fields in 4.6, but if the cluster was upgraded from 4.5 or earlier, the existing `DeploymentDegraded` status will continue to be reported

If the cluster was upgraded from 4.1 or earlier, `status.EndpointPublishingStrategy.LoadBalancer` (introduced in 4.2?) may not be present when `status.EndpointPublishingStrategy.Type` is set to `LoadBalancerService`. Internally, the assumption is made that this is equivalent to `status.EndpointPublishingStrategy.LoadBalancer` being set to `.LoadBalancer.scope: External`, but this PR makes that implicit assumption explicit.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -228,9 +228,9 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{Requeue: true}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


Miciah:
/lgtm
Thanks!

PR #542, opened 2021-02-02
Bug 1925245: Verify that Serivce resources have idle annotations from corresponding Endpoint resources, should they exist.

sgreene570:
**operator.go: Add endpoint/service idle annotations check**

pkg/operator/operator.go:

Add a new function, `ensureServicesHaveIdledAnnotation(...)` that
mirrors over any idling annotations from endpoint resources when the
idling annotations are not present on the corresponding service. Call
this function during the operator start up logic so that it only runs
once (rather than calling this function within the ingress controller's
reconcile function).

**vendor: Import openshift unidling API**

### DIFF
--- a/cmd/ingress-operator/start.go
+++ b/cmd/ingress-operator/start.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -16,8 +17,14 @@ import (
 	canarycontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/canary"
 	statuscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/status"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	unidlingapi "github.com/openshift/api/unidling/v1alpha1"
 )
 
 const (
@@ -88,6 +95,20 @@ func start(opts *StartOptions) error {
 
 	if opts.ReleaseVersion == statuscontroller.UnknownVersionValue {
 		log.Info("Warning: no release version is specified", "release version", statuscontroller.UnknownVersionValue)
+	}
+
+	// verify that all idled services have the correct idle annotations
+	// mirrored over from the corresponding endpoints resources.
+	// This is to ensure that applications idled with an older version of oc
+	// (and thus do not have the idle annotations on the service) are still
+	// safely un-idleable afer an upgrade that affects `oc idle` functionality.
+	// use a single-use client here separate from the client used by the operator.
+	cl, err := client.New(kubeConfig, client.Options{})
+	if err != nil {
+		return fmt.Errorf("failed to create client from kube config: %v", kubeConfig)
+	}
+	if err := ensureServicesHaveIdleAnnotation(cl); err != nil {
+		log.Error(err, "failed to verify idling endpoints between endpoints and services")
 	}
 
 	// Set up the channels for the watcher, operator, and metrics.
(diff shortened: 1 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -126,6 +129,15 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		}
 	}
 
+	// verify that all idled services have the correct idle annotations
+	// mirrored over from the corresponding endpoints resources.
+	// This is to ensure that applications idled with an older version of oc
+	// (and thus do not have the idle annotations on the service) are still
+	// safely un-idleable afer an upgrade that affects `oc idle` functionality.
+	if err := ensureServicesHaveIdleAnnotation(mgr.GetClient()); err != nil {
```
Miciah, line 137: Using `mgr.GetClient()` before the manager `Start` has been called (and thus caches synched) makes me a little nervous...  But I think this is all right because we explicitly specified `NewClient` not to use a caching client in the manager options.

pkg/operator/operator.go
```diff
@@ -198,3 +210,52 @@ func (o *Operator) ensureDefaultIngressController() error {
...
 }
+
+func ensureServicesHaveIdleAnnotation(cl client.Client) error {
+	endpointsList := &corev1.EndpointsList{}
+	err := cl.List(context.TODO(), endpointsList, &client.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list endpoints in all namespaces: %v", err)
+	}
+
+	for _, endpoints := range endpointsList.Items {
```
Miciah, line 221: If possible, I'd like to get data on how this affects start time on a cluster with, say, a hundred thousand endpoints.  This might best be done asynchronously (i.e., in a goroutine), even if we leave it it as a one-off for the time being.

sgreene570: Andy and I thought it would make the most sense for the operator to ensure that the idle annotations exist on services before creating the router deployment, since update the service annotations would not guarantee a router reload necessarily. Doing this asynchronously could make the behavior racey, so im not sure the best path forward. Thoughts?

pkg/operator/operator.go
sgreene570: > But I think this is all right because we explicitly specified `NewClient` not to use a caching client in the manager options.

Is there an alternative code path you'd like to see instead? I agree that since we are using a cacheless client anyways this is probably fine as is.

pkg/operator/operator.go
frobware: I share this concern. My concern is subsequently getting the idle annotations on a service but no corresponding endpoint changes. It's only in `createRouterEndpoints()` that we check to for _idleness_ and we only get there if there are endpoint additions/deletions. It seems that we could get out of sync if done asynchronously.

pkg/operator/operator.go
Miciah: Use a separate client (you can use either the controller-runtime or client-go package to get the client, but it should be a separately instantiated client).  

I'd go so far as to suggesting divorcing this migration logic from the controller-runtime logic.  Put it in `start.go`, or even put it in a separate subcommand, invoked using an init container.

pkg/operator/operator.go
Miciah: Keep in mind that this logic is only needed during upgrades, in which case the router deployment already exists.  But I think your reasoning holds if you s/creating the router deployment/updating the router deployment to use the image pullspec in the release payload/.  Based on that reasoning, doing this migration synchronously makes sense.  However, this puts the migration in the upgrade path, and I wonder (a) how much it increases time to upgrade and (b) how long the operator will be out of service and not reconciling its operands.

sgreene570: @Miciah I just ran a rough timing estimate here and on a 4.7 cluster with ~25k endpoints, and 10 idled services, the new idle annotation verification function took about 2 seconds to run. Is that a concerning amount of time in your opinion?

pkg/operator/operator.go
sgreene570: Alright, Ill move this logic to start.go and create a separate client there.

cmd/ingress-operator/start.go
```diff
@@ -168,3 +189,52 @@ func start(opts *StartOptions) error {
...
+		service := &corev1.Service{}
+		serviceName := types.NamespacedName{
+			Name:      endpoints.Name,
+			Namespace: endpoints.Namespace,
+		}
+		if err := cl.Get(context.TODO(), serviceName, service); err != nil {
+			return fmt.Errorf("failed to get service for endpoints %s/%s", service.Namespace, service.Name)
+		}
+
+		_, haveIdledAt = service.Annotations[unidlingapi.IdledAtAnnotation]
```
frobware, line 216: `service.Annotations` can be nil here?

pkg/operator/operator.go
sgreene570: Note that my tests were performed by running the operator locally, so in-cluster times would be probably be fast given lower API request latency.

Miciah: If it scales linearly (i.e., a cluster with 100k endpoints would take 8 seconds, and 200k endpoints would take 16 seconds), that doesn't seem too onerous.

sgreene570: A similar test with 50k endpoints took ~3.8 seconds, so I think we have enough evidence to suggest it scales linearly.

sgreene570:
`f2c5a63` to `716cb79`
 
 Modified `ensureServicesHaveIdledAnnotation(...)` so that any errors syncing endpoint + service annotations do not cause the operator to crash-loop.

frobware:
> `f2c5a63` to `716cb79`
> 
> Modified `ensureServicesHaveIdledAnnotation(...)` so that any errors syncing endpoint + service annotations do not cause the operator to crash-loop.

/lgtm

PR #543, opened 2021-02-05
Bug 1907315: Change AWS interal LB annotation to match documentation

rfredette:
Annotation `service.beta.kubernetes.io/aws-load-balancer-internal` is
currently set as `"0.0.0.0/0"` in the operator-created load balancer
service when using an internal load balancer on AWS. The prefered value
for this annotation is now `"true"`[1][2], although both values seem to be
valid.

[1] https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
[2] https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -97,6 +97,10 @@ var (
 	//
 	// https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
 	InternalLBAnnotations = map[configv1.PlatformType]map[string]string{
+		// Prior to 4.8, the aws internal LB annotation was set to "0.0.0.0/0".
+		// While "0.0.0.0/0" is valid, the preferred value, according to the
+		// documentation[1], is "true".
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -96,7 +96,7 @@ var (
 	// https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
 	InternalLBAnnotations = map[configv1.PlatformType]map[string]string{
 		configv1.AWSPlatformType: {
-			awsInternalLBAnnotation: "0.0.0.0/0",
```
sgreene570, line 99: Would it be worth dropping a comment here explaining that we used to set the annotation to `0.0.0.0/0` in prior OCP versions?

rfredette: Good idea. I've updated the PR with an explanation comment.

Miciah:
Thanks!
/lgtm

PR #549, opened 2021-02-11
Bug 1914127: Delete the ingress.openshift.io/operator finalizer

Miciah:
Add logic to remove the operator's finalizer on LoadBalancer services, which is no longer needed to clean up DNS records since the DNSRecord CRD was added.

* `pkg/operator/controller/ingress/load_balancer_service.go` (`ensureLoadBalancerService`): Use the new `deleteLoadBalancerService` method to delete the service as needed.  Use the new `deleteLoadBalancerServiceFinalizer` method to delete any finalizer on any existing service in the delete and update cases.
(`desiredLoadBalancerService`): Delete logic to add a finalizer to the service.
(`deleteLoadBalancerServiceFinalizer`): New method.  Delete any finalizer from the service.
(`finalizeLoadBalancerService`): Refactor to use the new `deleteLoadBalancerServiceFinalizer` method.
(`deleteLoadBalancerService`): New method.

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -16,6 +16,8 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -292,6 +324,18 @@ func (r *reconciler) createLoadBalancerService(service *corev1.Service) error {
 	return nil
 }
 
+// deleteLoadBalancerService deletes a load balancer service.
+func (r *reconciler) deleteLoadBalancerService(service *corev1.Service, options *crclient.DeleteOptions) error {
+	if err := r.client.Delete(context.TODO(), service, options); err != nil {
```
sgreene570, line 329: Is there a foreseeable reason to include an `*crclient.DeleteOptions` parameter? I'm not opposed to adding one here, but I recognize that the other `r.client.Delete` calls in this repo exclude this paramter.

sgreene570 (commented):
Overall LGTM, I just have one small question.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Good question, and I had to go back to see why I did this.  Yes, we will want to specify `PropagationPolicy: "Foreground"` when deleting and recreating the service if/when we re-introduce mutable scope.

sgreene570: aha! That makes sense. Looking forward to taking advantage of that :wink:

PR #552, opened 2021-02-16
Introduce ingress controller

awgreene:


pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	// Generate namespaceName mapping
+	mapToSecretName := map[types.NamespacedName]string{}
+	for _, componentRoute := range ingress.Spec.ComponentRoutes {
+		mapToSecretName[types.NamespacedName{Namespace: componentRoute.Namespace, Name: componentRoute.Name}] = componentRoute.ServingCertKeyPairSecret.Name
+	}
+
+	// For each Status.ComponentRoute resource, if a Spec.ComponentRoute with a matching
+	// namespace.name touple exists create the role and roleBinding for each consumer users.
+	// If no matching Spec.ComponentRoute exists, make sure any RBAC that was generated is removed.
+	for _, componentRoute := range ingress.Status.ComponentRoutes {
```
deads2k, line 96: you need to go in the other direction.  In order to be able to observe a status and know a thing needs to be removed, you need to find all the roles and rolebindings you've created and see which ones don't belong.  A label seesm appropriate.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	if err := r.client.Get(context.TODO(), request.NamespacedName, ingress); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("ingress cr not found; reconciliation will be skipped", "request", request)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get ingress %q: %v", request, err)
+	}
+
+	// Generate namespaceName mapping
+	mapToSecretName := map[types.NamespacedName]string{}
```
deads2k, line 88: I know it's a map.  Tell me what the key is: `keyToValue`

deads2k, line 88: don't reuse types.NamespacedName.  Create a local struct type for this key.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (r *reconciler) ensureServiceCertKeyPairSecretRole(owner metav1.Object, crName, secretName string) error {
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			// TODO: Generate Name and add labels to avoid componentRoute.Name collisions.
+			Name:      crName,
+			Namespace: r.config.SecretNamespace,
+			OwnerReferences: []metav1.OwnerReference{
```
deads2k, line 129: No.  I don't see a need or desire to tie this to a cluster resource that can never be deleted.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+					APIVersion: "config.openshift.io/v1",
+					Kind:       "Ingress",
+					Name:       owner.GetName(),
+					UID:        owner.GetUID(),
+				},
+			},
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				Verbs:         []string{"get", "list"},
```
deads2k, line 140: add watch

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+			return reconcile.Result{Requeue: true}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *reconciler) ensureServiceCertKeyPairSecretRole(owner metav1.Object, crName, secretName string) error {
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			// TODO: Generate Name and add labels to avoid componentRoute.Name collisions.
```
deads2k, line 126: not a TODO.  You need to do this to start

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+				Verbs:         []string{"get", "list"},
+				APIGroups:     []string{""},
+				Resources:     []string{"secrets"},
+				ResourceNames: []string{secretName},
+			},
+		},
+	}
+
+	// Attempt to create or update the roleBinding
+	if err := r.client.Create(context.TODO(), role); err != nil {
```
deads2k, line 149: pretty sure we have a library-go apply for this that handles a diff/create/update

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return nil
+}
+
+func (r *reconciler) ensureServiceCertKeyPairSecretRoleBinding(owner metav1.Object, roleName string, users []string) error {
+	subjects := []rbacv1.Subject{}
+	for _, user := range users {
+		subjects = append(subjects, rbacv1.Subject{
+			Kind:     rbacv1.ServiceAccountKind, // TODO: Should we support non service accounts?
+			Name:     user,
```
deads2k, line 175: this looks wrong for a serviceaccount.  find an example.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	return nil
+}
+
+func (r *reconciler) ensureServiceCertKeyPairSecretRoleBinding(owner metav1.Object, roleName string, users []string) error {
+	subjects := []rbacv1.Subject{}
+	for _, user := range users {
+		subjects = append(subjects, rbacv1.Subject{
+			Kind:     rbacv1.ServiceAccountKind, // TODO: Should we support non service accounts?
```
deads2k, line 174: hmmm.... I can't think of a reason right off, but probably.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+			Name:     user,
+			APIGroup: "",
+		})
+	}
+
+	roleBinding := &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      roleName,
+			Namespace: r.config.SecretNamespace,
+			OwnerReferences: []metav1.OwnerReference{
```
deads2k, line 184: no ownerref

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+		Subjects: subjects,
+		RoleRef: rbacv1.RoleRef{
+			Kind:     "Role",
+			Name:     roleName,
+			APIGroup: "rbac.authorization.k8s.io",
+		},
+	}
+
+	// Attempt to create or update the roleBinding
+	if err := r.client.Create(context.TODO(), roleBinding); err != nil {
```
deads2k, line 202: same library-go comment

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+// reconciler handles the actual ingress reconciliation logic in response to
+// events.
+type reconciler struct {
+	config Config
+	client client.Client
+}
+
+// Reconcile expects request to refer to a ingress in the operator namespace,
+// and will do all the work to ensure the ingress is in the desired state.
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
deads2k, line 74: how do errors here show up as degraded in the operator.  I think this and the lbirary-go apply are the most critical to work out ot decide if it can go here or needs to go elsewhere.

pkg/operator/controller/ingress-routes/controller.go
```diff
@@ -0,0 +1,250 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	controllerName = "ingress_routes_controller"
```
sgreene570, line 26: Would you be opposed to using a different controller name? Maybe `operator_route_rbac_controller`, or something along those lines? I think `ingress_routes_controller` is a little vague and somewhat similar to the existing `ingress-to-route` controller in OCM.

pkg/operator/operator.go
```diff
@@ -89,6 +90,13 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		return nil, fmt.Errorf("failed to create ingress controller: %v", err)
 	}
 
+	// Create and register the ingress routes controller with the operator manager.
+	if _, err := ingressroutescontroller.New(mgr, ingressroutescontroller.Config{
+		SecretNamespace: "openshift-config",
```
sgreene570, line 95: the `openshift-config` namespace is already defined in `pkg/operator/controller/names.go` as `GlobalMachineSpecifiedConfigNamespace`.

awgreene: [It looks like](https://github.com/awgreene/cluster-ingress-operator/blob/master/pkg/operator/controller/names.go#L16) the `GlobalMachineSpecifiedConfigNamespace` is actually `openshift-config-managed`.

pkg/operator/controller/ingress-routes/controller.go
awgreene: That sounds good to me, I'll include the change in the next update to the PR.

pkg/operator/operator.go
awgreene: I can add a new constant to the `names.go` file though.

sgreene570: Ah, my mistake (so close)! If you could add a new constant that would be great! thanks.

pkg/util/hash.go
```diff
@@ -0,0 +1,14 @@
+package k8s
+
+import (
+	"fmt"
+	"hash/fnv"
+
+	"k8s.io/apimachinery/pkg/util/rand"
+)
+
+func Hash(s string) string {
```
sgreene570, line 10: IMO, based on its usage, it might make more sense for this hash util function to take a namespaced name argument (rather than a string).
Thoughts?

test/e2e/ingress_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+func pollForValidComponentRouteRole(t *testing.T, componentRoute configv1.ComponentRouteSpec) error {
+	listOptions := []client.ListOption{
+		client.MatchingLabels{
+			componentRouteHashLabelKey: util.Hash(fmt.Sprintf("%s/%s", componentRoute.Namespace, componentRoute.Name)),
+		},
+	}
+
+	roleList := &rbacv1.RoleList{}
+	err := wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
sgreene570, line 209: Would you say that 10 seconds is a reasonable timeout here? Wondering if this timeout (or some of the others in this test file) might be too aggressive.

awgreene: We can absolutely relax the timeout constraints (say to a minute) but I would be concerned if the controller ever takes more than a second or so to reconcile an event.

sgreene570: I guess we can just keep on an eye on e2e operator test runs here?

sgreene570: Latest run (https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/552/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1366461389803098112) shows some interesting timeouts that sparked my asking.

pkg/operator/controller/ingress-routes/controller.go
sgreene570: Could we change the `controller/` sub dir for this controller as well? ie
`pkg/operator/controller/operator-route-rbac-controller`?

pkg/operator/controller/ingress-routes/controller.go
awgreene: The initial implementation will only support service accounts.

pkg/operator/controller/ingress-routes/controller.go
awgreene: I'll make sure to update the directory name.

test/e2e/ingress_test.go
awgreene: I'll update the code to wait for a minute.

jhadvig (commented):
@awgreene the change looks good. I was in an impression that the ingress-operator would also do a basic validation of the hostname and secret so the consuming components can just consume the API and check for the changes in the ComponentRoutesStatus.Conditions, eg:
- if the default and custom hostname are not the same
- if the custom hostname has cluster domain suffix, which indicates, if a secret that contains TLS certificate and key needs to exist in the:
    - If the suffix matches the cluster domain, then the secret is optional
    - If the suffix doesn't matches the cluster domain, then the secret is mandatory
- validate that the secret contains `tls.crt` and `tls.key` data keys where the certificate and key are stored and both need to be in valid format

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,376 @@
...
+
+		// Generate roleBinding.
+		if err := r.ensureServiceCertKeyPairSecretRoleBinding(ingress, componentRoute, roleName); err != nil {
+			return reconcile.Result{Requeue: true}, fmt.Errorf("failed to create rolebinding: %v", err)
+		}
+	}
+
+	// Delete any roles or roleBindings that were generated for componentRoutes that are no longer defined.
+	if err := r.cleanupOrphanedResources(componentRoutes); err != nil {
+		return reconcile.Result{Requeue: true}, fmt.Errorf("failed to delete role: %v", err)
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest commit.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,376 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	controllerName             = "configurable_route_controller"
+	componentRouteHashLabelKey = "ingress.operator.openshift.io/componentroutehash"
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest commit.

pkg/operator/controller/names.go
```diff
@@ -15,6 +15,12 @@ const (
 	// CA certificate in this namespace.
 	GlobalMachineSpecifiedConfigNamespace = "openshift-config-managed"
 
+	//CluseterScopedNamespace is an empty string representing a cluster scoped namespace.
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest commit.

manifests/01-role-binding.yaml
```diff
@@ -16,3 +16,23 @@ roleRef:
   kind: Role
   apiGroup: rbac.authorization.k8s.io
   name: ingress-operator
+---
+# RoleBinding for the operator to delete Role and RoleBindings
+# in the openshift-config namespace.
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ingress-operator
```
sttts, line 25: is this an established naming scheme for rolebindings?

awgreene: I have seen a number of operators simply name their role/rolebinding after the opreator. In this case, the operators seems to follow [this convention](https://github.com/openshift/cluster-ingress-operator/blob/a5525f387106974730ccb9f422e05115498cbc47/manifests/01-role-binding.yaml#L5).

manifests/01-role.yaml
```diff
@@ -36,3 +36,23 @@ rules:
...
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+rules:
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  verbs:
+  - delete
```
sttts, line 58: no creation?

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
sttts, line 22: order

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	componentRouteHashAnnotation = "ingress.operator.openshift.io/componentroutehash"
+)
+
+var (
+	log = logf.Logger.WithName(ControllerName)
+)
+
+// New creates the configurable route controller from configuration. This is the controller
+// that handles all the logic for generating roles and rolebindings for operators that
+// include routes with configurable hostnames and serving certificates.
```
sttts, line 45: would even like to see a few lines more here

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+// Reconcile expects request to refer to a ingress in the operator namespace,
+// and will do all the work to ensure the ingress is in the desired state.
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
+
+	// Only proceed if we can get the ingress resource.
+	ingress := &configv1.Ingress{}
+	if err := r.cache.Get(context.TODO(), request.NamespacedName, ingress); err != nil {
```
sttts, line 126: why do we care about ingresses? Aren't the route names in the config virtual, i.e. without 1:1 match with real routes?

manifests/01-role.yaml
awgreene: The operator already has a clusterRole granting [create, get, list, watch, and update permissions](https://github.com/openshift/cluster-ingress-operator/blob/a5525f387106974730ccb9f422e05115498cbc47/manifests/00-cluster-role.yaml#L51-L63) in all namespaces but I did not want to extend it the ability to delete roles/rolebindings in all namespaces.

pkg/operator/controller/configurable-route/controller.go
awgreene: The `ingresses.config.openshift.io/cluster` resource icontains the route configuration information. The CRD was updated in [this PR](https://github.com/openshift/api/pull/852).

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func (r *reconciler) ensureServiceCertKeyPairSecretRole(owner metav1.Object, componentRoute aggregatedComponentRoute) (string, error) {
+	role := &rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: componentRoute.Name + "-",
+			Namespace:    r.config.SecretNamespace,
+			Labels: map[string]string{
+				componentRouteHashAnnotation: componentRoute.Hash,
+			},
```
sttts, line 323: how about an owner reference to the operator?

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+			return "", err
+		}
+		return role.GetName(), nil
+	}
+
+	for i, curRole := range roleList.Items {
+		if i == 0 {
+			continue
+		}
+		if err := r.client.Delete(context.TODO(), &curRole); err != nil && !errors.IsNotFound(err) {
```
sttts, line 351: strange to see `ApplyRole` and this delete before. A create should be enough, no?

pkg/operator/controller/names.go
```diff
@@ -15,6 +15,12 @@ const (
 	// CA certificate in this namespace.
 	GlobalMachineSpecifiedConfigNamespace = "openshift-config-managed"
 
+	// ClusterScopedNamespace is an empty string representing a cluster scoped namespace.
+	ClusterScopedNamespace = ""
+
+	// OpenShiftConfigNamespace is the namespace for configuring OpenShift.
+	OpenShiftConfigNamespace = "openshift-config"
```
sttts, line 22: over-abstraction. This const just makes it harder to read.

pkg/operator/controller/names.go
```diff
@@ -15,6 +15,12 @@ const (
 	// CA certificate in this namespace.
 	GlobalMachineSpecifiedConfigNamespace = "openshift-config-managed"
 
+	// ClusterScopedNamespace is an empty string representing a cluster scoped namespace.
+	ClusterScopedNamespace = ""
```
sttts, line 19: this just makes it harder to read.

pkg/operator/controller/names.go
```diff
@@ -46,6 +52,14 @@ func IngressClusterOperatorName() types.NamespacedName {
 	}
 }
 
+// IngressClusterConfigName returns the namespaced name of the ingress.config.openshift.io
+// resource for the operator.
+func IngressClusterConfigName() types.NamespacedName {
```
sttts, line 57: over-abstraction

pkg/operator/operator.go
```diff
@@ -21,6 +22,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 
 	"k8s.io/client-go/rest"
+	"github.com/openshift/library-go/pkg/operator/events"
```
sttts, line 25: order

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	r.cache.List(context.TODO(), roleBindingList, allComponentRouteResources()...)
+	for _, item := range roleBindingList.Items {
+		expectedHash, ok := item.GetLabels()[componentRouteHashAnnotation]
+		if !ok {
+			return fmt.Errorf("Unable to find expected componentRoute hash label")
+		}
+
+		if _, ok := existingHashes[expectedHash]; !ok {
+			log.Info("deleting roleBinding", "name", item.GetName(), "namespace", item.GetNamespace())
+			if err := r.client.Delete(context.TODO(), &item); err != nil && !errors.IsNotFound(err) {
```
sttts, line 307: are we sure here this does not mutate the cache, i.e. when item has some pointer substruct. Won't they be reused by the delete client?

manifests/01-role.yaml
sttts: Would add delete there too. The operator is already basically cluster admin with what it can do (assigning itself more roles). Adding delete is not making it worse.

Wondering for what does it use the rbac permissions today?

pkg/operator/controller/names.go
awgreene: I was asked by the networking-team to include the constants [here](https://github.com/openshift/cluster-ingress-operator/pull/552#discussion_r578565736).

pkg/operator/controller/names.go
awgreene: I'm hesitant to break from the pattern established by the authors of the operator. @sgreene570 could you weigh in here?

pkg/operator/controller/names.go
awgreene: @sgreene570 what are your thoughts?

pkg/operator/controller/names.go
awgreene: This one I agree with.

pkg/operator/controller/ingress-routes/controller.go
awgreene: Added in the latest version of the PR.

pkg/operator/controller/ingress-routes/controller.go
awgreene: Addressed in the latest version of the PR.

test/e2e/ingress_test.go
awgreene: Also - the test cases are failing because the [PR to update the ingresses.config.openshift.io CRD](https://github.com/openshift/cluster-config-operator/pull/187) has not merged, so the CRD's status does not include any component routes.

manifests/01-role.yaml
awgreene: I agree with your assessment that the operator is effectively a cluster admin and that granting it delete permissions at the cluster level doesn't make anything worse.

However, I believe that by scoping these permissions to the `openshift-config` the purpose of rbac is better communicated to developers unfamiliar with this feature whereas the purpose is somewhat mystified if lumped into the clusterRole. If you do not believe that this information is valuable to other developers I can lump it into the clusterRole.

pkg/operator/controller/configurable-route/controller.go
awgreene: Updated in the latest version of the PR.

pkg/operator/controller/names.go
awgreene: Addressed in the latest version of the PR.

pkg/operator/operator.go
awgreene: Addressed in the latest version of the PR.

pkg/operator/controller/names.go
sgreene570: If you feel strongly about moving or removing the const I won't object. Just made sense to me to include the openshift config namespace as a const with all of the other namespace consts the operator may need.

pkg/operator/controller/names.go
sgreene570: I would prefer to keep things consistent with the existing operator code but if you feel strongly about moving / modifying / removing this function I won't object.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+		if len(roleBinding.Subjects) != len(componentRoute.ConsumingUsers) {
+			return false, nil
+		}
+
+		for _, expectedSubject := range componentRoute.ConsumingUsers {
+			found := false
+			splitExpectedSubjects := strings.Split(string(expectedSubject), ":")
+			for _, subject := range roleBinding.Subjects {
+				if subject.Namespace == splitExpectedSubjects[2] {
```
Miciah, line 271: Why check namespace and not name?  The test inputs refer to two component routes that have the same namespace, so it seems like this could lead to false positives.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+	roleBindingList := &rbacv1.RoleBindingList{}
+	if err := r.cache.List(context.TODO(), roleBindingList, componentRouteResources(componentRoute)...); err != nil {
+		return err
+	}
+
+	if len(roleBindingList.Items) == 1 {
+		roleBinding.Name = roleBindingList.Items[0].Name
+	}
+
+	_, _, err := resourceapply.ApplyRoleBinding(r.kclient.RbacV1(), r.eventRecorder, roleBinding)
```
Miciah, line 391: This logic seems a little wonky.  Presumably, `roleBinding.Name` and `roleBindingList.Items[0].Name` could have differed before this assignment, or else you wouldn't have written it, right?  Would it be simpler always to use the name of the just-created role (so the role name/reference and rolebinding name would always match)?  

Is this logic's purpose to solve the following problem?

1. User creates a route configuration.
2. Controller creates role and rolebinding.
3. User deletes route configuration.
4. Controller deletes the role and fails to delete the rolebinding.
5. User creates the same route configuration again.
6. Controller creates a new role, and you want it to re-use the old rolebinding instead of creating a new one.

This highlights that the life cycles of these resources are potentially nuanced and confusing.  A comment would be helpful.  Maybe it would be better still to create the rolebinding with an owner reference to the role, and let the garbage collector sort it out; that would simplify this logic, and obviate the rolebinding cleanup logic as well.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+	} else {
+		role.Name = roleList.Items[0].Name
+		role.GenerateName = ""
+		if _, _, err := resourceapply.ApplyRole(r.kclient.RbacV1(), r.eventRecorder, role); err != nil {
+			return "", err
+		}
+	}
+
+	return role.GetName(), nil
```
Miciah, line 362: Do you need to do another get to get the generated name?

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		ServingCertificateName: spec.ServingCertKeyPairSecret.Name,
+		ConsumingUsers:         status.ConsumingUsers,
+	}
+}
+
+// aggregatedComponeRoute contains information from the ComponentRouteSpec
+// and ComponentRouteStatus to generate the required Role and RoleBinding.
+type aggregatedComponentRoute struct {
+	Name                   string
+	Hash                   string
```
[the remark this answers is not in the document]
awgreene: Given the maximum length of namespaces and names and label length [requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) (see below), it is pretty easy to hit an unrecoverable edge case where the label value is too long.

Relevant section from k8s docs:
>Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+	for _, cr := range componentRoutes {
+		existingHashes[cr.Hash] = struct{}{}
+	}
+
+	roleList := &rbacv1.RoleList{}
+	r.cache.List(context.TODO(), roleList, allComponentRouteResources()...)
+	for _, item := range roleList.Items {
+		expectedHash, ok := item.GetLabels()[componentRouteHashAnnotation]
+		if !ok {
+			errors = append(errors, fmt.Errorf("Unable to find componentRoute hash label on roleBinding %s/%s", item.Namespace, item.Name))
```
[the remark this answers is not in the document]
awgreene: This was unintended, I will use `item.GetName{space}()` everywhere.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+	defer func() {
+		if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: "", Name: "cluster"}, ingress); err != nil {
+			t.Fatalf("failed to create ingress resource: %v", err)
+		}
+		ingress.Spec.ComponentRoutes = nil
+		if err := kclient.Update(context.TODO(), ingress); err != nil {
+			t.Errorf("failed to restore cluster ingress resource to original state: %v", err)
+		}
+		ingress.Status.ComponentRoutes = nil
+		if err := kclient.Status().Update(context.TODO(), ingress); err != nil {
```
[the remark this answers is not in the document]
awgreene: It does work now because the same ingress resource is used in each update call which modifies the resource version.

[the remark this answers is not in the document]
awgreene: I'll fix this in the next update to the PR.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+	// Make sure that the bar role and roleBinding are deleted
+	roleList := &rbacv1.RoleList{}
+	listOptions := []client.ListOption{
+		client.MatchingLabels{
+			componentRouteHashLabelKey: util.Hash("default/bar"),
+		},
+	}
+	if err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+		if err := kclient.List(context.TODO(), roleList, listOptions...); err != nil {
+			return false, err
```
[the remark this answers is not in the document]
awgreene: I understand why one would do this with a Get call but a list call that includes no resources should return an empty list rather than an error.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+		if len(roleBindingList.Items) != 0 {
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Errorf("role not deleted: %v", err)
+	}
+}
+
+func pollForValidComponentRouteRole(t *testing.T, componentRoute configv1.ComponentRouteSpec) error {
```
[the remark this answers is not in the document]
awgreene: Yes - my mistake. Thank you for catching this.

test/e2e/configurable_route_test.go
awgreene: Agreed, will add a check against the name.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+
+	// Trigger reconcile requests for the roles and roleBindings with the componentRoute label.
+	defaultPredicate := predicate.NewPredicateFuncs(func(meta metav1.Object, object runtime.Object) bool {
+		labels := meta.GetLabels()
+		_, ok := labels[componentRouteHashAnnotation]
+		return ok
+	})
+
+	roleInformer, err := mgr.GetCache().GetInformer(context.TODO(), &rbacv1.Role{})
+	if err := c.Watch(&source.Informer{Informer: roleInformer}, &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(reconciler.resourceToClusterIngressConfig)}, defaultPredicate); err != nil {
```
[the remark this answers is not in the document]
awgreene: We can use source.Kind, it removes the informer error checks as well.

pkg/operator/controller/configurable-route/controller.go
awgreene: Nope, the name is updated in the create call. I do however need to do a call later to get the UID of the role so it can be used on the roleBinding's ownerReference.

pkg/operator/controller/configurable-route/controller.go
awgreene: >Maybe it would be better still to create the rolebinding with an owner reference to the role, and let the garbage collector sort it out; that would simplify this logic, and obviate the rolebinding cleanup logic as well.

I agree with this approach and have implemented it in the latest version of the PR.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+		if len(roleBindingList.Items) != 1 {
+			return false, nil
+		}
+
+		roleBinding := roleBindingList.Items[0]
+		if roleBinding.RoleRef.APIGroup != "rbac.authorization.k8s.io" ||
+			roleBinding.RoleRef.Kind != "Role" ||
+			roleBinding.RoleRef.Name != roleBinding.GetName() {
+			return false, nil
```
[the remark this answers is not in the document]
awgreene: Updated the e2e test to address these concerns.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+	return reconcile.Result{}, nil
+}
+
+// newAggregatedComponentRoute returns an aggregatedComponentRoute.
+func newAggregatedComponentRoute(spec configv1.ComponentRouteSpec, status configv1.ComponentRouteStatus) aggregatedComponentRoute {
+	return aggregatedComponentRoute{
+		Name:                   spec.Name,
+		Hash:                   util.Hash(namespacedName(spec.Namespace, spec.Name)),
+		ServingCertificateName: spec.ServingCertKeyPairSecret.Name,
+		ConsumingUsers:         status.ConsumingUsers,
```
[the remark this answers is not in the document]
awgreene: I've the latest version of the PR copies the array, please let me know if that addresses all your concerns.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get ingress %q: %v", request, err)
+	}
+
+	// Get the list of componentRoutes defined in both the spec and status of the ingress resource
+	componentRoutes := r.intersectingComponentRoutes(ingress.Spec.ComponentRoutes, ingress.Status.ComponentRoutes)
+
+	// Ensure role and roleBindings exist for each valid componentRoute.
+	for _, componentRoute := range componentRoutes {
+		// Do not generate unnecessary roles or rolebindings if not required.
```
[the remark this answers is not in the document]
awgreene: Let's remove the comment.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+// and ComponentRouteStatus to generate the required Role and RoleBinding.
+type aggregatedComponentRoute struct {
+	Name                   string
+	Hash                   string
+	ServingCertificateName string
+	ConsumingUsers         []configv1.ConsumingUser
+}
+
+// getSubjects returns a list of subjects defined in the aggregatedComponentRoute.
+func (componentRoute *aggregatedComponentRoute) GetSubjects() []rbacv1.Subject {
```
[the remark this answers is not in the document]
awgreene: Agreed.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,393 @@
...
+	return []client.ListOption{
+		client.HasLabels{componentRouteHashAnnotation},
+		client.InNamespace(operatorcontroller.OpenShiftConfigNamespace),
+	}
+}
+
+func (r *reconciler) deleteOrphanedRoles(componentRoutes []aggregatedComponentRoute) []error {
+	errors := []error{}
+
+	existingHashes := map[string]struct{}{}
```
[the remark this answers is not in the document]
awgreene: Agreed, though the latest version of the PR leaves the cleanup of roleBindings to GC.

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,286 @@
...
+	}); err != nil {
+		t.Errorf("Failed to update ingress resource: %v", err)
+	}
+
+	// Make sure that the bar role and roleBinding are deleted
+	roleList := &rbacv1.RoleList{}
+	listOptions := []client.ListOption{
+		client.MatchingLabels{
+			componentRouteHashLabelKey: util.Hash("default/bar"),
+		},
```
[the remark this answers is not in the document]
awgreene: Added it.

awgreene:
@Miciah I've addressed or resolved all of your comments and have updated the PR with your suggestions.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,376 @@
...
+	return []client.ListOption{
+		client.HasLabels{componentRouteHashLabelKey},
+		client.InNamespace(operatorcontroller.GlobalUserSpecifiedConfigNamespace),
+	}
+}
+
+func (r *reconciler) deleteOrphanedRoles(componentRoutes []aggregatedComponentRoute, existingHashes sets.String) []error {
+	errors := []error{}
+	roleList := &rbacv1.RoleList{}
+	r.cache.List(context.TODO(), roleList, allComponentRouteResources()...)
```
Miciah, line 288: Check the error value.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -0,0 +1,376 @@
...
+	for _, item := range roleList.Items {
+		expectedHash, ok := item.GetLabels()[componentRouteHashLabelKey]
+		if !ok {
+			errors = append(errors, fmt.Errorf("Unable to find componentRoute hash label on role %s/%s", item.GetNamespace(), item.GetName()))
+			continue
+		}
+
+		if _, ok := existingHashes[expectedHash]; !ok {
```
Miciah, line 296:
```suggestion
		if if !existingHashes.Has(expectedHash) {
```

test/e2e/configurable_route_test.go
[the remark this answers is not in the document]
Miciah: We want to prevent transient errors from causing additional test flakiness.  It isn't the job of these tests to tell us if the API or network is flaky; just log the error so we have some clues if the test ultimately fails due to client calls failing, and retry. 
```suggestion
			return false, nil
```

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,531 @@
...
+	})
+}
+
+func eventuallyUpdateIngressStatus(t *testing.T, ingressStatus configv1.IngressStatus) error {
+	t.Helper()
+	ingress := &configv1.Ingress{}
+	return wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: "", Name: "cluster"}, ingress); err != nil {
+			t.Logf("error getting ingress: %v", err)
+			return false, err
```
Miciah, line 406:
```suggestion
			t.Logf("error getting ingress: %v", err)
			return false, nil
```

test/e2e/configurable_route_test.go
```diff
@@ -0,0 +1,531 @@
...
+			t.Logf("unexpected number of roles exist: expected (1), actual (%v)", len(roleList.Items))
+			return false, nil
+		}
+
+		role := roleList.Items[0]
+		if len(role.Rules) != 1 ||
+			len(role.Rules[0].Verbs) != 3 && role.Rules[0].Verbs[0] != "get" && role.Rules[0].Verbs[1] != "list" && role.Rules[0].Verbs[2] != "watch" ||
+			len(role.Rules[0].APIGroups) != 1 && role.Rules[0].APIGroups[0] != "" ||
+			len(role.Rules[0].Resources) != 1 && role.Rules[0].Resources[0] != "secrets" ||
+			len(role.Rules[0].ResourceNames) != 1 && role.Rules[0].ResourceNames[0] != componentRoute.ServingCertKeyPairSecret.Name {
```
Miciah, line 468:
```suggestion
			len(role.Rules[0].Verbs) != 3 || role.Rules[0].Verbs[0] != "get" || role.Rules[0].Verbs[1] != "list" || role.Rules[0].Verbs[2] != "watch" ||
			len(role.Rules[0].APIGroups) != 1 || role.Rules[0].APIGroups[0] != "" ||
			len(role.Rules[0].Resources) != 1 || role.Rules[0].Resources[0] != "secrets" ||
			len(role.Rules[0].ResourceNames) != 1 || role.Rules[0].ResourceNames[0] != componentRoute.ServingCertKeyPairSecret.Name {
```

pkg/operator/controller/configurable-route/controller.go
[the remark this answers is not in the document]
Miciah: Yeah, looks good.  Thanks!

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -227,21 +238,25 @@ func (componentRoute *aggregatedComponentRoute) requiresRBAC() bool {
 	return true
 }
 
-// intersectingComponentRoutes take a list of componentRouteSpecs and componentRouteStatuses and
-// identifies which namespace,name tuples appear in both lists, and returns an aggregatedComponentRoute
-// for each.
-func (r *reconciler) intersectingComponentRoutes(componentRouteSpecs []configv1.ComponentRouteSpec, componentRouteStatuses []configv1.ComponentRouteStatus) []aggregatedComponentRoute {
+// intersectingComponentRoutes takse a slice of componentRouteSpec and a slice
```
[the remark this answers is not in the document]
awgreene: I should have caught that, my mistake.

pkg/operator/controller/configurable-route/controller.go
```diff
@@ -132,50 +135,58 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		roleName, err := r.ensureServiceCertKeyPairSecretRole(ingress, componentRoute)
+		roleName, err := r.ensureServiceCertKeyPairSecretRole(componentRoute)
 		if err != nil {
-			return reconcile.Result{Requeue: true}, fmt.Errorf("failed to create role: %v", err)
+			return reconcile.Result{}, fmt.Errorf("failed to create role: %v", err)
+		}
+
+		// Get the role just created so the UID is available for the ownerReference on the roleBinding.
+		role := &rbacv1.Role{}
+		if err := r.cache.Get(context.TODO(), types.NamespacedName{Namespace: r.config.SecretNamespace, Name: roleName}, role); err != nil {
```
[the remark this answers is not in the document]
awgreene: I'll change this to a `r.client.Get(...)` call to avoid the race condition.

pkg/operator/controller/configurable-route/controller.go
awgreene: Oh that's a cool function. Thanks!

Miciah:
Lovely jubbly!
/lgtm

PR #554, opened 2021-02-24
daemonset should be able to tolerate the taints

bmeng:
We are using the taints on the node to avoid the customer workload to be scheduled on master or infra in OSD.
Currently, the pod for ds ingress-canary will not be scheduled if taints set on the node. We should be able to tolerate those taints.

Another thoughts is if we need to use `daemonset` here for the canary service? Maybe a `deployment` with some replicas running on random node is enough? 
If the `deamonset` is needed, then maybe we should better schedule the pods on all the nodes. If no, how about use `deployment` instead of `daemonset`?

cc @sgreene570 PTAL


sgreene570:
> If the `deamonset` is needed, then maybe we should better schedule the pods on all the nodes. If no, how about use `deployment` instead of `daemonset`?
> 
> cc @sgreene570 PTAL

We thought a daemonset would make the most sense since we could verify that the ingress controller is able to server workloads on any node in the cluster. With that in mind, it would probably make the most sense to tolerate all taints and run the canary daemonset on infra nodes as well. Does that reasoning sound good to you @bmeng ?

bmeng:
@sgreene570 looks like the e2e testing always fail...
Could you help take a look at the PR, if it is ok to merge?

sgreene570:
Thanks @bmeng !

This PR has been suprseded by https://github.com/openshift/cluster-ingress-operator/pull/560

We do not want to tolerate all taints for the canary daemonset. PR 560 should let the canary daemonset run on Infra nodes in OSD safely.

/close

sgreene570:
If theres any more questions please follow up on PR 560 or reach out to me on slack.

bmeng:
Thank you so much! @sgreene570

PR #558, opened 2021-02-24
 [release-4.7] Bug 1932649: Canary: Add edge termination to canary route 

sgreene570:
This is the manual expedited 4.7 cherry pick of https://github.com/openshift/cluster-ingress-operator/pull/556

---

**Add edge termination to canary route** (credit to @jomeier)

This commit adds edge termination to the canary route. If you use an external loadbalancer that redirects all http traffic to https (us in our case) this route will not serve its purpose and the openshift-ingress-operator will be degraded during an upgrade.
Setting it to 'edge' termination should always work.

**Call: 'make generate'.**

**Canary: Skip TLS verification in canary HTTP client**

Commit `c98416` switched the canary route from a cleartext route to an edge terminated route. In order for canary probes to succeed, the canary client needs to skip TLS verification, since the router's default certificate may be a self-signed certificate. This change is temporary, since ideally the router's default certificate (or a new certificate for the canary route) would be added to the canary client's trust bundle.


sgreene570:
```
    canary_test.go:120: failed to observe the expected canary response body: timed out waiting for the condition
--- FAIL: TestCanaryRoute (301.01s)
```

Forgot to update the canary e2e test earlier. Should be fixed by 485f7e0.

PR #560, opened 2021-02-25
Bug 1933102: Canary: Schedule canary server pods to worker and infra nodes

sgreene570:
**vendor: Pull in openshift/api/project/v1**

These API definitions are needed in pkg/operator/controller/canary/namespace.go.

---

**Canary: Add node-selector annotation to canary namespace**

assets/canary/namespace.yaml:

Add the `openshift.io/node-selector` annotation to the canary namespace yaml file. This will allow the canary daemonset to schedule pods onto infra nodes (in addition to worker nodes) by overriding the default behavior.

pkg/manifests/bindata.go:

Regenerate.

pkg/operator/controller/canary/namespace.go:

Add logic to update the canary namespace resource on a cluster specifically when the namespace's node-selector annotation changes.

pkg/operator/controller/canary/namespace_test.go:

Add a unit test to exercise the new `canaryNamespaceChanged` function.

pkg/operator/controller/canary/controller.go:

Update `ensureCanaryNamespace` calling site.

---

**Canary: Add node selector and tolerations to daemonset**

assets/canary/daemonset.go:

Add a nodeSelector field to the canary daemonset asset that selects linux nodes. Add a tolerations field to tolerate infra node taints, so that the canary daemonset schedules pods to worker and infra nodes only.

pkg/manifests/bindata.go:

Regenerate

pkg/operator/controller/canary/daemonset.go:

Add logic to update the canary daemonset resource on a cluster specifically when the daemonset's node selector, tolerations, or canary server image reference changes.

pkg/operator/controller/canary/daemonset_test.go

Add unit tests to exercise the new `canaryDaemonSetChanged` function.


michaelryanmcneill:
This looks great @sgreene570! Thanks for bringing the changes all together!

sgreene570:
unclear if this is actually a necessary change or if the problem hit by @michaelryanmcneill is a strange edge case in his home lab (since their cluster does not have node taints).

/hold for now, I will investigate more next week

pkg/operator/controller/canary/namespace.go
```diff
@@ -4,24 +4,93 @@ import (
...
+// canaryNamespaceChanged returns true if current and expected differ by the openshift
+// namespace node-selector annotation
+func canaryNamespaceChanged(current, expected *corev1.Namespace) (bool, *corev1.Namespace) {
+	updated := current.DeepCopy()
+
+	if updated.Annotations == nil {
+		updated.Annotations = map[string]string{}
+	}
+
+	if current.Annotations[NodeSelectorAnnotation] == expected.Annotations[NodeSelectorAnnotation] {
```
Miciah, line 89: You initialized `updated.Annotations`, but `current.Annotations` (which came from the API) could still be nil.
```suggestion
	if updated.Annotations[NodeSelectorAnnotation] == expected.Annotations[NodeSelectorAnnotation] {
```

Miciah: Never mind, this is file (reading from a nil map is safe; only assigning to a nil map is dangerous, because Go is weird).

pkg/operator/controller/canary/namespace.go
```diff
@@ -4,24 +4,93 @@ import (
...
+		return false, nil
+	}
+
+	updated.Annotations[NodeSelectorAnnotation] = expected.Annotations[NodeSelectorAnnotation]
```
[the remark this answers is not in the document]
sgreene570: Fixed! :grin:

pkg/operator/controller/canary/namespace.go
```diff
@@ -4,24 +4,93 @@ import (
...
-func (r *reconciler) ensureCanaryNamespace() error {
-	ns := manifests.CanaryNamespace()
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get ingress canary namespace %q: %v", ns.Name, err)
+func (r *reconciler) ensureCanaryNamespace() (bool, *corev1.Namespace, error) {
+	desired := manifests.CanaryNamespace()
+
+	haveNamespace, current, err := r.currentCanaryNamespace()
+
```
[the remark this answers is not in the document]
sgreene570: whoops. I had a few of these. I think I caught them all in my latest push.

sgreene570:
> /hold for now, I will investigate more next week

Looks like we want the canary daemonset to schedule pods on both worker nodes and infra nodes.
/hold cancel

sgreene570:
`7c8d04b` adds logic to update the canary daemonset when the canary server image reference changes. We forgot to code that logic in for 4.7, so backporting this change will resolve that gap. :grin:

pkg/operator/controller/canary/daemonset.go
```diff
@@ -53,6 +62,21 @@ func (r *reconciler) createCanaryDaemonSet(daemonset *appsv1.DaemonSet) error {
 	return nil
 }
 
+// updatedCanaryDaemonSet updates the canary daemonset if an appropriate change
```
Miciah, line 65:
```suggestion
// updateCanaryDaemonSet updates the canary daemonset if an appropriate change
```
But we can get that in a follow-up.

Miciah (commented):
/lgtm

pkg/operator/controller/canary/daemonset.go
sgreene570: So close!

pkg/operator/controller/canary/daemonset.go
```diff
@@ -73,3 +97,60 @@ func desiredCanaryDaemonSet(canaryImage string) *appsv1.DaemonSet {
 
 	return daemonset
 }
+
+// canaryDaemonSetChanged returns true if current and expected differ by the pod template's
+// node selector, tolerations, or container image reference.
+func canaryDaemonSetChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.DaemonSet) {
```
candita, line 103: Are namespace and command changes excluded?

sgreene570: Not sure I follow. The canary namespace is reconciled separately in `namespace.go`.

For PR 561, you could add logic here to trigger updates when the canary container command changes.
Does that make sense?

candita: To determine if there was a change, you check node selector, tolerations, and container image reference.
I was just wondering if you should also be checking namespace and command changes here to determine if there was a change.

sgreene570: Is there a situation we can think of where the canary daemonset's namespace would change? I think we want to keep that daemonset in the openshift-ingress-canary namespace always. 

For updates when the container command changes, nothing is changing `Spec.Template.Spec.Containers[0].Command` at the moment, so I dont think we need to check for updates there in this PR.

pkg/operator/controller/canary/daemonset_test.go
```diff
@@ -56,4 +58,71 @@ func TestDesiredCanaryDaemonSet(t *testing.T) {
...
+				}
+			},
+			expect: true,
+		},
+		{
+			description: "if pod template tolerations changes",
+			mutate: func(ds *appsv1.DaemonSet) {
+				ds.Spec.Template.Spec.Tolerations = []corev1.Toleration{
+					{
+						Key:      "test",
```
candita, line 98: Maybe add a check the functionality of the comparator, as this just checks for the existence of a new toleration.

sgreene570: `cmpTolerations` is for sorting a slice of tolerations. This test does exercise a comparison between 2 different tolerations as is.

sgreene570:
commit 00a9a6f fixes the `updatedCanaryDaemonSet` typo and adds a tolerations check to `TestDesiredCanaryDaemonSet`.

pkg/operator/controller/canary/daemonset_test.go
sgreene570: Ive added a quick tolerations check to `TestDesiredCanaryDaemonSet` to highlight this (and since that coverage was missing).

pkg/operator/controller/canary/daemonset.go
```diff
@@ -19,16 +23,21 @@ func (r *reconciler) ensureCanaryDaemonSet() (bool, *appsv1.DaemonSet, error) {
...
+	case haveDs:
+		if updated, err := r.updateCanaryDaemonSet(current, desired); err != nil {
+			return true, current, err
+		} else if updated {
+			return r.currentCanaryDaemonSet()
+		}
 	}
 
-	return true, desired, nil
+	return true, current, nil
```
candita, line 40: Couldn't you `return r.currentCanaryDaemonSet()` here too? Then you wouldn't need them on lines 31 and 36.

sgreene570: We use this flow to optimize out a few API calls.... we only need to fetch the current resource from the API if it were recently modified or created. If no change at all happens (ie no update necessary), then we can return `current` as is.

pkg/operator/controller/canary/daemonset_test.go
candita: Can one field of a toleration change to indicate a daemonset change?  I.e. if I change the Value or Operator of a toleration, do you mark the daemonset changed?

pkg/operator/controller/canary/daemonset.go
candita: Ok, yes, I'm getting ahead of myself, thinking about PR 561.

pkg/operator/controller/canary/daemonset_test.go
sgreene570: Changing any part of the toleration slice would trigger an update, yea. Should we add more test cases here? At a certain point we would just be testing that `cmp.Equal` works as expected.

Miciah: It's worth noting that `cmpTolerations` is copied from cluster-ingress-operator, which already has unit tests.  In particular, cluster-ingress-operator has a test case where the key changes:  https://github.com/openshift/cluster-ingress-operator/blob/7b41dc34fedc5107a8fe79a2a16d718b070aea09/pkg/operator/controller/ingress/deployment_test.go#L636-L643

It might be worth moving some of this common code to a shared package at some point.

PR #569, opened 2021-03-08
Bug 1959194: Ingress rollouts should specify minReadySeconds

smarterclayton:
Ingress components may run directly behind load balancers and thus
need to delay deployment rollout long enough for the load balancers
to see the new process. Without minReadySeconds, the upgrade process
will immediately delete the old pod as soon as the new pod is ready,
which means that the LB may briefly see no pods, which then triggers
on some LB the behavior of including all hosts (generally if all
hosts are unhealthy the LB will allow requests to go to any host
as opposed to failing closed).

while we test with this

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -200,6 +200,16 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	gracePeriod := int64(60 * 60)
 	deployment.Spec.Template.Spec.TerminationGracePeriodSeconds = &gracePeriod
 
+	// Services behind load balancers should roll out new instances only after we are certain
+	// the new instance is part of rotation. This is set based on the highest value across all
+	// platforms, excluding custom load balancers like an F5, but our recommendation for these
+	// values for those should be indentical to the slowest cloud, AWS (which does not allow
+	// health checks to be more frequent than 10 seconds).
+	deployment.Spec.MinReadySeconds =
+		(2 + /* max healthy checks required to be brought into rotation across all platforms */
+			1) * /* we could miss one */
+			10 /* the longest health check interval on any platform */
+
 	volumes := deployment.Spec.Template.Spec.Volumes
 	routerVolumeMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
 
@@ -994,6 +1004,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	// Copy metadata and spec fields to which any changes should trigger an
 	// update of the deployment but should not trigger a rolling update.
 	hashableDeployment.Labels = deployment.Labels
+	hashableDeployment.Spec.MinReadySeconds = deployment.Spec.MinReadySeconds
 	hashableDeployment.Spec.Strategy = deployment.Spec.Strategy
 	var replicas *int32
 	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != int32(1) {
@@ -1178,6 +1189,7 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 		replicas = *expected.Spec.Replicas
 	}
 	updated.Spec.Replicas = &replicas
+	updated.Spec.MinReadySeconds = expected.Spec.MinReadySeconds
 	return true, updated
 }
 
--- a/test/e2e/http_header_buffer_test.go
+++ b/test/e2e/http_header_buffer_test.go
@@ -212,7 +212,7 @@ func TestHTTPHeaderBufferSize(t *testing.T) {
 	}
 
 	// Wait for the new router pod for the updated ingresscontroller to become ready.
-	pollErr = wait.PollImmediate(2*time.Second, 3*time.Minute, func() (bool, error) {
+	pollErr = wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
 		podList := &corev1.PodList{}
 		if err := kclient.List(context.TODO(), podList, client.InNamespace(deployment.Namespace), client.MatchingLabels(labels)); err != nil {
 			t.Errorf("failed to list pods for ingress controllers %s: %v", ic.Name, err)
@@ -220,16 +220,19 @@ func TestHTTPHeaderBufferSize(t *testing.T) {
 
 		for _, pod := range podList.Items {
 			if pod.Name == oldRouterPodName {
-				continue
+				// wait until the old router pod is gone before continuing
+				return false, nil
 			}
+		}
 
+		for _, pod := range podList.Items {
 			for _, cond := range pod.Status.Conditions {
 				if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
 					return true, nil
 				}
 			}
-
 		}
+
 		t.Logf("waiting for new router pod for %s to become ready", ic.Name)
 		return false, nil
 	})
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -4,6 +4,7 @@ package e2e
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
@@ -1552,7 +1553,7 @@ func TestHTTPHeaderCapture(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create kube client: %v", err)
 	}
-	err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+	err = wait.PollImmediate(1*time.Second, 3*time.Minute, func() (bool, error) {
 		for _, pod := range podList.Items {
 			readCloser, err := client.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
 				Container: "logs",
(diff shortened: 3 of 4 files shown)


Miciah:
You'll need some logic to handle updates:
```diff
diff --git a/pkg/operator/controller/ingress/deployment.go b/pkg/operator/controller/ingress/deployment.go
index 44834362..8552bc1c 100644
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -941,6 +941,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	// Copy metadata and spec fields to which any changes should trigger an
 	// update of the deployment but should not trigger a rolling update.
 	hashableDeployment.Labels = deployment.Labels
+	hashableDeployment.Spec.MinReadySeconds = deployment.Spec.MinReadySeconds
 	hashableDeployment.Spec.Strategy = deployment.Spec.Strategy
 	var replicas *int32
 	if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != int32(1) {
@@ -1070,6 +1071,7 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 		updated.Spec.Template.Annotations[LivenessGracePeriodSecondsAnnotation] = val
 	}
 
+	updated.Spec.MinReadySeconds = expected.Spec.MinReadySeconds
 	updated.Spec.Strategy = expected.Spec.Strategy
 	volumes := make([]corev1.Volume, len(expected.Spec.Template.Spec.Volumes))
 	for i, vol := range expected.Spec.Template.Spec.Volumes {
diff --git a/pkg/operator/controller/ingress/deployment_test.go b/pkg/operator/controller/ingress/deployment_test.go
index cda60029..987c1de5 100644
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -925,6 +925,13 @@ func TestDeploymentConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if spec.minReadySeconds is changed",
+			mutate: func(deployment *appsv1.Deployment) {
+				deployment.Spec.MinReadySeconds = int32(30)
+			},
+			expect: true,
+		},
 	}
 
 	for _, tc := range testCases {
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -941,6 +951,9 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	// Copy metadata and spec fields to which any changes should trigger an
 	// update of the deployment but should not trigger a rolling update.
 	hashableDeployment.Labels = deployment.Labels
+	if deployment.Spec.MinReadySeconds != 30 {
+		hashableDeployment.Spec.MinReadySeconds = deployment.Spec.MinReadySeconds
```
Miciah, line 955: What is the purpose of the condition?  This will cause a deployment with `spec.minReadySeconds: 0` and a deployment with `spec.minReadySeconds: 30` to hash to the same value (if all other fields are the same), which would prevent the operator from updating the deployment if it had `spec.minReadySeconds: 0`.

Miciah:
Looks like this timeout needs to be increased:
https://github.com/openshift/cluster-ingress-operator/blob/7b41dc34fedc5107a8fe79a2a16d718b070aea09/test/e2e/operator_test.go#L1722
We might need to increase some other polling loop timeouts in tests to avoid increasing the flake rate with the higher `minReadySeconds` value

test/e2e/operator_test.go
```diff
@@ -360,6 +360,13 @@ func TestIngressControllerScale(t *testing.T) {
 		t.Fatalf("failed to get default ingresscontroller deployment: %v", err)
 	}
 
+	// use a shorter minReadySeconds in the test to keep the test fast and also
+	// to verify that customers can tune this value for their own load balancers
+	deployment.Spec.MinReadySeconds = 5
```
Miciah, line 365: You wanted to update the ingresscontroller, right?

test/e2e/operator_test.go
```diff
@@ -360,6 +360,13 @@ func TestIngressControllerScale(t *testing.T) {
...
+	if err := kclient.Update(context.TODO(), deployment); err != nil {
+		t.Fatalf("failed to get default ingresscontroller deployment: %v", err)
```
Miciah, line 367: "get" → "update".

test/e2e/operator_test.go
Miciah: Er, wait, this isn't exposed as configurable on the IngressController.  The operator should revert this change.

smarterclayton: Yes - when i run the test locally I get a pass in 29s.  Not sure why it's not passing in the job

smarterclayton: Operator will turn 0 into the default, but customers who have LB that may have wildly different configs can change the deployment to another value.  I *think* we allow them to already change the service LB annotations, which would also impact this, but will double check.

Miciah: Allowing the user to modify fields on operator-managed resources is scary and generally something we want to prohibit, but it would be expedient and easier than extending the IngressController API at this point...

smarterclayton: Don't we explicitly allow changes to other deployment objects?  I thought that was the whole point of your logic in the deployment sync code you had me change?

smarterclayton:
I'm waiting for cluster bot to come back before I test this again

smarterclayton:
/retest

cookie header, looks unrelated, as well as the kubelet kill

test/e2e/operator_test.go
Miciah: The update logic originally did as little as necessary (which was naïve in retrospect) and gradually grew as we added features, with fixes along the way for things like fighting the API when it defaults fields.  I don't believe the intent was ever that administrators would directly modify operands.  Unfortunately, administrators _did_ start directly modifying operands, which causes serious problems when we want the operator to start managing some particular field.  

It might be reasonable to make an exception for `minReadySeconds`; I just want to be careful that we don't create future problems.  In particular, it would make it difficult to extend IngressController with a field for configuring `minReadySeconds` if we ever wanted to do that.

smarterclayton: In the long run we probably want this to be detected or configured by the admin, so i’m fine not allowing the user to set it for now (it’s an improvement over other flows).

Still need to debug why the cookie test was failing, i’m wondering if i missed a spot in the reconcile loop where it’s doing two rollouts instead of one.

smarterclayton:
With debug output:

```
    http_header_buffer_test.go:242: 2021-05-10T19:52:41.590875988Z: Router pod available
    http_header_buffer_test.go:265: failed to read output from pod header-buffer-size-test-large-buffers-fail-case: container "curl" in pod "header-buffer-size-test-large-buffers-fail-case" is waiting to start: ContainerCreating
... // pod events:
2021-05-10 19:52:39.827785306+00:00 router-header-buffer-size-f578db66d-g88vx    1/1     Running             0          4s
2021-05-10 19:53:09.863275719+00:00 router-header-buffer-size-6847776996-7lnkx   1/1     Terminating         0          77s
2021-05-10 19:53:32.461407573+00:00 router-header-buffer-size-6847776996-7lnkx   0/1     Terminating         0          100s
2021-05-10 19:54:02.051643661+00:00 router-header-buffer-size-6847776996-7lnkx   0/1     Terminating         0          2m10s
2021-05-10 19:54:10.653217974+00:00 router-header-buffer-size-6847776996-7lnkx   0/1     Terminating         0          2m18s
2021-05-10 19:54:10.688566590+00:00 router-header-buffer-size-6847776996-7lnkx   0/1     Terminating         0          2m18s
2021-05-10 19:55:43.036789578+00:00 router-header-buffer-size-f578db66d-g88vx    1/1     Terminating         0          3m8s
2021-05-10 19:56:05.790032037+00:00 router-header-buffer-size-f578db66d-g88vx    0/1     Terminating         0          3m30s
2021-05-10 19:56:30.492768496+00:00 router-header-buffer-size-f578db66d-g88vx    0/1     Terminating         0          3m55s
2021-05-10 19:56:40.671552026+00:00 router-header-buffer-size-f578db66d-g88vx    0/1     Terminating         0          4m5s
2021-05-10 19:56:40.713757215+00:00 router-header-buffer-size-f578db66d-g88vx    0/1     Terminating         0          4m5s

```

smarterclayton:
Hrm, now I'm wondering whether this is truly required if we turn off the local traffic policy (although I guess people can turn it back on if they want the old behavior).  With local traffic policy on, we need this because we have to wait for another node to go ready (for the other endpoint to be healthy).  With that off, etcd should always propagate to have at least one ready pod:

1. new pod ready
2. deployment controller sees new pod ready and deletes old pod
3. endpoint controller either sees new pod + old pod, or just new pod ready.

That said minReadySeconds is good for other reasons (making sure we don't crash shortly after), so it's probably still a net win, but the original timing (cloud LB / kube-proxy propagation time, which are both about 30s) may not be the best guide (cloud lb is irrelevant without local traffic, kube-proxy should never see the intermediate no ready pods state).  OTOH 30s is nice and predictable and would work even onpremises and on all LB modes.

test/e2e/http_header_buffer_test.go
```diff
@@ -212,24 +212,25 @@ func TestHTTPHeaderBufferSize(t *testing.T) {
...
+				// wait until the old router pod is gone before continuing
+				return false, nil
 			}
+		}
 
-			for _, cond := range pod.Status.Conditions {
-				if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
-					return true, nil
-				}
+		for _, cond := range pod.Status.Conditions {
```
candita, line 228: Do you need to check status for each pod in the podlist?

sgreene570:
CI is green
/lgtm

smarterclayton:
/retest 

(unless this horrifically breaks install, which i don't think it does)

smarterclayton:
/retest

What is going on.

smarterclayton:
Fixed the commit message on the one test i changed, no functional changes (we had a green run).

PR #572, opened 2021-03-11
Use HAProxy log level "info" to enable access logs

Miciah:
Before this PR, the operator configured HAProxy with log level "debug" to enable access logs.  However, access logs are emitted at the "info" or "error" levels, so setting the log level to "info" is sufficient and avoids logging superfluous information.

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Set the `ROUTER_LOG_LEVEL` environment variable's value to "info" instead of to "debug" when enabling access logs.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Expect "info" instead of "debug" for `ROUTER_LOG_LEVEL`.


sgreene570:
good catch!
/lgtm

PR #575, opened 2021-03-17
Bug 1928931: Update DNSRecord CRD from v1beta1 to v1

rfredette:
CRD is being removed from v1beta1 in Kubernetes 1.22, so this PR changes DNSRecord to the v1 version of CRD.

The process I followed for this change was:
```bash
go get github.com/openshift/api@f67d7a61e06075e9c84bc731fca470fff2832227
go mod vendor
make generate
```

[openshift/api commit f67d7a61e06075e9c84bc731fca470fff2832227](https://github.com/openshift/api/commit/f67d7a61e06075e9c84bc731fca470fff2832227) is the commit that updates DNSRecord in the openshift/api tree, and as of the creation of this PR, the latest commit in openshift/api's master branch.

PR #576, opened 2021-03-18
NE-354: Implement Tunable router header buffers

sgreene570:
**Implement Router Tunable Buffer Sizes**

New func, `validateHTTPHeaderBufferValues`, to validate `HTTPHeaderBuffer` values when an ingress controller specifies them. Reject ingress controllers that specify improper `HTTPHeaderBuffer` values (ie `HeaderBufferBytes < HeaderBufferMaxRewrite`).

pkg/operator/controller/ingress/controller_test.go:

Add unit test coverage for `validateHTTPHederBufferValues`.

pkg/operator/controller/ingress/deployment.go:

Append `ROUTER_BUF_SIZE` and `ROUTER_MAX_REWRITE_SIZE` env variables to an IngressController's deployment when an IngressController specifies either `HeaderBufferBytes` or `HeaderBufferMaxRewriteBytes` via `Spec.HTTPHeaderBuffer`.

pkg/operator/controller/ingress/deployment_test.go:

Add valid `HTTPHeaderBuffer` values to test deployment in `TestDesiredRouterDeployment`.

**test/e2e: Add e2e test to test configurable header buffers**

test/e2e/http_header_buffer_test.go: New file.

Add new e2e test, `TestHTTPHeaderBufferSize`, that performs the following:

1) Create an Ingress Controller with header buffer values that are 2x larger than the defaults.

2) Send an HTTP request with a test header packed with random data to a header echo server and verify that the header is successfully received. Note that the test request has headers large enough such that the default ingress controller would return "400 Bad request" if it were handling the request.

3) Mutate the created Ingress Controller so that it is using header buffer values that are 1/2 the size of the defaults.

4) Send the same HTTP request with large headers as before.

5) Ensure that HAProxy responds with "400 Bad Request" since the request headers exceed the buffer limits.

---
Prequisites:

- [x] https://github.com/openshift/cluster-ingress-operator/pull/575
- [x] https://github.com/openshift/api/pull/854
- [x] https://github.com/openshift/router/pull/193

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -48,6 +48,9 @@ const (
 	IngressControllerDeploymentReplicasMinAvailableConditionType = "DeploymentReplicasMinAvailable"
 	IngressControllerDeploymentReplicasAllAvailableConditionType = "DeploymentReplicasAllAvailable"
 	IngressControllerCanaryCheckSuccessConditionType             = "CanaryChecksSucceeding"
+
+	routerDefaultHeaderBufferSize           = 32768
+	routerDefaultHeaderBufferMaxRewriteSize = 8192
 )
 
 var (
@@ -455,7 +458,9 @@ func (r *reconciler) validate(ic *operatorv1.IngressController) error {
 	if err := validateTLSSecurityProfile(ic); err != nil {
 		errors = append(errors, err)
 	}
-
+	if err := validateHTTPHeaderBufferValues(ic); err != nil {
+		errors = append(errors, err)
+	}
 	if err := utilerrors.NewAggregate(errors); err != nil {
 		return &admissionRejection{err.Error()}
 	}
@@ -557,6 +562,36 @@ func filterTLS13Ciphers(ciphers []string) []string {
 		}
 	}
 	return filteredCiphers
+}
+
+// validateHTTPHeaderBufferValues validates the given ingresscontroller's header buffer
+// size configuration, if it specifies one.
+func validateHTTPHeaderBufferValues(ic *operatorv1.IngressController) error {
+	bufSize := int(ic.Spec.HTTPHeaderBuffer.HeaderBufferBytes)
+	maxRewrite := int(ic.Spec.HTTPHeaderBuffer.HeaderBufferMaxRewriteBytes)
+
+	if bufSize == 0 && maxRewrite == 0 {
+		return nil
+	}
+
+	// HeaderBufferBytes and HeaderBufferMaxRewriteBytes are both
+	// optional fields. Substitute the default values used by the
+	// router when either field is empty so that we can ensure that
+	// tune.maxrewrite will never wind up being greater than tune.bufsize
+	// (which would break router reloads).
+	if bufSize == 0 {
+		bufSize = routerDefaultHeaderBufferSize
+	}
+	if maxRewrite == 0 {
+		maxRewrite = routerDefaultHeaderBufferMaxRewriteSize
+	}
+
+	if bufSize <= maxRewrite {
+		return fmt.Errorf("invalid spec.httpHeaderBuffer: headerBufferBytes (%d) "+
+			"must be larger than headerBufferMaxRewriteBytes (%d)", bufSize, maxRewrite)
+	}
+
+	return nil
 }
 
 // ensureIngressDeleted tries to delete ingress, and if successful, will remove
--- a/pkg/operator/controller/ingress/controller_test.go
+++ b/pkg/operator/controller/ingress/controller_test.go
@@ -317,3 +317,60 @@ func TestTLSProfileSpecForIngressController(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateHTTPHeaderBufferValues(t *testing.T) {
+	testCases := []struct {
+		description      string
+		httpHeaderBuffer operatorv1.IngressControllerHTTPHeaderBuffer
+		valid            bool
+	}{
+		{
+			description:      "no httpHeaderBuffer values",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{},
+			valid:            true,
+		},
+		{
+			description: "valid httpHeaderBuffer values",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferBytes:           32768,
+				HeaderBufferMaxRewriteBytes: 8192,
+			},
+			valid: true,
+		},
+		{
+			description: "invalid httpHeaderBuffer values",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferBytes:           8192,
+				HeaderBufferMaxRewriteBytes: 32768,
+			},
+			valid: false,
+		},
+		{
+			description: "invalid httpHeaderBuffer values, HeaderBufferMaxRewriteBytes not set",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferBytes: 1,
+			},
+			valid: false,
+		},
+		{
+			description: "invalid httpHeaderBuffer values, HeaderBufferBytes not set",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferMaxRewriteBytes: 65536,
+			},
+			valid: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		ic := &operatorv1.IngressController{}
+		ic.Spec.HTTPHeaderBuffer = tc.httpHeaderBuffer
+		err := validateHTTPHeaderBufferValues(ic)
+		if tc.valid && err != nil {
+			t.Errorf("%q: Expected valid HTTPHeaderBuffer to not return a validation error: %v", tc.description, err)
+		}
+
+		if !tc.valid && err == nil {
+			t.Errorf("%q: Expected invalid HTTPHeaderBuffer to return a validation error", tc.description)
+		}
+	}
+}
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -53,6 +53,9 @@ const (
 	RouterCaptureHTTPResponseHeaders = "ROUTER_CAPTURE_HTTP_RESPONSE_HEADERS"
 	RouterCaptureHTTPCookies         = "ROUTER_CAPTURE_HTTP_COOKIE"
 
+	RouterHeaderBufferSize           = "ROUTER_BUF_SIZE"
+	RouterHeaderBufferMaxRewriteSize = "ROUTER_MAX_REWRITE_SIZE"
+
 	RouterDisableHTTP2EnvName          = "ROUTER_DISABLE_HTTP2"
 	RouterDefaultEnableHTTP2Annotation = "ingress.operator.openshift.io/default-enable-http2"
 
@@ -658,6 +661,18 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 
 	if enabled, value := HardStopAfterIsEnabled(ci, ingressConfig); enabled {
 		env = append(env, corev1.EnvVar{Name: RouterHardStopAfterEnvName, Value: value})
+	}
+
+	// Apply HTTP Header Buffer size values to env
+	// when they are specified.
+	if ci.Spec.HTTPHeaderBuffer.HeaderBufferBytes != 0 {
+		env = append(env, corev1.EnvVar{Name: RouterHeaderBufferSize, Value: strconv.Itoa(
+			int(ci.Spec.HTTPHeaderBuffer.HeaderBufferBytes))})
+	}
+
+	if ci.Spec.HTTPHeaderBuffer.HeaderBufferMaxRewriteBytes != 0 {
+		env = append(env, corev1.EnvVar{Name: RouterHeaderBufferMaxRewriteSize, Value: strconv.Itoa(
+			int(ci.Spec.HTTPHeaderBuffer.HeaderBufferMaxRewriteBytes))})
 	}
 
 	deployment.Spec.Template.Spec.Volumes = volumes
--- a/test/e2e/http_header_buffer_test.go
+++ b/test/e2e/http_header_buffer_test.go
@@ -1,0 +1,311 @@
+// +build e2e
+
+package e2e
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"math/rand"
+	"strings"
+	"testing"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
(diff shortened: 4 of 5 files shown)


sgreene570:
Rebased now that https://github.com/openshift/cluster-ingress-operator/pull/575 has merged.

Leaving WIP while I verify that the new e2e test passes in CI.

sgreene570:
How ironic:
```
level=error msg=Error: error updating LB Target Group (arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-msdrpty1-43abb-k2lkx-aext/8aa060b0895fbb34) tags: error tagging resource (arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-msdrpty1-43abb-k2lkx-aext/8aa060b0895fbb34): TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-msdrpty1-43abb-k2lkx-aext/8aa060b0895fbb34' not found
````

/test e2e-aws-operator

**Comment by sgreene570** [comment]
failed on must-gather
/test e2e-aws-operator

File: pkg/operator/controller/ingress/controller.go [hunk 1]
```diff
@@ -559,6 +563,35 @@ func filterTLS13Ciphers(ciphers []string) []string {
 	return filteredCiphers
 }
 
+// validateHTTPHeaderBuffer validates the given ingresscontroller's header buffer
```
frobware, line 566: func name is different.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -317,3 +317,60 @@ func TestTLSProfileSpecForIngressController(t *testing.T) {
...
+		{
+			description: "valid, non-empty httpHeaderBuffer",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferBytes:           32768,
+				HeaderBufferMaxRewriteBytes: 8192,
+			},
+			valid: true,
+		},
+		{
+			description: "invalid, non-empty httpHeaderBuffer",
```
frobware, line 341: The wording confused me; we talked offline so I think you're going to propose "field is set??"

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -317,3 +317,60 @@ func TestTLSProfileSpecForIngressController(t *testing.T) {
...
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferBytes:           8192,
+				HeaderBufferMaxRewriteBytes: 32768,
+			},
+			valid: false,
+		},
+		{
+			description: "invalid httpHeaderBuffer, HeaderBufferMaxRewriteBytes empty",
```
frobware, line 349: nit: empty =~ "not specified"? Just trying to distinguish whether empty =~ the zero value.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -317,3 +317,60 @@ func TestTLSProfileSpecForIngressController(t *testing.T) {
...
+				HeaderBufferBytes: 1,
+			},
+			valid: false,
+		},
+		{
+			description: "invalid httpHeaderBuffer, HeaderBufferBytes empty",
+			httpHeaderBuffer: operatorv1.IngressControllerHTTPHeaderBuffer{
+				HeaderBufferMaxRewriteBytes: 16384,
+			},
+			valid: true,
```
frobware, line 360: The description says this is "invalid".

pkg/operator/controller/ingress/deployment.go
```diff
@@ -660,6 +663,20 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: RouterHardStopAfterEnvName, Value: value})
 	}
 
+	// Only check for custom header buffer settings if
+	// ci.Spec.HTTPHeaderBuffer is not empty.
+	if (ci.Spec.HTTPHeaderBuffer != operatorv1.IngressControllerHTTPHeaderBuffer{}) {
```
frobware, line 668: It seems like you could just pickup non-zero values and this check is not necessary.

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,311 @@
...
+	"bufio"
+	"context"
+	"fmt"
+	"math/rand"
+	"strings"
+	"testing"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
```
frobware, line 15: nit: group the openshift imports

pkg/operator/controller/ingress/controller.go
```diff
@@ -559,6 +563,35 @@ func filterTLS13Ciphers(ciphers []string) []string {
...
+	maxRewrite := int(ic.Spec.HTTPHeaderBuffer.HeaderBufferMaxRewriteBytes)
+
+	// HeaderBufferBytes and HeaderBufferMaxRewriteBytes are both
+	// optional fields. Substitute the default values used by the
+	// router when either field is empty so that we can ensure that
+	// tune.maxrewrite will never wind up being greater than tune.bufsize
+	// (which would break router reloads).
+	if bufSize == 0 {
+		bufSize = routerDefaultHeaderBufferSize
+	} else if maxRewrite == 0 {
```
frobware, line 583: Should this be a discrete check (i.e., drop the `else` here)?

frobware (commented):
I need to take another pass over the e2e test, just some initial feedback.

pkg/operator/controller/ingress/controller.go
sgreene570: fixed.

pkg/operator/controller/ingress/controller_test.go
sgreene570: Changed wording. Should be good now. Went with `values` phrasing instead of `field is set`.

pkg/operator/controller/ingress/controller_test.go
sgreene570: should be fixed now.

pkg/operator/controller/ingress/controller_test.go
sgreene570: whoops, needed to flip that and raise the value in the test. Fixed.

pkg/operator/controller/ingress/deployment.go
sgreene570: yea, we talked offline and I confirmed locally. Fixed.

test/e2e/http_header_buffer_test.go
sgreene570: fixed.

pkg/operator/controller/ingress/controller.go
sgreene570: No reason not to I suppose. Fixed.

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,310 @@
...
+		output := fmt.Sprintf("failed to observe the expected output: %v\nclient pod spec: %#v\nclient pod logs:\n%s", pollErr, pod, logs)
+		output = strings.Replace(output, random, fmt.Sprintf("<random data len %d btyes snipped>", length), 0)
+		t.Fatal(output)
+	}
+}
+
+// randomString quickly returns a string of
+// pseudo-random characters.
+func randomString(length int) (str string) {
+	b := make([]byte, length)
```
Miciah, line 306: So wasteful!
```suggestion
	b := make([]byte, (length+1)/2)
```
(Just kidding.  No need to micro-optimize this...)

pkg/operator/controller/ingress/controller.go
```diff
@@ -559,6 +563,36 @@ func filterTLS13Ciphers(ciphers []string) []string {
...
+	// tune.maxrewrite will never wind up being greater than tune.bufsize
+	// (which would break router reloads).
+	if bufSize == 0 {
+		bufSize = routerDefaultHeaderBufferSize
+	}
+	if maxRewrite == 0 {
+		maxRewrite = routerDefaultHeaderBufferMaxRewriteSize
+	}
+
+	if bufSize < maxRewrite {
```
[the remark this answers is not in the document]
sgreene570: Ah, good catch. I'll change `<` to `=<`.

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,310 @@
...
+	}()
+
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+	cl, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		t.Fatalf("failed to create kube client: %v", err)
+	}
```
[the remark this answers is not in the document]
sgreene570: Does the controller-runtime client have a way to get pod logs? I was basing this off of some of the code in https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/http_header_name_case_adjustment_test.go#L116 (Which I think you wrote)?

pkg/operator/controller/ingress/controller.go
```diff
@@ -57,6 +57,8 @@ var (
 	tlsVersion13Ciphers = sets.NewString("TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256",
 		"TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256", "TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384",
 		"TLS_CHACHA20_POLY1305_SHA256", "TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256")
+	routerDefaultHeaderBufferSize           = 32768
+	routerDefaultHeaderBufferMaxRewriteSize = 8192
```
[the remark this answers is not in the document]
sgreene570: Fixed. Thanks!

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
sgreene570: Fixed.

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,310 @@
...
+		logs, err := cl.CoreV1().Pods(clientPodValidRequest.Namespace).GetLogs(clientPodValidRequest.Name, &corev1.PodLogOptions{
+			Container: "curl",
+			Follow:    false,
+		}).DoRaw(context.TODO())
+		if err != nil {
+			t.Errorf("failed to get logs from pod %s: %v", clientPodValidRequest.Name, err)
+		}
+
+		output := fmt.Sprintf("failed to observe the expected output: %v\nclient pod spec: %#v\nclient pod logs:\n%s", pollErr, pod, logs)
+		output = strings.Replace(output, random, fmt.Sprintf("<random data of len %d btyes snipped>", length), 0)
```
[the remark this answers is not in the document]
sgreene570: Fixed. Thanks!

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller %s: %v", icName, err)
+	}
+
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, conditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Wait for the new router pod for the updated ingresscontroller to become ready.
```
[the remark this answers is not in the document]
sgreene570: Hm, maybe it would be simpler. Mind if I improve on this e2e test in a follow up? Wondering if my current approach might actually shave some time off of the test (But I guess we aren't super worried about how long these tests take).

test/e2e/http_header_buffer_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+	"k8s.io/client-go/kubernetes"
+)
+
+func TestHTTPHeaderBufferSize(t *testing.T) {
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "header-buffer-size"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	// Create an ingress controller with header buffer values that are 2x
+	// the defaults.
```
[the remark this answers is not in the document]
sgreene570: whoops. Fixed!

test/e2e/http_header_buffer_test.go
sgreene570: Alright I'll byte. How is this a micro-optimization? I must be missing something.

test/e2e/http_header_buffer_test.go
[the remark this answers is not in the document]
sgreene570: Might be able to re-use some helpers that might get added in https://github.com/openshift/cluster-ingress-operator/pull/580/

test/e2e/http_header_buffer_test.go
Miciah: You only need half the buffer space that the current logic allocates, and half the pseudo-random bytes that it generates.  The caller wants a string of the specified length, but `fmt.Sprintf("%x", ...)` produces 2 characters per byte, so the buffer only needs to be half as big as `length`, plus 1 to account for odd values for `length`.  What made me think of this was the `[:length]` subscript that you apply to the buffer, which didn't make sense to me until I realized `fmt.Sprintf` was producing twice as many characters as the caller requested.

test/e2e/http_header_buffer_test.go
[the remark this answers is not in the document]
Miciah: Follow-up is fine.

test/e2e/http_header_buffer_test.go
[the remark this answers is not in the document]
Miciah: Whoops!  I thought `kclient` was a client-go client.  These names make no sense to me.  Never mind!

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
Miciah: I guess using `<=` will catch a lot of errors, but it still allows values like `bufSize = 32768` and `maxRewrite = 32767`, which I imagine would cause all requests to fail, right?

test/e2e/http_header_buffer_test.go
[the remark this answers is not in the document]
Miciah: I was actually more concerned about the "btyes" than the "len".  * grin *.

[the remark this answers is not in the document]
sgreene570: Your suggested change kept "bytes" but removed "len"?

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
sgreene570: > I guess using `<=` will catch a lot of errors, but it still allows values like `bufSize = 32768` and `maxRewrite = 32767`, which I imagine would cause all requests to fail, right?

Yea, that's a good point. Not sure how much more validation we want to do here.... Can we think of a minimum difference between bufSize and maxRewrite that we can safely enforce?

test/e2e/http_header_buffer_test.go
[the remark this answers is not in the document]
sgreene570: Wow, My eyes have deceived me yet again.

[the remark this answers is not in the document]
sgreene570: Fixed (for real this time).

test/e2e/http_header_buffer_test.go
sgreene570: Fixed just because.

PR #577, opened 2021-03-21
Bug 1900819: Specify topology spread constraints

Miciah:
Specify topology spread constraints on router deployments to encourage the scheduler to spread router pod replicas across availability zones.

Add a to-do to replace the hard anti-affinity rule with soft anti-affinity once the descheduler has support for enforcing soft anti-affinity rules.

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Rename `needDeploymentHash` to `configureAffinity` as we now always need to compute the deployment hash but do not always specify an affinity policy. Specify `spec.template.spec.topologySpreadConstraints`.  Specify max skew 1, unsatisfiable-constraint action `ScheduleAnyway`, topology key `topology.kubernetes.io/zone`, and a label selector on the deployment hash label so as to specify a preference for spreading replicas of the same generation out but not prevent scheduling replicas of different generations on the same node or scheduling more replicas than there are availability zones.  Add to-do for changing to soft anti-affinity.
(`hashableDeployment`): Hash `spec.template.spec.topologySpreadConstraints`.
(`cmpMatchExpressions`, `zeroOutDeploymentHash`): New functions, factored out of hashableDeployment.
* `pkg/operator/controller/ingress/deployment_test.go` (`checkDeploymentHash`): New function, factored out of `TestDesiredRouterDeployment`.
(`TestDesiredRouterDeployment`): Use `checkDeploymentHash`.  Expect the deployment hash label always to be set.
(`TestDeploymentConfigChanged`): Add test cases for `spec.template.spec.topologySpreadConstraints`.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -209,7 +209,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 	deployment.Spec.Replicas = &desiredReplicas
 
-	needDeploymentHash := false
+	configureAffinity := false
 	switch ci.Status.EndpointPublishingStrategy.Type {
 	case operatorv1.HostNetworkStrategyType:
 		// Typically, an ingress controller will be scaled with replicas
@@ -272,7 +272,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		// that a node that had local endpoints at the start of a
 		// rolling update continues to have local endpoints for the
 		// duration of and at the completion of the update.
-		needDeploymentHash = true
+		configureAffinity = true
 		deployment.Spec.Template.Spec.Affinity = &corev1.Affinity{
 			PodAffinity: &corev1.PodAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
@@ -298,6 +298,10 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 					},
 				},
 			},
+			// TODO: Once https://issues.redhat.com/browse/RFE-1759
+			// is implemented, replace
+			// "RequiredDuringSchedulingIgnoredDuringExecution" with
+			// "PreferredDuringSchedulingIgnoredDuringExecution".
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
 					{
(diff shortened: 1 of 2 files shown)


Miciah:
In the last e2e-operator CI run, [pods.json](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/577/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1373728341844037632/artifacts/e2e-aws-operator/gather-extra/artifacts/pods.json) shows that the router pods were scheduled on worker nodes "ip-10-0-128-140.us-west-2.compute.internal" and "ip-10-0-184-104.us-west-2.compute.internal", which [nodes.json](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/577/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1373728341844037632/artifacts/e2e-aws-operator/gather-extra/artifacts/nodes.json) shows are both in zone "us-west-2a", which looks wrong.  However, looking at the last transition time of the "Ready" status conditions of the nodes, the two aforementioned nodes became ready at times "2021-03-21T21:12:14Z" and "2021-03-21T21:09:09Z" whereas the remaining worker node, "ip-10-0-237-234.us-west-2.compute.internal", became ready at time "2021-03-21T21:15:12Z", and the router pods were created at times "2021-03-21T21:12:50Z" and "2021-03-21T21:09:26Z", so it looks like the pods were scheduled to the same zone because the only available nodes at the time were in the same zone.  

The test failures look unrelated to these changes.

/test e2e-aws-operator

pkg/operator/controller/ingress/deployment.go
```diff
@@ -311,6 +311,27 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
 
+	// Configure topology constraints to spread replicas across availability
+	// zones.  We want to allow scheduling more replicas than there are AZs,
+	// so we specify "ScheduleAnyway".  We want to allow scheduling a
+	// newer-generation replica on the same node as an older-generation
+	// replica where the deployment strategy allows an depends on doing so,
+	// so we specify a label selector with the deployment's hash.
+	deployment.Spec.Template.Spec.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{{
+		MaxSkew:           int32(1),
+		TopologyKey:       "topology.kubernetes.io/zone",
```
sgreene570, line 322:
```suggestion
		TopologyKey:       corev1.LabelTopologyZone,
```

Might as well use that Topology key from corev1, right?

sgreene570:
Unlikely that this would cause a regression with SNO, but lets double check just to be safe
/test e2e-aws-single-node

pkg/operator/controller/ingress/deployment.go
```diff
@@ -311,6 +311,27 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		}
 	}
 
+	// Configure topology constraints to spread replicas across availability
+	// zones.  We want to allow scheduling more replicas than there are AZs,
+	// so we specify "ScheduleAnyway".  We want to allow scheduling a
+	// newer-generation replica on the same node as an older-generation
+	// replica where the deployment strategy allows an depends on doing so,
```
sgreene570, line 318:
```suggestion
	// replica where the deployment strategy allows and depends on doing so,
```
Typo?

pkg/operator/controller/ingress/deployment.go
Miciah: Done.  Thanks!

pkg/operator/controller/ingress/deployment.go
Miciah: Fixes.  Thanks!

Miciah:
Rebased and feedback addressed.

sgreene570:
Thanks for attending to this!
/lgtm

PR #578, opened 2021-03-23
Bug 1950284: Annotate services of type LoadBalancer with user tags (AWS only)

frobware:
This PR addresses https://issues.redhat.com/browse/NE-563.

This is waiting on:
- https://github.com/openshift/api/pull/864
- https://github.com/openshift/enhancements/pull/706
- https://github.com/openshift/installer/pull/4755
 
If there are `AWS.UserTags` defined then services of type LoadBalancer
will be annotated with the following annotation.

    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags

Values for this annotation are encoded as a comma separated list. For example:

    "Key1=Val1,Key2=Val2,KeyNoVal1=,KeyNoVal2"

The conceptual docs for this annotation are documented here:
https://kubernetes.io/docs/concepts/services-networking/service/#aws-load-balancer-additional-resource-tags

The operator will only add the annotation on create; there is no
ongoing reconciliation so changes to `AWS.UserTags` will not propagate
once the LoadBalancer service has been created.

There are restrictions on the key and values, documented here:
 https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html

Resources involved for classic (CLB) and network (NLB) differ.

# Atomic create/tag issues on CLBs:

When a CLB is created (w.r.t. tags) the following happens:

- the load balancer is created with tags (atomic create/tag)

then:

- the security groups are created
- then the security groups are updated with tag keys/values

# Atomic create/tag issues on NLBs:

When an NBL is created (w.r.t. tags) the following happens:

- the load balancer is created with tags (atomic create/tag)

then:

- a target group is created
- tags are added to the target group
- a listener is created
- *NO* tags are created for the listener (BUG?)

Given that we only tag on create the following issues will not bite
immediately, but capturing them here for completeness:

- https://github.com/kubernetes/kubernetes/issues/95106
- https://github.com/kubernetes/kubernetes/pull/96939


frobware:
/hold

Accumulating things to do as I discover them and also to provide a link for the Jira ticket.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -25,6 +26,13 @@ import (
 )
 
 const (
+	// awsLBAdditionalResourceTags is a comma separated list of
+	// Key=Value pairs that are additionally recorded on classic
+	// load balancer resources.
```
Miciah, line 31: ...and any security groups that we create for them.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -255,8 +263,18 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
 			} else {
+				var additionalTags []string
+				if platformSpec != nil && platformSpec.AWS != nil {
+					for _, userTag := range platformSpec.AWS.UserTags {
+						if len(userTag.Key) > 0 {
+							additionalTags = append(additionalTags, userTag.Key+"="+userTag.Value)
+						}
+					}
+				}
+				service.Annotations[awsLBAdditionalResourceTags] = strings.Join(additionalTags, ",")
```
Miciah, line 274: Does it make sense to set the annotation if `additionalTags` is empty?

Miciah, line 274: Doesn't the annotation work for NLBs as well?  Looking at the cloud provider code, I thought it did.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -394,6 +412,11 @@ func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev
 		changed = true
 	}
 
+	if current.Annotations[awsLBAdditionalResourceTags] != expected.Annotations[awsLBAdditionalResourceTags] {
```
sgreene570, line 415: nit: Add a comment explaining that we can safely reconcile these changes (See lines 396-401). Alternatively you could update the above comment.

pkg/operator/controller/ingress/controller.go
```diff
@@ -663,7 +670,7 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	var lbService *corev1.Service
 	var wildcardRecord *iov1.DNSRecord
-	if haveLB, lb, err := r.ensureLoadBalancerService(ci, deploymentRef, infraConfig); err != nil {
+	if haveLB, lb, err := r.ensureLoadBalancerService(ci, deploymentRef, infraConfig, &infraConfig.Spec.PlatformSpec); err != nil {
```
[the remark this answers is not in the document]
frobware: This is a stop-gap; I was waiting to see where the usertags field would actually end up in openshift/api#864.

pkg/operator/controller/ingress/load_balancer_service.go
frobware: > Does it make sense to set the annotation if `additionalTags` is empty?

I was wondering about that...

If you previously set "a=b,c=d" and this time around there are no tags then setting an empty value doesn't help here because it doesn't mean that "a=b,c=d" will be deleted.

I will change this to not set the annotation if `additionalTags` is empty.

frobware: I need to check. I just wanted something end-2-end at the moment to flesh out unknowns.

pkg/operator/controller/ingress/load_balancer_service.go
frobware: > nit: Add a comment explaining that we can safely reconcile these changes (See lines 396-401). Alternatively you could update the above comment.

Waiting on EP to see if we need to reconcile or only at creation time as proposed; https://github.com/openshift/enhancements/pull/706/commits/c0ccefa90fe8ac96607ac108adbbdf83874e3228#diff-9e7c9d80b69a46de2319af1610f54e1c0f04896901283d2e878bef4ecf915240R57

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
frobware: See also: https://github.com/openshift/enhancements/pull/706/commits/c0ccefa90fe8ac96607ac108adbbdf83874e3228#r600952570

pkg/operator/controller/ingress/load_balancer_service.go
frobware: NLBs work; updated.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
frobware: I temporarily vendored openshift/api#864. 

Still need to see whether user tags will end up in spec or status (and that only matters for `desiredLoadBalancerService` anyway, so dropped this unnecessary change and moving less to WIP.

frobware:
/hold cancel

Ready for review now.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -261,6 +269,17 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 			} else {
 				service.Annotations[awsLBHealthCheckIntervalAnnotation] = awsLBHealthCheckIntervalDefault
 			}
+
+			if platform != nil && platform.AWS != nil && len(platform.AWS.ResourceTags) > 0 {
```
sgreene570, line 273:
```suggestion
			if platform.AWS != nil && len(platform.AWS.ResourceTags) > 0 {
```

The `platform != nil` check here is redundant, right?

frobware: Yes, we're guarded by the block "above". Removed.

frobware: Done in https://github.com/openshift/cluster-ingress-operator/pull/578/commits/154643c1444bbbe4c7e68675680cc397b59d8f3e.

frobware:
Gathering artifacts ...
error: the server doesn't have a resource type "nodes"

/retest

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: This still sets the annotation if `platform.AWS.ResourceTags` is nonempty but every entry has an empty key, right?  Maybe that edge-case isn't worth worrying about.

Miciah (commented):
/lgtm
/hold
in case you want to address https://github.com/openshift/cluster-ingress-operator/pull/578#discussion_r616421539, but I don't mind merging as-is if you prefer.

Miciah:
We talked about incorporating #508 into this PR, but doing so would introduce merge conflicts during backports, so lets deal with #508 separately (including backports, if they are needed for #508).

frobware:
> in case you want to address [#578 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/578#discussion_r616421539), but I don't mind merging as-is if you prefer.

Thanks for catching this. Done in af1839e7a6cd5b36b5b46d928f98ef4ee5de7424.

Miciah:
failed to acquire lease: resources not found

/test e2e-aws-operator

frobware:
failed to acquire lease: resources not found

/test e2e-gcp-serial

frobware:
failed to acquire lease: resources not found

/test e2e-gcp-serial

PR #580, opened 2021-03-25
NE-302: Add field for configuring number of HAProxy threads in router

rfredette:
ingress operator reads the IngressController's `spec.tuningOptions.threadCount` field, and sets the `"ROUTER_THREADS"` environment variable in the desired router deployment according to its value.

this is dependent on [openshift/api #875](https://github.com/openshift/api/pull/875), and part of the implementation of [openshift/enhancements #708](https://github.com/openshift/enhancements/pull/708)

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -390,7 +390,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: "ROUTER_USE_PROXY_PROTOCOL", Value: "true"})
 	}
 
-	env = append(env, corev1.EnvVar{Name: "ROUTER_THREADS", Value: "4"})
+	threads := 4
+	if ci.Spec.TuningOptions.ThreadCount > 0 {
+		threads = int(ci.Spec.TuningOptions.ThreadCount)
+	}
+	env = append(env, corev1.EnvVar{Name: "ROUTER_THREADS", Value: strconv.Itoa(threads)})
 
 	nodeSelector := map[string]string{
 		"kubernetes.io/os":               "linux",
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -66,6 +66,9 @@ const (
 	RouterHardStopAfterAnnotation = "ingress.operator.openshift.io/hard-stop-after"
 
 	LivenessGracePeriodSecondsAnnotation = "unsupported.do-not-use.openshift.io/override-liveness-grace-period-seconds"
+
+	RouterHAProxyThreadsEnvName      = "ROUTER_THREADS"
+	RouterHAProxyThreadsDefaultValue = 4
 )
 
 // ensureRouterDeployment ensures the router deployment exists for a given
@@ -390,11 +393,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: "ROUTER_USE_PROXY_PROTOCOL", Value: "true"})
 	}
 
-	threads := 4
+	threads := RouterHAProxyThreadsDefaultValue
 	if ci.Spec.TuningOptions.ThreadCount > 0 {
 		threads = int(ci.Spec.TuningOptions.ThreadCount)
 	}
-	env = append(env, corev1.EnvVar{Name: "ROUTER_THREADS", Value: strconv.Itoa(threads)})
+	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
 	nodeSelector := map[string]string{
 		"kubernetes.io/os":               "linux",
--- a/test/e2e/nbthread_test.go
+++ b/test/e2e/nbthread_test.go
@@ -1,0 +1,83 @@
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const nbthreadRouterDeployTimeout time.Duration = 1 * time.Minute
+
+// NOTE: This test will mutate the default ingresscontroller
+func TestRouteNbthreadIngressController(t *testing.T) {
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	ic, err := getIngressController(t, kclient, defaultName, 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get ingress controller: %v", err)
+	}
+
+	routerDeployment, err := getDeployment(t, kclient, controller.RouterDeploymentName(ic), 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get router deployment: %v", err)
+	}
+
+	// by default, the router deployment should have ROUTER_THREADS set to 4
+	if err := waitForDeploymentEnvVar(t, kclient, routerDeployment, nbthreadRouterDeployTimeout, "ROUTER_THREADS", "4"); err != nil {
+		t.Fatalf("expected router deployment to have spec.threading.count unset: %v", err)
+	}
+
+	// set spec.threading.count to a nonstandard value, 7
+	if err := nbthreadSetThreadCount(t, kclient, 1*time.Minute, 7, ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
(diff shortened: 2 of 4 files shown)


test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	// when set back to default, the router deployment should have ROUTER_THREADS set to 4
+	if err := nbthreadWaitForRouterDeployment(t, kclient, nbthreadRouterDeployTimeout, ic, 4); err != nil {
+		t.Fatalf("expected router deployment to have spec.threading.count unset: %v", err)
+	}
+}
+
+func nbthreadGetIngressController(t *testing.T, client client.Client, timeout time.Duration) (*operatorv1.IngressController, error) {
```
Miciah, line 68: Most of these helpers look like they would be generally useful in other tests.  Any reason not to put them in `util.go` with more general names?  At some point, we can update other test code to use the helpers.  

Also, consider calling `t.Helper()` at the start of the function, at least for this and the other general helpers, so that logging and error output references the test that called the helper.

rfredette: True, there's no reason these can't be generic helper functions. I'll move them to `util.go`.

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,128 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const nbthreadRouterDeployTimeout time.Duration = 1 * time.Minute
+
+// NOTE: This test will mutate the default ingresscontroller and the
+// "cluster" ingress configuration.
```
sgreene570, line 28: Is this e2e test actually modifying the cluster ingress config?

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,128 @@
...
+func nbthreadVerifyRouterDeployment(deployment *appsv1.Deployment, nbthreads int) bool {
+	expectedThreads := nbthreads
+	if expectedThreads == 0 {
+		expectedThreads = 4
+	}
+	foundThreads := nbthreadGetThreadsFromEnv(deployment.Spec.Template.Spec.Containers[0].Env)
+	return foundThreads == expectedThreads
+}
+
+func nbthreadGetThreadsFromEnv(env []corev1.EnvVar) int {
```
sgreene570, line 119: Would `waitForDeploymentEnvVar` work in place of these new functions?

https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L1864

pkg/operator/controller/ingress/deployment.go
```diff
@@ -366,7 +366,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		env = append(env, corev1.EnvVar{Name: "ROUTER_USE_PROXY_PROTOCOL", Value: "true"})
 	}
 
-	env = append(env, corev1.EnvVar{Name: "ROUTER_THREADS", Value: "4"})
+	threads := 4
+	if ci.Spec.Threading.Count > 0 {
+		threads = ci.Spec.Threading.Count
+	}
+	env = append(env, corev1.EnvVar{Name: "ROUTER_THREADS", Value: strconv.Itoa(threads)})
```
sgreene570, line 373: Could you add unit test cases to cover this code? Should be pretty straightforward. See https://github.com/openshift/cluster-ingress-operator/pull/576 for an example.

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Fatalf("failed to get ingress controller: %v", err)
+	}
+
+	// by default, the router deployment should have ROUTER_THREADS set to 4
+	if err := nbthreadWaitForRouterDeployment(t, kclient, nbthreadRouterDeployTimeout, ic, 4); err != nil {
+		t.Fatalf("expected router deployment to have spec.threading.count unset: %v", err)
+	}
+
+	// set spec.threading.count to a nonstandard value, 7
+	if err := nbthreadSetThreadCount(t, kclient, 1*time.Minute, 7, ic); err != nil {
```
frobware, line 44: Perhaps we should generally define `4` as a constant in the deployment.go then do `2 * _DEFAULT_` just in case we ever change the "default" to 7 - we (or future "we") wouldn't notice any regression in that case.

frobware (approved):
LGTM overall - and depending how the week pans out we should consider just merging this (once @sgreene570's unit test case is covered) and do any other follow-ups post FF.

test/e2e/nbthread_test.go
rfredette: done, thanks for the suggestion

pkg/operator/controller/ingress/deployment.go
rfredette: I added some checks for `ROUTER_THREADS` to the existing `TestDesiredRouterDeployment` test case to cover this

test/e2e/nbthread_test.go
rfredette: That is a good fit, thanks!

test/e2e/nbthread_test.go
rfredette: It is not; it only modifies the default ingresscontroller. The comment has been updated

rfredette:
@Miciah PTAL

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
```
sgreene570, line 14: nit: name this import `ingresscontroller`

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,87 @@
...
+	routerDeployment, err := getDeployment(t, kclient, controller.RouterDeploymentName(ic), 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get router deployment: %v", err)
+	}
+
+	defaultThreadsStr := strconv.Itoa(ingress.RouterHAProxyThreadsDefaultValue)
+
+	// by default, the router deployment should have ROUTER_THREADS set to 4
+	if err := waitForDeploymentEnvVar(t, kclient, routerDeployment, nbthreadRouterDeployTimeout, "ROUTER_THREADS", defaultThreadsStr); err != nil {
+		t.Fatalf("expected router deployment to have spec.threading.count unset: %v", err)
```
sgreene570, line 44:
```suggestion
		t.Fatalf("expected router deployment to have ROUTER_THREADS env var set to %s: %v", defaultThreadsStr, err)
```

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,87 @@
...
+	if err := nbthreadSetThreadCount(t, kclient, 1*time.Minute, ingress.RouterHAProxyThreadsDefaultValue*2, ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, defaultName, defaultAvailableConditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	if err := waitForDeploymentEnvVar(t, kclient, routerDeployment, nbthreadRouterDeployTimeout, "ROUTER_THREADS", strconv.Itoa(ingress.RouterHAProxyThreadsDefaultValue*2)); err != nil {
+		t.Fatalf("expected router deployment to set spec.threading.count: %v", err)
```
sgreene570, line 57: Same comment as line 44

test/e2e/nbthread_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// set threading.count to 0, as if it was unset
+	if err := nbthreadSetThreadCount(t, kclient, 1*time.Minute, 0, ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	// when set back to default, the router deployment should have ROUTER_THREADS set to 4
+	if err := waitForDeploymentEnvVar(t, kclient, routerDeployment, nbthreadRouterDeployTimeout, "ROUTER_THREADS", defaultThreadsStr); err != nil {
+		t.Fatalf("expected router deployment to have spec.threading.count unset: %v", err)
```
sgreene570, line 67: same comment as line 44

test/e2e/nbthread_test.go
rfredette: done

test/e2e/nbthread_test.go
rfredette: Good catch! These have been fixed.

sgreene570: The router deployment should always have some value for `ROUTER_THREADS` set, right?
The new error messages imply that there are cases where the router deployment _should not_ have any corresponding value for `ROUTER_THREADS`.

rfredette: Correct, `ROUTER_THREADS` should always be set, even if `spec.tuningOptions.threadCount` isn't. Let me rework the error message for those cases

rfredette: Updated

Miciah:
Spiffy!  Thanks!  
/lgtm

sgreene570:
> Spiffy! Thanks!
> /lgtm

Plus 1

sgreene570:
One more to go! 
/test e2e-aws-operator

Miciah:
One of the workers failed to become schedulable.  

/test e2e-aws-operator

sgreene570:
OLM was firing alerts

/test e2e-upgrade

Miciah:
Just failures related to pod sandbox creation and Multus this time.  So close!

/test e2e-upgrade

sgreene570:
:partying_face:

PR #514, opened 2020-12-15
Bug 1906560: Revert "Support changing ingresscontroller load balancer scope"

Miciah:
This reverts #472.

----

This is the backup plan if https://github.com/openshift/api/pull/822 / https://github.com/openshift/cluster-ingress-operator/pull/510 cannot be merged in 4.7.

PR #582, opened 2021-03-27
NE-621: Support changing ingresscontroller load balancer scope

Miciah:
Add support for changing ingresscontroller load balancer scope. On AWS and IBM cloud, this requires deleting the existing load balancer service and recreating it with the desired scope; if the administrator changes the scope on the ingresscontroller, the operator will report `Progressing=True` until the administrator either reverts the change or deletes the service. If the administrator deletes the service, then the operator will recreate the service with the desired scope.  On Azure and GCP, it suffices to update the existing service's annotations.

This commit is based on #472, which was reverted by commit #514.  Unlike the original implementation of this feature, the operator does not delete the service but instead leaves it up to the administrator to do so as needed.

* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Update scope if needed.
(`ensureIngressController`): Pass `infraConfig` to `syncIngressControllerStatus`.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`externalLBAnnotations`): New variable.  Map platform type to the annotation for that platform that makes the load balancer external, if the platform requires an explicit annotation.
(`desiredLoadBalancerService`): Use new `externalLBAnnotations` variable to simplify logic.
(`updateLoadBalancerService`): Pass `platform` to `loadBalancerServiceScopeChanged`.
(`loadBalancerServiceChanged`): Add `platform` parameter.  Check if the scope changed and if the platform supports changing scope without recreating the service, and update the appropriate annotations if so.
(`loadBalancerServiceScopeChanged`): New function.  Check if the load balancer's scope changed.
(`IsServiceInternal`): New function.  Return a Boolean value indicating whether the provided service is annotated to request an internal load-balancer
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestLoadBalancerServiceChanged`): Update to pass platform status to `loadBalancerServiceChanged`.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Add `infraConfig` parameter.  Use `infraConfig` to get the platform type. Call the new `computeIngressProgressingCondition` function with the ingresscontroller, service, and platform to compute a "Progressing" status condition for the ingresscontroller.
(`computeIngressProgressingCondition`): New function.  Compute a "Progressing" status condition.  In particular, this status condition will indicate if it is necessary to delete the service so that it can be recreated with the updated scope.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressProgressCondition`): New test.  Verify that `computeIngressProgressingCondition` returns the expected status condition.
* `pkg/operator/controller/status/controller.go` (`Reconcile`): Pass ingresscontrollers to `computeOperatorProgressingCondition`.
(`computeOperatorProgressingCondition`): Report `Progressing=True` on the clusteroperator if any ingresscontroller reports `Progressing=True`.
* `pkg/operator/controller/status/controller_test.go` (`TestComputeOperatorProgressingCondition`): Add test case where an ingresscontroller is progressing.
* `test/e2e/operator_test.go` (`TestScopeChange`): New test.  Verify that the operator performs the appropriate behavior for the platform on which the test is running when an ingresscontroller's scope is changed from the default external scope to internal and then back to external.

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -183,6 +183,52 @@ var (
 			iksLBScopeAnnotation: iksLBScopePublic,
 		},
 	}
+
+	// managedLoadBalancerServiceAnnotations is a set of annotation keys for
+	// annotations that the operator manages for LoadBalancer-type services.
+	// The operator preserves all other annotations.
+	//
+	// Be careful when adding annotation keys to this set.  If a new release
+	// of the operator starts managing an annotation that it previously
+	// ignored, it could stomp annotations that the user has set when the
+	// user upgrades the operator to the new release (see
+	// <https://bugzilla.redhat.com/show_bug.cgi?id=1905490>).  In order to
+	// avoid problems, make sure the previous release blocks upgrades when
+	// the user has modified an annotation that the new release manages.
+	managedLoadBalancerServiceAnnotations = func() sets.String {
+		result := sets.NewString(
+			// AWS LB health check interval annotation (see
+			// <https://bugzilla.redhat.com/show_bug.cgi?id=1908758>).
+			awsLBHealthCheckIntervalAnnotation,
+			// GCP Global Access internal Load Balancer annotation
+			// (see <https://issues.redhat.com/browse/NE-447>).
+			GCPGlobalAccessAnnotation,
+			// local-with-fallback annotation for kube-proxy (see
+			// <https://bugzilla.redhat.com/show_bug.cgi?id=1960284>).
+			localWithFallbackAnnotation,
+		)
+
+		// Azure and GCP support switching between internal and external
+		// scope by changing the annotation, so the operator manages the
+		// corresponding load-balancer scope annotations for these
+		// platforms.  Other platforms require deleting and recreating
+		// the service, so the operator doesn't update the annotations
+		// that specify load-balancer scope for those platforms.  See
+		// <https://issues.redhat.com/browse/NE-621>.
+		for _, platform := range []configv1.PlatformType{
+			configv1.AzurePlatformType,
+			configv1.GCPPlatformType,
+		} {
+			for name := range InternalLBAnnotations[platform] {
+				result.Insert(name)
+			}
+			for name := range externalLBAnnotations[platform] {
+				result.Insert(name)
+			}
+		}
+
+		return result
+	}()
 )
 
 // ensureLoadBalancerService creates an LB service if one is desired but absent.
@@ -478,17 +524,17 @@ func (r *reconciler) updateLoadBalancerService(current, desired *corev1.Service,
 	return true, nil
 }
 
-// managedLoadBalancerServiceAnnotations is a set of annotation keys for
-// annotations that the operator manages for LoadBalancer-type services.
-var managedLoadBalancerServiceAnnotations = sets.NewString(
-	awsLBHealthCheckIntervalAnnotation,
-	GCPGlobalAccessAnnotation,
-	localWithFallbackAnnotation,
-)
-
 // loadBalancerServiceChanged checks if the current load balancer service
 // matches the expected and if not returns an updated one.
 func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev1.Service) {
+	// Preserve most fields and annotations.  If a new release of the
+	// operator starts managing an annotation or spec field that it
+	// previously ignored, it could stomp user changes when the user
+	// upgrades the operator to the new release (see
+	// <https://bugzilla.redhat.com/show_bug.cgi?id=1905490>).  In order to
+	// avoid problems, make sure the previous release blocks upgrades when
+	// the user has modified an annotation or spec field that the new
+	// release manages.
 	annotationCmpOpts := []cmp.Option{
 		cmpopts.IgnoreMapEntries(func(k, _ string) bool {
 			return !managedLoadBalancerServiceAnnotations.Has(k)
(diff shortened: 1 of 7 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -538,6 +544,45 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+			wantScope := ic.Status.EndpointPublishingStrategy.LoadBalancer.Scope
+			haveScope := operatorv1.ExternalLoadBalancer
+			if IsServiceInternal(service) {
+				haveScope = operatorv1.InternalLoadBalancer
+			}
+			if wantScope != haveScope {
+				message := fmt.Sprintf("Have load balancer with scope %q, want load balancer with scope %q.", haveScope, wantScope)
+				switch platform.Type {
+				case configv1.AWSPlatformType, configv1.IBMCloudPlatformType:
+					message = fmt.Sprintf("%s  You can delete the %s/%s service to proceed, and the service load-balancer will be deprovisioned and recreated.", message, service.Namespace, service.Name)
```
sgreene570, line 575: I think as is, this message could be confusing to a cluster administrator:

1) I think `recreated` in this context sort of implies that the new load-balancer will have the same hostname and IP as the last load-balancer. Should we explicitly state that when deleting the load-balancer service, a new load-balancer will be stood up that will most likely have a new hostname and IP?

2) Should we make it clear to the cluster administrator that they are free to "undo" their scope change without any reprocussions? I think some more details in general could be useful here as to not confuse a cluster administrator.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -398,9 +415,86 @@ func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev
...
+	for name := range InternalLBAnnotations[platform.Type] {
+		if currentAnnotations[name] != expectedAnnotations[name] {
+			return true
+		}
+	}
+	return false
+}
+
+// IsServiceInternal returns a Boolean indicating whether the provided service
+// Is annotated to request an internal load balancer.
```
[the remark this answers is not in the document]
Miciah: Thanks!  Fixed.

pkg/operator/controller/ingress/status.go
```diff
@@ -44,12 +44,17 @@ type expectedCondition struct {
...
-func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS) error {
+func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) error {
 	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
 	if err != nil {
 		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
 	}
 
+	platform, err := oputil.GetPlatformStatus(r.client, infraConfig)
+	if err != nil {
+		return fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %w", ic.Namespace, ic.Name, err)
```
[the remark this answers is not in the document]
Miciah: Yeah, IMO we should use `%w` in new code, and eventually update old code to use `%w` as well, so that we can unwrap errors and more easily check the type of an error (see https://blog.golang.org/go1.13-errors).  Eventually, consistent use of `%w` will enable us to simplify some error-handling code (e.g., the retryable-error logic).

pkg/operator/controller/ingress/status.go
Miciah: I've updated the message to convey these points.  Does it look all right?  (I hope it isn't _too_ verbose.)

Miciah:
Rebased.

pkg/operator/controller/ingress/status.go
sgreene570: Looks fine to me :grin:

Miciah:
Rebased with  7daeb6f → 37adb49.  I still need to make some changes based on feedback to the enhancement.

Miciah:
`TestIngressControllerScale` and `TestLocalWithFallbackOverrideForLoadBalancerService` failed, but `TestScopeChange` passed in [the last e2e-aws-operator run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/582/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1468750853321003008), indicating that the new functionality works on AWS (which requires deleting and recreating the service to change its scope).  Let's try e2e-gcp-operator to find out whether the failing tests pass on a second try and also find out whether the new functionality works on GCP (which *doesn't* require deleting and recreating the service to change its scope).  

/test e2e-gcp-operator

Miciah:
/retitle NE-621: Support changing ingresscontroller load balancer scope

Miciah:
/hold
CI is failing because of https://github.com/openshift/build-machinery-go/issues/56.

Miciah:
Must-gather failed.
/test e2e-aws-operator

Miciah:
`TestConfigurableRouteNoConsumingUserNoRBAC` and `TestRouteAdmissionPolicy` failed.
/test e2e-aws-operator

Miciah:
Unschedulable worker node.  
/test e2e-aws-operator

PR #586, opened 2021-03-30
Add the new NetworkPolicy-matching label to the namespace

danwinship:
while implementing the ["allow-from-router" feature](https://github.com/openshift/enhancements/blob/master/enhancements/network/allow-from-router-networkpolicy.md), we realized you can't have both `network.openshift.io/policy-group: ingress` and `network.openshift.io/policy-group: host-network` labels on the same namespace because they have the same name. So we're changing the host-network label to `policy-group.network.openshift.io/host-network: ""`, and figure we should change the ingress label to match. (But keeping the old form for backward compatibility.)

cc @abhat


Miciah:
You need to run `make generate` to regenerate `bindata.go`.

danwinship:
oops, fixed

PR #587, opened 2021-03-30
Bug 1948716: IBM cloud managed profile patch

csrwng:
Part 2 of enabling the IBM cloud managed profile.
This profile requires an alternate version of the operator's deployment manifest that does not have a master nodeselector. 

This PR vendors a common include from github.com/openshift/build-machinery-go that adds 'update' and 'verify' targets that handle applying a patch and keeping the alternate manifest up to date.


Makefile
```diff
@@ -34,6 +41,9 @@ generate: crd bindata
 bindata:
 	hack/update-generated-bindata.sh
 
+.PHONY: update
+update: bindata
```
sgreene570, line 45: I'm wondering if we should switch over our current `generate` target to just be `update`.
@Miciah thoughts? 
```suggestion
update: crd bindata
```

Miciah: Is `make update` more common than `make generate`?  I'd be fine with that.  We could keep `make generate` as an alias (possibly with a reminder to change to `make update`).

csrwng: @Miciah @sgreene570 I changed `update` and made `generate` simply invoke `update`. I also vendored the latest code from `build-machinery-go` which adds a comment at the top of the patched manifest to make it clear that it is a generated file.

Miciah: @csrwng, I'm really nervous about pulling in build-machinery-go, with its byzantine Makefile definitions, lack of documentation, and its use of Curl to download and run an old, deprecated binary off of GitHub, just to regenerate or verify the working tree (which shouldn't require network access to do at all).  If we merge this today, can we eventually replace build-machinery-go with something reasonable?  

Also, can you help me understand why Make doesn't complain about the duplicate definitions of the `update` and `verify` targets (first by `add-profile-manifests` and then explicitly here)?  From the manual, it seems like this should cause a warning:  https://www.gnu.org/software/make/manual/html_node/Multiple-Rules.html

Miciah: > Also, can you help me understand why Make doesn't complain about the duplicate definitions of the `update` and `verify` targets (first by `add-profile-manifests` and then explicitly here)?  From the manual, it seems like this should cause a warning:  https://www.gnu.org/software/make/manual/html_node/Multiple-Rules.html 

Ah, sorry, never mind—having multiple rules for the same target is fine in this case because none of the rules defines a recipe.

csrwng: @Miciah the main reasoning behind using build-machinery-go is that many projects under the openshift org already use it and there was a desire to not maintain separate code/scripts wherever we needed to process patches. Keep in mind that we're only importing the part of it that allows these patches to work. However, I had initially proposed an [implementation of this that used bash + the oc client](https://github.com/csrwng/cluster-network-operator/commit/69339a219db0c5beed69ab4f514a2aefb1fec9b1). If more acceptable, we can eventually replace the build-machinery-go vendor with it. I've already implemented different logic in other repos. For example in monitoring they like to do everything with jsonnet, and the patch there is implemented as a [jsonnet library](https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/ibm-cloud-managed-profile.libsonnet). Ultimately this is your repo and I will work with you to get to something you feel comfortable with.

Miciah: Thanks for presenting these options!  The Bash + oc solution looks nice—`oc patch --local` is nifty, although I'm still nervous about the Curl command, but at least it is from an OpenShift mirror, and presumably the oc executable is generally already installed, so the download logic will not trigger anyway.  The jsonnet approach looks very patch-as-procedure, which seems inconvenient, but maybe it's just my lack of familiarity with jsonnet.  I'll raise these options with my team so we can decide where we want to go.  For now, if build-machinery-go works and other teams are already using it, let's proceed with this PR as it is so we can get it merged in time.  Thanks again!

Miciah:
Need to rebase for https://github.com/openshift/cluster-ingress-operator/pull/595.

csrwng:
Rebased

Miciah:
Thanks!
/lgtm

Miciah:
This one seems to be a frequent flake on this and other PRs:

    : [sig-auth][Feature:OpenShiftAuthorization][Serial] authorization TestAuthorizationResourceAccessReview should succeed [Suite:openshift/conformance/serial]
    fail [github.com/openshift/origin/test/extended/authorization/authorization.go:375]: who can view deploymentconfigs in hammer by harold
    : evaluation errors does not match (clusterrole.rbac.authorization.k8s.io "machine-api-operator-ext-remediation" not found!=)

Makefile
Miciah: https://github.com/openshift/cluster-ingress-operator/pull/687 replaces build-machinery-go with `oc patch --local`.

PR #597, opened 2021-04-12
Bug 1941592: Alerts: Fix up HAProxyDown Alert Message

sgreene570:
Make the warning message for the `HAProxyDown` better by clarifying
that the alert only pertains to HAProxy instances running on router
pods, and not the status of router pod's themselves.

This resolves Bug 1941592.

It is most likely not worth the hassle of back-porting this change.


candita:
/lgtm

I would just make sure it gets a little mention in the docs in case anyone has alert monitoring based on a specifically formatted string.

PR #598, opened 2021-04-12
Bug 1948626: test/e2e: Make `TestRouteAdmissionPolicy` deterministic.

sgreene570:
This commit adds a few small e2e test improvements in support of Bug
1948626.

test/e2e/operator_test.go:

Helper functions:
Increase the polling timeouts in `waitForRouterIngressConditions` and
`waitForClusterOperatorConditions` from 10 seconds to 1 minute.
1 minute is not much longer than 10 seconds, and presumably a longer
timeout for these polling functions could result in less test flakes.
It's worth taking the additional 50 seconds if it saves us from any
future `/retest` comments.

`TestRouteAdmissionPolicy`:
Use specific names for `route1` and `route2`, as to avoid the common
`route` name, which makes debugging this e2e test much easier.
Sleep for 2 (two) seconds after creating route1. This is to ensure that
route1 and route2 _do not_ wind up having the same creation timestamp.
If route1 and route2 have a common timestamp, we cannot determine
whether route1 or route2 will be rejected.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -948,10 +948,11 @@ func TestRouteAdmissionPolicy(t *testing.T) {
 			},
 		}
 	}
-	route1Name := types.NamespacedName{Namespace: ns1.Name, Name: "route"}
+	// use unique names for each test route to simplify debugging.
+	route1Name := types.NamespacedName{Namespace: ns1.Name, Name: "route1"}
 	route1 := makeRoute(route1Name, "routeadmission.test.example.com", "/foo")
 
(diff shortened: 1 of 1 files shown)


Miciah:
Would it be just as well to verify that exactly 1 route is admitted with `spec.routeAdmission.namespaceOwnership: String` and that exactly 2 routes are admitted with `spec.routeAdmission.namespaceOwnership: InterNamespaceAllowed`?

Miciah:
Thanks!  A 2-second sleep _should_ make it extremely unlikely that the two routes will end up with the same creation timestamp.  
/lgtm

PR #599, opened 2021-04-15
Bug 1949880: ingress: Propagate GCP ProviderParameters to Status

sgreene570:

Fix `setDefaultPublishingStrategy` so that updates to an Ingress
Controller's `spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP`
field are propagated to `status` correctly. Prior to this commit, GCP
provider parameter changes were only propagated to `status` if the
Ingress Controller was originally created with GCP provider parameters.

This commit resolves Bug 1949880, in which setting the GCP Global Access
provider parameter for an Ingress Controller created prior to an upgrade
resulted in no Load Balancer Service updates.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -389,10 +389,21 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig 
 	// Detect changes to GCP LB provider parameters, which is something we can safely roll out.
 	statusLB := ic.Status.EndpointPublishingStrategy.LoadBalancer
 	specLB := effectiveStrategy.LoadBalancer
-	if specLB != nil && statusLB != nil && specLB.ProviderParameters != nil && statusLB.ProviderParameters != nil &&
-		specLB.ProviderParameters.GCP != statusLB.ProviderParameters.GCP {
-		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP = effectiveStrategy.LoadBalancer.ProviderParameters.GCP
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -389,10 +389,20 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig
...
 	specLB := effectiveStrategy.LoadBalancer
-	if specLB != nil && statusLB != nil && specLB.ProviderParameters != nil && statusLB.ProviderParameters != nil &&
-		specLB.ProviderParameters.GCP != statusLB.ProviderParameters.GCP {
-		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.GCP = effectiveStrategy.LoadBalancer.ProviderParameters.GCP
-		return true
+	if specLB != nil && statusLB != nil {
+		// If the ProviderParameters field does not exist for spec or status,
+		// just propagate (or remove) ProviderParameters in it's entirety.
+		if specLB.ProviderParameters == nil && statusLB.ProviderParameters != nil ||
+			specLB.ProviderParameters != nil && statusLB.ProviderParameters == nil {
```
rfredette, line 396: Is there any scenario where `specLB.ProviderParameters` will be nil but `statusLB.ProviderParameters` will be non-nil?

sgreene570: Yes, in the case that a user deletes the provider parameters field from spec (which is feasible if they no longer want to specify any GCP Global Access setting).

PR #601, opened 2021-04-16
Change status message to have a capital d in desired 

bryanro92:
Change status message to have a capital d in 'desired' to be consistent with other openshift operators 'ok' message. 

Looks better in my opinion, and as mentioned adds consistency in first char capitalization among operators status message.


bryanro92:
test flakes? can't imagine this pr would be causing it to fail.

PR #607, opened 2021-05-01
Bug 1955854: Compute Available and Degraded from default ingress

Miciah:
#### status: Combine status condition merging

* `pkg/operator/controller/status/controller.go` (`Reconcile`): Replace 3 separate invocations of `mergeConditions` with a single invocation.

#### canary: Delete extra "on" in log message

* `pkg/operator/controller/canary/metrics.go` (`StartMetricsListener`): Delete trailing "on" in the "starting metrics listener on " log message.

#### alerts: Add ingresscontroller degraded/unavailable

Add an `ingress_controller_conditions` Prometheus metric that reports the status conditions of ingresscontrollers, and add Prometheus rules to raise alerts if an ingresscontroller is unavailable or degraded.

* `cmd/ingress-operator/start.go` (`start`): Call `StartMetricsListener` from the operator package, and call `RegisterMetrics` from the ingress and canary controller packages.
* `manifests/0000_90_ingress-operator_03_prometheusrules.yaml`: Add alerts using the new `ingress_controller_conditions` metric to warn if an ingresscontroller is degraded or unavailable.
* `pkg/operator/controller/canary/metrics.go` (`registerCanaryMetrics`): Rename from this...
(`RegisterMetrics`): ...to this.
(`StartMetricsListener`): Move from here...
* `pkg/operator/metrics.go`: ...to here.  New file.
* `pkg/operator/controller/ingress/metrics.go` (`ingressControllerConditions`): New variable.  Define a `ingress_controller_conditions` Prometheus gauge.
(`metricsList`): New variable.  Define the list of metrics for this controller.  Currently this list comprises `ingressControllerConditions`.
(`reportedConditions`): New variable.  Define the ingresscontroller status conditions that are published in the `ingress_controller_conditions` metric.
(`SetIngressControllerConditionsMetric`): New function.  Update the new `ingress_controller_conditions` gauge with the status conditions of the given ingresscontroller.
(`RegisterMetrics`): New function.  Register `metricsList` with Prometheus.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Call `SetIngressControllerConditionsMetric` after updating status conditions.

#### status: Set Available/Degraded from default ingress

Use only the default ingresscontroller for computing the clusteroperator's `Available` and `Degraded` status conditions.  The status of other ingresscontrollers can be monitored using alerts; the clusteroperator should only report status as required for basic cluster functionality, and core components should depend only on the default ingresscontroller.

* `pkg/operator/controller/status/controller.go` (`Reconcile`): Pass `state.IngressControllers` to `computeOperatorStatusVersions`.
(`computeOperatorDegradedCondition`, `computeOperatorAvailableCondition`): Use only the default ingresscontroller to compute status conditions.


cmd/ingress-operator/start.go
```diff
@@ -123,7 +124,15 @@ func start(opts *StartOptions) error {
 	}
 
 	// Start operator metrics.
-	go canarycontroller.StartMetricsListener(opts.MetricsListenAddr, signal)
+	go operator.StartMetricsListener(opts.MetricsListenAddr, signal)
+	log.Info("registering Prometheus metrics for canary_controller")
+	if err := canarycontroller.RegisterMetrics(); err != nil {
+		log.Error(err, "unable to register metrics")
```
frobware, line 130: Would be useful to indicate which registration failed as we have two cases.

pkg/operator/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+	ctrlruntimemetrics.Registry.Unregister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+
+	// Create prometheus handler by combining the ingress-operator registry
+	// with the ingress-operator's controller runtime metrics registry.
+	handler := promhttp.HandlerFor(
+		prometheus.Gatherers{prometheus.DefaultGatherer, ctrlruntimemetrics.Registry},
+		promhttp.HandlerOpts{},
+	)
+
+	log.Info("starting metrics listener on ", "addr", addr)
```
frobware, line 29: nit: do you have to provide trailing whitespace on message?

pkg/operator/controller/status/controller.go
```diff
@@ -288,40 +286,50 @@ func checkAllIngressesAvailable(ingresses []operatorv1.IngressController) bool {
...
-	}
-	if len(degradedIngresses) == 0 {
-		return configv1.ClusterOperatorStatusCondition{
-			Type:   configv1.OperatorDegraded,
-			Status: configv1.ConditionFalse,
-			Reason: "NoIngressControllersDegraded",
+		if !foundDegradedStatusCondition {
+			degradedCondition.Status = configv1.ConditionUnknown
+			degradedCondition.Reason = "IngressDoesNotHaveDegradedCondition"
+			degradedCondition.Message = "The default ingress controller is not reporting an Degraded status condition."
```
frobware, line 323: nit: s/an/a.

pkg/operator/controller/status/controller.go
```diff
@@ -288,40 +286,50 @@ func checkAllIngressesAvailable(ingresses []operatorv1.IngressController) bool {
...
+		for _, cond := range ic.Status.Conditions {
+			if cond.Type != operatorv1.OperatorStatusTypeDegraded {
+				continue
+			}
+			foundDegradedStatusCondition = true
+			switch cond.Status {
+			case operatorv1.ConditionFalse:
+				degradedCondition.Status = configv1.ConditionFalse
+				degradedCondition.Reason = "IngressNotDegraded"
+				degradedCondition.Message = "The default ingress controller reports Degraded=False."
```
frobware, line 309: Can we use `manifests.DefaultIngressControllerName` for "default" here and in other messages?

frobware (changes requested):
Overall LGTM, just a few things to tidy up w.r.t messages and using `manifests.DefaultIngressControllerName` in lieu of "default".

manifests/0000_90_ingress-operator_03_prometheusrules.yaml
```diff
@@ -27,3 +27,21 @@ spec:
...
+          severity: warning
+        annotations:
+          message: |
+            The {{ $labels.namespace }}/{{ $labels.name }} ingresscontroller is
+            degraded: {{ $labels.reason }}.
+      - alert: IngressControllerUnavailable
+        expr: ingress_controller_conditions{condition="Available"} == 0
+        for: 5m
+        labels:
+          severity: critical
```
sgreene570, line 43: I'm wondering if this really warrants a critical level alert. A critical alert would already fire if the Default ingress controller were to become unavailable, right?

pkg/operator/controller/ingress/metrics.go
```diff
@@ -15,6 +17,49 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var (
+	// ingressControllerConditions reports the status conditions of each
+	// IngressController using the ingress_controller_conditions metric.
+	ingressControllerConditions = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
sgreene570, line 23: This metric seems like it might have high cardinality. Do we really want to report true/false for every ingress controller status condition? For the purpose of alerting, we only need a small subset of status conditions to be reported via metrics, right?

At the very least, I think we can ignore the canary status condition here since that's reported as it's own metric.

Thoughts?

sgreene570:
I have some high-level questions about the metric changes but overall LGTM

cmd/ingress-operator/start.go
Miciah: Done.

pkg/operator/metrics.go
Miciah: Deleted.

pkg/operator/controller/status/controller.go
Miciah: Fixed.

manifests/0000_90_ingress-operator_03_prometheusrules.yaml
Miciah: Downgraded to warning level.

Miciah:
Updated to address all feedback.

PR #608, opened 2021-05-03
Bug 1954330: Reconcile openshift-ingress namespace on upgrade

sgreene570:
**ingress: Reconcile openshift-ingress namespace on upgrade**

Commits `263c2aa` and `3505256` added new annotations and labels to the
openshift-ingress namespace, which is managed by the ingress-operator.

This commit adds logic to reconcile the openshift-ingress namespace
resource during cluster upgrades. To prevent overwriting changes made to
the namespace resource by other controllers (or cluster-admins), the
namespace reconciliation logic added in this commit explicitly syncs
namespace labels and annotations that the operator sets.

pkg/operator/controller/ingress/namespace.go:

Refactor `ensureRouterNamespace` so that it handles creating and updating
the openshift-ingress namespace resource (and nothing else).

Add new functions: `currentRouterNamespace`, `updateRouterNamespace`, &
`routerNamespaceChanged`.

Move existing Service Account and Cluster Role Binding creation logic
into new functions, `ensureRouterServiceAccount`, and
`ensureRouterClusterRoleBinding`.

Update `ensureRouterNamespace` calling site. Add new calling sites
for `ensureRouterServiceAccount` and `ensureRouterClusterRoleBinding`.

This commit resolves Bug 1954330.

**ingress: Add basic namespace update unit test**

pkg/operator/controller/ingress/namespace_test.go:
Add a trivial function to test `routerNamespaceChanged`
with a couple of relevant test cases.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -699,8 +699,16 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		return fmt.Errorf("failed to ensure cluster role: %v", err)
 	}
 
-	if err := r.ensureRouterNamespace(); err != nil {
+	if _, _, err := r.ensureRouterNamespace(); err != nil {
 		return fmt.Errorf("failed to ensure namespace: %v", err)
+	}
+
+	if err := r.ensureRouterServiceAccount(); err != nil {
+		return fmt.Errorf("failed to ensure service account: %v", err)
+	}
+
+	if err := r.ensureRouterClusterRoleBinding(); err != nil {
+		return fmt.Errorf("failed to ensure cluster role binding: %v", err)
 	}
 
 	var errs []error
--- a/pkg/operator/controller/ingress/namespace.go
+++ b/pkg/operator/controller/ingress/namespace.go
@@ -4,26 +4,127 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	corev1 "k8s.io/api/core/v1"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 )
 
-// ensureRouterNamespace ensures all the necessary scaffolding exists for
-// routers generally, including a namespace and all RBAC setup.
-func (r *reconciler) ensureRouterNamespace() error {
-	ns := manifests.RouterNamespace()
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, ns); err != nil {
-		if !errors.IsNotFound(err) {
-			return fmt.Errorf("failed to get router namespace %q: %v", ns.Name, err)
-		}
-		if err := r.client.Create(context.TODO(), ns); err != nil {
-			return fmt.Errorf("failed to create router namespace %s: %v", ns.Name, err)
-		}
-		log.Info("created router namespace", "name", ns.Name)
+// ensureRouterNamespace ensures that the router namespace exists.
+func (r *reconciler) ensureRouterNamespace() (bool, *corev1.Namespace, error) {
+	desired := manifests.RouterNamespace()
+
+	haveNamespace, current, err := r.currentRouterNamespace()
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -699,10 +699,18 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
...
+	if _, _, err := r.ensureRouterNamespace(); err != nil {
 		return fmt.Errorf("failed to ensure namespace: %v", err)
 	}
 
+	if err := r.ensureRouterServiceAccount(); err != nil {
+		return fmt.Errorf("failed to ensure service account: %v", err)
+	}
+
+	if err := r.ensureRouterClusterRoleBinding(); err != nil {
+		return fmt.Errorf("failed to ensure cluste role binding: %v", err)
```
Miciah, line 711: Missing an "r", and might as well use `%w`:
```suggestion
		return fmt.Errorf("failed to ensure cluster role binding: %w", err)
```

pkg/operator/controller/ingress/namespace.go
```diff
@@ -4,26 +4,128 @@ import (
...
+	if err != nil {
+		return false, nil, err
+	}
+
+	switch {
+	case !haveNamespace:
+		if err := r.client.Create(context.TODO(), desired); err != nil {
+			return false, nil, fmt.Errorf("failed to create router namespace: %v", err)
+		}
+		log.Info("created router namespace", "desired", desired)
```
Miciah, line 33: We're inconsistent in this operator with respect to logging objects and their names.  This isn't a problem with this specific PR, but we should decide whether it is generally useful to log the entire object or better to log just the name.

pkg/operator/controller/ingress/namespace.go
```diff
@@ -4,26 +4,128 @@ import (
...
+	return true, nil
+}
+
+// routerNamespaceChanged returns true if current and expected differ by any
+// of the annotations or the labels controlled by the operator.
+// Other namespace labels and annotations are not overwritten so that
+// values written by other controllers are preserved.
+func routerNamespaceChanged(current, expected *corev1.Namespace) (bool, *corev1.Namespace) {
+	knownAnnotations := []string{
+		"openshift.io/node-selector",
```
Miciah, line 85: This one worries me a little.  For 4.8, would it be worth setting `Upgradeable=False` if the administrator has set this annotation, and then reconcile it only in 4.9+?

pkg/operator/controller/ingress/namespace_test.go
```diff
@@ -0,0 +1,50 @@
...
+			},
+			expect: true,
+		},
+		{
+			description: "if namespace network-policy annotation changes",
+			mutate: func(ns *corev1.Namespace) {
+				ns.Labels["policy-group.network.openshift.io/ingress"] = "foo"
+			},
+			expect: true,
+		},
```
Miciah, line 35: Please add a test case where an unmanaged annotation and an unmanaged label are changed.

pkg/operator/controller/ingress/namespace.go
sgreene570: What if the operator did not manage this annotation at all?

pkg/operator/controller/ingress/namespace_test.go
sgreene570: done.

pkg/operator/controller/ingress/namespace.go
sgreene570: Right. I don't personally have a strong preference either way, although maybe we could get away with not logging entire object's during creation since we log diff's on updates?

pkg/operator/controller/ingress/controller.go
sgreene570: fixed. Thanks!

pkg/operator/controller/ingress/namespace.go
sgreene570: Removed the "openshift.io/node-selector" annotation here as per a conversation on slack.

Miciah: Yeah, if it isn't needed to fix [BZ#1954330](https://bugzilla.redhat.com/show_bug.cgi?id=1954330), we can drop reconciliation of that annotation from this PR.

pkg/operator/controller/ingress/namespace.go
Miciah: Yeah.  We can log interesting fields but otherwise avoid logging the entire object so as to reduce log noise.

Miciah:
Thanks!
/lgtm

PR #622, opened 2021-06-01
Bug 1960284: Set the "local-with-fallback" service annotation

Miciah:
#### Set the "local-with-fallback" service annotation

Set the `traffic-policy.network.alpha.openshift.io/local-with-fallback` annotation on LoadBalancer- and NodePort-type services that use the `Local` external traffic policy.

* `pkg/operator/controller/ingress/load_balancer_service.go` (`localWithFallbackAnnotation`): New const.  Define the annotation key for the `local-with-fallback` service annotation.
(`desiredLoadBalancerService`): Set the `local-with-fallback` annotation when using the `Local` external traffic policy.
(`managedLoadBalancerServiceAnnotations`): New variable.  Define the keys for annotations that the operator manages on LoadBalancer-type services, which now include the `local-with-fallback` annotation.
(`loadBalancerServiceChanged`): Fix the update logic to properly handle annotations with empty values using go-cmp and `managedLoadBalancerServiceAnnotations`.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestDesiredLoadBalancerService`): Verify that the `local-with-fallback` annotation is set when appropriate.
(`TestLoadBalancerServiceChanged`): Verify that `loadBalancerServiceChanged` properly handles updates to the `local-with-fallback` annotation.
* `pkg/operator/controller/ingress/nodeport_service.go` (`desiredNodePortService`): Set the `local-with-fallback` annotation.
(`managedNodePortServiceAnnotations`): New variable.  Define the annotation keys that the operator manages on NodePort-type services.
(`nodePortServiceChanged`): Check if any of the annotations in `managedNodePortServiceAnnotations` have changed.  Update managed annotations if they have changed.
* `pkg/operator/controller/ingress/nodeport_service_test.go` (`TestDesiredNodePortService`): Verify that the expected annotations are set.
(`TestNodePortServiceChanged`): Add a test case to verify that `nodePortServiceChanged` properly handles updates to the `local-with-fallback` annotation.

#### Add local-with-fallback unsupported config override

* `pkg/operator/controller/ingress/load_balancer_service.go` (`desiredLoadBalancerService`): Use the new `shouldUseLocalWithFallback`function to determine whether to set the `local-with-fallback` annotation.
(`shouldUseLocalWithFallback`): New function.  Check the service's external traffic policy and the ingresscontroller's unsupported config overrides and return a Boolean value indicating whether `local-with-fallback` should be enabled, or an error if the override could not be parsed.
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestShouldUseLocalWithFallback`): New test.  Verify that `shouldUseLocalWithFallback` behaves as expected.
* `pkg/operator/controller/ingress/nodeport_service.go` (`ensureNodePortService`): Check the error value from `desiredNodePortService`.
(`desiredNodePortService`): Add error return value.  Use the new `shouldUseLocalWithFallback` function (which can return an error) to determine whether to set the `local-with-fallback` annotation.
* `pkg/operator/controller/ingress/nodeport_service_test.go` (`TestDesiredNodePortService`): Check the error value from `desiredNodePortService`.
* test/e2e/operator_test.go (`TestLocalWithFallbackOverrideForLoadBalancerService`, `TestLocalWithFallbackOverrideForNodePortService`): New tests.  Verify that the operator does not set the `local-with-fallback` annotation on a LoadBalancer or NodePort service if the `localWithFallback` unsupported config override is set to "false".

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -21,6 +21,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -104,6 +105,12 @@ const (
 	// openstackInternalLBAnnotation is the annotation used on a service to specify an
 	// OpenStack load balancer as being internal.
 	openstackInternalLBAnnotation = "service.beta.kubernetes.io/openstack-internal-load-balancer"
+
+	// localWithFallbackAnnotation is the annotation used on a service that
+	// has "Local" external traffic policy to indicate that the service
+	// proxy should prefer using a local endpoint but forward traffic to any
+	// available endpoint if no local endpoint is available.
+	localWithFallbackAnnotation = "traffic-policy.network.alpha.openshift.io/local-with-fallback"
 )
 
 var (
@@ -298,6 +305,10 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef 
 		}
 		// Azure load balancers are not customizable and are set to (2 fail @ 5s interval, 2 healthy)
 		// GCP load balancers are not customizable and are set to (3 fail @ 8s interval, 1 healthy)
+
+		if service.Spec.ExternalTrafficPolicy == corev1.ServiceExternalTrafficPolicyTypeLocal {
+			service.Annotations[localWithFallbackAnnotation] = ""
+		}
 	}
 
 	service.SetOwnerReferences([]metav1.OwnerReference{deploymentRef})
@@ -394,14 +405,31 @@ func (r *reconciler) updateLoadBalancerService(current, desired *corev1.Service,
 	return true, nil
 }
 
+// managedLoadBalancerServiceAnnotations is a set of annotation keys for
+// annotations that the operator manages for LoadBalancer-type services.
+var managedLoadBalancerServiceAnnotations = sets.NewString(
+	awsLBHealthCheckIntervalAnnotation,
+	GCPGlobalAccessAnnotation,
+	localWithFallbackAnnotation,
+)
+
 // loadBalancerServiceChanged checks if the current load balancer service
 // matches the expected and if not returns an updated one.
 func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev1.Service) {
-	updated := current.DeepCopy()
-	changed := false
+	annotationCmpOpts := []cmp.Option{
+		cmpopts.IgnoreMapEntries(func(k, _ string) bool {
+			return !managedLoadBalancerServiceAnnotations.Has(k)
+		}),
+	}
+	if cmp.Equal(current.Annotations, expected.Annotations, annotationCmpOpts...) {
+		return false, nil
+	}
 
-	// Preserve everything but the AWS LB health check interval annotation &
-	// GCP Global Access internal Load Balancer annotation.
+	updated := current.DeepCopy()
+
+	// Preserve everything but the AWS LB health check interval annotation,
+	// GCP Global Access internal Load Balancer annotation, and
+	// local-with-fallback annotation
 	// (see <https://bugzilla.redhat.com/show_bug.cgi?id=1908758>).
 	// Updating annotations and spec fields cannot be done unless the
 	// previous release blocks upgrades when the user has modified those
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -298,12 +306,52 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
+// shouldUseLocalWithFallback returns a Boolean value indicating whether the
+// local-with-fallback annotation should be set for the given service, and
+// returns an error if the given ingresscontroller has an invalid unsupported
+// config override.
+func shouldUseLocalWithFallback(ic *operatorv1.IngressController, service *corev1.Service) (bool, error) {
+	var result bool
+
+	// By default, use local-with-fallback when using the "Local" external
+	// traffic policy.
+	if service.Spec.ExternalTrafficPolicy == corev1.ServiceExternalTrafficPolicyTypeLocal {
```
sgreene570, line 330: Should we return false if `service.Spec.ExternalTrafficPolicy` is not `local`?
As is, this logic would let a user use an unsupported config override to set the local-with-fallback annotation when `service.Spec.ExternalTrafficPolicy` is `cluster`, which is a state we shouldn't allow, right?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -394,33 +442,47 @@ func (r *reconciler) updateLoadBalancerService(current, desired *corev1.Service,
...
+var managedLoadBalancerServiceAnnotations = sets.NewString(
+	awsLBHealthCheckIntervalAnnotation,
+	GCPGlobalAccessAnnotation,
+	localWithFallbackAnnotation,
+)
+
 // loadBalancerServiceChanged checks if the current load balancer service
 // matches the expected and if not returns an updated one.
 func loadBalancerServiceChanged(current, expected *corev1.Service) (bool, *corev1.Service) {
+	annotationCmpOpts := []cmp.Option{
```
sgreene570, line 456: Awesome clean up!

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -394,33 +442,47 @@ func (r *reconciler) updateLoadBalancerService(current, desired *corev1.Service,
...
 
-	if !changed {
-		return false, nil
+	for annotation := range managedLoadBalancerServiceAnnotations {
+		currentVal, have := current.Annotations[annotation]
+		expectedVal, want := expected.Annotations[annotation]
+		if want && (!have || currentVal != expectedVal) {
+			updated.Annotations[annotation] = expected.Annotations[annotation]
+		} else if have && !want {
+			delete(updated.Annotations, annotation)
```
sgreene570, line 484: Could you add a unit test to cover this scenario?

pkg/operator/controller/ingress/nodeport_service.go
```diff
@@ -145,13 +164,38 @@ func nodePortServiceChanged(current, expected *corev1.Service) (bool, *corev1.Se
...
+	if updated.Annotations == nil {
+		updated.Annotations = map[string]string{}
+	}
+	for annotation := range managedNodePortServiceAnnotations {
+		currentVal, have := current.Annotations[annotation]
+		expectedVal, want := expected.Annotations[annotation]
+		if want && (!have || currentVal != expectedVal) {
+			updated.Annotations[annotation] = expected.Annotations[annotation]
+		} else if have && !want {
+			delete(updated.Annotations, annotation)
```
sgreene570, line 196: Could you add a unit test for this scenario?

test/e2e/operator_test.go
```diff
@@ -1925,6 +1925,128 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+	wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+			t.Logf("failed to get service %q: %v", serviceName, err)
+			return false, nil
+		}
+		_, ok := service.Annotations[annotation]
+		return !ok, nil
+	})
+	if _, ok := service.Annotations[annotation]; ok {
+		t.Fatalf("failed to observe removal of the %q annotation on ingresscontroller %q", annotation, defaultName)
```
sgreene570, line 1992:
```suggestion
		t.Fatalf("failed to observe removal of the %q annotation on service %q", annotation, serviceName)
```

test/e2e/operator_test.go
```diff
@@ -1925,6 +1925,128 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+	wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+			t.Logf("failed to get service %q: %v", serviceName, err)
+			return false, nil
+		}
+		_, ok := service.Annotations[annotation]
+		return !ok, nil
+	})
+	if _, ok := service.Annotations[annotation]; ok {
+		t.Fatalf("failed to observe removal of the %q annotation on the %q ingresscontroller", annotation, icName)
```
sgreene570, line 2046:
```suggestion
		t.Fatalf("failed to observe removal of the %q annotation on the %q service", annotation, serviceName)
```

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: I'm kind of ambivalent about allowing the user to do weird things with an unsupported config override.  I'll go ahead and make the suggested change.

sgreene570: Fair enough. Seems like some low-hanging fruit that we might as well patch while we can IMO.

Miciah:
Rebased.

test/e2e/operator_test.go
Miciah: Fixed.  Thanks!

test/e2e/operator_test.go
Miciah: Fixed.  Thanks!

pkg/operator/controller/ingress/nodeport_service.go
Miciah: Fixed.  However, I had to replace the test case for adding the annotation with test cases for updating and deleting.  To cover all three would make the test setup more complicated.  If you want coverage for all three, I'd just as soon add a separate test.  What do you think?

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Fixed.  Thanks!

pkg/operator/controller/ingress/nodeport_service.go
sgreene570: I see. Current coverage looks good enough to me! Thanks.

Miciah:
`TestHTTPCookieCapture` failed.  
/test e2e-aws-operator

28/289 e2e-gcp-serial tests failed, nothing that looks related to ingress.
/test e2e-gcp-serial

Miciah:
/test e2e-gcp-serial
GCP's DNS seems to be broken.

PR #623, opened 2021-06-03
Bug 1935058: Set AWS session region

rfredette:
Commit 64ab7b6:
Update AWS go sdk to v1.38.49

Commit aed9291:
If AWS DNS config specifies a region, also specify that region when establishing the AWS API session.


pkg/dns/aws/dns.go
```diff
@@ -113,6 +113,9 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	if config.CustomCABundle != "" {
 		sessionOpts.CustomCABundle = strings.NewReader(config.CustomCABundle)
 	}
+	if len(config.Region) > 0 {
+		sessionOpts.Config.Region = aws.String(config.Region)
+	}
```
candita, line 118: Does the setting of `region` in the `switch` statement on L129 need to change based on this change?

candita:
A unit test would be good idea for this addition.

pkg/dns/aws/dns.go
danehans: Agreed, I suggest moving the `region` var declaration before this change and include the switch logic up here.

rfredette: I had originally moved the `switch` logic to here, but I ran into some issues.

Because `sess` can't exist before `sessionOpts.Config.Region` is set, the `sess.Config.Region != nil` case can't be evaluated, so we have to figure out which region AWS would use manually. Internally, the AWS go API can pull region from at least 2 places, either a config file (`~/.aws/config` by default, although the file location can be changed), or from the environment variable `AWS_REGION`. I think the env variable takes precedence, but I'm not actually sure.

I doubt that this is a volatile part of the API, but I'd rather avoid trying to guess which region the AWS API uses as default. That said, I think `region` could be set here by default, the `switch` statement could be simplified to:

```go
if len(config.Region) == 0 {
        region = aws.StringVar(sess.Config.Region)
}
```

That part would still have to go after `session.NewSessionWithOptions(sessionOpts)` though.

candita: Is it valid or invalid to have config.Region != sess.Config.Region?

rfredette: I don't know for sure, but I think it's valid in most regions. For example, I think it's valid to authenticate using STS in us-west-1, and use that auth token for queries to route53 in us-east-1

PR #561, opened 2021-02-25
NE 484: Use ingress-operator subcommand instead of hello-openshift 

candita:
NE 484: Use ingress-operator subcommand instead of hello-openshift for canary server

- Add new subcommand `serve-healthcheck` 
- Add servehealthcheck code, same functionality as hello-openshift
- Change  CANARY_IMAGE to openshift/origin-cluster-ingress-operator:latest
- Add a check for Canary Image when computing the operator progressing condition
- Update tests

PR #627, opened 2021-06-22
[release-4.7] Bug 1973983: Canary: Handle downgrades from 4.8 to 4.7 properly

sgreene570:
Commit 8332af2 changed the Canary daemonset image and command values
in OCP 4.8.

Prior to commit 8332af2, the Canary's daemonset logic did not trigger
updates to the Canary daemonset when the `command` field was modified.

When downgrading from OCP 4.8 to OCP 4.7, the 4.7 Canary container
is fed a startup command intended for the 4.8 Canary image, which
results in Canary endpoints that don't actually send back the
expected response body.

This commit is a partial backport of commit 8332af2 in order
to remove unwanted daemonset `command` values available in OCP 4.8
that should not exist on the daemonset in OCP 4.7.

pkg/operator/controller/canary/daemonset.go:

Update the canary daemonset when the canary server, image, or container
name changes (e.g. handle downgrades from OCP 4.8 to OCP 4.7).

pkg/operator/controller/canary/daemonset_test.go:

Add crucial unit test coverage.

/assign @Miciah @candita

### DIFF
--- a/pkg/operator/controller/canary/daemonset.go
+++ b/pkg/operator/controller/canary/daemonset.go
@@ -104,11 +104,20 @@ func canaryDaemonSetChanged(current, expected *appsv1.DaemonSet) (bool, *appsv1.
 	changed := false
 	updated := current.DeepCopy()
 
-	// Update the canary daemonset when the canary server image changes
-	if len(current.Spec.Template.Spec.Containers) > 0 && len(expected.Spec.Template.Spec.Containers) > 0 &&
-		current.Spec.Template.Spec.Containers[0].Image != expected.Spec.Template.Spec.Containers[0].Image {
-		updated.Spec.Template.Spec.Containers[0].Image = expected.Spec.Template.Spec.Containers[0].Image
-		changed = true
+	// Update the canary daemonset when the canary server image, command, or container name changes
+	if len(current.Spec.Template.Spec.Containers) > 0 && len(expected.Spec.Template.Spec.Containers) > 0 {
+		if current.Spec.Template.Spec.Containers[0].Image != expected.Spec.Template.Spec.Containers[0].Image {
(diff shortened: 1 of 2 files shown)


Miciah:
Thanks!
/lgtm

candita:
In https://github.com/openshift/cluster-ingress-operator/pull/627 I also changed the image manifest and the deployment yaml, updated with a new image name.  Are you sure you don't need to change that for a downgrade, or else add the new image to this PR?  What if the downgrade happens after the old image is removed from quay?
* https://github.com/openshift/cluster-ingress-operator/pull/561/files#diff-6c639280ceb15ba85cc0bd847c4e3715e80cc8386f7e2ae78bc5e2ea191e6a26
* https://github.com/openshift/cluster-ingress-operator/pull/561/files#diff-c5de8e303795d333d4d685de85438af5df83ab9e18225ecbc79b4be35281ded3

sgreene570:
> In #561 I also changed the image manifest and the deployment yaml, updated with a new image name. Are you sure you don't need to change that for a downgrade, or else add the new image to this PR? What if the downgrade happens after the old image is removed from quay?
> 
> * https://github.com/openshift/cluster-ingress-operator/pull/561/files#diff-6c639280ceb15ba85cc0bd847c4e3715e80cc8386f7e2ae78bc5e2ea191e6a26
> * https://github.com/openshift/cluster-ingress-operator/pull/561/files#diff-c5de8e303795d333d4d685de85438af5df83ab9e18225ecbc79b4be35281ded3

The image references in both cases are actually just placeholders. The CVO updates those references with the correct pullspec from the given release image when applying the manifests. So if a user is able to downgrade to 4.7, that means they have a valid release image (that contains the proper hello-openshift image).

PR #628, opened 2021-06-23
Add unsupported config override for config manager

Miciah:
Add an unsupported config override to enable OpenShift router's dynamic config manager for HAProxy.

* `pkg/operator/controller/ingress/deployment.go` (`RouterHAProxyConfigManager`): New const.
(`desiredRouterDeployment`): Add unsupported config override for `ROUTER_HAPROXY_CONFIG_MANAGER`.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Verify that `desiredRouterDeployment` sets `ROUTER_HAPROXY_CONFIG_MANAGER` as expected.
* `test/e2e/operator_test.go` (`TestDynamicConfigManagerUnsupportedConfigOverride`): Verify that the operator sets `ROUTER_HAPROXY_CONFIG_MANAGER` on a router deployment for an IngressController that enables the unsupported config override.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1985,6 +1985,44 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
 	}
 }
 
+// TestDynamicConfigManagerUnsupportedConfigOverride verifies that the operator
+// configures router pod replicas to use the dynamic config manager if the
+// ingresscontroller is so configured using an unsupported config override.
+func TestDynamicConfigManagerUnsupportedConfigOverride(t *testing.T) {
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "dynamic-config-manager"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+		t.Errorf("failed to observe expected conditions: %w", err)
+	}
+
+	if err := kclient.Get(context.TODO(), icName, ic); err != nil {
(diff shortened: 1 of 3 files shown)


Miciah:
Rebased.

test/e2e/operator_test.go
```diff
@@ -1985,6 +1985,44 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+	}
+
+	ic.Spec.UnsupportedConfigOverrides = runtime.RawExtension{
+		Raw: []byte(`{"dynamicConfigManager":"true"}`),
+	}
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+	if err := waitForDeploymentEnvVar(t, kclient, deployment, 1*time.Minute, "ROUTER_HAPROXY_CONFIG_MANAGER", "true"); err != nil {
+		t.Fatalf("expected initial deployment to set ROUTER_HAPROXY_CONFIG_MANAGER=true: %v", err)
```
frobware, line 2022: "expected updated deployment"?

Miciah: Fixed.  Thanks!

showuon:
@Miciah , we're wondering for this feature, other than this being explicitly unsupported, are the any drawbacks as to why we should not consider using this a way to address our issues with connection termination? And when it will be officially supported? Thanks.

showuon:
@Miciah @frobware , could you help answer the questions above when you're available? Since we're planning to adopt this dynamic config manager feature in our production environment, we'd really like to get your suggestions about it. Thank you!

frobware:
> @Miciah , we're wondering for this feature, other than this being explicitly unsupported, are the any drawbacks as to why we should not consider using this a way to address our issues with connection termination? And when it will be officially supported? Thanks.

Enabling the override is not currently tested either manually or in any CI jobs that we have/use. I cannot vouch for its current behaviour.

showuon:
@frobware , thanks for the reply. So do you have plan when this feature will be "supported"?

frobware:
@showuon the RFE is tracked here: https://issues.redhat.com/browse/RFE-1439. Notably the spike https://issues.redhat.com/browse/NE-870 is still to be done. I don't have any answer for you as to "when'.

showuon:
Thank you for the comments. I see. Thanks.

PR #630, opened 2021-06-29
Add IBMCloud DNS support

SzucsAti:
With this PR, ingress-operators running on IBMCloud IPI/UPI clusters will be able to use CIS as their DNS provider.

This PR is required to implement IBMCloud IPI/UPI.
Enhancement doc: openshift/enhancements#773  
Relevant Installer PR: openshift/installer#4923


SzucsAti:
Removed unnecessary changes from go.mod.

manifests/00-ingress-credentials-request.yaml
```diff
@@ -68,3 +68,16 @@ spec:
...
+metadata:
+  name: openshift-ingress-ibmcloud
+  namespace: openshift-cloud-credential-operator
+spec:
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: IBMCloudProviderSpec
+  secretRef:
+    name: cloud-credentials
+    namespace: openshift-ingress-operator
```
sgreene570, line 83: Could you please add a newline to end of this file?

pkg/dns/ibm/dns.go
```diff
@@ -0,0 +1,194 @@
...
+		result, response, err := dnsService.ListAllDnsRecords(opt)
+		if err != nil {
+			if response.StatusCode != http.StatusNotFound {
+				return fmt.Errorf("delete: failed to list the dns record: %v", err)
+			}
+			continue
+		}
+		if result != nil && result.Result != nil {
+			for _, resultData := range result.Result {
+				if resultData.ID != nil {
```
sgreene570, line 111: Might be cleaner to check if `resultData.ID == nil` here and return the error on line 121 if it is (rather than indent the rest of the code path)

sgreene570: This suggestion might apply to other place as well

pkg/dns/ibm/dns_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+			dnsService.UpdateDnsRecordInputOutput = tc.updateDnsRecordInputOutput
+
+			err = provider.createOrUpdateDNSRecord(&record, zone)
+
+			if tc.expectedErr {
+				if err == nil {
+					t.Error("Expected error, but err is nil")
+				}
+			}
```
Miciah, line 263: Check `!tc.expectedErr && err != nil`: 
```go
			if !tc.expectedErr && err != nil {
				t.Errorf("Expected nil err, got %v", err)
			}
```
If you do that, I think you'll find that my other comment about checking for "not found" errors in the create/update logic has merit.  * grin *.

Miciah (commented):
Please run `go mod tidy` and squash.

pkg/dns/ibm/dns_test.go
Miciah: Hm, on second look, maybe the "not found" errors aren't relevant.  This check causes the following failure:
```
--- FAIL: TestCreateOrUpdate (0.00s)
    --- FAIL: TestCreateOrUpdate/listFail (0.00s)
        dns_test.go:261: Expected nil err, got createOrUpdateDNSRecord: failed to update the dns record: updateDnsRecord: inputs don't match
```
That can be fixed by adding `updateDnsRecordInputOutput` to the test case:
```go
		{
			desc:         "listFail",
			DNSName:      "testUpdate",
			recordedCall: "PUT",
			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
				OutputError:      nil,
				OutputStatusCode: http.StatusNotFound,
			},
			updateDnsRecordInputOutput: dnsclient.UpdateDnsRecordInputOutput{
				InputId:          "testUpdate",
				OutputError:      nil,
				OutputStatusCode: http.StatusOK,
			},
			expectedErr: false,
		},
```

Miciah: Ahh, I think the test case should actually look like this:
```go
		{
			desc:         "listFail",
			DNSName:      "testUpdate",
			recordedCall: "PUT",
			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
				OutputError:      errors.New("Error in ListAllDnsRecords"),
				OutputStatusCode: http.StatusNotFound,
			},
			updateDnsRecordInputOutput: dnsclient.UpdateDnsRecordInputOutput{
				InputId:          "testUpdate",
				OutputError:      nil,
				OutputStatusCode: http.StatusOK,
			},
			expectedErr: false,
		},
```
Then I get the following failure:
```
--- FAIL: TestCreateOrUpdate (0.00s)
    --- FAIL: TestCreateOrUpdate/listFail (0.00s)
        dns_test.go:266: Expected nil err, got createOrUpdateDNSRecord: failed to list the dns record: Error in ListAllDnsRecords
```
But if I fix `createOrUpdateDNSRecord` as follows:
```go
		result, response, err := dnsService.ListAllDnsRecords(listOpt)
		if err != nil {
			if response.StatusCode != http.StatusNotFound {
				return fmt.Errorf("createOrUpdateDNSRecord: failed to list the dns record: %w", err)
			}
			continue
		}
```
Then the test passes again.

SzucsAti:
Thank you for the reviews! Finished with resolving your comments and implemented the requested changes.

pkg/dns/ibm/dns.go
```diff
@@ -0,0 +1,201 @@
...
+		for _, resultData := range result.Result {
+			if resultData.ID == nil {
+				return fmt.Errorf("delete: record id is nil")
+			}
+			delOpt := dnsService.NewDeleteDnsRecordOptions(*resultData.ID)
+			_, delResponse, err := dnsService.DeleteDnsRecord(delOpt)
+			if err != nil && delResponse.StatusCode != http.StatusNotFound {
+				return fmt.Errorf("delete: failed to delete the dns record: %w", err)
+			}
+			log.Info("deleted DNS record", "record", record.Spec, "zone", zone)
```
Miciah, line 124: This will log "deleted DNS record" even if the record didn't exist.  Maybe that's all right?  ¯\\\_(ツ)\_/¯

pkg/operator/controller/dns/controller.go
```diff
@@ -559,6 +562,31 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 			return nil, fmt.Errorf("failed to create GCP DNS provider: %v", err)
 		}
 		dnsProvider = provider
+	case configv1.IBMCloudPlatformType:
+		if infraStatus.ControlPlaneTopology != configv1.ExternalTopologyMode {
```
Miciah, line 566: I asked asked about a similar `if infraStatus.ControlPlaneTopology != configv1.ExternalTopologyMode` check in `createDNSProviderIfNeeded`, but I forgot to comment on this.  Can this check be done for all providers?

SzucsAti: Yes it can. Changed it accordingly.

pkg/dns/ibm/dns.go
```diff
@@ -0,0 +1,203 @@
...
+		for _, resultData := range result.Result {
+			if resultData.ID == nil {
+				return fmt.Errorf("delete: record id is nil")
+			}
+			delOpt := dnsService.NewDeleteDnsRecordOptions(*resultData.ID)
+			_, delResponse, err := dnsService.DeleteDnsRecord(delOpt)
+			if err != nil && delResponse.StatusCode != http.StatusNotFound {
+				return fmt.Errorf("delete: failed to delete the dns record: %w", err)
+			}
+			if delResponse.StatusCode != http.StatusNotFound {
```
Miciah, line 124: Do we need to check to make sure `delResponse` isn't nil?  Come to think of it, we might need to check for nil values where we use the result or response value from `dnsService.ListAllDnsRecords` as well.

pkg/operator/controller/dns/controller.go
```diff
@@ -230,14 +233,25 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
 		if r.cloudCredentials == nil || !reflect.DeepEqual(creds.Data, r.cloudCredentials.Data) {
 			needUpdate = true
 		}
+	case configv1.IBMCloudPlatformType:
+		if infraConfig.Status.ControlPlaneTopology != configv1.ExternalTopologyMode {
```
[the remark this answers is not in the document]
SzucsAti: We discussed internally, that a separate case is needed for IBMCloud. If r.infraconfig == nil we don’t want to use the ibmcloud provider, however others would want to use their providers. Is this correct? Implemented the changes, but unfortunately it results in code duplication. Should we maybe create a function for this, or just leave it as it is for now?

pkg/operator/controller/dns/controller.go
```diff
@@ -222,13 +223,26 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
 	creds := &corev1.Secret{}
 	switch platformStatus.Type {
 	case configv1.AWSPlatformType, configv1.AzurePlatformType, configv1.GCPPlatformType:
-		name := types.NamespacedName{Namespace: r.config.Namespace, Name: cloudCredentialsSecretName}
-		if err := r.cache.Get(context.TODO(), name, creds); err != nil {
-			return fmt.Errorf("failed to get cloud credentials from secret %s: %v", name, err)
+		if (r.infraConfig != nil && infraConfig.Status.ControlPlaneTopology != configv1.ExternalTopologyMode) || r.infraConfig == nil {
```
Miciah, line 226: The condition can be simplified:
```suggestion
		if r.infraConfig == nil || infraConfig.Status.ControlPlaneTopology != configv1.ExternalTopologyMode {
```
But really, why check `r.infraConfig` at all in this `switch` statement?  If the condition `r.infraConfig == nil` is true, then that means that the `if r.infraConfig == nil || ...` statement below will set `needUpdate = true`, and if `infraConfig.Status.ControlPlaneTopology == configv1.ExternalTopologyMode` is true, then when we call `createDNSProvider`, it will return without using `creds`.  So I think what you had in the previous version of this PR is sufficient: 
```suggestion
		if infraConfig.Status.ControlPlaneTopology != configv1.ExternalTopologyMode {
```

pkg/operator/controller/dns/controller.go
```diff
@@ -222,13 +223,26 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
+			if err := r.cache.Get(context.TODO(), name, creds); err != nil {
+				return fmt.Errorf("failed to get cloud credentials from secret %s: %v", name, err)
+			}
+
+			if r.cloudCredentials == nil || !reflect.DeepEqual(creds.Data, r.cloudCredentials.Data) {
+				needUpdate = true
+			}
 		}
+	case configv1.IBMCloudPlatformType:
+		if r.infraConfig != nil && infraConfig.Status.ControlPlaneTopology != configv1.ExternalTopologyMode {
```
Miciah, line 237: In https://github.com/openshift/cluster-ingress-operator/compare/634e4b2c2a6f74899a3e310bc0c5c28d449d611d..7000e8a956b886d99353c456ddbb4a2744389b2f, you deleted this separate `case` branch and added `configv1.IBMCloudPlatformType` to the `case` above, and then in https://github.com/openshift/cluster-ingress-operator/compare/75f62202398fee26e230891ec54bc2878f67ba88..ca476de7671bcaadaef41712615bc9db0361d29f, you added this `case` back with this condition.  I don't see why we need the two separate `case` branches.  Is it necessary to skip initialization of the DNS provider on IBM Cloud when `r.infraConfig == nil`?  We should avoid special cases where possible (and it would be nice to avoid the code duplication as well if the special casing isn't required).

pkg/operator/controller/dns/controller.go
```diff
@@ -462,6 +476,11 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 		return &dns.FakeProvider{}, nil
 	}
 
+	if infraStatus != nil && infraStatus.ControlPlaneTopology == configv1.ExternalTopologyMode {
```
Miciah, line 479: `infraStatus` cannot be nil.
```suggestion
	if infraStatus.ControlPlaneTopology == configv1.ExternalTopologyMode {
```

Miciah:
Looks great!  Thanks for your patience!
/approve/
/lgtm

PR #631, opened 2021-06-30
Support gcp workload identity federation

akhil-rane:
To enable short-lived credentials in gcp cluster using workload identity
federation (similar to AWS STS), we need to make following set of changes

- google.golang.org/api to a version that supports workload identity
- add kubernetes service account names to gcp credentials request manifest
  which can be used by ccoctl tool to create gcp service accounts with
  limited access

ref: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity
docs PR: https://github.com/openshift/cloud-credential-operator/pull/376
x-ref: https://issues.redhat.com/browse/CCO-123


akhil-rane:
@sgreene570 this is ready for a review. PTAL

manifests/00-ingress-credentials-request.yaml
```diff
@@ -60,6 +60,8 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
 spec:
+  serviceAccountNames:
+  - ingress-operator
```
frobware, line 64: Will this affect non-GCP providers?

akhil-rane: It won't as this credentials request is for GCP only. You have to make changes in the credentials request of a particular provider as done for AWS before [here](https://github.com/openshift/cluster-ingress-operator/blob/e2cdf40beea5700785cc32db32b464850bd124dd/manifests/00-ingress-credentials-request.yaml#L14)

PR #635, opened 2021-07-12
NE-412: Add options for tuning connection timeouts in openshift ingress controller

rfredette:
Convert new api fields from openshift/api#968 into appropriate environment variables for openshift router. These environment variables are used within the router to generate the HAProxy config. `clientTimeout`, `clientFinTimeout`, `serverTimeout`, and `serverFinTimeout` are already implemented in openshift router, but `tlsInspectDelay` will not be rendered into the HAProxy config until implemented in a TBD PR.


pkg/operator/controller/ingress/deployment.go
```diff
@@ -470,6 +484,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
+	if len(ci.Spec.TuningOptions.ClientTimeout) > 0 {
```
frobware, line 487: Please can we extend the existing unig test case to assert these values are present when set on the spec.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -180,6 +185,15 @@ func HardStopAfterIsEnabled(ic *operatorv1.IngressController, ingressConfig *con
 	return HardStopAfterIsEnabledByAnnotation(ingressConfig.Annotations)
 }
 
+// HAProxyTimespec returns the first substring from `input` that matches the
```
frobware, line 188: Should we be more explicit here? Should it be exact vis-a-vis finding the first match?

pkg/operator/controller/ingress/deployment.go
rfredette: Will do

pkg/operator/controller/ingress/deployment.go
rfredette: I plan to switch to using `clipHAProxyTimeoutValue()` for all the new fields, which should be more explicit in its parsing.

pkg/operator/controller/ingress/deployment.go
rfredette: Done :+1:

vendor/modules.txt
```diff
@@ -187,7 +187,7 @@ github.com/modern-go/concurrent
 github.com/modern-go/reflect2
 # github.com/oklog/ulid v1.3.1
 github.com/oklog/ulid
-# github.com/openshift/api v0.0.0-20210628075413-841ec03e5006
+# github.com/openshift/api v0.0.0-20210628075413-841ec03e5006 => github.com/rfredette/api v0.0.0-20210719205548-fe5cb1848870
```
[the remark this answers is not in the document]
rfredette: Not yet. The API PR is open here: https://github.com/openshift/api/pull/968

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1436,3 +1456,29 @@ func clipHAProxyTimeoutValue(val string) (string, error) {
...
+// durationToHAProxyTimespec converts a time.Duration into a number that
+// HAProxy can consume, in the simplest unit possible. If the value would be
+// truncated by being converted to seconds, it outputs in milliseconds,
+// otherwise if the value wouldn't be truncated by converting to seconds, but
+// would if converted to minutes, it outputs in seconds, etc. up to a maximum
+// unit in hours (the largest unit natively supported by time.Duration).
+//
+// Also truncates values to the maximum length HAProxy allows if the value is
+// too large.
+func durationToHAProxyTimespec(duration time.Duration) string {
```
frobware, line 1469: Can we create a unit test for this to assert current functionality and so that we don't regress.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1436,3 +1456,29 @@ func clipHAProxyTimeoutValue(val string) (string, error) {
...
+	haproxyMaxTimeoutMilliseconds := int64(2147483647)
+	durationMilliseconds := duration.Milliseconds()
+	if durationMilliseconds > haproxyMaxTimeoutMilliseconds {
+		log.V(7).Info("Time value %s exceeds the maximum timeout length of %dms. Truncating timeout to match maximum value.", duration.String(), haproxyMaxTimeoutMilliseconds)
```
frobware, line 1473: Should we log at level 2? It's only going to be "noisy" if this happens and, arguably, you'd like to find out quickly because we can't dynamically change the log level.

rfredette: I don't mind making it more verbose, but I chose `V(7)` to match the level of a very similar message generated by `clipHAProxyTimeoutValue`. Both seem like they're similar severity, and only could crop up when generating a new router deployment config if I understand correctly. Do you think that should also be reduced to level 2?

pkg/operator/controller/ingress/deployment.go
rfredette: Done

pkg/operator/controller/ingress/deployment.go
frobware: We can deal with each separately. As as user I'd like to know if I hit some maximum. Logging at level 7 hides the clamping away.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1331,3 +1351,27 @@ func TestDeploymentConfigChanged(t *testing.T) {
 		}
 	}
 }
+
+func TestDurationToHAProxyTimespec(t *testing.T) {
+	testCases := []struct {
+		inputDuration  time.Duration
+		expectedOutput string
+	}{
+		{100 * time.Millisecond, "100ms"},
```
frobware, line 1360: Can we add the expectations for zero.

frobware: Actually, does that (a zero value) make any sense? Just looking for coverage here.

rfredette: It shouldn't get called for a duration of 0, at least at the moment, but for completeness it probably should be covered.

pkg/operator/controller/ingress/deployment.go
rfredette: Fair enough. I'll make that change here.

pkg/operator/controller/ingress/deployment_test.go
frobware: Yeah, I consider this a utility function. I don't know what or who the callers are, but "full coverage" helps its cause. Thanks.

pkg/operator/controller/ingress/deployment.go
rfredette: Done

pkg/operator/controller/ingress/deployment_test.go
rfredette: Done

rfredette:
I'm guessing something went wrong getting the tests scheduled? Prow says "no test found" for this PR

/retest

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1331,3 +1351,28 @@ func TestDeploymentConfigChanged(t *testing.T) {
...
+		{2 * time.Minute, "2m"},
+		{75 * time.Minute, "75m"},
+		{120 * time.Minute, "2h"},
+		{48 * time.Hour, "48h"},
+		{720 * time.Hour, "2147483647ms"},
+	}
+	for _, tc := range testCases {
+		output := durationToHAProxyTimespec(tc.inputDuration)
+		if output != tc.expectedOutput {
+			t.Fatalf("Expected \"%s\", got \"%s\"", tc.expectedOutput, output)
```
Miciah, line 1375: It'd make sense to use `t.Errorf` and `%q`, but no need to hold up this PR for that.  
```suggestion
			t.Errorf("Expected %q, got %q", tc.expectedOutput, output)
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -500,6 +501,25 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+	}
+	if ci.Spec.TuningOptions.ServerFinTimeout != nil && ci.Spec.TuningOptions.ServerFinTimeout.Duration != 0*time.Second {
+		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ServerFinTimeout.Duration)})
+	}
+	if ci.Spec.TuningOptions.TunnelTimeout != nil && ci.Spec.TuningOptions.TunnelTimeout.Duration != 0*time.Second {
+		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_TUNNEL_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.TunnelTimeout.Duration)})
+	}
+	if ci.Spec.TuningOptions.TLSInspectDelay != nil && ci.Spec.TuningOptions.TLSInspectDelay.Duration != 0*time.Second {
+		env = append(env, corev1.EnvVar{Name: "ROUTER_INSPECT_DELAY", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.TLSInspectDelay.Duration)})
+	}
```
Miciah, line 521: Is the zero check needed, or would the nil check suffice?  I guess there's no harm in doing both.  Now I'm wondering whether we should have set minimum values in the CRD.

Miciah:
> I'm guessing something went wrong getting the tests scheduled? Prow says "no test found" for this PR

Maybe some tests were still running when you did the last push?  I think I've seen that cause that error from Prow on other PRs.  In any case, Prow sometimes does things like that.  Just gotta retest (as you did).

pkg/operator/controller/ingress/deployment.go
rfredette: I'm not sure if the CRD supports minimum values for duration, since it's defined as a string in yaml form

This also probably would have been fine without the zero check, but it's little cost for a bit more insurance that the timeouts are sane

pkg/operator/controller/ingress/deployment_test.go
rfredette: Noted. If CI goes particularly bad I'll make this edit, but otherwise I'll try remember it for a future change.

rfredette: Included this change in the required rebase

PR #637, opened 2021-07-19
WIP: Enable config manager

Miciah:
Enable the dynamic config manager.  This PR is only for testing.

See also https://github.com/openshift/cluster-ingress-operator/pull/628, which adds an unsupported config override to enable the config manager (turned off by default).


Miciah:
Rebased.  #628 fell out.

Miciah:
We are continuing this work in #1113.
/close

PR #642, opened 2021-08-17
Bug 1989058: Watch CRL configmaps

Miciah:
#### crl: Delete unnecessary predicates.

Delete superfluous predicates.  The enqueue function already filters out irrelevant updates.

* `pkg/operator/controller/crl/controller.go` (`New`): Delete predicates in the configmap watch.
(`configmapIsInUse`): Delete now unused function.

#### crl: Watch client CA configmap in operand namespace

Change the crl controller watch on configmaps to watch for updates to the client CA configmap that is in the "openshift-ingress" operand namespace.

Before this change, the crl controller watched the user-managed client CA configmap in the "openshift-config" global config namespace.  However, the crl controller builds the CRL configmap from the operator-managed client CA configmap in the operand namespace, so it makes sense to watch this configmap instead.

* `pkg/operator/controller/crl/controller.go` (`New`): Use the name of the operator-managed configmap when building the index.
(`configmapToIngressController`): Ignore the configmap if it isn't in the operand namespace.

#### crl: Watch CRL configmaps

Add a watch on CRL configmaps so that the controller recreates a configmap if the user deletes it.

In principle, the same watch could be used for both client CA configmaps and CRL configmaps, but they are using separate indices and perform logically distinct functions, so it makes sense to use separate watches. The watches share an informer cache, so the overhead of using two watches instead of one is small.

* `pkg/operator/controller/crl/controller.go` (`crlConfigmapIndexFieldName`): New const.
(`New`): Add an index with name `crlConfigmapIndexFieldName` for ingresscontrollers keyed by CRL configmap names.  Add a watch for ingresscontrollers with mapfunc `crlConfigmapToIngressController`.
(`ingressControllersWithConfigmap`): Rename...
(`ingressControllersWithClientCAConfigmap`): ...to this.
(`configmapToIngressController`): Rename...
(`clientCAConfigmapToIngressController`): ...to this.
(`ingressControllersWithCRLConfigmap`): New method.  Return a slice of all ingresscontrollers that reference the specified CRL configmap.
(`crlConfigmapToIngressController`): New method.  Map a CRL configmap to a slice of reconcile requests, using `ingressControllersWithCRLConfigmap`.


pkg/operator/controller/crl/controller.go
```diff
@@ -107,13 +120,17 @@ func (r *reconciler) ingressControllersWithConfigmap(name string) ([]operatorv1.
...
-// configmapToIngressController maps a configmap to a slice of reconcile requests,
-// one request per ingresscontroller that references the configmap.
-func (r *reconciler) configmapToIngressController(o client.Object) []reconcile.Request {
+// clientCAConfigmapToIngressController maps a CRL configmap to a slice of
+// reconcile requests, one request per ingresscontroller that references the
+// configmap.
+func (r *reconciler) clientCAConfigmapToIngressController(o client.Object) []reconcile.Request {
 	requests := []reconcile.Request{}
-	controllers, err := r.ingressControllersWithConfigmap(o.GetName())
+	if o.GetNamespace() != operatorcontroller.DefaultOperandNamespace {
```
alebedev87, line 128: openshift's API is still referencing `openshift-config` namespace in the comment ([here](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L647)). I guess it's not a big deal except maybe `oc explain` command which seems to take the descriptions from the comments.

alebedev87:
Overall LGTM.

I was just wondering about the test: I don't know whether it's possible (or worth the effort) to test the enqueue functions. I don't know whether we can have a sort of fake cache (analogous to the fake client of controller-runtime) which we can index with the expected fields.

pkg/operator/controller/crl/controller.go
```diff
@@ -57,14 +58,27 @@ func New(mgr manager.Manager) (controller.Controller, error) {
...
+	})); err != nil {
+		return nil, fmt.Errorf("failed to create index for ingresscontroller: %w", err)
+	}
+
+	// Index ingresscontrollers over the CRL configmap name so that
+	// crlConfigmapToIngressController can look up the ingresscontroller
+	// associated with a given CRL configmap.
+	if err := operatorCache.IndexField(context.Background(), &operatorv1.IngressController{}, crlConfigmapIndexFieldName, client.IndexerFunc(func(o client.Object) []string {
+		ic := o.(*operatorv1.IngressController)
+		if len(ic.Spec.ClientTLS.ClientCA.Name) == 0 {
```
candita, line 78: Do you still need to check the ClientCA.Name here?

pkg/operator/controller/crl/controller.go
Miciah: The user still creates a configmap in the `openshift-config` namespace.  The clientca-controller copies the configmap from `openshift-config` to the `openshift-ingress` namespace, and then this controller reads it from there.  I've added some additional comments in the latest push; does that clear things up?

pkg/operator/controller/crl/controller.go
Miciah: I think it's appropriate to do so.  The `len(ic.Spec.ClientTLS.ClientCA.Name) == 0` check is really intended to make sure we only index the ingresscontroller if it is using a client CA.  If there's no client CA, then there should be no CRL (the CRL is a list of certificates that were signed using the CA but have been since revoked).  I've added a short comment.  Does it all make sense?

Miciah:
@alebedev87, 
> Overall LGTM.

Thanks!  
 
> I was just wondering about the test: I don't know whether it's possible (or worth the effort) to test the enqueue functions. I don't know whether we can have a sort of fake cache (analogously to the fake client of controller-runtime) which we can index with the expected fields.

I've looked into using envtest in the past, but it seemed heavy and cumbersome (I gave up when I learned I'd need to build the etcd and kube-apiserver binaries to run tests built using envtest).  I can look into using a fake client, and maybe that will be easier.

PR #644, opened 2021-08-17
Bug 1986575: Add e2e test cases for haproxy timeout api fields, and reject negative timeout values

rfredette:
This PR adds 2 tests to the e2e suite:

TestHAProxyTimeouts: 
- Set timeouts for all 6 new timeout fields, and verify that all timeout values are reflected in the router deployment's environment.
- Most of the values are set to durations that, if converted directly from `Duration` to `string`, would be invalid strings to HAProxy. E.g. `90 * time.Second` would string-ify as "1m30s", but HAProxy requires it to only use one time unit, such as "90s". This makes sure the `durationToHAProxyTimespec` function is working as intended.
- One timeout value is set higher than the maximum HAProxy will allow, to make sure that the timeout is clipped to HAProxy's max.

TestHAProxyTimeoutsRejection:
- Sets timeouts for all 6 new timeout fields that are allowed by the API, but are not actually valid timeout periods, like 0 or negative values. Verify that none of the environment variables are set, forcing the router pods to use the default timeouts.

In addition to the tests, this PR also includes a change to reject timeout values that are zero or less, instead of just rejecting zero-length timeouts. The router image will fall back to using default timeouts if negative ones are set, but rejecting them before they are set in the deployment environment is more obvious to the user.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -521,22 +521,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
-	if ci.Spec.TuningOptions.ClientTimeout != nil && ci.Spec.TuningOptions.ClientTimeout.Duration != 0*time.Second {
+	if ci.Spec.TuningOptions.ClientTimeout != nil && ci.Spec.TuningOptions.ClientTimeout.Duration > 0*time.Second {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_CLIENT_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ClientTimeout.Duration)})
 	}
-	if ci.Spec.TuningOptions.ClientFinTimeout != nil && ci.Spec.TuningOptions.ClientFinTimeout.Duration != 0*time.Second {
+	if ci.Spec.TuningOptions.ClientFinTimeout != nil && ci.Spec.TuningOptions.ClientFinTimeout.Duration > 0*time.Second {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_CLIENT_FIN_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ClientFinTimeout.Duration)})
 	}
-	if ci.Spec.TuningOptions.ServerTimeout != nil && ci.Spec.TuningOptions.ServerTimeout.Duration != 0*time.Second {
+	if ci.Spec.TuningOptions.ServerTimeout != nil && ci.Spec.TuningOptions.ServerTimeout.Duration > 0*time.Second {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_SERVER_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ServerTimeout.Duration)})
 	}
-	if ci.Spec.TuningOptions.ServerFinTimeout != nil && ci.Spec.TuningOptions.ServerFinTimeout.Duration != 0*time.Second {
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -521,22 +521,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
-	if ci.Spec.TuningOptions.ClientTimeout != nil && ci.Spec.TuningOptions.ClientTimeout.Duration != 0*time.Second {
+	if ci.Spec.TuningOptions.ClientTimeout != nil && ci.Spec.TuningOptions.ClientTimeout.Duration > 0*time.Second {
```
candita, line 524: What if Spec.TuningOptions is nil?  Is that possible?

test/e2e/haproxy_timeouts_test.go
```diff
@@ -0,0 +1,150 @@
...
+		case "ROUTER_DEFAULT_TUNNEL_TIMEOUT":
+			if envVar.Value != tunnelTimeoutOutput {
+				t.Errorf("expected %s = %q, got %q", envVar.Name, tunnelTimeoutOutput, envVar.Value)
+			}
+		case "ROUTER_INSPECT_DELAY":
+			if envVar.Value != tlsInspectDelayOutput {
+				t.Errorf("expected %s = %q, got %q", envVar.Name, tlsInspectDelayOutput, envVar.Value)
+			}
+		}
+	}
```
candita, line 88: It would be best to actually see these show up in the haproxy.config file with the values that you specified.  You would need to scan a router pod's haproxy.config for the timeout values that should be there.

pkg/operator/controller/ingress/deployment.go
Miciah: > What if Spec.TuningOptions is nil? Is that possible?

It isn't possible.  `Spec.TuningOptions` is not a pointer type.

test/e2e/haproxy_timeouts_test.go
rfredette: Good point. I will add checks for the values in haproxy's config.

test/e2e/haproxy_timeouts_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+	conditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
+		{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+		{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+	}
```
Miciah, line 52: You can use `IngressControllerAvailableConditionType`.

test/e2e/haproxy_timeouts_test.go
```diff
@@ -0,0 +1,150 @@
...
+			fallthrough
+		case "ROUTER_CLIENT_FIN_TIMEOUT":
+			fallthrough
+		case "ROUTER_DEFAULT_SERVER_TIMEOUT":
+			fallthrough
+		case "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT":
+			fallthrough
+		case "ROUTER_DEFAULT_TUNNEL_TIMEOUT":
+			fallthrough
+		case "ROUTER_INSPECT_DELAY":
```
Miciah, line 146:
```suggestion
		case "ROUTER_DEFAULT_CLIENT_TIMEOUT", "ROUTER_CLIENT_FIN_TIMEOUT", "ROUTER_DEFAULT_SERVER_TIMEOUT", "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT", "ROUTER_DEFAULT_TUNNEL_TIMEOUT", "ROUTER_INSPECT_DELAY":
```

PR #653, opened 2021-09-09
Minimize IBM Cloud IAM credentials scope

cjschaef:
Minimize the scope of credentials for the CredentialsRequest for
IBM Cloud provider.


frobware (approved):
LGTM, adding hold as this is a draft and not clear whether you want this to merge in its current state. But also curious: does adding the policies/roles reduce the IAM scope because by default everything is available?

/hold

cjschaef:
@frobware Sounds good, I need to verify the minimal scope is valid for the `ingress-operator`, via testing before I change this is a normal PR.

Yes, from my understanding, the default process will simply use a broad permissions scope. This PR is meant to use fine-grained permissions for IBM Cloud for the `ingress-operator`, by creating an IBM Cloud Service ID and API Key.

Similar and related PR's as an FYI
- [Enhanced ibmcloud ccoctl](https://github.com/openshift/cloud-credential-operator/pull/374)
- [MAO](https://github.com/openshift/machine-api-operator/pull/871)
-[CIRO]( https://github.com/openshift/cluster-image-registry-operator/pull/716)

candita:
Is it possible to add unit tests to validate that you can still create/delete/use certificates?
The `verify` CI test fails - probably because you need to check in generated files like `/pkg/manifests/bindata.go`.

cjschaef:
@candita That's a great question. I'll look to see if that's something I can figure out.

cjschaef:
@candita I added changes from `make update`.

I was wondering about your request for unit tests. I was hoping you could provide more details in regard to the components to unit test related to these changes. Also, I wasn't sure if this would be something targeted for the CI specifically (openshift/release), to gain potential access to credentials and outside API's/CLI's for test verification.

candita:
PR https://github.com/openshift/release/pull/20433 requires this change, to allow a CI IPI deployed cluster to successfully deploy on IBM Cloud and then run E2E conformance tests.

Per @cjschaef this PR "...requires the use of Manual credential mode, and with the recent changes to ccoctl, we must specify the scope of permissions for IAM for each operator... IBM Cloud only uses Manual mode, so our CI testing will have to use (test) these changes, since they are required for a successful IPI deployment on IBM Cloud."
/lgtm

PR #654, opened 2021-09-10
Bug 1943376: Preserve SyncLoadBalancerFailed status if LB stops reporting events

rfredette:
If the LB previously reported a failure event and is still pending, assume the reason is still valid even if the event describing the failure has been deleted

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -755,25 +755,42 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
 	case isPending(service):
 		reason := "LoadBalancerPending"
 		message := "The LoadBalancer service is pending"
+		lbReadyNeedsUpdate := true
 
-		// Try and find a more specific reason for for the pending status.
+		// Try and find a more specific reason for the pending status.
 		createFailedReason := "SyncLoadBalancerFailed"
 		failedLoadBalancerEvents := getEventsByReason(operandEvents, "service-controller", createFailedReason)
-		for _, event := range failedLoadBalancerEvents {
-			involved := event.InvolvedObject
-			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
-				reason = "SyncLoadBalancerFailed"
-				message = fmt.Sprintf("The %s component is reporting SyncLoadBalancerFailed events like: %s\n%s",
-					event.Source.Component, event.Message, "The kube-controller-manager logs may contain more details.")
-				break
+		if len(failedLoadBalancerEvents) > 0 {
+			for _, event := range failedLoadBalancerEvents {
+				involved := event.InvolvedObject
+				if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
+					reason = createFailedReason
+					message = fmt.Sprintf("The %s component is reporting SyncLoadBalancerFailed events like: %s\n%s",
+						event.Source.Component, event.Message, "The kube-controller-manager logs may contain more details.")
+					break
+				}
+			}
+		} else {
+			// The LB service eventually stops reporting events, even if there
+			// are still errors. If the service is still in the pending state
+			// and on the previous update, it was in a failed state, assume it
+			// is still in a failed state even if no events are present.
+			for _, condition := range ic.Status.Conditions {
+				if condition.Type == operatorv1.LoadBalancerReadyIngressConditionType &&
+					condition.Status == operatorv1.ConditionFalse &&
+					condition.Reason == createFailedReason {
+					lbReadyNeedsUpdate = false
+				}
 			}
 		}
-		conditions = append(conditions, operatorv1.OperatorCondition{
-			Type:    operatorv1.LoadBalancerReadyIngressConditionType,
-			Status:  operatorv1.ConditionFalse,
-			Reason:  reason,
-			Message: message,
-		})
+		if lbReadyNeedsUpdate {
+			conditions = append(conditions, operatorv1.OperatorCondition{
+				Type:    operatorv1.LoadBalancerReadyIngressConditionType,
+				Status:  operatorv1.ConditionFalse,
+				Reason:  reason,
+				Message: message,
+			})
+		}
 	}
 
 	return conditions
--- a/test/e2e/sync_lb_fail_test.go
+++ b/test/e2e/sync_lb_fail_test.go
@@ -1,0 +1,138 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var icName = types.NamespacedName{Namespace: "openshift-ingress", Name: "synclb"}
+
+var badLBService = corev1.Service{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "router-" + icName.Name,
+		Namespace: "openshift-ingress",
+		Annotations: map[string]string{
+			// this annotation is intentionally set to an invalid value, which
+			// should cause a SyncLoadBalancerFailed event
+			"service.beta.kubernetes.io/aws-load-balancer-proxy-protocol": "nonsense",
+		},
+		Labels: map[string]string{
+			"app": "router",
+			"ingresscontroller.operator.openshift.io/owning-ingresscontroller": icName.Name,
+			"router": "router-" + icName.Name,
+		},
+	},
+	Spec: corev1.ServiceSpec{
+		Type: corev1.ServiceTypeLoadBalancer,
+		Selector: map[string]string{
+			"ingresscontroller.operator.openshift.io/deployment-ingresscontroller": icName.Name,
+		},
+		Ports: []corev1.ServicePort{
+			{
+				Name:       "http",
+				Port:       80,
+				Protocol:   corev1.ProtocolTCP,
+				TargetPort: intstr.IntOrString{IntVal: 80},
+			},
+			{
+				Name:       "https",
+				Port:       443,
+				Protocol:   corev1.ProtocolTCP,
+				TargetPort: intstr.IntOrString{IntVal: 443},
+			},
+		},
+	},
+}
+
+func TestSyncLBFailEvent(t *testing.T) {
+	t.Parallel()
+	if infraConfig.Status.Platform != configv1.AWSPlatformType {
+		t.Skip("test skipped on non-aws platforms")
+	}
+	if err := kclient.Create(context.TODO(), &badLBService); err != nil {
+		t.Fatalf("failed to create service: %v", err)
+	}
+	defer kclient.Delete(context.TODO(), &badLBService)
+	ic := newLoadBalancerController(icName, icName.Name+dnsConfig.Spec.BaseDomain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer kclient.Delete(context.TODO(), ic)
+	eventList := &corev1.EventList{}
+	err := wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
+		if err := kclient.List(context.TODO(), eventList, client.InNamespace(badLBService.ObjectMeta.Namespace)); err != nil {
+			return false, nil
+		}
+		for _, event := range eventList.Items {
+			if event.Reason == "SyncLoadBalancerFailed" {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("Timed out waiting for SyncLoadBalancerFailed event")
+	}
+	expectedConditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionFalse},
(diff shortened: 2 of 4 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -619,13 +619,27 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+					break
+				}
+			}
+		} else {
+			// The LB service eventually stops reporting events, even if they
+			// are still valid. If the service is still in the pending state
+			// and on the previous update, it was in a failed state, assume it
+			// is still in a failed state even if no events are present.
+			for _, condition := range ic.Status.Conditions {
+				if condition.Type == operatorv1.LoadBalancerReadyIngressConditionType &&
```
candita, line 638: You probably also need to check if the condition.Status was true:
```suggestion
				if condition.Type == operatorv1.LoadBalancerReadyIngressConditionType && condition.Status == operatorv1.ConditionTrue &&
```

pkg/operator/controller/ingress/status.go
```diff
@@ -619,13 +619,27 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+					condition.Reason == "SyncLoadBalancerFailed" {
+					conditions = append(conditions, condition)
```
candita, line 640: Do you think you should add a log here, because the BZ was inquiring about a lack of log messages?

pkg/operator/controller/ingress/status.go
```diff
@@ -619,13 +619,27 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
-			involved := event.InvolvedObject
-			if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
-				reason = "SyncLoadBalancerFailed"
-				message = fmt.Sprintf("The %s component is reporting SyncLoadBalancerFailed events like: %s\n%s",
-					event.Source.Component, event.Message, "The kube-controller-manager logs may contain more details.")
-				break
+		if len(failedLoadBalancerEvents) > 0 {
+			for _, event := range failedLoadBalancerEvents {
+				involved := event.InvolvedObject
+				if involved.Kind == "Service" && involved.Namespace == service.Namespace && involved.Name == service.Name && involved.UID == service.UID {
```
candita, line 625: Just curious - did you happen to experiment to see if the filter on this was too strict, and we somehow lost messages that would have been relevant?

pkg/operator/controller/ingress/status.go
```diff
@@ -619,13 +619,27 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+					reason = "SyncLoadBalancerFailed"
+					message = fmt.Sprintf("The %s component is reporting SyncLoadBalancerFailed events like: %s\n%s",
+						event.Source.Component, event.Message, "The kube-controller-manager logs may contain more details.")
+					break
+				}
+			}
+		} else {
+			// The LB service eventually stops reporting events, even if they
+			// are still valid. If the service is still in the pending state
```
candita, line 634:
```suggestion
			// the last status was failed. If the service is still in the pending state
```

candita:
@rfredette I took a look through the long slack chat mentioned in the BZ, and it did point to a possible issue in the clusteroperator: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/status/controller.go#L378-L395
https://coreos.slack.com/archives/CCH60A77E/p1616706138136800?thread_ts=1616700653.126500&cid=CCH60A77E

Did you get a chance to look into that?

pkg/operator/controller/ingress/status.go
rfredette: There may be more checks than necessary, but I don't think it will reject any events that would be relevant. 

Specifically, maybe the namespace check is unnecessary since I think this is getting events in the `openshift-ingress` namespace, and all those events seem to have involved objects in the same namespace, but it's possible that's not a safe assumption on my part.

pkg/operator/controller/ingress/status.go
rfredette: the Reason `SyncLoadBalancerFailed` currently only exists when `LoadBalancerReady` is false, but I've added this check to be safe

pkg/operator/controller/ingress/status.go
rfredette: The log the BZ is asking about is the installer log, which I don't think we have a direct way to post messages to, but the degraded status condition already includes the error from `LoadBalancerReady` if there is one, and I believe the degraded message does get included in the installer log if ingress stays degraded.

pkg/operator/controller/ingress/status.go
```diff
@@ -693,13 +693,28 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+			// The LB service eventually stops reporting events, even if there
+			// are still errors. If the service is still in the pending state
+			// and on the previous update, it was in a failed state, assume it
+			// is still in a failed state even if no events are present.
+			for _, condition := range ic.Status.Conditions {
+				if condition.Type == operatorv1.LoadBalancerReadyIngressConditionType &&
+					condition.Status == operatorv1.ConditionFalse &&
+					condition.Reason == "SyncLoadBalancerFailed" {
+					conditions = append(conditions, condition)
+					return conditions
```
candita, line 716: Set up the condition and add it later in line 720
```suggestion
					condition.Status == operatorv1.ConditionFalse &&
					condition.Reason == "SyncLoadBalancerFailed" {
					reason = condition.Reason
					message = fmt.Sprintf("No more events occurred after SyncLoadBalancerFailed, and status is still Pending")
					break
```

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,125 @@
...
+			if event.Reason == "SyncLoadBalancerFailed" {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("Timed out waiting for SyncLoadBalancerFailed event")
+	}
+	expectedConditions := []operatorv1.OperatorCondition{
```
candita, line 93: Shouldn't there be an expected condition for `SyncLoadBalancerFailed` in this test?

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		expected := operatorConditionMap(expectedConditions...)
+		current := operatorConditionMap(ic.Status.Conditions...)
+		if conditionsMatchExpected(expected, current) {
+			return false, nil
+		} else {
+			return true, fmt.Errorf("ingresscontroller %s should remain degraded", icName.Name)
+		}
+	})
+	if err != nil && err != wait.ErrWaitTimeout {
```
candita, line 122: I don't think you can check for `ErrWaitTimeout` here.  You'll have to fix that error by making the time longer.

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,125 @@
...
+var icName = types.NamespacedName{Namespace: "openshift-ingress", Name: "synclb"}
+
+var badLBService = corev1.Service{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "router-" + icName.Name,
+		Namespace: "openshift-ingress",
+		Annotations: map[string]string{
+			// this annotation is intentionally set to an invalid value, which
+			// should cause a SyncLoadBalancerFailed event
+			"service.beta.kubernetes.io/aws-load-balancer-proxy-protocol": "nonsense",
```
candita, line 34: Why not find an annotation that would make every platform fail, or another reason why it would fail to sync?  Then you could run the test on the other platforms.  If you want to figure this out in a different PR, that's ok with me.

test/e2e/sync_lb_fail_test.go
rfredette: When `wait.PollImmediate` returns an error, it always returns `wait.ErrWaitTimeout` if the timeout is reached. The idea of how I wrote this is that if `LoadBalancerReady` stays `False` and `Degraded` stays `True` for the entire timeout, it's considered a success, but if it changes from that status, the poll loop will exit early with a different error, indicating a failure.

test/e2e/sync_lb_fail_test.go
rfredette: I wanted to use `waitForIngressControllerCondition` to cut down on rewriting similar functions, but it only matches condition type and status, while matching exactly on `SyncLoadBalancerFailed` would require reading the reason on the `LoadBalancerReady` status condition. But now that I try to justify that choice, I think you're right, that I should explicitly check the reason.

pkg/operator/controller/ingress/status.go
rfredette: That's a reasonable change in the way this function works, although I'd like to keep the existing message. Even if the event goes away, I think the existing message that lists out some of the messages from previous SyncLoadBalancerFailed events is more informative

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,125 @@
...
+		t.Fatalf("Timed out waiting for SyncLoadBalancerFailed event")
+	}
+	expectedConditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionFalse},
+		{Type: operatorv1.OperatorStatusTypeDegraded, Status: operatorv1.ConditionTrue},
+	}
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, expectedConditions...); err != nil {
+		t.Fatalf("Timed out waiting for status conditions LoadBalancerReady=false and Degraded=true for ingresscontroller %s", icName.Name)
+	}
+	err = wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
```
gcs278, line 100: another block comment here would be super helpful, this block of code is taking me a while to interpret.

test/e2e/sync_lb_fail_test.go
gcs278: With these type of E2Es, I'm a little surprised we aren't concerned about ballooning the E2E test time, since this test guarantees 5 mins runtime on pass, but I supposed better safe than sorry.

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,125 @@
...
+				TargetPort: intstr.IntOrString{IntVal: 443},
+			},
+		},
+	},
+}
+
+func TestSyncLBFailEvent(t *testing.T) {
+	if infraConfig.Status.Platform != configv1.AWSPlatformType {
+		t.Skip("test skipped on non-aws platforms")
+		return
```
gcs278, line 67: *nit* I don't think you need this return here, it appears t.Skip does the return for you.

gcs278, line 67: *nit* I don't think you need this return here, it appears t.Skip does the return for you.

gcs278 (commented):
Overall looks good, but I just have some general questions and some small nit picks.

pkg/operator/controller/ingress/status.go
```diff
@@ -743,25 +743,42 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+			}
+		} else {
+			// The LB service eventually stops reporting events, even if there
+			// are still errors. If the service is still in the pending state
+			// and on the previous update, it was in a failed state, assume it
+			// is still in a failed state even if no events are present.
+			for _, condition := range ic.Status.Conditions {
+				if condition.Type == operatorv1.LoadBalancerReadyIngressConditionType &&
+					condition.Status == operatorv1.ConditionFalse &&
+					condition.Reason == "SyncLoadBalancerFailed" {
```
gcs278, line 769: Use variable createFailedReason here

test/e2e/sync_lb_fail_test.go
gcs278: Bump

test/e2e/sync_lb_fail_test.go
```diff
@@ -0,0 +1,138 @@
...
+				Name:       "https",
+				Port:       443,
+				Protocol:   corev1.ProtocolTCP,
+				TargetPort: intstr.IntOrString{IntVal: 443},
+			},
+		},
+	},
+}
+
+func TestSyncLBFailEvent(t *testing.T) {
```
gcs278, line 63: If this goes in after Andy's parallelization of e2e tests, this would be a candidate for running in parallel, since you make your own ingress controller.

test/e2e/sync_lb_fail_test.go
gcs278: It's fine though, especially regarding my comment with parallelization.

gcs278 (commented):
Looks good just a couple of small things.

test/e2e/sync_lb_fail_test.go
rfredette: Do you know if adding `t.Parallel()` will do anything before his changes merge? If not, I can look into it, and if it's innocuous I'll add it now

test/e2e/sync_lb_fail_test.go
rfredette: With parallelization, the few minutes here will almost definitely finish before our longest tests, but I reduced the time to 3 minutes to be a bit faster. It seems to still be long enough in the manual test runs I did.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -545,7 +545,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 
 	if len(ci.Status.Domain) > 0 {
-		env = append(env, corev1.EnvVar{Name: "ROUTER_CANONICAL_HOSTNAME", Value: "router-" + ci.Name + "." + ci.Status.Domain})
+		cName := "router-" + ci.Name + "." + ci.Status.Domain
```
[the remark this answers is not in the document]
rfredette: I've rebased locally and it seems theres a commit in my branch that's no longer in master that has these changes to deployment.go. I'm not sure where it was reverted yet, but it's not my change, so I'll be dropping it

test/e2e/sync_lb_fail_test.go
gcs278: Looking back at this. Changes are merged so this will need to be updated.

test/e2e/sync_lb_fail_test.go
candita: > 

I think you just need to return `true, nil` where appropriate.

rfredette:
the bug this is for has been closed. If it is reopened I will revisit this, but for now I'm closing the PR as well

PR #674, opened 2021-11-03
NE-882: Set `ROUTER_DOMAIN` to enable route subdomain field

Miciah:
Set `ROUTER_DOMAIN` to the IngressController's domain in the router deployment so that the router can use the domain to construct a default host for routes that specify an empty value for `spec.host` and a nonempty value for `spec.subdomain`.

* `pkg/operator/controller/ingress/deployment.go` (`desiredRouterDeployment`): Set `ROUTER_DOMAIN`.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Verify that `ROUTER_DOMAIN` is set as expected.

### DIFF
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -312,6 +312,7 @@ func TestDesiredRouterDeployment(t *testing.T) {
 		{"STATS_PASSWORD_FILE", true, "/var/lib/haproxy/conf/metrics-auth/statsPassword"},
 		{"SSL_MIN_VERSION", true, "TLSv1.1"},
 		{WildcardRouteAdmissionPolicy, true, "false"},
+		{"ROUTER_DOMAIN", false, ""},
 	}
 	if err := checkDeploymentEnvironment(t, deployment, tests); err != nil {
 		t.Error(err)
@@ -510,6 +511,7 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 		{"ROUTER_IP_V4_V6_MODE", true, "v4v6"},
 		{RouterEnableCompression, true, "true"},
 		{RouterCompressionMIMETypes, true, "text/html application/*"},
+		{"ROUTER_DOMAIN", true, ic.Status.Domain},
 	}
 	if err := checkDeploymentEnvironment(t, deployment, tests); err != nil {
 		t.Error(err)
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -527,7 +527,11 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 
 	if len(ci.Status.Domain) > 0 {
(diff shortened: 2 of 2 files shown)


pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -232,6 +232,7 @@ func TestDesiredRouterDeployment(t *testing.T) {
 
 	checkDeploymentHasEnvVar(t, deployment, "ROUTER_USE_PROXY_PROTOCOL", false, "")
 
+	checkDeploymentHasEnvVar(t, deployment, "ROUTER_DOMAIN", false, "")
```
gcs278, line 235: If you control-f in this function, there are 4 `desiredRouterDeployment`, so 4 different generations of a deployment object. Looks like you did the private, loadbalancing, but there's another loadbalancing and the hostnetwork ingress controller tests.

To be thorough, would it make sense to just add the checks after the other two new deployment objects on what you expect? Specifically, I remember working a bug in which private and loadbalancing did one thing and hostNetwork did something different.

gcs278: Oh and I'll throw another plug that this function is super confusing and needs to be written into different test or something more readable (for another day I suppose...)

Miciah: > If you control-f in this function, there are 4 `desiredRouterDeployment`, so 4 different generations of a deployment object. Looks like you did the private, loadbalancing, but there's another loadbalancing and the hostnetwork ingress controller tests.
> 
> To be thorough, would it make sense to just add the checks after the other two new deployment objects on what you expect? Specifically, I remember working a bug in which private and loadbalancing did one thing and hostNetwork did something different.

I suppose I could check `ROUTER_DOMAIN` after each call to `desiredRouterDeployment`.  However, the intention with `TestDesiredRouterDeployment` was really that for each configuration option, we exercise every code path related to that option.  In some cases, there are only two possible code paths, so we only need to check the effect of two different settings to cover both code paths (that's the case with `status.domain` and `ROUTER_DOMAIN`: either the domain is present or it's empty, and we set the environment variable or omit it accordingly).  In other cases, we have one option with many code paths (such as `spec.endpointPublishingStrategy.type, which has 4 possible settings with different code paths; we also have some settings that interact in interesting ways, which we need to test).  It just happens that we mashed everything into a single test function, so we end up calling `desiredRouterDeployment` four times in the same test, which is needed to cover the possible code paths for some options, even if it isn't needed for other code paths and options.  

> Oh and I'll throw another plug that this function is super confusing and needs to be written into different tests or something more readable (for another day I suppose...)

Coming soon to a Git clone near you (courtesy of Candace): https://github.com/openshift/cluster-ingress-operator/pull/712

(Which means I'm going to need to rebase once that PR merges.  * grin *.)

gcs278: Sure, so you are basically saying you aren't exercising a different code path for the other `desiredRouterDeployment`. That's fair, I can buy that. I think it comes back to how large and confusing this function is, so I'm excited to see Candace's update.

candita: It's merged!  I don't know who's more excited, you or me!

Miciah:
Rebased for #712.

gcs278:
/lgtm
:taco:

PR #675, opened 2021-11-05
add a metric for the NLB usage on AWS

tjungblu:
for NE-704, next step there would be to expose through cluster-monitoring-operator


pkg/operator/controller/ingress/metrics.go
```diff
@@ -26,9 +26,15 @@ var (
 		Help: "Report the conditions for ingress controllers. 0 is False and 1 is True.",
 	}, []string{"name", "condition"})
 
+	nlbUsage = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ingress_controller_aws_nlb_usage",
```
candita, line 30: "usage" usually implies the amount of a limited resource (cpu usage, memory usage, etc.).  Is there a better name?  I would suggest *_count but you're not counting them.  It might be better to count the number in use over time.

pkg/operator/controller/ingress/metrics.go
```diff
@@ -68,6 +74,23 @@ func DeleteIngressControllerConditionsMetric(ic *operatorv1.IngressController) {
 	}
 }
 
+func UpdateIngressControllerUsageMetrics(ci *operatorv1.IngressController) error {
```
candita, line 77: You may need to create a set and delete func for the metric.  If a customer tried AWS NLB for a proof of concept and then deleted it, do we still want to record that?

pkg/operator/controller/ingress/metrics.go
tjungblu: good point, I was just briefly looking through the others in cluster-monitoring-operator and came across "_info":
https://github.com/openshift/cluster-monitoring-operator/blob/master/jsonnet/rules.libsonnet#L343-L346

`ingress_controller_aws_nlb_info` maybe? wdyt? Or just dropping the suffix altogether.

pkg/operator/controller/ingress/metrics.go
tjungblu: wouldn't that set the gauge to zero on the next reconciliation? Or do you explicitly want to cover the scenario of: created + deleted?

I reckon in the latter scenario it would show as NLB = 1 during the proof of concept time and then drop back to 0.

candita: Deletion is taken care of during reconciliation, but the metrics need to be specifically deleted, as in https://github.com/openshift/cluster-ingress-operator/pull/675/files#diff-0c9271ea6402ecb6233e9e0a629cd72156e12bb7a2c41ef99f0d0265394d7ec4R199

pkg/operator/controller/ingress/metrics.go
candita: Maybe `ingress_controller_aws_nlb_active` if you don't count them.  But I still think it would be better to count them.

pkg/operator/controller/ingress/metrics.go
tjungblu: we want to tell whether an ingress controller in AWS has the NLB or not - why do we need to delete the metric in the latter case? 

Otherwise it will be difficult to tell later, what the percentage is of NLB enabled ingress controllers as the baseline is missing.

pkg/operator/controller/ingress/metrics.go
tjungblu: > But I still think it would be better to count them.

gotta look up the Gauge implementation in Golang, but IIRC it will sum across all gauges along the "name" label. So for a cluster you should be able to get `sum(ingress_controller_aws_nlb_usage)` that would give you the number of active NLB ingress controllers, wouldn't it?

Or do you think we need to use a counter instead of a gauge?

pkg/operator/controller/ingress/metrics.go
tjungblu: is that what you mean?

```
the IngressController CRD "fizz" is created 
 ....  -> we'll report "fizz" -> 1
the IngressController CRD is deleted
```

and here we continue to report "fizz" -> 1 since we didn't delete the metric for "fizz"?

tjungblu: added the deletion :)

tjungblu:
other PR for exposing said metric in: https://github.com/openshift/cluster-monitoring-operator/pull/1509

pkg/operator/controller/ingress/controller.go
```diff
@@ -674,7 +674,7 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 	}
 
 	// Delete the metrics related to the ingresscontroller
-	DeleteIngressControllerConditionsMetric(ingress)
+	DeleteIngressControllerMetrics(ingress)
```
candita, line 677: I would keep two separate functions, one for controller conditions and one for this new metric.

candita (changes requested):

pkg/operator/controller/ingress/metrics.go
```diff
@@ -26,9 +26,15 @@ var (
 		Help: "Report the conditions for ingress controllers. 0 is False and 1 is True.",
 	}, []string{"name", "condition"})
 
+	activeNLBs = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ingress_controller_aws_nlb_active",
+		Help: "Report the number of active NLBs on AWS clusters. 0 is False, 1 is True",
+	}, []string{"name"})
```
jan--f, line 32: As mentioned in https://github.com/openshift/cluster-monitoring-operator/pull/1509 the `name` label is potentially problematic, as it seems to have an unbounded value set.
Consider dropping this to keep cardinality of this metric in check.

candita: If you remove the name, please validate what the metric looks like when one ingress-controller has NLB, but another one does not.  We should still see a metric with the value 1.

tjungblu: within a cluster, the amount of ingress controllers is not that high (2 at max?).
In openshift/cluster-monitoring-operator#1509  I changed it to a sum across the whole cluster, so you only get a single integer back without a label. I hope that's OK @jan--f?

.gitignore
```diff
@@ -16,6 +16,9 @@ auto-save-list
 tramp
 .\#*
 
+# Goland
```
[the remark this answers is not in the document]
tjungblu: can I add https://www.toptal.com/developers/gitignore/api/go,vim,emacs,visualstudiocode,goland instead?

tjungblu:
hi @candita,

applied all your comments. Let me know if there's anything left to be done.

cc @mcurry-rh

pkg/operator/controller/ingress/metrics.go
jan--f: Sure the recording rule is fine, however we do ingest this metric into the in-cluster stack.
That being said, a dozen or so label values is perfectly fine. I suppose creating those is nothing the average user can do right?
The only concern here would be that someone can create lot of label values in a short time frame (by creating a lot of ingress controllers), blowing up the resources needed to track this metric. Often upgrades can be an issue too where things get restarted and names change rapidly because they are (at least partly) autogenerated. None of that seems to be a concern here though so :+1:

tjungblu: yeah, I think the user will be in bigger trouble by creating the equivalent of ingress controllers before Prometheus would blow up on the metric dimension. :+1:

pkg/operator/controller/ingress/controller.go
candita: Now you should revert the name change too.

pkg/operator/controller/ingress/metrics.go
```diff
@@ -68,6 +74,23 @@ func DeleteIngressControllerConditionsMetric(ic *operatorv1.IngressController) {
 	}
 }
 
+func DeleteActiveNLBMetrics(ic *operatorv1.IngressController) {
+	activeNLBs.DeleteLabelValues(ic.Name)
+}
+
+func SetIngressControllerNLBMetric(ci *operatorv1.IngressController) {
+	labelVal := 0
+	if ci.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
```
candita, line 83: ci.Status.EndpointPublishingStrategy should be checked for non-nil, as it is an optional field.  All of these `&&` statements can be put in one `if` clause, there is no need for two `if` clauses.

pkg/operator/controller/ingress/metrics_test.go
```diff
@@ -140,6 +140,67 @@ func TestDeleteIngressControllerConditionsMetric(t *testing.T) {
...
+						}},
+				}}),
+			expectedMetricFormat: `
+			# HELP ingress_controller_aws_nlb_active Report the number of active NLBs on AWS clusters.
+			# TYPE ingress_controller_aws_nlb_active gauge
+			ingress_controller_aws_nlb_active{name="test1"} 1
+			`,
+		},
+		{
+			name: "classic ELB metrics happy path",
```
candita, line 166: And a test with no EndpointPublishingStrategy would be good.

candita (changes requested):
I have a few more comments, if you can change those and squash your commits it should be ready to merge

tjungblu:
hi @candita,

applied all your comments. Let me know if there's anything left to be done.

cc @mcurry-rh

pkg/operator/controller/ingress/metrics.go
```diff
@@ -68,6 +74,23 @@ func DeleteIngressControllerConditionsMetric(ic *operatorv1.IngressController) {
...
+
+func SetIngressControllerNLBMetric(ci *operatorv1.IngressController) {
+	labelVal := 0
+	if ci.Status.EndpointPublishingStrategy != nil &&
+		ci.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.Type == operatorv1.AWSLoadBalancerProvider &&
+		ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil &&
+		ci.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+		labelVal = 1
```
candita, line 89: There are better ways to express this, but I'll let it go for now. If you come back to this, can you clean this up ?

tjungblu: I'm happy to apply any suggestion you have here

PR #677, opened 2021-11-10
Bug 1918651: ingress operator doesn't report any error when one of router pods is not ready

brandisher:
**Issue 1**: The cluster-ingress-operator reports Available: true and Progressing: false even when the number of defined in ingresscontroller/default.spec.replicas are not all available (e.g. 1/2 replicas available).

**Issue 2**: The cluster-ingress-operator reports Available: true and Progressing: false even when the ingresscontroller deployment reports Progressing: true.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -58,6 +58,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.TLSProfile = computeIngressTLSProfile(ic.Status.TLSProfile, deployment)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentPodsScheduledCondition(deployment, pods))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentProgressingCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/ingress/status_test.go
```diff
@@ -959,6 +959,46 @@ func TestComputeIngressAvailableCondition(t *testing.T) {
...
+	}
+	testCases := []struct {
+		description string
+		deployment  *appsv1.Deployment
+		expect      operatorv1.OperatorCondition
+	}{
+		{
+			description: "deployment has a progressing condition",
+			deployment: deployment(appsv1.DeploymentProgressing, corev1.ConditionTrue),
+			expect: operatorv1.OperatorCondition{Type: IngressControllerDeploymentProgressingConditionType, Status: operatorv1.ConditionTrue},
```
frobware: Should we also assert on the contents of `Reason` and `Message`?

frobware (changes requested):

pkg/operator/controller/ingress/status.go
```diff
@@ -246,6 +247,32 @@ func computeIngressAvailableCondition(conditions []operatorv1.OperatorCondition)
...
 }
 
+// computeDeploymentProgressingCondition computes the current Progressing status
+// of the related Deployment
+func computeDeploymentProgressingCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+	var progressingConditions []*appsv1.DeploymentCondition
+
+	for _, dc := range deployment.Status.Conditions {
+		if dc.Type == appsv1.DeploymentProgressing {
+			progressingConditions = append(progressingConditions, &dc)
```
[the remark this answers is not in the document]
brandisher: That's a good point; originally I intended on surfacing the progressing reason but that felt like that was redundant since you could always check the Deployment itself. I'll simplify this.

pkg/operator/controller/ingress/status_test.go
brandisher: I think that'd be a good idea. I noticed that Deployments have a nuance where they show progressing even after the replicaset has finished progressing so testing that the reason and message are expected makes sense.

pkg/operator/controller/ingress/status.go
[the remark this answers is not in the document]
brandisher: Did a rewrite of this in the latest push so marking this as resolved for now.

pkg/operator/controller/ingress/status_test.go
brandisher: Same here; rewritten in the latest push

brandisher:
Rerunning this one more time as there were some registry issues at the time of the last run.

/test e2e-aws-single-node

brandisher:
Random container build failures

/test e2e-aws-single-node

brandisher:
Trying one more time as there's several terraform related errors

/test e2e-aws-single-node

brandisher:
Closing this out as this is fixed in OCP 4.9+

PR #685, opened 2021-12-13
CFE-72: make infrastructure resource tags updatable

Elbehery:
This PR adds a watch for Infrastructure type to update AWS tags when changed

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -105,6 +105,9 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &configv1.Ingress{}}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
 		return nil, err
 	}
+	if err := c.Watch(&source.Kind{Type: &configv1.Infrastructure{}}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
+		return nil, err
+	}
 	return c, nil
 }
 
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -221,6 +221,7 @@ var (
 			// local-with-fallback annotation for kube-proxy (see
 			// <https://bugzilla.redhat.com/show_bug.cgi?id=1960284>).
 			localWithFallbackAnnotation,
+			awsLBAdditionalResourceTags,
 		)
 
 		// Azure and GCP support switching between internal and external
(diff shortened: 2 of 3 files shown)


Elbehery:
This error exists on the master branch, not from my commit 

@arjunrn 
@lmzuccarelli 
@Connect2naga

pkg/operator/controller/ingress/controller.go
```diff
@@ -105,6 +105,9 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &configv1.Ingress{}}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
 		return nil, err
 	}
+	if err := c.Watch(&source.Kind{Type: &configv1.Infrastructure{}}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
```
lmzuccarelli, line 108: Nice simple change

lmzuccarelli:
@Elbehery Can you check why the verify job is failing.

Elbehery:
> @Elbehery Can you check why the verify job is failing.

there is a fix here https://github.com/openshift/build-machinery-go/pull/58 .. I will try to bump in the `machinary` version in this repo

Elbehery:
@lmzuccarelli  added this PR https://github.com/openshift/cluster-ingress-operator/pull/688 to fix the issue

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -468,6 +468,7 @@ var managedLoadBalancerServiceAnnotations = sets.NewString(
 	awsLBHealthCheckIntervalAnnotation,
 	GCPGlobalAccessAnnotation,
 	localWithFallbackAnnotation,
+	awsLBAdditionalResourceTags,
```
Miciah, line 471: This will cause the operator to override user changes to the annotation.  @frobware, is there a reason the operator didn't manage this annotation before?  Is documenting the change sufficient, or do we need to backport some logic to set `Upgradeable=False` if the user has set an annotation that doesn't match the tags specified in the infrastructure config?

frobware: > is there a reason the operator didn't manage this annotation before?

It's not clear to me whether this should have been specified previously or is/was a glaring omission; that's perhaps not terribly helpful but if we go ahead with this change then we should also go with the additional logic (and backports) for smooth upgrades (IMO).

Elbehery: @Miciah  @frobware  thanks a lot for your review 

The target of this PR is a watch the AWS resource tags on the `infrastructure` crd and update the tags accordingly. 
Would this change be sufficient for this purpose ? 

cc @arjunrn @lmzuccarelli

PR #691, opened 2021-12-18
Bug 1997226: setDefaultPublishingStrategy: Deep copy, tests

Miciah:
Make a copy of `spec.endpointPublishingStrategy` to avoid mutating it, and add unit tests for `setDefaultDomain` and `setDefaultPublishingStrategy`.

Follow-up to #681.

* `pkg/operator/controller/ingress/controller.go` (`setDefaultPublishingStrategy`): Use a deep copy of `ic.Spec.EndpointPublishingStrategy`.
* `pkg/operator/controller/ingress/controller_test.go` (`TestSetDefaultDomain`, `TestSetDefaultPublishingStrategySetsPlatformDefaults`, `TestSetDefaultPublishingStrategyHandlesUpdates`): New tests.

---

@candita, @miheer

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -352,7 +352,7 @@ func setDefaultDomain(ic *operatorv1.IngressController, ingressConfig *configv1.
 }
 
 func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig *configv1.Infrastructure) bool {
-	effectiveStrategy := ic.Spec.EndpointPublishingStrategy
+	effectiveStrategy := ic.Spec.EndpointPublishingStrategy.DeepCopy()
 	if effectiveStrategy == nil {
 		var strategyType operatorv1.EndpointPublishingStrategyType
 		switch infraConfig.Status.Platform {
--- a/pkg/operator/controller/ingress/controller_test.go
+++ b/pkg/operator/controller/ingress/controller_test.go
@@ -4,9 +4,425 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
+
+// TestSetDefaultDomain verifies that setDefaultDomain behaves correctly.
+func TestSetDefaultDomain(t *testing.T) {
+	ingressConfig := &configv1.Ingress{
+		Spec: configv1.IngressSpec{
+			Domain: "apps.mycluster.com",
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -352,7 +352,7 @@ func setDefaultDomain(ic *operatorv1.IngressController, ingressConfig *configv1.
 }
 
 func setDefaultPublishingStrategy(ic *operatorv1.IngressController, infraConfig *configv1.Infrastructure) bool {
-	effectiveStrategy := ic.Spec.EndpointPublishingStrategy
+	effectiveStrategy := ic.Spec.EndpointPublishingStrategy.DeepCopy()
```
frobware, line 355: Is there any need to backport this semantic change?

frobware (approved):
Given that we now "Make a copy of spec.endpointPublishingStrategy" is there a need outside of 4.10 to backport this change?

Otherwise LGTM.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -4,10 +4,421 @@ import (
...
+			name:        "vSphere",
+			infraConfig: makeInfra(configv1.VSpherePlatformType),
+			expectedIC:  ingressControllerWithHostNetwork,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ic := &operatorv1.IngressController{}
+			infraConfig := tc.infraConfig.DeepCopy()
+			if actualResult := setDefaultPublishingStrategy(ic, infraConfig); actualResult != true {
```
frobware, line 211: Using coverage I don't see a case for: 

```go
                case configv1.LibvirtPlatformType:
                        strategyType = operatorv1.HostNetworkStrategyType
```

in `setDefaultPublishingStrategy()`.

pkg/operator/controller/ingress/controller.go
Miciah: No, this change only fixes an issue revealed by unit tests.  The change isn't needed to resolve the original issue in the BZ.

pkg/operator/controller/ingress/controller_test.go
Miciah: I could add a case for Libvirt, but it isn't a supported platform, so I didn't see a need.

Miciah: We do specify the defaulting behavior for Libvirt in the IngressController API godoc though, so I'll add a case.

pkg/operator/controller/ingress/controller.go
Miciah: If we backport the fix, I'd rather backport just #681 and not #691.  The deep copy is to fix a correctness issue revealed by the unit tests but isn't required to fix the original issue.

PR #692, opened 2021-12-21
manifests/00-ingress-credentials-request: Fill in cluster-profile annotations

wking:
All of these platforms require manual credential creation (e.g. via `ccoctl`), so it doesn't really matter if the CredentialRequests get pushed into the cluster or not.  But having them in the cluster will allow the Kube API-server and possibly the cloud-credential operator to complain about anything surprising they see ("that's not a valid value for that property", "I'm watching over your manual-mode shoulder, and you missed a requested secret", etc.).  Regardless of whether we have that coverage today, getting these into the cluster is consistent with our other, existing CredentialsRequests.

Also tweak the namespace of the Alibaba request to match our usual `openshift-cloud-credential-operator`.  Neither the in-cluster cloud-credential operator nor `ccoctl` care about cred-request namespaces; they both happily process requests in any namespace.  But all of our other, existing CredentialsRequests use the cloud-cred namespace, so stick with that for consistency too.


Miciah:
/lgtm
/hold
@kwoodson, @csrwng, would you mind double-checking that these changes are appropriate for Alibaba and IBM Cloud?

csrwng:
hypershift/roks doesn't use the cloud credential operator, so the `ibm-cloud-managed` annotation can be left out. However, it doesn't do any harm since there's no operator to do anything with them.

wking:
Waiting while we decide whether the cloud-credential operator should be installed for more profiles than [just `self-managed-high-availability`][1].

[1]: https://github.com/openshift/cloud-credential-operator/blob/9906c6a5d26a5326a80a236fb56dd3244892db34/manifests/03-deployment.yaml#L7

manifests/00-ingress-credentials-request.yaml
```diff
@@ -117,7 +125,11 @@ apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
   name: openshift-ingress-alibabacloud
-  namespace: openshift-ingress-operator
+  namespace: openshift-cloud-credential-operator
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
```
kwoodson, line 130: I don't think these will do anything for Alibaba but I'd prefer if this wasn't added as I don't see why it would be ibm-cloud-managed.

wking:
ok, fbcf0197 -> ef473577 drops `ibm-cloud-managed` from the CredentialsRequests (following the above comments), and also drops `single-node-developer` across the board (it isn't used, see openshift/cluster-version-operator#685).

manifests/00-ingress-credentials-request.yaml
wking: I've [dropped them][1].

[1]: https://github.com/openshift/cluster-ingress-operator/pull/692#issuecomment-999338209

manifests/00-ingress-credentials-request.yaml
```diff
@@ -117,7 +125,11 @@ apiVersion: cloudcredential.openshift.io/v1
...
+    include.release.openshift.io/self-managed-high-availability: "true"
```
wking, line 120: Talking this over with @joelddiaz , we probably want to keep these out of this profile, to make it more clear to cluster admins that the cloud-cred operator will completely ignore CredentialsRequests on these platforms, regardless of whether it's configured for manual or mint or whatever mode.

joelddiaz: this does imply we should do this for ALL CredentialsRequests for platforms that are only supported in "Manual" mode...

PR #693, opened 2022-01-10
Bug 2039339: Upgradeable Condition in Operator and IC status

arjunrn:
This change sets the `Upgradeable` condition for the IngressController. If the corresponding load balancer _Service_ has been modified, particularly the AWS resource tags fields then the IngressController is marked as `Upgradable=False`. This in turn marks the operator as `Upgradeable=False`.


Miciah:
Thanks!
/lgtm
/approve

PR #697, opened 2022-01-24
WIP: Bug 2005351: desiredRouterDeployment: Default 1 replica on SNO

Miciah:
#### Bump to openshift/api for `spec.replicas` godoc

Bump to github.com/Miciah/api@d8ba5065c5396381d20d95f3be8c75b6befa295d to get updated godoc for the `IngressController.spec.replicas` API field.

* `go.mod`: Bump.
* `go.sum`:
* `pkg/manifests/bindata.go`:
* `vendor/*`: Regenerate.

#### `desiredRouterDeployment`: Default 1 replica on SNO

When determining the default number of router pod replicas for an ingresscontroller that does not specify replicas, check the infrastructure topology, and default to 1 replica if the topology is "SingleReplica" (i.e., Single-Node OpenShift).

When the publishing strategy is "Internal", the installer generates a default ingresscontroller manifest, and this manifest does not specify replicas.  As a result, the ingress operator used the default value of 2 replicas.  On single-node clusters, this caused the ingresscontroller to reported degraded status because only 1 pod replica can be scheduled per ingresscontroller per node.

This issue does not arise when the publishing strategy is "External" because in this case, the installer does not generate a default ingresscontroller manifest; instead, the ingress operator creates the default ingresscontroller, and the ingress operator does check the topology and specify an appropriate, explicit number of replicas in the ingresscontroller that it creates.

Follow-up to #519.

* `pkg/operator/controller/ingress/deployment.go` (`ensureRouterDeployment`): Pass `infraConfig` to `desiredRouterDeployment`.
(`desiredRouterDeployment`): Add a parameter for the infrastructure config. Use the infrastructure to check whether the infrastructure topology.  If the topology is "SingleReplica" and replicas is not specified, default to 1 replica.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeployment`): Verify that `desiredRouterDeployment` behaves correctly when the infrastructure topology is "SingleReplica".

---

Depends on https://github.com/openshift/api/pull/1103.


candita:
/retitle [WIP] Bug 2005351: desiredRouterDeployment: Default 1 replica on SNO

candita:
/retitle WIP: Bug 2005351: desiredRouterDeployment: Default 1 replica on SNO

PR #701, opened 2022-02-08
BUG 2037447: Disable keepalive for canary probe

frobware:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2037447

If keepalives are enabled then we see an accumulation of established
canary connections that remain in ESTABLISHED state. Given that we
only probe once per minute it's not clear we gain anything from
keepalives so let's just disable them when probing the canary host.

This helps prevent resource exhaustion.


Miciah:
Thanks!
/lgtm

PR #702, opened 2022-02-10
Properly reconcile and react to changes in cluster operator resource

mfojtik:
When an external actor modifies the clusteroperators/ingress resource, it takes 10h[1] for the ingress operator to react and reconcile.
This fix (hopefully) makes the ingress operator reconcile the status when clusteroperator resource change, so it should react to any modification immediately. 

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#example-New


Miciah:
/close
Superseded by <https://github.com/openshift/cluster-ingress-operator/pull/714>.

PR #703, opened 2022-02-17
Add support for route53 in the us-isob-east-1 region

dmc5179:
Currently the operator defaults to using the us-east-1 region endpoint for route53 in the us-isob-east-1 region which does not work. Add support for the route53 in the us-isob-east-1 region.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -142,15 +142,20 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	elbConfig := aws.NewConfig().WithRegion(region)
 	tagConfig := aws.NewConfig()
 
+	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
+	if !ok {
+		return nil, fmt.Errorf("Unable to determine partition from region: %q", region)
+	}
+
 	// If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
 	// 1. hard code route53 api endpoint to https://route53.amazonaws.com.cn and region to "cn-northwest-1"
 	//    as route53 is not GA in AWS China and aws sdk didn't have the endpoint.
 	// 2. use the aws china region cn-northwest-1 to setup tagging api correctly instead of "us-east-1"
-	switch region {
-	case endpoints.CnNorth1RegionID, endpoints.CnNorthwest1RegionID:
+	switch partition.ID() {
+	case endpoints.AwsCnPartitionID:
 		tagConfig = tagConfig.WithRegion(endpoints.CnNorthwest1RegionID)
 		r53Config = r53Config.WithRegion(endpoints.CnNorthwest1RegionID).WithEndpoint(chinaRoute53Endpoint)
-	case endpoints.UsGovEast1RegionID, endpoints.UsGovWest1RegionID:
+	case endpoints.AwsUsGovPartitionID:
 		// Route53 for GovCloud uses the "us-gov-west-1" region id:
 		// https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html
 		r53Config = r53Config.WithRegion(endpoints.UsGovWest1RegionID)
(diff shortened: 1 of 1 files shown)


pkg/dns/aws/dns.go
```diff
@@ -158,10 +158,10 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 		// in the same region as the Route53 client to find the hosted zone
 		// of managed records.
 		tagConfig = tagConfig.WithRegion(endpoints.UsGovWest1RegionID)
-	case endpoints.UsIsoEast1RegionID:
-		// The resourcetagging API is not available in C2S
+	case endpoints.UsIsoEast1RegionID, endpoints.UsIsobEast1RegionID:
```
staebler, line 161: I suggest that we take this opportunity to consider the AWS partition rather than the region. Otherwise, we are going to have to come back and make changes every time there are no regions added to the ISO or ISOB partitions.

dmc5179: agreed. I can take a look how this is being managed in the installer and apply it here.

dmc5179: Pushed new commit to switch from hard coded regions to partition based logic.

pkg/dns/aws/dns.go
```diff
@@ -142,23 +142,28 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	elbConfig := aws.NewConfig().WithRegion(region)
 	tagConfig := aws.NewConfig()
 
+	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
+	if !ok {
+		return nil, fmt.Errorf("Unable to determine partition from region: %q", region)
```
staebler, line 147: I don't know if it may be better here to continue on with an unknown partition and use the default case in the following switch--as opposed to failing when the partition cannot be determined.

pkg/dns/aws/dns.go
```diff
@@ -142,26 +142,31 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
...
 		// in the same region as the Route53 client to find the hosted zone
 		// of managed records.
 		tagConfig = tagConfig.WithRegion(endpoints.UsGovWest1RegionID)
-	case endpoints.UsIsoEast1RegionID:
-		// The resourcetagging API is not available in C2S
+	case endpoints.AwsIsoPartitionID, endpoints.AwsIsoBPartitionID:
+		// The resourcetagging API is not available in C2S or SC2S
 		tagConfig = nil
-		// Do not override the region in C2s
+		// Do not override the region in C2S or SC2S
```
staebler, line 169: Does the AWS behavior in C2S and SC2S follow other partitions where there is one region used for global resources? In particular, if the region used for the cluster is us-iso-west-1, will the us-iso-east-1 region need to be used when querying the resource tagging API for route53 resources?

pkg/dns/aws/dns.go
dmc5179: I'll update the code to use the default case.

pkg/dns/aws/dns.go
dmc5179: I have to ask AWS about this. I will submit a support case to AWS when I go on site today but might not hear back from them until next week.

dmc5179: I submitted a case to AWS on this topic and heard back today. The "correct" method is to point all route53 requests to route53.c2s.ic.gov but AWS is leaving the route53.us-iso-east-1.c2s.ic.gov alias in place because it has been around for so long that they don't want to break anything by removing it. This means that I need to change the PR because the us-iso-west-1 region will result in route53.us-iso-west-1.cs2.ic.gov which will not exist. 

I could change the code to say, if isoB, use route53.sc2s.sgov.gov, if iso use route53.c2s.ic.gov

Thoughts?

staebler: I don't think that we need any of the special logic in this switch any more, regardless of the partition. The AWS SDK selects the correct endpoints to use, and we should be making use of the selections made by the SDK.

staebler: https://github.com/openshift/cluster-ingress-operator/blob/fe1165625070372bb0e5d6b1be53ba5b76d3a8cc/vendor/github.com/aws/aws-sdk-go/aws/endpoints/defaults.go#L10893-L10905

staebler:
/lgtm
@Miciah Could you take a look at this for approval?

pkg/dns/aws/dns.go
rfredette: Is this change still planned? I also agree that it'd be preferable to use the default for unknown regions, rather than error out

mjpytlak:
Providing doc ack. No doc is required for this PR

Miciah:
/label docs-approved
based on https://github.com/openshift/cluster-ingress-operator/pull/703#issuecomment-1110996241.

PR #704, opened 2022-02-17
Bug 2055470: Normalize the AWS internal LB annotation value

Miciah:
The AWS cloud-provider implementation checks the `service.beta.kubernetes.io/aws-load-balancer-internal` service annotation to determine whether a service load-balancer (SLB) should be internal.  The cloud-provider implementation recognizes both the value "0.0.0.0/0" and the value "true" as indicating that the SLB should be internal.  OpenShift 4.7 and earlier use the value "0.0.0.0/0", and OpenShift 4.8 and later use the value "true".  A service that was created on an older cluster might have the value "0.0.0.0/0", which can cause comparisons that check for the "true" value to return the wrong result.  This PR adds logic to normalize the annotation value to "true".

Follow-up to <https://github.com/openshift/cluster-ingress-operator/pull/543> and to <https://github.com/openshift/cluster-ingress-operator/pull/582>.

* `pkg/operator/controller/ingress/load_balancer_service.go` (`ensureLoadBalancerService`): Call the new `normalizeLoadBalancerServiceAnnotations` method to normalize the AWS internal LB annotation.
(`normalizeLoadBalancerServiceAnnotations`): New method.  Normalize the `service.beta.kubernetes.io/aws-load-balancer-internal` service annotation by replacing the value "0.0.0.0/0" with the value "true".
* `test/e2e/operator_test.go` (`TestInternalLoadBalancer`): Verify that the operator updates the old "0.0.0.0/0" annotation value to "true".

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -291,6 +291,14 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 				return haveLBS, currentLBService, err
 			}
 		}
+		if updated, err := r.normalizeLoadBalancerServiceAnnotations(currentLBService); err != nil {
+			return true, currentLBService, fmt.Errorf("failed to normalize annotations for load balancer service: %w", err)
+		} else if updated {
+			haveLBS, currentLBService, err = r.currentLoadBalancerService(ci)
+			if err != nil {
+				return haveLBS, currentLBService, err
+			}
+		}
 		deleteIfScopeChanged := false
 		if _, ok := ci.Annotations[autoDeleteLoadBalancerAnnotation]; ok {
 			deleteIfScopeChanged = true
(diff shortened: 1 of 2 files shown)


Miciah:
Unfortunately [the operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/704/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1494204979970314240/artifacts/e2e-aws-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-78f4779ffb-cjxmq_ingress-operator.log) start a few minutes after the `TestInternalLoadBalancer` test ran, so we don't see the "normalized annotation" log entry.  However, `TestInternalLoadBalancer` still passes, and we can see in [the test output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/704/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1494204979970314240/artifacts/e2e-aws-operator/test/build-log.txt) that it ran the new verification and passed:
```
=== RUN   TestInternalLoadBalancer
    operator_test.go:835: expected service.beta.kubernetes.io/aws-load-balancer-internal=true, found service.beta.kubernetes.io/aws-load-balancer-internal=0.0.0.0/0
    operator_test.go:844: deleted ingresscontroller test
```
Let's make sure the test change didn't break other platforms.

/test e2e-azure-operator
/test e2e-gcp-operator

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -484,6 +492,34 @@ func (r *reconciler) deleteLoadBalancerServiceFinalizer(service *corev1.Service)
...
+		// (i.e. it could have been from a cache).
+		updated := service.DeepCopy()
+		updated.Annotations[awsInternalLBAnnotation] = "true"
+		if err := r.client.Update(context.TODO(), updated); err != nil {
+			return false, fmt.Errorf("failed to normalize %s annotation on service %s/%s: %w", awsInternalLBAnnotation, service.Namespace, service.Name, err)
+		}
+
+		log.Info("normalized annotation", "namespace", service.Namespace, "name", service.Name, "annotation", awsInternalLBAnnotation, "old", v, "new", updated.Annotations[awsInternalLBAnnotation])
+
+		return true, nil
```
miheer, line 517: may be we shall say return true, err ?  anyway the logic will remain same.

PR #707, opened 2022-02-24
BUG 2054200: Fix removing custom created service in openshift-ingress with same name

gcs278:
If custom created service in the openshift-ingress namespace has the same naming convention of what the ingress controllers uses, it will be deleted. This fix ensures that the only services that get deleted are owned by the ingress controller responsible for creating it.

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -261,10 +261,16 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 		return false, nil, err
 	}
 
+	// BZ2054200: Don't modify/delete services that are not directly owned by this controller.
+	ownLBS := isServiceOwnedByIngressController(currentLBService, ci)
+
 	switch {
 	case !wantLBS && !haveLBS:
 		return false, nil, nil
 	case !wantLBS && haveLBS:
+		if !ownLBS {
+			return false, nil, fmt.Errorf("a conflicting load balancer service exists that is not owned by the ingress controller: %s", controller.LoadBalancerServiceName(ci))
+		}
 		if err := r.deleteLoadBalancerService(currentLBService, &crclient.DeleteOptions{}); err != nil {
 			return true, currentLBService, err
 		}
@@ -275,6 +281,9 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 		}
 		return r.currentLoadBalancerService(ci)
 	case wantLBS && haveLBS:
+		if !ownLBS {
+			return false, nil, fmt.Errorf("a conflicting load balancer service exists that is not owned by the ingress controller: %s", controller.LoadBalancerServiceName(ci))
+		}
 		if updated, err := r.normalizeLoadBalancerServiceAnnotations(currentLBService); err != nil {
 			return true, currentLBService, fmt.Errorf("failed to normalize annotations for load balancer service: %w", err)
 		} else if updated {
@@ -294,6 +303,14 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 		}
 	}
 	return true, currentLBService, nil
+}
+
+// isServiceOwnedByIngressController determines whether a service is owned by an ingress controller.
+func isServiceOwnedByIngressController(service *corev1.Service, ic *operatorv1.IngressController) bool {
+	if service != nil && service.Labels[manifests.OwningIngressControllerLabel] == ic.Name {
+		return true
+	}
+	return false
 }
 
 // desiredLoadBalancerService returns the desired LB service for a
--- a/pkg/operator/controller/ingress/load_balancer_service_test.go
+++ b/pkg/operator/controller/ingress/load_balancer_service_test.go
@@ -8,6 +8,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -735,4 +736,67 @@ func TestLoadBalancerServiceChanged(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestServiceIngressOwner(t *testing.T) {
+	testCases := []struct {
+		description string
+		service     *corev1.Service
+		ingressName string
+		expect      bool
+	}{
+		{
+			description: "if owner is set correctly",
+			service: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						manifests.OwningIngressControllerLabel: "foo",
+					},
+				},
+			},
+			ingressName: "foo",
+			expect:      true,
+		},
+		{
+			description: "if owner is not set correctly",
+			service: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						manifests.OwningIngressControllerLabel: "foo",
+					},
+				},
+			},
+			ingressName: "bar",
+			expect:      false,
+		},
+		{
+			description: "if owner label is not set at all",
+			service: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{},
+			},
+			ingressName: "bar",
+			expect:      false,
+		},
+		{
+			description: "if service is nil",
+			service:     nil,
+			ingressName: "bar",
+			expect:      false,
+		},
+	}
+
+	for _, tc := range testCases {
+		ic := &operatorv1.IngressController{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: tc.ingressName,
+			},
+			Spec:   operatorv1.IngressControllerSpec{},
+			Status: operatorv1.IngressControllerStatus{},
+		}
+
+		if actual := isServiceOwnedByIngressController(tc.service, ic); actual != tc.expect {
+			t.Errorf("expected ownership %t got %t", tc.expect, actual)
+		}
+	}
+
 }
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -2723,6 +2723,89 @@ func TestTunableRouterKubeletProbesForCustomIngressController(t *testing.T) {
 	}
 }
 
+// TestIngressControllerServiceNameCollision validates BZ2054200: Don't delete services that are not directly owned by this controller.
+// It creates a service with the same naming convention as the ingress controller creates its own load balancing services.
+// Then it triggers a reconcilation of the ingress operator to see if it will delete our service.
+func TestIngressControllerServiceNameCollision(t *testing.T) {
+	// Create the new private controller that we will later create a service to collide with the naming scheme of this.
+	icName := types.NamespacedName{
+		Namespace: operatorNamespace,
+		Name:      "e2e-name-collision-test",
+	}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+
+	// Clean up our new Ingress Controller after we are done.
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Wait for new ingress controller to come online.
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Create services that could collide with the new ingress controller's naming convention for loadbalancer and nodeport.
+	// TRICKY: Our private ingress controller will reconcile and delete extra loadBalancerServices, even though the
+	//         ingress controller isn't a loadbalancer type itself.
+	conflictingLoadBalancerServiceName := types.NamespacedName{
+		Name:      "router-" + icName.Name,
+		Namespace: "openshift-ingress",
+	}
+	conflictingLoadBalancerService := buildEchoService(conflictingLoadBalancerServiceName.Name, conflictingLoadBalancerServiceName.Namespace, nil)
+	if err := kclient.Create(context.TODO(), conflictingLoadBalancerService); err != nil {
+		t.Fatalf("failed to create service %s: %v", conflictingLoadBalancerServiceName, err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), conflictingLoadBalancerService); err != nil {
+			t.Fatalf("failed to delete service %s: %v", conflictingLoadBalancerServiceName, err)
+		}
+	}()
+
+	conflictingNodeportServiceName := types.NamespacedName{
+		Name:      "router-nodeport-" + icName.Name,
+		Namespace: "openshift-ingress",
+	}
+	conflictingNodeportService := buildEchoService(conflictingNodeportServiceName.Name, conflictingNodeportServiceName.Namespace, nil)
+	if err := kclient.Create(context.TODO(), conflictingNodeportService); err != nil {
+		t.Fatalf("failed to create service %s: %v", conflictingNodeportServiceName, err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), conflictingNodeportService); err != nil {
+			t.Fatalf("failed to delete service %s: %v", conflictingNodeportServiceName, err)
+		}
+	}()
+
+	ic, err := getIngressController(t, kclient, icName, 1*time.Minute)
+	if err != nil {
+		t.Fatalf("failed to get ingress controller: %v", err)
+	}
+
+	// Annotate the ingress operator, to trigger a reconcilation to determine if our service is deleted.
+	// This may not be needed, but it ensures a reconcilation occurs ASAP.
+	if ic.Annotations == nil {
+		ic.Annotations = map[string]string{}
+	}
+	ic.Annotations["ingress.operator.openshift.io/e2e-name-collision-test"] = ""
+
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait to see if our service gets deleted by the operator due to name collision.
+	oldLoadBalancerUID := conflictingLoadBalancerService.UID
+	oldNodePortUID := conflictingNodeportService.UID
+	wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		// Check if LoadBalancer and Nodeport Service don't get deleted for the entire duration of this loop.
+		// Will throw fatal error if deleted or marked for deletion and this loop stops.
+		assertServiceNotDeleted(t, conflictingLoadBalancerServiceName, oldLoadBalancerUID)
+		assertServiceNotDeleted(t, conflictingNodeportServiceName, oldNodePortUID)
+
+		return false, nil
+	})
+}
+
 func newLoadBalancerController(name types.NamespacedName, domain string) *operatorv1.IngressController {
 	repl := int32(1)
 	return &operatorv1.IngressController{
@@ -2985,6 +3068,25 @@ func deleteIngressController(t *testing.T, cl client.Client, ic *operatorv1.Ingr
 		return fmt.Errorf("timed out waiting for ingresscontroller to be deleted: %v", err)
 	}
 	return nil
+}
+
+// assertServiceNotDeleted asserts that a provide service wasn't deleted.
+func assertServiceNotDeleted(t *testing.T, serviceName types.NamespacedName, oldUid types.UID) {
+	t.Helper()
+
+	// First check our LoadBalancer Service.
+	service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+		t.Fatalf("expected %s to be present: %v", serviceName, err)
+	}
+	// If there is a DeletionTimestamp, it has been marked for deletion.
+	if service.DeletionTimestamp != nil {
+		t.Fatalf("expected service %s to not be marked for deletion: %v", serviceName, service.DeletionTimestamp)
+	}
+	// If the UID has changed, then the service has been recreated.
+	if service.UID != oldUid {
+		t.Fatalf("expected service %s to have UID %v, got %v", serviceName, oldUid, service.UID)
+	}
 }
 
 func createDefaultCertTestSecret(cl client.Client, name string) (*corev1.Secret, error) {
--- a/pkg/operator/controller/ingress/nodeport_service.go
+++ b/pkg/operator/controller/ingress/nodeport_service.go
@@ -52,10 +52,16 @@ func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, dep
 		return false, nil, err
 	}
 
+	// BZ2054200: Don't modify/delete services that are not directly owned by this controller.
+	ownLBS := isServiceOwnedByIngressController(current, ic)
+
 	switch {
 	case !wantService && !haveService:
 		return false, nil, nil
 	case !wantService && haveService:
+		if !ownLBS {
+			return false, nil, fmt.Errorf("a conflicting nodeport service exists that is not owned by the ingress controller: %s", controller.LoadBalancerServiceName(ic))
+		}
 		if err := r.client.Delete(context.TODO(), current); err != nil {
 			if !errors.IsNotFound(err) {
 				return true, current, fmt.Errorf("failed to delete NodePort service: %v", err)
@@ -71,6 +77,9 @@ func (r *reconciler) ensureNodePortService(ic *operatorv1.IngressController, dep
 		log.Info("created NodePort service", "service", desired)
 		return r.currentNodePortService(ic)
 	case wantService && haveService:
+		if !ownLBS {
+			return false, nil, fmt.Errorf("a conflicting nodeport service exists that is not owned by the ingress controller: %s", controller.LoadBalancerServiceName(ic))
+		}
 		if updated, err := r.updateNodePortService(current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update NodePort service: %v", err)
 		} else if updated {


test/e2e/operator_test.go
```diff
@@ -2585,6 +2585,72 @@ func TestCustomErrorpages(t *testing.T) {
...
+	}
+
+	// Wait to see if our service gets deleted by the operator due to name collision
+	wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), echoServiceName, echoService); err != nil {
+			t.Fatalf("expected %s/%s to not be deleted: %v", echoServiceName.Namespace, echoServiceName.Name, err)
+			return true, nil
+		}
+		return false, nil
+	})
```
Miciah, line 2651: Please use full sentences, including the period, here and in the other comments that you are adding.  (I know I'm being picky, but I want to set good habits.)

You can simplify the format slightly by using `echoServiceName` directly; a `type.NamespacedName` value gets stringified as "namespace/name".  Same applies for lines 2618 and 2624.  

We can make this test a little more thorough by verifying that `deletionTimestamp` is nil and that the uid hasn't changed (which would be the case if the service were deleted and recreated).  

```suggestion
	// Wait to see if our service gets deleted by the operator due to name collision.
	oldUID := echoService.oldUID
	wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
		if err := kclient.Get(context.TODO(), echoServiceName, echoService); err != nil {
			t.Fatalf("expected service %s to be present: %v", echoServiceName, err)
			return true, nil
		}
		if echoService.DeletionTimestamp != nil {
			t.Fatalf("expected service %s to not be marked for deletion: %v", echoServiceName, echoService.DeletionTimestamp)
			return true, nil
		}
		if echoService.UID != oldUID {
			t.Fatalf("expected service %s to have UID %v, got %v", echoServiceName, oldUID, echoService.UID)
			return true, nil
		}
		return false, nil
	})
```

gcs278: Thanks - I addressed these points in my latest push

gcs278:
@Miciah could I get you to take a look when you get a chance

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -465,6 +465,12 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
 		}
 		return false, nil, err
 	}
+
+	// BZ2054200: Don't delete services that are not directly owned by this controller
+	if service.Labels[manifests.OwningIngressControllerLabel] != ci.Name {
+		return false, nil, nil
```
Miciah, line 471: We should do something similar in `currentNodePortService` in `nodeport_service.go`.  

What happens if (1) the ingresscontroller has `spec.endpointPublishingStrategy.type: LoadBalancerService`, (b) a service with the desired name exists, and (c) the service isn't owned by the operator?

gcs278: Okay I'll add it in nodeport_service.go, good point. I was able to create another edge case (maybe this is what you were saying?) in which I create a ingress controller called "test" with `spec.endpointPublishingStrategy.type: LoadBalancerService` then created a service called `router-nodeport-test`. When the IngressController reconciles, it deletes `router-nodeport-test`, despite being a `LoadBalancerService` type.

This is because the nodeport_service.go reconciles regardless of `EndpointPublishingStrategy.type` and attempts to clean up unwanted services. Good catch.

gcs278: Added. Resolving this.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -465,6 +465,12 @@ func (r *reconciler) currentLoadBalancerService(ci *operatorv1.IngressController
 		}
 		return false, nil, err
 	}
+
+	// BZ2054200: Don't delete services that are not directly owned by this controller
+	if service.Labels[manifests.OwningIngressControllerLabel] != ci.Name {
```
candita, line 470: Couldn't there be more than one service to check?

candita, line 470: `currrentLoadBalancer` is called by other methods that are not deleting.  I think you need to check this only when deleting.

candita:
@gcs278 can you add some unit tests just to validate that the service gets deleted when it should, and doesn't get deleted when it shouldn't?

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Correct - this method determines whether we have a `currentLoadBalancerService`. Either deleting, creating, or updating, we still want to filter out these services that aren't owned by our ingress controller, so I'd think it makes sense to keep this logic generic.

test/e2e/operator_test.go
```diff
@@ -2585,6 +2585,93 @@ func TestCustomErrorpages(t *testing.T) {
 	}
 }
 
+// This test validates BZ2054200: Don't delete services that are not directly owned by this controller.
```
candita, line 2588: We usually do a unit test for bug fixes, and e2e tests for new features.  It doesn't hurt to have an e2e test, but a unit test should be done.

gcs278: I agree - but I don't think we can do a unit test here, right? I'm curious, can you unit test a function that makes an API (see r.client.Get...)? I think that requires a cluster.

pkg/operator/controller/ingress/load_balancer_service.go
candita: If this api call can return multiple services, then you need to query for a list and iterate over a list.

gcs278: Hmmm - I don't think that's possible with r.client.Get since names are unique.

test/e2e/operator_test.go
candita: You can mock out two services on the ic, one that is owned and one that is not.

pkg/operator/controller/ingress/load_balancer_service.go
candita: Isn't there an r.client.List?

test/e2e/operator_test.go
gcs278: So here's the function I made changes to and we'd want to unit test:

```
func (r *reconciler) currentNodePortService(ic *operatorv1.IngressController) (bool, *corev1.Service, error) {
	service := &corev1.Service{}
	if err := r.client.Get(context.TODO(), controller.NodePortServiceName(ic), service); err != nil {
		if errors.IsNotFound(err) {
			return false, nil, nil
		}
		return false, nil, err
	}

	// BZ2054200: Don't delete services that are not directly owned by this controller
	if service.Labels[manifests.OwningIngressControllerLabel] != ic.Name {
		return false, nil, nil
	}

	return true, service, nil
}

```
I don't see a way to mock out the two services when the function itself queries the API for the services. It would fail every time in a unit test because there is no API in a unit test right?

pkg/operator/controller/ingress/load_balancer_service.go
candita: Do you agree that it's possible to find more than once service attached to this ic?

candita: I believe this is what you need to use: 
https://github.com/kubernetes/client-go/blob/master/listers/core/v1/service.go#L82

gcs278: Hmm, I don't believe so. We are looking for a single service that the load_balancer_service.go with a unique name; there wouldn't ever be more than 1 service with the same name in the same namespace.

But I think I see potentially where you are headed. If the loadbalancer service already exists, but isn't owned by the ingress controller, then it will declare it doesn't have a service. The operator will go attempt to create it, but that create will fail because it already exists, leaving it in a broken state. I will fix that.

test/e2e/operator_test.go
candita: There's probably more complicated ways to do it, but if you extract the code you added into its own function, you can test it more easily by feeding it a mock service.  You'll want to make this iterate over of list of service, but you get the idea:

```
func checkIfServiceOwned(service corev1.Service, owner string)bool {
        if service.Labels[manifests.OwningIngressControllerLabel] != owner {
		return false
	}
        return true
}
```

gcs278:
/hold

@candita helped me realize there's a bug here

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Ah yea sorry! You were correct with your first comment. I think we only want to check when we delete otherwise this causes issues. I think we are still looking for a singular service object, I just had everything backwards.

gcs278: Okay there is definitely some nuance to this. I pushed up a new update, it wasn't quite that straight forward because if you put only wrapped the delete logic to avoid deleting, then it would indefinitely hang the deletion of the ingress controller because it still thought it had a service.

Also, if a service existed, the ingress controller would try to fix it and inevitably put it in a broken state. So it now just throws an error if you have an existing service with a name collision and it won't touch it.

test/e2e/operator_test.go
gcs278: Oh I see now. Yes - I will address this in my next update since I made it a separate function like you said. thanks!

gcs278:
/unhold

Addressed multiple things:
1. There wasn't any error message or indication there was a name conflict so I added those verbose messages in
2. There were issues with how the logic was originally written in that, the ingress operator thought it didn't have a service because it didn't own it, so it attempted to create it and failed, providing non-descriptive error messages. It now doesn't do anything if it doesn't own the service regardless if it wants to create it or delete it.
3. Added unit test for the new function that determines whether a service is owned by an ingress controller
4. Ensured that if a ingress controller had a service that it didn't own, it'd still be able to be deleted.

thanks @candita for pointing me in the right direction.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -526,10 +542,11 @@ func (r *reconciler) normalizeLoadBalancerServiceAnnotations(service *corev1.Ser
 // on existing resources.
 func (r *reconciler) finalizeLoadBalancerService(ci *operatorv1.IngressController) (bool, error) {
 	haveLBS, service, err := r.currentLoadBalancerService(ci)
+	isMyLBS := isServiceOwnedByIngressController(service, ci)
 	if err != nil {
 		return false, err
 	}
-	if !haveLBS {
+	if !haveLBS || !isMyLBS {
```
Miciah, line 549: We have an unusual situation here.  The `finalizeLoadBalancerService` method, and calls to it, can be deleted at this point because we changed the operator not to use the finalizer, and to remove it if it existed, in OpenShift 4.8.  See the comment on the `deleteLoadBalancerServiceFinalizer` method (which likewise can be deleted).  I can file a BZ and PR to remove this unnecessary code.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -312,6 +319,15 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 	return true, currentLBService, nil
 }
 
+// Determines whether a service is owned by an ingress controller.
+func isServiceOwnedByIngressController(service *corev1.Service, ci *operatorv1.IngressController) bool {
+	if service == nil || service.Labels[manifests.OwningIngressControllerLabel] != ci.Name {
+		return false
+	} else {
+		return true
+	}
```
Miciah, line 328: Slightly more idiomatic Go:
```suggestion
	if service != nil || service.Labels[manifests.OwningIngressControllerLabel] == ci.Name {
		return true
	}
	return true
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -261,9 +261,16 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
...
+	// BZ2054200: Don't modify/delete services that are not directly owned by this controller
+	ownLBS := isServiceOwnedByIngressController(currentLBService, ci)
+
 	switch {
 	case !wantLBS && !haveLBS:
 		return false, nil, nil
+	case !wantLBS && haveLBS && !ownLBS:
+		return false, nil, fmt.Errorf("refusing to delete load balancer service not owned by ingress controller: %s", controller.LoadBalancerServiceName(ci))
+	case wantLBS && haveLBS && !ownLBS:
+		return false, nil, fmt.Errorf("refusing to update load balancer service not owned by ingress controller: %s", controller.LoadBalancerServiceName(ci))
```
Miciah, line 273: The switch statement follows a pattern that appears throughout cluster-ingress-operator and cluster-dns-operator, and I'd prefer we stick to that pattern for consistency.  Could we instead modify the existing cases as follows?
```go
	case !wantLBS && haveLBS:
		if !ownLBS {
			return false, nil, fmt.Errorf("refusing to delete load balancer service not owned by ingress controller: %s", controller.LoadBalancerServiceName(ci))
		}
		// [existing logic]
	case wantLBS && !haveLBS:
		// [existing logic]
	case wantLBS && haveLBS:
		if !ownLBS {
			return false, nil, fmt.Errorf("refusing to update load balancer service not owned by ingress controller: %s", controller.LoadBalancerServiceName(ci))
		}
		// [existing logic]
```

Miciah: By the way, what actually happens with this logic when the service exists but isn't owned by the operator?  If `ensureLoadBalancerService` returns an error, I believe the reconciliation loop will retry, correct?  Does the reconciliation loop retry a bunch of times with backoff?

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: https://github.com/openshift/cluster-ingress-operator/pull/729

gcs278: Ah yes I saw those comments about it not really being necessary. Thanks for cleaning that up.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Ah yea that's fair. I kept toying over the best way to keep logic consistent, but I agree with your suggestion.

As best as I can see, if `ensureLoadBalancerService` returns an error, it will retry if ALL errors in list are type `retryable.Error`. However, if there is one error that is not, it doesn't retry. This logic seems funny to me. So our error that we are returning won't make it constantly retry. In fact, it will prevent it from retrying if there is another `retryable.Error` in the error list. See:
https://github.com/openshift/cluster-ingress-operator/blob/542124811a01039852d8b0846da91d33665037f8/pkg/util/retryableerror/retryableerror.go#L40

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: TIL GO this style. checks out :+1: 

Got it - `||` should be `&&` but I understand.

gcs278: TIL GO this style. checks out :+1: 

Got it - `||` should be `&&` but I understand.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Done

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Done

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: > As best as I can see, if `ensureLoadBalancerService` returns an error, it will retry if ALL errors in list are type `retryable.Error`. However, if there is one error that is not, it doesn't retry. 

That isn't completely accurate.  The `Reconciler` interface has nonintuitive semantics:

```go
type Reconciler interface {
	// Reconcile performs a full reconciliation for the object referred to by the Request.
	// The Controller will requeue the Request to be processed again if an error is non-nil or
	// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
	Reconcile(context.Context, Request) (Result, error)
}
```
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

```go
type Result struct {
	// Requeue tells the Controller to requeue the reconcile key.  Defaults to false.
	Requeue bool

	// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
	// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
	RequeueAfter time.Duration
}
```
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result

(Even that documentation is confusing—the `Reconciler` godoc doesn't mention `RequeueAfter`.  I guess the case could be made that the godoc isn't flat-out _wrong_ as the `RequeueAfter` godoc says that `RequeueAfter` "Implies that Requeue is true".  It's darned confusing though and should be fixed IMO.  Good first issue?)

The operator's `Reconcile` method returns `reconcile.Result{RequeueAfter: e.After()}, nil` if `err` is a `retryable.Error`-type value.  Otherwise, `Reconcile` returns `reconcile.Result{}, err`, which **_still_** causes a retry if `err` is not nil.

Miciah: Anyway, that's all to say that I still believe that returning an error here in case of a conflicting service will cause a bunch of spurious reconciliations.  That doesn't mean that we absolutely cannot use the approach you are using here, especially if the alternative would require significant changes to the operator's design, but I want to make sure we understand the implications of this change, and that the implications aren't _too_ onerous.

gcs278: Hmmm, yea you are right. I missed that, thought it was only based on RequeueAfter. 

So I think we both agree some type of message is appropriate here. What about doing a log.Info() or log.Error() without returning it as an error? Would that be enough?

Do we want intentionally want it to continually try to reconcile? In my testing:
1. I create a conflicting service
2. Create the ingress controller
3. Deleted the conflicting service
4. I found that it eventually reconciles and recreates the service that is now not longer under conflict.

gcs278: @Miciah thoughts?

gcs278: Revisiting this now that I know more about how `watch` in operators work. The issue is: if we don't return an error and reconcile again, it may never reconcile, since deleting a service that isn't owned by the ingresscontroller, will not trigger a reconcilation.

So based on @Miciah's comment "it's better to reconcile more than you need than less than you need", I think the logic should stay the way it is, especially since this is a bit of an edge case. Constant errors make sense because the ingress controller isn't able to do it's job, and justifiably should be complaining often.

Understood the implications.

gcs278:
My e2e failed, but I'm not sure how:
`operator_test.go:2805: expected openshift-ingress/router-nodeport-e2e-name-collision-test to be present: rpc error: code = Unknown desc = malformed header: missing HTTP content-type`

I don't think I have any control over this and must have been a flake.

/retest

quarterpin:
/label qe-approved
Verified via pre-merge test process. For more details refer to: https://bugzilla.redhat.com/show_bug.cgi?id=2054200#c6

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -260,10 +260,16 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
...
 
+	// BZ2054200: Don't modify/delete services that are not directly owned by this controller
+	ownLBS := isServiceOwnedByIngressController(currentLBService, ci)
+
 	switch {
 	case !wantLBS && !haveLBS:
 		return false, nil, nil
 	case !wantLBS && haveLBS:
+		if !ownLBS {
+			return false, nil, fmt.Errorf("refusing to delete load balancer service not owned by ingress controller: %s", controller.LoadBalancerServiceName(ci))
```
Miciah, line 271: I think this wording is a little clearer:
```suggestion
			return false, nil, fmt.Errorf("a conflicting load balancer service exists that is not owned by the ingresscontroller: %s", controller.LoadBalancerServiceName(ci))
```
Likewise for the other 3 similar error messages.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -295,6 +304,14 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 	return true, currentLBService, nil
 }
 
+// isServiceOwnedByIngressController determines whether a service is owned by an ingress controller.
+func isServiceOwnedByIngressController(service *corev1.Service, ci *operatorv1.IngressController) bool {
+	if service != nil && service.Labels[manifests.OwningIngressControllerLabel] == ci.Name {
```
Miciah, line 309: We should avoid using `ci` in new code (we renamed `ClusterIngress` to `IngressController` long ago).
```suggestion
func isServiceOwnedByIngressController(service *corev1.Service, ic *operatorv1.IngressController) bool {
	if service != nil && service.Labels[manifests.OwningIngressControllerLabel] == ic.Name {
```

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -736,3 +737,66 @@ func TestLoadBalancerServiceChanged(t *testing.T) {
...
+	for _, tc := range testCases {
+		ci := &operatorv1.IngressController{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: tc.ingressName,
+			},
+			Spec:   operatorv1.IngressControllerSpec{},
+			Status: operatorv1.IngressControllerStatus{},
+		}
+
+		if actual := isServiceOwnedByIngressController(tc.service, ci); actual != tc.expect {
```
Miciah, line 797:
```suggestion
		ic := &operatorv1.IngressController{
			ObjectMeta: metav1.ObjectMeta{
				Name: tc.ingressName,
			},
			Spec:   operatorv1.IngressControllerSpec{},
			Status: operatorv1.IngressControllerStatus{},
		}

		if actual := isServiceOwnedByIngressController(tc.service, ic); actual != tc.expect {
```

test/e2e/operator_test.go
```diff
@@ -2723,6 +2723,93 @@ func TestTunableRouterKubeletProbesForCustomIngressController(t *testing.T) {
...
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Wait for new ingress controller to come online.
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Create services that could collide with the new ingress controller's naming convention for loadbalancer and nodeport.
+	// TRICKY: Our private ingress controller will reconcile (delete) extra loadBalancerServices, even though it isn't one.
+	echoLoadBalancerServiceName := types.NamespacedName{
```
Miciah, line 2751: Why "echo"?  Copypasta, or maybe you meant to write "extra"?  In other tests where "echo" appears, the tests are setting up an echo server (that is, a server that echoes back client requests).  

I see you use `buildEchoService` below, and that's fine (it happens to return a suitable enough service definition).  However, the "echo" in the variable names is a little confusing.

test/e2e/operator_test.go
```diff
@@ -2723,6 +2723,93 @@ func TestTunableRouterKubeletProbesForCustomIngressController(t *testing.T) {
...
+	echoNodeportService := buildEchoService(echoNodeportServiceName.Name, echoNodeportServiceName.Namespace, nil)
+	if err := kclient.Create(context.TODO(), echoNodeportService); err != nil {
+		t.Fatalf("failed to create service %s: %v", echoNodeportServiceName, err)
+	}
+
+	// Clean up our new services after we are done.
+	defer func() {
+		if err := kclient.Delete(context.TODO(), echoLoadBalancerService); err != nil {
+			t.Fatalf("failed to delete service %s: %v", echoLoadBalancerServiceName, err)
+		}
```
Miciah, line 2773: It would be slightly safer to use two separate deferred functions and put the deferred cleanup for the load-balancer service right after the creation of that service (and, more importantly, before the creation of the nodeport service, which could fail and bail out).

test/e2e/operator_test.go
```diff
@@ -2723,6 +2723,93 @@ func TestTunableRouterKubeletProbesForCustomIngressController(t *testing.T) {
...
+	}
+
+	// Wait to see if our service gets deleted by the operator due to name collision.
+	oldLoadBalancerUID := echoLoadBalancerService.UID
+	oldNodePortUID := echoNodeportService.UID
+	wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		// First Check our LoadBalancer Service.
+		if !assertServiceNotDeleted(t, echoLoadBalancerServiceName, oldLoadBalancerUID) {
+			return true, nil
+		}
```
Miciah, line 2802: `!assertServiceNot` is a double-negative.  We should avoid double-negatives, to improve clarity, if doing so isn't too difficult.  In this case, you don't really need the return value because `assertServiceNotDeleted` calls `f.Fatalf` itself if the assertion fails, so you can remove the double-negative and simplify the logic:
```suggestion
		assertServiceNotDeleted(t, echoLoadBalancerServiceName, oldLoadBalancerUID)
```

test/e2e/operator_test.go
```diff
@@ -2987,6 +3074,30 @@ func deleteIngressController(t *testing.T, cl client.Client, ic *operatorv1.Ingr
...
+		return false
+	}
+	// If the UID has changed, then the service has been recreated.
+	if service.UID != oldUid {
+		t.Fatalf("expected service %s to have UID %v, got %v", serviceName, oldUid, service.UID)
+		return false
+	}
+
+	return true
+}
```
Miciah, line 3099:
```suggestion
func assertServiceNotDeleted(t *testing.T, serviceName types.NamespacedName, oldUid types.UID) {
	t.Helper()

	// First Check our LoadBalancer Service.
	service := &corev1.Service{}
	if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
		t.Fatalf("expected %s to be present: %v", serviceName, err)
	}
	// If there is a DeletionTimestamp, it has been marked for deletion.
	if service.DeletionTimestamp != nil {
		t.Fatalf("expected service %s to not be marked for deletion: %v", serviceName, service.DeletionTimestamp)
	}
	// If the UID has changed, then the service has been recreated.
	if service.UID != oldUid {
		t.Fatalf("expected service %s to have UID %v, got %v", serviceName, oldUid, service.UID)
	}
}
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -260,10 +260,16 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 		return false, nil, err
 	}
 
+	// BZ2054200: Don't modify/delete services that are not directly owned by this controller
```
Miciah, line 263:
```suggestion
	// BZ2054200: Don't modify/delete services that are not directly owned by this controller.
```
(Sorry, I know I'm being picky.)

Miciah (commented):
Looks good overall.  I have a few small suggestions.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Agreed. Done.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Done

pkg/operator/controller/ingress/load_balancer_service_test.go
gcs278: Done.

test/e2e/operator_test.go
gcs278: Sure yea it's not very descriptive. Done.

test/e2e/operator_test.go
gcs278: Agreed. Done.

test/e2e/operator_test.go
gcs278: Ah good point. Done.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Done and in nodeport_service.go as well.

test/e2e/operator_test.go
gcs278: Agree. Removed the if statement all together so it's just 1 negative now.

PR #708, opened 2022-02-25
Bug 2057762: Set Upgradeable=False if default cert has no SAN

Miciah:
If an ingresscontroller's default certificate has a Common Name (CN) for the ingress domain but has no Subject Alternative Name (SAN) for the same, report that the cluster cannot be upgraded by setting the `Upgradeable=False` status condition on the ingresscontroller and clusteroperator.

Clients built using Go 1.17 reject certificates without SANs.  OpenShift 4.10 is built using Go 1.17, which means that various operators that connect to routes that use the default certificate would reject the certificate and fail to complete the TLS handshake after upgrading to OpenShift 4.10 if the ingress operator didn't block the upgrade on a cluster with a problematic certificate.

* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Get the default certificate secret and pass it to `computeIngressUpgradeableCondition`.
(`computeIngressUpgradeableCondition`): Add a parameter for the default certificate secret.  Use the argument value to call the new `checkDefaultCertificate` function to check for problematic certificates.
(`checkDefaultCertificate`): New function.  Return a non-nil error value if the default certificate in the provided secret has a CN for the ingress domain and no SAN for the same.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressUpgradeableCondition`): Verify that `computeIngressUpgradeableCondition` reports `Upgradeable=False` if the default certificate has a CN and no SAN for the ingress domain and reports `Upgradeable=True` otherwise.

---

This PR needs to be backport to 4.9, after it can be reverted in 4.11 and 4.10.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -2,6 +2,8 @@ package ingress
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"reflect"
@@ -12,6 +14,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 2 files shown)


Miciah:
I anticipate that conflicts will necessitate a manual cherry-pick, but now that https://github.com/openshift/cluster-ingress-operator/pull/709 has merged, maybe an automated merge will succeed...
/cherry-pick release-4.9

PR #709, opened 2022-02-25
Bug 2058699: Upgradeable Condition in Operator and IC status

arjunrn:
This change sets the `Upgradeable` condition for the IngressController. If the corresponding load balancer _Service_ has been modified, particularly the AWS resource tags fields then the IngressController is marked as `Upgradable=False`. This in turn marks the operator as `Upgradeable=False`.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -11,12 +11,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
-	iov1 "github.com/openshift/api/operatoringress/v1"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	iov1 "github.com/openshift/api/operatoringress/v1"
+
 	oputil "github.com/openshift/cluster-ingress-operator/pkg/util"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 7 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -66,6 +68,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions, updated.Name)
 	errs = append(errs, err)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, infraConfig.Status.PlatformStatus))
```
Miciah, line 71: Oof, I missed this during review—`infraConfig.Status` could be nil.  That's why we use the `GetPlatformStatus` helper.  I'll fix this in my backport of #710.

Miciah: https://github.com/openshift/cluster-ingress-operator/pull/711/commits/654c1a9f8a80cfcee59ecbe645764a292f0d0e5c

Miciah: On further investigation, `infraConfig.Status` cannot be nil, but `infraConfig.Status.PlatformStatus` can be, and the consequence of passing a nil value for `plsatform` to `computeIngressUpgradeableCondition` would be the following:

* `computeIngressUpgradeableCondition` passes the nil value to `loadBalancerServiceIsUpgradeable`.
* `loadBalancerServiceIsUpgradeable` passes the nil value to `desiredLoadBalancerService`.
* `desiredLoadBalancerService` passes the nil value to `IsProxyProtocolNeeded`.
* `IsProxyProtocolNeeded` checks for nil and returns an error value.
* The error value gets returned up through `desiredLoadBalancerService` and `loadBalancerServiceIsUpgradeable` to `computeIngressUpgradeableCondition`.
* `computeIngressUpgradeableCondition` reports `Upgradeable=False`.

Thus, if my analysis is correct, this issue will inappropriately block upgrades on born-as-4.1 clusters.

Miciah: @wking pointed https://github.com/openshift/cluster-config-operator/pull/174 out to me; this change in cluster-config-operator should prevent this issue with born-as-4.1 clusters.

PR #711, opened 2022-03-02
[release-4.9] Bug  2060111: Set Upgradeable=False if default cert has no SAN

Miciah:
This is a manual cherry-pick of https://github.com/openshift/cluster-ingress-operator/pull/708 with an additional fix for an issue introduced in #709 (see https://github.com/openshift/cluster-ingress-operator/pull/709#discussion_r817929252).  

---

#### status: Use `GetPlatformStatus`

Don't use `infraConfig.Status.PlatformStatus` directly; `infraConfig.Status` could be nil.

Follow-up to #709.

* `pkg/operator/controller/ingress/status.go`: Use `GetPlatformStatus`.

#### Set `Upgradeable=False` if default cert has no SAN

If an ingresscontroller's default certificate has a Common Name (CN) for the ingress domain but has no Subject Alternative Name (SAN) for the same, report that the cluster cannot be upgraded by setting the `Upgradeable=False` status condition on the ingresscontroller and clusteroperator.

Clients built using Go 1.17 reject certificates without SANs.  OpenShift 4.10 is built using Go 1.17, which means that various operators that connect to routes that use the default certificate would reject the certificate and fail to complete the TLS handshake after upgrading to OpenShift 4.10 if the ingress operator didn't block the upgrade on a cluster with a problematic certificate.

* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Get the default certificate secret and pass it to `computeIngressUpgradeableCondition`.
(`computeIngressUpgradeableCondition`): Add a parameter for the default certificate secret.  Use the argument value to call the new `checkDefaultCertificate` function to check for problematic certificates.
(`checkDefaultCertificate`): New function.  Return a non-nil error value if the default certificate in the provided secret has a CN for the ingress domain and no SAN for the same.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressUpgradeableCondition`): Verify that `computeIngressUpgradeableCondition` reports `Upgradeable=False` if the default certificate has a CN and no SAN for the ingress domain and reports `Upgradeable=True` otherwise.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -52,6 +52,11 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
 	}
 
+	platform, err := oputil.GetPlatformStatus(r.client, infraConfig)
+	if err != nil {
+		return fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %w", ic.Namespace, ic.Name, err)
+	}
+
 	var errs []error
 
 	updated := ic.DeepCopy()
@@ -68,7 +73,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions, updated.Name)
 	errs = append(errs, err)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, infraConfig.Status.PlatformStatus))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, platform))
 	updated.Status.Conditions = PruneConditions(updated.Status.Conditions)
 
 	if !IngressStatusesEqual(updated.Status, ic.Status) {
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -2,6 +2,8 @@ package ingress
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"reflect"
@@ -12,6 +14,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -23,6 +26,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
@@ -57,6 +61,12 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		return fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %w", ic.Namespace, ic.Name, err)
 	}
 
+	secret := &corev1.Secret{}
+	secretName := controller.RouterEffectiveDefaultCertificateSecretName(ic, deployment.Namespace)
+	if err := r.client.Get(context.TODO(), secretName, secret); err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get the default certificate secret %s for ingresscontroller %s/%s: %w", secretName, ic.Namespace, ic.Name, err)
+	}
+
 	var errs []error
 
 	updated := ic.DeepCopy()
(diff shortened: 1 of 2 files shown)


Miciah:
`TestConfigurableRouteNoConsumingUserNoRBAC` [failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/711/pull-ci-openshift-cluster-ingress-operator-release-4.9-e2e-aws-operator/1499145014159085568); the same test [failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/707/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1498771366109777920) on #707, so it looks like the test may be flaky.

/test e2e-aws-operator

melvinjoseph86:
https://bugzilla.redhat.com/show_bug.cgi?id=2060111#c3
/label qe-approved

Miciah:
Must-gather failed.
/test e2e-aws-operator

Miciah:
Installer failed:
```
level=error msg=Error: error updating LB Target Group (arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-9ddchgy8-265e5-qlmhd-aint/638dbe06869a2be6) tags: error tagging resource (arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-9ddchgy8-265e5-qlmhd-aint/638dbe06869a2be6): TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-9ddchgy8-265e5-qlmhd-aint/638dbe06869a2be6' not found
level=error msg=	status code: 400, request id: ca03fac1-2eea-451c-9a63-3d90143670ac
level=error
level=error msg=  on ../tmp/openshift-install-cluster-393099372/vpc/master-elb.tf line 45, in resource "aws_lb_target_group" "api_internal":
level=error msg=  45: resource "aws_lb_target_group" "api_internal" {
level=error
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
```
/test e2e-aws-operator

pkg/operator/controller/ingress/status.go
```diff
@@ -556,6 +574,50 @@ func computeIngressUpgradeableCondition(ic *operatorv1.IngressController, deploy
...
+		if err != nil {
+			continue
+		}
+		foundSAN := false
+		for i := range cert.DNSNames {
+			if cert.DNSNames[i] == domain {
+				foundSAN = true
+			}
+		}
+		if cert.Subject.CommonName == domain && !foundSAN {
```
candita, line 613: Is there always going to be a common name, even if there is a SAN?

pkg/operator/controller/ingress/status.go
```diff
@@ -556,6 +574,50 @@ func computeIngressUpgradeableCondition(ic *operatorv1.IngressController, deploy
...
+// certificate specifies a Subject Alternative Name (SAN) for the ingress
+// domain, then it is safe to upgrade, and the return value is nil.  Otherwise,
+// if the certificate has a legacy Common Name (CN) for the ingress domain and
+// no SAN for the same, then the return value is an error indicating that the
+// certificate must be replaced by one with a SAN before upgrading is allowed.
+// This check is necessary because OpenShift 4.10 is built using Go 1.17, which
+// rejects certificates without SANs.  Note that this function only checks the
+// validity of the certificate insofar as it affects upgrades.  See
+// <https://bugzilla.redhat.com/show_bug.cgi?id=2057762>.  This check can be
+// removed after OpenShift 4.9.
```
candita, line 588: This is in 4.9, right?

```suggestion
// <https://bugzilla.redhat.com/show_bug.cgi?id=2057762>.  
```

pkg/operator/controller/ingress/status.go
Miciah: No, but what we really care about is the situation where a serving certificate is accepted by older clients and rejected by newer clients.  Checking that the CN matches the ingresscontroller domain and that no SAN does is the best approach I've come up with to warn when there is a problem without causing false positives.  Consider these four cases:

* The certificate has no SAN and no CN; such a certificate wouldn't work at all as a serving certificate even for older clients, so we should ignore it.
* The certificate has no CN and does have a SAN; this is fine: if the SAN is valid, the certificate will be accepted by both old clients as well as new clients (and if the SAN is invalid, it won't be accepted by old or new clients), so we should ignore it.
* The certificate has a CN that doesn't match the ingresscontroller domain; in this case, the certificate is presumably not the serving certificate but rather an intermediate, and anyway, as with the previous case, either it has a valid SAN, in which case both old as well as new clients will accept it, or it does not, in which case both old as well as new clients will reject it, so we should ignore it.
* The certificate has a CN that does match the ingresscontroller domain; in this case, the certificate most likely is the serving certificate, and we need to make sure it has a SAN with the same domain in order to still be accepted by newer Go clients.  

The last case is really the only one that matters for upgrades, so it is the only one where we set `Upgradeable=True` if we find a mismatch.

pkg/operator/controller/ingress/status.go
Miciah: Yes, this is the 4.9 backport.  Usually I keep the backport as close as possible to the original change (in this case, #708) to minimize risk of breaking something and to simplify future backports.

candita: I was thinking that if someone saw "This check can be removed after OpenShift 4.9." that they might remove it in a later 4.9 version.  Or it would at least cause some questions.

Miciah: Hm, I guess someone could interpret "after OpenShift 4.9" as including newer z-stream releases _of_ 4.9.  Maybe I should have said "in OpenShift 4.10 or later" in the original commit.  However, even if someone interpreted "after OpenShift 4.9" in that way, I think the risk that someone would remove this in a 4.9 z-stream is low since we have extra levels of review and scrutiny for backports.

candita:
> level=error msg=Bootstrap failed to complete: Get "https://api.ci-op-5g2hxix6-265e5.origin-ci-int-aws.dev.rhcloud.com:6443/version?timeout=32s": dial tcp 23.22.248.35:6443: connect: connection refused
...
> level=info msg=Pull failed. Retrying registry.build02.ci.openshift.org/ci-op-5g2hxix6/release@sha256:745b411dc13d6fe198c528e91f77c8c1be6f108e7d9a99283bc84ea5464dfedc...
> level=info msg=Error: Error initializing source docker://registry.build02.ci.openshift.org/ci-op-5g2hxix6/release@sha256:745b411dc13d6fe198c528e91f77c8c1be6f108e7d9a99283bc84ea5464dfedc: Error reading manifest sha256:745b411dc13d6fe198c528e91f77c8c1be6f108e7d9a99283bc84ea5464dfedc in registry.build02.ci.openshift.org/ci-op-5g2hxix6/release: unauthorized: authentication required
> level=fatal msg=Bootstrap failed to complete

/retest

pkg/operator/controller/ingress/status.go
Miciah: I've amended the comment.

PR #713, opened 2022-03-03
Bug 2079517: Use externalTrafficPolicy: Cluster with OVN

Miciah:
OVN-Kubernetes in OpenShift 4.9 and earlier does not support `externalTrafficPolicy: Local`, and specifying it is reported to result in imbalanced traffic for some users.  This change checks the cluster network type and sets `externalTrafficPolicy: Cluster` on the LoadBalancer-type service that the operator creates for ingress if the network type is "OVN-Kubernetes".

* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Pass the network config to `ensureLoadBalancerService` and `syncIngressControllerStatus`.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`ensureLoadBalancerService`, `loadBalancerServiceIsUpgradeable`): Add a parameter for the network config.  Pass the network type from the network config to `desiredLoadBalancerService`.
(`desiredLoadBalancerService`): Add a parameter for the network type.  Set `externalTrafficPolicy: Cluster` if the network type is "OVN-Kubernetes".
* `pkg/operator/controller/ingress/load_balancer_service_test.go` (`TestDesiredLoadBalancerService`): Add a test case for OVN-Kubernetes.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Add a parameter for the network config.  Pass the config to `computeIngressUpgradeableCondition`.
(`computeIngressUpgradeableCondition`): Add a parameter for the network config.  Pass the network config to `loadBalancerServiceIsUpgradeable`.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputeIngressUpgradeableCondition`): Pass the network type to `desiredLoadBalancerService` and the network config to `computeIngressUpgradeableCondition`.

---

This PR includes commits from #711 because we want #711 to merge first, and it includes conflicting changes.  The first two commits in this PR can be ignored by reviewers.  

/hold

---

Surya, can you verify that this change is correct?  In particular, I want to be sure we're checking the right API field for the right value and that the logic behind this PR is sound at a high level.  

/assign @tssurya

### DIFF
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -156,12 +156,12 @@ var (
 // ensureLoadBalancerService creates an LB service if one is desired but absent.
 // Always returns the current LB service if one exists (whether it already
 // existed or was created during the course of the function).
-func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController, deploymentRef metav1.OwnerReference, infraConfig *configv1.Infrastructure) (bool, *corev1.Service, error) {
+func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController, deploymentRef metav1.OwnerReference, infraConfig *configv1.Infrastructure, networkConfig *configv1.Network) (bool, *corev1.Service, error) {
 	platform, err := oputil.GetPlatformStatus(r.client, infraConfig)
 	if err != nil {
 		return false, nil, fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %v", ci.Namespace, ci.Name, err)
 	}
-	wantLBS, desiredLBService, err := desiredLoadBalancerService(ci, deploymentRef, platform)
+	wantLBS, desiredLBService, err := desiredLoadBalancerService(ci, deploymentRef, platform, networkConfig.Spec.NetworkType)
 	if err != nil {
 		return false, nil, err
 	}
@@ -222,7 +222,7 @@ func (r *reconciler) ensureLoadBalancerService(ci *operatorv1.IngressController,
 // ingresscontroller, or nil if an LB service isn't desired. An LB service is
 // desired if the high availability type is Cloud. An LB service will declare an
 // owner reference to the given deployment.
-func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef metav1.OwnerReference, platform *configv1.PlatformStatus) (bool, *corev1.Service, error) {
+func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef metav1.OwnerReference, platform *configv1.PlatformStatus, networkType string) (bool, *corev1.Service, error) {
 	if ci.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
 		return false, nil, nil
 	}
@@ -268,6 +268,12 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef 
 					service.Annotations[GCPGlobalAccessAnnotation] = strconv.FormatBool(globalAccessEnabled)
 				}
 			}
+		}
+		// externalTrafficPolicy: Local is unsupported by OVN in
+		// OpenShift 4.9 and earlier; see
+		// <https://bugzilla.redhat.com/show_bug.cgi?id=2060542>.
+		if networkType == string(operatorv1.NetworkTypeOVNKubernetes) {
+			service.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeCluster
 		}
 		switch platform.Type {
 		case configv1.AWSPlatformType:
@@ -560,8 +566,8 @@ func loadBalancerServiceTagsModified(current, expected *corev1.Service) (bool, *
 // return value is nil.  Otherwise, if something or someone else has modified
 // the service, then the return value is a non-nil error indicating that the
 // modification must be reverted before upgrading is allowed.
-func loadBalancerServiceIsUpgradeable(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference, current *corev1.Service, platform *configv1.PlatformStatus) error {
-	want, desired, err := desiredLoadBalancerService(ic, deploymentRef, platform)
+func loadBalancerServiceIsUpgradeable(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference, current *corev1.Service, platform *configv1.PlatformStatus, networkConfig *configv1.Network) error {
+	want, desired, err := desiredLoadBalancerService(ic, deploymentRef, platform, networkConfig.Spec.NetworkType)
 	if err != nil {
 		return err
 	}
--- a/pkg/operator/controller/ingress/load_balancer_service_test.go
+++ b/pkg/operator/controller/ingress/load_balancer_service_test.go
@@ -23,7 +23,9 @@ func TestDesiredLoadBalancerService(t *testing.T) {
 		proxyNeeded          bool
 		expect               bool
 		platformStatus       configv1.PlatformStatus
+		networkType          string
 		expectedResourceTags string
+		expectedETP          string
 	}{
 		{
 			description:  "external classic load balancer with scope for aws platform",
@@ -159,7 +161,8 @@ func TestDesiredLoadBalancerService(t *testing.T) {
 			lbStrategy: operatorv1.LoadBalancerStrategy{
 				Scope: operatorv1.ExternalLoadBalancer,
 			},
-			expect: true,
+			expect:      true,
+			expectedETP: "Cluster",
 		},
 		{
 			description:  "internal load balancer for ibm platform",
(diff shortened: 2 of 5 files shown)


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -261,6 +261,12 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 				}
 			}
 		}
+		// externalTrafficPolicy: Local is unsupported by OVN in
+		// OpenShift 4.9 and earlier; see
+		// <https://bugzilla.redhat.com/show_bug.cgi?id=2060542>.
+		if networkType == "OVN-Kubernetes" {
```
tssurya, line 267: its `OVNKubernetes`, we should probably hardcode tha value as a constant to avoid the typo?
reference: https://github.com/openshift/api/blob/fb6f933bb8d5ce8454a8777c0c4782c193ef5674/operator/v1/types_network.go#L536

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -524,8 +530,8 @@ func loadBalancerServiceTagsModified(current, expected *corev1.Service) (bool, *
 // return value is nil.  Otherwise, if something or someone else has modified
 // the service, then the return value is a non-nil error indicating that the
 // modification must be reverted before upgrading is allowed.
-func loadBalancerServiceIsUpgradeable(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference, current *corev1.Service, platform *configv1.PlatformStatus) error {
-	want, desired, err := desiredLoadBalancerService(ic, deploymentRef, platform)
+func loadBalancerServiceIsUpgradeable(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference, current *corev1.Service, platform *configv1.PlatformStatus, networkConfig *configv1.Network) error {
```
tssurya, line 533: @danwinship or @squeed : I have always been confused as to when to use the operv1 versus configv1 api for things. Could you please configm if using configv1 ^ is ok here? Seems like for config we support sdn and ovn: https://github.com/openshift/api/blob/d5252bac47154049f80d49eff7fec5bb642be9af/config/v1/types_network.go#L145 and for operator we support the other ones like kuryr etc..sorry for the dumb question, I probably should be knowing these things by now.

tssurya (changes requested):

tssurya:
thanks @Miciah for the quick fix, appreciate it!

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Thanks!  I didn't think to check for named constants because the `Spec.NetworkType` field of the `Network` type in `config/v1/types_network.go` is a string type.  I wonder whether the field could be changed to use the `NetworkType` type; I suppose if the same values are allowed, it may be all right to change the type, although it would require changes in any Go code using the Go type definitions.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: It is confusing that `operator/v1/types_network.go` defines these named constants, but `config/v1/types_network.go` just uses a string type.

Miciah:
Bootstrap failures.
/retest

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -261,6 +261,12 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 				}
 			}
 		}
+		// externalTrafficPolicy: Local is unsupported by OVN in
+		// OpenShift 4.9 and earlier; see
+		// <https://bugzilla.redhat.com/show_bug.cgi?id=2060542>.
+		if networkType == string(operatorv1.NetworkTypeOVNKubernetes) {
+			service.Spec.ExternalTrafficPolicy = "Cluster"
```
tssurya, line 268: nit: we could use the https://github.com/kubernetes/kubernetes/blob/e6c093d87ea4cbb530a7b2ae91e54c0842d8308a/pkg/apis/core/types.go#L3672 constant from upstream

pkg/operator/controller/ingress/load_balancer_service.go
squeed: @tssurya good question! In this case, it doesn't really matter, because configv1 NetworkType is inferred from configv1. Since configv1 is guaranteed to be set from cluster creation, it is slightly better to use. But it really, really, really doesn't matter.

squeed: > It is confusing that `operator/v1/types_network.go` defines these named constants [but configv1 doesn't]

That's because you can run a cluster with a third-party network operator - so we need to support arbitrary network types in configv1. But operv1 defines what the openshift cluster network operator supports.

tssurya: >  because configv1 NetworkType is inferred from configv1. 

you meant from operv1?

> Since configv1 is guaranteed to be set from cluster creation, it is slightly better to use. But it really, really, really doesn't matter. 

ack thanks casey! makes sense. @Miciah so I think we are good here.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Fixed.  Thanks!

Miciah:
Failed to launch the cluster:
```
level=error msg=Error: error listing tags for LB Target Group (arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-fjqjgw7q-265e5-b5vk5-sint/84bf5875e5bc1488): TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-1:460538899914:targetgroup/ci-op-fjqjgw7q-265e5-b5vk5-sint/84bf5875e5bc1488' not found
level=error msg=	status code: 400, request id: d43be720-0c69-49e4-9094-98e9ad332ddb
level=error
level=error msg=  on ../tmp/openshift-install-cluster-733658679/vpc/master-elb.tf line 99, in resource "aws_lb_target_group" "services":
level=error msg=  99: resource "aws_lb_target_group" "services" {
level=error
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
```
/test e2e-aws-operator

Miciah:
#711 merged, and I've rebased this PR; the ignoring the changes from #711, the diff for this PR is exactly the same but for some line numbers (compare <https://github.com/openshift/cluster-ingress-operator/commit/15202ad247d0dedaadec7c2564b60db9800f50ff.patch> and <https://github.com/openshift/cluster-ingress-operator/commit/7b3de002a1bf5373504d2b5dde2985cec012aecf.patch>).  
/hold cancel

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -320,8 +344,10 @@ func TestDesiredLoadBalancerService(t *testing.T) {
 				if err := checkServiceHasAnnotation(svc, awsLBHealthCheckHealthyThresholdAnnotation, true, awsLBHealthCheckHealthyThresholdDefault); err != nil {
 					t.Errorf("annotation check for test %q failed: %v", tc.description, err)
 				}
-				if err := checkServiceHasAnnotation(svc, localWithFallbackAnnotation, true, ""); err != nil {
-					t.Errorf("local-with-fallback annotation check for test %q failed: %v", tc.description, err)
+				if svc.Spec.ExternalTrafficPolicy == "Local" {
```
candita, line 347: Shouldn't this be `if svc.Spec.ExternalTrafficPolicy != "Cluster"` ?

pkg/operator/controller/ingress/status_test.go
```diff
@@ -1312,7 +1312,12 @@ func TestComputeIngressUpgradeableCondition(t *testing.T) {
 					},
 				},
 			}
-			wantSvc, service, err := desiredLoadBalancerService(ic, deploymentRef, platformStatus)
+			networkConfig := &configv1.Network{
+				Spec: configv1.NetworkSpec{
+					NetworkType: "OpenShiftSDN",
+				},
+			}
```
candita, line 1319: Normally we would add a new variable to the test case itself, and add a couple of test cases to test the variant values.

candita:
@miciah was there any update for e2e tests for this?

candita:
Error: Failed to download metadata for repo 'localdev-rhel-8-server-ose-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
...
No match for argument: skopeo
Error: Unable to find a match: skopeo

/test e2e-gcp-serial

candita:
>  fail [github.com/openshift/origin/test/extended/oauth/requestheaders.go:204]: unexpected error response status (401) while trying to reach '/oauth/authorize?client_id=openshift-challenging-client&response_type=token' endpoint: HTTP/1.1 401 Unauthorized
> ...
> Referrer-Policy: strict-origin-when-cross-origin
> Warning: 199 Origin "A non-empty X-CSRF-Token header is required to receive basic-auth challenges"
> ...
> X-Frame-Options: DENY
> X-Xss-Protection: 1; mode=block
> A non-empty X-CSRF-Token header is required to receive basic-auth challenges
> failed: (2m23s) 2022-04-28T22:33:08 "[Serial] [sig-auth][Feature:OAuthServer] [RequestHeaders] [IdP] test RequestHeaders IdP [Suite:openshift/conformance/serial]"
>

lihongan:
verified with pre-merge test process, see https://bugzilla.redhat.com/show_bug.cgi?id=2079517#c1
/label qe-approved

PR #715, opened 2022-03-03
Bug 1995953: Add unit test for verifying router deployment env is always sorted

gcs278:
Unit test to check all EndpointPublishingStrategys sort the container env as a response to https://bugzilla.redhat.com/show_bug.cgi?id=1995953.

### DIFF
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"testing"
@@ -74,6 +75,18 @@ func checkDeploymentDoesNotHaveEnvVar(t *testing.T, deployment *appsv1.Deploymen
 	}
 	if foundValue {
 		t.Errorf("router Deployment has unexpected %s setting: %q", name, actualValue)
+	}
+}
+
+func checkDeploymentHasEnvSorted(t *testing.T, deployment *appsv1.Deployment) {
+	t.Helper()
+
+	env := deployment.Spec.Template.Spec.Containers[0].Env
+	isSorted := sort.SliceIsSorted(env, func(i, j int) bool {
+		return env[i].Name < env[j].Name
+	})
+	if !isSorted {
+		t.Errorf("router Deployment environment variables are unsorted")
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -327,6 +344,8 @@ func TestDesiredRouterDeployment(t *testing.T) {
 	checkDeploymentHasEnvVar(t, deployment, RouterEnableCompression, false, "")
 	checkDeploymentHasEnvVar(t, deployment, RouterCompressionMIMETypes, false, "")
 
+	checkDeploymentHasEnvSorted(t, deployment)
```
candita, line 347: Checking once at the end is sufficient.

gcs278: So that's what I did in the beginning; however, the bug actually was that env was not sorted specifically for EndpointPublishingStrategy = hostNetwork, though it was sorted for LoadBalancingService. So to be completely thorough, I placed it after every `desiredRouterDeployment` which generates a new ENV list. This test function is a bit overloaded as you might be able to tell.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -77,6 +78,22 @@ func checkDeploymentDoesNotHaveEnvVar(t *testing.T, deployment *appsv1.Deploymen
 	}
 }
 
+func checkDeploymentHasEnvSorted(t *testing.T, deployment *appsv1.Deployment) {
```
candita, line 81: Could you just use https://pkg.go.dev/sort#SliceIsSorted ?

pkg/operator/controller/ingress/deployment_test.go
candita: If you've got the time, it might be better to group all the env var checks together in a function that iterates over the list of env variable structs, applies them, then checks they are sorted at the end.  It would make this test easier to read and maintain, especially for new env variables.

gcs278: Yes I agree. I think this test function is a bit cluttered. Let me take a stab at cleaning some stuff up like you mentioned.

candita: I'm going to do some work on this for my current PR, so unless you've started, this lgtm.

gcs278:
@candita thanks for the push...my first real source code contribution to openshift!

PR #716, opened 2022-03-04
[WIP] Bug 2024946: Canary: Add ingress controller's routeSelectors to canary route labels to support sharding

gcs278:
Canary route object didn't update when the ingress controller's routeSelectors were updated as in the case of route sharding. This change links the canary route metadata.labels to the ingress controller's spec.routeSelectors.matchLabels so the canary wont fail when ingress controller's routeSelectors are added.


gcs278:
I just discovered routeSelector.matchExpressions...which makes this not trivial. So I think it's back to the drawing board on this one.

PR #720, opened 2022-03-14
Bug 2071139: delete default ingress pod if it is scheduled where another router pod already is

deads2k:
Anti-affinity in the scheduler has a bug and isn't properly spreading pods.  This fits a new library-go controller into the ingress-operator to delete a mal-scheduled pod.  If it finds two non-deleted pods using the selector on the same node, then the oldest pod is deleted.  Oldest is chosen so that a deployment controller could still make progress.  This *different* that is commonly done to preserve the oldest to disrupt the fewest connections, but it's better to avoid blocking the deployment controller.

I chose to start with ingress because it has the biggest multiplier and this doesn't fully resolve the bug for this component, it only impacts the default ingress.

/hold
This PR is proof for https://github.com/openshift/library-go/pull/1330


deads2k:
"Operator \"cluster-monitoring-operator\" produces more watch requests than expected: watchrequestcount=104, upperbound=96, ratio=1.0833333333333333",

/retest

deads2k:
Fix here https://github.com/openshift/library-go/pull/1344, will pick when it lands.

Miciah:
Thanks!
/lgtm

Miciah:
/retitle Bug 2071139: delete default ingress pod if it is scheduled where another router pod already is

PR #721, opened 2022-03-14
Bug 2066444: update relatedObjects for clusteroperator status

deads2k:
Adding to the CVO manifest will ensure collection even during installation failures.

relatedObjects is based on resources, not kinds.  This ensures that collection always happens without a RESTMapping which may or may not succeed.


manifests/03-cluster-operator.yaml
```diff
@@ -13,3 +13,19 @@ status:
...
+  relatedObjects:
+    - group: ""
+      name: openshift-ingress-operator
+      resource: namespaces
+    - group: ""
+      name: openshift-ingress
+      resource: namespaces
+    - group: ""
+      name: openshift-ingress-canary
+      resource: namespaces
```
Miciah, line 25: The operator creates these namespaces; could having them in the clusteroperator before the operator has created them cause problems?

wking: `must-gather` and `oc adm inspect ...` will both continue and eventually exit 0 in the face of unresolved references.  For example, in a recent 4.11 nightly run:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.11-e2e-aws-serial/1503335183393230848/artifacts/e2e-aws-serial/gather-must-gather/artifacts/must-gather.tar | tar xOz must-gather.log | grep 'have a resource type'
[must-gather-wxxwd] POD 2022-03-14T13:17:47.583259366Z     the server could not find the requested resource], skipping gathering EgressFirewall.k8s.ovn.org due to error: the server doesn't have a resource type "EgressFirewall", skipping gathering EgressIP.k8s.ovn.org due to error: the server doesn't have a resource type "EgressIP", skipping gathering endpoints/host-etcd-2 due to error: endpoints "host-etcd-2" not found, skipping gathering sharedconfigmaps.sharedresource.openshift.io due to error: the server doesn't have a resource type "sharedconfigmaps", skipping gathering sharedsecrets.sharedresource.openshift.io due to error: the server doesn't have a resource type "sharedsecrets"]
```

Including these references in the must-gather will ensure they are collected in the even that your operator wedges before writing `relatedObjects`.  So including `openshift-ingress-operator` here is important, because that would allow you to debug why your operator failed to come up.  I'm agnostic about the other entries...

dgoodwin:
Bump and

/retest

Miciah:
/retitle Bug 2066444: update relatedObjects for clusteroperator status

Miciah:
`TestClusterOperatorStatusRelatedObjects` needs to be updated.

PR #723, opened 2022-03-16
Bug 2021446: Set canary status as unknown if not admitted to default ingress controller

gcs278:
Adds a new status for IngressControllerCanaryCheckSuccessConditionType as unknown if the canary route is not admitted to the default ingress controller. It also doesn't attempt to probe canary route if probe routine is already started and canary route is not admitted.

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -165,7 +165,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return result, fmt.Errorf("failed to get canary service: %v", err)
 	}
 
-	haveRoute, route, err := r.ensureCanaryRoute(service)
+	haveRoute, _, err := r.ensureCanaryRoute(service)
 	if err != nil {
 		return result, fmt.Errorf("failed to ensure canary route: %v", err)
 	} else if !haveRoute {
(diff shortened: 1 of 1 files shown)


melvinjoseph86:
https://bugzilla.redhat.com/show_bug.cgi?id=2021446#c8
/label qe-approved

pkg/operator/controller/canary/controller.go
```diff
@@ -244,6 +241,14 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			return
 		}
 
+		// Don't attempt to probe if route is not actually admitted.
+		if !checkRouteAdmitted(route) {
+			if err := r.setCanaryNotAdmittedStatusCondition(); err != nil {
+				log.Error(err, "error updating canary status condition")
+			}
+			return
+		}
```
Miciah, line 250: This raises the question, Should we also set the status to "Unknown" in the case that the route doesn't exist?  We can leave that as a follow-up if you prefer.

gcs278: Even if you go unmanaged, doesn't the ingress controller still ensure the canary route? Is there a situation in which a user can delete the canary route and it not come back (at least try to come back)? I suppose you can never be too sure...

Miciah:
Hunky dory!
/lgtm

PR #724, opened 2022-03-17
Bug 1944851: Clear route status when an ingress controller is deleted or a route is un-admitted

gcs278:
Updates for fixing stale route status which is impacting various operations such as sharding and ingress operator status.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"regexp/syntax"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 
@@ -33,8 +34,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -98,6 +101,15 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, enqueueRequestForOwningIngressController(config.Namespace)); err != nil {
 		return nil, err
 	}
+	// Add watch for deleted pods specifically for ensuring ingress deletion.
+	if err := c.Watch(&source.Kind{Type: &corev1.Pod{}}, enqueueRequestForOwningIngressController(config.Namespace), predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}); err != nil {
+		return nil, err
+	}
 	// add watch for changes in DNS config
 	if err := c.Watch(&source.Kind{Type: &configv1.DNS{}}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
 		return nil, err
@@ -136,6 +148,16 @@ func enqueueRequestForOwningIngressController(namespace string) handler.EventHan
 		func(a client.Object) []reconcile.Request {
 			labels := a.GetLabels()
 			if ingressName, ok := labels[manifests.OwningIngressControllerLabel]; ok {
+				log.Info("queueing ingress", "name", ingressName, "related", a.GetSelfLink())
+				return []reconcile.Request{
+					{
+						NamespacedName: types.NamespacedName{
+							Namespace: namespace,
+							Name:      ingressName,
+						},
+					},
+				}
+			} else if ingressName, ok := labels[operatorcontroller.ControllerDeploymentLabel]; ok {
 				log.Info("queueing ingress", "name", ingressName, "related", a.GetSelfLink())
 				return []reconcile.Request{
 					{
@@ -200,7 +222,13 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	// If the ingresscontroller is deleted, handle that and return early.
 	if ingress.DeletionTimestamp != nil {
 		if err := r.ensureIngressDeleted(ingress); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to ensure ingress deletion: %v", err)
+			switch e := err.(type) {
+			case retryable.Error:
+				log.Error(e, "got retryable error; requeueing", "after", e.After())
+				return reconcile.Result{RequeueAfter: e.After()}, nil
+			default:
+				return reconcile.Result{}, fmt.Errorf("failed to ensure ingress deletion: %v", err)
+			}
 		}
 		log.Info("ingresscontroller was successfully deleted", "ingresscontroller", ingress)
 		return reconcile.Result{}, nil
@@ -773,6 +801,22 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 			errs = append(errs, fmt.Errorf("failed to get deployment for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 		} else if haveDepl {
 			errs = append(errs, fmt.Errorf("deployment still exists for ingress %s/%s", ingress.Namespace, ingress.Name))
+		} else {
+			// Wait for all the router pods to be deleted. This is important because the router deployment
+			// gets deleted a handful of milliseconds before the router pods process the graceful shutdown. This causes
+			// a race condition in which we clear route status, then the router pod will race to re-admit the status in
+			// these few milliseconds before it initiates the graceful shutdown. The only way to avoid is to wait
+			// until all router pods are deleted.
+			if allDeleted, err := r.allRouterPodsDeleted(ingress); err != nil {
+				errs = append(errs, err)
+			} else if allDeleted {
+				// Deployment has been deleted and there are no more pods left.
+				// Clear all routes status for this ingress controller.
+				statusErrs := r.clearAllRoutesStatusForIngressController(ingress.ObjectMeta.Name)
+				errs = append(errs, statusErrs...)
+			} else {
+				errs = append(errs, retryable.New(fmt.Errorf("not all router pods have been deleted for %s/%s", ingress.Namespace, ingress.Name), 15*time.Second))
+			}
 		}
 	}
 
@@ -790,7 +834,7 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 			}
 		}
 	}
-	return utilerrors.NewAggregate(errs)
+	return retryable.NewMaybeRetryableAggregate(errs)
 }
 
 // ensureIngressController ensures all necessary router resources exist for a
@@ -907,9 +951,22 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		errs = append(errs, fmt.Errorf("failed to list pods in namespace %q: %v", operatorcontroller.DefaultOperatorNamespace, err))
 	}
 
-	errs = append(errs, r.syncIngressControllerStatus(ci, deployment, deploymentRef, pods.Items, lbService, operandEvents.Items, wildcardRecord, dnsConfig, infraConfig))
+	syncStatusErr, updated := r.syncIngressControllerStatus(ci, deployment, deploymentRef, pods.Items, lbService, operandEvents.Items, wildcardRecord, dnsConfig, infraConfig)
+	errs = append(errs, syncStatusErr)
+
+	// If syncIngressControllerStatus updated our ingress status, it's important we query for that new object.
+	// If we don't, then the next function syncRouteStatus would always fail because it has a stale ingress object.
+	if updated {
+		updatedIc := &operatorv1.IngressController{}
+		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ci.Namespace, Name: ci.Name}, updatedIc); err != nil {
+			errs = append(errs, fmt.Errorf("failed to get ingresscontroller: %w", err))
+		}
+		ci = updatedIc
+	}
 
 	SetIngressControllerNLBMetric(ci)
+
+	errs = append(errs, r.syncRouteStatus(ci)...)
 
 	return retryable.NewMaybeRetryableAggregate(errs)
 }
@@ -952,4 +1009,22 @@ func IsProxyProtocolNeeded(ic *operatorv1.IngressController, platform *configv1.
 		}
 	}
 	return false, nil
+}
+
+// allRouterPodsDeleted determines if all the router pods for a given ingress controller are deleted.
+func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController) (bool, error) {
+	// List all pods that are owned by the ingress controller.
+	podList := &corev1.PodList{}
+	labels := map[string]string{
+		operatorcontroller.ControllerDeploymentLabel: ingress.Name,
+	}
+	if err := r.client.List(context.TODO(), podList, client.InNamespace(operatorcontroller.DefaultOperandNamespace), client.MatchingLabels(labels)); err != nil {
+		return false, fmt.Errorf("failed to list all pods owned by %s: %w", ingress.Name, err)
+	}
+	// If any pods exist, return false since they haven't all been deleted.
+	if len(podList.Items) > 0 {
+		return false, nil
+	}
+
+	return true, nil
 }
--- a/pkg/operator/controller/ingress/router_status.go
+++ b/pkg/operator/controller/ingress/router_status.go
@@ -1,0 +1,197 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
+	"reflect"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// A brief overview of the operator's interactions with route status:
+// Though the openshift-router is mainly responsible for route object status, the operator plays a small, but
+// significant role in ensuring the route status is accurate. The openshift-router updates the route object's status
+// when it is admitted to an ingress controller. However, the openshift-router is unable to reliably update the route's
+// status when it stops managing the route. Here are the scenarios where the operator steps in:
+// #1 When the ingress controller, the corresponding router deployment, and its pods are deleted.
+//    - The operator knows when a router is deleted because it is the one responsible for deleting it. So it
+//      simply calls clearRouteStatus to clear status of routes that openshift-router has admitted.
+// #2 When the ingress controller sharding configuration (i.e., selectors) is changed.
+//    - When the selectors (routeSelector and namespaceSelector) are updated, the operator simply clears the status of
+//      any route that it is no longer selecting using the updated selectors.
+//    - We determine what routes are admitted by the current state of the selectors (just like the openshift-router).
+
+// syncRouteStatus ensures that all routes status have been synced with the ingress controller's state.
+func (r *reconciler) syncRouteStatus(ic *operatorv1.IngressController) []error {
+	// Clear routes that are not admitted by this ingress controller if route selectors have been updated.
+	if routeSelectorsUpdated(ic) {
+		// Only clear once we are done rolling out routers.
+		// We want to avoid race condition in which we clear status and an old router re-admits it before terminated.
+		if done, err := r.isRouterDeploymentRolloutComplete(ic); err != nil {
+			return []error{err}
+		} else if done {
+			// Clear routes status not admitted by this ingress controller.
+			if errs := r.clearRoutesNotAdmittedByIngress(ic); len(errs) > 0 {
+				return errs
+			}
+
+			// Now sync the selectors from the spec to the status, so we indicate we are done clearing status.
+			if err := r.syncIngressControllerSelectorStatus(ic); err != nil {
+				return []error{err}
+			}
+		}
+	}
+	return nil
+}
+
+// isRouterDeploymentRolloutComplete determines whether the rollout of the ingress router deployment is complete.
+func (r *reconciler) isRouterDeploymentRolloutComplete(ic *operatorv1.IngressController) (bool, error) {
+	deployment := appsv1.Deployment{}
+	deploymentName := operatorcontroller.RouterDeploymentName(ic)
+	if err := r.cache.Get(context.TODO(), deploymentName, &deployment); err != nil {
+		return false, fmt.Errorf("failed to get deployment %s: %w", deploymentName, err)
+	}
+
+	if deployment.Status.Replicas != deployment.Status.UpdatedReplicas {
+		return false, nil
+	}
+	return true, nil
+}
+
+// clearAllRoutesStatusForIngressController clears any route status that have been
+// admitted by provided ingress controller.
+func (r *reconciler) clearAllRoutesStatusForIngressController(icName string) []error {
+	// List all routes.
+	errs := []error{}
+	start := time.Now()
+	routeList := &routev1.RouteList{}
+	routesCleared := 0
+	if err := r.client.List(context.TODO(), routeList); err != nil {
+		return append(errs, fmt.Errorf("failed to list all routes in order to clear route status for deployment %s: %w", icName, err))
+	}
+	// Clear status on the routes that belonged to icName.
+	for i := range routeList.Items {
+		if cleared, err := r.clearRouteStatus(&routeList.Items[i], icName); err != nil {
+			errs = append(errs, err)
+		} else if cleared {
+			routesCleared++
+		}
+	}
+	elapsed := time.Since(start)
+	log.Info("cleared all route status for ingress", "Ingress Controller",
+		icName, "Routes Status Cleared", routesCleared, "Time Elapsed", elapsed)
+
+	return errs
+}
+
+// clearRouteStatus clears a route's status that is admitted by a specific ingress controller.
+func (r *reconciler) clearRouteStatus(route *routev1.Route, icName string) (bool, error) {
+	// Go through each route and clear status if admitted by this ingress controller.
+	var updated routev1.Route
+	for i := range route.Status.Ingress {
+		if condition := findCondition(&route.Status.Ingress[i], routev1.RouteAdmitted); condition != nil {
+			if route.Status.Ingress[i].RouterName == icName {
+				// Remove this status since it matches our routerName.
+				route.DeepCopyInto(&updated)
+				updated.Status.Ingress = append(route.Status.Ingress[:i], route.Status.Ingress[i+1:]...)
+				if err := r.client.Status().Update(context.TODO(), &updated); err != nil {
+					return false, fmt.Errorf("failed to clear route status of %s/%s for routerName %s: %w",
+						route.Namespace, route.Name, icName, err)
+				}
+				log.Info("cleared admitted status for route", "Route", route.Namespace+"/"+route.Name,
+					"Ingress Controller", icName)
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
+}
+
+// routeSelectorsUpdated returns whether any of the route selectors have been updated by comparing
+// the status selector fields to the spec selector fields.
+func routeSelectorsUpdated(ingress *operatorv1.IngressController) bool {
+	if !reflect.DeepEqual(ingress.Spec.RouteSelector, ingress.Status.RouteSelector) ||
+		!reflect.DeepEqual(ingress.Spec.NamespaceSelector, ingress.Status.NamespaceSelector) {
+		return true
+	}
+	return false
+}
+
+// clearRoutesNotAdmittedByIngress clears routes status that are not selected by a specific ingress controller.
+func (r *reconciler) clearRoutesNotAdmittedByIngress(ingress *operatorv1.IngressController) []error {
+	start := time.Now()
+	errs := []error{}
+
+	// List all routes.
+	routeList := &routev1.RouteList{}
+	if err := r.client.List(context.TODO(), routeList); err != nil {
+		return append(errs, fmt.Errorf("failed to list all routes in order to clear route status: %w", err))
+	}
+
+	// List namespaces filtered by our ingress's namespace selector.
+	namespaceSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.NamespaceSelector)
+	if err != nil {
+		return append(errs, fmt.Errorf("ingresscontroller %s has an invalid namespace selector: %w", ingress.Name, err))
+	}
+	filteredNamespaceList := &corev1.NamespaceList{}
+	if err := r.client.List(context.TODO(), filteredNamespaceList,
+		client.MatchingLabelsSelector{Selector: namespaceSelector}); err != nil {
+		return append(errs, fmt.Errorf("failed to list all namespaces in order to clear route status for %s: %w", ingress.Name, err))
+	}
+
+	// Create a set of namespaces to easily look up namespaces in this shard.
+	namespacesInShard := sets.NewString()
+	for i := range filteredNamespaceList.Items {
+		namespacesInShard.Insert(filteredNamespaceList.Items[i].Name)
+	}
+
+	// List routes filtered by our ingress's route selector.
+	routeSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.RouteSelector)
+	if err != nil {
+		return append(errs, fmt.Errorf("ingresscontroller %s has an invalid route selector: %w", ingress.Name, err))
+	}
+
+	// Iterate over the entire route list and clear if not selected by route selector OR namespace selector.
+	routesCleared := 0
+	for i := range routeList.Items {
+		route := &routeList.Items[i]
+
+		routeInShard := routeSelector.Matches(labels.Set(route.Labels))
+		namespaceInShard := namespacesInShard.Has(route.Namespace)
+
+		if !routeInShard || !namespaceInShard {
+			if cleared, err := r.clearRouteStatus(route, ingress.ObjectMeta.Name); err != nil {
+				errs = append(errs, err)
+			} else if cleared {
+				routesCleared++
+			}
+		}
+
+	}
+	elapsed := time.Since(start)
+	log.Info("cleared route status after selector update", "Ingress Controller", ingress.Name, "Routes Status Cleared", routesCleared, "Time Elapsed", elapsed)
+	return errs
+}
+
+// findCondition locates the first condition that corresponds to the requested type.
+func findCondition(ingress *routev1.RouteIngress, t routev1.RouteIngressConditionType) *routev1.RouteIngressCondition {
+	for i := range ingress.Conditions {
+		if ingress.Conditions[i].Type == t {
+			return &ingress.Conditions[i]
+		}
+	}
+	return nil
+}
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -50,21 +50,22 @@ type expectedCondition struct {
 
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
-func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) error {
+func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, infraConfig *configv1.Infrastructure) (error, bool) {
+	updatedIc := false
 	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
 	if err != nil {
-		return fmt.Errorf("deployment has invalid spec.selector: %v", err)
+		return fmt.Errorf("deployment has invalid spec.selector: %v", err), updatedIc
 	}
 
 	platform, err := oputil.GetPlatformStatus(r.client, infraConfig)
 	if err != nil {
-		return fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %w", ic.Namespace, ic.Name, err)
+		return fmt.Errorf("failed to determine infrastructure platform status for ingresscontroller %s/%s: %w", ic.Namespace, ic.Name, err), updatedIc
 	}
 
 	secret := &corev1.Secret{}
 	secretName := controller.RouterEffectiveDefaultCertificateSecretName(ic, deployment.Namespace)
 	if err := r.client.Get(context.TODO(), secretName, secret); err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("failed to get the default certificate secret %s for ingresscontroller %s/%s: %w", secretName, ic.Namespace, ic.Name, err)
+		return fmt.Errorf("failed to get the default certificate secret %s for ingresscontroller %s/%s: %w", secretName, ic.Namespace, ic.Name, err), updatedIc
 	}
 
 	var errs []error
@@ -92,11 +93,27 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		if err := r.client.Status().Update(context.TODO(), updated); err != nil {
 			errs = append(errs, fmt.Errorf("failed to update ingresscontroller status: %v", err))
 		} else {
+			updatedIc = true
 			SetIngressControllerConditionsMetric(updated)
 		}
 	}
 
-	return retryableerror.NewMaybeRetryableAggregate(errs)
+	return retryableerror.NewMaybeRetryableAggregate(errs), updatedIc
+}
+
+// syncIngressControllerSelectorStatus syncs the routeSelector and namespaceSelector
+// from the spec to the status for tracking selector state.
+func (r *reconciler) syncIngressControllerSelectorStatus(ic *operatorv1.IngressController) error {
+	// Sync selectors from Spec to Status. This allows us to determine if either of these were updated.
+	updated := ic.DeepCopy()
+	updated.Status.RouteSelector = ic.Spec.RouteSelector
+	updated.Status.NamespaceSelector = ic.Spec.NamespaceSelector
+
+	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+		return fmt.Errorf("failed to update ingresscontroller status: %w", err)
+	}
+
+	return nil
 }
 
 // MergeConditions adds or updates matching conditions, and updates
--- a/test/e2e/router_status_test.go
+++ b/test/e2e/router_status_test.go
@@ -1,0 +1,301 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	routev1 "github.com/openshift/api/route/v1"
+
+	corev1 "k8s.io/api/core/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	admittedCondition = routev1.RouteIngressCondition{Type: routev1.RouteAdmitted, Status: corev1.ConditionTrue}
+)
+
+// TestDeleteIngressControllerShouldClearRouteStatus deletes an Ingress Controller and
+// makes sure a previously admitted route has its status cleared since.
+func TestDeleteIngressControllerShouldClearRouteStatus(t *testing.T) {
+	t.Parallel()
+	// Create an ingress controller that can admit our route.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "ic-delete-test"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	ic.Spec.RouteSelector = &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"type": icName.Name,
+		},
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Create a route to be admitted by this ingress controller.
+	// Use openshift-console namespace to get a namespace outside the ingress-operator's cache.
+	routeName := types.NamespacedName{Namespace: "openshift-console", Name: "route-" + icName.Name}
+	route := newRouteWithLabel(routeName, icName.Name)
+	if err := kclient.Create(context.TODO(), route); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), route); err != nil {
+			t.Fatalf("failed to delete route %s: %v", routeName, err)
+		}
+	}()
+
+	// Wait for route to be admitted.
+	if err := waitForRouteIngressConditions(t, kclient, routeName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Delete the ingress controller to trigger route status clean up.
+	assertIngressControllerDeleted(t, kclient, ic)
+
+	// Wait until route status goes unadmitted.
+	if err := waitForRouteStatusClear(t, kclient, routeName, ic.Name); err != nil {
+		t.Fatalf("failed to observe route status clear when ingress controller was deleted: %v", err)
+	}
+}
+
+// TestIngressControllerRouteSelectorUpdateShouldClearRouteStatus modifies an ingress controller's selectors to
+// unadmit a route and verifies it clears the admitted route's status.
+func TestIngressControllerRouteSelectorUpdateShouldClearRouteStatus(t *testing.T) {
+	t.Parallel()
+	// Create an ingress controller that can admit our route.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "ic-route-selector-test"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	ic.Spec.RouteSelector = &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"type": "foo",
+		},
+	}
+
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Create a route to be immediately admitted by this ingress controller and then when the IC label selectors are
+	// updated, the status should clear.
+	// Use openshift-console namespace to get a namespace outside the ingress-operator's cache.
+	routeFooLabelName := types.NamespacedName{Namespace: "openshift-console", Name: "route-foo-label"}
+	routeFooLabel := newRouteWithLabel(routeFooLabelName, "foo")
+	if err := kclient.Create(context.TODO(), routeFooLabel); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), routeFooLabel); err != nil {
+			t.Fatalf("failed to delete route %s: %v", routeFooLabelName, err)
+		}
+	}()
+
+	// Create a route that will NOT be immediately admitted by the ingress controller, but will be admitted AFTER
+	// the IC selectors are updated. The status SHOULD be successfully admitted.
+	// Use openshift-console namespace to get a namespace outside the ingress-operator's cache.
+	routeBarLabelName := types.NamespacedName{Namespace: "openshift-console", Name: "route-bar-label"}
+	routeBarLabel := newRouteWithLabel(routeBarLabelName, "bar")
+	if err := kclient.Create(context.TODO(), routeBarLabel); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), routeBarLabel); err != nil {
+			t.Fatalf("failed to delete route %s: %v", routeBarLabel, err)
+		}
+	}()
+
+	// Wait for routeFooLabel to be admitted upon creation.
+	if err := waitForRouteIngressConditions(t, kclient, routeFooLabelName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Update the ingress controller to not select routeFooLabel to be admitted, but to select
+	// routeBarLabel instead.
+	if err := kclient.Get(context.TODO(), icName, ic); err != nil {
+		t.Fatalf("failed to get ingress resource: %v", err)
+	}
+	ic.Spec.RouteSelector = &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"type": "bar",
+		},
+	}
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	// Wait until routeFooLabel status is cleared.
+	if err := waitForRouteStatusClear(t, kclient, routeFooLabelName, ic.Name); err != nil {
+		t.Fatalf("failed to observe route status clear when the ingress controller route selector was updated: %v", err)
+	}
+
+	// Wait until routeBarLabel status is admitted.
+	if err := waitForRouteIngressConditions(t, kclient, routeBarLabelName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe route status admitted when the ingress controller route selector was updated: %v", err)
+	}
+}
+
+func TestIngressControllerNamespaceSelectorUpdateShouldClearRouteStatus(t *testing.T) {
+	t.Parallel()
+	// Create an ingress controller that can admit our route.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "ic-namespace-selector-test"}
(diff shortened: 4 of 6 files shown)


pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +720,56 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+	if err := r.client.List(context.TODO(), routeList); err != nil {
+		return fmt.Errorf("failed to list all routes in order to clear route status")
+	}
+	elapsed := time.Since(start)
+	fmt.Println(elapsed)
+
+	// Go through each route and clear status if admitted by this ingress controller.
+	start2 := time.Now()
+	for _, route := range routeList.Items {
+		updated := route.DeepCopy()
```
Miciah, line 742: The deepcopy could be moved to imediately before it's used below to avoid unnecessary copies.  

It might be worth declaring `var updated routev1.Route` here and then using `route.DeepCopyInto(&updated)` to do the deepcopy to save some memory allocations.

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +720,56 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+					if err := r.client.Update(context.TODO(), updated); err != nil {
+						return fmt.Errorf("failed to clear route status for %s with routerName %s: %v", route.Name, icName, err)
+					}
+					continue
+				}
+			}
+		}
+	}
+	elapsed2 := time.Since(start2)
+	fmt.Println(elapsed2)
```
Miciah, line 757: This timing information could be quite useful.  Can you make these print statements into formal log messages?

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +720,56 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+
+// findMostRecentIngress returns the name of the ingress status with the most recent Admitted condition transition time,
+// or an empty string if no such ingress exists.
+func (r *reconciler) clearRouteStatus(icName string) error {
+	routeList := &routev1.RouteList{}
+
+	// List all routes.
+
+	start := time.Now()
+	if err := r.client.List(context.TODO(), routeList); err != nil {
```
Miciah, line 733: Use a cache if possible.

pkg/operator/controller/ingress/controller.go
Miciah: > It might be worth declaring `var updated routev1.Route` here and then using `route.DeepCopyInto(&updated)` to do the deepcopy to save some memory allocations.

Actually, the variable declaration should be outside the outermost `for` loop to minimize allocations, if you decide to take this approach.

gcs278: ah of course, good idea

pkg/operator/controller/ingress/controller.go
gcs278: Sure

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+			return err
+		}
+	}
+	elapsed := time.Since(start)
+	log.Info("cleared route status for ingress deletion", "Ingress Controller", icName, "Routes Status Cleared", routesCleared, "Time Elapsed", elapsed)
+	return nil
+}
+
+// Clears this route's status that is admitted by provided ingressController
+func (r *reconciler) clearRouteStatus(route routev1.Route, icName string) (bool, error) {
```
Miciah, line 753:
```suggestion
func (r *reconciler) clearRouteStatus(route *routev1.Route, icName string) (bool, error) {
```
To avoid a copy.

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+	var updated routev1.Route
+	clearedStatus := false
+	for i := range route.Status.Ingress {
+		if condition := findCondition(&route.Status.Ingress[i], routev1.RouteAdmitted); condition != nil {
+			if routerName := &route.Status.Ingress[i].RouterName; routerName != nil && *routerName == icName {
+				// Remove this status since it matches our routerName.
+				route.DeepCopyInto(&updated)
+				updated.Status.Ingress = append(route.Status.Ingress[:i], route.Status.Ingress[i+1:]...)
+				if err := r.client.Status().Update(context.TODO(), &updated); err != nil {
+					return clearedStatus, fmt.Errorf("failed to clear route status for %s with routerName %s: %v", route.Name, icName, err)
```
Miciah, line 764: Include the route's namespace to reduce ambiguity.  (It maybe `err` already includes the route's name and namespace?)  Also, it's good practice to use `%w` when wrapping error values.  
```suggestion
					return clearedStatus, fmt.Errorf("failed to clear route status of route %s/%s for router %s: %w", route.Namespace, route.Name, icName, err)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+			if routeLabelSelector == nil {
+				routeSelectorMatch = true
+			} else if routeLabelSelector.Matches(labels.Set(route.Labels)) {
+				routeSelectorMatch = true
+			}
+		} else {
+			routeSelectorMatch = true
+		}
+
+		// TODO: What if namespace Expression is invalid? This will panic
```
Miciah, line 802: If the selector is invalid, the router won't be able to parse it either, and so the router won't start up.  We can log an error and skip cleanup in this case.

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+		if routeSelector != nil {
+			routeLabelSelector, _ := labels.Parse(routeSelector.String())
+			if routeLabelSelector == nil {
+				routeSelectorMatch = true
+			} else if routeLabelSelector.Matches(labels.Set(route.Labels)) {
+				routeSelectorMatch = true
+			}
+		} else {
+			routeSelectorMatch = true
+		}
```
Miciah, line 800: Apply the selector when you list routes:
```suggestion
	routeSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.RouteSelector)
	if err != nil {
		return fmt.Errorf("ingresscontroller %s/%s has an invalid route selector: %w", ingress.Namespace, ingress.Name, err)
	}
	listOpts := []client.ListOption{
		client.MatchingLabelsSelector{Selector: routeSelector},
	}
	if err := r.cache.List(context.TODO(), routeList, listOpts); err != nil {
		fmt.Errorf("failed to list all routes in order to clear route status")
	}
	for _, route := range routeList.Items {
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+		if namespaceLabelSelector == nil {
+			namespaceSelectorMatch = true
+		} else if namespaceLabelSelector.Matches(labels.Set(ns.Labels)) {
+			namespaceSelectorMatch = true
+		}
+		// If not my route, then clear my state (if any...)
+		if !namespaceSelectorMatch || !routeSelectorMatch {
+			fmt.Printf("clearing route %s from ingress controller %s", route.Name, ingress.Name)
+			cleared, err := r.clearRouteStatus(route, ingress.ObjectMeta.Name)
```
Miciah, line 818: I'm tempted to suggest you just clear the route status unconditionally and force the router to re-admit all the routes, but that would probably cause too much churn.

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+		namespaceLabelSelector, _ := labels.Parse(namespaceSelector.String())
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: route.Namespace}}
+		if err := r.cache.Get(context.TODO(), types.NamespacedName{Name: route.Namespace}, ns); err != nil {
+			fmt.Errorf("failed to get namespace of route")
+		}
+		if namespaceLabelSelector == nil {
+			namespaceSelectorMatch = true
+		} else if namespaceLabelSelector.Matches(labels.Set(ns.Labels)) {
+			namespaceSelectorMatch = true
+		}
```
Miciah, line 814: Phooey, GitHub lost a comment that I meant to include in my review, so I gotta retype it...

What OpenShift router does is list/watch namespaces using the namespace selector and maintain a list of namespaces that it's supposed to handle, and then each time the router processes a route, it checks whether the route is in that list.  You could do something similar:

1. Create a string set (`namespaces := sets.NewString()`).
2. List namespaces using the namespace selector, similar to what I suggested for listing routes using the route selector.
3. Iterate over the namespaces, adding them to this string set (`namespaces.Insert(namespace.Name)`).  

Then use that set when you iterate over routes (`if !namespaces.Has(route.Namespace) { continue }`).

Miciah: > Then use that set when you iterate over routes (`if !namespaces.Has(route.Namespace) { continue }`).

Errr, actually you want to clear the status in this case:
```go
if !namespaces.Has(route.Namespace) {
	r.clearRouteStatus(route, ingress.ObjectMeta.Name)
}
```

pkg/operator/controller/ingress/controller.go
Miciah: Sorry, my suggestion isn't right; you would need to list routes using the _old_ route selector, if you had it.  :thinking:...

Miciah: I don't think it's possible to negate a selector in the general case because a selector is conjunction, and the negation of a conjunction is a disjunction, which cannot be represented as a selector.  So what you could do is list routes using the route selector to get a filtered list (as I described in my earlier comment), then list _all_ routes, and then iterate through the list of all routes and clear the status of any that aren't in the filtered list.  This would require a little bit of cleverness to do efficiently though; maybe something like the following.  Build a set from the filtered list:
```go
	type routeKey struct {
		name, namespace string
	}
	routesInShard := map[routeKey]struct{}{}
	for i := range filteredRouteList.Items {
		route := &filteredRouteList.Items[i]
		routesInShard[routeKey{route.Name, route.Namespace}] = struct{}{}
	}
```
Then use this set when iterating over the full list of routes:
```go
	for i := range routeList.Items {
		route := &routeList.Items[i]
		if _, ok := routesInShard[routeKey{route.Name, route.Namespace]; !ok {
			r.clearRouteStatus(...)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -3,17 +3,19 @@ package ingress
...
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	retryable "github.com/openshift/cluster-ingress-operator/pkg/util/retryableerror"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"regexp"
+	"regexp/syntax"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"strings"
+	"time"
```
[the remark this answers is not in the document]
gcs278: Got it, not sure how that happend, but fixed it.

pkg/operator/controller/ingress/controller.go
```diff
@@ -242,6 +243,10 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Clear routes that are not admitted by this ingress controller
+	// TODO: Only call this when selectors are updated!
+	r.clearRoutesNotAdmittedByIngress(ingress)
```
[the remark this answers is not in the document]
gcs278: Yup 100% agree, just hadn't gotten to it yet. I remember you telling me to make sure it happens after they roll out.

pkg/operator/controller/ingress/controller.go
```diff
@@ -713,6 +726,121 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+		return fmt.Errorf("failed to list all routes in order to clear route status")
+	}
+	for _, route := range routeList.Items {
+		cleared, err := r.clearRouteStatus(route, icName)
+		if cleared {
+			routesCleared++
+		}
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
gcs278: Great, got it.

pkg/operator/controller/ingress/controller.go
gcs278: Great, got it.

pkg/operator/controller/ingress/controller.go
gcs278: Great, got it

pkg/operator/controller/ingress/controller.go
gcs278: Agreed. I suppose it wouldn't be too bad when we update it to only clear the status when the selectors are updated (that should be very rare right?). It would majorly reduce the logic here, that would be a benefit. But would not be the most efficient solution.

pkg/operator/controller/ingress/controller.go
gcs278: done

pkg/operator/controller/ingress/controller.go
gcs278: Done

pkg/operator/controller/ingress/controller.go
gcs278: done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return nil
+}
+
+// isRouterDeploymentRolloutComplete determines whether the rollout of the ingress router deployment is complete.
+func (r *reconciler) isRouterDeploymentRolloutComplete(ic *operatorv1.IngressController) (bool, error) {
+	deployment := appsv1.Deployment{}
+	deploymentName := operatorcontroller.RouterDeploymentName(ic)
+	if err := r.cache.Get(context.TODO(), deploymentName, &deployment); err != nil {
+		return false, fmt.Errorf("failed to get deployment: %w", deploymentName)
```
Miciah, line 64:
```suggestion
		return false, fmt.Errorf("failed to get deployment %s: %w", deploymentName, err)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -3,12 +3,11 @@ package ingress
 import (
 	"context"
 	"fmt"
+	"github.com/pkg/errors"
 	"regexp"
 	"regexp/syntax"
 	"strings"
 
-	"github.com/pkg/errors"
-
```
Miciah, line 11: Please drop this change.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+	}
+	elapsed := time.Since(start)
+	log.Info("cleared route status after selector update", "Ingress Controller", ingress.Name, "Routes Status Cleared", routesCleared, "Time Elapsed", elapsed)
+	return nil
+}
+
+// findCondition locates the first condition that corresponds to the requested type.
+func findCondition(ingress *routev1.RouteIngress, t routev1.RouteIngressConditionType) (_ *routev1.RouteIngressCondition) {
```
Miciah, line 204: Kind of weird to use the syntax for named results and then use the blank identifier.  * grin *.  
```suggestion
func findCondition(ingress *routev1.RouteIngress, t routev1.RouteIngressConditionType) *routev1.RouteIngressCondition {
```

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+	for i, existing := range ingress.Conditions {
+		if existing.Type != t {
+			continue
+		}
+		return &ingress.Conditions[i]
+	}
```
Miciah, line 210: We can micro-optimize by avoiding the copying into `existing`, and simplify the logic slightly by inverting the `if` statement and dropping the `continue`:
```suggestion
	for i := range ingress.Conditions {
		if ingress.Conditions[i].Type == t {
			return &ingress.Conditions[i]
		}
	}
```

pkg/operator/controller/ingress/status.go
```diff
@@ -99,6 +99,20 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	return retryableerror.NewMaybeRetryableAggregate(errs)
 }
 
+// Syncs the routeSelector and namespaceSelector in the spec to the status for tracking selector state.
+func (r *reconciler) syncIngressControllerSelectorStatus(ic *operatorv1.IngressController) error {
```
Miciah, line 103:
```suggestion
// syncIngressControllerSelectorStatus syncs the routeSelector and
// namespaceSelector in the spec to the status for tracking selector state.
func (r *reconciler) syncIngressControllerSelectorStatus(ic *operatorv1.IngressController) error {
```

pkg/operator/controller/ingress/status.go
```diff
@@ -99,6 +99,20 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
...
+	// Sync selectors from Spec to Status. This allows us to determine if either of these were updated.
+	updated := ic.DeepCopy()
+	updated.Status.RouteSelector = ic.Spec.RouteSelector
+	updated.Status.NamespaceSelector = ic.Spec.NamespaceSelector
+
+	if err := r.client.Status().Update(context.TODO(), updated); err != nil {
+		return fmt.Errorf("failed to update ingresscontroller status: %v", err)
```
Miciah, line 110:
```suggestion
		return fmt.Errorf("failed to update ingresscontroller status: %w", err)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestDeleteIngressControllerShouldClearRouteStatus deletes an Ingress Controller and
+// makes sure a previously admitted route has its status cleared since.
+func TestDeleteIngressControllerShouldClearRouteStatus(t *testing.T) {
+	// Create an ingress controller that can admit our route.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "ic-delete-test"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newLoadBalancerController(icName, domain)
```
Miciah, line 26: Better to use `newPrivateController` if you don't need a load balancer.  Provisioning an LB requires more time and money.  
```suggestion
	ic := newPrivateController(icName, domain)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	ic.Spec.RouteSelector = &metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"type": icName.Name,
+		},
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
```
Miciah, line 34: Make sure you delete the ingresscontroller even if `t.Fatalf` is called on line 41, line 45, or line 52.
```suggestion
	if err := kclient.Create(context.TODO(), ic); err != nil {
		t.Fatalf("failed to create ingresscontroller: %v", err)
	}
	defer func() {
		if err := kclient.Get(context.TODO(), icName, ic); err != nil {
			if apierrors.IsNotFound(err) {
				return
			}
			t.Fatalf("failed to get ingresscontroller %q: %v", icName, err)
		}
	 	assertIngressControllerDeleted(t, kclient, ic)
	}()
```
Actually, it might be worth tweaking `assertIngressControllerDeleted` to handle this case for you, but that involves a few steps (which I haven't tested):
1. Rename the `"k8s.io/apimachinery/pkg/api/errors"` import to `apierrors "k8s.io/apimachinery/pkg/api/errors"`, and update all references to the package (s/errors/apierrors).  (This step is optional—you could import the `"errors"` package under a different name—but it's more consistent with other code to rename `"k8s.io/apimachinery/pkg/api/errors"` and then import `"errors"` as "errors".)
2. Import `"errors"`.
3. Update `deleteIngressController` to use `%w` so that its error return value can be unwrapped:

```go
func deleteIngressController(t *testing.T, cl client.Client, ic *operatorv1.IngressController, timeout time.Duration) error {
	t.Helper()
	name := types.NamespacedName{Namespace: ic.Namespace, Name: ic.Name}
	if err := cl.Delete(context.TODO(), ic); err != nil {
		return fmt.Errorf("failed to delete ingresscontroller: %w", err)
	}
	// [existing logic]
```

4. Update `assertIngressControllerDeleted` to unwrap the error value from `deleteIngressController` and return without calling `t.Fatalf` if the unwrapped error value is a not-found error.

```go
func assertIngressControllerDeleted(t *testing.T, cl client.Client, ing *operatorv1.IngressController) {
	t.Helper()
	if err := deleteIngressController(t, cl, ing, 2*time.Minute); err != nil {
		if apierrors.IsNotFound(errors.Unwrap(err)) {
			return
		}
		t.Fatalf("WARNING: cloud resources may have been leaked! failed to delete ingresscontroller %s: %v", ing.Name, err)
	// [existing logic]
```

5. Use your fancy new `assertIngressControllerDeleted` here:

```suggestion
	if err := kclient.Create(context.TODO(), ic); err != nil {
		t.Fatalf("failed to create ingresscontroller: %v", err)
	}
	defer assertIngressControllerDeleted(t, kclient, ic)
```

Up to you which approach you prefer.

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	routeName := types.NamespacedName{Namespace: "default", Name: "route-" + icName.Name}
+	route := newRouteWithLabel(routeName, icName.Name)
+
+	// Create a route to be admitted by this ingress controller.
+	if err := kclient.Create(context.TODO(), route); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), route); err != nil {
+			t.Fatalf("failed to delete router %s: %v", routeName.Name, err)
```
Miciah, line 45:
```suggestion
			t.Fatalf("failed to delete route %s: %v", routeName.Name, err)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// Wait until route status is cleared.
+	if err := waitForRouteStatusClear(t, kclient, routeName, ic.Name); err != nil {
+		t.Fatalf("failed to observe route status clear when ingress controller selector was updated: %v", err)
+	}
+}
+
+// waitForRouteStatusClear waits for route to be unadmitted (e.g. admitted status cleared) by given ingress controller.
+func waitForRouteStatusClear(t *testing.T, cl client.Client, routeName types.NamespacedName, routerName string) error {
+	t.Helper()
+	return wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
```
Miciah, line 123: How long does the operator usually take to update the route status, including the time it takes for the router pods to terminate?  Graceful termination of a deployment with 1 router pod replica can take almost a minute, and listing all routes could possibly take a while on a strained cluster.  We should set the timeout period high enough to avoid flakiness on strained CI clusters.  Probably at least 2 minutes, maybe more.  

Remember, using a high timeout period here only makes the test take longer in the unexpected case that it fails, and it is worth the tradeoff for a failing test to take a few extra minutes to fail if it makes the test less flaky.

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// TestIngressControllerSelectorUpdateShouldClearRouteStatus modifies an ingress controller's selectors to
+// unadmit a route and verifies it clears the admitted route's status.
+func TestIngressControllerSelectorUpdateShouldClearRouteStatus(t *testing.T) {
+	// Create an ingress controller that can admit our route.
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: "ic-selector-test"}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newLoadBalancerController(icName, domain)
```
Miciah, line 70:
```suggestion
	ic := newPrivateController(icName, domain)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	routeName := types.NamespacedName{Namespace: "default", Name: "route" + icName.Name}
+	route := newRouteWithLabel(routeName, icName.Name)
+
+	// Create a route to be admitted by this ingress controller.
+	if err := kclient.Create(context.TODO(), route); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), route); err != nil {
+			t.Fatalf("failed to delete router %s: %v", routeName.Name, err)
```
Miciah, line 91:
```suggestion
			t.Fatalf("failed to delete route %s: %v", routeName.Name, err)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+// waitForRouteStatusClear waits for route to be unadmitted (e.g. admitted status cleared) by given ingress controller.
+func waitForRouteStatusClear(t *testing.T, cl client.Client, routeName types.NamespacedName, routerName string) error {
+	t.Helper()
+	return wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		route := &routev1.Route{}
+		if err := cl.Get(context.TODO(), routeName, route); err != nil {
+			t.Logf("failed to get route %s: %v", routeName.Name, err)
```
Miciah, line 126: Might as well log the whole route name:
```suggestion
			t.Logf("failed to get route %s: %v", routeName, err)
```

pkg/operator/controller/ingress/router_status.go
gcs278: Done

pkg/operator/controller/ingress/controller.go
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// A brief overview of the ingress controller interactions with route status:
+// Though the openshift-router is mainly responsible for route object status, the ingress controller plays a small, but
+// significant role in ensuring the route status is accurate. The openshift-router updates the route object's status
+// when it is admitted to an ingress controller. However, the openshift-router is unable to reliably update the route's
+// status when it is un-admitted. Here are the scenarios where the ingress controller steps in:
+// #1 When the ingress controller/router is deleted or killed
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// A brief overview of the ingress controller interactions with route status:
+// Though the openshift-router is mainly responsible for route object status, the ingress controller plays a small, but
+// significant role in ensuring the route status is accurate. The openshift-router updates the route object's status
+// when it is admitted to an ingress controller. However, the openshift-router is unable to reliably update the route's
+// status when it is un-admitted. Here are the scenarios where the ingress controller steps in:
```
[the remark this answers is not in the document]
gcs278: Agreed. Done.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+
+// routeSelectorsUpdated returns whether any of the route selectors have been updated by comparing
+// the status selector fields to the spec selector fields.
+func routeSelectorsUpdated(ingress *operatorv1.IngressController) bool {
+	if !reflect.DeepEqual(ingress.Spec.RouteSelector, ingress.Status.RouteSelector) ||
+		!reflect.DeepEqual(ingress.Spec.NamespaceSelector, ingress.Status.NamespaceSelector) {
+		return true
+	} else {
+		return false
+	}
```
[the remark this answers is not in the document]
gcs278: Done

test/e2e/router_status_test.go
gcs278: Done

test/e2e/router_status_test.go
gcs278: Done

test/e2e/router_status_test.go
gcs278: Done

test/e2e/router_status_test.go
gcs278: Ah that's a great point. Delete happens in average of 1 second, but selector updates takes an average of 45 seconds :confused: Understood the tradeoff as you mention, making longer.

test/e2e/router_status_test.go
gcs278: Assuming you prefer the private controller because it is created faster and deleted faster and admits routes in the same way?

Done.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+				return err
+			}
+
+			// Now sync the selectors from the spec to the status, so we indicate we are done clearing status.
+			if err := r.syncIngressControllerSelectorStatus(ic); err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
gcs278: Sorry - for some reason I missed these comments. I think they got hidden.

Done.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+
+// clearAllRoutesStatusForIngressController clears any route status that have been
+// admitted by provided ingress controller.
+func (r *reconciler) clearAllRoutesStatusForIngressController(icName string) error {
+	// List all routes.
+	start := time.Now()
+	routeList := &routev1.RouteList{}
+	routesCleared := 0
+	if err := r.cache.List(context.TODO(), routeList); err != nil {
+		return fmt.Errorf("failed to list all routes in order to clear route status for %s", icName)
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// clearRouteStatus clears a route's status that is admitted by a specific ingress controller.
+func (r *reconciler) clearRouteStatus(route *routev1.Route, icName string) (bool, error) {
+	// Go through each route and clear status if admitted by this ingress controller.
+	var updated routev1.Route
+	clearedStatus := false
+	for i := range route.Status.Ingress {
+		if condition := findCondition(&route.Status.Ingress[i], routev1.RouteAdmitted); condition != nil {
+			if routerName := &route.Status.Ingress[i].RouterName; routerName != nil && *routerName == icName {
```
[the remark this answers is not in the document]
gcs278: Ah force of habit. Done.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+				// Remove this status since it matches our routerName.
+				route.DeepCopyInto(&updated)
+				updated.Status.Ingress = append(route.Status.Ingress[:i], route.Status.Ingress[i+1:]...)
+				if err := r.client.Status().Update(context.TODO(), &updated); err != nil {
+					return clearedStatus, fmt.Errorf("failed to clear route status of %s/%s for routerName %s: %v",
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+						route.Namespace, route.Name, icName, err)
+				}
+				clearedStatus = true
```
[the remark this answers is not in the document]
gcs278: Ah yea you are right.

Done.

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// clearRoutesNotAdmittedByIngress clears routes status that are not admitted by a specific ingress controller.
+func (r *reconciler) clearRoutesNotAdmittedByIngress(ingress *operatorv1.IngressController) error {
+	start := time.Now()
+
+	// List all routes.
+	routeList := &routev1.RouteList{}
+	if err := r.cache.List(context.TODO(), routeList); err != nil {
+		return fmt.Errorf("failed to list all routes in order to clear route status")
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+	// List namespaces filtered by our ingress's namespace selector.
+	namespaceSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.NamespaceSelector)
+	if err != nil {
+		return fmt.Errorf("ingresscontroller %s has an invalid namespace selector: %w", ingress.Name, err)
+	}
+	namespaceLabelSelector, _ := labels.Parse(namespaceSelector.String())
+	filteredNamespaceList := &corev1.NamespaceList{}
+	if err := r.cache.List(context.TODO(), filteredNamespaceList,
+		client.MatchingLabelsSelector{Selector: namespaceLabelSelector}); err != nil {
+		return fmt.Errorf("failed to list all routes in order to clear route status for %s", ingress.Name)
```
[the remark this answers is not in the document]
gcs278: Whoops good eye.

Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// List namespaces filtered by our ingress's namespace selector.
+	namespaceSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.NamespaceSelector)
+	if err != nil {
+		return fmt.Errorf("ingresscontroller %s has an invalid namespace selector: %w", ingress.Name, err)
+	}
+	namespaceLabelSelector, _ := labels.Parse(namespaceSelector.String())
+	filteredNamespaceList := &corev1.NamespaceList{}
+	if err := r.cache.List(context.TODO(), filteredNamespaceList,
+		client.MatchingLabelsSelector{Selector: namespaceLabelSelector}); err != nil {
```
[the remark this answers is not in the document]
gcs278: Ah no you are aren't. I was following some example code, but that was from parsing ENV strings, so yes, this was needless.

Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// List routes filtered by our ingress's route selector.
+	routeSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.RouteSelector)
+	if err != nil {
+		return fmt.Errorf("ingresscontroller %s has an invalid route selector: %w", ingress.Name, err)
+	}
+	routeLabelSelector, _ := labels.Parse(routeSelector.String())
+	filteredRouteList := &routev1.RouteList{}
+	if err := r.cache.List(context.TODO(), filteredRouteList,
+		client.MatchingLabelsSelector{Selector: routeLabelSelector}); err != nil {
```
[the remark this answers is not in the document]
gcs278: Yup. Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+		return fmt.Errorf("failed to list all routes in order to clear route status")
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Iterate over the entire route list and clear if not selected by route selector OR namespace selector.
+	routesCleared := 0
+	for i := range routeList.Items {
+		route := &routeList.Items[i]
+		_, routeInShard := routesInShard[routeKey{route.Name, route.Namespace}]
+		namespaceInShard := namespacesInShard.Has(route.Namespace)
+
+		if !routeInShard || !namespaceInShard {
+			cleared, _ := r.clearRouteStatus(route, ingress.ObjectMeta.Name)
```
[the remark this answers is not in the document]
gcs278: ah good catch, wasn't handling this error.

Done. Added an error array to return multiple errors.

pkg/operator/controller/ingress/router_status.go
gcs278: ...I have to admit something...I stole this code from: https://github.com/openshift/router/blob/88ecceb8953751fcde2fe2eab52a5c15e07a955e/pkg/router/controller/status.go#L250

But yes this is silly. Done.

pkg/operator/controller/ingress/router_status.go
gcs278: Agree. This is more readable in my opinion. Like I said, I just found this as existing code in openshift-router. Lesson learned to review copied code.

Done.

pkg/operator/controller/ingress/status.go
gcs278: ah thanks missed this one.

Done.

pkg/operator/controller/ingress/status.go
gcs278: Done.

test/e2e/router_status_test.go
gcs278: Agreed.

Done.

test/e2e/router_status_test.go
gcs278: Sure, so I did call on line 36:
```
// Don't use assertIngressControllerDeleted because it's okay if it's not deleted
defer deleteIngressController(t, kclient, ic, 2*time.Minute)
```

However, I do agree with the update that it will make assertIngressControllerDeleted. TIL unwrapping errors, that's pretty cool. I'll make that update.

Done.

gcs278:
/hold

found race condition with deleting an ingress controller, clearing status, and the router updating the status back right before it's told to shut down.

pkg/operator/controller/ingress/controller.go
```diff
@@ -688,6 +687,23 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+			// a race condition in which we clear route status, then the router pod will race to re-admit the status in
+			// these few milliseconds it before it processes the graceful shutdown. The only way to avoid is to wait
+			// until all router pods are deleted.
+			if allDeleted, err := r.allRouterPodsDeleted(ingress); err != nil {
+				errs = append(errs, err)
+			} else if allDeleted {
+				// Deployment has been deleted and there are no more pods left.
+				// Clear all routes status for this ingress controller.
+				if statusErrs := r.clearAllRoutesStatusForIngressController(ingress.ObjectMeta.Name); len(statusErrs) > 0 {
+					errs = append(errs, statusErrs...)
```
gcs278, line 702: A concern I have, like the above, which applies to all thing of this pattern is: if you return an error on `ensureIngressDeletion`, it doesn't retry reconcilation.

So in my testing, I found a bug in which it failed the 1st time this was run. It'd then take another 6 minutes to Reconcile again. Why would we not try again if we failed to delete the ingress controller? The user's delete command is being blocked and not sure if this is the right approach (that is, to not even try again). We are stuck in terminating.

pkg/operator/controller/ingress/controller.go
```diff
@@ -688,6 +687,23 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
 		} else if haveDepl {
 			errs = append(errs, fmt.Errorf("deployment still exists for ingress %s/%s", ingress.Namespace, ingress.Name))
+		} else {
+			// Wait for all the router pods to be deleted. This is important because the router deployment
+			// gets deleted a handful of milliseconds before the router pods process the graceful shutdown. This causes
+			// a race condition in which we clear route status, then the router pod will race to re-admit the status in
+			// these few milliseconds it before it processes the graceful shutdown. The only way to avoid is to wait
+			// until all router pods are deleted.
+			if allDeleted, err := r.allRouterPodsDeleted(ingress); err != nil {
+				errs = append(errs, err)
```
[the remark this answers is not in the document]
gcs278: Update: I added reconcilation watch on "pods" so we reconcile on pod udpates now. This fixes this issue.

pkg/operator/controller/ingress/controller.go
```diff
@@ -142,6 +145,16 @@ func enqueueRequestForOwningIngressController(namespace string) handler.EventHan
 						},
 					},
 				}
+			} else if ingressName, ok := labels[operatorcontroller.ControllerDeploymentLabel]; ok {
```
gcs278, line 148: this is the logic that adds reconcilation for pods (different label name), but same value.

pkg/operator/controller/ingress/controller.go
```diff
@@ -95,6 +94,10 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, enqueueRequestForOwningIngressController(config.Namespace)); err != nil {
 		return nil, err
 	}
+	// Add watch for pods specifically for ensuring we know when pods get deleted for ensuring ingress deletion.
+	if err := c.Watch(&source.Kind{Type: &corev1.Pod{}}, enqueueRequestForOwningIngressController(config.Namespace)); err != nil {
```
gcs278, line 98: Is it okay that we reconcile on pod updates? It will generally increase the number of reconcilations, but is needed to reconcile again when the pods disappear.

The other option, eluded to below, is to reconcile on `ensureIngressDeleted` failure (or errors exist). If we did this, then arguably there is no need for watching the pods.

pkg/operator/controller/ingress/controller.go
gcs278: Side note: The reason it fails on the first time is because `r.client` lists duplicate canary routes for some reason. [Thread here](https://coreos.slack.com/archives/CBWMXQJKD/p1649704358332479)

pkg/operator/controller/ingress/controller.go
```diff
@@ -95,6 +96,15 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(&source.Kind{Type: &corev1.Service{}}, enqueueRequestForOwningIngressController(config.Namespace)); err != nil {
 		return nil, err
 	}
+	// Add watch for deleted pods specifically for ensuring ingress deletion.
+	if err := c.Watch(&source.Kind{Type: &corev1.Pod{}}, enqueueRequestForOwningIngressController(config.Namespace), predicate.Funcs{
```
gcs278, line 100: watching pods only on delete events so it triggers a reconcilation when they are cleaned up.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: Update2: Miciah said it really should reconcile on an error and maybe I was seeing the exponential backoff logic delays and getting confused.

pkg/operator/controller/ingress/controller.go
gcs278: Update: Miciah said it really should reconcile on an error and maybe I was seeing the exponential backoff logic delays and getting confused.

Update2: We found an issue with `r.cache` in which it has duplicate objects. That is being fixed separately.

pkg/operator/controller/ingress/controller.go
```diff
@@ -822,10 +859,23 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
...
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ci.Namespace, Name: ci.Name}, updatedIc); err != nil {
+				errs = append(errs, fmt.Errorf("failed to get ingresscontroller: %w", err))
+			}
+			ci = updatedIc
+		}
+	}
 
 	SetIngressControllerNLBMetric(ci)
 
+	errs = append(errs, r.syncRouteStatus(ci)...)
```
gcs278, line 877: Do you think I should do the same pattern above? Since I potentially update the ingress controller object here, subsequent updates after this function (in the coding of the future) will fail. Or is that something we can just leave to future developers?

gcs278:
Fixed race condition by updating the ingress controller deletion process to now to wait for all pods to be deleted before removing the finalizer on the ingress controller. This is a pretty big change, but the only impact should really be a longer wait time (45 seconds) for deleting an ingress controller and potentially longer E2E tests.

gcs278:
I think it's reasonable that we should make https://github.com/openshift/cluster-ingress-operator/pull/740 a dependency of this PR.  Though this PR would still work, it'd throw an error when an ingress controller is deleted. It would then try again and succeed.

gcs278:
Update: the e2e test is most likely indeed failing because of the change that doesn't delete ingress until all pods are gone. It's timing out of deleteIngressController after 1 hour.

I found that if you create and delete an ingress controller too quickly, it will never get delete (or take a really long time) because the router gets stuck in terminating.

gcs278:
Update: This PR exposed another bug: https://github.com/openshift/router/pull/383. I feel pretty confident that this will fix the E2E test problem. Once checked in, we can resume this PR.

gcs278:
Update: https://github.com/openshift/router/pull/383 has been merged. We can move forward with this PR.

gcs278:
The E2E tests are still failing. I believe I need to increase the timeout on the `assertIngressControllerDeleted`; however there is a really weird bug in which some E2E tests are just timing out after 1 hour. Some on timing out on `assertIngressControllerDeleted` [like this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/724/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1516473475324186624), but I just saw one time out on `waitForIngressControllerCondition` [like this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/724/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1516524743493160960#1:build-log.txt%3A1207). I do believe it's this PR that is causing this because I haven't seen these errors anywhere else.

gcs278:
I've discovered that this PR is pushing our e2e-aws-operator test over 1h TOTAL and the whole E2E times out. That because the ingress deletion adds 45 seconds and is pushing us over the 1h edge.

gcs278:
Increasing the test time helped, but my E2E test for router status failed. Back to the drawing board.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
candita: Yeah, I would remove the watch and start over.

pkg/operator/controller/ingress/controller.go
candita: This should not be needed, as Miciah mentioned.

gcs278:
/retest
Finally figured out the issue. Routes were not getting listed because the cache doesn't contain all routes in all namespaces.

gcs278:
/hold
@frobware has a PR that runs E2E in parallel and that will affect the timeout value that this PR is attempting to update.

gcs278:
TestScopeChange failed. I've never seen this before. Testing locally.
/retest

gcs278:
Just found that if there is a ImgPullError, then this change causes it to never be deleted. That's an issue.

pkg/operator/controller/ingress/controller.go
gcs278: I looked into this again - it does indeed reconcile again on the error "pods aren't deleted", however, it begins to trigger exponential back off, which causes the deletion to be a little slower than it needs to be (5-10 seconds slower).

This watch makes the final reconcile (when all pods are deleted) to be more timely which results in a 5-10 second quicker deletion. I'd say it's worth to keep it in here since that 5-10 seconds can add up.

Makefile
```diff
@@ -53,7 +53,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e:
-	$(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	$(GO) test -timeout 2h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
```
gcs278, line 56: This is the only TBD change. When https://github.com/openshift/cluster-ingress-operator/pull/756 gets introduced, than this can go away (so regardless, there will be 1 more change)

gcs278:
/test all
Going to retest everything to make sure I'm not getting lucky

gcs278:
/test e2e-aws-single-node
Unrelated CI failure

Miciah:
Faiiled on `TestIngressOperatorCacheIsNotGlobal`.
/test e2e-aws-operator

gcs278:
=== CONT  TestAll/parallel/TestIngressOperatorCacheIsNotGlobal
    operator_test.go:3117: failed to get ingresscontroller ignored-ns/test-cache-is-not-global: rpc error: code = Unknown desc = malformed header: missing HTTP content-type
    panic.go:642: deleted ingresscontroller test-cache-is-not-global

Appears like a flake to me.

/test e2e-aws-operator

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+			"type": icName.Name,
+		},
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Create a route to be admitted by this ingress controller.
+	// Use openshift-console namespace to get a namespace outside the ingress-operator's cache.
```
Miciah, line 41: I had to think a while about why the namespace would matter.  It's because you want to verify that the cleanup logic isn't restricted to routes in the multi-namespace cache, right?

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Wait for route to be admitted.
+	admittedCondition := routev1.RouteIngressCondition{Type: routev1.RouteAdmitted, Status: corev1.ConditionTrue}
+	if err := waitForRouteIngressConditions(t, kclient, routeName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Delete the ingress controller to trigger route status clean up.
+	// Warning: Deleting ingress so quickly will cause race conditions in updating ingress status and
+	//          cause some errors to be thrown; however, it doesn't matter the purpose of this E2E test.
```
Miciah, line 61: What kind of race conditions and errors?  Do these race conditions happen even with the `waitForRouteIngressConditions` above?

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+	if err := kclient.Create(context.TODO(), routeBarLabel); err != nil {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), routeBarLabel); err != nil {
+			t.Fatalf("failed to delete route %s: %v", routeBarLabel, err)
+		}
+	}()
+
+	// Wait for routeFooLabel to be admitted upon creation.
```
Miciah, line 117:
```suggestion
	// Wait for routeBarLabel to be admitted upon creation.
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+	admittedCondition := routev1.RouteIngressCondition{Type: routev1.RouteAdmitted, Status: corev1.ConditionTrue}
```
Miciah, line 118: You could declare a `routeAdmittedStatusCondition` variable at the top of the file to avoid the need to declare the same variable in all three tests.

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// Wait until routeFooLabel status is cleared.
+	if err := waitForRouteStatusClear(t, kclient, routeFooLabelName, ic.Name); err != nil {
+		t.Fatalf("failed to observe route status clear when the ingress controller route selector was updated: %v", err)
+	}
+
+	// Wait until routeBarLabel status is admitted.
+	if err := waitForRouteIngressConditions(t, kclient, routeBarLabelName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe route status admitted when te ingress controller route selector was updated: %v", err)
```
Miciah, line 144:
```suggestion
		t.Fatalf("failed to observe route status admitted when the ingress controller route selector was updated: %v", err)
```

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Create a new namespace for the route that we can immediately match with the IC's namespace selector.
+	nsFoo := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "foo",
```
Miciah, line 168: This name is rather generic; maybe better to pick a name that is unlikely to be used by something outside of the ingress operator's tests.

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+		if err := kclient.Delete(context.TODO(), routeBarLabel); err != nil {
+			t.Fatalf("failed to delete route %s: %v", routeBarLabel, err)
+		}
+	}()
+
+	// Wait for routeFooLabel to be admitted upon creation.
+	admittedCondition := routev1.RouteIngressCondition{Type: routev1.RouteAdmitted, Status: corev1.ConditionTrue}
+	if err := waitForRouteIngressConditions(t, kclient, routeFooLabelName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
```
Miciah, line 231: Would it make sense also to check ` 	if err := waitForRouteStatusClear(t, kclient, routeBarLabelName, ic.Name)` here?

test/e2e/router_status_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Wait until routeFooLabel status is cleared.
+	if err := waitForRouteStatusClear(t, kclient, routeFooLabelName, ic.Name); err != nil {
+		t.Fatalf("failed to observe route status clear when the ingress controller route selector was updated: %v", err)
+	}
+
+	// Wait until routeBarLabel status is admitted.
+	if err := waitForRouteIngressConditions(t, kclient, routeBarLabelName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe route status admitted when te ingress controller route selector was updated: %v", err)
```
Miciah, line 254:
```suggestion
		t.Fatalf("failed to observe route status admitted when the ingress controller route selector was updated: %v", err)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -773,6 +801,23 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
 			errs = append(errs, fmt.Errorf("failed to get deployment for ingress %s/%s: %v", ingress.Namespace, ingress.Name, err))
 		} else if haveDepl {
 			errs = append(errs, fmt.Errorf("deployment still exists for ingress %s/%s", ingress.Namespace, ingress.Name))
+		} else {
+			// Wait for all the router pods to be deleted. This is important because the router deployment
+			// gets deleted a handful of milliseconds before the router pods process the graceful shutdown. This causes
+			// a race condition in which we clear route status, then the router pod will race to re-admit the status in
+			// these few milliseconds it before it processes the graceful shutdown. The only way to avoid is to wait
```
Miciah, line 808:
```suggestion
			// these few milliseconds before it initiates the graceful shutdown. The only way to avoid is to wait
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -773,6 +801,23 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+			// until all router pods are deleted.
+			if allDeleted, err := r.allRouterPodsDeleted(ingress); err != nil {
+				errs = append(errs, err)
+			} else if allDeleted {
+				// Deployment has been deleted and there are no more pods left.
+				// Clear all routes status for this ingress controller.
+				if statusErrs := r.clearAllRoutesStatusForIngressController(ingress.ObjectMeta.Name); len(statusErrs) > 0 {
+					errs = append(errs, statusErrs...)
+				}
```
Miciah, line 817: Simpler:
```suggestion
				statusErrs := r.clearAllRoutesStatusForIngressController(ingress.Name)
				errs = append(errs, statusErrs...)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -773,6 +801,23 @@ func (r *reconciler) ensureIngressDeleted(ingress *operatorv1.IngressController)
...
+			} else if !allDeleted {
```
Miciah, line 818: Simpler:
```suggestion
			} else {
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -907,10 +952,23 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
...
+		// If syncIngressControllerStatus updated our ingress status, it's important we query for that new object.
+		// If we don't, then the next function syncRouteStatus would always fail because it has a stale ingress object.
+		if updated {
+			updatedIc := &operatorv1.IngressController{}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ci.Namespace, Name: ci.Name}, updatedIc); err != nil {
+				errs = append(errs, fmt.Errorf("failed to get ingresscontroller: %w", err))
+			}
+			ci = updatedIc
+		}
+	}
```
Miciah, line 966: Why do you only update `ci` if `syncStatusErr != nil`?  Couldn't `syncIngressControllerStatus` update the status, return `nil, true`, and cause `syncRouteStatus` to fail its update?

Miciah: BTW, `NewMaybeRetryableAggregate` ignores nil values.  I think you want the following:
```suggestion
	syncStatusErr, updated := r.syncIngressControllerStatus(ci, deployment, deploymentRef, pods.Items, lbService, operandEvents.Items, wildcardRecord, dnsConfig, infraConfig)
	errs = append(errs, syncStatusErr)
	// If syncIngressControllerStatus updated our ingress status, it's important we query for that new object.
	// If we don't, then the next function syncRouteStatus would always fail because it has a stale ingress object.
	if updated {
		updatedIc := &operatorv1.IngressController{}
		if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ci.Namespace, Name: ci.Name}, updatedIc); err != nil {
			errs = append(errs, fmt.Errorf("failed to get ingresscontroller: %w", err))
		}
		ci = updatedIc
	}
```

pkg/operator/controller/ingress/router_status.go
```diff
@@ -0,0 +1,198 @@
...
+	errs := []error{}
+	start := time.Now()
+	routeList := &routev1.RouteList{}
+	routesCleared := 0
+	if err := r.client.List(context.TODO(), routeList); err != nil {
+		return append(errs, fmt.Errorf("failed to list all routes in order to clear route status for deployment %s: %w", icName, err))
+	}
+	// Clear status on the routes that belonged to icName.
+	for _, route := range routeList.Items {
+		if cleared, err := r.clearRouteStatus(&route, icName); err != nil {
```
Miciah, line 86: We did this micro-optimization in `clearRouteStatus`, might as well do it here:
```suggestion
	for i := range routeList.Items {
		if cleared, err := r.clearRouteStatus(&route[i], icName); err != nil {
```

test/e2e/router_status_test.go
gcs278: Yup exactly. I was once erroneously searching the ingress-operator's cache for routes to clear status on, but routes can be in any namespace (unlike the ingress controllers...). This just verifies the clearing works for routes in all namespaces.

pkg/operator/controller/ingress/controller.go
gcs278: Yep, I removed this logic and double checked I was doing something valid. It works without with this logic removed, but I get this error on the first reconcilation :
>  failed to update ingresscontroller status: Operation cannot be fulfilled on ingresscontrollers.operator.openshift.io \"ic-delete-test\": the object has been modified; please apply your changes to the latest version and try again

So just a sanity check for why I did this.

gcs278 (commented):
Thanks for the review!

gcs278:
Errors don't look related.
/retest

Miciah:
Lovely jubbly!
/lgtm

PR #725, opened 2022-03-18
Fix incorrect related object references in ClusterOperator status.

dgoodwin:
origin will be adding tests shortly to fail on this, so we want to get
this fixed before those go in.


dgoodwin:
Oops, likely already done in https://github.com/openshift/cluster-ingress-operator/pull/721

PR #727, opened 2022-03-23
NE-825: Update router to use random balancing algorithm once again

gcs278:
After attempting to use random as the default algorithm and subsequently
reverting, we discovered that reducing haproxy's weight to 1 for backends
with a single service significantly reduces the memory overhead of
the random balance algorithm and provides the same exactly functionality.

We are now turning back on random again since we have mitigated the memory
overhead issue.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -492,18 +492,20 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 
 	// For non-TLS, edge-terminated, and reencrypt routes, use the
-	// "leastconn" balancing algorithm by default, but allow an unsupported
+	// "random" balancing algorithm by default, but allow an unsupported
 	// config override to override it.  For passthrough routes, use the
 	// "source" balancing algorithm in order to provide some
-	// session-affinity.  This code at one time used "random" as the default
-	// for non-passthrough routes, but we changed it to "leastconn" upon
-	// discovering that "random" incurs significant memory overhead for each
-	// backend that uses it.  See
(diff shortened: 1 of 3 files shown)


gcs278:
This PR should NOT be merged until https://github.com/openshift/router/pull/377 is merged.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -473,18 +473,20 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
-	// discovering that "random" incurs significant memory overhead for each
-	// backend that uses it.  See
-	// <https://bugzilla.redhat.com/show_bug.cgi?id=2007581>.
-	loadBalancingAlgorithm := "leastconn"
+	// session-affinity.
+	// We've had issues with "random" in the past due to significant memory
+	// overhead; however we mitigated that in openshift-router by effectively
+	// setting all endpoints in "random" backends to weight 1 to avoid incurring
+	// extraneous memory allocations. Be aware that using "random" with large
+	// weights can significantly increase memory overhead.
```
Miciah, line 484: This comment could be clearer.  As written, it could be interpreted to be saying that setting a large weight on a service can cause high memory usage.  However, with your change to OpenShift router, the high memory usage actually occurs under the specific scenario where a route has multiple associated services, irrespective of the weights that are specified on the services, right?  That is, the high memory usage is caused by large weights on the _backend server_, which is an implementation detail; from the user's perspective, the high memory usage is caused by having multiple services.

gcs278:
/unhold

We are ready for lgtms. Perf & Scale results have been reviewed.

PR #735, opened 2022-04-07
NE-577: Support a Configurable ROUTER_MAX_CONNECTIONS in HAproxy

frobware:
This implements https://github.com/openshift/enhancements/pull/1084.

Requires https://github.com/openshift/api/pull/1161 [MERGED]


frobware:
/hold

Needs the EP to be approved. 
Needs the API to be approved.

frobware:
Errors:

- [sig-scheduling][Early] The openshift-etcd pods should be scheduled on different nodes [Suite:openshift/conformance/parallel]

/retest

frobware:
configurable_route_test.go:101: error updating ingress.status: timed out waiting for the condition

/test e2e-aws-operator

test/e2e/maxconn_test.go
```diff
@@ -0,0 +1,81 @@
...
+		t.Fatalf("expected router deployment to have %s unset: %v", ingress.RouterMaxConnectionsEnvName, err)
+	}
+
+	for _, testCase := range []struct {
+		description    string
+		maxConnections int32
+		expectedEnvVar string
+	}{
+		{"set maxconn 12345", 12345, "12345"},
+		{"set maxconn auto", -1, "auto"},
```
gcs278, line 50: Just curious, have you prepared the PR to support "-1" as auto in openshift-router? I know you are playing with some changes with maxConn reservations.

gcs278 (commented):
Looks good to me.

test/e2e/maxconn_test.go
frobware: When we specify -1 that becomes "auto". The template only populates a value if it is an integer (https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L48). Setting a non-integer (i..e, "auto") omits any value when the template generation is run.

gcs278: Ah that's right. I forgot if you omit a maxconn value, it does auto.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -454,8 +454,9 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 	var expectedReplicas int32 = 8
 	ic.Spec.Replicas = &expectedReplicas
 	ic.Spec.UnsupportedConfigOverrides = runtime.RawExtension{
-		Raw: []byte(`{"loadBalancingAlgorithm":"random","dynamicConfigManager":"false","maxConnections":-1,"reloadInterval":15}`),
+		Raw: []byte(`{"loadBalancingAlgorithm":"random","dynamicConfigManager":"false","reloadInterval":15}`),
 	}
+	ic.Spec.TuningOptions.MaxConnections = -1
```
Miciah, line 459: This could be moved above, here: https://github.com/openshift/cluster-ingress-operator/blob/6620e18087417fc5f0c922602826ce4e30ae1ed2/pkg/operator/controller/ingress/deployment_test.go#L430-L434

test/e2e/maxconn_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func TestTunableMaxConnectionsValidValues(t *testing.T) {
+	updateMaxConnections := func(t *testing.T, client client.Client, timeout time.Duration, maxConnections int32, c *operatorv1.IngressController) error {
```
Miciah, line 23: Seems you only use `c` as a way to get the name of the object and then as a place to store the object that you get from the API, which is subtle.  What do you think about replacing the `*operatorv1.IngressController` parameter with a `types.NamespacedName` parameter and using a local variable to store the object from the API?

test/e2e/maxconn_test.go
```diff
@@ -0,0 +1,164 @@
...
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+	conditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
+		{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+		{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionFalse},
+	}
+	t.Logf("waiting for ingresscontroller %v", icName)
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, conditions...); err != nil {
```
Miciah, line 53:
```suggestion
	t.Logf("waiting for ingresscontroller %v", icName)
	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -538,8 +539,9 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 	// Any value for loadBalancingAlgorithm other than "random" should be
 	// ignored.
 	ic.Spec.UnsupportedConfigOverrides = runtime.RawExtension{
-		Raw: []byte(`{"loadBalancingAlgorithm":"source","dynamicConfigManager":"true","maxConnections":40000}`),
+		Raw: []byte(`{"loadBalancingAlgorithm":"source","dynamicConfigManager":"true"}`),
 	}
+	ic.Spec.TuningOptions.MaxConnections = 40000
```
Miciah, line 544: `TestDesiredRouterDeploymentSpecAndNetwork` feels like it should be at least two separate tests, but that can be a further cleanup for another day.

test/e2e/maxconn_test.go
Miciah: Would it make more sense just to use `kclient` and `time.Minute` instead of parameterizing the client and timeout?

Miciah (commented):
I have a few suggestions to simplify the code, but overall I would be fine with merging this as-is.

test/e2e/maxconn_test.go
frobware: I want the caller to specify how long to wait for. And I'd rather not have a mismatch of globals and discrete parameter values.

pkg/operator/controller/ingress/deployment_test.go
frobware: Did you mean above? I added `MaxConnections` as an addition after `ThreadCount`.

test/e2e/maxconn_test.go
frobware: > types.NamespacedName

Done.

frobware:
@Miciah I addressed your feedback with my latest push.

/hold cancel

frobware:
FAIL: TestHTTPCookieCapture

/test e2e-aws-operator

PR #740, opened 2022-04-12
Bug 2075671: Fix k8s client cache object global inclusion and duplication.

gcs278:
An empty string was included in MultiNamespacedCacheBuilder which was intended to capture
  Cluster Scoped Resources, but it erroneously included ALL objects in ANY namespace.
  It also duplicated other objects that had already been selected by a namespace. It is not
  needed since cluster scoped resources are already included in the cache by default.
  This may fix many inefficiencies such as double reconcilations and cache overloading.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -77,9 +77,7 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 			operatorcontroller.DefaultOperandNamespace,
 			operatorcontroller.DefaultCanaryNamespace,
 			operatorcontroller.GlobalMachineSpecifiedConfigNamespace,
-			"",
 			operatorcontroller.SourceConfigMapNamespace,
-			operatorcontroller.GlobalUserSpecifiedConfigNamespace,
 		}),
 		// Use a non-caching client everywhere. The default split client does not
 		// promise to invalidate the cache during writes (nor does it promise


gcs278:
/assign @Miciah 
Since you worked with me on finding this bug.

gcs278:
/cc @Miciah 
Trying to figure out what these do...

gcs278:
/unassign @Miciah 
I cc'ed you for review, I think that's more appropriate than assigning

pkg/operator/operator.go
```diff
@@ -77,7 +77,6 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 			operatorcontroller.DefaultOperandNamespace,
 			operatorcontroller.DefaultCanaryNamespace,
 			operatorcontroller.GlobalMachineSpecifiedConfigNamespace,
-			"",
 			operatorcontroller.SourceConfigMapNamespace,
 			operatorcontroller.GlobalUserSpecifiedConfigNamespace,
```
candita, line 81: Do we need `operatorcontroller.GlobalUserSpecifiedConfigNamespace` twice?

gcs278:
/hold
Still want to assess impact to existing controllers (hopefully CI picks up any issues).

pkg/operator/operator.go
gcs278: Oh good catch! No, there is no reason to have them twice.

gcs278: Done.

gcs278:
For reference: The controllers added after the original global-inclusion-of-all-namespaces change was made:
1. Ingressclass
2. sync-http-error-code-configmap
3. crl
4. clientca-configmap
5. configurable-route

These are the controller more susceptible to using this existing "catch all" for including objects in cache they didn't explicitly say to be included in the cache.

gcs278:
Failures from e2e-aws-operator-gather-must-gather and TestRouterCompressionOperation which are pretty common.

/retest

gcs278:
e2e-aws-singel-node failed with `alert KubeAggregatedAPIDown fired for 60 seconds with labels: {name="v1.apps.openshift.io", namespace="default", severity="warning"}`

Not related to ingress controller.

/retest

gcs278:
Something about hostname could not be determined for integrated registry. Not related.

/retest

Miciah:
Excellent!
/lgtm

gcs278:
/unhold
I did assessments of every controller and cache call. I don't believe we are using objects outside of the specified namespaces in `MultiNamespacedCacheBuilder`

PR #743, opened 2022-04-19
Bug 2007246: Add allowPrivilegeEscalation to the router container

brandisher:
The reason for this change is that the pods associated with the default
IngressController don't explicitly request allowPrivilegeEscalation
which allows them to acquire an SCC that's too restrictive.


suleymanakbas91 (commented):
In general, it looks good to me. Can we maybe add a unit test to check if security context is really set on the deployment? Something like [this](https://github.com/openshift/cluster-ingress-operator/blob/fc4ec60cbefe370ab1d24d5846dd8ea89a958876/pkg/operator/controller/ingress/deployment_test.go#L38) should be sufficient.

assets/router/deployment.yaml
```diff
@@ -18,6 +18,9 @@ spec:
         - name: router
           # image is set at runtime.
           imagePullPolicy: IfNotPresent
+          securityContext:
+            # See BZ 2007246
```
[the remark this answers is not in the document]
brandisher: Fixed!

brandisher:
> In general, it looks good to me. Can we maybe add a unit test to check if security context is really set on the deployment? Something like [this](https://github.com/openshift/cluster-ingress-operator/blob/fc4ec60cbefe370ab1d24d5846dd8ea89a958876/pkg/operator/controller/ingress/deployment_test.go#L38) should be sufficient.

Added in the latest push.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1253,6 +1253,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate, defaultIngr
 			LivenessProbe:   hashableProbe(container.LivenessProbe, defaultIngressController),
 			ReadinessProbe:  hashableProbe(container.ReadinessProbe, defaultIngressController),
 			StartupProbe:    hashableProbe(container.StartupProbe, defaultIngressController),
+			SecurityContext: container.SecurityContext,
```
Miciah, line 1256: You need also need to do the update in `deploymentConfigChanged`:
```diff
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1467,6 +1467,7 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment, defaultIngres
 	}
 	updated.Spec.Template.Spec.Volumes = volumes
 	updated.Spec.Template.Spec.NodeSelector = expected.Spec.Template.Spec.NodeSelector
+	updated.Spec.Template.Spec.Containers[0].SecurityContext = expected.Spec.Template.Spec.Containers[0].SecurityContext
 	updated.Spec.Template.Spec.Containers[0].Env = expected.Spec.Template.Spec.Containers[0].Env
 	updated.Spec.Template.Spec.Containers[0].Image = expected.Spec.Template.Spec.Containers[0].Image
 	copyProbe(expected.Spec.Template.Spec.Containers[0].LivenessProbe, updated.Spec.Template.Spec.Containers[0].LivenessProbe, defaultIngressController)
```
Please add a test case for this:
```diff
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -1317,6 +1317,17 @@ func TestDeploymentConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
+			description: "if the container security context changes",
+			mutate: func(deployment *appsv1.Deployment) {
+				v := true
+				sc := &corev1.SecurityContext{
+					AllowPrivilegeEscalation: &v,
+				}
+				deployment.Spec.Template.Spec.Containers[0].SecurityContext = sc
+			},
+			expect: true,
+		},
 	}
 
 	for _, tc := range testCases {
```

assets/router/deployment.yaml
```diff
@@ -18,6 +18,9 @@ spec:
         - name: router
           # image is set at runtime.
           imagePullPolicy: IfNotPresent
+          securityContext:
+            # See https://bugzilla.redhat.com/2007246
```
stlaz, line 22: with bugzilla going away, you may not want to leave links in your code

brandisher: I'd like to keep this reference for the sake of providing context; I believe Bugzilla will remain in read-only mode at least for a period of time, and this reference can still be found in JIRA via the BZ ID.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -68,6 +68,20 @@ func checkDeploymentHasContainer(t *testing.T, deployment *appsv1.Deployment, na
 	}
 }
 
+func checkRouterContainerSecurityContext(t *testing.T, deployment *appsv1.Deployment) {
+	routerContainerName := "router"
```
Miciah, line 72: Can you make this a constant and move it to the top of the file?
```go
const (
	routerContainerName = "router"
)
```
We should also update other tests to use the named constant, but that can be left for future PRs.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -68,6 +68,20 @@ func checkDeploymentHasContainer(t *testing.T, deployment *appsv1.Deployment, na
...
+	t.Helper()
+	checkDeploymentHasContainer(t, deployment, routerContainerName, true)
+
+	for _, container := range deployment.Spec.Template.Spec.Containers {
+		if container.Name == routerContainerName {
+			if *(container.SecurityContext.AllowPrivilegeEscalation) != true {
```
Miciah, line 78:
```suggestion
			if v := container.SecurityContext.AllowPrivilegeEscalation; v == nil || *v != true {
```

pkg/operator/controller/ingress/deployment_test.go
brandisher: Fixed!

pkg/operator/controller/ingress/deployment_test.go
brandisher: Fixed!

Miciah:
Thanks!
/lgtm

Miciah:
@enxebre,

> @brandisher @Miciah can you help me understand why the need to use allowPrivilegeEscalation? what of https://kubernetes.io/docs/concepts/security/pod-security-policy/#privilege-escalation / https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt is needed?

If I recall correctly, we need `allowPrivilegeEscalation` because we have `setcap 'cap_net_bind_service=ep' /usr/sbin/haproxy` in https://github.com/openshift/router/blob/49dce8fe45184a4b52121b60a39561fedd4a2cfb/images/router/haproxy/Dockerfile#L9 to enable HAProxy to bind to lower-numbered ports 80 and 443.

PR #744, opened 2022-04-20
BUG 2076984: test/e2e: add resilience to RBAC test teardown

frobware:
RBAC tests fail in teardown with:

    configurable_route_test.go:301: failed to restore cluster ingress resource to original state: Operation cannot be fulfilled on ingresses.config.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again
    configurable_route_test.go:305: failed to restore cluster ingress resource to original state: Operation cannot be fulfilled on ingresses.config.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again

Change the defer logic for all the RBAC e2e tests to make multiple
attempts to restore both the spec and status when the test completes.

### DIFF
--- a/test/e2e/configurable_route_test.go
+++ b/test/e2e/configurable_route_test.go
@@ -39,11 +39,11 @@ func TestConfigurableRouteRBAC(t *testing.T) {
 			t.Fatalf("failed to get ingress resource: %v", err)
 		}
 		ingress.Spec.ComponentRoutes = nil
-		if err := kclient.Update(context.TODO(), ingress); err != nil {
+		if err := eventuallyUpdateIngressSpec(t, ingress.Spec); err != nil {
 			t.Errorf("failed to restore cluster ingress.spec resource to original state: %v", err)
 		}
 		ingress.Status.ComponentRoutes = nil
-		if err := kclient.Status().Update(context.TODO(), ingress); err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/configurable_route_test.go
```diff
@@ -39,12 +39,12 @@ func TestConfigurableRouteRBAC(t *testing.T) {
 			t.Fatalf("failed to get ingress resource: %v", err)
 		}
 		ingress.Spec.ComponentRoutes = nil
-		if err := kclient.Update(context.TODO(), ingress); err != nil {
-			t.Errorf("failed to restore cluster ingress.spec resource to original state: %v", err)
+		if err := eventuallyUpdateIngressSpec(t, ingress.Spec); err != nil {
+			t.Fatalf("failed to restore cluster ingress spec resource to original state: %v", err)
```
Miciah, line 43: This is the cleanup handler; does it make sense for a failure to revert the spec to prevent reverting the status?

frobware: It's not always a failure though. I'm attempting to deflake the test. If we want to make more wholesale changes than that warrants a different PR.

frobware: @Miciah I changed the calls from `Fatalf` back to `Errorf`.

Miciah:
Thanks!
/lgtm

PR #718, opened 2022-03-10
Bug 2007246: Ingress Controller does not set allowPrivilegeEscalation in the router deployment

brandisher:
This will ensure that the router pod gets selected into the correct SCC to avoid conflicts with custom SCCs.

PR #746, opened 2022-04-21
router deployment: add NET_BIND_SERVICE capability

stlaz:
Router requires the 80 and 443 ports in order to function properly.
These are reserved ports and so the container needs the NET_BIND_SERVICE
capability in order to work in a drop ALL caps environments.

---

This PR is currently blocking https://github.com/openshift/cluster-kube-apiserver-operator/pull/1338 from succeeding

/cc @s-urbaniak


stlaz:
Aaaand these PRs are now blocking each other :facepalm:

Miciah:
Is this needed in addition to <https://github.com/openshift/cluster-ingress-operator/pull/718>?

Miciah:
Note that #718 was reverted, and the change is re-added with <https://github.com/openshift/cluster-ingress-operator/pull/743>.

deads2k:
I think enumerating the capabilities is both different and valuable

Miciah:
> I think enumerating the capabilities is both different and valuable

Do they warrant separate Bugzilla reports?  Is there a reason we wouldn't want to backport both changes to 4.10?  (We plan to backport #743.)  Adding the `NET_BIND_SERVICE` capability and adding `allowPrivilegeEscalation: true` still seem similar enough to combine into a single PR (I can ask Chad to cherry-pick the commit from this PR, preserving authorship).  If you prefer to track this separately though, that's fine; we'll just need a separate BZ, and we should merge #743 first since it adds logic to handle updates to the security context.

stlaz:
> I think enumerating the capabilities is both different and valuable
absolutely

I temporarily added the restricted-v2 SCC to unblock this PR.

> Do they warrant separate Bugzilla reports?
No bugzilla needed for this PR

> Is there a reason we wouldn't want to backport both changes to 4.10?
Unfortunately, the way SCCs were previously set up, I think only the `privileged` SCC would match the added capabilities :frowning_face: 

>  Adding the `NET_BIND_SERVICE` capability and adding `allowPrivilegeEscalation: true`
I don't understand, this is only adding NET_BIND_SERVICE. I don't think the container needs `allowPrivilegeEscalation: true`, at least it seemed to work just fine in my testing (see the SCC that's attached now).

stlaz:
Okay, it seems that I was wrong and was checking what happens in an environment where the privilege escalation was still possible.

In that case we can ignore this PR in favour of https://github.com/openshift/cluster-ingress-operator/pull/743

Miciah:
> No bugzilla needed for this PR

I mean, Prow won't let you merge without either a BZ or a Jira issue with required approvals.  

> > Is there a reason we wouldn't want to backport both changes to 4.10?
> 
> Unfortunately, the way SCCs were previously set up, I think only the `privileged` SCC would match the added capabilities frowning_face

I think we used "privileged" when using container networking and "hostnetwork" when using the host network.  Is there a more restrictive SCC that would work for one or both of these types of networking?  

> This is only adding NET_BIND_SERVICE. I don't think the container needs `allowPrivilegeEscalation: true`, at least it seemed to work just fine in my testing (see the SCC that's attached now).

The reason `allowPrivilegeEscalation: true` is needed is that some users add custom SCCs that match the requirements as described in the pod template spec but that don't specify `allowPrivilegeEscalation: true`.  IIRC, privilege escalation is required because the router image uses a filesystem attribute to imbue `/bin/openshift-router` with the capability to bind low-numbered ports.

Miciah:
Ah, the moment I posted my comment, your comment appears.  I should have reloaded before commenting, sorry.

PR #750, opened 2022-04-29
Bug 2080379: Add some resilience to TestIngressControllerServiceNameCollision

frobware:
In CI this test flakes. The assertServiceNotDeleted() calls t.Helper()
but that function goes on to call t.Fatalf() serveral times. What gets
reported in test runs are line numbers with respect to the caller
which doesn't help isolate which part of assertServiceNotDeleted() was
failing.

Looking at CI logs we see cases like:

```console
operator_test.go:2792: expected openshift-ingress/router-e2e-name-collision-test to be present: Get "https://api.ci-op-rjh6srkl-43abb.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/openshift-ingress/services/router-e2e-name-collision-test": dial tcp 13.58.253.110:6443: connect: connection refused
```

I restructured the test to repeatedly retry for 1 minute if we failed
to update the service in case that error was really a transient
failure (e.g., "connection refused").


frobware:
Timed out waiting for node count (5) to equal or exceed machine count (6).

/retest

PR #751, opened 2022-04-29
h2spec: Print errors from Cobra and h2spec

Miciah:
* `test/h2spec/client.go` (`NewClientCommand`): Don't silence Cobra errors.
(`run`): Log errors from `h2spec.Run`.

### DIFF
--- a/test/h2spec/client.go
+++ b/test/h2spec/client.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"time"
 
+	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
 	"github.com/spf13/cobra"
 	"github.com/summerwind/h2spec"
 	h2specconfig "github.com/summerwind/h2spec/config"
@@ -15,6 +16,8 @@ import (
 var (
 	version string = "2.6.0"
 	commit  string = "70ac2294010887f48b18e2d64f5cccd48421fad1"
+
+	log = logf.Logger.WithName("h2spec")
 )
 
 func NewClientCommand() *cobra.Command {
@@ -28,7 +31,6 @@ func NewClientCommand() *cobra.Command {
 	}
 
 	cmd.SilenceUsage = true
-	cmd.SilenceErrors = true
 
 	flags := cmd.Flags()
 	flags.StringP("host", "h", "127.0.0.1", "Target host")
@@ -148,6 +150,7 @@ func run(cmd *cobra.Command, args []string) error {
 
 	success, err := h2spec.Run(c)
 	if !success {
+		log.Error(err, "error")
 		os.Exit(1)
 	}
 


test/h2spec/client.go
```diff
@@ -148,6 +150,7 @@ func run(cmd *cobra.Command, args []string) error {
 
 	success, err := h2spec.Run(c)
 	if !success {
+		log.Error(err, "error")
```
candita, line 153: The only error from `h2spec.Run`comes from the attempt to generate a junit report, so this is probably not useful, but otherwise lgtm.  https://github.com/openshift/cluster-ingress-operator/blob/8d9b42212a8ed0484f51d5a3cf9ede958878ad48/vendor/github.com/summerwind/h2spec/h2spec.go#L62

candita:
>  operator_test.go:1803: failed to update infrastructure config: Put "https://api.ci-op-y04gntz8-43abb.origin-ci-int-aws.dev.rhcloud.com:6443/apis/config.openshift.io/v1/infrastructures/cluster": http2: client connection lost
> --- FAIL: TestIngressControllerCustomEndpoints (45.00s)

/test e2e-aws-operator

test/h2spec/client.go
frobware: As @candita already mentioned. The only error from `Run()` is when junit fails. If a conformance run fails some of the tests then `err` will be `nil`.
 

```console
$ ./ingress-operator h2spec -t -h www.redhat.com
Finished in 13.2878 seconds
146 tests, 137 passed, 4 skipped, 5 failed
2022-05-03T11:51:55.190+0100	ERROR	operator.h2spec	h2spec/client.go:29	error
```

frobware: On the openshift/origin side we assert success or failure according to the comments here: https://github.com/openshift/origin/blob/master/test/extended/router/h2spec.go#L247

candita:
> === RUN   TestHTTPCookieCapture
>     operator_test.go:2098: failed to read logs from pod router-cookiecapture-56fbf87d97-78nfx: pods "router-cookiecapture-56fbf87d97-78nfx" not found
/test e2e-aws-operator

test/h2spec/client.go
```diff
@@ -28,7 +31,6 @@ func NewClientCommand() *cobra.Command {
 	}
 
 	cmd.SilenceUsage = true
-	cmd.SilenceErrors = true
```
frobware, line 31: This allows timeout errors to be reported, but not as a returned error value; the h2spec code just prints these directly: https://github.com/openshift/cluster-ingress-operator//blob/master/vendor/github.com/spf13/cobra/command.go#L986.

frobware:
/lgtm

Given the structure of h2spec there's limited return on the changes. But printing an error if junit fails may be useful, although we try and accommodate that failure in the origin tests anyway: https://github.com/openshift/origin/blob/master/test/extended/router/h2spec.go#L247.

Adding /hold. If you're happy with the observed comprises then feel free to remove the hold.


/hold

PR #754, opened 2022-05-04
 Add support for route53 in the us-isob-east-1 region + fix

r4f4:
This PR is just #703 with the requested fix implemented.

### DIFF
--- a/pkg/dns/aws/dns.go
+++ b/pkg/dns/aws/dns.go
@@ -142,15 +142,20 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	elbConfig := aws.NewConfig().WithRegion(region)
 	tagConfig := aws.NewConfig()
 
+	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
+	if !ok {
+		return nil, fmt.Errorf("Unable to determine partition from region: %q", region)
+	}
+
 	// If the region is in aws china, cn-north-1 or cn-northwest-1, we should:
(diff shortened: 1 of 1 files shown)


r4f4:
CC @rfredette for review

pkg/dns/aws/dns.go
```diff
@@ -142,26 +142,31 @@ func NewProvider(config Config, operatorReleaseVersion string) (*Provider, error
 	elbConfig := aws.NewConfig().WithRegion(region)
 	tagConfig := aws.NewConfig()
 
+	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
+	if !ok {
```
rna-afk, line 146: Not sure but did you check if partition would be nil if ok is false? The switch case would throw error in that case since it's a function call with nil reference.

r4f4: @rna-afk thanks for the review! Yes, I checked the code for `PartitionForRegion` in the SDK and an empty struct is returned in case of error https://github.com/aws/aws-sdk-go/blob/v1.44.6/aws/endpoints/endpoints.go#L384-L392. So it should be safe to call `ID()` on it, which will result in an empty string.

rfredette:
thanks!
/lgtm

mjpytlak:
Providing doc ack. No doc is required for this PR

patrickdillon:
We have the buy in from dos, qe, and px. Furthermore, as we are (currently) a FF team, I should be able to apply these labels myself:

/label docs-approved
/label qe-approved
/label px-approved

PR #756, opened 2022-05-06
Bug 2080379: Group all e2e tests as parallel or serial

frobware:
This PR changes the entry point for the e2e test to the new function `TestAll`.

The new test/e2e/`TestAll` function groups tests that can be run in parallel and those that must run in series via two overarching sub-tests. The new `TestAll` function guarantees consistent ordering of all e2e tests; serial tests will not start until the parallel tests run to completion. Parallel tests are those that do not mutate the ingressconfig object, the default ingress controller, the infrastructure object, default certificates, et al. Those tests that stand up a private ingress controller, for example, are deemed to be self-contained and are grouped into the parallel set. Each test that can run in parallel must call `t.Parallel()`.

- All tests can be run via `make test-e2e`. 
- The set of all serial tests can be run with `make test-e2e TEST=TestAll/serial`
- The set of all parallel tests can be run with `make test-e2e TEST=TestAll/parallel`
- And individual tests (e.g., during development) can be run with `make test-e2e TEST=TestHstsPolicyWorks`

I have been soak testing both the serial and parallel tests to use `go test -count=20`. The intent is to flush out cases where, in particular, the serial tests continue to flake. Typically any error has show up to be a conflict error on update. For those cases I have identified I have updated the tests to retry on conflict errors.

Running some of the e2e tests in parallel reduces the elapsed time for the e2e tests from ~55 minutes to ~10 mins. The reduction in time for the set of identified parallel tests is ~28 mins => ~5 mins.

There is a new step to the `make verify` target. If a new test is added it must be added to the set of parallel or serial tests defined in ./test/e2e/all_test.go. Running `make verify` will identify any tests (via the `./hack/verify-e2e-test-all-presence.sh` script) that are not listed in `all_test.go`. For example: 

```console
$ make verify
hack/verify-gofmt.sh
hack/verify-generated-crd.sh
hack/verify-profile-manifests.sh
hack/verify-generated-bindata.sh
hack/verify-deps.sh
hack/verify-e2e-test-all-presence.sh
error: test function 'TestTunableMaxConnectionsValidValues' not called by 'TestAll' in /home/aim/src/github.com/openshift/cluster-ingress-operator/hack/../test/e2e/all_test.go
make: *** [Makefile:70: verify] Error 1

```

[...]

PR #764, opened 2022-05-27
Bug 2075671: Add a e2e test that verifies the ingress operator's cache doesn't include everything

gcs278:
Adding a simple e2e for BZ2075671 (that was already closed) to validate that the ingress operator's cache isn't including all objects.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -3078,6 +3078,52 @@ func newLoadBalancerController(name types.NamespacedName, domain string) *operat
 	}
 }
 
+// TestIngressOperatorCacheIsNotGlobal validates BZ2075671: Don't include all objects in all namespaces in the
+// Ingress Operator's cache. This tests adds an Ingress Controller in another namespace that isn't in the
+// Ingress Operator's cache and ensures the Ingress Operator ignores it.
+func TestIngressOperatorCacheIsNotGlobal(t *testing.T) {
+	t.Parallel()
+	// Setup a namespace that will be ignored by the ingress operator.
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ignored-ns",
+		},
+	}
+	if err := kclient.Create(context.TODO(), ns); err != nil {
+		t.Fatalf("failed to create namespace: %v", err)
+	}
+	defer func() {
+		if err := kclient.Delete(context.TODO(), ns); err != nil {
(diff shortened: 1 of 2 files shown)


test/e2e/operator_test.go
```diff
@@ -3035,6 +3035,51 @@ func newLoadBalancerController(name types.NamespacedName, domain string) *operat
...
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+
+	// Clean up our new Ingress Controller after we are done.
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Verify new Ingress Controller is ignored by ensuring its status never gets modified.
```
candita, line 3070: Is an unmodified status the only way to check the default IC cache?

gcs278: Seems like the best way as far as I can tell. I diffed the YAML of a IC that was admitted and one that was not. In comparison, there are a couple things different, but it seems like status is this most direct way to test  The other things are default values that get added by the ingress operator (that might change in the future).

test/e2e/operator_test.go
```diff
@@ -3035,6 +3035,51 @@ func newLoadBalancerController(name types.NamespacedName, domain string) *operat
...
+	wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), icName, ic); err != nil {
+			t.Fatalf("failed to get ingresscontroller %s/%s: %v", ic.Namespace, ic.Name, err)
+		}
+
+		if ic.Status.Conditions != nil {
```
Miciah, line 3076: I get confused by the semantics of nil versus empty with API objects.  Might be safer to check `len(ic.Status.Conditions) != 0`.

Miciah:
```
error: test function 'TestIngressOperatorCacheIsNotGlobal' not called by 'TestAll' in /go/src/github.com/openshift/cluster-ingress-operator/hack/../test/e2e/all_test.go 
```
The verify job is now failing since #756 merged. It looks like the new E2E test can run in parallel with other tests, so it should call `t.Parallel()` and be included in the list of parallel tests in `test/e2e/all_test.go`.

test/e2e/operator_test.go
gcs278: That's a fair point.

PR #767, opened 2022-06-01
Bug 2094932: MGMT-10403: Set the `defaultPlacement` for none-platform SNO clusters installed before 4.11

omertuc:
Added a call to a new `handleSingleNode4Dot11Upgrade` function that sets
the defaultPlacement status in the ingress config CR of none-platform
single node clusters to "ControlPlane" if it's not already set. The
situations in which this value is not set are in clusters upgraded from
<4.11 to >=4.11. When that field is not set, it is assumed by this
operator's controllers to have the value "Workers", which is fine for
most clusters, but not a good default for none-platform single-node
clusters as it causes issues when adding workers to them (and is why
starting with 4.11 freshly installed none-platform have it set to
"ControlPlane"). The purpose of this function is to then "fix" those
older clusters so they behave more like freshly installed >=4.11
clusters. Before doing that we also verify that the cluster truly has
only one node, as we wouldn't want to affect clusters that have workers.
That's because It's likely that such clusters already have a custom
ingress configuration that we wouldn't want to potentially mess up, even
though adding workers was previously unsupported.
If defaultPlacement is unset and the cluster doesn't match the criteria
mentioned above, it is anyway given the explicit "Workers" value to
ensure this function doesn't affect it in the future no matter what
changes (e.g. number of nodes), and to make it more consistent with
freshly installed >=4.11 clusters.

This was originally a [non-goal of the enhancement](https://github.com/openshift/enhancements/blob/c2ceb84d86c030650b7b4092d4ee924b16f43faf/enhancements/single-node/single-node-openshift-with-workers.md?plain=1#L185-L190)
but I don't see any harm in fixing it in this manner, it will simplifiy
the need for manual actions and the need to document the difference
between 4.11 clusters and "freshly installed 4.11 clusters" when it
comes to SNO worker expansion.


pkg/operator/operator.go
```diff
@@ -259,6 +265,74 @@ func (o *Operator) determineReplicasForDefaultIngressController() (int32, error)
...
+	}
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		if err := o.client.Get(context.TODO(), ingressConfigName, ingressConfig); err != nil {
+			return err
+		}
+
+		patch := client.MergeFrom(ingressConfig.DeepCopy())
+		ingressConfig.Status.DefaultPlacement = desiredDefaultPlacement
+		err := o.client.Status().Patch(context.TODO(), ingressConfig, patch)
```
romfreiman, line 326: log

Miciah, line 326: You'll need to update `manifests/00-cluster-role.yaml` to allow this.

manifests/00-cluster-role.yaml
```diff
@@ -118,6 +118,7 @@ rules:
   - config.openshift.io
   resources:
   - clusteroperators/status
+  - ingresses
```
Miciah, line 121:
```suggestion
  - ingresses/status
```

pkg/operator/operator.go
```diff
@@ -232,6 +233,11 @@ func (o *Operator) Start(ctx context.Context) error {
 		}
 	}, 1*time.Minute, ctx.Done())
 
+	err := o.handleSingleNode4Dot11Upgrade()
+	if err != nil {
```
Miciah, line 237:
```suggestion
	if err := o.handleSingleNode4Dot11Upgrade(); err != nil {
```

pkg/operator/operator.go
```diff
@@ -259,6 +265,76 @@ func (o *Operator) determineReplicasForDefaultIngressController() (int32, error)
...
+// it's not already set. The situations in which this value is not set are in
+// clusters upgraded from <4.11 to >=4.11. When that field is not set, it is
+// assumed by this operator's controllers to have the value "Workers", which is
+// fine for most clusters, but not a good default for none-platform single-node
+// clusters as it causes issues when adding workers to them (and is why
+// starting with 4.11 freshly installed none-platform have it set to
+// "ControlPlane"). The purpose of this function is to then "fix" those older
+// clusters so they behave more like freshly installed >=4.11 clusters. Before
+// doing that we also verify that the cluster truly has only one node, as we
+// wouldn't want to affect clusters that have workers. That's because It's
```
Miciah, line 279:
```suggestion
// wouldn't want to affect clusters that have workers. That's because it's
```

pkg/operator/operator.go
```diff
@@ -259,6 +264,76 @@ func (o *Operator) determineReplicasForDefaultIngressController() (int32, error)
...
+func (o *Operator) handleSingleNode4Dot11Upgrade() error {
+	ingressConfigName := types.NamespacedName{Name: "cluster"}
+
+	infraConfig := &configv1.Infrastructure{}
+	if err := o.client.Get(context.TODO(), ingressConfigName, infraConfig); err != nil {
+		return fmt.Errorf("failed fetching infrastructure config: %w", err)
+	}
+
+	ingressConfig := &configv1.Ingress{}
+	if err := o.client.Get(context.TODO(), ingressConfigName, ingressConfig); err != nil {
```
Miciah, line 296:
```suggestion
	if err := o.client.Get(context.TODO(), operatorcontroller.IngressClusterConfigName(), ingressConfig); err != nil {
```

pkg/operator/operator.go
```diff
@@ -259,6 +264,76 @@ func (o *Operator) determineReplicasForDefaultIngressController() (int32, error)
...
+// If defaultPlacement is unset and the cluster doesn't match the criteria
+// mentioned above, it is anyway given the explicit "Workers" value to ensure
+// this function doesn't affect it in the future no matter what changes (e.g.
+// number of nodes), and to make it more consistent with freshly installed
+// >=4.11 clusters.
+func (o *Operator) handleSingleNode4Dot11Upgrade() error {
+	ingressConfigName := types.NamespacedName{Name: "cluster"}
+
+	infraConfig := &configv1.Infrastructure{}
+	if err := o.client.Get(context.TODO(), ingressConfigName, infraConfig); err != nil {
```
Miciah, line 291: Using `ingressConfigName` for the infra config object's name works because the objects have the same name, but it looks strange.  Would you mind adding a function to `names.go` like `IngressClusterConfigName` and using that here?

pkg/operator/operator.go
```diff
@@ -259,6 +264,76 @@ func (o *Operator) determineReplicasForDefaultIngressController() (int32, error)
...
+
+	if len(nodes.Items) == 1 &&
+		infraConfig.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode &&
+		infraConfig.Status.InfrastructureTopology == configv1.SingleReplicaTopologyMode &&
+		infraConfig.Status.PlatformStatus.Type == configv1.NonePlatformType {
+		desiredDefaultPlacement = configv1.DefaultPlacementControlPlane
+	}
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		if err := o.client.Get(context.TODO(), ingressConfigName, ingressConfig); err != nil {
```
Miciah, line 319:
```suggestion
		if err := o.client.Get(context.TODO(), operatorcontroller.IngressClusterConfigName(), ingressConfig); err != nil {
```

pkg/operator/operator.go
omertuc: Ah absolutely, didn't even notice I did that

Miciah:
must-gather failed.
/test e2e-aws-operator

Miciah:
```
Failing invariants:

[bz-monitoring][invariant] alert/Watchdog must have no gaps or changes
```
I'm hoping that the latest fix for [BZ#2079679](https://bugzilla.redhat.com/show_bug.cgi?id=2079679) will help with this failure.  The fix is in [the 4.11.0-0.ci-2022-06-03-123056 payload](https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4.11.0-0.ci/release/4. 11.0-0.ci-2022-06-03-123056), which has been accepted.  
/test e2e-aws-single-node

Miciah:
must-gather/post stage failed.
/test e2e-aws-operator

Miciah:
`TestRouteAdmissionPolicy` failed, and must-gather failed.
/test e2e-aws-operator

Miciah:
Still getting this:
```
Failing invariants:

[bz-monitoring][invariant] alert/Watchdog must have no gaps or changes
```
https://search.ci.openshift.org/?search=Watchdog+must+have+no+gaps+or+changes&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=cluster-ingress-operator&maxMatches=5&maxBytes=20971520&groupBy=job shows beaucoup failures in other repositories, so I'm fairly confident that the failure isn't caused by changes in this PR.
/test e2e-aws-single-node

Miciah:
Must-gather.
/test e2e-aws-single-node

omertuc:
The e2e tests step fails to generate the junit file at the end because it runs out of memory and gets OOM'd even though the pod has a limit of 4GB. One of the tests is probably generating a huge output or something...

Has nothing to do with this PR though, seems to affect all instances of this job everywhere :/

omertuc:
Looks like the junit issue is gone.

But e2e-aws-single-node is permafailing even on periodics due to this watchdog thing, so it can be safely ignored

PR #769, opened 2022-06-04
Bug 2066560: Make ingress clusteroperator progressing=true when router deployment is rolling out

gcs278:
When a router deployment the ingress controller is rolling out, set progressing=true in the cluster operator status with a message.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -73,12 +73,14 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentRollingOutCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeLoadBalancerProgressingStatus(ic, service, platformStatus))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDNSStatus(ic, wildcardRecord, platformStatus, dnsConfig)...)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressAvailableCondition(updated.Status.Conditions))
 	degradedCondition, err := computeIngressDegradedCondition(updated.Status.Conditions, updated.Name)
 	errs = append(errs, err)
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressProgressingCondition(updated.Status.Conditions, ic, service, platformStatus))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressProgressingCondition(updated.Status.Conditions))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, platformStatus, secret))
 
@@ -457,6 +459,55 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
 	}
 }
 
+// computeDeploymentRollingOutCondition computes the ingress controller's
+// "DeploymentRollingOut" status condition by examining the number of updated
+// replicas reported in the deployment's status. The "DeploymentRollingOut"
+// condition is true if the number of updated replicas is not equal to the number
+// of expected or available replicas.
+// See Reference: https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go
+func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+	// If have replicas is less than want replicas, then we are waiting for replicas to be updated.
+	if deployment.Spec.Replicas != nil && deployment.Status.UpdatedReplicas < *deployment.Spec.Replicas {
+		return operatorv1.OperatorCondition{
+			Type:   IngressControllerDeploymentRollingOutConditionType,
+			Status: operatorv1.ConditionTrue,
+			Reason: "DeploymentRollingOut",
+			Message: fmt.Sprintf(
+				"Waiting for router deployment rollout to finish: %d out of %d new replica(s) have been updated...\n",
+				deployment.Status.UpdatedReplicas, *deployment.Spec.Replicas),
+		}
+	}
+	// If have replicas greater than updated replicas, then we are waiting for old replicas to terminate.
+	if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
+		return operatorv1.OperatorCondition{
+			Type:   IngressControllerDeploymentRollingOutConditionType,
+			Status: operatorv1.ConditionTrue,
+			Reason: "DeploymentRollingOut",
+			Message: fmt.Sprintf(
+				"Waiting for router deployment rollout to finish: %d old replica(s) are pending termination...\n",
+				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
+		}
+	}
+	// If available replicas less than updated replicas, then we are waiting for updated replicas to become available.
+	if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
+		return operatorv1.OperatorCondition{
+			Type:   IngressControllerDeploymentRollingOutConditionType,
+			Status: operatorv1.ConditionTrue,
+			Reason: "DeploymentRollingOut",
+			Message: fmt.Sprintf(
+				"Waiting for router deployment rollout to finish: %d of %d updated replica(s) are available...\n",
+				deployment.Status.AvailableReplicas, deployment.Status.UpdatedReplicas),
+		}
+	}
+
+	return operatorv1.OperatorCondition{
+		Type:    IngressControllerDeploymentRollingOutConditionType,
+		Status:  operatorv1.ConditionFalse,
+		Reason:  "DeploymentNotRollingOut",
+		Message: "Deployment is not actively rolling out",
+	}
+}
+
 // computeIngressDegradedCondition computes the ingresscontroller's "Degraded"
 // status condition, which aggregates other status conditions that can indicate
 // a degraded state.  In addition, computeIngressDegradedCondition returns a
@@ -641,43 +692,42 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
 	return formatted
 }
 
-// computeIngressProgressingCondition computes the ingresscontroller's
-// "Progressing" status condition, which indicates if the ingresscontroller's
-// current state matches its desired state.
-func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition, ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus) operatorv1.OperatorCondition {
-	condition := operatorv1.OperatorCondition{
-		Type:   operatorv1.OperatorStatusTypeProgressing,
-		Status: operatorv1.ConditionFalse,
+// computeIngressProgressingCondition computes the IngressController's current Progressing status state
+// by inspecting the following:
+// 1) the Deployment Replicas Updated condition of the IngressController
+// 2) the LoadBalancer Progressing condition of the IngressController
+// The IngressController is judged NOT Progressing only if all 2 conditions are true; otherwise
+// it is considered to be Progressing.
+func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+	expected := []expectedCondition{
+		{
+			condition:        IngressControllerLoadBalancerProgressingConditionType,
+			status:           operatorv1.ConditionFalse,
+			ifConditionsTrue: []string{operatorv1.LoadBalancerManagedIngressConditionType},
+		},
+		{
+			condition: IngressControllerDeploymentRollingOutConditionType,
+			status:    operatorv1.ConditionFalse,
+		},
 	}
-	if ic.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType {
-		switch {
-		case ic.Status.EndpointPublishingStrategy.LoadBalancer == nil:
-			condition.Message = "status.endpointPublishingStrategy.loadBalancer is not set."
-			condition.Reason = "IncompleteStatus"
-			condition.Status = operatorv1.ConditionUnknown
-		case service == nil:
-			condition.Message = "LoadBalancer Service not created."
-			condition.Reason = "NoService"
-			condition.Status = operatorv1.ConditionTrue
-		default:
-			wantScope := ic.Status.EndpointPublishingStrategy.LoadBalancer.Scope
-			haveScope := operatorv1.ExternalLoadBalancer
-			if IsServiceInternal(service) {
-				haveScope = operatorv1.InternalLoadBalancer
-			}
-			if wantScope != haveScope {
-				message := fmt.Sprintf("The IngressController scope was changed from %q to %q.", haveScope, wantScope)
-				switch platform.Type {
-				case configv1.AWSPlatformType, configv1.IBMCloudPlatformType:
-					message = fmt.Sprintf("%[1]s  To effectuate this change, you must delete the service: `oc -n %[2]s delete svc/%[3]s`; the service load-balancer will then be deprovisioned and a new one created.  This will most likely cause the new load-balancer to have a different host name and IP address from the old one's.  Alternatively, you can revert the change to the IngressController: `oc -n openshift-ingress-operator patch ingresscontrollers/%[4]s --type=merge --patch='{\"spec\":{\"endpointPublishingStrategy\":{\"loadBalancer\":{\"scope\":\"%[5]s\"}}}}'", message, service.Namespace, service.Name, ic.Name, haveScope)
-				}
-				condition.Reason = "ScopeChanged"
-				condition.Message = message
-				condition.Status = operatorv1.ConditionTrue
+
+	// Check for the rare case of no conditions
+	if len(conditions) != 0 {
+		_, progressingConditions, _ := checkConditions(expected, conditions)
+		if len(progressingConditions) != 0 {
+			progressing := formatConditions(progressingConditions)
+			return operatorv1.OperatorCondition{
+				Type:    operatorv1.OperatorStatusTypeProgressing,
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "IngressControllerProgressing",
+				Message: "One or more status conditions indicate progressing: " + progressing,
 			}
 		}
 	}
-	return condition
+	return operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeProgressing,
+		Status: operatorv1.ConditionFalse,
+	}
 }
 
 // IngressStatusesEqual compares two IngressControllerStatus values.  Returns true
@@ -713,9 +763,11 @@ func conditionChanged(a, b operatorv1.OperatorCondition) bool {
 	return a.Status != b.Status || a.Reason != b.Reason || a.Message != b.Message
 }
 
-// computeLoadBalancerStatus returns the complete set of current
-// LoadBalancer-prefixed conditions for the given ingress controller.
+// computeLoadBalancerStatus returns the set of current
+// LoadBalancer-prefixed conditions for the given ingress controller, which are
+// used later to determine the ingress controller's Degraded or Available status.
 func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1.Service, operandEvents []corev1.Event) []operatorv1.OperatorCondition {
+	// Compute the LoadBalancerManagedIngressConditionType condition
 	if ic.Status.EndpointPublishingStrategy == nil ||
 		ic.Status.EndpointPublishingStrategy.Type != operatorv1.LoadBalancerServiceStrategyType {
 		return []operatorv1.OperatorCondition{
@@ -737,6 +789,7 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
 		Message: "The endpoint publishing strategy supports a managed load balancer",
 	})
 
+	// Compute the LoadBalancerReadyIngressConditionType condition
 	switch {
 	case service == nil:
 		conditions = append(conditions, operatorv1.OperatorCondition{
@@ -775,8 +828,57 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
 			Message: message,
 		})
 	}
-
 	return conditions
+}
+
+// computeLoadBalancerProgressingStatus returns the LoadBalancerProgressing
+// conditions for the given ingress controller. These conditions subsequently determine
+// the ingress controller's Progressing status.
+func computeLoadBalancerProgressingStatus(ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus) operatorv1.OperatorCondition {
+	// Compute the IngressControllerLoadBalancerProgressingConditionType condition for the LoadBalancer
+	if ic.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType {
+		switch {
+		case ic.Status.EndpointPublishingStrategy.LoadBalancer == nil:
+			return operatorv1.OperatorCondition{
+				Type:    IngressControllerLoadBalancerProgressingConditionType,
+				Status:  operatorv1.ConditionUnknown,
+				Reason:  "StatusIncomplete",
+				Message: "status.endpointPublishingStrategy.loadBalancer is not set.",
+			}
+		case service == nil:
+			return operatorv1.OperatorCondition{
+				Type:    IngressControllerLoadBalancerProgressingConditionType,
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "NoService",
+				Message: "LoadBalancer Service not created.",
+			}
+		default:
+			wantScope := ic.Status.EndpointPublishingStrategy.LoadBalancer.Scope
+			haveScope := operatorv1.ExternalLoadBalancer
+			if IsServiceInternal(service) {
+				haveScope = operatorv1.InternalLoadBalancer
+			}
+			if wantScope != haveScope {
+				message := fmt.Sprintf("The IngressController scope was changed from %q to %q.", haveScope, wantScope)
+				switch platform.Type {
+				case configv1.AWSPlatformType, configv1.IBMCloudPlatformType:
+					message = fmt.Sprintf("%[1]s  To effectuate this change, you must delete the service: `oc -n %[2]s delete svc/%[3]s`; the service load-balancer will then be deprovisioned and a new one created.  This will most likely cause the new load-balancer to have a different host name and IP address from the old one's.  Alternatively, you can revert the change to the IngressController: `oc -n openshift-ingress-operator patch ingresscontrollers/%[4]s --type=merge --patch='{\"spec\":{\"endpointPublishingStrategy\":{\"loadBalancer\":{\"scope\":\"%[5]s\"}}}}'", message, service.Namespace, service.Name, ic.Name, haveScope)
+				}
+				return operatorv1.OperatorCondition{
+					Type:    IngressControllerLoadBalancerProgressingConditionType,
+					Status:  operatorv1.ConditionTrue,
+					Reason:  "ScopeChanged",
+					Message: message,
+				}
+			}
+		}
+	}
+	return operatorv1.OperatorCondition{
+		Type:    IngressControllerLoadBalancerProgressingConditionType,
+		Status:  operatorv1.ConditionFalse,
+		Reason:  "LoadBalancerNotProgressing",
+		Message: "LoadBalancer is not progressing",
+	}
 }
 
 func isProvisioned(service *corev1.Service) bool {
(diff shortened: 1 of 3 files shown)


gcs278:
/hold
Still have some edge cases I need to work through.

pkg/operator/controller/ingress/status.go
```diff
@@ -632,21 +633,41 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+	haveReplicas := deployment.Status.Replicas
+	// First check if we have the total number of desired replicas, then check if they are all up-to-date.
+	if haveReplicas != wantReplicas {
+		progressingReasons = append(progressingReasons, "Reconciling")
+		messages = append(messages, fmt.Sprintf("Have %d available router pods, want %d.", haveReplicas, wantReplicas))
+	} else {
+		updatedReplicas := deployment.Status.UpdatedReplicas
```
gcs278, line 655: I figured there no need to check update == have if have != want. If Have != want, then update usually != want and doesn't really tell us anything new.

pkg/operator/controller/ingress/status.go
```diff
@@ -632,21 +633,41 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+
+	// If deployment doesn't have the desired number of replicas, mark as progressing.
+	wantReplicas := int32(0)
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+	haveReplicas := deployment.Status.Replicas
+	// First check if we have the total number of desired replicas, then check if they are all up-to-date.
+	if haveReplicas != wantReplicas {
+		progressingReasons = append(progressingReasons, "Reconciling")
```
gcs278, line 652: I agree this seems funky, but I copied the pattern from https://github.com/openshift/cluster-dns-operator/blob/d50df32df68f53c1d47db8f5e51a8b27c402f278/pkg/operator/controller/dns_status.go#L221. 

It does seem a little awkward in the output:

> NAME      VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
> ingress   unknown   True        True          False      14m     ingresscontroller "default" is progressing: ReconcilingIncompleteStatus: Have 3 available router pods, want 2....

It only will output the first part of the error message that gets generated first (I put these new errors before than the loadbalancing checks...so always will be this new progressing error).

pkg/operator/controller/ingress/status.go
```diff
@@ -659,12 +680,20 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
...
+				messages = append(messages, message)
 			}
 		}
 	}
+	if len(progressingReasons) != 0 {
+		// Let ConditionUnknown status take precedence over ConditionTrue if it was set by logic above.
+		if condition.Status != operatorv1.ConditionUnknown {
+			condition.Status = operatorv1.ConditionTrue
+		}
+		condition.Reason = strings.Join(progressingReasons, "")
```
gcs278, line 693: Like I said, I copied from [here](https://github.com/openshift/cluster-dns-operator/blob/d50df32df68f53c1d47db8f5e51a8b27c402f278/pkg/operator/controller/dns_status.go#L221.). Should I join with "," since that might look cleaner or keep this consistent with DNS Operator like I have it now?

gcs278:
bunch of time outs
/retest

pkg/operator/controller/ingress/status.go
```diff
@@ -632,21 +633,41 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
 	}
+	messages := []string{}
+	progressingReasons := []string{}
+
+	// If deployment doesn't have the desired number of replicas, mark as progressing.
+	wantReplicas := int32(0)
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+	haveReplicas := deployment.Status.Replicas
```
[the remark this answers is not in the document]
Miciah: Never mind, that's what you do below.  Is it necessary to check both `Status.Replicas` and `Status.UpdatedReplicas`?

pkg/operator/controller/ingress/status.go
Miciah: `Reason` shouldn't have punctuation or spaces, so joining using `""` is appropriate.

Miciah: You could join on `"And"` so you'd get "ReconcilingAndIncompleteStatus" instead of "ReconcilingIncompleteStatus".  Both "Reconciling" and "IncompleteStatus" seem too vague though; they could be changed to "DeploymentProgressing" and "LoadBalancerPending" or similar, respectively.

pkg/operator/controller/ingress/status.go
```diff
@@ -632,21 +633,41 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition, ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus, deployment *appsv1.Deployment) operatorv1.OperatorCondition {
 	condition := operatorv1.OperatorCondition{
 		Type:   operatorv1.OperatorStatusTypeProgressing,
 		Status: operatorv1.ConditionFalse,
 	}
+	messages := []string{}
+	progressingReasons := []string{}
+
+	// If deployment doesn't have the desired number of replicas, mark as progressing.
+	wantReplicas := int32(0)
```
Miciah, line 645: If a deployment's `spec.Replicas` field is nil, the default value is 1.

pkg/operator/controller/ingress/status.go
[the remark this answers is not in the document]
Miciah: > Though rare, couldn't you get into a situation where you have have == want != updated? Where the wanted replicas and have replicas are equal but one is not updated?

Yes, that can happen during a rolling update (especially for an ingresscontroller with `spec.endpointPublishingStrategy.type: HostNetwork` where we cannot use surge).  We should report when want != updated; I was wondering rather whether we need to report when have != want.

pkg/operator/controller/ingress/status.go
```diff
@@ -644,21 +645,41 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
 // computeIngressProgressingCondition computes the ingresscontroller's
 // "Progressing" status condition, which indicates if the ingresscontroller's
 // current state matches its desired state.
-func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition, ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus) operatorv1.OperatorCondition {
+func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition, ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus, deployment *appsv1.Deployment) operatorv1.OperatorCondition {
```
candita, line 648: I wonder if it's necessary to recompute the replicas when we already have conditions for `computeDeploymentReplicasMinAvailableCondition` and `computeDeploymentReplicasAllAvailableCondition`.  Perhaps you could just add one more method `computeDeploymentReplicasAllUpdatedCondition` and use those conditions to determine Progressing status?

gcs278: Yea I'll admit this is pretty confusing to me, I've gone back and forth. cluster-ingress-operator and cluster-dns-operator do things in different patterns. But you've hit on a good point.

So I think those functions are doing slightly different things. It looks like they always add individual `status.condition` entries and are used for calculating the "degraded" (`operatorv1.OperatorStatusTypeDegraded`) in `oc get co` clusteroperator status. So each function adds independent `status.condition`.

Right now, this new logic stuffs all of the "progressing" errors messages in to 1 single condition, the progressing condition (`operatorv1.OperatorStatusTypeProgressing`). We could make a new `computeDeploymentProgressingCondition` function which would make a new individual `status.condition` (e.g. `DeploymentProgressing`) that would always need to be there. Then we could calculate `operatorv1.OperatorStatusTypeProgressing` just like `operatorv1.OperatorStatusTypeDegraded` is calculated.

However, it would involve a little bit of refactoring. Since the previous developers just packed (lazily? or intentionally?) the `operatorv1.OperatorStatusTypeProgressing` condition with `LoadBalancerStatusIncomplete` and `NoService`, these would need to become new individual status conditions (e.g. stop overloading `operatorv1.OperatorStatusTypeProgressing`).

I'm conflicted. I haven't found any standards on error messages. For example, `cluster-dns-operator` doesn't make individual condition entries for each status. It just packs the progressing, degraded, available condition messages with multiple appended error messages (https://github.com/openshift/cluster-dns-operator/blob/master/pkg/operator/controller/dns_status.go#L66-L70). 

I think the important thing here, is that the "real" status messages are `OperatorStatusTypeAvailable`, `OperatorStatusTypeProgressing`, and `OperatorStatusTypeDegraded` because these types of conditions determine the status of `oc get co`. I suppose we can do it however we want. If we want multiple individual error conditions, then aggregate it into a single `OperatorStatusType*` condition, I suppose that is fine.

Sorry for the long response....I had to spend a lot of time working this out in my head. I think I'm indifferent. I lean towards making them individual conditions because that's how `OperatorStatusTypeAvailable` and `OperatorStatusTypeProgressing` are done.

gcs278: Not sure if I'm following. I don't think the existing `status.conditions` offer a condition based on `status.UpdatedReplicas`. For reference, the `kubectl rollout` command uses mostly `status.UpdatedReplicas` as well as `status.AvailableReplicas` to determine if pod is actively rolling out:
https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go#L76-L89. I think that's what we should use here to determine the progressing status.

Both `computeDeploymentReplicasMinAvailableCondition` and `computeDeploymentReplicasAllUpdatedCondition` compute a status solely based on `Status.AvailableReplicas` as well as the `deployment.Spec.Strategy` specifiers, and though closely related, I don't think that's completely correct when calculating the progressing condition. I don't think either of these functions/conditions tell us if a pod is progressing.

Specifically `computeDeploymentReplicasAllUpdatedCondition` is very close to telling us if the deployment is progressing, but it just checks `deployment.Status.AvailableReplicas < replicas` which isn't enough info. You could be still rolling out and not trigger this condition. `computeDeploymentReplicasMinAvailableCondition` just says if we don't have enough replicas running, doesn't tell us if it's progressing.

candita: > Perhaps you could just add one more method `computeDeploymentReplicasAllUpdatedCondition` and use those conditions to determine Progressing status?

gcs278: What do you mean by "those conditions"? Are you suggesting to add a new `DeploymentReplicasAllUpdated` condition `status.condition.type` and use that new condition to calculate the progressing status?

Or are you suggesting to use the existing `status.conditions` generated by `computeDeploymentReplicasMinAvailableCondition` and `computeDeploymentReplicasAllUpdatedCondition` to calculate the progressing status?

candita: Add a new method called `computeDeploymentReplicasAllUpdatedCondition` which does what you do in:
https://github.com/openshift/cluster-ingress-operator/blob/ec2c9bcada6a46026cc3f5e758b7c03565dc056f/pkg/operator/controller/ingress/status.go#L667-L671
Call it before you call `computeIngressProgressingCondition`, maybe after L76.  Then, instead of passing the `deployment` variable into `computeIngressProgressingCondition`, you can look up the `conditions` to get the information you need to mark the `DeploymentProgressing` condition.

gcs278:
/hold
Still fixing unit tests.

gcs278:
Still need to double check things, but should be mostly completed if anyone wants to review again.

pkg/operator/controller/ingress/status.go
```diff
@@ -775,10 +827,58 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+			}
+			return operatorv1.OperatorCondition{
+				Type:    IngressControllerLoadBalancerProgressingConditionType,
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "ScopeChanged",
+				Message: message,
+			}
+		}
+	}
+	return operatorv1.OperatorCondition{
```
gcs278, line 874: Same comment as above, should we omit this condition if it is false, or better to be verbose?

pkg/operator/controller/ingress/status.go
```diff
@@ -641,43 +691,42 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
-					message = fmt.Sprintf("%[1]s  To effectuate this change, you must delete the service: `oc -n %[2]s delete svc/%[3]s`; the service load-balancer will then be deprovisioned and a new one created.  This will most likely cause the new load-balancer to have a different host name and IP address from the old one's.  Alternatively, you can revert the change to the IngressController: `oc -n openshift-ingress-operator patch ingresscontrollers/%[4]s --type=merge --patch='{\"spec\":{\"endpointPublishingStrategy\":{\"loadBalancer\":{\"scope\":\"%[5]s\"}}}}'", message, service.Namespace, service.Name, ic.Name, haveScope)
-				}
-				condition.Reason = "ScopeChanged"
-				condition.Message = message
-				condition.Status = operatorv1.ConditionTrue
-			}
+
+	// Cover the rare case of no conditions
+	if len(conditions) == 0 {
+		return operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeProgressing, Status: operatorv1.ConditionTrue}
```
gcs278, line 714: I'm 50/50 on whether this should be ConditionTrue or ConditionFalse if all of the conditions don't exist

But if 1 condition exists, and the other is missing, it doesn't account for the the missing condition. Seems a bit arbitrary.

pkg/operator/controller/ingress/status.go
```diff
@@ -72,13 +72,14 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentPodsScheduledCondition(deployment, pods))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
```
Miciah, line 75: Did you mean to delete this line?  E2E still looks for the `DeploymentReplicasAllAvailable` status condition to be set.

pkg/operator/controller/ingress/status.go
```diff
@@ -457,6 +459,55 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
+	}
+	// If have replicas greater than updated replicas, then we are waiting for old replicas to terminate.
+	if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
+		return operatorv1.OperatorCondition{
+			Type:   IngressControllerDeploymentRollingOutConditionType,
+			Status: operatorv1.ConditionTrue,
+			Reason: "DeploymentRollingOut",
+			Message: fmt.Sprintf(
+				"Waiting for router deployment rollout to finish: %d old replicas are pending termination...\n",
+				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
```
candita, line 488: nit
```suggestion
				"Waiting for router deployment rollout to finish: %d old replica/s are pending termination...\n",
				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
```

pkg/operator/controller/ingress/status.go
```diff
@@ -457,6 +459,55 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
+		}
+	}
+	// If available replicas less than updated replicas, then we are waiting for updated replicas to become available.
+	if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
```
candita, line 492: It's fine to return only one condition (instead of appending conditions to a list), but maybe this one should be the second one evaluated, and the "pending termination" one be evaluated last?

pkg/operator/controller/ingress/status.go
candita: The default from the previous incarnation of this method was to return ConditionFalse; that probably shouldn't change.

pkg/operator/controller/ingress/status.go
candita: Better to set it False, since it is valuable information we don't want to lose track of.

pkg/operator/controller/ingress/status.go
candita: Actually, the more I think about this....shouldn't this be considered a Degraded condition, not  a DeploymentRollingOut condition?

pkg/operator/controller/ingress/status.go
```diff
@@ -457,6 +459,55 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
+				"Waiting for router deployment rollout to finish: %d of %d updated replicas are available...\n",
+				deployment.Status.AvailableReplicas, deployment.Status.UpdatedReplicas),
+		}
+	}
+
+	return operatorv1.OperatorCondition{
+		Type:    IngressControllerDeploymentRollingOutConditionType,
+		Status:  operatorv1.ConditionFalse,
+		Reason:  "DeploymentNotRollingOut",
+		Message: "Deployment is not actively rolling out",
```
candita, line 507: Maybe:
```suggestion
		Message: "Deployment completed",
```

gcs278 (commented):
Thanks for another review!

pkg/operator/controller/ingress/status.go
```diff
@@ -775,10 +828,58 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+			}
+		case service == nil:
+			return operatorv1.OperatorCondition{
+				Type:    IngressControllerLoadBalancerProgressingConditionType,
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "NoService",
+				Message: "LoadBalancer Service not created.",
+			}
+		}
+		wantScope := ic.Status.EndpointPublishingStrategy.LoadBalancer.Scope
```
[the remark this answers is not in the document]
gcs278: Whoops, I missed this comment, sorry about that, but good catch. Technically, it doesn't matter because of the `returns` in the previous cases, but should stay consistent.
Fixed.

pkg/operator/controller/ingress/status.go
candita: Does that mean you're going to change it?

gcs278: Whoops sorry, I misinterpreted. I think it's slightly an apples-to-oranges comparison. The previous method did return false by default, but it was at least was attempting to calculate the progressing status.

On the other hand, this logic says "if there are no `IngressControllerLoadBalancerProgressingConditionType` or `IngressControllerDeploymentRollingOutConditionType` status conditions, then say we are progressing". It determines progressing based on if the other conditions exist (which, by all means, should always exist, unless something is wrong).

Like I said, it seems a bit arbitrary, but I suppose we should err on the side of not progressing if the conditions don't exist, so I will change it.

pkg/operator/controller/ingress/status.go
gcs278: @candita Just wanted to recap the individual conditions types that we use for ingress.
This is the "Type: <type>" in the condition message of the IC object. **BOLDED** are status used for cluster operator status `oc get co` and in parentheses is the condition referenced in the calculation:

1. **Available**
    a. DeploymentAvailable (# 7)
    b. DNSReady (# 14)
    c. LoadBalancerReady (# 12)
2. **Progressing**
    a. LoadBalancerProgressing (# 16, NEW: Previously wasn't a separate condition)
    b. DeploymentRollingOut (# 15, NEW: Previously wasn't a separate condition)
3. **Degraded**
    a. PodsScheduled (# 6)
    b. DeploymentAvailable (# 7)
    c. DeploymentReplicasMinAvailable (# 8)
    d. DeploymentReplicasAllAvailable (# 9)
    e. LoadBalancerReady (# 12)
    f. DNSReady (# 14)
4. **Upgradeable**
    a. _Default Cert Check_ (NOT A SEPARATE CONDITION)
    b. _Load Balancer Upgradeable_ (NOT A SEPARATE CONDITION)
5. Admitted
6. PodsScheduled
7. DeploymentAvailable
8. DeploymentReplicasMinAvailable
9. DeploymentReplicasAllAvailable
10. CanaryChecksSucceeding
11. LoadBalancerManaged
12. LoadBalancerReady
13. DNSManaged
14. DNSReady
15. (NEW) DeploymentRollingOut
16. (NEW) LoadBalancerProgressing

Previously, 15 and 16 were stuffed into 3 (`Degraded`) and not individualized conditions. By this, I mean the conditions were not separate and were concatenated together into the `Degraded` condition's Message field, just like `Upgradeable` still is today. Does that help visualize the structure? 

To your point:

> Actually, the more I think about this....shouldn't this be considered a Degraded condition, not a DeploymentRollingOut condition?

"Rolliing Out" says `Progressing` to me. The `Degraded` condition makes sense the way it is calculated today in which it uses `DeploymentAvailable` and the other conditions. Do you disagree with that? If it helps, we can have a call to review ingress status conditions in their entirety

pkg/operator/controller/ingress/status.go
gcs278: Thoughts on using positive or negative verbiage? I think it should reference the fact it is "rolling out" or not.

gcs278:
@candita Let me know when you are available to review/respond to comments again.

pkg/operator/controller/ingress/status.go
candita: Good overview.  This is worth posting to our FAQ.

pkg/operator/controller/ingress/status.go
candita: I agree, the "Deployment has successfully rolled out" message is fine.

pkg/operator/controller/ingress/status.go
```diff
@@ -457,6 +459,55 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
+			Type:   IngressControllerDeploymentRollingOutConditionType,
+			Status: operatorv1.ConditionTrue,
+			Reason: "DeploymentRollingOut",
+			Message: fmt.Sprintf(
+				"Waiting for router deployment rollout to finish: %d old replica(s) are pending termination...\n",
+				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
+		}
+	}
+	// If available replicas less than updated replicas, then we are waiting for updated replicas to become available.
+	if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
```
candita, line 492: Is it possible for AvailableReplicas to change even if the deployment is not actively rolling out?  I.e. if a replica was deleted, do we consider the deployment to actually be rolling out just because the AvailableReplicas might have changed?

pkg/operator/controller/ingress/status.go
```diff
@@ -641,43 +692,42 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+	_, progressingConditions, _ := checkConditions(expected, conditions)
+	if len(progressingConditions) != 0 {
+		progressing := formatConditions(progressingConditions)
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "IngressControllerProgressing",
+			Message: "One or more status conditions indicate progressing: " + progressing,
 		}
 	}
```
candita, line 726:
```suggestion
	if len(conditions) != 0 {
	    _, progressingConditions, _ := checkConditions(expected, conditions)
	    if len(progressingConditions) != 0 {
	    	progressing := formatConditions(progressingConditions)
	  	return operatorv1.OperatorCondition{
			Type:    operatorv1.OperatorStatusTypeProgressing,
			Status:  operatorv1.ConditionTrue,
			Reason:  "IngressControllerProgressing",
			Message: "One or more status conditions indicate progressing: " + progressing,
		}
	    }
	}
```

pkg/operator/controller/ingress/status.go
```diff
@@ -713,9 +763,11 @@ func conditionChanged(a, b operatorv1.OperatorCondition) bool {
 	return a.Status != b.Status || a.Reason != b.Reason || a.Message != b.Message
 }
 
-// computeLoadBalancerStatus returns the complete set of current
-// LoadBalancer-prefixed conditions for the given ingress controller.
+// computeLoadBalancerStatus returns the a set of current
```
candita, line 766:
```suggestion
// computeLoadBalancerStatus returns the set of current
```

pkg/operator/controller/ingress/status.go
```diff
@@ -713,9 +763,11 @@ func conditionChanged(a, b operatorv1.OperatorCondition) bool {
...
+// LoadBalancer-prefixed conditions for the given ingress controller which are
```
candita, line 767:
```suggestion
// LoadBalancer-prefixed conditions for the given ingress controller, which are
```

pkg/operator/controller/ingress/status.go
```diff
@@ -775,10 +828,59 @@ func computeLoadBalancerStatus(ic *operatorv1.IngressController, service *corev1
...
+// the ingress controller's Progressing status.
+func computeLoadBalancerProgressingStatus(ic *operatorv1.IngressController, service *corev1.Service, platform *configv1.PlatformStatus) operatorv1.OperatorCondition {
+	// Compute the IngressControllerLoadBalancerProgressingConditionType condition for the LoadBalancer
+	if ic.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType {
+		switch {
+		case ic.Status.EndpointPublishingStrategy.LoadBalancer == nil:
+			return operatorv1.OperatorCondition{
+				Type:    IngressControllerLoadBalancerProgressingConditionType,
+				Status:  operatorv1.ConditionUnknown,
+				Reason:  "LoadBalancerStatusIncomplete",
```
candita, line 845: I hesitate to change the name of the Reason in case users have monitoring based on a specific name.  If you feel this wouldn't impact existing monitoring, then it can be changed.

pkg/operator/controller/ingress/status.go
candita: In fact, I would go so far as to recommend that we publish some kind of release note that describes the change to status.  I suppose there could be CI, monitoring, or other operators that rely on these statuses being unchanging over time, so we at least need to document the changes.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -1031,6 +1143,103 @@ func TestComputeLoadBalancerStatus(t *testing.T) {
...
+			},
+			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeProgressing, Status: operatorv1.ConditionFalse},
+		},
+		{
+			description: "all progressing unknown",
+			conditions: []operatorv1.OperatorCondition{
+				{Type: IngressControllerLoadBalancerProgressingConditionType, Status: operatorv1.ConditionUnknown},
+				{Type: IngressControllerDeploymentRollingOutConditionType, Status: operatorv1.ConditionUnknown},
+			},
+			expect: operatorv1.OperatorCondition{Type: operatorv1.OperatorStatusTypeProgressing, Status: operatorv1.ConditionTrue},
```
candita, line 1222: I wonder, if both conditions are Unknown, if the Progessing condition should also be Unknown.

pkg/operator/controller/ingress/status.go
```diff
@@ -641,43 +692,42 @@ func formatConditions(conditions []*operatorv1.OperatorCondition) string {
...
+// 1) the Deployment Replicas Updated condition of the IngressController
+// 2) the LoadBalancer Progressing condition of the IngressController
+// The IngressController is judged NOT Progressing only if all 2 conditions are true; otherwise
+// it is considered to be Progressing.
+func computeIngressProgressingCondition(conditions []operatorv1.OperatorCondition) operatorv1.OperatorCondition {
+	expected := []expectedCondition{
+		{
+			condition: IngressControllerLoadBalancerProgressingConditionType,
+			status:    operatorv1.ConditionFalse,
+		},
```
candita, line 706: You might need to add the `ifConditionsTrue` clause when defining `expectedCondition` for `IngressControllerLoadBalancerProgressingConditionType`, as we needed to for `LoadBalancerReadyIngressConditionType` in:
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/status.go#L257

pkg/operator/controller/ingress/status_test.go
```diff
@@ -497,9 +498,120 @@ func TestComputeIngressDegradedCondition(t *testing.T) {
...
 
-// TestComputeIngressProgressCondition verifies that
-// computeIngressProgressingCondition returns the expected status condition.
-func TestComputeIngressProgressCondition(t *testing.T) {
+// TestComputeDeploymentRollingOutCondition verifies that
+// computeDeploymentRollingOutCondition returns the expected status condition.
+func TestComputeDeploymentRollingOutCondition(t *testing.T) {
+	tests := []struct {
+		name                        string
+		conditions                  []operatorv1.OperatorCondition
```
candita, line 506: Looks like `conditions` are no longer used.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -497,9 +498,120 @@ func TestComputeIngressDegradedCondition(t *testing.T) {
...
+		replicasWanted              *int32
+		replicasHave                *int32
+		replicasUpdated             *int32
+		replicasAvailable           *int32
+		expectStatus                operatorv1.ConditionStatus
+		expectMessageContains       string
+		expectMessageDoesNotContain string
```
candita, line 513: `expectMessageDoesNotContain` is also not used.

candita:
@gcs278 sorry for the delay.  It's a great addition, but just a few tweaks needed.

pkg/operator/controller/ingress/status.go
Miciah: More documentation is good.  I'm wary of documenting these in a way that commits us to maintaining backwards compatibility, but a release note and a FAQ would be useful,.  Maybe something along the lines of this to emphasize that these are not formally part of the API: 'To simplify the computation of the "Progressing" status condition, this fix introduces two new status conditions for internal use: "DeploymentRollingOut" and "LoadBalancerProgressing".'

gcs278 (commented):
thanks for the review!

PR #772, opened 2022-06-08
Fix grep command to stop after one match

brandisher:


test/e2e/haproxy_timeouts_test.go
```diff
@@ -141,6 +141,7 @@ func TestHAProxyTimeouts(t *testing.T) {
 		cmd := []string{
 			"grep",
 			"-oP",
+			"-m1",
```
gcs278, line 144: I'm a little conflicted about this. This logic seems not fully deterministic because it's simply grabbing the first `timeout server` from the configuration. In our case, it will always be from the `default` section, which is what we want.

However, the correct solution would be to write a full haproxy.config parser. I'm not convinced we really want to spend a lot of time doing that.

Think you should put  comment in the code explaining the "m1" and the implications? Maybe more specifically, that we are targeting the `default` section and it only works because the default section comes first.

gcs278 (commented):
I think I'm good with this change, but consider a detailed comment for the future.

test/e2e/haproxy_timeouts_test.go
frobware: @rfredette was working on a fix for this. When test `TestAWSELBConnectionIdleTimeout` and `TestHAProxyTimeouts` run simultaneously the latter is likely to fail because these two tests choose `90s` for the `timeout server` value. TestHAProxyTimeouts fails because it tries to match "90s" against "90s\n90s" and fails. So your `-m1` would work, but the test assertion is incomplete and/or just lucky. We could just ensure that these two tests choose values in the millisecond range that are unlikely to conflict (at the time of writing/coding).

PR #775, opened 2022-06-10
Bug 2094962: Fix flakey logic in haproxy timeout tests

rfredette:
Adjust both TestHAProxyTimeouts and TestHAProxyTimeoutsRejection to search for their timeout values explicitly, rather than finding all timeouts of a particular type. In order to make it more likely that the test is finding the correct line, also use more unique timeout values in TestHAProxyTimeouts

Also fix the output of stderr in these tests when they fail

### DIFF
--- a/test/e2e/haproxy_timeouts_test.go
+++ b/test/e2e/haproxy_timeouts_test.go
@@ -25,17 +25,19 @@ import (
 
 func TestHAProxyTimeouts(t *testing.T) {
 	t.Parallel()
+	// Use timeout values that are more likely to be unique across the ENTIRE
+	// test suite
 	const (
-		clientTimeoutInput     = 45 * time.Second
-		clientTimeoutOutput    = "45s"
+		clientTimeoutInput     = 47 * time.Second
+		clientTimeoutOutput    = "47s"
 		clientFinTimeoutInput  = 1500 * time.Millisecond
 		clientFinTimeoutOutput = "1500ms"
-		serverTimeoutInput     = 90 * time.Second
-		serverTimeoutOutput    = "90s"
-		serverFinTimeoutInput  = 5 * time.Second
-		serverFinTimeoutOutput = "5s"
-		tunnelTimeoutInput     = 90 * time.Minute
-		tunnelTimeoutOutput    = "90m"
+		serverTimeoutInput     = 92 * time.Second
+		serverTimeoutOutput    = "92s"
+		serverFinTimeoutInput  = 7 * time.Second
+		serverFinTimeoutOutput = "7s"
+		tunnelTimeoutInput     = 93 * time.Minute
+		tunnelTimeoutOutput    = "93m"
 		tlsInspectDelayInput   = 720 * time.Hour
 		tlsInspectDelayOutput  = "2147483647ms" // 720h is greater than the maximum timeout, so it should get clipped to max
 	)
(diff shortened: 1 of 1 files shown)


test/e2e/haproxy_timeouts_test.go
```diff
@@ -26,16 +26,16 @@ import (
 func TestHAProxyTimeouts(t *testing.T) {
 	t.Parallel()
 	const (
-		clientTimeoutInput     = 45 * time.Second
-		clientTimeoutOutput    = "45s"
+		clientTimeoutInput     = 47 * time.Second
```
gcs278, line 29: My only comment is that, this issue resolution isn't super obvious by reading the code (magic numbers). i.e. a future developer may "clean things up" and set all the timeouts to even numbers.

Do you think you could add a comment explaining that these numbers are specifically different for a reason and to be careful when changing them?

rfredette: That's fair. I'll add a comment to help explain why these numbers were used

test/e2e/haproxy_timeouts_test.go
```diff
@@ -141,12 +142,12 @@ func TestHAProxyTimeouts(t *testing.T) {
...
 			"grep",
 			"-oP",
-			"timeout\\s+" + timeout.Name + "\\s*\\K[0-9]+(?:us|ms|s|m|h|d)",
+			"timeout\\s+" + timeout.Name + "\\s*\\K" + timeout.Value,
 			"/var/lib/haproxy/conf/haproxy.config",
 		}
 		if err := podExec(t, routerPod, &stdout, &stderr, cmd); err != nil {
 			t.Errorf("Error executing %s: %v", strings.Join(cmd, " "), err)
-			t.Errorf("stderr: %v", stderr)
+			t.Errorf("stderr: %v", stderr.Bytes())
```
frobware, line 150: The format needs to be %s, otherwise on error you get:

```console
=== CONT  TestHAProxyTimeouts
    haproxy_timeouts_test.go:149: Error executing grep-x -oP timeout\s+client\s*\K47s /var/lib/haproxy/conf/haproxy.config: command terminated with exit code 255
    haproxy_timeouts_test.go:150: stderr: [116 105 109 101 61 34 50 48 50 50 45 48 54 45 49 52 84 49 49 58 50 50 58 49 49 90 34 32 108 101 118 101 108 61 101 114
```

test/e2e/haproxy_timeouts_test.go
```diff
@@ -141,12 +142,12 @@ func TestHAProxyTimeouts(t *testing.T) {
 		cmd := []string{
 			"grep",
 			"-oP",
-			"timeout\\s+" + timeout.Name + "\\s*\\K[0-9]+(?:us|ms|s|m|h|d)",
+			"timeout\\s+" + timeout.Name + "\\s*\\K" + timeout.Value,
```
frobware, line 145: Given that we are changing the RE in this commit the second whitespace (i.e., \s*) should be \s+.

test/e2e/haproxy_timeouts_test.go
```diff
@@ -279,12 +280,12 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
```
frobware, line 283: Given that we are changing the RE in this commit the second whitespace (i.e., \s*) should be \s+.

frobware (changes requested):

PR #781, opened 2022-06-15
NE-568: Add dnsManagementPolicy option to ingresscontrollers

thejasn:
Add `dnsManagementPolicy` option to ingresscontrollers
---
- `pkg/operator/controller/dns/controller.go` 
    - Create DNS provider based on `dnsManagementPolicy` on `DNSRecord` CR.
    - Add sanitize function for purging deprecated fields on DNS record status.
    - Refactor `publishRecordToZones()`
- `pkg/operator/controller/ingress/controller.go` - Reconcile changes made to `dnsManagementPolicy` on ingress spec.
- `pkg/operator/controller/ingress/dns.go` 
    - Create desired `DNSRecord` with `dnsManagementPolicy` derived from ingresscontroller status.
    - Converge domain mismatch check with `UmanagedDNS` policy. 
- `pkg/operator/controller/ingress/status.go` - Update status of types `DNSManaged` and `DNSReady`.
- Renamed `test/e2e/util.go` -> `test/e2e/util_test.go` to allow `gopls` to index variables(`kclient`) declared under other test files.

EP: https://github.com/openshift/enhancements/pull/1114
CFE: [CFE-339](https://issues.redhat.com/browse/CFE-339), [CFE-342](https://issues.redhat.com/browse/CFE-342)

---
Signed-off-by: thejasn <thn@redhat.com>

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -25,7 +25,7 @@ import (
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
-	operatorutil "github.com/openshift/cluster-ingress-operator/pkg/util"
+	oputil "github.com/openshift/cluster-ingress-operator/pkg/util"
 	awsutil "github.com/openshift/cluster-ingress-operator/pkg/util/aws"
 	"github.com/openshift/cluster-ingress-operator/pkg/util/slice"
 
@@ -39,7 +39,6 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
-	operatoringressv1 "github.com/openshift/api/operatoringress/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -132,10 +131,6 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, fmt.Errorf("failed to get dns 'cluster': %v", err)
 	}
 
-	if err := r.createDNSProviderIfNeeded(dnsConfig); err != nil {
-		return reconcile.Result{}, err
-	}
-
 	record := &iov1.DNSRecord{}
 	if err := r.client.Get(ctx, request.NamespacedName, record); err != nil {
 		if errors.IsNotFound(err) {
@@ -144,6 +139,16 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 		log.Error(err, "failed to get dnsrecord; will retry", "dnsrecord", request.NamespacedName)
 		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
+	}
+
+	if updated, err := r.migrateRecordStatusConditions(record); err != nil {
+		return reconcile.Result{}, err
+	} else if updated {
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	if err := r.createDNSProviderIfNeeded(dnsConfig, record); err != nil {
+		return reconcile.Result{}, err
 	}
 
 	// If the DNS record was deleted, clean up and return.
@@ -190,7 +195,14 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	if dnsConfig.Spec.PublicZone != nil {
 		zones = append(zones, *dnsConfig.Spec.PublicZone)
 	}
-	statuses, result := r.publishRecordToZones(zones, record)
+	requeue, statuses := r.publishRecordToZones(zones, record)
+
+	// Requeue if publishing records failed.
+	result := reconcile.Result{}
+	if requeue {
+		result.RequeueAfter = 30 * time.Second
+	}
+
 	if !dnsZoneStatusSlicesEqual(statuses, record.Status.Zones) {
 		updated := record.DeepCopy()
 		updated.Status.Zones = statuses
@@ -202,6 +214,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 			log.Info("updated dnsrecord", "dnsrecord", updated)
 		}
 	}
+
 	return result, nil
 }
 
@@ -210,8 +223,12 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 // changed since the current provider was created.  After creating a new
 // provider, createDNSProviderIfNeeded updates the reconciler state
 // with the new provider and current platform status and cloud credentials.
-func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
+func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS, record *iov1.DNSRecord) error {
 	var needUpdate bool
+
+	if record.Spec.DNSManagementPolicy == iov1.UnmanagedDNS {
+		return nil
+	}
 
 	infraConfig := &configv1.Infrastructure{}
 	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
@@ -256,61 +273,105 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
 	return nil
 }
 
-func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1.DNSRecord) ([]iov1.DNSZoneStatus, reconcile.Result) {
-	result := reconcile.Result{}
+// replacePublishedRecord replaces a previously published record with the given record,
+// and the result is returned as a condition. Upon errors during publishing,
+// an error object is returned.
+func (r *reconciler) replacePublishedRecord(zone configv1.DNSZone, record *iov1.DNSRecord) (iov1.DNSZoneCondition, error) {
+	condition := iov1.DNSZoneCondition{
+		Status:             string(operatorv1.ConditionUnknown),
+		Type:               iov1.DNSRecordPublishedConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+
+	err := r.dnsProvider.Replace(record, zone)
+	if err != nil {
+		log.Error(err, "failed to replace DNS record in zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionFalse)
+		condition.Reason = "ProviderError"
+		condition.Message = fmt.Sprintf("The DNS provider failed to replace the record: %v", err)
+	} else {
+		log.Info("replaced DNS record in zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionTrue)
+		condition.Reason = "ProviderSuccess"
+		condition.Message = "The DNS provider succeeded in replacing the record"
+	}
+
+	return condition, err
+}
+
+// publishRecord ensures the given record is published to the provided zone
+// and the result is returned as a condition. Upon errors during publishing
+// an error object is returned.
+func (r *reconciler) publishRecord(zone configv1.DNSZone, record *iov1.DNSRecord) (iov1.DNSZoneCondition, error) {
+	condition := iov1.DNSZoneCondition{
+		Status:             string(operatorv1.ConditionUnknown),
+		Type:               iov1.DNSRecordPublishedConditionType,
+		LastTransitionTime: metav1.Now(),
+	}
+
+	err := r.dnsProvider.Ensure(record, zone)
+	if err != nil {
+		log.Error(err, "failed to publish DNS record to zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionFalse)
+		condition.Reason = "ProviderError"
+		condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
+	} else {
+		log.Info("published DNS record to zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionTrue)
+		condition.Reason = "ProviderSuccess"
+		condition.Message = "The DNS provider succeeded in ensuring the record"
+	}
+
+	return condition, err
+}
+
+// publishRecordToZones attempts to publish records and returns a bool
+// indicating if we need to requeue due to errors and list of latest DNS Zone status.
+func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1.DNSRecord) (bool, []iov1.DNSZoneStatus) {
 	var statuses []iov1.DNSZoneStatus
+	var requeue bool
+	dnsPolicy := record.Spec.DNSManagementPolicy
 	for i := range zones {
-		zone := zones[i]
+		isRecordPublished := recordIsAlreadyPublishedToZone(record, &zones[i])
 
 		// Only publish the record if the DNSRecord has been modified
 		// (which would mean the target could have changed) or its
 		// status does not indicate that it has already been published.
-		if record.Generation == record.Status.ObservedGeneration && recordIsAlreadyPublishedToZone(record, &zone) {
-			log.Info("skipping zone to which the DNS record is already published", "record", record.Spec, "dnszone", zone)
+		if record.Generation == record.Status.ObservedGeneration && isRecordPublished {
+			log.Info("skipping zone to which the DNS record is already published", "record", record.Spec, "dnszone", zones[i])
 			continue
 		}
 
-		condition := iov1.DNSZoneCondition{
-			Status:             string(operatorv1.ConditionUnknown),
-			Type:               iov1.DNSRecordFailedConditionType,
-			LastTransitionTime: metav1.Now(),
-		}
-
-		if recordIsAlreadyPublishedToZone(record, &zone) {
-			log.Info("replacing DNS record", "record", record.Spec, "dnszone", zone)
-
-			if err := r.dnsProvider.Replace(record, zone); err != nil {
-				log.Error(err, "failed to replace DNS record in zone", "record", record.Spec, "dnszone", zone)
-				condition.Status = string(operatorv1.ConditionTrue)
-				condition.Reason = "ProviderError"
-				condition.Message = fmt.Sprintf("The DNS provider failed to replace the record: %v", err)
-				result.RequeueAfter = 30 * time.Second
-			} else {
-				log.Info("replaced DNS record in zone", "record", record.Spec, "dnszone", zone)
-				condition.Status = string(operatorv1.ConditionFalse)
-				condition.Reason = "ProviderSuccess"
-				condition.Message = "The DNS provider succeeded in replacing the record"
+		var err error
+		var condition iov1.DNSZoneCondition
+		if dnsPolicy == iov1.UnmanagedDNS {
+			log.Info("DNS record not published", "record", record.Spec)
+			condition = iov1.DNSZoneCondition{
+				Message:            "DNS record is currently not being managed by the operator",
+				Reason:             "UnmanagedDNS",
+				Status:             string(operatorv1.ConditionUnknown),
+				Type:               iov1.DNSRecordPublishedConditionType,
+				LastTransitionTime: metav1.Now(),
 			}
+		} else if isRecordPublished {
+			condition, err = r.replacePublishedRecord(zones[i], record)
 		} else {
-			if err := r.dnsProvider.Ensure(record, zone); err != nil {
-				log.Error(err, "failed to publish DNS record to zone", "record", record.Spec, "dnszone", zone)
-				condition.Status = string(operatorv1.ConditionTrue)
-				condition.Reason = "ProviderError"
-				condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
-				result.RequeueAfter = 30 * time.Second
-			} else {
-				log.Info("published DNS record to zone", "record", record.Spec, "dnszone", zone)
-				condition.Status = string(operatorv1.ConditionFalse)
-				condition.Reason = "ProviderSuccess"
-				condition.Message = "The DNS provider succeeded in ensuring the record"
-			}
+			condition, err = r.publishRecord(zones[i], record)
 		}
+
+		// Check if replacing or publishing record resulted in an error.
+		// If it did, re-enqueue the request for processing.
+		if err != nil {
+			requeue = true
+		}
+
 		statuses = append(statuses, iov1.DNSZoneStatus{
-			DNSZone:    zone,
+			DNSZone:    zones[i],
 			Conditions: []iov1.DNSZoneCondition{condition},
 		})
 	}
-	return mergeStatuses(zones, record.Status.DeepCopy().Zones, statuses), result
+
+	return requeue, mergeStatuses(zones, record.Status.DeepCopy().Zones, statuses)
 }
 
 // recordIsAlreadyPublishedToZone returns a Boolean value indicating whether the
@@ -323,8 +384,8 @@ func recordIsAlreadyPublishedToZone(record *iov1.DNSRecord, zoneToPublish *confi
 		}
 
 		for _, condition := range zoneInStatus.Conditions {
-			if condition.Type == iov1.DNSRecordFailedConditionType {
-				return condition.Status == string(operatorv1.ConditionFalse)
+			if condition.Type == iov1.DNSRecordPublishedConditionType {
+				return condition.Status == string(operatorv1.ConditionTrue)
 			}
 		}
 	}
@@ -415,6 +476,75 @@ func conditionChanged(a, b iov1.DNSZoneCondition) bool {
 	return a.Status != b.Status || a.Reason != b.Reason || a.Message != b.Message
 }
 
+// migrateRecordStatusConditions purges deprecated fields from DNS Record
+// status. Removes "Failed" (DNSRecordFailedConditionType) and replaces with
+// "Published" (DNSRecordPublishedConditionType). If the status needs to be
+// updated, this function performs the update.  The function returns a Boolean
+// value indicating whether it has performed an update along with the error
+// value from any update it has performed.
+//
+// This function is at least required for 1 minor release and can be removed after
+// Openshift 4.13 is released.
+func (r *reconciler) migrateRecordStatusConditions(record *iov1.DNSRecord) (bool, error) {
+	var updateConditions bool
+	for i := range record.Status.Zones {
+		zone := record.Status.DeepCopy().Zones[i]
+		updateZone, cond := migrateRecordStatusCondition(zone.Conditions)
+		if updateZone {
+			updateConditions = true
+			record.Status.Zones[i].Conditions = cond
+		}
+	}
+	if updateConditions {
+		if err := r.client.Status().Update(context.TODO(), record); err != nil {
+			return false, fmt.Errorf("failed to update dnsrecord status: %w", err)
+		}
+		log.Info("dnsrecord status migrated", "dnsrecord", record)
+		return true, nil
+	}
+	return false, nil
+}
+
+func migrateRecordStatusCondition(conditions []iov1.DNSZoneCondition) (bool, []iov1.DNSZoneCondition) {
+	var result []iov1.DNSZoneCondition
+	var updatedCondition iov1.DNSZoneCondition
+	var updated bool
+	now := metav1.NewTime(clock.Now())
+	for _, cond := range conditions {
+		if cond.Type != iov1.DNSRecordFailedConditionType {
+			updatedCondition = cond
+		} else {
+			updated = true
+			switch cond.Status {
+			case string(operatorv1.ConditionTrue):
+				updatedCondition = iov1.DNSZoneCondition{
+					Type:               iov1.DNSRecordPublishedConditionType,
+					Status:             string(operatorv1.ConditionFalse),
+					LastTransitionTime: now,
+					Reason:             cond.Reason,
+					Message:            cond.Message,
+				}
+			case string(operatorv1.ConditionFalse):
+				updatedCondition = iov1.DNSZoneCondition{
+					Type:               iov1.DNSRecordPublishedConditionType,
+					Status:             string(operatorv1.ConditionTrue),
+					LastTransitionTime: now,
+					Reason:             cond.Reason,
+					Message:            cond.Message,
+				}
+			default:
+				// We don't need to map Failed=Unknown condition
+				// since the operator never set this condition to Unknown.
+				// Since no version of the operator had this condition, we
+				// don't need to migrate it.
+				continue
+			}
+		}
+		result = mergeConditions(result, []iov1.DNSZoneCondition{updatedCondition})
+	}
+	return updated, result
+}
+
 // dnsZoneStatusSlicesEqual compares two DNSZoneStatus slice values.  Returns
 // true if the provided values should be considered equal for the purpose of
 // determining whether an update is necessary, false otherwise.  The comparison
@@ -437,7 +567,7 @@ func dnsZoneStatusSlicesEqual(a, b []iov1.DNSZoneStatus) bool {
 // ToDNSRecords returns reconciliation requests for all DNSRecords.
 func (r *reconciler) ToDNSRecords(o client.Object) []reconcile.Request {
 	var requests []reconcile.Request
-	records := &operatoringressv1.DNSRecordList{}
+	records := &iov1.DNSRecordList{}
 	if err := r.cache.List(context.Background(), records, client.InNamespace(r.config.Namespace)); err != nil {
 		log.Error(err, "failed to list dnsrecords", "related", o.GetSelfLink())
 		return requests
@@ -499,32 +629,32 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 					break
 				case ep.Name == awsdns.Route53Service:
 					route53Found = true
-					scheme, err := operatorutil.URI(ep.URL)
+					scheme, err := oputil.URI(ep.URL)
 					if err != nil {
 						return nil, fmt.Errorf("failed to validate URI %s: %v", ep.URL, err)
 					}
-					if scheme != operatorutil.SchemeHTTPS {
-						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, operatorutil.SchemeHTTPS)
+					if scheme != oputil.SchemeHTTPS {
+						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, oputil.SchemeHTTPS)
 					}
 					cfg.ServiceEndpoints = append(cfg.ServiceEndpoints, awsdns.ServiceEndpoint{Name: ep.Name, URL: ep.URL})
 				case ep.Name == awsdns.ELBService:
 					elbFound = true
-					scheme, err := operatorutil.URI(ep.URL)
+					scheme, err := oputil.URI(ep.URL)
 					if err != nil {
 						return nil, fmt.Errorf("failed to validate URI %s: %v", ep.URL, err)
 					}
-					if scheme != operatorutil.SchemeHTTPS {
-						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, operatorutil.SchemeHTTPS)
+					if scheme != oputil.SchemeHTTPS {
+						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, oputil.SchemeHTTPS)
 					}
 					cfg.ServiceEndpoints = append(cfg.ServiceEndpoints, awsdns.ServiceEndpoint{Name: ep.Name, URL: ep.URL})
 				case ep.Name == awsdns.TaggingService:
 					tagFound = true
-					scheme, err := operatorutil.URI(ep.URL)
+					scheme, err := oputil.URI(ep.URL)
 					if err != nil {
 						return nil, fmt.Errorf("failed to validate URI %s: %v", ep.URL, err)
 					}
-					if scheme != operatorutil.SchemeHTTPS {
-						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, operatorutil.SchemeHTTPS)
+					if scheme != oputil.SchemeHTTPS {
+						return nil, fmt.Errorf("invalid scheme for URI %s; must be %s", ep.URL, oputil.SchemeHTTPS)
 					}
 					cfg.ServiceEndpoints = append(cfg.ServiceEndpoints, awsdns.ServiceEndpoint{Name: ep.Name, URL: ep.URL})
 				}
@@ -591,7 +721,7 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 			return &dns.FakeProvider{}, nil
 		}
 	case configv1.PowerVSPlatformType:
-		//Power VS platform will use the ibm dns implementation
+		// Power VS platform will use the ibm dns implementation
 		var err error
 		if platformStatus.PowerVS.CISInstanceCRN != "" {
 			dnsProvider, err = getIbmDNSProvider(dnsConfig, creds, platformStatus.PowerVS.CISInstanceCRN, userAgent, true)
--- a/pkg/operator/controller/dns/controller_test.go
+++ b/pkg/operator/controller/dns/controller_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
@@ -15,17 +16,28 @@ import (
 )
 
 func TestPublishRecordToZones(t *testing.T) {
-	var tests = []struct {
-		name   string
-		zones  []configv1.DNSZone
-		expect []string
+	tests := []struct {
+		name         string
+		zones        []configv1.DNSZone
+		unManagedDNS bool
+		expect       []iov1.DNSZoneStatus
 	}{
 		{
 			name: "testing for a successful update of 1 record",
-			zones: []configv1.DNSZone{{
-				ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+			zones: []configv1.DNSZone{
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
 			},
-			expect: []string{string(operatorv1.ConditionFalse)},
+			expect: []iov1.DNSZoneStatus{{
+				DNSZone: configv1.DNSZone{
+					ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io",
+				},
+				Conditions: []iov1.DNSZoneCondition{
+					{
+						Type:   "Published",
+						Status: "True",
+					},
+				},
+			}},
 		},
 		{
 			name:   "when no zones available",
@@ -36,32 +48,80 @@ func TestPublishRecordToZones(t *testing.T) {
 			name: "when one zone available and one not available",
 			zones: []configv1.DNSZone{
 				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
-				{ID: ""}},
-			expect: []string{string(operatorv1.ConditionFalse), string(operatorv1.ConditionFalse)},
+				{ID: ""},
+			},
+			expect: []iov1.DNSZoneStatus{
+				{
+					DNSZone: configv1.DNSZone{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+					Conditions: []iov1.DNSZoneCondition{
+						{
+							Type:   "Published",
+							Status: "True",
+						},
+					},
+				},
+				{
+					DNSZone: configv1.DNSZone{ID: ""},
+					Conditions: []iov1.DNSZoneCondition{
+						{
+							Type:   "Published",
+							Status: "True",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "when one zone available and one not available and unmanaged DNS policy",
+			zones: []configv1.DNSZone{
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+				{ID: ""},
+			},
+			unManagedDNS: true,
+			expect: []iov1.DNSZoneStatus{
+				{
+					DNSZone: configv1.DNSZone{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
+					Conditions: []iov1.DNSZoneCondition{
+						{
+							Type:   "Published",
+							Status: "Unknown",
+						},
+					},
+				},
+				{
+					DNSZone: configv1.DNSZone{ID: ""},
+					Conditions: []iov1.DNSZoneCondition{
+						{
+							Type:   "Published",
+							Status: "Unknown",
+						},
+					},
+				},
+			},
 		},
 	}
 
 	for _, test := range tests {
 		record := &iov1.DNSRecord{
 			Spec: iov1.DNSRecordSpec{
-				DNSName:    "subdomain.dnszone.io.",
-				RecordType: iov1.ARecordType,
-				Targets:    []string{"55.11.22.33"},
+				DNSName:             "subdomain.dnszone.io.",
+				RecordType:          iov1.ARecordType,
+				DNSManagementPolicy: iov1.ManagedDNS,
+				Targets:             []string{"55.11.22.33"},
 			},
 		}
+		if test.unManagedDNS {
+			record.Spec.DNSManagementPolicy = iov1.UnmanagedDNS
+		}
 		r := &reconciler{
-			//TODO To write a fake provider that can return errors and add more test cases.
+			// TODO To write a fake provider that can return errors and add more test cases.
 			dnsProvider: &dns.FakeProvider{},
 		}
-		actual, _ := r.publishRecordToZones(test.zones, record)
-		var conditions []string
-		for _, dnsStatus := range actual {
-			for _, condition := range dnsStatus.Conditions {
-				conditions = append(conditions, condition.Status)
-			}
-		}
-		if !cmp.Equal(conditions, test.expect) {
-			t.Fatalf("%q: expected:\n%#v\ngot:\n%#v", test.name, test.expect, conditions)
+
+		_, actual := r.publishRecordToZones(test.zones, record)
+		opts := cmpopts.IgnoreFields(iov1.DNSZoneCondition{}, "Reason", "Message", "LastTransitionTime")
+		if !cmp.Equal(actual, test.expect, opts) {
+			t.Fatalf("%q: found diff between actual and expected:\n%s", test.name, cmp.Diff(actual, test.expect, opts))
 		}
 	}
 }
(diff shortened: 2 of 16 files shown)


go.mod
```diff
@@ -125,3 +125,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/openshift/api => github.com/thejasn/api v0.0.0-20220718063518-93fe28a445b0
```
candita, line 129: This needs to change prior to approval.

pkg/operator/controller/ingress/dns.go
```diff
@@ -34,7 +34,7 @@ func (r *reconciler) ensureWildcardDNSRecord(ic *operatorv1.IngressController, p
 		return false, nil, nil
 	}
 
-	if !manageDNSForDomain(ic.Status.Domain, platformStatus, dnsConfig) {
+	if !manageDNSForDomain(ic.Status.Domain, platformStatus, dnsConfig) && ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy != operatorv1.UnmanagedLoadBalancerDNS {
```
rfredette, line 37: Shouldn't this be an or? If the LB's DNS policy is managed but the domain is unmanaged, I'd expect the operator not to create a wildcard record (and vice versa)

thejasn: So there is a variation in the implementation of this and #761. The new added field controls DNS record management on the said cloud provider only i.e. the `DNSRecord` CR is still created in an `Unmanaged` state. The CR is created so that user/admin can use the information (`target`) present in it to create the required DNS records manually. ([REF](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/configurable-dns-loadbalancerservice.md#implementation-detailsnotesconstraints))

In PR #761, the logic that was implemented was that in addition to skipping creation of DNS records on the cloud provider it also skipped the creation of the `DNSRecord` CR. 

It was decided we could converge the 2 implementations and settle with the current solution, where the `DNSRecord` CR is created in an `Unmanaged` state. I'll be raising the needed PRs post getting this merged.

pkg/operator/controller/ingress/status.go
```diff
@@ -822,25 +819,37 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 			},
 		}
 	}
-
-	if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) {
-		return []operatorv1.OperatorCondition{
+	var conditions []operatorv1.OperatorCondition
+	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS ||
+		(ic.Spec.EndpointPublishingStrategy != nil && ic.Spec.EndpointPublishingStrategy.LoadBalancer != nil &&
+			ic.Spec.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS &&
```
gcs278, line 825: What situation would `ic.Spec` be important when `ic.Status` don't have it set? Is this just being extra cautious?

pkg/operator/controller/ingress/status.go
```diff
@@ -822,25 +819,37 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
+			ic.Spec.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy != ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy) {
```
gcs278, line 826: Does this extra `ic.Spec.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy != ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy` provide value?

Isn't this always true because you just checked `ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS`?

test/e2e/http_header_buffer_test.go
```diff
@@ -115,7 +115,7 @@ func TestHTTPHeaderBufferSize(t *testing.T) {
 		"-H",
 		headerString,
 	}
-
+	extraCurlArgs = append(extraCurlArgs, "--resolve", echoRoute.Spec.Host+":80:"+service.Spec.ClusterIP)
```
gcs278, line 118: Not 100% sure how this is related, not that I care too much. But I see you are manually resolving the IP. Why exactly? Is this related to this PR?

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	}()
+
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s", address), nil)
+	if err != nil {
+		t.Fatalf("failed to build client request: %v", err)
+	}
+	req.Host = hostname
+
+	httpClient := http.Client{Timeout: 5 * time.Second}
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
```
gcs278, line 287: Why does `verifyUnmanagedIngressController` build a pod, service, route and try to curl it? I guess it's not super clear to me how this tests DNS.

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if len(wildcardRecord.Status.Zones) != 2 {
+		t.Fatalf("DNSRecord %s expected allocated dnsZones but found none", wildcardRecordName.Name)
+	}
+
+	verifyUnmanagedIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
+}
+
+func verifyUnmanagedIngressController(t *testing.T, name types.NamespacedName, hostname, address string) {
```
gcs278, line 249: Is this really testing UnmanagedDNS? Would it make sense to add `DNS` in the function name to just be a little clearer?

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	httpClient := http.Client{Timeout: 5 * time.Second}
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
+		if r.StatusCode == http.StatusOK {
+			t.Logf("verified connectivitiy with workload with req %v and response %v", req.URL, r.StatusCode)
+			return true
+		}
+		return false
+	})
+	if err != nil {
+		t.Fatalf("failed verify condition with external client: %v", err)
```
gcs278, line 295: If this is really testing DNS, does it make sense for this error message be more specific? Or generally, could "verify condition" be more specific to help debugging?

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	wildcardRecord := &iov1.DNSRecord{}
+	if err := kclient.Get(context.TODO(), wildcardRecordName, wildcardRecord); err != nil {
+		t.Fatalf("failed to get wildcard dnsrecord %s: %v", wildcardRecordName, err)
+	}
+
+	if wildcardRecord.Spec.DNSManagementPolicy != iov1.ManagedDNS {
+		t.Fatalf("DNSRecord %s expected in ManagedDNS state but got %s", wildcardRecordName.Name, wildcardRecord.Spec.DNSManagementPolicy)
+	}
+
+	verifyUnmanagedIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
```
gcs278, line 137: (see comments below about `verifyUnmanagedIngressController`). This confuses me a bit. This is a DNS managed ingress controller, but we are calling a function called `verifyUnmanagedIngressController`.

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if err := kclient.Get(context.TODO(), wildcardRecordName, wildcardRecord); err != nil {
+		t.Fatalf("failed to get wildcard dnsrecord %s: %v", wildcardRecordName, err)
+	}
+
+	if wildcardRecord.Spec.DNSManagementPolicy != iov1.UnmanagedDNS {
+		t.Fatalf("DNSRecord %s expected in UnmanagedDNS state but got %s", wildcardRecordName.Name, wildcardRecord.Spec.DNSManagementPolicy)
+	}
+
```
gcs278, line 161: So the enhancement specificially stated:

> - Test the following update paths on a custom ingresscontroller
>   - Updating `.loadbalancer.dnsManagementPolicy` from `Managed` -> `Unmanaged` -> `Managed` and
>     to ensure no conflicts during creation/recreation of DNS record on the cloud provider.

I don't think that whole test (Managed -> Unmanaged -> Managed) has exactly been accomplished yet. I'm hesitant to hold you to the exact words, but is it worth going back to Managed one last time here?

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	verifyUnmanagedIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
+
+	t.Logf("Updating ingresscontroller %s to Managed", ic.Name)
```
gcs278, line 69: *nit* could you specify more specifically "DNS" is managed? "managed" is a little vague. This also applies to a couple other log statements.

gcs278 (commented):
Overall great looking review. I like your formatting updates.

test/e2e/unmanaged_dns_test.go
thejasn: So the route that is created is configured to use the created ingresscontroller's domain. The curl tries to ensure proper connectivity to the test workload (echo pod) and this also ensures functionality of the router that was created.

test/e2e/unmanaged_dns_test.go
thejasn: Thanks, good catch. This function evolved too much during implementation. I'll rename both the functions (add a godoc as well), 
```
verifyUnmanagedIngressController => verifyExternalIngressController
verifyUnmanagedInternalIngressController => verifyInternalIngressController
```
WDYT?

test/e2e/unmanaged_dns_test.go
thejasn: So my initial thought was to add this transition and along with it manually delete the `DNS` records that are created on the cloud provider. But this would require a lot of additional test code to delete the records across cloud providers with very little returns (thoughts?). So without this logic, the last transition from `Unmanaged` -> `Transition` doesn't really make sense. 

So I'm thinking of just updating the EP w.r.t testing strategy along with the follow up change that is discussed [here](https://github.com/openshift/cluster-ingress-operator/pull/781#discussion_r929852173).

pkg/operator/controller/ingress/status.go
thejasn: So this is just me being over cautious in the scenario where `dnsManagementPolicy` isn't set on the spec (relying on the defaults, which later gets synced to `ic.Status`) and we migrate to `Unmanaged` by editing the `spec`. (if `spec` didn't sync to `status`).

I've removed it, on hindsight this makes it less readable.

test/e2e/http_header_buffer_test.go
thejasn: So the `--resolve` option was added by default in the func `buildCurlPod()` and was being mapped to `address` (specifically IP addresses). It **does not** allow CNAME to be provided as the destination address.

For our use-case I'm using `--resolve` to map a dummy domain to the target associated with the ingresscontroller (on AWS this is a CNAME). So I've moved all the usages of `--resolve` out of the function.

pkg/operator/controller/ingress/status.go
gcs278: Yea I don't have strong feelings either way. If `ic.Status` isn't eventually getting updated, then I suppose something went wrong, and DNS could be considered still managed. There might be more nuance to these situations than I am aware up. 

But I'm good with just using `ic.Status`.

test/e2e/http_header_buffer_test.go
gcs278: Ah okay I see it now. Thanks for explaining.

test/e2e/unmanaged_dns_test.go
gcs278: I think that's much clearer

test/e2e/unmanaged_dns_test.go
gcs278: That's a fair point. I think it would be good to keep the EP aligned with we are actually doing.

test/e2e/unmanaged_dns_test.go
gcs278: That makes sense and is valid. Did you consider using just a `dnslookup` instead of HTTP client request? An entire http request makes sense, but seems to be quite broad. Just curious, not necessarily recommending a change.

My other comment is, now that I understand these functions, would it make more sense to put them in the util_test.go? Because these are really quite generic IngressController tests that could be reused.

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+
+	t.Logf("Updating ingresscontroller %s to dnsManagementPolicy=Managed", ic.Name)
+
+	if err := updateIngressControllerSpecWithRetryOnConflict(t, name, 5*time.Minute, func(ics *operatorv1.IngressControllerSpec) {
+		ics.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.ManagedLoadBalancerDNS
+	}); err != nil {
+		t.Fatalf("failed to update ingresscontroller %s: %v", name, err)
+	}
+
+	t.Log("Waiting for stable conditions on ingresscontroller after Managed")
```
gcs278, line 77: *nit* I like your update with `dnsManagementPolicy=` Could you do that with the other logs messages?

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+	wildcardRecord := &iov1.DNSRecord{}
+	if err := kclient.Get(context.TODO(), wildcardRecordName, wildcardRecord); err != nil {
+		t.Fatalf("failed to get wildcard dnsrecord %s: %v", wildcardRecordName, err)
+	}
+
+	if wildcardRecord.Spec.DNSManagementPolicy != iov1.UnmanagedDNS {
+		t.Fatalf("DNSRecord %s expected in UnmanagedDNS state but got %s", wildcardRecordName.Name, wildcardRecord.Spec.DNSManagementPolicy)
+	}
+
+	verifyExternalIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
```
gcs278, line 67: I'm a little confused, how is this successful? You just created an IC with `DNSManagementPolicy: Unmanaged`, so it shouldn't create a DNSRecord, right?

I suppose a DNSRecord gets created regardless even when starting out with `Unmanaged`? Is that what this means (from enhancement):
> ### Non-Goals
> - Provide a day-0 solution for cluster installations involving external DNS
  management by the customer.

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,380 @@
...
+
+	if err := kclient.Get(context.TODO(), wildcardRecordName, wildcardRecord); err != nil {
+		t.Fatalf("failed to get wildcard dnsrecord %s: %v", wildcardRecordName, err)
+	}
+
+	if wildcardRecord.Spec.DNSManagementPolicy != iov1.UnmanagedDNS {
+		t.Fatalf("DNSRecord %s expected in UnmanagedDNS state but got %s", wildcardRecordName.Name, wildcardRecord.Spec.DNSManagementPolicy)
+	}
+
+	verifyExternalIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
```
gcs278, line 162: I think I understand this, but isn't obvious by reading the code. After putting an IC into an `Unmanaged` state, the DNS record should not be deleted.

Could you put a comment here just for future test debugging?
```suggestion
        // Even though the ingress controller's DNSManagementPolicy is now unmanaged, DNSRecords will not
        // get deleted and should still continue to function normally.
	verifyExternalIngressController(t, name, "apps."+ic.Spec.Domain, wildcardRecord.Spec.Targets[0])
```

test/e2e/unmanaged_dns_test.go
thejasn: > That makes sense and is valid. Did you consider using just a dnslookup instead of HTTP client request? An entire http request makes sense, but seems to be quite broad. Just curious, not necessarily recommending a change.

I initially did think of using `dnslookup`, but found a couple of issues with it.
- In the scenario where we have a `Unmanaged` external ingresscontroller, we would need to create new DNS records on the cloud providers (and their cleanup). This kind of adds a lot of test code to maintain.
- Since the test creates a lot of new objects (router,certs,secrets,ingresscontrollers) just testing using `dnslookup` doesn't exactly cover all aspects. So adding a HTTP call, ensures a broad test covering all the new objects.

> My other comment is, now that I understand these functions, would it make more sense to put them in the util_test.go? Because these are really quite generic IngressController tests that could be reused.

Agreed, moved to `util_test.go` along with `godocs`.

test/e2e/unmanaged_dns_test.go
thejasn: > I suppose a DNSRecord gets created regardless even when starting out with Unmanaged.

Yes, the DNSRecord CR is always created and is ensured that it always exists in both `Managed` and `Unmanaged` policies. When transitioning from `Managed`=>`Unmanaged` the DNS records on the cloud provider will not be managed by the operator. So `dnsManagementPolicy` essentially controllers how the operator manages DNS records on the cloud (not the DNSRecord CR).

test/e2e/unmanaged_dns_test.go
thejasn: OK, so I finally understood what the confusion was about. I've added additional logs to help understand this better. Let me know if this makes it clear.

test/e2e/unmanaged_dns_test.go
gcs278: Got it, thanks

test/e2e/unmanaged_dns_test.go
gcs278: :+1:  Very good comment added!

gcs278:
LGTM, but pointless to add LGTM since you wait for your API vendor bump.

Thanks for the comment responses.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -15,17 +15,28 @@ import (
...
+				{ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io"},
 			},
-			expect: []string{string(operatorv1.ConditionFalse)},
+			expect: []iov1.DNSZoneStatus{{
+				DNSZone: configv1.DNSZone{
+					ID: "/subscriptions/E540B02D-5CCE-4D47-A13B-EB05A19D696E/resourceGroups/test-rg/providers/Microsoft.Network/dnszones/dnszone.io",
+				},
+				Conditions: []iov1.DNSZoneCondition{
+					{
+						Type:   "Published",
```
gcs278, line 35: is this a reason these `"Published"` can't be `iov1.DNSRecordPublishedConditionType` instead of a hard-coded string?

pkg/operator/controller/ingress/status.go
```diff
@@ -874,12 +874,29 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
 		for _, zone := range wildcardRecord.Status.Zones {
-			for _, cond := range zone.Conditions {
-				if cond.Type == iov1.DNSRecordFailedConditionType && cond.Status == string(operatorv1.ConditionTrue) {
-					// check to see if the zone is in the dnsConfig.Spec
-					// fix:BZ1942657 - relates to status changes when updating DNS PrivateZone config
-					if checkZoneInConfig(dnsConfig, zone.DNSZone) {
-						failedZones = append(failedZones, zone.DNSZone)
+			if len(zone.Conditions) > 0 {
+
+				// When scope of load balancer is changed from internal to external
```
gcs278, line 879: +1 I really like comments like these that explain why; saves me so much time reviewing.

gcs278 (commented):
Changes look good just one comment.

pkg/operator/controller/dns/controller_test.go
thejasn: I have a couple of reasons, 
- Using the variable kind of doesn't ensure the value of the literal is exactly what we want. 
- And for literals that would almost never change, this would ensure the tests would break even if someone changed the value of the variable. 

Moreover, most of the other tests are also written similarly. (assuming its due to the same reasoning)

pkg/operator/controller/ingress/dns.go
alebedev87: > It was decided we could converge the 2 implementations and settle with the current solution, where the DNSRecord CR is created in an Unmanaged state. I'll be raising the needed PRs once this is merged.

So, the missing peace here is just setting `DNSRecord` as `Unmanaged` if the domain doesn't match and IC is `Managed`, right? Because if IC is `Unmanaged` we will create `Unmanaged` `DNSRecord` in any case (matching the domain or not).
Isn't it like half way implemented? I mean if a followup PR is expected, why not leave this condition untouched - letting IC not create `DNSRecord` if the domain doesn't match. Or just add the missing peace I said about above.

pkg/operator/controller/dns/controller.go
```diff
@@ -275,7 +285,13 @@ func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1
 			LastTransitionTime: metav1.Now(),
 		}
 
-		if recordIsAlreadyPublishedToZone(record, &zone) {
+		if dnsPolicy != iov1.ManagedDNS {
```
thejasn, line 288: @gcs278 I'm planning to refactor this function to make it less verbose and also thinking of modifying the behaviour where the status would have all the condition types always. Meaning, upon
- successfully publishing the record: `Published=True` and `Failed=False`
- when failed to publish the record: `Publish=False` and `Failed=True`
- when unmanaged: `Published=False` and `Failed=False`

This would simplify the `computeDNS()` where we wouldn't have to sort and probably can reuse the previous function. WDYT?

pkg/operator/controller/dns/controller_test.go
gcs278: Okay go it. Those are valid points.

pkg/operator/controller/dns/controller.go
gcs278: That sounds much better to me if you can remove the logic of sorting and checking the last condition. I'm using this pattern of always including all conditions with either true or false in my PR (https://github.com/openshift/cluster-ingress-operator/pull/769). So I think it's a valid idea.

thejasn: Interesting pattern, might use it in the future. So post [this](https://coreos.slack.com/archives/CE4L0F143/p1659622155245749?thread_ts=1659356092.005309&cid=CE4L0F143), we will not need to complicate this further. It'll mostly go back to the previous implementation.

pkg/operator/controller/ingress/dns.go
thejasn: Hmm, I just checked the change seems small as well so might as well add it. 
cc: @gcs278

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +266,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
+		condition.Status = string(operatorv1.ConditionTrue)
+		condition.Type = iov1.DNSRecordPublishedConditionType
+		condition.Reason = "ProviderSuccess"
+		condition.Message = "The DNS provider succeeded in replacing the record"
+	}
+
+	return condition, err
+}
+
+// publishRecord ensures the given record is publised to the provided zone
```
gcs278, line 297: *published*

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +266,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
 	return nil
 }
 
+// replacePlublishedRecord replaces previously published records
+// and the result is returned as a condition. Upon errors during publishing
+// an error object is returned.
+func (r *reconciler) replacePlublishedRecord(zone configv1.DNSZone, record *iov1.DNSRecord) (iov1.DNSZoneCondition, error) {
```
gcs278, line 272: Replace replacePlublishedRecord with replacePublishedRecord

pkg/operator/controller/dns/controller.go
```diff
@@ -433,6 +473,42 @@ func conditionChanged(a, b iov1.DNSZoneCondition) bool {
...
+		} else if cond.Type == iov1.DNSRecordFailedConditionType && cond.Status == string(operatorv1.ConditionFalse) {
+			result = append(result, iov1.DNSZoneCondition{
+				Type:               iov1.DNSRecordPublishedConditionType,
+				Status:             string(operatorv1.ConditionTrue),
+				LastTransitionTime: cond.LastTransitionTime,
+				Reason:             cond.Reason,
+				Message:            cond.Message,
+			})
+		} else {
+			result = append(result, iov1.DNSZoneCondition{
```
gcs278, line 500: Is there a reason you couldn't just do:
`result = append(result, cond)`?

pkg/operator/controller/dns/controller.go
```diff
@@ -433,6 +473,42 @@ func conditionChanged(a, b iov1.DNSZoneCondition) bool {
 	return a.Status != b.Status || a.Reason != b.Reason || a.Message != b.Message
 }
 
+// sanitizeRecordStatusCondition purges deprecated fields from DNS Record
+// status. Removes "Failed" (DNSRecordFailedConditionType) and replaces with
+// "Published" (DNSRecordPublishedConditionType). Returns updated condition
+// list.
```
gcs278, line 479: Is there a time in which this function can be removed? Say a OCP release? Could we include instructions on when this can be removed?

pkg/operator/controller/dns/controller.go
thejasn: fixed

pkg/operator/controller/dns/controller.go
thejasn: Added to godoc.

gcs278 (commented):
Just took another pass, everything looks good to me, waiting on official API vendoring.

test/e2e/util_test.go
```diff
@@ -19,6 +22,7 @@ import (
 	"k8s.io/utils/pointer"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
Miciah, line 25: Duplicate import.

test/e2e/util_test.go
```diff
@@ -402,3 +434,149 @@ func updateInfrastructureConfigSpecWithRetryOnConflict(t *testing.T, name types.
...
 	})
 }
+
+// verifyExternalIngressController verifies connectivity between the router
+// and a test workload by making a http call using the hostname passed to it.
+// This hostname must be the domain associated with the ingresscontroller under test.
+// This function overrides the HTTP HOST header with hostname argument, which is needed
+// when no DNS records have been created on the cloud provider for said hostname.
+func verifyExternalIngressController(t *testing.T, name types.NamespacedName, hostname, address string) {
+	echoPod := buildEchoPod(name.Name, name.Namespace)
```
Miciah, line 444:
```suggestion
func verifyExternalIngressController(t *testing.T, name types.NamespacedName, hostname, address string) {
	t.Helper()
	echoPod := buildEchoPod(name.Name, name.Namespace)
```

test/e2e/util_test.go
```diff
@@ -402,3 +434,149 @@ func updateInfrastructureConfigSpecWithRetryOnConflict(t *testing.T, name types.
...
+	}
+
+	// we use HOST header to map to the domain associated on the ingresscontroller.
+	// This ensures our http call is routed to the correct router.
+	req.Host = hostname
+
+	httpClient := http.Client{Timeout: 5 * time.Second}
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
+		if r.StatusCode == http.StatusOK {
+			t.Logf("verified connectivitiy with workload with req %v and response %v", req.URL, r.StatusCode)
```
Miciah, line 486:
```suggestion
			t.Logf("verified connectivity with workload with req %v and response %v", req.URL, r.StatusCode)
```

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -564,17 +565,18 @@ func TestTLSProfileSpecForSecurityProfile(t *testing.T) {
 	invalidTLSVersion := configv1.TLSProtocolVersion("abc")
 	invalidCiphers := []string{"ECDHE-ECDSA-AES256-GCM-SHA384", "invalid cipher"}
 	validCiphers := []string{"ECDHE-ECDSA-AES256-GCM-SHA384"}
-	tlsVersion13Ciphers := []string{"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256",
+	tlsVersion13Ciphers := []string{
+		"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256",
 		"TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256", "TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384",
-		"TLS_CHACHA20_POLY1305_SHA256", "TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256"}
+		"TLS_CHACHA20_POLY1305_SHA256", "TLS_AES_128_CCM_SHA256", "TLS_AES_128_CCM_8_SHA256",
+	}
```
Miciah, line 572: Thanks!  This has been bothering me for a while.

test/e2e/operator_test.go
```diff
@@ -81,6 +81,7 @@ var (
 		{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionTrue},
 		{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionTrue},
 		{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
+		{Type: operatorv1.OperatorStatusTypeProgressing, Status: operatorv1.ConditionFalse},
```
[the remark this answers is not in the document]
thejasn: I think one of the tests failed once or twice (when updating from `Umanaged` to `Managed`) because the condition was satisfied even though the operator was still progressing.

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +274,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
-				condition.Message = "The DNS provider succeeded in ensuring the record"
-			}
+			condition, err = r.publishRecord(zone, record)
 		}
+
+		// checks if replacing/publishing record resulted in an error
+		// if so, we re-enqueue request for processing
+		if err != nil {
+			result.RequeueAfter = 30 * time.Second
+		}
```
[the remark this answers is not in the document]
thejasn: Would this refactor involve some kind of error aggregation as well?

pkg/operator/controller/dns/controller.go
```diff
@@ -170,6 +180,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, nil
 	}
 
+	dnsPolicy := record.Spec.DNSManagementPolicy
+	if dnsPolicy == iov1.UnmanagedDNS {
+		r.recorder.Eventf(record, "Normal", "UnmanagedDNS", "Ingresscontroller DNS management policy configured as Unmanaged; No DNS records created on the cloud provider and this Record will be ignored.")
+	}
```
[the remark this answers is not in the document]
thejasn: I've gone ahead and removed the event.

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +274,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
+	if err != nil {
+		log.Error(err, "failed to replace DNS record in zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionFalse)
+		condition.Reason = "ProviderError"
+		condition.Message = fmt.Sprintf("The DNS provider failed to replace the record: %v", err)
+
+	} else {
+		log.Info("replaced DNS record in zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionTrue)
+		condition.Type = iov1.DNSRecordPublishedConditionType
```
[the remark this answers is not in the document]
thejasn: Fixed

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +274,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
+	err := r.dnsProvider.Ensure(record, zone)
+	if err != nil {
+		log.Error(err, "failed to publish DNS record to zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionFalse)
+		condition.Reason = "ProviderError"
+		condition.Message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
+	} else {
+		log.Info("published DNS record to zone", "record", record.Spec, "dnszone", zone)
+		condition.Status = string(operatorv1.ConditionTrue)
+		condition.Type = iov1.DNSRecordPublishedConditionType
```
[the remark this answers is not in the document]
thejasn: Fixed

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +274,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
+	var err error
+	dnsPolicy := record.Spec.DNSManagementPolicy
 	for i := range zones {
 		zone := zones[i]
 
 		// Only publish the record if the DNSRecord has been modified
 		// (which would mean the target could have changed) or its
 		// status does not indicate that it has already been published.
-		if record.Generation == record.Status.ObservedGeneration && recordIsAlreadyPublishedToZone(record, &zone) {
+		if record.Generation == record.Status.ObservedGeneration && dnsPolicy != iov1.UnmanagedDNS && recordIsAlreadyPublishedToZone(record, &zone) {
```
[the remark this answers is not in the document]
thejasn: Fixed

pkg/operator/controller/dns/controller.go
```diff
@@ -255,60 +274,108 @@ func (r *reconciler) createDNSProviderIfNeeded(dnsConfig *configv1.DNS) error {
...
 			log.Info("skipping zone to which the DNS record is already published", "record", record.Spec, "dnszone", zone)
 			continue
 		}
 
 		condition := iov1.DNSZoneCondition{
 			Status:             string(operatorv1.ConditionUnknown),
-			Type:               iov1.DNSRecordFailedConditionType,
+			Type:               iov1.DNSRecordPublishedConditionType,
 			LastTransitionTime: metav1.Now(),
 		}
```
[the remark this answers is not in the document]
thejasn: Fixed.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -162,6 +225,53 @@ func TestPublishRecordToZonesMergesStatus(t *testing.T) {
...
+					Status: string(operatorv1.ConditionFalse),
+				},
+			},
+			expected: []iov1.DNSZoneCondition{
+				{
+					Type:   iov1.DNSRecordPublishedConditionType,
+					Status: string(operatorv1.ConditionTrue),
+				},
+			},
+		},
```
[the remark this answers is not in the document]
thejasn: Added the test case.

pkg/operator/controller/ingress/status.go
```diff
@@ -822,25 +819,34 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
+	var conditions []operatorv1.OperatorCondition
+	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+		conditions = []operatorv1.OperatorCondition{
+			{
+				Type:    operatorv1.DNSManagedIngressConditionType,
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "UnmanagedLoadBalancerDNS",
+				Message: "The DNS management policy is set to Unmanaged.",
+			},
+		}
```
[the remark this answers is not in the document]
thejasn: Fixed.

pkg/operator/controller/ingress/status.go
```diff
@@ -862,7 +875,7 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 		var failedZones []configv1.DNSZone
 		for _, zone := range wildcardRecord.Status.Zones {
 			for _, cond := range zone.Conditions {
-				if cond.Type == iov1.DNSRecordFailedConditionType && cond.Status == string(operatorv1.ConditionTrue) {
+				if cond.Type == iov1.DNSRecordPublishedConditionType && cond.Status == string(operatorv1.ConditionFalse) {
```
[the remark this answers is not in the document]
thejasn: Hmm, there is no user action needed when "Published=Unknown" so I don't think we should. Is there any other use case for adding it to `status`?

pkg/operator/controller/ingress/dns.go
```diff
@@ -110,6 +107,14 @@ func desiredWildcardDNSRecord(ic *operatorv1.IngressController, service *corev1.
 		target = ingress.IP
 	}
 
+	dnsPolicy := iov1.ManagedDNS
+
+	// Set the DNS management policy on the dnsrecord to "Unmanaged" if ingresscontroller has "Unmanaged" DNS policy or
+	// if the ingresscontroller domain isn't a subdomain of the cluster's base domain.
+	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS || !domainMatchesBaseDomain {
```
Miciah, line 114: Is it necessary to check `domainMatchesBaseDomain` here if `Reconcile` or `setDefaultPublishingStrategy` sets `ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy`?  I'm assuming `Reconcile` will correctly set `ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy` if an older version of the operator left it unset.

pkg/operator/controller/ingress/status.go
```diff
@@ -822,25 +819,28 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
-		{
+	var conditions []operatorv1.OperatorCondition
+	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.DNSManagedIngressConditionType,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "UnmanagedLoadBalancerDNS",
+			Message: "The DNS management policy is set to Unmanaged.",
+		})
+	} else if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) {
```
Miciah, line 830: Same question as for `desiredWildcardDNSRecord`: Is it necessary to check `domainMatchesBaseDomain` here if `Reconcile` or `setDefaultPublishingStrategy` sets `ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy`?

pkg/operator/controller/ingress/controller.go
```diff
@@ -277,12 +277,21 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	domainMatchesBaseDomain := manageDNSForDomain(ingress.Status.Domain, platformStatus, dnsConfig)
```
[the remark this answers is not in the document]
thejasn: Initially this was the implementation. But turns out for the default ingress controller, domain isn't set during creation but the status updated within `admit()` (`spec.domain` is still blank). So kind of forced to call `domainMatchesBaseDomain` twice.

pkg/operator/controller/ingress/controller.go
```diff
@@ -277,12 +277,21 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-			Scope: operatorv1.ExternalLoadBalancer,
+			Scope:               operatorv1.ExternalLoadBalancer,
+			DNSManagementPolicy: operatorv1.ManagedLoadBalancerDNS,
 		}
+
+		// Set dnsManagementPolicy based on current domain on the ingresscontroller
+		// and base domain on dns config.
+		if !domainMatchesBaseDomain {
+			ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
+		}
```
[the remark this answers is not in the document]
thejasn: > so by now, ingress.Status.EndpointPublishingStrategy.LoadBalancer should always already be non-nil, making this dead code.

Since you mentioned this is more or less dead code as of now, I'll go ahead and remove it, since the same is being done within `setDefaultPublishingStrategy()`.

>Is it your intention to initialize ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy if the ingresscontroller has been admitted by an earlier version of the operator that didn't set DNSManagementPolicy?

Yes, but I figured I'd keep the invocations being done in `desiredWildcardDNSRecord` and `computeDNSStatus` just as a precaution if `ic.status` wasn't updated. I've gone ahead and removed them.

pkg/operator/controller/ingress/status.go
```diff
@@ -860,32 +867,49 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
+				switch cond.Status {
+				case string(operatorv1.ConditionFalse):
 					// check to see if the zone is in the dnsConfig.Spec
 					// fix:BZ1942657 - relates to status changes when updating DNS PrivateZone config
-					if checkZoneInConfig(dnsConfig, zone.DNSZone) {
-						failedZones = append(failedZones, zone.DNSZone)
-					}
+					failedZones = append(failedZones, zone.DNSZone)
+				case string(operatorv1.ConditionUnknown):
+					unknownZones = append(failedZones, zone.DNSZone)
```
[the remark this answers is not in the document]
thejasn: While adding tests, I remembered that with the current implementation we kind of give precedence to `Failed` over `Unknown`. Do you think we need to document it in addition to the current docs?
```go
		} else if len(unknownZones) != 0 {
			// This condition is an edge case where DNSManaged=True but
			// there was an internal error during publishing record.
			conditions = append(conditions, operatorv1.OperatorCondition{
				Type:    operatorv1.DNSReadyIngressConditionType,
				Reason:  "UnknownZones",
				Message: fmt.Sprintf("Provisioning of the record is in an unknown state in some zones: %v", unknownZones),
			})
```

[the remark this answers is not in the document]
Miciah: Are you asking whether `computeDNSStatus` should report if there are both failed zones and unknown zones?  That would be fine, but I don't think it's necessary.

pkg/operator/controller/dns/controller.go
```diff
@@ -480,18 +478,21 @@ func conditionChanged(a, b iov1.DNSZoneCondition) bool {
...
 func (r *reconciler) migrateRecordStatusConditions(record *iov1.DNSRecord) (bool, error) {
 	var updateConditions bool
 	for i := range record.Status.Zones {
-		var updateZone bool
-		updateZone, record.Status.Zones[i].Conditions = migrateRecordStatusCondition(record.Status.Zones[i].Conditions)
+		zone := record.Status.DeepCopy().Zones[i]
+		updateZone, cond := migrateRecordStatusCondition(zone.Conditions)
 		if updateZone {
 			updateConditions = true
+			record.Status.Zones[i].Conditions = mergeConditions(zone.Conditions, cond)
```
Miciah, line 495: Ah, was the logic setting the last transition time incorrectly before this change?

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
Miciah: > > so by now, ingress.Status.EndpointPublishingStrategy.LoadBalancer should always already be non-nil, making this dead code.
> 
> Since you mentioned this is more or less dead code as of now, I'll go ahead and remove it, since the same is being done within `setDefaultPublishingStrategy()`.

Sorry if I wasn't clear.  Note that `ingress.Status.EndpointPublishingStrategy.LoadBalancer` should always be non-nil, but `ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy` could still be empty, on upgraded clusters.  Or am I missing something?  Unless I am, you still need some logic to set `DNSManagementPolicy` if it is empty but the ingresscontroller is already admitted.

pkg/operator/controller/dns/controller.go
thejasn: It wasn't merging the condition types resulting in multiple entries of the same condition type.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
thejasn: Hmm, you are correct in one specific scenario. In most scenarios, `DNSManagementPolicy` will be to the default value i.e. `Managed` since our defaulting fix which results in all reads from `etcd` having the default value. The only time it makes sense to not rely on this would be when an already admitted controller has its domain not matching (also need to be on AWS platform).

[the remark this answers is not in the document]
thejasn: I've replaced the older logic with this,
```go
    // During upgrades, an already admitted controller might require overriding
    // default dnsManagementPolicy to "Unmanaged" due to mismatch in its domainand
    // and the pre-configured base domain.
    // TODO: Remove this in 4.13
    domainMatchesBaseDomain := manageDNSForDomain(ingress.Status.Domain, platformStatus, dnsConfig)
    if ingress.Status.EndpointPublishingStrategy != nil &&
    ingress.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType &&
    ingress.Status.EndpointPublishingStrategy.LoadBalancer != nil {

        // Set dnsManagementPolicy based on current domain on the ingresscontroller
        // and base domain on dns config.
       if !domainMatchesBaseDomain {
                ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS

                if err := r.client.Status().Update(context.TODO(), ingress); err != nil {
                     return reconcile.Result{}, fmt.Errorf("failed to update status: %v", err)
                }
                log.Info("Updated ingresscontroller status: dnsManagementPolicy as Unmanaged", "ingresscontroller", ingress.Status)
	}
}

```

pkg/operator/controller/dns/controller.go
thejasn: There was an edge case within migration code, I've added an additional test made the required changes in migration function.

pkg/operator/controller/ingress/controller.go
```diff
@@ -277,16 +277,26 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	// and the pre-configured base domain.
+	// TODO: Remove this in 4.13
+	domainMatchesBaseDomain := manageDNSForDomain(ingress.Status.Domain, platformStatus, dnsConfig)
+	if ingress.Status.EndpointPublishingStrategy != nil &&
+		ingress.Status.EndpointPublishingStrategy.Type == operatorv1.LoadBalancerServiceStrategyType &&
+		ingress.Status.EndpointPublishingStrategy.LoadBalancer != nil {
+
+		// Set dnsManagementPolicy based on current domain on the ingresscontroller
+		// and base domain on dns config.
+		if !domainMatchesBaseDomain && ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy != operatorv1.UnmanagedLoadBalancerDNS {
```
Miciah, line 291:
```suggestion
	if eps := ingress.Status.EndpointPublishingStrategy; eps != nil && eps.Type == operatorv1.LoadBalancerServiceStrategyType && eps.LoadBalancer != nil {

		// Set dnsManagementPolicy based on current domain on the ingresscontroller
		// and base domain on dns config.
		if !domainMatchesBaseDomain && eps.LoadBalancer.DNSManagementPolicy != operatorv1.UnmanagedLoadBalancerDNS {
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -277,16 +277,26 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+			ingress.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
+
+			if err := r.client.Status().Update(context.TODO(), ingress); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to update status: %v", err)
```
Miciah, line 295:
```suggestion
				return reconcile.Result{}, fmt.Errorf("failed to update status: %w", err)
```

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
Miciah: It might warrant a comment on the second call explaining why we need to call twice.

pkg/operator/controller/dns/controller.go
```diff
@@ -415,6 +476,73 @@ func conditionChanged(a, b iov1.DNSZoneCondition) bool {
...
+					Reason:             cond.Reason,
+					Message:            cond.Message,
+				}
+				// We don't need to map Failed=Unknown condition
+				// since the operator never set this condition to Unknown.
+				// Since no version of the operator had this condition, we
+				// don't need to migrate it.
+			}
+		}
+		result = mergeConditions(result, []iov1.DNSZoneCondition{updatedCondition})
```
Miciah, line 541: Won't this add a zero status condition for `Published=Unknown`?  I know we don't expect that status condition ever to be set, but I would still prefer to have a default case:
```suggestion
			switch cond.Status {
			case string(operatorv1.ConditionTrue):
				updatedCondition = iov1.DNSZoneCondition{
					Type:               iov1.DNSRecordPublishedConditionType,
					Status:             string(operatorv1.ConditionFalse),
					LastTransitionTime: now,
					Reason:             cond.Reason,
					Message:            cond.Message,
				}
			case string(operatorv1.ConditionFalse):
				updatedCondition = iov1.DNSZoneCondition{
					Type:               iov1.DNSRecordPublishedConditionType,
					Status:             string(operatorv1.ConditionTrue),
					LastTransitionTime: now,
					Reason:             cond.Reason,
					Message:            cond.Message,
				}
			default:
				// We don't need to map Failed=Unknown condition
				// since the operator never set this condition to Unknown.
				// Since no version of the operator had this condition, we
				// don't need to migrate it.
				continue
			}
		}
		result = mergeConditions(result, []iov1.DNSZoneCondition{updatedCondition})
```

test/e2e/unmanaged_dns_test.go
```diff
@@ -0,0 +1,287 @@
...
+	iov1 "github.com/openshift/api/operatoringress/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestUnmanagedDNSToManagedIngressController(t *testing.T) {
```
gcs278, line 21: *nit* if you look at the names of the e2e tests, they are not quite consistent:
`TestUnmanagedDNSToManagedIngressController`
`TestManagedDNSToUnmanagedDNSIngressController`
`TestUnmanagedDNSToManagedInternalIngressController`
Maybe just put DNS always after Unmanaged/Managed.

gcs278 (commented):
Couple of wording nits, but otherwise looks good.

gcs278:
/lgtm
@Miciah need another review from you.

Miciah:
The new changes look fine.
/lgtm

Miciah:
/hold cancel
per https://github.com/openshift/cluster-ingress-operator/pull/781#issuecomment-1195419077 since https://github.com/openshift/api/pull/1200 has been merged and incorporated into this PR.

thejasn:
`e2e-gcp-ovn-serial` currently fails due to https://issues.redhat.com/browse/OCPBUGS-1083

Miciah:
/lgtm
because the last push was only to rebase.

thejasn:
/test e2e-aws-operator
unrelated failure 
```
        --- FAIL: TestAll/serial/TestRouterCompressionOperation (75.09s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference 
```

thejasn:
/test e2e-aws-operator
test failed while collecting must gather

thejasn:
/test e2e-aws-operator
failed during cluster installation due to AWS

PR #790, opened 2022-06-23
Enable switching of AWS CLB to NLB without deletion of IngressController object.

miheer:
As an user/admin, I want to switch an OCP Ingress from an AWS Classic Load Balancer to AWS Network Load Balancer. I should be able to do this without having to delete and recreate the IngressController object.
Using this fix one can update the ingress controller object with change of LB type details and get the LB service reflected with the necessary changes without the need for recreation.

Fixes https://issues.redhat.com/browse/NE-865

`pkg/operator/controller/configurable-route/controller.go` We need to mention the key or an error is thrown.
`pkg/operator/controller/ingress/controller.go` Logic added for setting the right endpoint publishing strategy while switching from ELB to NLB or vice-versa.
`pkg/operator/controller/ingress/controller_test.go` Unit tests for switching between NLB and CLB.
`pkg/operator/controller/ingress/deployment_test.go`  We need to mention the key or an error is thrown.
`pkg/operator/controller/ingress/load_balancer_service.go` Logic to check difference between desired and current service for the lb type annotation and proxy protocol annotation and update if the current service with the desired service annotations if necessary.
`pkg/operator/controller/ingress/status.go` Adds progressing status condition logic while switching between NLB and ELB and vice-versa.
` pkg/operator/controller/ingress/status_test.go` Adds a nil as the computeIngressProgressingCondition now also accepts current service as a parameter but for this unit test current lb service is not needed.
`test/e2e/all_test.go` Adds e2e test written for this PR.
`test/e2e/checkinterval_test.go` We need to mention the key or an error is thrown.
`test/e2e/operator_test.go` e2e test for this PR.
`test/e2e/reloadinterval_test.go` We need to mention the key or an error `./reloadinterval_test.go:61:31: k8s.io/apimachinery/pkg/apis/meta/v1.Duration composite literal uses unkeyed fields` is thrown.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -478,38 +478,44 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 			}
 			switch lbType {
 			case operatorv1.AWSLoadBalancerProvider:
-				// The only provider parameter that is supported
-				// for AWS is the connection idle timeout for
-				// classic ELBs.
-				var specIdleTimeout, statusIdleTimeout metav1.Duration
-				if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					specIdleTimeout = specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
+				if statusLB.ProviderParameters == nil {
+					statusLB.ProviderParameters = &operatorv1.ProviderLoadBalancerParameters{}
 				}
-				if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					statusIdleTimeout = statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
+				if len(statusLB.ProviderParameters.Type) == 0 {
+					statusLB.ProviderParameters.Type = operatorv1.AWSLoadBalancerProvider
 				}
-				if specIdleTimeout != statusIdleTimeout {
-					if statusLB.ProviderParameters == nil {
-						statusLB.ProviderParameters = &operatorv1.ProviderLoadBalancerParameters{}
-					}
-					if len(statusLB.ProviderParameters.Type) == 0 {
-						statusLB.ProviderParameters.Type = operatorv1.AWSLoadBalancerProvider
-					}
-					if statusLB.ProviderParameters.AWS == nil {
-						statusLB.ProviderParameters.AWS = &operatorv1.AWSLoadBalancerParameters{}
-					}
-					if len(statusLB.ProviderParameters.AWS.Type) == 0 {
-						statusLB.ProviderParameters.AWS.Type = operatorv1.AWSClassicLoadBalancer
-					}
+				if statusLB.ProviderParameters.AWS == nil {
+					statusLB.ProviderParameters.AWS = &operatorv1.AWSLoadBalancerParameters{}
+				}
+				// Assume the LB type is "Classic" if the field is empty.
+				specLBType := operatorv1.AWSClassicLoadBalancer
+				if specLB.ProviderParameters.AWS != nil && len(specLB.ProviderParameters.AWS.Type) != 0 {
+					specLBType = specLB.ProviderParameters.AWS.Type
+				}
+				if specLBType != statusLB.ProviderParameters.AWS.Type {
+					statusLB.ProviderParameters.AWS.Type = specLBType
+					changed = true
+				}
+				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSClassicLoadBalancer {
 					if statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil {
 						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters = &operatorv1.AWSClassicLoadBalancerParameters{}
 					}
-					var v metav1.Duration
-					if specIdleTimeout.Duration > 0 {
-						v = specIdleTimeout
+					// The only provider parameter that is
+					// supported for AWS Classic ELBs is the
+					// connection idle timeout.
+					var specIdleTimeout metav1.Duration
+					if specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
+						specIdleTimeout = specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
 					}
-					statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = v
-					changed = true
+					statusIdleTimeout := statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
+					if specIdleTimeout != statusIdleTimeout {
+						var v metav1.Duration
+						if specIdleTimeout.Duration > 0 {
+							v = specIdleTimeout
+						}
+						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = v
+						changed = true
+					}
 				}
 			case operatorv1.GCPLoadBalancerProvider:
 				// The only provider parameter that is supported
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1096,6 +1096,78 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
 	}
 }
 
+func TestAWSLBTypeChange(t *testing.T) {
+	t.Parallel()
+
+	if infraConfig.Status.Platform != "AWS" {
+		t.Skipf("test skipped on platform %q", infraConfig.Status.Platform)
+	}
+
+	name := types.NamespacedName{Namespace: operatorNamespace, Name: "awslb"}
+	ic := newLoadBalancerController(name, name.Name+"."+dnsConfig.Spec.BaseDomain)
+	ic.Spec.EndpointPublishingStrategy.LoadBalancer = &operatorv1.LoadBalancerStrategy{
+		Scope: operatorv1.ExternalLoadBalancer,
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	// Wait for the load balancer and DNS to be ready.
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name, availableConditionsForIngressControllerWithLoadBalancer...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	lbService := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), lbService); err != nil {
+		t.Fatalf("failed to get LoadBalancer service: %v", err)
+	}
+	if v := lbService.Annotations[ingresscontroller.AWSLBTypeAnnotation]; len(v) != 0 {
+		t.Fatalf("load balancer service has unexpected %s=%s annotation", ingresscontroller.AWSLBTypeAnnotation, v)
+	}
+
+	if err := kclient.Get(context.TODO(), name, ic); err != nil {
+		t.Fatalf("failed to get ingresscontroller %s: %v", name, err)
+	}
+
+	pp := &operatorv1.ProviderLoadBalancerParameters{
+		Type: operatorv1.AWSLoadBalancerProvider,
+		AWS: &operatorv1.AWSLoadBalancerParameters{
+			Type: operatorv1.AWSNetworkLoadBalancer,
+		},
+	}
+	ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters = pp
+
+	if err := kclient.Update(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	// Wait for the load balancer and DNS to be ready.
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name, availableConditionsForIngressControllerWithLoadBalancer...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	err := wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
+		service := &corev1.Service{}
+		if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), service); err != nil {
+			t.Logf("failed to get service %s: %v", controller.LoadBalancerServiceName(ic), err)
+			return false, nil
+		}
+		if actual, ok := service.Annotations[ingresscontroller.AWSLBTypeAnnotation]; !ok {
+			t.Logf("load balancer has no %q annotation: %v", ingresscontroller.AWSLBTypeAnnotation, service.Annotations)
+			return false, nil
+		} else if actual != ingresscontroller.AWSNLBAnnotation {
+			t.Logf("expected %s=%s, found %s=%s", ingresscontroller.AWSLBTypeAnnotation, ingresscontroller.AWSNLBAnnotation,
+				ingresscontroller.AWSLBTypeAnnotation, actual)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("timed out waiting for the service LB type annotation to be updated: %v", err)
+	}
+}
+
 // TestScopeChange creates an ingresscontroller with the "LoadBalancerService"
 // endpoint publishing strategy type and verifies that the operator behaves
 // correctly when the ingresscontroller's scope is mutated.  The correct
--- a/pkg/operator/controller/configurable-route/controller.go
+++ b/pkg/operator/controller/configurable-route/controller.go
@@ -101,7 +101,7 @@ func New(mgr manager.Manager, config Config, eventRecorder events.Recorder) (con
 func (r *reconciler) resourceToClusterIngressConfig(o client.Object) []reconcile.Request {
 	return []reconcile.Request{
 		{
-			operatorcontroller.IngressClusterConfigName(),
+			NamespacedName: operatorcontroller.IngressClusterConfigName(),
 		},
 	}
 }
--- a/pkg/operator/controller/ingress/controller_test.go
+++ b/pkg/operator/controller/ingress/controller_test.go
@@ -386,16 +386,28 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
 			expectedIC:     makeIC(spec(lb(operatorv1.ExternalLoadBalancer)), status(lb(operatorv1.ExternalLoadBalancer))),
 		},
 		{
+			name:           "loadbalancer type set to ELB",
+			ic:             makeIC(spec(elb()), status(elb())),
+			expectedResult: false,
+			expectedIC:     makeIC(spec(elb()), status(elbWithNullParameters())),
+		},
+		{
+			name:           "loadbalancer type set to NLB",
+			ic:             makeIC(spec(nlb()), status(nlb())),
+			expectedResult: false,
+			expectedIC:     makeIC(spec(nlb()), status(nlb())),
+		},
+		{
 			name:           "loadbalancer type changed from ELB to NLB",
 			ic:             makeIC(spec(nlb()), status(elb())),
-			expectedResult: false,
-			expectedIC:     makeIC(spec(nlb()), status(elb())),
+			expectedResult: true,
+			expectedIC:     makeIC(spec(nlb()), status(nlb())),
 		},
 		{
 			name:           "loadbalancer type changed from NLB to ELB",
 			ic:             makeIC(spec(elb()), status(nlb())),
-			expectedResult: false,
-			expectedIC:     makeIC(spec(elb()), status(nlb())),
+			expectedResult: true,
+			expectedIC:     makeIC(spec(elb()), status(elbWithNullParameters())),
 		},
 		{
 			name:           "loadbalancer ELB connection idle timeout changed from unset with null provider parameters to 2m",
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -157,14 +157,14 @@ func TestTuningOptions(t *testing.T) {
 	ic, ingressConfig, infraConfig, apiConfig, networkConfig, _ := getRouterDeploymentComponents(t)
 
 	// Set up tuning options
-	ic.Spec.TuningOptions.ClientTimeout = &metav1.Duration{45 * time.Second}
-	ic.Spec.TuningOptions.ClientFinTimeout = &metav1.Duration{3 * time.Second}
-	ic.Spec.TuningOptions.ServerTimeout = &metav1.Duration{60 * time.Second}
-	ic.Spec.TuningOptions.ServerFinTimeout = &metav1.Duration{4 * time.Second}
-	ic.Spec.TuningOptions.TunnelTimeout = &metav1.Duration{30 * time.Minute}
-	ic.Spec.TuningOptions.TLSInspectDelay = &metav1.Duration{5 * time.Second}
-	ic.Spec.TuningOptions.HealthCheckInterval = &metav1.Duration{15 * time.Second}
-	ic.Spec.TuningOptions.ReloadInterval = metav1.Duration{30 * time.Second}
+	ic.Spec.TuningOptions.ClientTimeout = &metav1.Duration{Duration: 45 * time.Second}
+	ic.Spec.TuningOptions.ClientFinTimeout = &metav1.Duration{Duration: 3 * time.Second}
+	ic.Spec.TuningOptions.ServerTimeout = &metav1.Duration{Duration: 60 * time.Second}
+	ic.Spec.TuningOptions.ServerFinTimeout = &metav1.Duration{Duration: 4 * time.Second}
+	ic.Spec.TuningOptions.TunnelTimeout = &metav1.Duration{Duration: 30 * time.Minute}
+	ic.Spec.TuningOptions.TLSInspectDelay = &metav1.Duration{Duration: 5 * time.Second}
+	ic.Spec.TuningOptions.HealthCheckInterval = &metav1.Duration{Duration: 15 * time.Second}
+	ic.Spec.TuningOptions.ReloadInterval = metav1.Duration{Duration: 30 * time.Second}
 
 	deployment, err := desiredRouterDeployment(ic, ingressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, false, false, nil)
 	if err != nil {
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -226,6 +226,13 @@ var (
 			// local-with-fallback annotation for kube-proxy (see
 			// <https://bugzilla.redhat.com/show_bug.cgi?id=1960284>).
 			localWithFallbackAnnotation,
+			// AWS load balancer type annotation to set either CLB/ELB or NLB
+			AWSLBTypeAnnotation,
+			// awsLBProxyProtocolAnnotation is used to enable the PROXY protocol on any
+			// AWS load balancer services created.
+			//
+			// https://kubernetes.io/docs/concepts/services-networking/service/#proxy-protocol-support-on-aws
+			awsLBProxyProtocolAnnotation,
 		)
 
 		// Azure and GCP support switching between internal and external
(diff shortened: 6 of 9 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -674,6 +674,36 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
 				condition.Reason = "ScopeChanged"
 				condition.Message = message
 				condition.Status = operatorv1.ConditionTrue
+			} else {
+				switch platform.Type {
+				case configv1.AWSPlatformType:
+					if _, ok := service.Annotations[AWSLBTypeAnnotation]; ok {
+						if (ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != "NLB") && (service.Annotations[AWSLBTypeAnnotation] == "nlb") {
```
gcs278, line 681: What about nesting the `if` and doing an `or`? I feel like it might be a little cleaner opposed to having repeated code blocks.
```suggestion
if service.Annotations[AWSLBTypeAnnotation] == AWSNLBAnnotation {
  if ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters == nil || 
     ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS == nil || 
     ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != AWSNetworkLoadBalancer {
 	condition.Reason = "AWSLBTypeChangedToNLBFromClassic"
	condition.Message = message
	condition.Status = operatorv1.ConditionTrue
}
```

gcs278: In the e2e, we updated ingress controller's `endpointPublishingStrategy` spec, then we wait for the annotation to appear on the service. Here we compare the service annotation with the IC status. Which I think it valid, but what about when the IC.spec is updated, but nor the IC status or the service reflect the change yet? Is that a situation that we should update progressing? Or is that transient enough not to care?

pkg/operator/controller/ingress/controller.go
```diff
@@ -473,17 +473,30 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
-				var specIdleTimeout, statusIdleTimeout metav1.Duration
-				if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					specIdleTimeout = specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
-				}
-				if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					statusIdleTimeout = statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
-				}
-				if specIdleTimeout != statusIdleTimeout {
+				if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+					if statusLB == nil {
```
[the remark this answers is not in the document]
gcs278: Optionally, if you do need that logic for `statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil ` you do it this way (but I am unsure because your code had both `!= nil` and `== nil`:
```suggestion
if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil {
  if statusLB.ProviderParameters.AWS.Type != operatorv1.AWSNetworkLoadBalancer || statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil  {
     statusLB.ProviderParameters = specLB.ProviderParameters
     changed = true
}
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -493,18 +506,30 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
+					// for AWS is the connection idle timeout for
+					// classic ELBs.
+					var specIdleTimeout, statusIdleTimeout metav1.Duration
+					if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
+						specIdleTimeout = specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
+					}
+					if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
+						statusIdleTimeout = statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
+					}
+					if specIdleTimeout != statusIdleTimeout {
```
gcs278, line 525: Consider initializing statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout to nil and then setting it to specIdleTimeout in this if statement. I think that might be more explicit and readable

```suggestion
if specIdleTimeout != statusIdleTimeout {
  statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = nil
  if specIdleTimeout.Duration > 0 {
    statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = specIdleTimeout
  }
  changed = true
}
```

gcs278 (commented):
Nice PR - mainly just comments on style, otherwise looks good.

pkg/operator/controller/ingress/controller.go
```diff
@@ -473,17 +473,30 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
+					} else if statusLB != nil && statusLB.ProviderParameters == nil {
+						statusLB.ProviderParameters = specLB.ProviderParameters
+						changed = true
+					} else if statusLB != nil && statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS == nil {
+						statusLB.ProviderParameters = specLB.ProviderParameters
+						changed = true
+					} else if statusLB != nil && statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil {
+						statusLB.ProviderParameters = specLB.ProviderParameters
+						changed = true
+					} else if statusLB != nil && statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
```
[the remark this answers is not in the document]
miheer: When we use ingress controller with CLB the ClassicLoadBalancerParameters can be nil or not nil. Eg- The default IC with clb one gets created with  no  ClassicLoadBalancerParameters i.e nil. So having ClassicLoadBalancerParameters is not mandatory for user to set it in the spec of the IC when he wants CLB.

pkg/operator/controller/ingress/controller.go
```diff
@@ -961,7 +986,11 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		errs = append(errs, fmt.Errorf("failed to list pods in namespace %q: %v", operatorcontroller.DefaultOperatorNamespace, err))
 	}
 
-	syncStatusErr, updated := r.syncIngressControllerStatus(ci, deployment, deploymentRef, pods.Items, lbService, operandEvents.Items, wildcardRecord, dnsConfig, platformStatus)
+	var currentLBService *corev1.Service
+	if _, currentLBService, err = r.currentLoadBalancerService(ci); err != nil {
```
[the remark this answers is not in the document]
miheer: lbService is the desiredLBService object created w.r.t the IC and not the the current LBService.

pkg/operator/controller/ingress/status.go
```diff
@@ -50,7 +50,7 @@ type expectedCondition struct {
 
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
-func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (error, bool) {
+func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, service *corev1.Service, currentLBService *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (error, bool) {
```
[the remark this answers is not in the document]
miheer: answered over there

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
miheer: When we IC has CLB then the ClassicLoadBalancerParameters is not mandatory. Eg- the default IC with CLB has no ClassicLoadBalancerParameters. So it can be set or not set.  So, we need to check for `statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters` for nil and not nil along with the value set by the user in the spec for IC which in this is NLB and then set the provider parameters of the spec to the status which then is used by the lb service code.

pkg/operator/controller/ingress/controller.go
miheer: Actually this part of the code is not related to this PR so did not touch the logic. Nil value was not set in the earlier code as well. I think may be a separate PR can be filed for this to refactor. However I feel setting nil is not needed.

pkg/operator/controller/ingress/status.go
miheer: regarding suggestion for `What about nesting the if and doing an or? I feel like it might be a little cleaner opposed to having repeated code blocks.` it will change the logic. We need to check for the nil value at every level after Provider Params as the go code will crash if nil value is encountered.

pkg/operator/controller/ingress/status.go
```diff
@@ -674,6 +674,36 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
 				condition.Reason = "ScopeChanged"
 				condition.Message = message
 				condition.Status = operatorv1.ConditionTrue
+			} else {
```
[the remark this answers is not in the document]
miheer: > Nice PR - mainly just comments on style, otherwise looks good.

Thanks!

pkg/operator/controller/ingress/status.go
miheer: > In the e2e, we updated ingress controller's `endpointPublishingStrategy` spec, then we wait for the annotation to appear on the service. Here we compare the service annotation with the IC status. Which I think it valid, but what about when the IC.spec is updated, but nor the IC status or the service reflect the change yet? Is that a situation that we should update progressing? Or is that transient enough not to care?

I did try to test that but the annotation in the service gets quickly updated and the progressing status quickly becomes false as the conditions to make progressing true become false very quickly.

miheer:
Not sure why the e2e tests are failing.  The details link next to ci tests say ->

~~~~~~~~~

Application is not available
The application is currently not serving requests at this endpoint. It may not have been started or is still starting.

Possible reasons you are seeing this page:

The host doesn't exist. Make sure the hostname was typed correctly and that a route matching this hostname exists.
The host exists, but doesn't have a matching path. Check if the URL path was typed correctly and that the route was created using the desired path.
Route and path matches, but all pods are down. Make sure that the resources exposed by this route (pods, services, deployment configs, etc) have at least one pod running.

~~~~~~~~~

I ran the e2e on local which is working fine.

[miheer@localhost cluster-ingress-operator]$ make test-e2e TEST=TestAWSLBTypeChange
GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run "TestAWSLBTypeChange" ./test/e2e
I0728 16:26:55.283754   81670 request.go:665] Waited for 1.008819894s due to client-side throttling, not priority and fairness, request: GET:https://api.misalunkswitchlb.devcluster.openshift.com:6443/apis/ingress.operator.openshift.io/v1?timeout=32s
=== RUN   TestAWSLBTypeChange
=== PAUSE TestAWSLBTypeChange
=== CONT  TestAWSLBTypeChange
    operator_test.go:1133: load balancer has no "service.beta.kubernetes.io/aws-load-balancer-type" annotation: map[service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold:2 service.beta.kubernetes.io/aws-load-balancer-healthcheck-interval:5 service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout:4 service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold:2 service.beta.kubernetes.io/aws-load-balancer-proxy-protocol:* traffic-policy.network.alpha.openshift.io/local-with-fallback:]
    operator_test.go:1145: deleted ingresscontroller awslb
--- PASS: TestAWSLBTypeChange (104.52s)
PASS
ok  	github.com/openshift/cluster-ingress-operator/test/e2e	107.041s
[miheer@localhost cluster-ingress-operator]$

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: Got it thanks for the details, that makes more sense. But checking for something that is nil or not nil is always going to enter one of this `else if` statements, it can't be anything else other than nil or not nil, so what's the point in checking `ClassicLoadBalancerParameters` in the first place? Aren't you effectively doing `exists or not exists` which is always true?

pkg/operator/controller/ingress/controller.go
gcs278: Oh yea now I see the diff. Sorry about that, this can stay as is.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: It doesn't look like the desiredLBService, it appears it's the currentLBService if I'm reading the code correctly.
`ensureLoadBalancerService` returns the currentLBService as the 2nd return argument, not the `desiredLBSerivce`:
`return true, currentLBService, nil`.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: At minimum, consider line 458 and that `statusLB` can never be null, this is a redundant check in each `else if` statement.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: To further my point, if `statusLB.ProviderParameters.AWS != nil`, what logic would make it past this point? `statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters` would always be either `nil` or not `nil`. So the `if else` statements below appear to be dead code, never able to be reached.

miheer:
@gcs278 nice review :) Thanks a lot!

miheer:
@gcs278 i have implemented the changes you asked for. The e2e test also is working fine

[miheer@localhost cluster-ingress-operator]$ make test-e2e TEST=TestAWSLBTypeChange
GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run "TestAWSLBTypeChange" ./test/e2e
I0729 11:11:36.201785   32724 request.go:665] Waited for 1.049038965s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-ln-5wfns0k-76ef8.origin-ci-int-aws.dev.rhcloud.com:6443/apis/controlplane.operator.openshift.io/v1alpha1?timeout=32s
=== RUN   TestAWSLBTypeChange
=== PAUSE TestAWSLBTypeChange
=== CONT  TestAWSLBTypeChange
    operator_test.go:1145: deleted ingresscontroller awslb
--- PASS: TestAWSLBTypeChange (114.65s)
PASS
ok  	github.com/openshift/cluster-ingress-operator/test/e2e	119.247s
[miheer@localhost cluster-ingress-operator]$

pkg/operator/controller/ingress/status.go
gcs278: > regarding suggestion for What about nesting the if and doing an or? I feel like it might be a little cleaner opposed to 
 having repeated code blocks. it will change the logic. We need to check for the nil value at every level after Provider Params as the go code will crash if nil value is encountered.

Agreed, but this is what my suggestion actually does. The code has 3 if conditions with identical outputs. This suggestion is `nil` safe because if the first condition `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters == nil` is true, then it won't execute the next `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS == nil`, so on and so forth.

Please double check my work to make sure I'm not missing something, but I'd argue this is more concise and readable. Pseudo Code:
```
if service has NLB annotation {
      if ProviderParameters is nil or ProviderParameters.AWS is nil or ProviderParameters.AWS.type is not NLB {
          set condition to AWSLBTypeChangedToNLBFromClassic
      }
}
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -473,17 +473,10 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
-				var specIdleTimeout, statusIdleTimeout metav1.Duration
-				if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					specIdleTimeout = specLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
-				}
-				if statusLB.ProviderParameters != nil && statusLB.ProviderParameters.AWS != nil && statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil {
-					statusIdleTimeout = statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout
-				}
-				if specIdleTimeout != statusIdleTimeout {
+				if specLB.ProviderParameters != nil && specLB.ProviderParameters.AWS != nil && specLB.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+					statusLB.ProviderParameters = specLB.ProviderParameters
```
gcs278, line 477: Interesting, this looks a lot cleaner. So we didn't need to check if the lbStatus's ProviderParameters to see if it was empty or a `AWSClassicLoadBalancer`?

miheer: The else part will take of this.

pkg/operator/controller/ingress/status.go
```diff
@@ -674,6 +674,29 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
...
+				switch platform.Type {
+				case configv1.AWSPlatformType:
+					if _, ok := service.Annotations[AWSLBTypeAnnotation]; ok {
+						if (service.Annotations[AWSLBTypeAnnotation] == AWSNLBAnnotation) &&
+							(ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != operatorv1.AWSNetworkLoadBalancer) ||
```
gcs278, line 682: Thanks for trying out this change. Do you really need the first stanza that checks for `nil` on `ProvidedParameters` and `AWS` if you rearrange properly?
`(ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != operatorv1.AWSNetworkLoadBalancer)`?

Couldn't you just do (in this order specifically):
```
if (service.Annotations[AWSLBTypeAnnotation] == AWSNLBAnnotation) &&
(ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters == nil ||
ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS == nil ||
ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != operatorv1.AWSNetworkLoadBalancer) {
```

The same short circuiting you are using with "don't check `ProviderParameters.AWS.Type` if a prior `&&` is false, is the same with `||` if it is true. It won't attempt to evaluate the next `||` or conditions. Based on the fact you are using `if` conditional short-circuiting, I don't see a reason not to use it with `||`. It is just as `nil` safe as this. Thoughts?

miheer: We need to check if `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != operatorv1.AWSNetworkLoadBalancer)`  is not set to NLB(generally for classic LB this is not set but there might be a case where Classic is set when someone wants to specify params in the provider) so for doing that you will need to check if ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil to ensure go does not throw nil error.

I don't think `OR` and  `== nil` will help.

Also we will also need to check the condition if  `(ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters == nil || ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS == nil)`

as for having classic LB  in ingresscontroller it is not mandatory to have the ProviderParameters set in the ingress controller.

gcs278: Sorry, I don't mean to drag this one out, but I want to make sure we are on the same page. Totally agree with what you said, but just pointing out it could be simplified which makes it a bit more readable.

So do you agree that:
```
if (a.b != nil && a.b.c != nil && a.b.c.d == "x") ||
    (a.b == nil || a.b.c == nil) {
    do something
}
```
the same as this:
```
if a.b == nil || a.b.c == nil || a.b.c.d == "x" {
    do something
}
```
?

pkg/operator/controller/ingress/status.go
miheer: @gcs278 cool I got it :) thanks! made the changes.

pkg/operator/controller/ingress/status.go
miheer: @gcs278  Agree. Thanks. PTAL at the changes I made as per your request.

pkg/operator/controller/ingress/status.go
```diff
@@ -50,7 +50,7 @@ type expectedCondition struct {
 
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
-func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, service *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (error, bool) {
+func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressController, deployment *appsv1.Deployment, deploymentRef metav1.OwnerReference, pods []corev1.Pod, currentLBService *corev1.Service, operandEvents []corev1.Event, wildcardRecord *iov1.DNSRecord, dnsConfig *configv1.DNS, platformStatus *configv1.PlatformStatus) (error, bool) {
```
gcs278, line 53: this is a nit pick, but should we really change the existing variable name `service` to `currentLBService`? In context, all of the other variable names are `deployment`, `pod`, `ic`, not `currentDeployment` or `currentIC`. Seems a little inconsistent.

I think the going assumption is that all parameters are "current" objects and not desired. I think it's better to stay consistent with these variable names to avoid confusion.

gcs278:
I just added one more comment, otherwise lgtm. @Miciah said he would review it after I did, so it's ready for you @Miciah now.

pkg/operator/controller/ingress/status.go
miheer: @gcs278  done

Miciah (commented):
Looks good overall.  I have some suggestions and questions about a few details.

pkg/operator/controller/ingress/status.go
```diff
@@ -674,6 +674,31 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
...
+							}
+						}
+					} else if !ok {
+						if ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil && ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+							message := fmt.Sprintf("The IngressController LBType was changed from NLB to Classic")
+							condition.Reason = "AWSLBTypeChangedToClassicFromNLB"
+							condition.Message = message
+							condition.Status = operatorv1.ConditionTrue
+						}
+					}
```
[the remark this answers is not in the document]
miheer: Yes it is fast and we don't see it as it is quick.

I did not find any other way.

test/e2e/operator_test.go
```diff
@@ -1096,6 +1096,89 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name, availableConditionsForIngressControllerWithLoadBalancer...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	err := wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
+		service := &corev1.Service{}
+		if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), service); err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
```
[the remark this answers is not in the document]
miheer: just added a check. do you want to delete it ? I think it does not harm to keep a check.

miheer:
@Miciah PTAL

pkg/operator/controller/ingress/controller.go
```diff
@@ -498,18 +491,30 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
+						var v metav1.Duration
+						if specIdleTimeout.Duration > 0 {
+							v = specIdleTimeout
+						}
+						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = v
+						changed = true
 					}
-					statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = v
-					changed = true
 				}
```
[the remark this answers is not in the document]
Miciah: Did you consider this suggestion?  I think it makes the logic much simpler.

test/e2e/operator_test.go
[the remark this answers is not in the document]
Miciah: It seems to me like this condition should never occur.  Anyway, either the condition cannot occur or it can.  If it cannot occur, then the check is pointless.  If it can but shouldn't occur, then why wouldn't you want the test to log it?

Miciah:
Just so it doesn't get lost due to the way GitHub threads comments, I'll note that the following questions are outstanding:

* https://github.com/openshift/cluster-ingress-operator/pull/790#discussion_r966467715
* https://github.com/openshift/cluster-ingress-operator/pull/790#discussion_r966469721

test/e2e/operator_test.go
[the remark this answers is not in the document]
miheer: OK will remove it.

miheer:
@Miciah PTAL. Thanks a lot for the suggestions!

pkg/operator/controller/ingress/controller.go
```diff
@@ -498,19 +540,31 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
+					}
+					if specIdleTimeout != statusIdleTimeout {
+						var v metav1.Duration
+						if specIdleTimeout.Duration > 0 {
+							v = specIdleTimeout
+						}
+						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters.ConnectionIdleTimeout = v
+						changed = true
+					}
+				}*/
```
Miciah, line 567: No need to keep the commented-out code.

pkg/operator/controller/ingress/controller.go
```diff
@@ -477,39 +477,46 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 				lbType = specLB.ProviderParameters.Type
 			}
 			switch lbType {
+			//case operatorv1.AWSLoadBalancerProvider:
```
Miciah, line 480: No need to keep this commented-out code.

miheer:
@Miciah PTAL

miheer:
@gcs278 can you please /approved ?

candita:
Failing invariants reported in https://coreos.slack.com/archives/C01CQA76KMX/p1664216773179949, as this started after a change in openshift/origin made for https://issues.redhat.com//browse/TRT-474

[sig-network-edge] ns/openshift-authentication route/oauth-openshift disruption/ingress-to-oauth-server connection/new should be available throughout the test
[sig-network-edge] ns/openshift-console route/console disruption/ingress-to-console connection/new should be available throughout the test

PR #793, opened 2022-06-24
Bug 2086887: Add better debug logging to TestIngressOperatorCacheIsNotGlobal to help debug flakes

gcs278:
Add simple debug logic because @frobware and I saw a flake for TestIngressOperatorCacheIsNotGlobal. We need to log the condition to help us understand why the test is failing.


brandisher:
/lgtm
/hold

Sticking the hold on for another ack from the team.

Miciah:
/retitle Bug 2100918: Add better debug logging to TestIngressOperatorCacheIsNotGlobal to help debug flakes

PR #794, opened 2022-06-28
Bug 2101878: Fix route status clearing race condition caused by using the cache

gcs278:
This is a follow up fix to https://github.com/openshift/cluster-ingress-operator/pull/724 which resolves a race condition caused using the cache for determining if the router deployment is still rolling out. The cache says it isn't rolling out immediately after it started rolling out.


brandisher:
Based on the comment [here](https://github.com/openshift/cluster-ingress-operator/blob/1dab76cc978beac595dca72c3c8af7dd2f6787ae/pkg/operator/operator.go#L84-L90) this looks like the correct approach to me. Using the client to get the deployment ensures that you're not getting stale data from the cache.

/lgtm

PR #795, opened 2022-06-29
[release-4.9] Bug 2100631: Fix flakey logic in haproxy timeout tests

rfredette:
Adjust both TestHAProxyTimeouts and TestHAProxyTimeoutsRejection to search for their timeout values explicitly, rather than finding all timeouts of a particular type. In order to make it more likely that the test is finding the correct line, also use more unique timeout values in TestHAProxyTimeouts

Also fix the output of stderr in these tests when they fail

---
This is a manual cherry-pick of #775

### DIFF
--- a/test/e2e/haproxy_timeouts_test.go
+++ b/test/e2e/haproxy_timeouts_test.go
@@ -23,17 +23,19 @@ import (
 )
 
 func TestHAProxyTimeouts(t *testing.T) {
+	// Use timeout values that are more likely to be unique across the ENTIRE
+	// test suite
 	const (
-		clientTimeoutInput     = 45 * time.Second
-		clientTimeoutOutput    = "45s"
+		clientTimeoutInput     = 47 * time.Second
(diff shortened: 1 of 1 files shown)


test/e2e/haproxy_timeouts_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
Miciah, line 1: The 4.9 branch is built using Go 1.16, and `//go:build` is a Go 1.17 artifact.  The original PR didn't add this, and it isn't needed in 4.9, so better if the backport not pull in this gratuitious change.

rfredette: Good catch. I have my editor configured to automatically run gofmt when I save files, and it looks like it added this when I resolved the merge conflicts. This should be fixed now.

candita:
CI tests have changed over the last few weeks, retesting all in order to assess backport risk.
/test all

candita (approved):

PR #798, opened 2022-07-04
In the infra config cluster object cluster object we need to store AwsLBType info for persistence.

miheer:
In the infra config cluster object cluster object we need to set for AwsLBType which will store and persist the information
of the desired LB type mentioned by the installer config API so that the ingress operator will refer the LB type from this object whenever someone accidentally deletes the ingress controller.
Currently, the default LB type i.e CLB/ELB gets set if you delete the ingress controller object even when the desired the LB type for default ingress controller
was set to NLB as ingress operator does not have a way to check the LB type as we don't store it.
Fixes http://issues.redhat.com/browse/NE-942


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -226,6 +226,8 @@ var (
 			// local-with-fallback annotation for kube-proxy (see
 			// <https://bugzilla.redhat.com/show_bug.cgi?id=1960284>).
 			localWithFallbackAnnotation,
+			// Sets annotation for NLB in AWS
+			AWSLBTypeAnnotation,
```
Miciah, line 230: This is needed for #790 so that the operator correctly _updates_ the annotation, but why is it needed for _this_ PR?  This PR changes the behavior when the operator _creates_ an ingresscontroller, in which case the operator should also be _creating_ a new service rather than updating an existing one.

pkg/operator/operator.go
```diff
@@ -236,7 +236,12 @@ func (o *Operator) Start(ctx context.Context) error {
 				log.Error(nil, "failed to sync cache before ensuring default ingresscontroller")
 				return
 			}
-			err := o.ensureDefaultIngressController(infraConfig)
+			ingressConfigName := operatorcontroller.IngressClusterConfigName()
+			ingressConfig := &configv1.Ingress{}
+			if err := o.client.Get(context.TODO(), ingressConfigName, ingressConfig); err != nil {
+				log.Error(err, "failed fetching ingress config: %w")
```
Miciah, line 242:
```suggestion
				log.Error(err, "failed to fetch cluster ingress config")
				return
```

pkg/operator/operator.go
```diff
@@ -370,6 +376,24 @@ func (o *Operator) ensureDefaultIngressController(infraConfig *configv1.Infrastr
 			Replicas: &replicas,
 		},
 	}
+	if ingressConfig.Spec.LoadBalancer.Platform.Type == configv1.AWSPlatformType {
+		if ingressConfig.Spec.LoadBalancer.Platform.AWS.Type == configv1.NLB {
```
Miciah, line 380: Check for nil `ingressConfig.Spec.LoadBalancer.Platform.AWS`.

pkg/operator/operator.go
```diff
@@ -370,6 +376,24 @@ func (o *Operator) ensureDefaultIngressController(infraConfig *configv1.Infrastr
...
+					Scope: "External",
+					ProviderParameters: &operatorv1.ProviderLoadBalancerParameters{
+						Type: operatorv1.AWSLoadBalancerProvider,
+						AWS: &operatorv1.AWSLoadBalancerParameters{
+							Type: operatorv1.AWSNetworkLoadBalancer,
+						},
+					},
+				},
+			}
+			ic.Spec.EndpointPublishingStrategy = &loadBalancerStrategyWithNLB
```
Miciah, line 393:
```suggestion
			ic.Spec.EndpointPublishingStrategy = &operatorv1.EndpointPublishingStrategy{
				Type: operatorv1.LoadBalancerServiceStrategyType,
				LoadBalancer: &operatorv1.LoadBalancerStrategy{
					Scope: "External",
					ProviderParameters: &operatorv1.ProviderLoadBalancerParameters{
						Type: operatorv1.AWSLoadBalancerProvider,
						AWS: &operatorv1.AWSLoadBalancerParameters{
							Type: operatorv1.AWSNetworkLoadBalancer,
						},
					},
				},
			}
```

Miciah (commented):
The commit message could be a little more succinct and formatted according to conventions.  I think the following conveys the same information:

> Use ingress config to set default LB type on AWS
> 
> The cluster ingress config object can now specify the AWS ELB type that was
> specified when installing the cluster.  The ingress operator can refer to
> this object whenever someone deletes the default ingresscontroller and the
> operator has to recreate it.
> 
> Before this commit, the operator would always set the default ELB type
> (Classic ELB) when creating the default ingresscontroller.
> 
> Fixes https://issues.redhat.com/browse/NE-975
> 
> * `pkg/operator/operator.go`: Check if infrastructure object was set with the LB type "NLB" and create the ingresscontroller accordingly.
> * `pkg/operator/controller/configurable-route/controller.go`:
> * `pkg/operator/controller/ingress/deployment_test.go`:
> * `test/e2e/checkinterval_test.go`:
> * `test/e2e/reloadinterval_test.go`: Add keys in some struct values to prevent "composite literal uses unkeyed fields" errors.

I'm assuming the changes in `go.mod`, `go.sum`, `manifests/`, `pkg/manifests/bindata.go`, and `vendor/` will be dropped.

pkg/operator/controller/ingress/load_balancer_service.go
miheer: yeah i will remove it.

miheer:
> The commit message could be a little more succinct and formatted according to conventions. I think the following conveys the same information:
> 
> > Use ingress config to set default LB type on AWS
> > The cluster ingress config object can now specify the AWS ELB type that was
> > specified when installing the cluster.  The ingress operator can refer to
> > this object whenever someone deletes the default ingresscontroller and the
> > operator has to recreate it.
> > Before this commit, the operator would always set the default ELB type
> > (Classic ELB) when creating the default ingresscontroller.
> > Fixes https://issues.redhat.com/browse/NE-975
> > 
> > * `pkg/operator/operator.go`: Check if infrastructure object was set with
> >   the LB type "NLB" and create the ingresscontroller accordingly.
> > * `test/e2e/checkinterval_test.go`:
> > * `pkg/operator/controller/configurable-route/controller.go`:
> > * `pkg/operator/controller/ingress/deployment_test.go`: Add keys in some
> >   struct values to prevent "composite literal uses unkeyed fields error".
> 
> I'm assuming the changes in `go.mod`, `go.sum`, `manifests/`, `pkg/manifests/bindata.go`, and `vendor/` will be dropped.

I did this go mod edit -dropreplace=github.com/openshift/api

go get github.com/openshift/api@1858b752d4b6f7dc82ab9dd147a129994221e01a
https://github.com/openshift/api/commit/1858b752d4b6f7dc82ab9dd147a129994221e01a

go mod tidy

go mod vendor

miheer:
@Miciah  PTAL. Your followup API also has been used to override the openshift api

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1,16 +1,14 @@
...
 	"reflect"
 	"testing"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
-
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
Miciah, line 13: Did your editor do this?  This is an odd re-ordering.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -270,9 +281,47 @@ func TestSetDefaultPublishingStrategySetsPlatformDefaults(t *testing.T) {
...
+								AWS: &configv1.AWSIngressSpec{
+									Type: configv1.NLB,
+								},
+							},
+						},
+					},
+				}
+			}
+			return nil
+		}
```
Miciah, line 322: Why not just define two variables for the two ingress configs?

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -300,6 +349,29 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
...
+			lbStrategy := lbs(scope, policy)
+			lbStrategy.ProviderParameters = &operatorv1.ProviderLoadBalancerParameters{
+				Type: operatorv1.AWSLoadBalancerProvider,
+				AWS: &operatorv1.AWSLoadBalancerParameters{
+					Type:                          operatorv1.AWSClassicLoadBalancer,
+					ClassicLoadBalancerParameters: &operatorv1.AWSClassicLoadBalancerParameters{},
+				},
+			}
+			return eps(lbStrategy)
+		}
```
Miciah, line 373: I wonder whether we can do without these functions.  The purpose of these helpers is to make the test cases easier to understand, but it's becoming confusing what the difference is between, say, `classicScope` and `elbWithNullParameters`.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -413,250 +485,304 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
...
 	}{
 		{
 			name:                    "loadbalancer scope changed from external to internal",
 			ic:                      makeIC(spec(eps(lbs(operatorv1.InternalLoadBalancer, &managedDNS))), status(eps(lbs(operatorv1.ExternalLoadBalancer, &managedDNS)))),
 			expectedResult:          true,
-			expectedIC:              makeIC(spec(eps(lbs(operatorv1.InternalLoadBalancer, &managedDNS))), status(eps(lbs(operatorv1.InternalLoadBalancer, &managedDNS)))),
+			platformStatus:          makePlatformStatus(configv1.AWSPlatformType),
+			ingressConfig:           makeIngressConfig(configv1.NLB, configv1.AWSPlatformType),
 			domainMatchesBaseDomain: true,
+			expectedIC:              makeIC(spec(eps(lbs(operatorv1.InternalLoadBalancer, &managedDNS))), status(nlbScope(operatorv1.InternalLoadBalancer, &managedDNS))),
```
Miciah, line 500: Why do you need to change the expected results of existing test cases?  This makes me nervous.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -270,9 +281,47 @@ func TestSetDefaultPublishingStrategySetsPlatformDefaults(t *testing.T) {
 // spec.endpointPublishingStrategy.
 func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
```
Miciah, line 282: I think you also need some new test cases in `TestSetDefaultPublishingStrategySetsPlatformDefaults`.  The new defaulting logic is the main point of this PR, so we should have test cases that ensure that `TestSetDefaultPublishingStrategySetsPlatformDefaults` does the right thing when setting defaults for the first time.  For `TestSetDefaultPublishingStrategySetsPlatformDefaults`, we just need test cases to verify that it correctly handles when someone clears `spec.endpointPublishingStrategy.loadBalancer.providerParameters`, right?  

I think it will be clearer what I mean if I submit a PR, so I'll do so.

pkg/operator/controller/ingress/controller.go
```diff
@@ -409,7 +409,7 @@ func setDefaultDomain(ic *operatorv1.IngressController, ingressConfig *configv1.
 	return false
 }
 
-func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStatus *configv1.PlatformStatus, domainMatchesBaseDomain bool) bool {
+func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStatus *configv1.PlatformStatus, domainMatchesBaseDomain bool, ingressConfig *configv1.Ingress) bool {
```
Miciah, line 412: I think some of the refactoring in `setDefaultPublishingStrategy` is unnecessary, but you're also missing some defaulting logic for the case `ic.Status.EndpointPublishingStrategy` is nil.  I'll submit a PR rather than try to explain in a bunch of comments.

pkg/operator/operator.go
```diff
@@ -343,7 +339,7 @@ func (o *Operator) handleSingleNode4Dot11Upgrade() error {
 
 // ensureDefaultIngressController creates the default ingresscontroller if it
 // doesn't already exist.
-func (o *Operator) ensureDefaultIngressController(infraConfig *configv1.Infrastructure) error {
+func (o *Operator) ensureDefaultIngressController(infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress) error {
```
Miciah, line 342: Are these changes still needed if `setDefaultPublishingStrategy` handles defaulting when `spec.endpointPublishingStrategy.loadBalancer.providerParameters.aws.type` is omitted?

Miciah (commented):
You should no longer need to use a `replace` in `go.mod` now that the API is merged in openshift/api.  

The commit message needs to be updated.  

It looks like you have some vestigial changes from earlier iterations of this PR.  Please remove the unnecessary changes.  

https://github.com/miheer/cluster-ingress-operator/pull/2 has some changes I propose for `controller.go` and `controller_test.go` in order to improve correctness, simplify logic, and reduce delta.

pkg/operator/controller/ingress/controller.go
```diff
@@ -503,28 +503,35 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
 			// Currently the only platforms with configurable
 			// provider-specific parameters are AWS and GCP.
-			var lbType operatorv1.LoadBalancerProviderType
-			if specLB.ProviderParameters != nil {
-				lbType = specLB.ProviderParameters.Type
-			}
-			switch lbType {
-			case operatorv1.AWSLoadBalancerProvider:
+			switch platformStatus.Type {
+			case configv1.AWSPlatformType:
```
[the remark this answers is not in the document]
miheer: I think we need this change from my understanding because nothing will happen if specLB.ProviderParameters == nil

pkg/operator/operator.go
miheer: anyway changes done

pkg/operator/controller/ingress/replicas.go
```diff
@@ -9,17 +9,17 @@ import (
 // determining the number of replicas for the default IngressController and in
 // determining the number of replicas in the Deployments corresponding to
 // IngressController resources in which the number of replicas is unset
-func DetermineReplicas(ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure) int32 {
+func DetermineReplicas(ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure) (int32, error) {
```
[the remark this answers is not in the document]
Miciah: If there are never any errors, then there is no reason to do error handling.

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
Miciah: Yeah, my suggested change set a default value for `specLB.ProviderParameters` if it is nil.  My concern is that if the code is reading from a discriminated union (as `spec.ProviderParameters` is), the code should check the union's discriminator.

pkg/operator/controller/ingress/controller.go
```diff
@@ -632,6 +625,44 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
...
+		if len(lbs.ProviderParameters.AWS.Type) == 0 {
+			lbs.ProviderParameters.AWS.Type = defaultLBType
+		}
+		switch lbs.ProviderParameters.AWS.Type {
+		case operatorv1.AWSClassicLoadBalancer:
+			if lbs.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil {
+				lbs.ProviderParameters.AWS.ClassicLoadBalancerParameters = &operatorv1.AWSClassicLoadBalancerParameters{}
+			}
+		}
+	}
```
Miciah, line 662: Sorry, I need to revise my earlier suggestion.  `setDefaultPublishingStrategy` now assumes that `specLB.ProviderParameters.GCP` is set by `setDefaultProviderParameters`, so you will need something like this.
```suggestion
		}
	case operatorv1.GCPLoadBalancerProvider:
		if lbs.ProviderParameters.GCP == nil {
			lbs.ProviderParameters.GCP = &operatorv1.GCPLoadBalancerParameters{}
		}
	}

File: pkg/operator/controller/ingress/controller_test.go [hunk 15]
```diff
@@ -255,11 +332,16 @@ func TestSetDefaultPublishingStrategySetsPlatformDefaults(t *testing.T) {
...
+			ingressConfig := tc.ingressConfig
+			if ingressConfig == nil {
+				ingressConfig = &configv1.Ingress{}
+			}
+			if actualResult := setDefaultPublishingStrategy(ic, platformStatus, tc.domainMatchesBaseDomain, ingressConfig); actualResult != true {
 				t.Errorf("expected result %v, got %v", true, actualResult)
 			}
 			if diff := cmp.Diff(tc.expectedIC, ic); len(diff) != 0 {
 				t.Errorf("got expected ingresscontroller: %s", diff)
+				//t.Errorf("got expected ingresscontroller: %s\n\n%+v", diff,ic.Status.EndpointPublishingStrategy.LoadBalancer)
```
Miciah, line 344: Oops, I didn't mean to include this commented code in https://github.com/miheer/cluster-ingress-operator/pull/2.

pkg/operator/operator.go
```diff
@@ -236,7 +236,12 @@ func (o *Operator) Start(ctx context.Context) error {
 				log.Error(nil, "failed to sync cache before ensuring default ingresscontroller")
 				return
 			}
-			err := o.ensureDefaultIngressController(infraConfig)
+			ingressConfigName := operatorcontroller.IngressClusterConfigName()
+			ingressConfig := &configv1.Ingress{}
+			if err := o.client.Get(context.TODO(), ingressConfigName, ingressConfig); err != nil {
+				log.Error(err, "failed to fetch ingress config: %w")
```
Miciah, line 242:
```suggestion
				log.Error(err, "failed to fetch ingress config")
```

pkg/operator/controller/ingress/controller_test.go
miheer: yeah the editor does it

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -458,6 +567,30 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
...
+			domainMatchesBaseDomain: true,
+		},
+		{
+			name:                    "loadbalancer type changed from Classic LB to unset, with NLB as default",
+			ic:                      makeIC(spec(eps(lbs(operatorv1.ExternalLoadBalancer, &managedDNS))), status(elb())),
+			ingressConfig:           ingressConfigWithDefaultNLB,
+			expectedResult:          true,
+			expectedIC:              makeIC(spec(eps(lbs(operatorv1.ExternalLoadBalancer, &managedDNS))), status(nlb())),
+			domainMatchesBaseDomain: true,
+		},
```
Miciah, line 585: Thinking about this further, what this test case shows is that if I have a cluster on AWS with an ingresscontroller that has a null `spec.endpointPublishingStrategy.loadBalancer.providerParameters` value and I do the following:
```shell
oc patch ingresses.config.openshift.io/cluster -p '{"spec":{"loadbalancer":{"platform":{"type":"AWS","aws":{"type":"NLB"}}}}}'
```
then the operator will change my ingresscontroller from a classic ELB to an NLB, right?

pkg/operator/controller/ingress/controller_test.go
gcs278: Done.

pkg/operator/operator.go
gcs278: Done

pkg/operator/controller/ingress/controller.go
gcs278: Done

pkg/operator/controller/ingress/controller_test.go
miheer: yes

miheer: Once we patch ingress config to NLB, the new ingress controller(not the existing running default IC or any other existing running IC) with nil ProviderParams will be created with NLB, however the existing running default IC and any other running exisiting IC until it is deleted will be Classic because the logic for recreating is in operatot.go

PR #799, opened 2022-07-04
Run internal load balancer test on OpenStack

stephenfin:
Implements OSASINFRA-2550.


candita:
/retitle OSASINFRA-2550 Run internal load balancer test on OpenStack

candita:
@stephenfin is this a work in progress?  Could you check into the CI tests and let us know if you need advice?

candita:
> @candita I think it's good to go? Those failures seem to be flakes since they're happening on aws and gcp, neither of which this should affect?

@stephenfin please check into it when you can, as we don't merge PRs with failures on required CI tests unless they are thoroughly researched.

Miciah:
/lgtm
@stephenfin, is there a CI job that will run the operator's E2E tests on OpenStack?

PR #801, opened 2022-07-06
pkg/operator/controller/status: Inline ingress controller issues

wking:
Instead of the generic:

    Not all ingress controllers are available.

Make debugging easier by explaining which IngressControllers are not available and why they aren't available.  This saves folks one step of "what do I check next?" when debugging this situation, and may include information like:

    One or more status conditions indicate unavailable: DNSReady=False (FailedZones: The record failed to provision in some zones: ...)

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -24,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -421,8 +426,8 @@ func computeOperatorProgressingCondition(ingresscontrollers []operatorv1.Ingress
 		}
 	}
 
-	if !allIngressesAvailable {
-		messages = append(messages, "Not all ingress controllers are available.")
+	if allIngressesAvailable != nil {
+		messages = append(messages, allIngressesAvailable.Error())
```
wking, line 430: possibly worth packing the `Reason` along as structured information, with some sort off aggregation in the event that there are multiple underlying reasons among different ingress controllers, to avoid the generic `progressingCondition.Reason = "Reconciling"` below, and make it easier to see what sort of failure modes we're having in Telemetry and other systems that can process `reason` on ClusterOperator conditions.  But I haven't bitten that off in this pull request.

pkg/operator/controller/status/controller.go
```diff
@@ -297,21 +298,25 @@ func (r *reconciler) computeOperatorStatusVersions(oldVersions []configv1.Operan
...
+	errs := make([]error, 0, len(ingresses))
 	for _, ing := range ingresses {
-		available := false
+		found := false
 		for _, c := range ing.Status.Conditions {
-			if c.Type == operatorv1.IngressControllerAvailableConditionType && c.Status == operatorv1.ConditionTrue {
-				available = true
+			if c.Type == operatorv1.IngressControllerAvailableConditionType {
+				if c.Status != operatorv1.ConditionTrue {
+					errs = append(errs, fmt.Errorf("ingress controller %s in namespace %s is Available=%s: %s: %s", ing.Name, ing.Namespace, c.Status, c.Reason, c.Message))
```
candita, line 308:
```suggestion
					errs = append(errs, fmt.Errorf("ingress controller %s in namespace %s is not Available=%s: %s: %s", ing.Name, ing.Namespace, c.Status, c.Reason, c.Message))
```

frobware: Isn't `c.Status` dynamic here -- "not" doesn't always hold true.

candita: Or `"ingress controller test-1 is Available=Unknown`

candita:
@wking Could you add a valid BZ reference to this?

PR #805, opened 2022-07-12
NE-586: add ReloadInterval TuningOption

Ethany-RH:
add ReloadInterval as a TuningOption and remove it from unsupportedConfigOverrides


pkg/operator/controller/ingress/deployment.go
```diff
@@ -1678,3 +1669,11 @@ func GetMIMETypes(mimeTypes []operatorv1.CompressionMIMEType) []string {
 
 	return mimes
 }
+
+func calcReloadIntervalFromInput(interval *metav1.Duration) string {
+	reloadInterval := 5 * time.Second
+	if interval != nil && interval.Duration >= 1*time.Second && interval.Duration <= 720*time.Hour {
+		reloadInterval = interval.Duration
+	}
```
Miciah, line 1677: I figured values outside of the allowed range would be capped, as the operator does for other timeouts, rather than ignored entirely:
```suggestion
	const (
		defaultInterval = 5 * time.Second
		minInterval = 1 * time.Second
		maxInterval = 720 * time.Hour
	)
	reloadInterval = defaultInterval
	if interval != nil {
		if interval.Duration < minInterval {
			reloadInterval = minInterval
		} else if [etc.]
	}
```
Does that make sense?

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestReloadInterval(t *testing.T) {
+	t.Parallel()
```
Miciah, line 23: You need to add an entry for this test in `TestAll` in <https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/all_test.go>.

Miciah:
`TestReloadIntervalUnsupportedConfigOverride` can be deleted from `operator_test.go`, and the reference to the test can be deleted from `all_test.go`, since your changes remove the override entirely.

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,94 @@
...
+		// set Spec.TuningOptions.ReloadInterval to a nondefault value within allowed bounds
+		{"set reloadInterval 100s", metav1.Duration{100 * time.Second}, "100s"},
+		{"set reloadInterval 1m", metav1.Duration{1 * time.Minute}, "1m"},
+
+		// set ReloadInterval to 0, which should give default of 5s
+		{"set reloadInterval default", metav1.Duration{0 * time.Second}, "5s"},
+
+		// set ReloadInterval to a value outside of allowed bounds
+		{"set reloadInterval 130s", metav1.Duration{130 * time.Second}, "120s"},
+		{"set reloadInterval 0.5s", metav1.Duration{500 * time.Millisecond}, "1s"},
```
Miciah, line 63: 0.5ms is rejected by the API validation.  "130s" gets capped to 120 seconds and then converted to "2m".  
```suggestion
		{"set reloadInterval 130s", metav1.Duration{130 * time.Second}, "2m"},
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1678,3 +1669,27 @@ func GetMIMETypes(mimeTypes []operatorv1.CompressionMIMEType) []string {
...
 }
+
+// caps the value of ReloadInterval between the bounds of 1s and 120s
+// returns the default of 5s if the user gives a 0 value
+func capReloadIntervalValue(interval time.Duration) time.Duration {
+	const (
+		maxInterval     = 120 * time.Second
+		minInterval     = 1 * time.Second
+		defaultInterval = 5 * time.Second
+	)
```
candita, line 1680:
```suggestion
	maxInterval     := 120 * time.Second
	minInterval     := 1 * time.Second
	defaultInterval := 5 * time.Second
	
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1678,3 +1669,27 @@ func GetMIMETypes(mimeTypes []operatorv1.CompressionMIMEType) []string {
...
+
+	reloadInterval := defaultInterval
+	if interval == 0 {
+		return reloadInterval
+	}
```
candita, line 1685:
```suggestion
	if interval == 0 {
		return defaultInterval
	}
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1678,3 +1669,27 @@ func GetMIMETypes(mimeTypes []operatorv1.CompressionMIMEType) []string {
...
+
+	if interval < minInterval {
+		reloadInterval = minInterval
+	} else if interval > maxInterval {
+		reloadInterval = maxInterval
+	} else {
+		reloadInterval = interval
+	}
+	return reloadInterval
```
candita, line 1694:
```suggestion
	switch interval {
	case interval == 0 :
		return defaultlInterval
	case interval < minInterval :
		return minInterval
	case interval > maxInterval :
		return maxInterval
	default:
		return interval
	}
```

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	deployment := &appsv1.Deployment{}
+	if err := kclient.Get(context.TODO(), controller.RouterDeploymentName(ic), deployment); err != nil {
+		t.Fatalf("failed to get ingresscontroller router deployment: %v", err)
+	}
+
+	for _, testCase := range []struct {
+		description    string
+		reloadInterval metav1.Duration
+		expectedEnvVar string
```
candita, line 52:
```suggestion
		description    string
		reloadInterval metav1.Duration
		expectedEnvVar string
		expectedSuccess bool
```

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,95 @@
...
+		// set ReloadInterval to 0, which should give default of 5s
+		{"set reloadInterval default", metav1.Duration{0 * time.Second}, "5s", true},
+
+		// set ReloadInterval to a value outside of allowed bounds
+		{"set reloadInterval 130s", metav1.Duration{130 * time.Second}, "2m", true},
+		{"set reloadInterval 0.5s", metav1.Duration{500 * time.Millisecond}, "1s", false},
+		{"set reloadInterval 1s", metav1.Duration{-1 * time.Second}, "1s", false},
+	} {
+		t.Log(testCase.description)
+		if err := setReloadInterval(t, kclient, 1*time.Minute, testCase.reloadInterval, icName); err != nil {
```
candita, line 67:
```suggestion
		if err := setReloadInterval(t, kclient, 1*time.Minute, testCase.reloadInterval, icName); err != nil && expectedSuccess {
```

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,95 @@
...
+			t.Fatalf("failed to update ingresscontroller with reloadInterval=%v: %v", testCase.reloadInterval, err)
+		}
+		if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+			t.Fatalf("failed to observe expected conditions: %v", err)
+		}
+		if err := waitForDeploymentEnvVar(t, kclient, deployment, 1*time.Minute, ingresscontroller.RouterReloadIntervalEnvName, testCase.expectedEnvVar); err != nil {
```
candita, line 73:
```suggestion
		if err := waitForDeploymentEnvVar(t, kclient, deployment, 1*time.Minute, ingresscontroller.RouterReloadIntervalEnvName, testCase.expectedEnvVar); err != nil && expectedSuccess {
```

pkg/operator/controller/ingress/deployment.go
frobware: What's the reason to make them non-const? I would have thought immutability would be preferable.

pkg/operator/controller/ingress/deployment.go
Ethany-RH: How do you use boolean expression (e.g. < and >) with switch case statements?

candita: Oops, sorry.  You can just use switch alone:
```suggestion
	switch  {
	case interval == 0 :
		return defaultInterval
	case interval < minInterval :
		return minInterval
	case interval > maxInterval :
		return maxInterval
	default:
		return interval
	}
```

candita: And you can get rid of the variable declarations if you do:

```suggestion
	switch  {
	case interval == 0 :
		return 5 * time.Second
	case interval < minInterval :
		return 1 * time.Second
	case interval > maxInterval :
		return 120 * time.Second
	default:
		return 120 * time.Second
	}
```

pkg/operator/controller/ingress/deployment.go
candita: I thought it was weird, albeit not incorrect, to declare a `const` within the same function it is used.  On second thought I think it's probably better just to return the duration value if the const isn't used outside this function.

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+}
+
+// Tests values that do not pass
+// the regex validation (i.e. values that result in an error).
+func TestReloadIntervalInvalidValues(t *testing.T) {
+	t.Parallel()
+
+	const expectedErr string = `spec.tuningOptions.reloadInterval in body should match '^(0|([0-9]+(\.[0-9]+)?(s|m))+)$'`
```
candita, line 93: Per our chat, this expectedErr could be added to your testCases.

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, testCase := range []struct {
+		description    string
+		reloadInterval metav1.Duration
+		expectedEnvVar string
+	}{
+		{"set reloadInterval 500 ms", metav1.Duration{500 * time.Millisecond}, "5s"},
+		{"set reloadInterval -1s", metav1.Duration{-1 * time.Second}, "5s"},
+	} {
+		t.Log(testCase.description)
+		err := setReloadIntervalInvalidValue(t, kclient, testCase.reloadInterval, icName)
```
candita, line 128: Also, rather than duplicate this, you should add it to the function I described in the previous comment, passing in 5, kclient, icName, deployment, and testcases as input.

Miciah:
Something went wrong with the build:
```
[33mWARN[0m[2022-08-05T14:58:05Z] Failed gathering successful build src logs into artifacts.  [33merror[0m=error: Unable to retrieve logs for build src: pod "src-build" not found
```
/test e2e-aws-operator

pkg/operator/controller/ingress/deployment.go
Miciah: @candita, are you suggesting replacing the named constants with literals?  I prefer using named consts—it should compile to the same code, and the names make the logic clearer.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1658,6 +1660,37 @@ func TestDurationToHAProxyTimespec(t *testing.T) {
...
+		// Values below the minimum 1s returns 1s.
+		{1 * time.Millisecond, 1 * time.Second},
+		{-1, 1 * time.Second},
+
+		// Values above the maximum 120s returns 120s.
+		{6 * time.Minute, 120 * time.Second},
+		{1 * time.Hour, 120 * time.Second},
+		{365 * time.Second, 120 * time.Second},
+
+		// Values in the allowed range returns itself (i.e. b/w 1s and 5 mins).
```
candita, line 1677: If you need to change anything else, you should update this comment, which has the wrong range.

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,133 @@
...
+		return true, nil
+	})
+}
+
+func setReloadIntervalInvalidValue(t *testing.T, client client.Client, reloadInterval metav1.Duration, name types.NamespacedName) error {
+	t.Helper()
+
+	ic := operatorv1.IngressController{}
+	if err := client.Get(context.TODO(), name, &ic); err != nil {
+		t.Logf("Get %q failed: %v, retrying ...", name, err)
```
candita, line 128: nit: This error message is not quite right, as it wouldn't be retrying here, since it is no longer wrapped in the `wait.PollImmediate` loop

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+		// set ReloadInterval to a value outside of allowed bounds
+		// expect no error
+		{"set reloadInterval 130s", metav1.Duration{130 * time.Second}, "2m", true},
+	} {
+		t.Log(testCase.description)
+		// cases with values that pass the regex validation
+		if testCase.expectSuccess {
+			if err := setReloadIntervalValidValue(t, kclient, 1*time.Minute, testCase.reloadInterval, icName); err != nil {
+				t.Fatalf("failed to update ingresscontroller with reloadInterval=%v: %v", testCase.reloadInterval, err)
```
candita, line 83: It would be good if this continued on to do the other tests, so if you use `Errorf` instead of `Fatalf` you can continue instead of exiting.  
```suggestion
				t.Errorf("failed to update ingresscontroller with reloadInterval=%v: %v", testCase.reloadInterval, err)
```

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			// cases with values that fail the regex validation
+		} else {
+			err := setReloadIntervalInvalidValue(t, kclient, testCase.reloadInterval, icName)
+			if err == nil {
+				t.Fatalf("expected an error")
```
candita, line 89:
```suggestion
				t.Errorf("expected an error when attempting invalid value")
```

test/e2e/reloadinterval_test.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			if !strings.Contains(err.Error(), expectedErr) {
+				t.Fatalf("expected error message %q, got %v", expectedErr, err)
```
candita, line 92:
```suggestion
				t.Errorf("expected error message %q, got %v", expectedErr, err)
```

Ethany-RH: Did you want me to change the other 2 `t.Fatalf()` to `t.Errorf()` as well? (`waitForIngressControllerCondition()` and `waitForDeploymentEnvVar`)

candita: Good question.  You could do so, but for these two steps, if it doesn't succeed then there may be something going on that could affect the subsequent. tests.

candita: And in that case, you should probably add a `continue` in the loop after the `t.Errorf`s
 you just added.

pkg/operator/controller/ingress/deployment.go
candita: That's fine with me.  I prefer consts too, but not usually within the same function they are declared.

Ethany-RH:
cc @ShudiLi for QE approval
cc @ahardin-rh for docs approval
cc @CFields651 for px approval

ShudiLi:
/label qe-approved
@Ethany-RH thanks

Ethany-RH:
@ShudiLi , @ahardin-rh , @CFields651 , if you have any questions or concerns, please contact @Miciah or @candita . My summer internship just ended last week, so I will no longer be able to receive messages on Slack or email. Thanks for all the help!

PR #819, opened 2022-08-24
Add IBM Cloud DNS support for Power VS

Karthik-K-N:
Power VS platform uses the IBM Cloud DNS Service. This PR contains the changes to support IBM Cloud DNS Service for Power VS.

Jira Story: https://issues.redhat.com/browse/MULTIARCH-2796

IBM Cloud DNS support has been added throgh the PR: https://github.com/openshift/cluster-ingress-operator/pull/796


pkg/operator/controller/dns/controller.go
```diff
@@ -595,13 +595,20 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
...
-		dnsProvider, err = getIbmDNSProvider(dnsConfig, creds, cisInstanceCRN, userAgent, true)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create IBM DNS manager: %v", err)
+		if platformStatus.PowerVS.CISInstanceCRN != "" {
+			dnsProvider, err = getIbmDNSProvider(dnsConfig, creds, platformStatus.PowerVS.CISInstanceCRN, userAgent, true)
+			if err != nil {
+				return nil, err
+			}
+		} else if platformStatus.PowerVS.DNSInstanceCRN != "" {
+			matches := ibm.IBMResourceCRNRegexp.FindStringSubmatch(platformStatus.PowerVS.DNSInstanceCRN)
```
SzucsAti, line 604: Wondering if we could get rid of the code duplication by moving the common parts to `getIbmDNSProvider` as provider creation seems to be the same for both IBMCloud VPC and PowerVS.

SzucsAti (commented):
Thank you!

pkg/operator/controller/dns/controller.go
Karthik-K-N: Thank you, I have update the code accordingly, Please take a look.

SzucsAti: Looks good to me.

Karthik-K-N:
@rfredette @lmzuccarelli could you please help in reviewing this

pkg/operator/controller/dns/controller.go
```diff
@@ -661,26 +660,30 @@ func getIbmDNSProvider(dnsConfig *configv1.DNS, creds *corev1.Secret, instanceID
...
 	if isPublic {
+		providerCfg.InstanceID = instanceCRN
 		provider, err := ibmpublicdns.NewProvider(providerCfg)
 		if err != nil {
 			return nil, fmt.Errorf("failed to initialize IBM CIS DNS provider: %w", err)
 		}
 		log.Info("successfully initialized IBM CIS DNS provider")
 		return provider, nil
 	} else {
+		matches := ibm.IBMResourceCRNRegexp.FindStringSubmatch(instanceCRN)
```
mkumatag, line 677: It will be great if you can add some comment about the regex format we are talking about.

pkg/operator/controller/dns/controller.go
```diff
@@ -661,26 +660,30 @@ func getIbmDNSProvider(dnsConfig *configv1.DNS, creds *corev1.Secret, instanceID
...
+		if len(matches) <= 0 {
+			return nil, fmt.Errorf("CRN does not match expected format: %s", instanceCRN)
```
mkumatag, line 679: can we include that regex format we are expecting here otherwise this error is going to be really difficult to digest.

mkumatag (changes requested):

pkg/operator/controller/dns/controller.go
Karthik-K-N: We have it over [here](https://github.com/Karthik-K-N/cluster-ingress-operator/blob/10966557ef8592858fe068065e526a87c37315c5/pkg/dns/ibm/common.go#L14-L16).

pkg/operator/controller/dns/controller.go
Karthik-K-N: Updated.

pkg/operator/controller/dns/controller.go
```diff
@@ -661,26 +660,30 @@ func getIbmDNSProvider(dnsConfig *configv1.DNS, creds *corev1.Secret, instanceID
...
+		providerCfg.InstanceID = instanceCRN
 		provider, err := ibmpublicdns.NewProvider(providerCfg)
 		if err != nil {
 			return nil, fmt.Errorf("failed to initialize IBM CIS DNS provider: %w", err)
 		}
 		log.Info("successfully initialized IBM CIS DNS provider")
 		return provider, nil
 	} else {
+		matches := ibm.IBMResourceCRNRegexp.FindStringSubmatch(instanceCRN)
+		if len(matches) <= 0 {
```
frobware, line 678: Can this be just `== 0`?

frobware: > Can this be just `== 0`?

On reflection, can we just check for `nil` as the Go doc says a return value of nil indicates no match.

```go
// FindStringSubmatch returns a slice of strings holding the text of the
// leftmost match of the regular expression in s and the matches, if any, of
// its subexpressions, as defined by the 'Submatch' description in the
// package comment.
// **A return value of nil indicates no match**.
```

Karthik-K-N: I have updated it to check against nil.

frobware: I will take another look tomorrow -- ran out of time today.

pkg/operator/controller/dns/controller.go
```diff
@@ -661,26 +660,30 @@ func getIbmDNSProvider(dnsConfig *configv1.DNS, creds *corev1.Secret, instanceID
...
 			return nil, fmt.Errorf("failed to initialize IBM CIS DNS provider: %w", err)
 		}
 		log.Info("successfully initialized IBM CIS DNS provider")
 		return provider, nil
 	} else {
+		matches := ibm.IBMResourceCRNRegexp.FindStringSubmatch(instanceCRN)
+		if matches == nil {
+			return nil, fmt.Errorf("CRN: %s does not match expected format: %s", instanceCRN, ibm.IBMResourceCRNRegexp)
+		}
+		providerCfg.InstanceID = matches[ibm.IBMResourceCRNRegexp.SubexpIndex("guid")]
```
frobware, line 681: I think you need to check the bound on `matches`. Looking at `ibm.IBMResourceCRNRegexp.SubexpIndex("guid")`, that can return `-1`. I think we should be more defensive here.

Karthik-K-N: As [IBMResourceCRNTegexp](https://github.com/Karthik-K-N/cluster-ingress-operator/blob/10966557ef8592858fe068065e526a87c37315c5/pkg/dns/ibm/common.go#L16-L15) is declared as constant it always expected to return 6. So do you think we need to check additionally for matches length >= 6?

frobware: Will `matches` always have 6 entries if the lookup succeeds?

Karthik-K-N: With a valid instanceCRN the matches length will be 9,
```
        IBMResourceCRNRegexp := regexp.MustCompile(`^crn:v[0-9]:(?P<cloudName>[^:]*):(?P<cloudType>[^:]*):(?P<serviceName>[^:]*):(?P<location>[^:]*):(?P<scope>[^:]*):(?P<guid>[^:]*):(?P<resourceType>[^:]*):(?P<resourceID>[^:]*)$`)
	crn := "crn:v1:bluemix:public:internet-svcs:global:a/65b64c1f1c11160e8c2e4bbfbd893c2c:453c4cff-2ee0-4309-95f1-2e9384111b96::"
	matches := IBMResourceCRNRegexp.FindStringSubmatch(crn)
	if matches == nil {
		panic(fmt.Errorf("CRN: %s does not match expected format: %s", crn, IBMResourceCRNRegexp))
	}
	fmt.Println(IBMResourceCRNRegexp.SubexpIndex("guid")) // 6
	fmt.Println(len(matches)) // 9
        fmt.Println(matches[IBMResourceCRNRegexp.SubexpIndex("guid")]) // 453c4cff-2ee0-4309-95f1-2e9384111b96
```

pkg/operator/controller/dns/controller.go
```diff
@@ -583,25 +583,24 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
...
+		if platformStatus.PowerVS.CISInstanceCRN != "" {
+			dnsProvider, err = getIbmDNSProvider(dnsConfig, creds, platformStatus.PowerVS.CISInstanceCRN, userAgent, true)
+			if err != nil {
+				return nil, err
+			}
+		} else if platformStatus.PowerVS.DNSInstanceCRN != "" {
+			dnsProvider, err = getIbmDNSProvider(dnsConfig, creds, platformStatus.PowerVS.DNSInstanceCRN, userAgent, false)
+			if err != nil {
+				return nil, err
+			}
```
frobware, line 603: Is it possible for both `CISInstanceCRN` and `DNSInstanceCRN` to be `""`?

```go
	// CISInstanceCRN is the CRN of the Cloud Internet Services instance managing
	// the DNS zone for the cluster's base domain
	CISInstanceCRN string `json:"cisInstanceCRN,omitempty"`

	// DNSInstanceCRN is the CRN of the DNS Services instance managing the DNS zone
	// for the cluster's base domain
	DNSInstanceCRN string `json:"dnsInstanceCRN,omitempty"`
```

Do we assert elsewhere that one of these fields is always set for `PowerVS`?

Karthik-K-N: No, both CISInstanceCRN and DNSInstanceCRN cannot be empty at the same time, installer will set  values for one of these.

frobware:
@network-edge I took a look over the PR and LGTM. Needs another reviewer to now (review and) approve.

candita:
@frobware - I'll add to the PR Scrub list.  Not sure why it didn't start there?

candita:
@Karthik-K-N Please add a JIRA reference and start work on getting the px-approved, qe-approved, docs-approved labels for this PR.

mkumatag:
/cc @alishaIBM @juliemathew 

for docs and QA

mkumatag:
/cc @CFields651

for px

pkg/operator/controller/dns/controller.go
Miciah, line 603: If `platformStatus.PowerVS.CISInstanceCRN` and `platformStatus.PowerVS.DNSInstanceCRN` are both empty, should `createDNSProvider` log something and return `&dns.FakeProvider{}`?  Come to think of it, the same question applies for `case configv1.IBMCloudPlatformType`.

Karthik-K-N: We expect either of this to be set by the installer but to be on the safer side  I have made changes to use the `&dns.FakeProvider{}` in case both of them are empty

Miciah:
/approve

#826 should have fixed the issue that caused e2e-aws-operator to fail.
/test e2e-aws-operator

mkumatag (approved):
/lgtm

gcs278:
I'm late to the party, but had no comments.
/lgtm

mkumatag:
@sferich888 @CFields651 any help is appreciated!

mkumatag:
@Miciah the CI jobs are failing forever and they aren't related to the changes we made in this PR, wondering if we can override the job?

PR #821, opened 2022-08-29
NE-366: Set router go process log level based on ingresscontroller.spec.logLevel

rfredette:
Add the `logLevel` field to the ingresscontroller api, which sets the log verbosity of the router's main process, and add the `operatorLogLevel` field to the cluster-wide config.ingress api to set the log verbosity of the ingress operator.

`logLevel` is set as `CONTROLLER_LOG_LEVEL` in the router pod environment, and will apply the update to the router with a pod restart. Later, the ability to change this without restarting the router pod is planned.

The ingress operator will change its log level at runtime to sync with the current value of `operatorLogLevel`

Unit tests and CI tests still todo for this PR.


rfredette:
/hold
for openshift/api#1300 api PR to merge

PR #822, opened 2022-08-30
Makefile: build tests with CGO_ENABLED=0

frobware:
This is about consistency. If we build the operator binary with
CGO_ENABLED=0 then we should do the same for both the unit tests and
the e2e tests. Using CGO_ENABLED=0 means that the tests run with the
same resolver implementation as the operator binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ crd:
 
 .PHONY: test
 test:
-	$(GO) test ./...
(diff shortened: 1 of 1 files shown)


frobware:
Step e2e-aws-operator-gather-must-gather failed after 7m0s. 

/retest

PR #824, opened 2022-09-06
OCPBUGS-853: certificate-publisher: Don't publish extraneous certificates

Miciah:
#### certificate-publisher: Delete `secretIsInUse`

Delete the `secretIsInUse` predicate for the watch on secrets.  The predicate is superfluous because the map func maps the event to an empty slice anyway if no ingresscontroller uses the secret.

* `pkg/operator/controller/certificate-publisher/controller.go` (`New`): Delete the predicates on the watch on secrets.
(`secretIsInUse`): Delete function.

#### certificate-publisher: Simplify the map func

Refactor the `secretToIngressController` map func and the `ingressControllersWithSecret` helper.  The helper is no longer used by any other functions besides `secretToIngressController`, so the helper can be inlined into `secretToIngressController`.

Remove selflink from log messages since OpenShift 4.8 turned off selflinks.

* `pkg/operator/controller/certificate-publisher/controller.go` (`secretToIngressController`): Inline `ingressControllersWithSecret`.
(`ingressControllersWithSecret`): Delete function.

#### certificate: Log reconcile requests.

* `pkg/operator/controller/certificate/controller.go` (`Reconcile`): Log the reconcile request.

#### Move "default-ingress-cert" configmap publishing

Move the reconciliation of the "default-ingress-cert" configmap from the "certificate" controller to the "certificate-publisher" controller in order to ensure that the configmap gets updated when the default-certificate secret is updated.

Before this change, the certificate-publisher controller already published the "router-certs" secret in the "openshift-config-managed" namespace.  The router-certs secret is supposed to have the certificate and key of the ingresscontroller that has the cluster ingress domain, so the controller watches secrets as well as ingresscontrollers in order to update the router-certs secret when the ingresscontroller is updated to reference a different secret or when the content of the referenced secret is updated.

In contrast, the certificate controller was originally written to generate a self-signed CA and generate default certificates for ingresscontrollers using this CA, and for these purposes, the controller only needs to watch ingresscontrollers and not the secrets themselves.

#331 added reconciliation of the default-ingress-cert configmap to the certificate controller.  This configmap has the default certificate of the "default" ingresscontroller. The configmap should be updated when the secret reference or content is updated.  Moreover, it makes sense conceptually for the certificate-publisher controller to handle both publication tasks: both the router-certs secret as well as the default-ingress-certs configmap.

* `pkg/operator/controller/certificate/controller.go` (`Reconcile`): Move the call to `ensureDefaultIngressCertConfigMap` from here...
* `pkg/operator/controller/certificate-publisher/controller.go` (`Reconcile`): ...to here.
* `pkg/operator/controller/certificate/publish_ca.go`: Rename...
* `pkg/operator/controller/certificate-publisher/publish_ca.go`: ...to this.

#### certificate-publisher: Don't publish extra certs

Publish the default certificate and key of only whichever ingresscontroller has the cluster ingress domain.

Before this change, the certificate-publisher controller published the certificates and keys of all ingresscontrollers, in the "router-certs" secret.  However, the only component that uses the router-certs secret is the authentication operator, which only needs the certificate and key for the cluster ingress domain.  Moreover, collecting the certificates and keys for all ingresscontrollers can produce a result that exceeds the maximum secret size of 1 mebibyte, causing the certificate-publisher controller to fail to create or update the router-certs secret.  This PR changes the certificate-publisher controller not to publish the extraneous certificates and keys.

* `pkg/operator/controller/certificate-publisher/controller.go`: Update comments to reflect that the controller only publishes the certificate and key of the ingresscontroller for the cluster ingress domain in the "router-certs" secret.
(`New`): Add a new predicate to the watch on ingresscontrollers, using the new `hasClusterIngressDomain` method and `isDefaultIngressController` function.
(`secretToIngressController`): Skip ingresscontrollers for domains other than the cluster ingress domain.
(`hasClusterIngressDomain`): New method.  Get the cluster ingress config and return true iff the given ingresscontroller has the same domain.
(`isDefaultIngressController`): New function.  Return true iff the given ingresscontroller is the "default" ingresscontroller.
(`Reconcile`): Get the cluster ingress config, and pass it to `ensureRouterCertsGlobalSecret`.
* `pkg/operator/controller/certificate-publisher/publish_certs.go` (`ensureRouterCertsGlobalSecret`): Add a parameter for the ingress config, and pass the domain from the ingress config to `desiredRouterCertsGlobalSecret`.
(`desiredRouterCertsGlobalSecret`): Add a parameter for the cluster ingress domain, use it to filter out extraneous ingresscontrollers, and publish the certificate and key for only the ingresscontroller that has the cluster ingress domain.
(`getDefaultCertificateSecretForIngressController`): New function used by `desiredRouterCertsGlobalSecret`.
* `pkg/operator/controller/certificate-publisher/publish_certs_test.go` (`newSecret`): Delete the example PEM data and put the secret's name in the "tls.crt" and "tls.key" data fields of the returned secret so that the data fields differ for different secrets.
(`TestDesiredRouterCertsGlobalSecret`): Modify test cases so that they verify that the router-cert secret's data fields have the expected values.  Add a "default certificate, explicit" test case.  Add "custom certificate" and "missing custom certificate" test cases for the default ingresscontroller with a custom default certificate.  Add a "custom ingresscontroller with the cluster ingress domain" test case.  Change the "no secrets" test case to use the default ingresscontroller so that the test case would expect a router-cert secret but for the missing secret for the ingresscontroller. Update the "missing secret", "extra secret", and "perfect match" test cases not to expect the router-cert secret to include data for other ingresscontrollers.
* `test/e2e/all_test.go` (`TestAll`): Update the lists of tests.
* `test/e2e/certificate_publisher_test.go`: Delete file.  This deletes the `TestCreateIngressControllerThenSecret` and `TestCreateSecretThenIngressController` tests.
* `test/e2e/operator_test.go` (`TestUpdateDefaultIngressController`): Rename...
(`TestUpdateDefaultIngressControllerSecret`): ...to this.  Expand the test to verify that the operator updates both the router-certs secret as well as the default-ingress-cert configmap correctly, as well as verifying that the operator *does not* update the router-certs secret and the configmap if the ingresscontroller is updated to reference a non-existent secret.

#### `TestDesiredRouterCertsGlobalSecret`: Use `t.Run`

* `pkg/operator/controller/certificate-publisher/publish_certs_test.go` (`TestDesiredRouterCertsGlobalSecret`): Use `t.Run`.

#### Move `IsAdmitted` to new util package

Move the `isAdmitted` function from the ingress controller package to a new util package that can be imported into other controllers.

* `pkg/operator/controller/ingress/controller.go` (`isAdmitted`): Move from here...
* `pkg/util/ingresscontroller/ingresscontroller.go` (`IsAdmitted`): ...to here.  New file for IngressController-related util functions.

#### certificate-publisher: Ignore not-admitted ingresscontrollers

When determining the desired "router-certs" secret, ignore ingresscontrollers that haven't been admitted.  Multiple ingresscontrollers can all have the cluster ingress domain, but only one can be admitted at a given time.  Any ingresscontrollers that haven't been admitted should be ignored so that the default certificate of any one that has been admitted is published.

* `pkg/operator/controller/certificate-publisher/publish_certs.go` (`desiredRouterCertsGlobalSecret`): Ignore ingresscontrollers that haven't been admitted.
* `pkg/operator/controller/certificate-publisher/publish_certs_test.go` (`newIngressController`): Add an "admitted" parameter for specifying the status of the "Admitted" status condition.
(`TestDesiredRouterCertsGlobalSecret`): Add a test case for "default certificate and custom ingresscontroller that conflicts on domain".


Miciah:
`TestCanaryRoute` failed due to pod security.  I filed <https://issues.redhat.com/browse/OCPBUGS-1049> to track the issue and posted <https://github.com/openshift/cluster-ingress-operator/pull/826> to fix it.

Miciah:
#826 merged.
/test e2e-aws-operator

pkg/operator/controller/certificate-publisher/publish_certs_test.go
```diff
@@ -100,15 +58,24 @@ func TestDesiredRouterCertsGlobalSecret(t *testing.T) {
...
-		ci1 = newIngressController("ci1", "s1", "dom1")
-		ci2 = newIngressController("ci2", "s2", "dom2")
-		s1  = newSecret("s1")
-		s2  = newSecret("s2")
-		// data has the PEM for defaultCert, s1, and s2 (which all have
-		// the same certificate and key).
-		data = bytes.Join([][]byte{
-			s1.Data["tls.crt"],
-			s1.Data["tls.key"],
+		defaultCIExplicitDefaultDefaultCert = newIngressController("default", "router-certs-default", "apps.my.devcluster.openshift.com")
```
suleymanakbas91, line 61: I suppose this is a typo.
```suggestion
		defaultCIExplicitDefaultCert = newIngressController("default", "router-certs-default", "apps.my.devcluster.openshift.com")
```

test/e2e/operator_test.go
```diff
@@ -558,6 +662,27 @@ func TestUpdateDefaultIngressController(t *testing.T) {
 		t.Logf("secret content=%v", string(secret.Data["tls.crt"]))
 		t.Fatalf("failed to observe update of default ingress CA certificate configmap: %v\noriginal=%v\ncurrent=%v", err, previousDefaultIngressCAConfigmap.Data["ca-bundle.crt"], defaultIngressCAConfigmap.Data["ca-bundle.crt"])
 	}
+
+	// Verify that the router-certs secret gets updated.
+	previousRouterCertsSecret = routerCertsSecret.DeepCopy()
+	wait.PollImmediate(1*time.Second, 20*time.Second, func() (bool, error) {
```
suleymanakbas91, line 668: Timeout values look arbitrary (e.g. 10s, 15s, 20s). Would it make sense to use the same timeout values for consistency or to write a brief explanation of why a particular value is preferred?

suleymanakbas91 (changes requested):

pkg/operator/controller/certificate-publisher/publish_certs_test.go
Miciah: It actually isn't; it's just confusing.  * grin *.  I've added a comment to explain:
```
		// defaultCIExplicitDefaultDefaultCert is an ingresscontroller
		// named "default" that specifies an explicit reference for
		// default certificate secret, where that secret is the same
		// default one that the operator generates when none is
		// specified (the "default default certificate").
```
Is the comment sufficient, or would it still be less confusing if I renamed the variable?

test/e2e/operator_test.go
Miciah: There was probably some off-the-cuff reasoning about how long it might take the operator to update a resource, or assumptions that if resource A has been updated, then resource B should have been updated too.  However, in lieu of detailed explanations, I've updated the polling loops all to use `timeout` for consistency.  This increases the timeouts for some polling loops; at worst, it will take a little longer for the test to fail, and at best, this can make the test less flaky.

Miciah:
Rebased and addressed Suleyman's comments.

pkg/operator/controller/certificate-publisher/publish_certs_test.go
suleymanakbas91: It is clear now with the comment.

suleymanakbas91:
/lgtm 

Failures look unrelated..
/retest

pkg/operator/controller/certificate-publisher/publish_certs.go
```diff
@@ -56,55 +57,63 @@ func (r *reconciler) ensureRouterCertsGlobalSecret(secrets []corev1.Secret, ingr
...
+		// ingress.Spec.DefaultCertificate is updated to point to a
+		// non-existent secret, the certificate controller does not
+		// delete the operator-generated default certificate, so it is
+		// safe to fall back to using it.  See
+		// <https://bugzilla.redhat.com/show_bug.cgi?id=1887441>.
+		for i := range secrets {
+			// defaultCertName can be the same as customCertName, so
+			// both of the following conditions can be true; see
+			// <https://bugzilla.redhat.com/show_bug.cgi?id=1912922>.
+			if defaultCertName.Name == secrets[i].Name {
```
gcs278, line 89: I've stared at this for a while and still it's not really making sense. We didn't fix https://bugzilla.redhat.com/show_bug.cgi?id=1912922 here right? How'd did the old code handle this?

gcs278 (commented):
Overall looks good to me

pkg/operator/controller/certificate-publisher/controller.go
```diff
@@ -168,6 +166,26 @@ func (r *reconciler) secretChanged(old, new runtime.Object) bool {
...
+func (r *reconciler) hasClusterIngressDomain(o client.Object) bool {
+	var ingressConfig configv1.Ingress
+	if err := r.cache.Get(context.Background(), controller.IngressClusterConfigName(), &ingressConfig); err != nil {
+		log.Error(err, "failed to get ingress 'cluster'")
+		// Assume it might be a match.  Better to reconcile an extra
+		// time than to miss an update.
+		return true
+	}
+	ic := o.(*operatorv1.IngressController)
+	return ic.Status.Domain == ingressConfig.Spec.Domain
```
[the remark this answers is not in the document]
Miciah: When cluster-authentication-operator reads the secret, it uses (effectively) `secret.Data[ingressConfig.Spec.Domain]`: https://github.com/openshift/cluster-authentication-operator/blob/686e2897e0d6c4a0b2d68ef0cb8e28e56a6d190d/pkg/controllers/routercerts/controller.go#L137-L159
When cluster-ingress-operator publishes the secret, the old logic uses `ic.Status.Domain` as the key in the secret's data: https://github.com/openshift/cluster-ingress-operator/blob/362a3ec2a92f501477dff3ec71348c8150437709/pkg/operator/controller/certificate-publisher/publish_certs.go#L105
So we could normalize the key in the new logic, but not doing so doesn't break anything that would have worked before.

pkg/operator/controller/certificate-publisher/controller.go
```diff
@@ -100,14 +106,27 @@ func New(mgr manager.Manager, operatorNamespace, operandNamespace string) (runti
...
+		return requests
+	}
+	if err := r.cache.Get(context.Background(), controller.IngressClusterConfigName(), &ingressConfig); err != nil {
+		log.Error(err, "failed to get ingress 'cluster'")
 		return requests
 	}
-	for _, ic := range controllers {
-		log.Info("queueing ingresscontroller", "name", ic.Name, "related", o.GetSelfLink())
+	for _, ic := range list.Items {
+		if ic.Status.Domain != ingressConfig.Spec.Domain {
```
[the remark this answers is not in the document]
Miciah: The analysis in <https://github.com/openshift/cluster-ingress-operator/pull/824#discussion_r1006198353> applies here.  If the ingresscontroller's domain doesn't match the cluster ingress config's domain in a case-sensitive match, the end result is the same with either the old logic or the new logic (namely that cluster-authentication-operator won't find the certificate in the secret).

pkg/operator/controller/certificate-publisher/publish_certs.go
```diff
@@ -56,55 +57,63 @@ func (r *reconciler) ensureRouterCertsGlobalSecret(secrets []corev1.Secret, ingr
...
-	}
+func desiredRouterCertsGlobalSecret(secrets []corev1.Secret, ingresses []operatorv1.IngressController, operandNamespace, clusterIngressDomain string) (*corev1.Secret, error) {
+	for i := range ingresses {
+		// The authentication operator only requires the certificate and
+		// key for the ingresscontroller that handles the cluster
+		// ingress domain, and publishing certificates for all
+		// ingresscontrollers could cause the secret's size to exceed
+		// the maximum secret size of 1 mebibyte.  See
+		// <https://issues.redhat.com/browse/OCPBUGS-853>.
+		if ingresses[i].Status.Domain != clusterIngressDomain {
```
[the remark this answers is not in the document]
Miciah: The analysis in <https://github.com/openshift/cluster-ingress-operator/pull/824#discussion_r1006198353> applies here.

pkg/operator/controller/certificate-publisher/controller.go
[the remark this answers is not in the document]
gcs278: Got it, sounds good.

pkg/operator/controller/certificate-publisher/controller.go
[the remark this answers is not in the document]
gcs278: Got it, it's fine the way it is. Thanks for looking into this.

pkg/operator/controller/certificate-publisher/publish_certs.go
Miciah: Right, https://github.com/openshift/cluster-ingress-operator/pull/521/commits/00f17ef70bff89c1f9ad13e2c8c36968c3915be1 fixed https://bugzilla.redhat.com/show_bug.cgi?id=1912922 but didn't require changes in this code to do so.  In this code, the old logic was fine, but it is a little subtle, so I figured it would be helpful to add a comment to explain why I wrote the logic this way.  

The old logic went something like this:

1. Build `nameToSecret`.  
2. For each ingresscontroller, 
    a. If `ingress.Spec.DefaultCertificate` is set, add the secret named `ingress.Spec.DefaultCertificate` to `globalSecret`.  
    b. Otherwise add the secret named `controller.RouterOperatorGeneratedDefaultCertificateSecretName(&ingress, operandNamespace)` to `globalSecret`.  
3. Publish `globalSecret`.  

The new logic, because it only cares about the default ingresscontroller, inverts this logic:

1. For each secret, 
    a. If the secret's name matches `controller.RouterOperatorGeneratedDefaultCertificateSecretName(&ingress[i], operandNamespace)` where `&ingresses[i]` is the default ingresscontroller, set `cert` to the secret.
    b. If the secret's name matches `ingresses[i].Spec.DefaultCertificate` where `&ingresses[i]` is the default ingresscontroller, set `cert` to the secret`.
2. If `cert` is set, publish it in `globalSecret`.

Now suppose `controller.RouterOperatorGeneratedDefaultCertificateSecretName(&ingresses[i], operandNamespace)` is "router-certs-default" and `ingresses[i].Spec.DefaultCertificate` is "foo" and that both the "router-certs-default" secret and the "foo" secret exist.  Then the following logic could do the wrong thing:
```go
		for i := range secrets {
			if defaultCertName.Name == secrets[i].Name {
				cert = &secrets[i]
			}
			if customCertName != nil && customCertName.Name == secrets[i].Name {
				cert = &secrets[i]
			}
		}
```
The following test case (added to `TestDesiredRouterCertsGlobalSecret`) fails with the faulty logic:
```go
		{
			description: "custom certificate, with secrets having the custom one and then default one",
			inputs: testInputs{
				[]operatorv1.IngressController{defaultCICustomDefaultCert},
				[]corev1.Secret{customDefaultCert, defaultCert},
			},
			output: testOutputs{
				&corev1.Secret{
					Data: map[string][]byte{"apps.my.devcluster.openshift.com": customDefaultCertData},
				},
			},
		},
```
If `secrets` has "router-certs-default" after "foo", then `cert` would get assigned the wrong value.  

Having worked through this example, I'll add the aforementioned test case to the PR.  Does my explanation make the logic clearer?

gcs278: Yea thanks this helps - it was definitely subtle in the existing code. I think your code is a bit more verbose, which is good.

Test case makes sense.

Miciah: I went ahead and refactored the tricky logic into a new function, `getDefaultCertificateSecretForIngressController`, with more comments.

gcs278: The new variables names makes it MUCH more readable in my opinion, thanks!

gcs278:
Thanks for responses
/lgtm

Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1585398230055653376) because kube-apiserver, kube-controller-manager, kube-scheduler, and ovnkube-master had problems rolling out their pods.  Also, must-gather and deprovisioning failed.
/test e2e-aws-operator
/test e2e-azure-operator
/test e2e-gcp-operator

[e2e-gcp-ovn-serial failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-ovn-serial/1585398230252785664) because disruption tests failed:
```
{  ingress-to-oauth-server-new-connections was unreachable during disruption testing for at least 41s of 2h19m2s (maxAllowed=21s):

}
```
```
{  ingress-to-console-new-connections was unreachable during disruption testing for at least 49s of 2h19m2s (maxAllowed=30s):

}
```
/test e2e-gcp-ovn-serial

pkg/operator/controller/certificate-publisher/publish_certs_test.go
frobware: Maybe renaming the variable would further aid clarity because I first wondered whether the run-on of DefaultDefault was intentional. And CI is Custom Ingress? Or Cluster Ingress? Maybe we could choose names like "production", or "staging", or "dev"...?

pkg/operator/controller/certificate-publisher/publish_certs_test.go
```diff
@@ -139,55 +177,48 @@ func TestDesiredRouterCertsGlobalSecret(t *testing.T) {
...
 		},
 	}
 
 	for _, tc := range testCases {
 		expected := tc.output.secret
-		actual, err := desiredRouterCertsGlobalSecret(tc.inputs.secrets, tc.inputs.ingresses, "openshift-ingress")
+		actual, err := desiredRouterCertsGlobalSecret(tc.inputs.secrets, tc.inputs.ingresses, "openshift-ingress", "apps.my.devcluster.openshift.com")
 		if err != nil {
 			t.Errorf("failed to get desired router-ca global secret: %v", err)
 			continue
```
frobware, line 224: I know this line wasn't modified but I changed the overall structure to use subtests so that I could fiddle with the expected inputs/output and just invoke a particular test case to observe deliberation changes to the test conditions didn't allow the test to pass. But the change also helped convince me that the early `continue` statements didn't mask any other assertions that should/could have fired.

```suggestion
	for _, tc := range testCases {
		t.Run(tc.description, func(t *testing.T) {
			expected := tc.output.secret
			actual, err := desiredRouterCertsGlobalSecret(tc.inputs.secrets, tc.inputs.ingresses, "openshift-ingress", "apps.my.devcluster.openshift.com")
			if err != nil {
				t.Errorf("failed to get desired router-ca global secret: %v", err)
			}
			if expected == nil || actual == nil {
				if expected != nil {
					t.Errorf("expected %v, got nil", expected)
				}
				if actual != nil {
					t.Errorf("expected nil, got %v", actual)
				}
			}

			if expected != nil && actual != nil {
				if !routerCertsSecretsEqual(expected, actual) {
					t.Errorf("expected %v, got %v", expected, actual)
				}
			}
		})
	}
```

frobware (commented):
LGTM but I think the variable naming as mentioned in one of the comments would help.

pkg/operator/controller/certificate-publisher/controller.go
```diff
@@ -106,12 +112,20 @@ func (r *reconciler) secretToIngressController(o client.Object) []reconcile.Requ
...
 			"defaultCertificateName": o.GetName(),
 		})
+		ingressConfig configv1.Ingress
 	)
 	if err := r.cache.List(context.Background(), &list, listOpts); err != nil {
 		log.Error(err, "failed to list ingresscontrollers for secret", "secret", o.GetName())
 		return requests
 	}
+	if err := r.cache.Get(context.Background(), controller.IngressClusterConfigName(), &ingressConfig); err != nil {
+		log.Error(err, "failed to get ingress 'cluster'")
```
frobware, line 122: Nit: Use `controller.IngressClusterConfigName()` in lieu of the hard-coded 'cluster'?

pkg/operator/controller/certificate-publisher/publish_certs_test.go
Miciah: Are you suggesting that I change `TestDesiredRouterCertsGlobalSecret` to use `t.Run`?

Miciah: I went ahead and added a commit with your suggestion.

pkg/operator/controller/certificate-publisher/controller.go
Miciah: Updated.

pkg/operator/controller/certificate-publisher/publish_certs_test.go
Miciah: I've renamed the variables: `defaultCIWithDefaultCertUnspecified`, `defaultCIWithDefaultCertSetToDefault`, and `defaultCIWithDefaultCertSetToCustom`.

Miciah: Hmm, I suppose while I'm renaming variables, I can get rid of the archaic "CI" initialism...  Done.

Miciah:
[e2e-azure-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1585652785167208448) because a bunch of image pulls failed; [`pods.json`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1585652785167208448/artifacts/e2e-azure-operator/gather-extra/artifacts/pods.json) shows that the following pods fails to pull their images:
```
% jq < pods.json -r '.items|.[]|select(.status|.containerStatuses,.initContainerStatuses|.[]?|select(.state.waiting.reason=="ImagePullBackOff"))|.metadata.name'
oauth-openshift-59db778d9c-vt5rl
oauth-openshift-59db778d9c-z4sn6
console-operator-599b69fbd-fdt7v
console-operator-599b69fbd-fdt7v
node-ca-wcnff
certified-operators-5pxv8
community-operators-fpptm
redhat-operators-5d8f7
node-exporter-dml6g
node-exporter-t5cvq
```
/test e2e-azure-operator

Miciah:
/test all
because #823 merged.

Miciah:
[e2e-aws-ovn-single-node failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-single-node/1585713551698300928) because `[sig-arch][Late] operators should not create watch channels very often [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]` failed:
```
{  fail [github.com/openshift/origin/test/extended/apiserver/api_requests.go:446]: Expected
    <[]string | len:1, cap:1>: [
        "Operator \"kube-controller-manager-operator\" produces more watch requests than expected: watchrequestcount=294, upperbound=290, ratio=1.013793103448276",
    ]
to be empty
```

Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/824/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1585713551622803456) because kube-apiserver reported `NodeInstallerProgressing`, must-gather failed, and deprovisioning failed.  However, `TestUpdateDefaultIngressControllerSecret` passed in e2e-aws-operator, e2e-azure-operator, and e2e-gcp-operator.

Miciah:
because e2e-azure-operator and e2e-gcp-operator passed and e2e-aws-operator failed on known issues unrelated to the changes in this PR.
/override ci/prow/e2e-aws-ovn-single-node
because it failed on issues with kube-controller-manager-operator that are unrelated to the changes in this PR.
/override ci/prow/e2e-gcp-ovn-serial
because it failed on disruption issues that are unrelated to the changes in this PR.

PR #827, opened 2022-09-10
Enable relative resource names for gcp dns zones

jstuever:
Enable the gcp dns provider to use relative resource names for the managedZones. This allows the project to be specified as part of the zone ID and override the defalut cluster project when managing the dns records.

In addition to accepting the zone name as before, the zone ID can now accept the following format:

projects/project-id/managedZones/zone-name

https://issues.redhat.com/browse/CORS-2072

### DIFF
--- a/pkg/dns/gcp/provider.go
+++ b/pkg/dns/gcp/provider.go
@@ -2,7 +2,9 @@ package gcp
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"strings"
 
 	"google.golang.org/api/googleapi"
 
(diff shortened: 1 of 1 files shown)


sadasu:
This logic seems to occur several times:
project, zoneName, err := parseZone(zone.ID)
	if err != nil {
		return err
	}
	if project == "" {
		project = p.config.Project
	}

Could we add a function to emcalsulate this logic:
func getProjectAndZone(zone string) (string, string, error) {
    project, zoneName, err := parseZone(zone.ID)
	if err != nil {
		return err
	}
	if project == "" {
		project = p.config.Project
	}
        return project, zone, nil
}

That way if this logic changes in the future, we only need to change it in one place.

sadasu:
@miheer Could you please take a look?

PR #833, opened 2022-09-29
Bug 2107462: Update CGO_ENABLED=1

prb112:
I am updating with `CGO_ENABLED=1` per discussion/debugging of @miheer 

I've asked @miheer and @manojnkumar to review.

We are currently verifying with fips enabled on ppc64le.

Signed-off-by: Paul Bastide <pbastide@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ GO_GCFLAGS ?= -gcflags=all="-N -l"
 endif
 
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
-GO_BUILD_RECIPE=CGO_ENABLED=0 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
+GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
 
 TEST ?= TestAll
 
@@ -45,7 +45,7 @@ crd:
 
 .PHONY: test
 test:
-	$(GO) test ./...
+	CGO_ENABLED=1 $(GO) test ./...
 
 .PHONY: release-local
 release-local:
@@ -53,7 +53,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e:
-	$(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	CGO_ENABLED=1 $(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
 
 .PHONY: test-e2e-list
 test-e2e-list:


prb112:
Closing per @miheer

Makefile
```diff
@@ -10,7 +10,7 @@ GO_GCFLAGS ?= -gcflags=all="-N -l"
 endif
 
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
-GO_BUILD_RECIPE=CGO_ENABLED=0 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
+GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
```
miheer, line 13: I don't think we can simply generalize enabling of CGO for everything.
You need to add conditions to check if FIPS was enabled or not.
If FIPS is enabled then only we shall be enabling CGO.
We could check contents of /proc/sys/crypto/fips_enabled however there is a catch that may be the system building this image does not have this file set to 1.
@Miciah WDYT? Do you think just enabling CGO will be OK ?

miheer (changes requested):

Makefile
manojnkumar: Isn't FIPS enabled by the customer?  How can we check whether it is enabled during build?

prb112:
Reopened per Slack Thread

Makefile
miheer: > generalize enabling of CGO for e

May be we might need to involve the art team in this and see what they say on this.

miheer: > I don't think we can simply generalize enabling of CGO for everything.
> You need to add conditions to check if FIPS was enabled or not.
> If FIPS is enabled then only we shall be enabling CGO.
> We could check contents of /proc/sys/crypto/fips_enabled however there is a catch that may be the system building this image does not have this file set to 1.
> @Miciah WDYT? Do you think just enabling CGO will be OK ?

May be we might need to involve the art team in this and see what they say on this.

miheer: If the art team agrees on building on CGO enabled image for ingress operator then I think we might also need code changes in the where cvo will need manifests with the fips image if fips was set by the installer during installation.

miheer: > I don't think we can simply generalize enabling of CGO for everything.
> You need to add conditions to check if FIPS was enabled or not.
> If FIPS is enabled then only we shall be enabling CGO.
> We could check contents of /proc/sys/crypto/fips_enabled however there is a catch that may be the system building this image does not have this file set to 1.
> @Miciah WDYT? Do you think just enabling CGO will be OK ?

Ignore this as we don't need build system to have FIPS enabled

miheer: I think it will be very complicated to check if FIPS was enabled during installation and then have art build image using CGO enabled. I think this will require code changes in the art-build system if we were to pass a Dockerfile called Docker.cgoenabled which would then build a image with this Dockerfile. However it will also need to understand if FIPS was enabled or not.

miheer: @Miciah WDYT ? May be we shall use CGO=1 only ?

frobware: > I don't think we can simply generalize enabling of CGO for everything.

Why not? At worst you'll end up always using the system resolver instead of Go's builtin resovler.

miheer: > > I don't think we can simply generalize enabling of CGO for everything.
> 
> Why not? At worst you'll end up always using the system resolver instead of Go's builtin resovler.

Yeah I know about that. We discussed on this on slack.
https://coreos.slack.com/archives/CB95J6R4N/p1664504382385589?thread_ts=1663335999.495519&cid=CB95J6R4N

frobware: So go with CGO_ENABLED=1. The openshift-router doesn't do a vast amount of DNS lookups beyond the kube API. We don't build or run the tests with CGO_ENABLED=0: https://github.com/openshift/router/blob/master/Makefile#L39.

manojnkumar: Can this be merged as is?

prb112: *Follow-ups*

1. As-Is: (and per Manoj) Is this an acceptable fix? per @frobware it'll go to the local system for resolutions.
2. Should [openshift/router: Makefile](https://github.com/openshift/router/blob/master/Makefile#L26) be updated as well? (follow on from frobware's comment and focused on the build, not test)
3. `ci/prow/e2e-gcp-ovn-serial` reported test thresholds being exceeded. 33s of 2h4m28s (maxAllowed=17s) as did the others `ci/prow/e2e-azure-ovn` and `ci/prow/e2e-aws-ovn-single-node` some are flakes. I started a `/retest`
4. Follow up on multiple-image / CVO approach.

> If the art team agrees on building on CGO enabled image for ingress operator then I think we might also need code changes in the where cvo will need manifests with the fips image if fips was set by the installer during installation.

That's understandable, however, it creates a larger surface area for the builds/streams and hard to uniformly diagnose issues.
5. @miheer I think I caught up on the review thread. Do you think this is sufficient?

miheer: > So go with CGO_ENABLED=1. The openshift-router doesn't do a vast amount of DNS lookups beyond the kube API. We don't build or run the tests with CGO_ENABLED=0: https://github.com/openshift/router/blob/master/Makefile#L39.

@frobware can you approve this PR if you think everything is fine ?

Miciah:
/retitle Bug 2107462: Update CGO_ENABLED=1
Would you mind updating the commit message to reference <https://bugzilla.redhat.com/show_bug.cgi?id=2107462> and briefly describe the issue (namely that we are setting `CGO_ENABLED=1` in order to use OpenSSL for FIPS compliance)?

Makefile
frobware: > _Follow-ups_
> 
> 1. As-Is: (and per Manoj) Is this an acceptable fix? per @frobware it'll go to the local system for resolutions.
> 2. Should [openshift/router: Makefile](https://github.com/openshift/router/blob/master/Makefile#L26) be updated as well? (follow on from frobware's comment and focused on the build, not test)

Yes, but obviously a separate PR and not urgent. 

I quoted the wrong repo though; I went looking in the router repo to see how we build the tests to see if it was different to how they are built here in the cluster-ingress-operator repo. The answer is we don't use the `GO_BUILD_RECIPE` flags in either repo for the `test`, `test-e2e`, or `check` make targets. Note: the `check` target only occurs in the openshift-router repo.

As we plan to switch to building with `CGO_ENABLED=1` we should be consistent in how we build and run the tests so I'd update this PR to also ensure that the `test` and `test-e2e` also build with `CGO_ENABLED=1`.

> 3. `ci/prow/e2e-gcp-ovn-serial` reported test thresholds being exceeded. 33s of 2h4m28s (maxAllowed=17s) as did the others `ci/prow/e2e-azure-ovn` and `ci/prow/e2e-aws-ovn-single-node` some are flakes. I started a `/retest`
> 4. Follow up on multiple-image / CVO approach.
> 
> > If the art team agrees on building on CGO enabled image for ingress operator then I think we might also need code changes in the where cvo will need manifests with the fips image if fips was set by the installer during installation.
> 
> That's understandable, however, it creates a larger surface area for the builds/streams and hard to uniformly diagnose issues. 5. @miheer I think I caught up on the review thread. Do you think this is sufficient?

prb112:
> `CGO_ENABLED=1` in order to use OpenSSL for FIPS compliance

@Miciah changes made per request, please let me know if it's sufficient.

prb112:
/retest-required

Note, these are timeouts related to the tests

frobware:
> > `CGO_ENABLED=1` in order to use OpenSSL for FIPS compliance
> 
> @Miciah changes made per request, please let me know if it's sufficient.

Will we make changes so that the build of tests is consistent ala https://github.com/openshift/cluster-ingress-operator/pull/833#discussion_r985493108?

prb112:
> > > `CGO_ENABLED=1` in order to use OpenSSL for FIPS compliance
> > 
> > 
> > @Miciah changes made per request, please let me know if it's sufficient.
> 
> Will we make changes so that the build of tests is consistent ala [#833 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/833#discussion_r985493108)?

Apologies - I missed that comment (I think my page had a dirty cache). 
I'll do so.

Makefile
prb112: Router PR https://github.com/openshift/router/pull/423

prb112: Hi @frobware Further discussion with @Prashanth684 it may be best that 423 is a bug on its own.  Would you want me to raise a new bug in OCPBUGS?

frobware: If we don't do the change in openshift/router what breaks specifically for the router?

prb112: i am not sure @miheer thoughts?

prb112:
the e2e-aws-operator is very flaky

PR #839, opened 2022-10-14
OCPBUGS-2334: Added nil check for service object on load balancer scope change

gcs278:
Added nil check for service object which without causes the ingress operator to panic when using ingress.operator.openshift.io/auto-delete-load-balancer annotation and changing scope of loadbalancer from external to internal.

- `pkg/operator/controller/ingress/status.go`: Add nil check for `computeAllowedSourceRanges`
- `pkg/operator/controller/ingress/status_test.go`: Add unit test for `computeAllowedSourceRanges`

PR #842, opened 2022-10-14
OCPBUGS-2429: Fix TestReloadInterval min value test cases and race conditions

gcs278:
OCPBUGS-2429: Fix TestReloadInterval min value test cases to 1s and fix race condition in which it passes as a false positive by forcing the ingress operator to recreated the router deployment instead of waiting on the ingress operator to report Available.

- `test/e2e/reloadinterval_test.go`: Add deletion of router deployment to trigger reconcile and fix min value test cases

### DIFF
--- a/test/e2e/reloadinterval_test.go
+++ b/test/e2e/reloadinterval_test.go
@@ -73,9 +73,9 @@ func TestReloadInterval(t *testing.T) {
 
 		// set ReloadInterval to a value outside allowed bounds
 		// expect no error
-		{"set reloadInterval 200ns", metav1.Duration{Duration: 200 * time.Nanosecond}, "5s", true},
-		{"set reloadInterval 5us", metav1.Duration{Duration: 5 * time.Microsecond}, "5s", true},
-		{"set reloadInterval 500ms", metav1.Duration{Duration: 500 * time.Millisecond}, "5s", true},
+		{"set reloadInterval 200ns", metav1.Duration{Duration: 200 * time.Nanosecond}, "1s", true},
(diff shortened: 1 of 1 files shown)


Miciah:
/retest-required
now that #839 has merged.

test/e2e/reloadinterval_test.go
```diff
@@ -73,9 +73,9 @@ func TestReloadInterval(t *testing.T) {
 
 		// set ReloadInterval to a value outside allowed bounds
 		// expect no error
-		{"set reloadInterval 200ns", metav1.Duration{Duration: 200 * time.Nanosecond}, "5s", true},
-		{"set reloadInterval 5us", metav1.Duration{Duration: 5 * time.Microsecond}, "5s", true},
-		{"set reloadInterval 500ms", metav1.Duration{Duration: 500 * time.Millisecond}, "5s", true},
+		{"set reloadInterval 200ns", metav1.Duration{Duration: 200 * time.Nanosecond}, "1s", true},
+		{"set reloadInterval 5us", metav1.Duration{Duration: 5 * time.Microsecond}, "1s", true},
```
miheer, line 77: does setting to 1s  really help ? Won't  it cause load on the router ?

miheer:
The PR lgtm so /lgtm

test/e2e/reloadinterval_test.go
gcs278: Not exactly - it's the value the test is looking for in the router deployment ENV variable. So no load affects, just a comparison value that was wrong.

ShudiLi:
cancel /label qe-approve

PR #843, opened 2022-10-15
OCPBUGS-2435: TestRouterCompressionOperation: Nil-pointer fix

Miciah:
Fix a nil-pointer dereference in the `getHttpHeaders` helper for `TestRouterCompressionOperation`.

Follow-up to #679.

* `test/e2e/router_compression_test.go` (`getHttpHeaders`): Don't dereference the response value from `client.Do` if it returned a non-nil error value.

PR #845, opened 2022-10-18
OCPBUGS-2493: Fix TestUnmanagedDNSToManagedDNSInternal E2E test race conditions

gcs278:
OCPBUGS-2493: Fix TestUnmanagedDNSToManagedDNSInternalIngressController E2E flake

`test/e2e/unmanaged_dns_test.go`:
  - Fixed service deletion race condition by ensuring loadbalancer service changed
  - Fixed check for IsServiceInternal which was actually checking for internal
    when should have been checking for external
  - Fixed IsServiceInternal using outdated service object
 
`test/e2e/util_test.go`:
  - Add HTTP responce body close in waitForHTTPClientCondition (the main issue)
  - Increased timeout to 10 minutes for verifyInternalIngressController and verifyExternalIngressController
  - Added curl pod restart in case of failure to verifyInternalIngressController
  - Added wait for DNS Resolution in waitForHTTPClientCondition for debuggability

### DIFF
--- a/test/e2e/unmanaged_dns_test.go
+++ b/test/e2e/unmanaged_dns_test.go
@@ -5,17 +5,23 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // TODO: Remove this once this condition is added to all e2e test
@@ -186,6 +192,23 @@ func TestManagedDNSToUnmanagedDNSIngressController(t *testing.T) {
 func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 	t.Parallel()
 
+	if infraConfig.Status.PlatformStatus == nil {
+		t.Skip("test skipped on nil platform")
+	}
+	platform := infraConfig.Status.PlatformStatus.Type
+
+	supportedPlatforms := map[configv1.PlatformType]struct{}{
+		configv1.AlibabaCloudPlatformType: {},
+		configv1.AWSPlatformType:          {},
+		configv1.AzurePlatformType:        {},
+		configv1.GCPPlatformType:          {},
+		configv1.IBMCloudPlatformType:     {},
+		configv1.PowerVSPlatformType:      {},
+	}
+	if _, supported := supportedPlatforms[platform]; !supported {
+		t.Skipf("test skipped on platform %q", platform)
+	}
+
 	name := types.NamespacedName{Namespace: operatorNamespace, Name: "unmanaged-migrated-internal"}
 	ic := newLoadBalancerController(name, name.Name+"."+dnsConfig.Spec.BaseDomain)
 	ic.Spec.EndpointPublishingStrategy.LoadBalancer = &operatorv1.LoadBalancerStrategy{
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 		t.Fatalf("failed to update ingresscontroller %s: %v", name, err)
 	}
 
-	if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
-		t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+	var oldLoadBalancerStatus corev1.LoadBalancerStatus
+	lbService.Status.LoadBalancer.DeepCopyInto(&oldLoadBalancerStatus)
+
+	// Only delete the service on platforms that don't automatically update the service's scope.
+	switch platform {
+	case configv1.AlibabaCloudPlatformType, configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType:
+		if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
+			t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+		}
+	}
+
+	// Ensure the service's load-balancer status changes.
+	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+		lbService := &corev1.Service{}
+		if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), lbService); err != nil {
+			t.Logf("Get %q failed: %v, retrying ...", controller.LoadBalancerServiceName(ic), err)
+			return false, nil
+		}
+		if reflect.DeepEqual(lbService.Status.LoadBalancer, oldLoadBalancerStatus) {
+			t.Logf("Waiting for service %q to be updated", controller.LoadBalancerServiceName(ic))
+			return false, nil
+		} else if ingresscontroller.IsServiceInternal(lbService) {
+			// The service got updated, but is not external.
+			return true, fmt.Errorf("load balancer %s is internal but should be external", lbService.Name)
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("error updating the %q service: %v", controller.LoadBalancerServiceName(ic), err)
 	}
 
 	t.Logf("Waiting for stable conditions on ingresscontroller %s after dnsManagementPolicy=Managed", ic.Name)
@@ -249,10 +299,6 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 	// Wait for the load balancer and DNS to reach stable conditions.
 	if err := waitForIngressControllerCondition(t, kclient, 10*time.Minute, name, append(availableConditionsForIngressControllerWithLoadBalancer, operatorProgressingFalse)...); err != nil {
 		t.Fatalf("failed to observe expected conditions: %v", err)
-	}
-
-	if !ingresscontroller.IsServiceInternal(lbService) {
-		t.Fatalf("load balancer %s is internal but should be external", lbService.Name)
 	}
 
 	// Ensure DNSRecord CR is present.
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"testing"
@@ -83,6 +84,7 @@ func waitForHTTPClientCondition(t *testing.T, httpClient *http.Client, req *http
 	return wait.PollImmediate(interval, timeout, func() (done bool, err error) {
 		resp, err := httpClient.Do(req)
 		if err == nil {
+			defer resp.Body.Close()
 			return compareFunc(resp), nil
 		} else {
 			t.Logf("retrying client call due to: %+v", err)
@@ -127,6 +129,7 @@ func buildCurlPod(name, namespace, image, host, address string, extraArgs ...str
 			Namespace: namespace,
 		},
 		Spec: corev1.PodSpec{
+			TerminationGracePeriodSeconds: pointer.Int64(0),
 			Containers: []corev1.Container{
 				{
 					Name:    "curl",
@@ -472,17 +475,31 @@ func verifyExternalIngressController(t *testing.T, name types.NamespacedName, ho
 		}
 	}()
 
+	// If we have a DNS as an external IP address, make sure we can resolve it before moving on.
+	// This just limits the number of "could not resolve host" errors which can be confusing.
+	if net.ParseIP(address) == nil {
+		if err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+			_, err := net.LookupIP(address)
+			if err != nil {
+				t.Logf("waiting for loadbalancer domain %s to resolve...", address)
+				return false, nil
+			}
+			return true, nil
+		}); err != nil {
+			t.Fatalf("loadbalancer domain %s was unable to resolve:", address)
+		}
+	}
+
 	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s", address), nil)
 	if err != nil {
 		t.Fatalf("failed to build client request: %v", err)
 	}
-
 	// we use HOST header to map to the domain associated on the ingresscontroller.
 	// This ensures our http call is routed to the correct router.
 	req.Host = hostname
 
 	httpClient := http.Client{Timeout: 5 * time.Second}
-	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 10*time.Minute, func(r *http.Response) bool {
 		if r.StatusCode == http.StatusOK {
 			t.Logf("verified connectivity with workload with req %v and response %v", req.URL, r.StatusCode)
 			return true
@@ -545,20 +562,32 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
 		"--retry-delay", "20",
 		"--max-time", "10",
 	}
-	clientPod := buildCurlPod("curl-"+name.Name, echoRoute.Namespace, image, address, echoRoute.Spec.Host, extraArgs...)
+	clientPodName := types.NamespacedName{Namespace: name.Namespace, Name: "curl-" + name.Name}
+	clientPodSpec := buildCurlPod(clientPodName.Name, clientPodName.Namespace, image, address, echoRoute.Spec.Host, extraArgs...)
+	clientPod := clientPodSpec.DeepCopy()
 	if err := kclient.Create(context.TODO(), clientPod); err != nil {
-		t.Fatalf("failed to create pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
+		t.Fatalf("failed to create pod %q: %v", clientPodName, err)
 	}
 	defer func() {
 		if err := kclient.Delete(context.TODO(), clientPod); err != nil {
 			if errors.IsNotFound(err) {
 				return
 			}
-			t.Fatalf("failed to delete pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
+			t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
 		}
 	}()
 
-	err = wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+	var curlPodLogs string
+	err = wait.PollImmediate(10*time.Second, 10*time.Minute, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), clientPodName, clientPod); err != nil {
+			t.Logf("error getting client pod %q: %v, retrying...", clientPodName, err)
+			return false, nil
+		}
+		// First check if client curl pod is still starting or not running.
+		if clientPod.Status.Phase == corev1.PodPending {
+			t.Logf("waiting for client pod %q to start", clientPodName)
+			return false, nil
+		}
 		readCloser, err := client.CoreV1().Pods(clientPod.Namespace).GetLogs(clientPod.Name, &corev1.PodLogOptions{
 			Container: "curl",
 			Follow:    false,
@@ -573,16 +602,39 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
 				t.Errorf("failed to close reader for pod %s: %v", clientPod.Name, err)
 			}
 		}()
+		curlPodLogs = ""
 		for scanner.Scan() {
 			line := scanner.Text()
+			curlPodLogs += line + "\n"
 			if strings.Contains(line, "HTTP/1.0 200 OK") {
 				t.Logf("verified connectivity with workload with address: %s with response %s", address, line)
 				return true, nil
 			}
 		}
+		// If failed or succeeded, the pod is stopped, but didn't provide us 200 response, let's try again.
+		if clientPod.Status.Phase == corev1.PodFailed || clientPod.Status.Phase == corev1.PodSucceeded {
+			t.Logf("client pod %q has stopped...restarting. Curl Pod Logs:\n%s", clientPodName, curlPodLogs)
+			if err := kclient.Delete(context.TODO(), clientPod); err != nil && errors.IsNotFound(err) {
+				t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
+			}
+			// Wait for deletion to prevent a race condition. Use PollInfinite since we are already in a Poll.
+			wait.PollInfinite(5*time.Second, func() (bool, error) {
+				err = kclient.Get(context.TODO(), clientPodName, clientPod)
+				if !errors.IsNotFound(err) {
+					t.Logf("waiting for %q: to be deleted", clientPodName)
+					return false, nil
+				}
+				return true, nil
+			})
+			clientPod = clientPodSpec.DeepCopy()
+			if err := kclient.Create(context.TODO(), clientPod); err != nil {
+				t.Fatalf("failed to create pod %q: %v", clientPodName, err)
+			}
+			return false, nil
+		}
 		return false, nil
 	})
 	if err != nil {
-		t.Fatalf("failed to verify connectivity with workload with address: %s using internal curl client: %v", address, err)
+		t.Fatalf("failed to verify connectivity with workload with address: %s using internal curl client. Curl Pod Logs:\n%s", address, curlPodLogs)
 	}
 }


Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1582221833082507264) because `TestUnmanagedDNSToManagedDNSInternalIngressController` failed, but the failure is different now:
```
util_test.go:88: retrying client call due to: Get "http://a5d2a6a35aec849c99e6189f6a877f2e-492060313.us-west-2.elb.amazonaws.com/": EOF
    unmanaged_dns_test.go:294: failed to verify connectivity with workload with reqURL http://a5d2a6a35aec849c99e6189f6a877f2e-492060313.us-west-2.elb.amazonaws.com/ using external client: timed out waiting for the condition
```
We might just need a longer timeout on the polling loop; the scope change is known to take as long as ~6 minutes on AWS: https://bugzilla.redhat.com/show_bug.cgi?id=2034795

Miciah:
[e2e-azure-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1582227739467845632) too, with a similar failure:
```
util_test.go:88: retrying client call due to: Get "http://52.143.243.13/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
    unmanaged_dns_test.go:294: failed to verify connectivity with workload with reqURL http://52.143.243.13/ using external client: timed out waiting for the condition
```
Searching for `*.unmanaged-migrated-internal` in [the ingress-operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1582227739467845632/artifacts/e2e-azure-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-59d58f6775-xvktn_ingress-operator.log), it looks like the dnsrecord is created unmanaged with an internal IP address, as expected:
```
2022-10-18T05:50:54.927Z	INFO	operator.ingress_controller	ingress/controller.go:1036	created dnsrecord	{"dnsrecord": {"metadata":{"name":"unmanaged-migrated-internal-wildcard","namespace":"openshift-ingress-operator","uid":"6cf23620-0cfb-450e-9a15-a72c0d2f8b7c","resourceVersion":"44312","generation":1,"creationTimestamp":"2022-10-18T05:50:54Z","labels":{"ingresscontroller.operator.openshift.io/owning-ingresscontroller":"unmanaged-migrated-internal"},"ownerReferences":[{"apiVersion":"operator.openshift.io/v1","kind":"IngressController","name":"unmanaged-migrated-internal","uid":"2a6d5ac0-5722-45a2-bf7d-3306fd4afd8a","controller":true,"blockOwnerDeletion":true}],"finalizers":["operator.openshift.io/ingress-dns"],"managedFields":[{"manager":"ingress-operator","operation":"Update","apiVersion":"ingress.operator.openshift.io/v1","time":"2022-10-18T05:50:54Z","fieldsType":"FieldsV1","fieldsV1":{"f:metadata":{"f:finalizers":{".":{},"v:\"operator.openshift.io/ingress-dns\"":{}},"f:labels":{".":{},"f:ingresscontroller.operator.openshift.io/owning-ingresscontroller":{}},"f:ownerReferences":{".":{},"k:{\"uid\":\"2a6d5ac0-5722-45a2-bf7d-3306fd4afd8a\"}":{}}},"f:spec":{".":{},"f:dnsManagementPolicy":{},"f:dnsName":{},"f:recordTTL":{},"f:recordType":{},"f:targets":{}}}}]},"spec":{"dnsName":"*.unmanaged-migrated-internal.ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com.","targets":["10.0.128.7"],"recordType":"A","recordTTL":30,"dnsManagementPolicy":"Unmanaged"},"status":{}}}
2022-10-18T05:50:54.928Z	INFO	operator.dns_controller	controller/controller.go:121	reconciling	{"request": "openshift-ingress-operator/unmanaged-migrated-internal-wildcard"}
2022-10-18T05:50:54.945Z	INFO	operator.dns_controller	dns/controller.go:198	DNS record not published	{"record": {"dnsName":"*.unmanaged-migrated-internal.ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com.","targets":["10.0.128.7"],"recordType":"A","recordTTL":30,"dnsManagementPolicy":"Unmanaged"}}
```
Then it is updated to managed with a public IP address, again as expected:
```
2022-10-18T05:51:55.280Z	INFO	operator.dns	dns/controller.go:312	upserted DNS record	{"record": {"dnsName":"*.unmanaged-migrated-internal.ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com.","targets":["10.0.128.7"],"recordType":"A","recordTTL":30,"dnsManagementPolicy":"Managed"}, "zone": {"id":"/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-ikmtjxjn-04a70-8svq6-rg/providers/Microsoft.Network/privateDnsZones/ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com"}}
2022-10-18T05:51:55.280Z	INFO	operator.dns_controller	dns/controller.go:359	published DNS record to zone	{"record": {"dnsName":"*.unmanaged-migrated-internal.ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com.","targets":["10.0.128.7"],"recordType":"A","recordTTL":30,"dnsManagementPolicy":"Managed"}, "dnszone": {"id":"/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-ikmtjxjn-04a70-8svq6-rg/providers/Microsoft.Network/privateDnsZones/ci-op-ikmtjxjn-04a70.ci.azure.devcluster.openshift.com"}}
```
So it isn't obvious to me what is going wrong.

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,21 +243,40 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
...
+			t.Logf("Get %q failed: %v, retrying ...", controller.LoadBalancerServiceName(ic), err)
+			return false, nil
+		}
+		// Ensure the UID of the service changed.
+		if lbService.UID == oldUID {
+			t.Logf("Waiting for service %q to be deleted", controller.LoadBalancerServiceName(ic))
+			return false, nil
+		} else if ingresscontroller.IsServiceInternal(lbService) {
+			// The service got recreated, but is not external
+			t.Fatalf("load balancer %s is internal but should be external", lbService.Name)
```
frobware, line 264: Shouldn't this just return an error? We catch and report any error after the call to PollImmediate.

gcs278:
@Miciah must be another problem. Increased `verifyInternalIngressController` to 8 minutes to see if that helps.

I've never seen `EOF` or `Client.Timeout exceeded while awaiting headers` but they seem related. Do you believe this to be related to a load balances being initialized? Or maybe DNS for the external loadbalancer domain hasn't initialized?

gcs278:
I added `resp.body.Close()` and `req.Close = true` to util_test.go since I found https://bugzilla.redhat.com/show_bug.cgi?id=2037447 and some [old threads](https://stackoverflow.com/questions/17714494/golang-http-request-results-in-eof-errors-when-making-multiple-requests-successi) about closing HTTP connection before trying again.

gcs278:
Whoops I broke something..hang on

Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1582399591662227456), and the ingress clusteroperator is reporting the following:
```
{Operator progressing (Reconciling): ingresscontroller "sourcerangesstatus" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: LoadBalancerProgressing=True (OperandsProgressing: One or more managed resources are progressing: You have manually edited an operator-managed object. You must revert your modifications by removing the service.beta.kubernetes.io/load-balancer-source-ranges annotation on service "router-sourcerangesstatus". You can use the new AllowedSourceRanges API field on the ingresscontroller object to configure this setting instead.).  Operator progressing (Reconciling): ingresscontroller "sourcerangesstatus" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: LoadBalancerProgressing=True (OperandsProgressing: One or more managed resources are progressing: You have manually edited an operator-managed object. You must revert your modifications by removing the service.beta.kubernetes.io/load-balancer-source-ranges annotation on service "router-sourcerangesstatus". You can use the new AllowedSourceRanges API field on the ingresscontroller object to configure this setting instead.).}
```

test/e2e/unmanaged_dns_test.go
Miciah: Andy's point is that if you returned an error here from the polling loop, then the `if err != nil { t.Fatalf(...) }` immediately after the loop would still suffice to terminate the test.

gcs278:
> Also, `TestUnmanagedDNSToManagedDNSIngressController` had a nil-pointer dereference in `waitForHTTPClientCondition`. Probably `resp` was nil in `defer resp.Body.Close()`.

Yea sorry I thought I fixed it in the last push but forget to do a `git add`

test/e2e/util_test.go
```diff
@@ -545,20 +547,43 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
+			return false, nil
+		}
+		// First check if client curl pod is still starting or not running.
+		// If not running, the pod most likely exited, but didn't provide a 200 response, let's try again.
+		if clientPod.Status.Phase == corev1.PodPending {
+			t.Logf("waiting for client pod %q to start", clientPodName)
+			return false, nil
+		} else {
+			t.Logf("client pod %q has failed...restarting", clientPodName)
+			if err := kclient.Delete(context.TODO(), clientPod); err != nil && errors.IsNotFound(err) {
```
gcs278, line 579: this is a race condition that I'm working on fixing

gcs278: WHOOPS I uploaded my test code...

test/e2e/unmanaged_dns_test.go
gcs278: Ah sorry I thought you meant `t.Errorf` vs. `t.Fatalf`.
Will fix in next push.

gcs278:
Successful [e2e-aws-operator](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1582446445049417728) run

gcs278:
Failed [ci/prow/e2e-azure-operator ](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1582447058818699264)

Just never connected to external loadbalancer. I feel like 8 mins is a LONG time to not be ready.

gcs278:
Increasing to 10 mins and trying again, also added `verifyInternalIngressController` fixes which I think are really getting quite messy, so need some feedback.

Miciah:
For what it's worth, I have a cluster running OCP 4.12.0-0.nightly-2022-10-18-041406 on GCP, and the `TestUnmanagedDNSToManagedDNSInternalIngressController` test from this PR passes there when run in isolation.  I tried it twice.  The test finished both times in under 5 minutes.

Miciah:
On Azure, changing the scope and then immediately deleting the service seems to cause the problem.  If the test doesn't delete the service, it succeeds on Azure in my testing.  Try this on for size:
```patch
diff --git a/test/e2e/unmanaged_dns_test.go b/test/e2e/unmanaged_dns_test.go
index 5a8de68d..c06f2dd3 100644
--- a/test/e2e/unmanaged_dns_test.go
+++ b/test/e2e/unmanaged_dns_test.go
@@ -6,9 +6,11 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
@@ -190,6 +192,23 @@ func TestManagedDNSToUnmanagedDNSIngressController(t *testing.T) {
 func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 	t.Parallel()
 
+	if infraConfig.Status.PlatformStatus == nil {
+		t.Skip("test skipped on nil platform")
+	}
+	platform := infraConfig.Status.PlatformStatus.Type
+
+		configv1.AlibabaCloudPlatformType: {},
+		configv1.AWSPlatformType:          {},
+		configv1.AzurePlatformType:        {},
+		configv1.GCPPlatformType:          {},
+		configv1.IBMCloudPlatformType:     {},
+		configv1.PowerVSPlatformType:      {},
+	}
+	if _, supported := supportedPlatforms[platform]; !supported {
+		t.Skipf("test skipped on platform %q", platform)
+	}
+
 	name := types.NamespacedName{Namespace: operatorNamespace, Name: "unmanaged-migrated-internal"}
 	ic := newLoadBalancerController(name, name.Name+"."+dnsConfig.Spec.BaseDomain)
 	ic.Spec.EndpointPublishingStrategy.LoadBalancer = &operatorv1.LoadBalancerStrategy{
@@ -244,31 +263,33 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 		t.Fatalf("failed to update ingresscontroller %s: %v", name, err)
 	}
 
-	oldUID := lbService.UID
-	if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
-		t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+	var oldLoadBalancerStatus corev1.LoadBalancerStatus
+	lbService.Status.LoadBalancer.DeepCopyInto(&oldLoadBalancerStatus)
+	switch platform {
+	case configv1.AlibabaCloudPlatformType, configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType:
+		if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
+			t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+		}
 	}
 
-	// Ensure the service has been recreated
+	// Ensure the service's load-balancer status changes.
 	err := wait.PollImmediate(10*time.Second, 3*time.Minute, func() (bool, error) {
 		lbService := &corev1.Service{}
 		if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), lbService); err != nil {
 			t.Logf("Get %q failed: %v, retrying ...", controller.LoadBalancerServiceName(ic), err)
 			return false, nil
 		}
-		// Ensure the UID of the service changed.
-		if lbService.UID == oldUID {
-			t.Logf("Waiting for service %q to be deleted", controller.LoadBalancerServiceName(ic))
+		if reflect.DeepEqual(lbService.Status.LoadBalancer, oldLoadBalancerStatus) {
+			t.Logf("Waiting for service %q to be updated", controller.LoadBalancerServiceName(ic))
 			return false, nil
 		} else if ingresscontroller.IsServiceInternal(lbService) {
-			// The service got recreated, but is not external
+			// The service got updated, but is not external.
 			return true, fmt.Errorf("load balancer %s is internal but should be external", lbService.Name)
 		}
 		return true, nil
 	})
-
 	if err != nil {
-		t.Fatalf("error recreating the %q service: %v", controller.LoadBalancerServiceName(ic), err)
+		t.Fatalf("error updating service %q: %v\nold status: %v\ncurrent status: %v", controller.LoadBalancerServiceName(ic), err, oldLoadBalancerStatus, lbService.Status.LoadBalancer)
 	}
 
 	t.Logf("Waiting for stable conditions on ingresscontroller %s after dnsManagementPolicy=Managed", ic.Name)
```

gcs278:
Only getting `TestRouterCompressionOperation` failures, so we are passing.
/retest

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
...
 
-	if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
-		t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+	var oldLoadBalancerStatus corev1.LoadBalancerStatus
+	lbService.Status.LoadBalancer.DeepCopyInto(&oldLoadBalancerStatus)
+
+	// Only delete the service on platforms that don't automatically update the service's scope.
+	switch platform {
+	case configv1.AlibabaCloudPlatformType, configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType:
+		if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
```
frobware, line 272: Should we have multiple attempts here to mitigate transient failures? Somebody else updated the object, for example.

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
...
+	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+		lbService := &corev1.Service{}
+		if err := kclient.Get(context.TODO(), controller.LoadBalancerServiceName(ic), lbService); err != nil {
+			t.Logf("Get %q failed: %v, retrying ...", controller.LoadBalancerServiceName(ic), err)
+			return false, nil
+		}
+		if reflect.DeepEqual(lbService.Status.LoadBalancer, oldLoadBalancerStatus) {
+			t.Logf("Waiting for service %q to be updated", controller.LoadBalancerServiceName(ic))
+			return false, nil
+		} else if ingresscontroller.IsServiceInternal(lbService) {
```
frobware, line 287: Moving this check seems odd - can it magically flip between internal and external?

test/e2e/unmanaged_dns_test.go
```diff
@@ -186,6 +192,23 @@ func TestManagedDNSToUnmanagedDNSIngressController(t *testing.T) {
...
+		configv1.AlibabaCloudPlatformType: {},
+		configv1.AWSPlatformType:          {},
+		configv1.AzurePlatformType:        {},
+		configv1.GCPPlatformType:          {},
+		configv1.IBMCloudPlatformType:     {},
+		configv1.PowerVSPlatformType:      {},
+	}
+	if _, supported := supportedPlatforms[platform]; !supported {
+		t.Skipf("test skipped on platform %q", platform)
+	}
```
candita, line 210: I feel like over time, this will cause problems.  When a new platform is added, we will have to remember to add it here.  If there is a specific platform that we don't want to run this on, you could do this:

```suggestion
	unsupportedPlatforms := map[configv1.PlatformType]struct{}{
		configv1.someSpecificPlatformType: {},
	}
	if _, unsupported := unsupportedPlatforms[platform]; unsupported {
		t.Skipf("test skipped on platform %q", platform)
	}
```

test/e2e/util_test.go
```diff
@@ -545,20 +563,50 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
 		"--retry-delay", "20",
 		"--max-time", "10",
 	}
-	clientPod := buildCurlPod("curl-"+name.Name, echoRoute.Namespace, image, address, echoRoute.Spec.Host, extraArgs...)
+	clientPodName := types.NamespacedName{echoRoute.Namespace, "curl-" + name.Name}
```
frobware, line 566: Be explicit with fields. 

```suggestion
	clientPodName := types.NamespacedName{Namespace: echoRoute.Namespace, Name: "curl-" + name.Name}

```

test/e2e/util_test.go
```diff
@@ -80,9 +81,11 @@ func buildEchoPod(name, namespace string) *corev1.Pod {
 
 func waitForHTTPClientCondition(t *testing.T, httpClient *http.Client, req *http.Request, interval, timeout time.Duration, compareFunc func(*http.Response) bool) error {
 	t.Helper()
+	req.Close = true
```
frobware, line 84: This forces/changes the behaviour of this utility function. Let the caller decide and, if necessary, the caller can set the desired behaviour.

test/e2e/util_test.go
```diff
@@ -472,17 +476,31 @@ func verifyExternalIngressController(t *testing.T, name types.NamespacedName, ho
...
 	}
-
 	// we use HOST header to map to the domain associated on the ingresscontroller.
 	// This ensures our http call is routed to the correct router.
 	req.Host = hostname
 
 	httpClient := http.Client{Timeout: 5 * time.Second}
-	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 10*time.Minute, func(r *http.Response) bool {
 		if r.StatusCode == http.StatusOK {
```
frobware, line 504: I think we should be draining the response buffer, ala: 

```go
io.Copy(ioutil.Discard, r.Body)
```

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 		t.Fatalf("failed to update ingresscontroller %s: %v", name, err)
 	}
```
candita, line 265: Should this step wait for stability too?
```suggestion
// Wait for the load balancer and DNS to reach stable conditions.
	if err := waitForIngressControllerCondition(t, kclient, 10*time.Minute, name, append(availableConditionsForIngressControllerWithLoadBalancer, operatorProgressingFalse)...); err != nil {
		t.Fatalf("failed to observe expected conditions: %v", err)
	}
```

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
...
+	lbService.Status.LoadBalancer.DeepCopyInto(&oldLoadBalancerStatus)
+
+	// Only delete the service on platforms that don't automatically update the service's scope.
+	switch platform {
+	case configv1.AlibabaCloudPlatformType, configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType:
+		if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
+			t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+		}
+	}
+
```
candita, line 276: Perhaps a wait for stability here too?

test/e2e/util_test.go
```diff
@@ -573,8 +621,10 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
 				t.Errorf("failed to close reader for pod %s: %v", clientPod.Name, err)
 			}
 		}()
+		curlPodLogs = ""
```
frobware, line 624: Can this become the declaration?

If, after, processing the scanner loop you can't return as "HTTP/1.0 200 OK" was never matched, then print `curlPodLogs`. The output would correspond to each iteration through the polling loop. Right now we just accumulate and print in one fell swoop.

frobware (changes requested):

test/e2e/util_test.go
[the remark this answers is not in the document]
candita: If it's needed at all.

test/e2e/util_test.go
```diff
@@ -80,9 +81,11 @@ func buildEchoPod(name, namespace string) *corev1.Pod {
 
 func waitForHTTPClientCondition(t *testing.T, httpClient *http.Client, req *http.Request, interval, timeout time.Duration, compareFunc func(*http.Response) bool) error {
 	t.Helper()
+	req.Close = true
 	return wait.PollImmediate(interval, timeout, func() (done bool, err error) {
 		resp, err := httpClient.Do(req)
 		if err == nil {
+			defer resp.Body.Close()
```
candita, line 88: Good find!

test/e2e/util_test.go
```diff
@@ -472,17 +476,31 @@ func verifyExternalIngressController(t *testing.T, name types.NamespacedName, ho
 		}
 	}()
 
+	// If we have a DNS as an external IP address, make sure we can resolve it before moving on.
+	// This just limits the number of "could not resolve host" errors which can be confusing.
+	if net.ParseIP(address) == nil {
+		if err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
```
candita, line 482: That's a long wait in an e2e test, can we get away with 1m?

test/e2e/util_test.go
```diff
@@ -545,20 +563,50 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
+			return false, nil
+		} else if clientPod.Status.Phase == corev1.PodFailed {
+			// If not running, the pod most likely exited, but didn't provide a 200 response, let's try again.
+			t.Logf("client pod %q has failed...restarting", clientPodName)
+			if err := kclient.Delete(context.TODO(), clientPod); err != nil && errors.IsNotFound(err) {
+				t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
+			}
+			// Wait for deletion to prevent a race condition.
+			err = kclient.Get(context.TODO(), clientPodName, clientPod)
+			for !errors.IsNotFound(err) {
```
frobware, line 599: These inner loops should/could make use of wait.Poll* too.

test/e2e/util_test.go
```diff
@@ -472,17 +476,31 @@ func verifyExternalIngressController(t *testing.T, name types.NamespacedName, ho
...
 		t.Fatalf("failed to build client request: %v", err)
 	}
-
 	// we use HOST header to map to the domain associated on the ingresscontroller.
 	// This ensures our http call is routed to the correct router.
 	req.Host = hostname
 
 	httpClient := http.Client{Timeout: 5 * time.Second}
-	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 5*time.Minute, func(r *http.Response) bool {
+	err = waitForHTTPClientCondition(t, &httpClient, req, 10*time.Second, 10*time.Minute, func(r *http.Response) bool {
```
candita, line 503: What do we gain with doubling the timeout?  Does just adding 1m help?

test/e2e/util_test.go
```diff
@@ -545,20 +563,50 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
 				return
 			}
-			t.Fatalf("failed to delete pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
+			t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
 		}
 	}()
 
-	err = wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+	curlPodLogs := ""
+	err = wait.PollImmediate(10*time.Second, 10*time.Minute, func() (bool, error) {
```
candita, line 582: Ditto about doubling the timeout here.  Can we get away with just adding 1m?

gcs278:
So e2e-aws-operator failed on `TestUnmanagedDNSToManagedDNSInternalIngressController` but it appears there was a installation problem as ALL curl-type to loadbalancer failed (also `TestHstsPolicyWorks`, `TestForwardedHeaderPolicyAppend`, `TestHTTPHeaderBufferSize`...)
e2e-gcp-operator failed on fix for https://github.com/openshift/cluster-ingress-operator/pull/843

test/e2e/util_test.go
```diff
@@ -545,20 +563,50 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
+			t.Logf("client pod %q has failed...restarting", clientPodName)
+			if err := kclient.Delete(context.TODO(), clientPod); err != nil && errors.IsNotFound(err) {
+				t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
+			}
+			// Wait for deletion to prevent a race condition.
+			err = kclient.Get(context.TODO(), clientPodName, clientPod)
+			for !errors.IsNotFound(err) {
+				t.Logf("waiting for %q: to be deleted", clientPodName)
+				err = kclient.Get(context.TODO(), clientPodName, clientPod)
+				time.Sleep(5 * time.Second)
```
candita, line 602: Another `pollImmediate` with a 5s timeout is always better than a 5s sleep.

test/e2e/util_test.go
```diff
@@ -545,20 +563,50 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
+		if err := kclient.Get(context.TODO(), clientPodName, clientPod); err != nil {
+			t.Logf("error getting client pod %q: %v, retrying...", clientPodName, err)
+			return false, nil
+		}
+		// First check if client curl pod is still starting or not running.
+		if clientPod.Status.Phase == corev1.PodPending {
+			t.Logf("waiting for client pod %q to start", clientPodName)
+			return false, nil
+		} else if clientPod.Status.Phase == corev1.PodFailed {
+			// If not running, the pod most likely exited, but didn't provide a 200 response, let's try again.
```
candita, line 592: I'm not sure we need to restart.  Do we really gain a better test with a restart, or with some log that tells us why the pod failed?

gcs278: Yes it does make the test better. There is an issue in which the loadbalancer is still initializing and will return an EOF instead of a 200 return code for a small period of time. The curl pod will exit and not try again. This will flake the test.

This is protecting against a flake, not an real-life cluster-ingress-operator issue. I provide curl pod logs regardless if the fails even after a restart.

test/e2e/util_test.go
gcs278: So yes, 10 mins is pretty aggressive, but I've seen AWS, Azure, GCP take over 5 mins. Here's Miciah's answer to when I asked why it failed at 5 mins: https://github.com/openshift/cluster-ingress-operator/pull/845#issuecomment-1281846948. 

It needs to be at least 7 mins, since we've seen loadbalancers take 6 mins to initialize, but we've been having issues with very long initialization periods for loadbalancers in this test. I could reduce to 7 or 8 mins, but I don't really see the point if it introduces even more flakes. There isn't much to gain by reducing the wait time here, it's parallel. 

I'm hesitant, but I can go down a minute or two if you think there is value.

test/e2e/util_test.go
gcs278: It's not long considering DNS can take very well up to 5 mins to propagate for AWS. Keep in mind, these times are dependent on external processes, not something we have control over. I think we want to give platforms as much time as it needs.

test/e2e/unmanaged_dns_test.go
```diff
@@ -251,10 +301,6 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
 		t.Fatalf("failed to observe expected conditions: %v", err)
 	}
 
-	if !ingresscontroller.IsServiceInternal(lbService) {
```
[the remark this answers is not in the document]
gcs278: Yep, just in this location it was opposite of what it actually should have been. I checked other usages to make sure thinks weren't flipped.

test/e2e/util_test.go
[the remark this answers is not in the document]
gcs278: Yea I agree, I put this here mainly for testing, but I don't think it's needed. I'm going to remove and test again.

gcs278: Ack - removing as I was using it to test with and don't believe it is needed.

test/e2e/util_test.go
gcs278: Done

test/e2e/unmanaged_dns_test.go
gcs278: Yea I see your point, opt-out vs. opt-in. @Miciah wrote this, and I think he wanted to be explicit with an "opt-in" approach to this E2E test.

@Miciah what are your thoughts since you added this? What platforms are we really trying to skip here (Nutanix? Libvirt? Openstack?...) can we even run e2e-test on them?

test/e2e/unmanaged_dns_test.go
gcs278: We do wait for DNS stable conditions on [line 300](https://github.com/openshift/cluster-ingress-operator/pull/845/files/bcf092c41b7cd5d1e4b74197fe2d45cbda038d6e#diff-a660dcb9b4adcc19e854b3f294ec260efabed12bb55eb2f2b8fb03ff6eb92527R300) before we do our verification test.

test/e2e/unmanaged_dns_test.go
gcs278: This check was moved just because we needed an up-to-date version of `lbService`. It's in the loop since we have an up-to-date `lbService`. Flipping isn't a concern here, if it detects the lbService was updated, then it checks if it's still internal only once then exits (it's not checking internal/external repeatedly).

This could be moved outside the loop, but just requires another `kclient.Get` prior to checking, so I just threw it in the loop to be more efficient. If you feel being outside of the loop would be clearer, let me know.

test/e2e/unmanaged_dns_test.go
gcs278: Hmm, that's a good, but loaded question. It's never been an issue in the multiple times I have tried it. Part of me agrees with you, but the other part says "Well, if we do it here, we need to rewrite a ton of other Gets/Deletes/Updates in this e2e-test and others" as this doesn't just apply here.

I think we need standardized common wrapper functions for Delete/Get/Update if we want to mitigate this, opposed as doing it as 1-offs everywhere. Could have a standard "deleteAndRetry" function (I think we have updateAndRetry-type functions already?).

I think probably a good shift-week-type refactoring, I'm hesitant to start doing 1-off-type refactoring since it's in many more than this one place here, and I think the effort should be coordinated. Let me know if you disagree, or have an existing function I can use.

test/e2e/util_test.go
gcs278: Yup I agree, but `resp.Body.close()` does that for us if we don't do it ourselves: https://groups.google.com/g/golang-nuts/c/4Rr8BYVKrAI/m/HRY-N-8_orcJ

`resp.Body.close()` was added in `waitForHTTPClientCondition`.

test/e2e/unmanaged_dns_test.go
Miciah: We're trying to skip platforms that don't use or support LoadBalancer-type services.  We have the same or similar logic repeated in several tests.  This should probably be refactored as part of a general cleanup of how we handle platform-specific skips.

test/e2e/unmanaged_dns_test.go
Miciah: +1 to what Grant said.  We've added retries to E2E tests _ad hoc_ in the past, and absent any _hoc_, it's reasonable to defer that to some more comprehensive effort to add retries to Kube API calls in E2E tests.

test/e2e/util_test.go
gcs278: Sure okay, I debated for awhile, but I don't think adding another timebox within a timebox to further artificially restrict certain parts of the code is beneficial (I think 1 timebox is enough), but if I switched it to a `pollInfinite` (noting how you said `wait.Poll*` so hopefully that addresses your concern with style, even though it's effectively the same thing.

I confirmed the outer poll will kill the inner poll if it exceeds the outer poll's deadline.

test/e2e/util_test.go
gcs278: Andy said something similar, hopefully this helps address what you mean: https://github.com/openshift/cluster-ingress-operator/pull/845/files/bcf092c41b7cd5d1e4b74197fe2d45cbda038d6e#r999840724.

test/e2e/unmanaged_dns_test.go
```diff
@@ -240,8 +263,35 @@ func TestUnmanagedDNSToManagedDNSInternalIngressController(t *testing.T) {
...
-		t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+	var oldLoadBalancerStatus corev1.LoadBalancerStatus
+	lbService.Status.LoadBalancer.DeepCopyInto(&oldLoadBalancerStatus)
+
+	// Only delete the service on platforms that don't automatically update the service's scope.
+	switch platform {
+	case configv1.AlibabaCloudPlatformType, configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType:
+		if err := kclient.Delete(context.TODO(), lbService); err != nil && !errors.IsNotFound(err) {
+			t.Fatalf("failed to delete svc %s: %v", lbService.Name, err)
+		}
```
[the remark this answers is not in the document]
Miciah: There's already a wait after the polling loop.  That should be enough:

1. Verify the service reports a different LB endpoint and is now internal.  
2. Verify the operator doesn't report that it is still updating the configuration.  
3. Verify the dnsrecord now reports "Managed".  
4. Verify we can connect to the LB at its updated address.  

Is there a need to repeat one of those checks, or is there some missing check?

test/e2e/util_test.go
gcs278: > Can this become the declaration?

You mean `var curlPodLogs string`? Done.

> If, after processing the scanner loop, you can't return as "HTTP/1.0 200 OK" was never matched, then print curlPodLogs. The output would correspond to each iteration through the polling loop. Right now we just accumulate and print in one fell swoop.

So yea I see what your saying, but this happens a lot where 200 OK isn't matched after each scanner loop (usually DNS issues...). I think if we print every time, we are gonna clog up the logs with the same repeated failures. Keep in mind, it's not additive: it scans the whole entire log each poll loop. So, with what you are suggesting, I believe you'd get something like this:
```
log line 1
log line 1
log line 2
log line 1
log line 2
log line 3
log line 1
log line 2
log line 3
log line 4
success!
```
I just chose to store the last log scan and only output on failures to keep the logs clean. Let me know if this makes sense or if I'm not seeing what you are saying.

test/e2e/unmanaged_dns_test.go
gcs278: @Miciah Yea I think your point is that this is a larger issue than just this single E2E test. Should we just remove this logic for the time being in anticipation for a more comprehensive effort? (kind of the same debate with Andy's comment on retries https://github.com/openshift/cluster-ingress-operator/pull/845/files#r999688986)

gcs278:
All E2E tests (I care about in this PR) passed! Round 4 I think. minus other issues (RouteCompression NilPointer issue and must gather...)

test/e2e/util_test.go
gcs278: Just for posterity, here's what it currently looks like (I intentionally broke it with an invalid IP):
```
    util_test.go:637: failed to verify connectivity with workload with address: 10.0.128.15 using internal curl client. Curl Pod Logs:
        * Rebuilt URL to: http://test.10.0.128.15/
        * Could not resolve host: test.10.0.128.15
        * Closing connection 0
        * Rebuilt URL to: http://test.10.0.128.15/
        * Could not resolve host: test.10.0.128.15
        * Closing connection 1
        * Rebuilt URL to: http://test.10.0.128.15/
        * Could not resolve host: test.10.0.128.15
        * Closing connection 2
    panic.go:482: deleted ingresscontroller unmanaged-migrated-internal
--- FAIL: TestUnmanagedDNSToManagedDNSInternalIngressController (173.51s)
FAIL
FAIL	github.com/openshift/cluster-ingress-operator/test/e2e	176.610s
FAIL
make: *** [Makefile:56: test-e2e] Error 1

```

test/e2e/unmanaged_dns_test.go
Miciah: I don't see a reason to remove this logic—it is strictly better to skip this test on platforms on which we know the test would fail than to run the test and fail on those platforms.

test/e2e/util_test.go
candita: >We might just need a longer timeout on the polling loop; the scope change is known to take as long as ~6 minutes on AWS...

I think we should take a minute to consider changes that add 10-15 minutes to a single test.  We're already waiting up to 3 hrs on some tests due to things we can't control.  

Are customers willing to wait up to 10 minutes for a loadbalancer to initialize?  If not, we need to find out why this activity fails to perform on time, not just give it more time to pass CI.  Just my opinion.

test/e2e/util_test.go
candita: That link doesn't work for me...

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
gcs278: Oh sorry I commented over here https://github.com/openshift/cluster-ingress-operator/pull/845/files#r999787438, I think the comments said the same thing.

test/e2e/util_test.go
candita: Why not just add a little bit of code to wait until the loadbalancer is stable?

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
candita: You mean, it was removed because it is erroneous?

test/e2e/unmanaged_dns_test.go
candita: Before I commented, I searched for uses of this pattern, and I didn't find any in c-i-o, except in the case where we knew we only wanted to run the test on a single platform.  https://github.com/openshift/cluster-ingress-operator/search?q=supportedplatforms

test/e2e/unmanaged_dns_test.go
candita: I've seen (and written) e2e tests where whenever there's an operation that has side effects, you call `waitForIngressControllerCondition`  Here, since he called `updateIngressControllerSpecWithRetryOnConflict`, it might be beneficial to call `waitForIngressControllerCondition`.

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
candita: If it doesn't matter whether the service was actually deleted or not, and it has no side effects, then you don't need to wait.

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
gcs278: Sorry I misunderstood, yes, this was definitely incorrect, it was fixed and moved above with other changes, https://github.com/openshift/cluster-ingress-operator/pull/845/files#diff-a660dcb9b4adcc19e854b3f294ec260efabed12bb55eb2f2b8fb03ff6eb92527R287

test/e2e/unmanaged_dns_test.go
gcs278: I agree, but we do exactly `waitForIngressControllerCondition` on [line 300](https://github.com/openshift/cluster-ingress-operator/pull/845/files/bcf092c41b7cd5d1e4b74197fe2d45cbda038d6e#diff-a660dcb9b4adcc19e854b3f294ec260efabed12bb55eb2f2b8fb03ff6eb92527R300) after this update operation.

It's not immediately below the IC update because we can go do some other things (like check for lbService update) before we `waitForIngressControllerCondition`, being more efficient, but it still comes after the update like you suggest to make sure we don't have any side effects.

Is that sufficient?

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
gcs278: Oh sorry, this comment was different, ignore my comment above.

We do wait on [Line 278](https://github.com/openshift/cluster-ingress-operator/pull/845/files#diff-a660dcb9b4adcc19e854b3f294ec260efabed12bb55eb2f2b8fb03ff6eb92527R278) to make sure the service was updated. We don't need to specifically wait for the service to be deleted, since we just want to trigger an update, and on some platform that's how you need to do it.

test/e2e/unmanaged_dns_test.go
Miciah: GitHub's search is unreliable.  The pattern appears in several tests.

- `TestAllowedSourceRanges`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/allowed_source_ranges_test.go#L28-L41
- `TestAllowedSourceRangesStatus`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/allowed_source_ranges_test.go#L127-L140
- `TestSourceRangesProgressingAndEvaluationConditionsDetectedStatuses`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/allowed_source_ranges_test.go#L246-L259
- `TestInternalLoadBalancer`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/operator_test.go#L935-L954
- `TestInternalLoadBalancerGlobalAccessGCP`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/operator_test.go#L1029-L1039
- `TestScopeChange`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/operator_test.go#L1299-L1315
- `TestLocalWithFallbackOverrideForLoadBalancerService`: https://github.com/openshift/cluster-ingress-operator/blob/9456d9fade4bf9afab26a26e089b6c44a51bc444/test/e2e/operator_test.go#L2879-L2890

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
Miciah: It still isn't clear to me whether there is a missing wait, or an extra wait.  If there is a missing wait, it isn't clear to me what it would be or why the test is passing now.  If there is an extra wait, I'm not too bothered by it since the test is passing now.  

Remember, the goal here is to get CI working again ASAP so that we can be confident that we haven't broken anything, and have time to fix anything that is broken before branching day; barring a major problem with the test (such as breaking it for some job or changing it in a way such that it doesn't cover the intended functionality at all), I'd rather not delay it for issues that can be addressed in a future clean-up.

test/e2e/util_test.go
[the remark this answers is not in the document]
gcs278: Removed

test/e2e/util_test.go
gcs278: Oh I probably linked on a specific changed, and updated that so it disappeared: 
https://github.com/openshift/cluster-ingress-operator/pull/845#discussion_r999732895

test/e2e/util_test.go
gcs278: Oh and I should also mention, I'm planning to redo this logic on shift-week :crossed_fingers:, like @Miciah suggested.

This logic is gnarly and has edge cases I'm trying to address here, but a refactoring hopefully should make this look a lot better.

test/e2e/util_test.go
Miciah: > I think we should take a minute to consider changes that add 10-15 minutes to a single test. We're already waiting up to 3 hrs on some tests due to things we can't control.

This only increases the job runtime if (a) this is the longest running test and (b) the polling loop times out.  If the LB is ready after 2 minutes, then it makes no different whether we specify a timeout of 5 minutes or 10 minutes.  If the LB takes 6 minutes to be ready (as is [known to be possible on AWS](https://bugzilla.redhat.com/show_bug.cgi?id=2034795)), then setting the timeout <10 minutes can increase frequency of flakes, and then you get to wait for the whole E2E job to rerun.  

> Are customers willing to wait up to 10 minutes for a loadbalancer to initialize? If not, we need to find out why this activity fails to perform on time, not just give it more time to pass CI. Just my opinion.

If there's a regression, it should be tracked as a separate bug.  As Grant said, optimizing LB initialization is out of scope for this PR.  The goal here is to get CI working again ASAP.

gcs278:
One last push, but e2e-aws-operator was successful (round 5 I think).

Small change, just wanted to ensure we restart the curl client pod regardless if it failed or successfully exited while we still didn't get a 200 response code. Just keep trying.

test/e2e/util_test.go
Miciah: > Why not just add a little bit of code to wait until the loadbalancer is stable before creating the pod?

We already have a `waitForIngressControllerCondition` before the call to `verifyInternalIngressController`.  What code do you have in mind?

test/e2e/unmanaged_dns_test.go
candita: It's a toss up on whether you wait for things to come back to normal after a change, or you wait and redo things over and over in a loop, adding time when you find you need to tweak the test.

test/e2e/unmanaged_dns_test.go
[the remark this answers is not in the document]
candita: As an interim, lgtm.

Miciah:
[e2e-gcp-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1582878615115141120) because `TestRouterCompressionOperation` failed (which #843 should fix); `TestUnmanagedDNSToManagedDNSInternal` passed.  :tada:!

Miciah:
[e2e-azure-operator passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1582878615077392384).  🎉 !

gcs278:
e2e-aws-operator has successful e2e but failed on `e2e-aws-operator-ipi-deprovision-deprovision`. other issues look to be install issues.
/retest
/test e2e-azure-operator
/test e2e-gcp-operator

Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1582917888962990080) because etcd failed to come up.
/test e2e-aws-operator

[e2e-gcp-ovn-serial failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-ovn-serial/1582878314899443712) because the disruption/ingress-to-oauth-server connection/new and ingress-to-console connection/new disruption tests failed.
/test e2e-gcp-ovn-serial

Miciah:
[e2e-gcp-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1582917889063653376) on both `TestRouterCompressionOperation` and `TestInternalLoadBalancer`.  The latter is interesting.  

The test output shows that `TestInternalLoadBalancer` timed out waiting for ingresscontroller to become available before the load balancer became ready.  From [the ingress-operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1582917889063653376/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-6db7bd4dd4-dg9db_ingress-operator.log), I see the following timeline:

- At 2022-10-20T03:10:19.895Z, the operator created the service.  
- At 2022-10-20T03:14:34.921Z, the service was still pending.  
- At 2022-10-20T03:15:19.125Z, the ingresscontroller was deleted, presumably because the test timed out.  

[The kube-controller-manager logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1582917889063653376/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-kube-controller-manager_kube-controller-manager-ci-op-7909m04i-76b3b-dstgx-master-0_kube-controller-manager.log) show that k-c-m ensured the LB at 03:14:34.646931.  

Strangely, I cannot find the create for that particular service in [the kube-apiserver access logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1582917889063653376/artifacts/e2e-gcp-operator/gather-extra/artifacts/audit_logs/kube-apiserver/), even though I can see various get and patch requests for that service and creates for other services that the operator created.  

My guess is that kube-apiserver or k-c-m was overloaded, and the create in kube-apiserver, the watch in k-c-m, or the provisioning of the LB in GCP was delayed, causing `TestInternalLoadBalancer` to hit its 5-minute timeout.  If this isn't a one-off flake, we might need to reduce the parallelism of the tests or increase timeouts where provisioning load balancers is involved.

gcs278:
[e2e-aws-operator](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1583077608508100608): `e2e-aws-operator-ipi-deprovision-deprovision container test` failure again with other `Operator progressing (NodeInstaller): NodeInstallerProgressing` issues.

/retest-required

gcs278:
[e2e-gcp-ovn-serial](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-ovn-serial/1583077608596180992) failed on known distruption issues
/retest-required

Miciah:
Looks like [e2e-aws-operator is going to fail](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/845/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1583122038648737792) because the e2e-aws-operator-gather-must-gather step failed and because `kube-apiserver`, `kube-controller-manager`, and `kube-scheduler` reported `NodeInstallerProgressing`.  The E2E tests all passed.  

PR #779, opened 2022-06-13
[WIP] NE-682: Added support for exporting Route type and Routes per Shard metrics

arkadeepsen:
This PR gathers and exports metrics for Route Type and Routes per Shard. @TrilokGeer

PR #849, opened 2022-10-25
[WIP] OCPBUGS-1689: Clear route admission status on route and namespace labels change

gcs278:
OCPBUGS-1689: Add the route-status control loop for clearing route admitted status when labels change on routes and namespaces. Refactored route_status.go and functions to be organized under the route-status directory. Added unit testing for route_status.go with the addition of the fake_client.go to simplify future creation of fake clients for unit testing.

- `pkg/operator/controller/ingress/controller.go`: Moving functions from/to route_status.go.
- `pkg/operator/controller/route-metrics/controller.go`: De-duplicate logic between route-status and route-metric.
- `pkg/operator/controller/route-status/controller.go`: New control loop that reconciles route status if a route or namespace label changes.
- `pkg/operator/controller/route-status/controller_test.go`: Unit testing for the new control loop using fake client.
- `pkg/operator/controller/route-status/router_status.go`: Moved from pkg/operator/controller/ingress to be co-located with route status control loop.
- `pkg/operator/controller/route-status/router_status_test.go`: New route_status.go unit testing using fake clients.
- `pkg/operator/operator.go`: Start the new route_status control loop and share the cache with the route_metric control loop.
- `test/e2e/router_status_test.go`: E2E test for route status control loop.
- `test/unit/fake_client.go`: Added common fake client initialization functions.
- `test/unit/utils.go`: Added IngressController, Namespace, and Route builder functions to for reusable helper functions.


pkg/operator/controller/route-status/router_status.go
```diff
@@ -0,0 +1,254 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+)
+
+// A brief overview of the operator's interactions with route status:
```
gcs278, line 22: todo update

gcs278:
TODO: Clean up some comments and write an E2E test

gcs278:
`route_status.go` shows as a new file, but I specifically created a commit to isolate the rename and moving. Maybe because more than 50% changes. So I would highly recommend reviewing the[ second commit](https://github.com/openshift/cluster-ingress-operator/pull/849/commits/9b79503bc83a9d2d8a5caec4418a7e23eeb0a17d) (link may break in future) to see changes to `route_status.go`

gcs278:
CI failed to update on my last push
/retest

pkg/operator/controller/route-metrics/controller.go
```diff
@@ -172,55 +163,26 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	// Variable to store the number of routes admitted by the Shard (Ingress Controller).
 	routesAdmitted := 0
 
 	// Iterate through the list Routes.
-	for _, route := range routeList.Items {
+	for _, route := range routesInShard.Items {
 		// Check if the Route's Namespace matches one of the Namespaces in the set namespacesSet and
 		// the Route is admitted by the Ingress Controller.
-		if namespacesSet.Has(route.Namespace) && routeStatusAdmitted(route, ingressController.Name) {
+		if namespacesInShard.Has(route.Namespace) {
```
arkadeepsen, line 185: I think you missed out calling the `IsRouteStatusAdmitted` function here.
```suggestion
		if namespacesInShard.Has(route.Namespace) && routestatus.IsRouteStatusAdmitted(route, ingressController.Name) {
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,714 @@
...
+			if err := client.Get(context.Background(), actualRouteName, &actualRoute); err != nil {
+				t.Errorf("error retrieving route from client: %v", err)
+			}
+			assert.Equal(t, tc.expectedRoute.Status, actualRoute.Status, "route name", tc.expectedRoute.Name)
+		})
+	}
+}
+
+// Test_RouteStatusAdmitted verifies that IsRouteStatusAdmitted behaves correctly.
+func Test_RouteStatusAdmitted(t *testing.T) {
```
arkadeepsen, line 530:
```suggestion
func Test_IsRouteStatusAdmitted(t *testing.T) {
```

arkadeepsen (commented):
I was just browsing through the PR. Found these issues, so leaving these comments here :slightly_smiling_face:

pkg/operator/controller/route-metrics/controller.go
gcs278: Thanks for the review (we were going to actually ping you to review this).

Yea I debated this, dropped it intentionally, and wanted to have a discussion. I think our options are:
1. Just check route/namespace label matching (my current revision)
2. Check route/namespace label matching AND route status (current master)
3. Check just route status

I forgot about option 3 until now - since we are actually fixing route status in the PR, technically we could just call `IsRouteStatusAdmitted` solely, without doing ANY lookups, which I think is much more efficient, but doesn't protect us against bugs in the future (not sure if that's a valid argument though, bugs should be fixed, and other things rely solely on route status like external dns).

I'm leaning toward option 3, let me know what you think or if I forgot anything.

pkg/operator/controller/route-status/route_status_test.go
gcs278: Ah good catch, thanks

pkg/operator/controller/route-metrics/controller.go
arkadeepsen: > 1. Just check route/namespace label matching (my current revision)

Is the route guaranteed to be admitted (immediately) by the Ingress if the route/namespace label match? If so, then checking the route status would not be required. If not (due to delay, misconfiguration, failure, etc.), then route metrics would be incorrect. I am not sure about how route admission works, so leaving this open.

> 2. Check route/namespace label matching AND route status (current master)

Out of the 3, this option as of now looks optimal to me as we are filtering out the route first and then also checking the status to ensure the route metrics are in sync.

> 3. Check just route status

Are all the routes going to be fetched in this scenario and then for each route the `IsRouteStatusAdmitted` function will be called? If so, then aren't we adding a lot of additional work to check the status of the routes which could be easily filtered out by the labelselectors? This would add a lot of load to ingress operator if a cluster has a large number of routes.

Let me know if I have understood the 3 scenarios correctly.

arkadeepsen: I kind of approximated the time complexity of both the approaches: 1) With namespace filtering, 2) Without namespace filtering. Not sure if the assumptions are correct, but here it goes.

1\) Let's assume that there are `n` namespaces filtered out by the namespace labelselector. Assuming that fetching the namespaces from the k8s api is dependent on the number of namespaces being fetched, then the time complexity will be `O(n)`. After the namespaces are fetched, each namespace is iterated over to construct the Set, the time complexity of which is again `O(n)`. Now, let there be `p+q` routes fetched using the `routeSelector`. `p` routes don't match the `namespaceSelector` and `q` routes match it. Let's assume that for each route there are `r` ingresses in the status. Then, following is the time complexity of the below code-snippet:
```go
// Iterate through the list Routes.
for _, route := range routesInShard.Items {
// Check if the Route's Namespace matches one of the Namespaces in the set namespacesSet and
// the Route is admitted by the Ingress Controller.
if namespacesInShard.Has(route.Namespace) && routestatus.IsRouteStatusAdmitted(route, ingressController.Name) {
	// If the Route is admitted then, the routesAdmitted should be incremented by 1 for the Shard.
	routesAdmitted++
}
```
The namespace filtering will be done for all routes, so the time complexity will be `O(p+q)` as the implementation of the Set is done using hashmaps and the lookup in a hashmap is `O(1)`. The check for route's admitted status will run for only `q` routes and the time complexity will be `O(qr)`.
So, the overall time complexity for this approach will be: `O(n) + O(n) + O(p+q) + O(qr)`. This will boil down to `O(qr)` as the lower order time complexity will not have as much impact as the higher order time complexity. Though I am not completely sure whether the time complexity of fetching the namespaces should be `O(n)`.

2\) As there are no namespace filtering, the time complexity calculation will only involve calculating it for the following code snippet:
  ```go
  // Iterate through the list Routes.
  for _, route := range routesInShard.Items {
  // Check if the Route's Namespace matches one of the Namespaces in the set namespacesSet and
  // the Route is admitted by the Ingress Controller.
  if routestatus.IsRouteStatusAdmitted(route, ingressController.Name) {
	  // If the Route is admitted then, the routesAdmitted should be incremented by 1 for the Shard.
	  routesAdmitted++
  }
  ```
Going with the same assumption about the number routes and ingresses per route, the time complexity for checking the route's admitted status will be `O(pr+qr)` as it is executed for all the `p+q` routes.

So, looking at the time complexities of both the approaches the factor of `O(pr)` is the additional impact we get for not filtering out the routes using the `namespaceSelector`. For clusters where the number of routes are small the benefit might not be significant. In fact, the extra work done for namespace filtering (fetching and constructing the Set) might sometimes (but not always) result in a little bit higher time. Whereas, for clusters with large routes the namespace filtering will beneficial as we are avoiding running through the nested loop for the filtered out routes. Again, all the calculations depend on the assumption of the time complexity for fetching the namespaces from the k8s api.

Thoughts?

gcs278: Great analysis, I'm a little rusty on O notation, but I appreciate you taking this analytical approach and keeping me honest.

As written, yes I think you have a valid argument. Did you mean an exponential here: `O(qʳ)` and `O(pʳ)`?

> Though I am not completely sure whether the time complexity of fetching the namespaces should be `O(n)`.

I think it's a little more than just O(n). Looking at the list code: https://github.com/kubernetes-sigs/controller-runtime/blob/ff622bd6042d2ebecb5cfc1dcf8c5b63cf47ec5c/pkg/cache/internal/cache_reader.go#L107

1. List all namespaces and filter by label selector. we'll call N all namespaces `O(N)`
2. Return filtered namespaces.

Then, we build a set from these I think that's `O(n)`. So wouldn't this be `O(N+n)`? But I don't think this matters because this function boils down to the highly complexity of `O(qʳ)`. 

Regardless, I think you have a valid point for efficiency here regarding filtering. I think we still want to check the route status regardless to make sure there is feedback from the router pod that it actually admitted the route (that's why I need to put back `IsRouteStatusAdmitted`).

De-duplicating logic was my priority (especially since https://issues.redhat.com/browse/OCPBUGS-2848), then efficiency next, but I think you have a good point about efficient. I just wonder if there is a nuance we are missing, but for the time being, I'll put it back the way it was.

Sorry for this tangent, but thanks for responding.

arkadeepsen: > As written, yes I think you have a valid argument. Did you mean an exponential here: `O(q)` and `O(pʳ)`?

`O(pr)` and `O(qr)` are simply `O(p*r)` and `O(q*r)` which can be interpreted as `O(n^2)`.

> 1. List all namespaces and filter by label selector. we'll call N all namespaces `O(N)`
> 2. Return filtered namespaces.
> 
> Then, we build a set from these I think that's `O(n)`. So wouldn't this be `O(N+n)`?

We can take the upper limit and consider it as `O(2*N)` which is same as `O(N)` as the constant 2 would not have much of an effect. But I was mostly referring to the overhead of the RPC which I am not sure about.

> De-duplicating logic was my priority (especially since https://issues.redhat.com/browse/OCPBUGS-2848), then efficiency next, but I think you have a good point about efficient. I just wonder if there is a nuance we are missing, but for the time being, I'll put it back the way it was.

Maybe we can think of a different way of implementing the route metrics controller. We can do something like this (https://github.com/openshift/cluster-ingress-operator/pull/779/commits/04df4b7994bb6db3eecd048c065087e47637b32c). Though this implementation makes a lot of assumptions about how `predicate.Funcs` work and valid questions were raised (https://github.com/openshift/cluster-ingress-operator/pull/779#discussion_r939069420). But the same logic can be implemented without using them as well. Here we only update the metric for the particular route for which a CRUD event happened. We just depend on the route status of the particular.

gcs278: > O(pr) and O(qr) are simply O(p*r) and O(q*r) which can be interpreted as O(n^2).

Okay yea you are right, I had to brush up on O Notation.

> Maybe we can think of a different way of implementing the route metrics controller. We can do something like this (https://github.com/openshift/cluster-ingress-operator/commit/04df4b7994bb6db3eecd048c065087e47637b32c). Though this implementation makes a lot of assumptions about how predicate.Funcs work and valid questions were raised (https://github.com/openshift/cluster-ingress-operator/pull/779#discussion_r939069420). But the same logic can be implemented without using them as well. Here we only update the metric for the particular route for which a CRUD event happened. We just depend on the route status of the particular route.

Yea that's a valid point and I'll keep that pattern in mind, but I think I'm going to abstain from further improvements to the route-metrics controller in this PR and we can rewrite if it efficiency issues come up later with the route-metrics controller in a different PR. I'll keep my PRs modifications to make logic common between route-metric and route-status for the time being.

arkadeepsen: > I think I'm going to abstain from further improvements to the route-metrics controller in this PR and we can rewrite if it efficiency issues come up later with the route-metrics controller in a different PR

Makes sense. That can be taken up separately whenever required.

gcs278:
DNS Umanaged To Managed E2E problem and `The HAProxy router converges when multiple routers are writing status`
/retest

gcs278:
/hold
I think this PR may be triggering issues with: `[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router converges when multiple routers are writing conflicting status`

pkg/operator/controller/route-status/route_status.go
```diff
@@ -0,0 +1,273 @@
...
+
+// ClearAllRoutesStatusForIngressController clears any route status that has been admitted by the ingress controller,
+// regardless if it is actually admitted or not. This function should be used for deletions of ingress controllers.
+func ClearAllRoutesStatusForIngressController(kclient client.Client, icName string) []error {
+	// List all routes.
+	errs := []error{}
+	start := time.Now()
+	routeList := &routev1.RouteList{}
+	routesCleared := 0
+	if err := kclient.List(context.TODO(), routeList); err != nil {
```
rfredette, line 52: Does this list all routes in the entire cluster? That seems like that could be a lot of items to loop over. 

It looks like `List()` can filter based on namespace, selector, or labels that the items have (and maybe there are other filtering options I haven't seen). Would any of those be applicable to cut down on the number of items that `List()` returns?

gcs278: Good question. Full disclosure, I didn't change this logic in this PR (not saying I wont' fix it now though), but I did write it. Use https://github.com/openshift/cluster-ingress-operator/pull/849/commits/e2c30c47ef5ccf467d7a6ad924386c5a506c0147 for your comparison since `route_status.go` shows as added/removed in the PR diff.

Potentially, we could read the IC's route selector and namespace selector (if exists) and filter by those. In theory, we should only have to clear routes that are current selected by the selectors, however, clearing all of the routes is a bit more fail-safe. If another part of this route status logic didn't clear a status when say the routeSelector was updated, doing all routes would cover our tracks.

I agree it's a performance hit, but since this only happens when you delete an IC, I'd argue it's not that big of  problem. Do you agree that deleting an IC should be pretty rare?

pkg/operator/controller/route-status/route_status.go
```diff
@@ -0,0 +1,273 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+)
+
+// A brief overview of the operator's interactions with route status:
```
rfredette, line 21: I really appreciate this overview. The scenarios listed really help contextualize what's going on in this controller, and it made it much easier to follow.

pkg/operator/controller/route-status/route_status.go
rfredette: Yeah, I agree that IC deletion is pretty rare, and making sure you definitely catch every route makes sense.

rfredette:
This looks good to me. Once you figure out the failure in `[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router converges when multiple routers are writing conflicting status`, let me know and I'll take another pass at it.

gcs278:
@rfredette I pushed a fix for the E2E test: https://github.com/openshift/cluster-ingress-operator/compare/e2c30c47ef5ccf467d7a6ad924386c5a506c0147..5724eeda2bd6d0317cea2c53d666e9e3d15a70ce

Turns out we have E2E tests that test the routers status functionality by configuring the routers to write an admitted status and waiting for all of them to agree. Well, my logic removed the status since they weren't actually admitted to an IC. I changed the logic to not remove route status if the IC doesn't exist. In theory, it should be removed by the deletion finalizer, so it is a bit extraneous.

gcs278:
Errors not related
/retest

gcs278:
All known issues
/retest

gcs278:
all known issues
/retest

gcs278:
@rfredette any updates on reviewing this? I apologize, I know it's a painful review. I did rework some things by de-duplicating logic, moving files around, and we talked about overloading PRs at one point, so I'm open to opinions on whether I should separate this into multiple PRs.

gcs278:
@rfredette thoughts? This review is starting to languish.

ShudiLi:
/label qe-approved
thanks

gcs278:
/hold
Chatted with @rfredette today - I'm going to look at potential updates to the origin test to see if we keep the robust logic to remove status if an ingress controller doesn't exist.

gcs278:
**Update**: Chatted with Miciah and Ryan, we decided it's best that the ingress operator does NOT remove stale status if an associated ingress controller object doesn't exist, which considers that there may be unconventional deployments of the router in different openshift models/platforms in the future.

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,859 @@
...
+			ingressController: newIngressControllerWithSelectors("foo", "shard", "", "", "shard"),
+			expectedRoutes: routev1.RouteList{
+				Items: []routev1.Route{
+					*newRouteWithLabelWithAdmittedStatuses("foo-route", "ns", "shard", "foo", "bar"),
+					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns", "", "baz", "bar"),
+				},
+			},
+		},
+		{
+			name: "clear route test that is no longer admitted by ingress controller by namespaceSelectors expression",
```
gcs278, line 247: this is the same test function as above

gcs278:
@rfredette I reviewed my code, made some update (mainly to unit tests), and push.

Updates [here](https://github.com/openshift/cluster-ingress-operator/compare/5724eeda2bd6d0317cea2c53d666e9e3d15a70ce..9c6165a667299feec0d4f2b2096b29fa4d87130a):

- Added a Reconcile function unit test in `pkg/operator/controller/route-status/controller_test.go`
- `isRouteAdmittedByIngressController`didn't need to be public
- Updated names of objects in unit tests to be "*-ic" or "*-route" or "*-label" for readability
- Removed redundant test cases
- Removed unit tests `Test_IsRouteAdmittedByIngressController` and `Test_clearStaleRouteAdmittedStatus` because they quite literally had the same test cases as the new `Test_Reconcile`. I feel it's better to test at the Public level (see https://redhat-internal.slack.com/archives/CBWMXQJKD/p1673537933270319)

Let me know if you have any questions, it's ready for you to review again.

gcs278:
Fixed verify with `go mod tidy` and `go mod vendor`

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,199 @@
...
+			if tc.ingressController != nil {
+				if err := r.client.Create(context.Background(), tc.ingressController); err != nil {
+					t.Errorf("error creating ingress controller: %v", err)
+				}
+			}
+
+			if _, err := r.Reconcile(context.Background(), tc.request); tc.expectedErr {
+				if err == nil {
+					t.Errorf("expected error, got no error")
+				}
```
rfredette, line 174:
```suggestion
			if _, err := r.Reconcile(context.Background(), tc.request); tc.expectedErr && err == nil {
				t.Errorf("expected error, got no error")
```
You could flatten this if condition without affecting the test functionality

pkg/operator/controller/route-status/route_status.go
```diff
@@ -0,0 +1,298 @@
...
+		return append(errs, err)
+	}
+
+	// List routes filtered by our ingress's route selector.
+	routeSelector, err := metav1.LabelSelectorAsSelector(ingress.Spec.RouteSelector)
+	if err != nil {
+		return append(errs, fmt.Errorf("ingresscontroller %s has an invalid route selector: %w", ingress.Name, err))
+	}
+
+	// Iterate over the entire route list and clear if not selected by route selector OR namespace selector.
```
rfredette, line 124:
```suggestion
	// Iterate over the entire route list and clear if either the route selector OR the namespace selector does not select it
```
This comment reads to me as it will clear the status if `!(routeInShard || namespaceInShard)`, which isn't consistent with the way it actually works. Maybe reword it to help prevent future confusion

rfredette (approved):
I have a couple of nit-picks. Neither of them actually changes the functionality, so I think of them as optional improvements for clarity

rfredette:
/lgtm
/hold

Take a look at the suggestions I made, but if you don't feel they're necessary, feel free to remove this hold.

pkg/operator/controller/route-status/route_status.go
gcs278: Done

pkg/operator/controller/route-status/controller_test.go
gcs278: Actually, it would slightly change the test functionality. If `tc.expectedErr` is `true` and `err` is not `nil` (which is a successful/passing test), this logic flow says to not check against the expectedRoute (aka the error check is enough).

However, this doesn't matter because I don't use the `tc.expectedErr`. I felt it nice to have it there for future unit test capabilities. I modified it to flatten and check `} else if tc.expectedRoute != nil && !tc.expectedErr {` down in the last if/else case.

gcs278:
@rfredette thanks for the review. I added the suggestions, could you PTAL?

gcs278:
/hold
Actually - there's a CI flake on `TestAll/parallel/TestIngressControllerRouteSelectorUpdateShouldClearRouteStatus`
`    router_status_test.go:135: failed to update ingresscontroller: Operation cannot be fulfilled on ingresscontrollers.operator.openshift.io "ic-route-selector-test": the object has been modified; please apply your changes to the latest version and try again
`

gcs278:
@rfredette I believe the test failure was from naming collisions in my parallel status tests. I just deconflicted names of everything: https://github.com/openshift/cluster-ingress-operator/compare/cb5cadeee17a7d91c113bfc3650f8591ca7dfd8a..2856c79b19f652fc0972b320eb3be780ee15cc2b

Mind taking another look?

gcs278:
I made the route names too long so I got: `spec.host: Invalid value: "route-e2e-ic-routestatus-route-label-qs52b-test-route-label-4fnfc.apps.ci-op-pk079vzc-43abb.origin-ci-int-aws.dev.rhcloud.com": must be no more than 63 characters`

Fixed that.

frobware:
> I made the route names too long so I got: `spec.host: Invalid value: "route-e2e-ic-routestatus-route-label-qs52b-test-route-label-4fnfc.apps.ci-op-pk079vzc-43abb.origin-ci-int-aws.dev.rhcloud.com": must be no more than 63 characters`

We should add a helper and just panic if > 63. Would save a lot of time.

gcs278:
Doesn't appear to be any of my new E2E tests failing.
/retest-required

gcs278:
@rfredette take a look when you can, but E2E `TestRouteMetricsControllerRouteAndNamespaceSelector` failed, I will look into that to see if I was my fault.

gcs278:
Known DNS To Unmanaged DNS issue:
/test ci/prow/e2e-aws-operator

gcs278:
Cluster install failure
/test ci/prow/e2e-aws-ovn

gcs278:
Cluster install failure
/test e2e-aws-ovn

gcs278:
Known DNS To Unmanaged DNS issue:
/test e2e-aws-operator

rfredette:
Looks like it hit the same DNS test failure again.
/test e2e-aws-operator

gcs278:
`TestUnmanagedDNSToManagedDNSInternalIngressController` failure
/test e2e-aws-operator

gcs278:
/test e2e-aws-operator
/test e2e-gcp-operator

Making sure I'm not introducing flakes.

gcs278:
Known issues
/test e2e-aws-operator
/test e2e-gcp-operator

gcs278:
Known issues
/test e2e-aws-operator
/test e2e-gcp-operator

ShudiLi:
/label qe-approved
thanks

gcs278:
Known issues
/test e2e-aws-operator
/test e2e-gcp-operator

gcs278:
Just another round to make sure this isn't introducing any more flakes
/test e2e-aws-operator
/test e2e-gcp-operator

gcs278:
CI is looking good with a couple of spins.

test/e2e/router_status_test.go
```diff
@@ -259,6 +260,139 @@ func TestIngressControllerNamespaceSelectorUpdateShouldClearRouteStatus(t *testi
...
+			t.Fatalf("failed to delete route %s: %v", routeName, err)
+		}
+	}()
+
+	// Wait for route to be admitted upon creation.
+	if err := waitForRouteIngressConditions(t, kclient, routeName, ic.Name, admittedCondition); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	// Update namespace to not match the namespaceSelector.
```
Miciah, line 381:
```suggestion
	// Update route to not match the routeSelector.
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			// Load up client with test objects.
+			client := unit.NewFakeClient(tc.ingressController)
+			for _, route := range tc.routes.Items {
+				if err := client.Create(context.Background(), &route); err != nil {
+					t.Errorf("error creating route: %v", err)
+				}
+			}
+
+			if errs := ClearAllRoutesStatusForIngressController(client, tc.ingressController.Name); tc.expectedErr && len(errs) == 0 {
```
Miciah, line 102: Should there be a test case with `expectedErr: true`?

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+					var actualRoute routev1.Route
+					for _, route := range actualRoutes.Items {
+						if route.Name == expectedRoute.Name {
+							actualRoute = route
+						}
+					}
+					assert.Equal(t, expectedRoute.Status, actualRoute.Status, "route name", expectedRoute.Name)
+				}
+			} else {
+				t.Errorf("error retrieving routes from client: %v", err)
```
Miciah, line 121:
```suggestion
			if err := client.List(context.Background(), &actualRoutes); err != nil {
				t.Fatalf("error retrieving routes from client: %v", err)
			}
			for _, expectedRoute := range tc.expectedRoutes.Items {
				// Find the actual route that we should compare status with this expected route
				var actualRoute routev1.Route
				for _, route := range actualRoutes.Items {
					if route.Name == expectedRoute.Name {
						actualRoute = route
					}
				}
				assert.Equal(t, expectedRoute.Status, actualRoute.Status, "route name", expectedRoute.Name)
			}
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+			// Load up client with test objects.
+			client := unit.NewFakeClient(tc.ingressController)
+			for _, route := range tc.routes.Items {
+				if err := client.Create(context.Background(), &route); err != nil {
+					t.Errorf("error creating route: %v", err)
+				}
+			}
+
+			if errs := ClearAllRoutesStatusForIngressController(client, tc.ingressController.Name); tc.expectedErr && len(errs) == 0 {
+				t.Errorf("expected errors, got no errors")
```
Miciah, line 103:
```suggestion
				t.Error("expected errors, got no errors")
```
or
```suggestion
				t.Fatal("expected errors, got no errors")
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns", "", "baz-ic", "bar-ic"),
+				},
+			},
+		},
+		{
+			name: "clear route that is no longer admitted by ingress controller by namespaceSelectors",
+			routes: routev1.RouteList{
+				Items: []routev1.Route{
+					*newRouteWithLabelWithAdmittedStatuses("foo-route", "ns", "shard-label", "foo-ic", "bar-ic"),
+					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns2", "", "baz-ic", "foo-ic", "bar-ic"),
```
Miciah, line 205: You want the route to have the right label and just the namespace not to, right?  
```suggestion
					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns2", "shard-label", "baz-ic", "foo-ic", "bar-ic"),
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns", "", "baz-ic", "bar-ic"),
+				},
+			},
+		},
+		{
+			name: "clear route that is no longer admitted by ingress controller by namespaceSelectors expression",
+			routes: routev1.RouteList{
+				Items: []routev1.Route{
+					*newRouteWithLabelWithAdmittedStatuses("foo-route", "ns", "shard-label", "foo-ic", "bar-ic"),
+					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns2", "", "baz-ic", "foo-ic", "bar-ic"),
```
Miciah, line 227:
```suggestion
					*newRouteWithLabelWithAdmittedStatuses("bar-route", "ns2", "shard-label", "baz-ic", "foo-ic", "bar-ic"),
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+			for _, ns := range tc.namespace.Items {
+				if err := client.Create(context.Background(), &ns); err != nil {
+					t.Errorf("error creating ns: %v", err)
+				}
+			}
+
+			if errs := ClearRoutesNotAdmittedByIngress(client, tc.ingressController); tc.expectedErr && len(errs) == 0 {
+				t.Errorf("expected errors, got no errors")
+			} else if !tc.expectedErr && len(errs) != 0 {
+				t.Errorf("did not expected errors: %v", errs)
```
Miciah, line 277: These can probably all use `t.Fatal` or `t.Fatalf` instead of `t.Errorf`.

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+					var actualRoute routev1.Route
+					for _, route := range actualRoutes.Items {
+						if route.Name == expectedRoute.Name {
+							actualRoute = route
+						}
+					}
+					assert.Equal(t, expectedRoute.Status, actualRoute.Status, "route name", expectedRoute.Name)
+				}
+			} else {
+				t.Errorf("error retrieving all from client: %v", err)
```
Miciah, line 293:
```suggestion
			if err := client.List(context.Background(), &actualRoutes); err != nil {
				t.Fatalf("error retrieving all from client: %v", err)
			}
			for _, expectedRoute := range tc.expectedRoutes.Items {
				// Find the actual route that we should compare status with this expected route
				var actualRoute routev1.Route
				for _, route := range actualRoutes.Items {
					if route.Name == expectedRoute.Name {
						actualRoute = route
					}
				}
				assert.Equal(t, expectedRoute.Status, actualRoute.Status, "route name", expectedRoute.Name)
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+		},
+	}
+	if len(label) != 0 {
+		namespace.ObjectMeta.Labels = map[string]string{"type": label}
+	}
+	return &namespace
+}
+
+// newRouteWithAdmittedStatuses returns a new route that is admitted by ingress controllers.
+func newRouteWithLabelWithAdmittedStatuses(name, namespace, label string, icAdmitted ...string) *routev1.Route {
```
Miciah, line 437:
```suggestion
// newRouteWithLabelWithAdmittedStatuses returns a new route that is admitted by ingress controllers
// and has the "type" label set with the specified value.
func newRouteWithLabelWithAdmittedStatuses(name, namespace, label string, icAdmitted ...string) *routev1.Route {
```

pkg/operator/controller/route-status/route_status_test.go
```diff
@@ -0,0 +1,469 @@
...
+					{
+						Type:   routev1.RouteAdmitted,
+						Status: "True",
+					},
+				},
+			})
+		}
+	}
+	return &route
+}
```
Miciah, line 469: Why do these all return pointer values?  Is that just for the one test case in `Test_Reconcile` where `expectedRoute` is nil?  

I wonder how we could make the test code easier to read.  One option would be to use the builder pattern, like what Andrey did here: https://github.com/openshift/aws-load-balancer-operator/blob/54506d26e0de8cbc71289d897b71b39e1c185391/test/e2e/util.go#L274-L337

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
+)
+
+// Test_Reconcile verifies Reconcile
+// Note: This effectively tests clearStaleRouteAdmittedStatus as well
```
Miciah, line 18: Please use full sentences.  * grin *.  
```suggestion
// Test_Reconcile verifies Reconcile behaves as expected.
// Note: This effectively tests clearStaleRouteAdmittedStatus as well.
```

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "", "", "shard-label"),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", ""),
+		},
+		{
+			name:              "route is nil",
+			request:           reconcileRequestRoute,
+			route:             nil,
+			namespace:         newNamespace("ns", "not-shard-label"),
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "", "", "shard-label"),
+			expectedErr:       false,
```
Miciah, line 145: This is the only case that specifies `expectedErr`, and it specifies `expectedErr: false`.  Did you mean to put `expectedErr: true` here?  Do we really need `expectedErr` at all?

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+		},
+		{
+			name:              "admitted with no routeSelector and no namespaceSelector with route with correct labels in namespace with no labels",
+			request:           reconcileRequestRoute,
+			route:             newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+			namespace:         newNamespace("ns", ""),
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "", "", ""),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+		},
```
Miciah, line 50: These two test cases are identical.

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+		{
+			name:              "admitted with no routeSelector and no namespaceSelector with route with no labels in namespace with correct labels",
```
Miciah, line 52: If there is no namespace selector, what makes the namespace's labels "correct"?

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "", "", ""),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+		},
+		{
+			name:              "admitted with routeSelector and no namespaceSelector with route with correct labels in namespace with no labels",
+			request:           reconcileRequestRoute,
+			route:             newRouteWithLabelWithAdmittedStatuses("route", "ns", "shard-label", "default-ic"),
+			namespace:         newNamespace("ns", ""),
+			ingressController: newIngressControllerWithSelectors("default-ic", "shard-label", "", "", ""),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
```
Miciah, line 65: Why did the route's label go away?  I guess it doesn't matter because the test checks the status, not the labels, but it looks funny.

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+		},
+		{
+			name:              "not admitted with routeSelector and no namespaceSelector with route with no labels in namespace with no labels",
+			request:           reconcileRequestRoute,
+			route:             newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
```
Miciah, line 70: It looks like the route's status reports that it *was* admitted, right?  Should I read "not admitted" as "stale Admitted status condition"?

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+		{
+			name:              "not admitted with expression routeSelector and no namespaceSelector with route with incorrect labels in namespace with no labels",
+			request:           reconcileRequestRoute,
+			route:             newRouteWithLabelWithAdmittedStatuses("route", "ns", "not-shard-label", "default-ic"),
+			namespace:         newNamespace("ns", ""),
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "shard-label", "", ""),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", ""),
+		},
+		{
+			name:              "admitted with no routeSelector and expression namespaceSelector with route with correct labels in namespace with no labels",
```
Miciah, line 124:
```suggestion
			name:              "admitted with no routeSelector and expression namespaceSelector with route with correct labels in namespace with matching label",
```

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+			request:           reconcileRequestRoute,
+			route:             newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+			namespace:         newNamespace("ns", "shard-label"),
+			ingressController: newIngressControllerWithSelectors("default-ic", "", "", "", "shard-label"),
+			expectedRoute:     newRouteWithLabelWithAdmittedStatuses("route", "ns", "", "default-ic"),
+		},
+		{
+			name:              "not admitted with no routeSelector and expression namespaceSelector with route with incorrect labels in namespace with no labels",
```
Miciah, line 132:
```suggestion
			name:              "stale admitted with no routeSelector and expression namespaceSelector with route with no labels in namespace with no matching labels",
```

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			r := newFakeReconciler(tc.namespace)
+			if tc.route != nil {
+				if err := r.client.Create(context.Background(), tc.route); err != nil {
+					t.Errorf("error creating route: %v", err)
+				}
+			}
+			if tc.ingressController != nil {
+				if err := r.client.Create(context.Background(), tc.ingressController); err != nil {
+					t.Errorf("error creating ingress controller: %v", err)
```
Miciah, line 167: You can use `t.Fatalf` here.

pkg/operator/controller/route-status/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+func Test_Reconcile(t *testing.T) {
+	reconcileRequestRoute := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      "route",
+			Namespace: "ns",
+		},
+	}
+	testCases := []struct {
+		name              string
+		request           reconcile.Request
```
Miciah, line 28: Does `request` need to be part of the test input?  It appears to be the same in every test case.

pkg/operator/operator.go
```diff
@@ -209,11 +210,27 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
...
+	// Create a new cache to watch on Route objects from every namespace for the route metrics and the route status
+	// controller to share.
+	globalCache, err := cache.New(mgr.GetConfig(), cache.Options{
+		Scheme: mgr.GetScheme(),
+	})
+	if err != nil {
+		return nil, err
+	}
+	// Add the cache to the manager so that the cache is started along with the other runnables.
+	mgr.Add(globalCache)
```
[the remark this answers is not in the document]
gcs278: I called it `globalCache` because it includes every namespace. Wouldn't it be okay if other resources/controllers used it for cluster-scoped watches? Isn't that the point? Or am I misunderstanding something?

I don't mind changing it, but `routeCache` seems artificially limited, it's not just for routes, right? It could be for any object. is that what you are saying we want?

And `clusterScopedCache` is an overloaded term, aren't cluster scoped resources technically not apart of a namespace (when this cache watches all namespaces)?

pkg/operator/controller/route-status/route_status_test.go
gcs278: Done

gcs278 (commented):
Thanks for the very detailed review.

gcs278:
/hold
Things left for me TODO
1. Try to refactor unit test code with builder functions
2. Analyze startup time of CIO with new change

pkg/operator/controller/route-status/route_status_test.go
gcs278: **Update**: I added the builder pattern in https://github.com/openshift/cluster-ingress-operator/pull/849/commits/efc08f91fc67ffb060cb4616f78894171f08ccb7. It's reusable, and hopefully much more readable.

gcs278:
My commits are pretty messy I think better to squash:
/label tide/merge-method-squash

ShudiLi:
From QE side, after rebel the namespace, the route was not admitted anymore.
1.
%oc label namespace test type=loadbalancer

2.
% oc get route router-loadbalancer-test -oyaml
...
status:
  ingress:
  - conditions:
    - lastTransitionTime: "2023-02-20T09:44:54Z"
      status: "True"
      type: Admitted
    host: router-loadbalancer-test-test.apps.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerCanonicalHostname: router-default.apps.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerName: default
    wildcardPolicy: None
  - conditions:
    - lastTransitionTime: "2023-02-20T09:44:54Z"
      status: "True"
      type: Admitted
    host: router-loadbalancer-test-test.apps.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerCanonicalHostname: router-loadbalancer.loadbalancer.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerName: loadbalancer
    wildcardPolicy: None
%

3.
% oc label namespace test type=opt-out --overwrite

4.
% oc get route router-loadbalancer-test -oyaml 
...
status:
  ingress:
  - conditions:
    - lastTransitionTime: "2023-02-20T09:44:54Z"
      status: "True"
      type: Admitted
    host: router-loadbalancer-test-test.apps.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerCanonicalHostname: router-default.apps.ci-ln-kbq4082-72292.origin-ci-int-gce.dev.rhcloud.com
    routerName: default
    wildcardPolicy: None
% 

/label qe-approved
thanks

alebedev87:
/unassign

I withdraw myself from the review because the PR has enough of quality reviews already made.

gcs278:
Still holding. Going to push through https://github.com/openshift/cluster-ingress-operator/pull/869, which is a simpler PR that introduces the builder pattern that these unit tests need.

gcs278:
Closing in favor of a router-specific implementation which is much cleaner (in my humble opinion): https://github.com/openshift/router/pull/514

PR #852, opened 2022-10-27
OCPBUGS-2247: Add explicit message to condition when domain doesn't match basedomain

thejasn:
### Description
Add additional reason to status to explain why `dnsManagementPolicy` was set to `Unmanaged` when domain of Ingress Controller doesn't match cluster base domain.

Currently, when a user creates an Ingress Controller with a custom domain (doesn't match cluster base domain) the `.spec.endpointPublishingStrategy.loadBalancer.dnsManagementPolicy` defaults to `Managed` whereas the operator correctly updates the `.status.endpointPublishingStrategy.loadBalancer.dnsManagementPolicy` to be `Unmanaged`. This result although odd is expected but may not be very clear to the end user. 

Adding a more specific reason to the existing condition `DNSManaged` when the domains are mismatched would be more clear to the end user.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -979,7 +979,14 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 		}
 	}
 	var conditions []operatorv1.OperatorCondition
-	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+	if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) && ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.DNSManagedIngressConditionType,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "DomainNotMatching",
+			Message: "DNS management is not supported for ingresscontrollers with domain not matching the baseDomain of the cluster DNS config.",
+		})
+	} else if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
 		conditions = append(conditions, operatorv1.OperatorCondition{
 			Type:    operatorv1.DNSManagedIngressConditionType,
 			Status:  operatorv1.ConditionFalse,
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -6,12 +6,13 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
-	"k8s.io/utils/pointer"
 	"math/big"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
+
+	"k8s.io/utils/pointer"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -1581,15 +1582,61 @@ func TestComputeDNSStatus(t *testing.T) {
 			}},
 		},
 		{
-			name: "DNSManaged false due to UnmanagedLoadBalancerDNS",
+			name: "DNSManaged false due to domain not matching base domain",
 			dnsConfig: &configv1.DNS{
 				Spec: configv1.DNSSpec{
+					BaseDomain:  "basedomain.com",
 					PublicZone:  &configv1.DNSZone{},
 					PrivateZone: &configv1.DNSZone{},
 				},
 			},
+			platformStatus: &configv1.PlatformStatus{
+				Type: configv1.AWSPlatformType,
+			},
 			controller: &operatorv1.IngressController{
 				Status: operatorv1.IngressControllerStatus{
+					Domain: "apps.example.com",
+					EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
+						Type: operatorv1.LoadBalancerServiceStrategyType,
+						LoadBalancer: &operatorv1.LoadBalancerStrategy{
+							DNSManagementPolicy: operatorv1.UnmanagedLoadBalancerDNS,
+						},
+					},
+				},
+			},
+			record: &iov1.DNSRecord{
+				Spec: iov1.DNSRecordSpec{
+					DNSManagementPolicy: iov1.UnmanagedDNS,
+				},
+			},
+			expect: []operatorv1.OperatorCondition{
+				{
+					Type:   "DNSManaged",
+					Status: operatorv1.ConditionFalse,
+					Reason: "DomainNotMatching",
+				},
+				{
+					Type:   "DNSReady",
+					Status: operatorv1.ConditionUnknown,
+					Reason: "UnmanagedDNS",
+				},
+			},
+		},
+		{
+			name: "DNSManaged false due to UnmanagedLoadBalancerDNS",
+			dnsConfig: &configv1.DNS{
+				Spec: configv1.DNSSpec{
+					BaseDomain:  "basedomain.com",
+					PublicZone:  &configv1.DNSZone{},
+					PrivateZone: &configv1.DNSZone{},
+				},
+			},
+			platformStatus: &configv1.PlatformStatus{
+				Type: configv1.AWSPlatformType,
+			},
+			controller: &operatorv1.IngressController{
+				Status: operatorv1.IngressControllerStatus{
+					Domain: "apps.basedomain.com",
 					EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
 						Type: operatorv1.LoadBalancerServiceStrategyType,
 						LoadBalancer: &operatorv1.LoadBalancerStrategy{
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -1664,6 +1664,45 @@ func TestComputeDNSStatus(t *testing.T) {
 			},
 		},
 		{
+			name: "DNSManaged is false due to DomainNotMatching because platformStatus is nil",
+			dnsConfig: &configv1.DNS{
+				Spec: configv1.DNSSpec{
+					BaseDomain:  "basedomain.com",
+					PublicZone:  &configv1.DNSZone{},
+					PrivateZone: &configv1.DNSZone{},
+				},
+			},
+			platformStatus: nil,
+			controller: &operatorv1.IngressController{
+				Status: operatorv1.IngressControllerStatus{
+					Domain: "apps.basedomain.com",
+					EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
+						Type: operatorv1.LoadBalancerServiceStrategyType,
+						LoadBalancer: &operatorv1.LoadBalancerStrategy{
+							DNSManagementPolicy: operatorv1.UnmanagedLoadBalancerDNS,
+						},
+					},
+				},
+			},
+			record: &iov1.DNSRecord{
+				Spec: iov1.DNSRecordSpec{
+					DNSManagementPolicy: iov1.UnmanagedDNS,
+				},
+			},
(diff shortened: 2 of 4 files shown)


thejasn:
/retest 
failed during must-gather

thejasn:
/test e2e-aws-operator
Failed during deprovisioning of cluster

candita:
@thejasn can you add a better description? I can't tell from the bug report what the expectation is here.  Is it that the log event isn't enough?  We don't usually add a status condition for something like this, and if we did, we'd probably want to note what was incorrect and how to fix it.

pkg/operator/controller/ingress/status.go
```diff
@@ -979,7 +979,14 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
 	var conditions []operatorv1.OperatorCondition
-	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+	if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) && ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+		conditions = append(conditions, operatorv1.OperatorCondition{
+			Type:    operatorv1.DNSManagedIngressConditionType,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "DomainNotMatching",
+			Message: "DNS management is not supported for ingresscontrollers with domain not matching the baseDomain of the cluster DNS config.",
+		})
+	} else if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
```
candita, line 989:
```suggestion
	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
	    if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) {
	  	    conditions = append(conditions, operatorv1.OperatorCondition{
			Type:    operatorv1.DNSManagedIngressConditionType,
			Status:  operatorv1.ConditionFalse,
			Reason:  "DomainNotMatching",
			Message: fmt.Sprintf("Domain %s not matching baseDomain %s.",ic.Status.Domain, dnsConfig.Spec.BaseDomain),
		    })
	} else {
```

pkg/operator/controller/ingress/status.go
```diff
@@ -979,7 +979,14 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 		}
 	}
 	var conditions []operatorv1.OperatorCondition
-	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
+	if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) && ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
```
candita, line 982: What if someone wants it to be unmanaged, and set it that way?  Is there any reason to set this condition in that case?

pkg/operator/controller/ingress/status_test.go
```diff
@@ -1580,16 +1581,62 @@ func TestComputeDNSStatus(t *testing.T) {
...
 		{
 			name: "DNSManaged false due to UnmanagedLoadBalancerDNS",
 			dnsConfig: &configv1.DNS{
 				Spec: configv1.DNSSpec{
+					BaseDomain:  "basedomain.com",
 					PublicZone:  &configv1.DNSZone{},
 					PrivateZone: &configv1.DNSZone{},
 				},
 			},
+			platformStatus: &configv1.PlatformStatus{
```
candita, line 1634: Please keep the testcase for when the platformStatus is not set, and add a new testcase for the new case.

thejasn: This test is applicable only when `dnsManagementPolicy` is set to `Unmanaged`. The domains match in this case. I had to introduce `configv1.PlatformStatus{}` since code now always depends on `manageDNSForDomain()` which expects `PlatformStatus` to be non-nil always. 

Previously this dependency on `manageDNSForDomain()` didn't exist when `dnsManagementPolicy` is set to `Unamanaged`.

pkg/operator/controller/ingress/status.go
thejasn:
```suggestion
	if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {
		if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) {
			conditions = append(conditions, operatorv1.OperatorCondition{
				Type:    operatorv1.DNSManagedIngressConditionType,
				Status:  operatorv1.ConditionFalse,
				Reason:  "DomainNotMatching",
				Message: "DNS management is not supported for ingresscontrollers with domain not matching the baseDomain of the cluster DNS config.",
			})
		} else {
			conditions = append(conditions, operatorv1.OperatorCondition{
				Type:    operatorv1.DNSManagedIngressConditionType,
				Status:  operatorv1.ConditionFalse,
				Reason:  "UnmanagedLoadBalancerDNS",
				Message: "The DNS management policy is set to Unmanaged.",
			})
		}
	} else {
		conditions = append(conditions, operatorv1.OperatorCondition{
			Type:    operatorv1.DNSManagedIngressConditionType,
			Status:  operatorv1.ConditionTrue,
			Reason:  "Normal",
			Message: "DNS management is supported and zones are specified in the cluster DNS config.",
		})
	}
```
Hmm, are you suggesting I change the `if..else..if` like this?

ShudiLi:
/label qe-approved
thanks

pkg/operator/controller/ingress/status.go
candita: You had `if ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS ` in both places, and I was suggesting to put that in the outer `if` clause.

pkg/operator/controller/ingress/status_test.go
candita: > expects PlatformStatus to be non-nil always.
I don't think we can make that expectation - manageDNSforDomain needs to be prepared for when it is nil. Am I missing something?

thejasn: Agreed, will patch `manageDNSforDomain()` and add new tests accordingly.

thejasn: Hey @candita, what would your suggestion be for this case? 
```go
--- a/pkg/operator/controller/ingress/dns.go
+++ b/pkg/operator/controller/ingress/dns.go
@@ -208,6 +208,10 @@ func manageDNSForDomain(domain string, status *configv1.PlatformStatus, dnsConfi
                return false
        }

+       if dnsConfig == nil || status == nil {
+               return false
+       }
+
        mustContain := "." + dnsConfig.Spec.BaseDomain
        switch status.Type {
``` 
Returning `false` when either `platformStatus` or `dnsConfig` is `nil` will lead to incorrectly updating the condition to `DomainNotMatching`. And returning a `true` would also be incorrect.

There is a check for the same already ([here](https://github.com/openshift/cluster-ingress-operator/blob/d9d1a2b44cc7955a18fbedfdc973daddba67bccd/pkg/operator/controller/ingress/controller.go#L264)) but it isn't local to `manageDNSForDomain()` and thereby not covered by the tests in `status_test.go`.

gcs278: Sorry ahead of time if you already have accomplished this, but making sure we are all on the same page:

I think we need nil pointer checks before using pointers in functions, regardless if we do a nil pointer check outside of the function. We should also unit test for nil pointer exceptions. Sounds like we are talking about two different functions with possible Nil Pointer Exceptions:

**computeDNSStatus**:
Seems like the `status *configv1.PlatformStatus` is passed directly into `manageDNSForDomain`, so really it's `manageDNSForDomain`'s problem. So I think we okay here, but still worth keeping a unit test in `TestCoHope this helps and feel free to push back on some of this, but I do think this is a good opportunity to do these small maintenance tasks.mputeDNSStatus` like Candace said to make sure `computeDNSStatus` doesn't get a NIL Pointer exception in the future when code changes.

Unrelated to this PR, but`dnsConfig` this is missing a NIL Pointer check, that should probably be fixed sometime.

**manageDNSForDomain**:
I'd vote to just return `false` when given a nil pointer, since saying DNS is unmanageable is more accurate than saying`true` (it's manageable). Either way, it's much better than a nil pointer exception.

For `TestManageDNSForDomain` should have NIL Pointer test cases in it, but the test is set up a little weird, so maybe you could modify it to do something like this:
```
var status *configv1.PlatformStatus
if tc.platformType != "" {
  status = configv1.PlatformStatus{
    Type: tc.platformType,
  }

var dnsConfig *configv1.DNS
if tc.baseDomain != "" {
  dnsConfig = configv1.DNS{
    Spec: configv1.DNSSpec{
      BaseDomain: tc.baseDomain,
    },
 }
}
actual := manageDNSForDomain(tc.domain, status, dnsConfig)
```
And add the unit test cases that would trigger this `false` return value.

thejasn: I've addressed the comments.

pkg/operator/controller/ingress/status.go
```diff
@@ -979,7 +979,14 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
...
+		conditions = append(conditions, operatorv1.OperatorCondition{
```
candita, line 983: Should we just return the single condition `return []operatorv1.OperatorCondition` here?  Do we really want to go on and check the `wildcardRecord` and append more conditions after this?

candita:
> Step e2e-aws-operator-gather-must-gather failed after 2m50s. 
/retest-required

pkg/operator/controller/ingress/status.go
thejasn: The two sections are for 2 different condition types (`DNSManagedIngressConditionType` and `DNSReadyIngressConditionType`), so even if domain doesn't match and ingresscontroller status has `dnsManagementPolicy` as `Unmanaged` we still need to update `DNSReady` condition based on wildcard record.

thejasn:
>level=error msg=Error: retrieving Target Group Attributes: TargetGroupNotFound: Target groups 'arn:aws:elasticloadbalancing:us-east-2:460538899914:targetgroup/ci-op-gj117i6q-43abb-z94f4-aext/696bdce4ee7f12d6' not found
level=error msg=	status code: 400, request id: f[56](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/852/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1600517195798941696#1:build-log.txt%3A56)fa335-aba4-48d7-a2d8-e7fc1e83ee18

/retest-required

pkg/operator/controller/ingress/status.go
candita: Doesn't this mean that the domains don't match and it is unmanaged?

> 	if !manageDNSForDomain(ic.Status.Domain, status, dnsConfig) && ic.Status.EndpointPublishingStrategy.LoadBalancer.DNSManagementPolicy == operatorv1.UnmanagedLoadBalancerDNS {

candita:
On hold until all comments are addressed.
/hold

pkg/operator/controller/ingress/status.go
candita: In other words, if you rely on the mismatch between `domain` and `baseDomain` to determine whether `UnmanagedLoadBalancerDNS` was intentional or not, that is incorrect because the spec domains are not cleaned up when it is intentionally set to unmanaged.  What you would need to do is check the spec against the status DNSManagementPolicy, and if they aren't the same, and there is a mismatch, then you could add this condition.

gcs278:
@thejasn is there any update on this PR? Still planning to work it?

thejasn:
Yes, Candace had given a few suggestions on this, will get updated ASAP (had to de-prioritize this in favor of another task).

candita:
@thejasn will you be working on this?  If so, can you remove the `rotten` label and rebase?

PR #854, opened 2022-11-04
OCPBUGS-434: Absorb PodsScheduled condition into MinAvailable

Miciah:
#### Absorb PodsScheduled condition into MinAvailable

Remove the `PodsScheduled` status condition on IngressController objects, and report the information that `PodsScheduled` reported instead in the `DeploymentReplicasMinAvailable` status condition's message when its status is "False".

Before this change, the `PodsScheduled` status condition could erroneously signal that there was a problem when some misscheduled pods existed even if there were enough correctly scheduled and available pods.

The `PodsScheduled` status condition was added in order to provide diagnostic information when there was a problem.  That is, `PodsScheduled` was intended to indicate *why* there was a problem, not *whether* there was a problem. The `DeploymentReplicasMinAvailable` status condition definitively indicates *whether* there is a problem, and its status message should indicate *why* there is a problem.  Having a separate status condition for the "why" is superfluous and unnecessarily complicated in this instance.

Follow-up to #465.

* `pkg/operator/controller/ingress/controller.go` (`IngressControllerPodsScheduledConditionType`): Delete const.
* `pkg/operator/controller/ingress/metrics_test.go` (`TestDeleteIngressControllerConditionsMetric`): Delete a reference to the "PodsScheduled" status condition in a test case.
* `pkg/operator/controller/ingress/status.go` (`syncIngressControllerStatus`): Delete the call to `computeDeploymentPodsScheduledCondition`.  Pass the list of pods to `computeDeploymentReplicasMinAvailableCondition`.
(`PruneConditions`): Delete pruning of the "DeploymentDegraded" status condition (which was removed in 4.6.0) and instead prune the "PodsScheduled" status condition.
(`computeDeploymentPodsScheduledCondition`): Rename...
(`checkPodsScheduledForDeployment`): ...to this.  Change the return value from an operator status condition to an error value.
(`computeDeploymentReplicasMinAvailableCondition`): Call `checkPodsScheduledForDeployment` and include the error string in the status condition's message if the function returns an error.
(`computeIngressDegradedCondition`): Remove `IngressControllerPodsScheduledConditionType` from the list of conditions to check.
* `pkg/operator/controller/ingress/status_test.go` (`TestComputePodsScheduledCondition`): Rename...
(`Test_checkPodsScheduledForDeployment`): ...to this.  Update for the function's new name and result value.
(`TestComputeIngressDegradedCondition`): Delete test cases for `PodsScheduled`.
(`TestComputeDeploymentReplicasMinAvailableCondition`): Update to set the deployment's label selector and pass in a list of pods (which is empty).

#### `checkPodsScheduledForDeployment`: Add nil check

`checkPodsScheduledForDeployment` takes a pointer to a deployment, which should never be nil, so return an error indicating that there is a bug if the deployment is somehow nil.

* `pkg/operator/controller/ingress/status.go` (`checkPodsScheduledForDeployment`): Check whether deployment is nil and return an error if it is.
* `pkg/operator/controller/ingress/status_test.go` (`Test_checkPodsScheduledForDeployment`): Add a "nil deployment" test case.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -75,9 +75,8 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		updated.Status.EndpointPublishingStrategy.LoadBalancer.AllowedSourceRanges = computeAllowedSourceRanges(service)
 	}
 
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentPodsScheduledCondition(deployment, pods))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentAvailableCondition(deployment))
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasMinAvailableCondition(deployment, pods))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentReplicasAllAvailableCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeDeploymentRollingOutCondition(deployment))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeLoadBalancerStatus(ic, service, operandEvents)...)
@@ -153,9 +152,9 @@ func MergeConditions(conditions []operatorv1.OperatorCondition, updates ...opera
 // Returns the updated condition array.
 func PruneConditions(conditions []operatorv1.OperatorCondition) []operatorv1.OperatorCondition {
 	for i, condition := range conditions {
-		// TODO: Remove this fix-up logic in 4.8
-		if condition.Type == "DeploymentDegraded" {
-			// DeploymentDegraded was removed in 4.6.0
+		// PodsScheduled was removed in 4.13.0.
+		// TODO: Remove this fix-up logic in 4.14.
+		if condition.Type == "PodsScheduled" {
 			conditions = append(conditions[:i], conditions[i+1:]...)
 		}
 	}
@@ -204,18 +203,14 @@ func computeAllowedSourceRanges(service *corev1.Service) []operatorv1.CIDR {
 	return nil
 }
 
-// computeDeploymentPodsScheduledCondition computes the ingress controller's
-// current PodsScheduled status condition state by inspecting the PodScheduled
-// conditions of the pods associated with the deployment.
-func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods []corev1.Pod) operatorv1.OperatorCondition {
+// checkPodsScheduledForDeployment checks whether the given deployment has a
+// valid label selector and whether all of the deployment's pods are reporting
+// PodsScheduled=True.  This function returns an error value indicating the
+// result of that check.
+func checkPodsScheduledForDeployment(deployment *appsv1.Deployment, pods []corev1.Pod) error {
 	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
 	if err != nil || selector.Empty() {
-		return operatorv1.OperatorCondition{
-			Type:    IngressControllerPodsScheduledConditionType,
-			Status:  operatorv1.ConditionUnknown,
-			Reason:  "InvalidLabelSelector",
-			Message: "Deployment has an invalid label selector.",
-		}
+		return errors.New("Deployment has an invalid label selector.")
 	}
 	unscheduled := make(map[*corev1.Pod]corev1.PodCondition)
 	for i, pod := range pods {
@@ -255,17 +250,9 @@ func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods
 		if haveUnschedulable {
 			message = message + " Make sure you have sufficient worker nodes."
 		}
-		return operatorv1.OperatorCondition{
-			Type:    IngressControllerPodsScheduledConditionType,
-			Status:  operatorv1.ConditionFalse,
-			Reason:  "PodsNotScheduled",
-			Message: message,
-		}
+		return errors.New(message)
 	}
-	return operatorv1.OperatorCondition{
-		Type:   IngressControllerPodsScheduledConditionType,
-		Status: operatorv1.ConditionTrue,
-	}
+	return nil
 }
 
 // computeIngressAvailableCondition computes the ingress controller's current Available status state
@@ -411,7 +398,7 @@ func computeDeploymentAvailableCondition(deployment *appsv1.Deployment) operator
 // update parameters.  The "DeploymentReplicasMinAvailable" condition is true if
 // the number of available replicas is equal to or greater than the number of
 // desired replicas less the number minimum available.
-func computeDeploymentReplicasMinAvailableCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+func computeDeploymentReplicasMinAvailableCondition(deployment *appsv1.Deployment, pods []corev1.Pod) operatorv1.OperatorCondition {
 	replicas := int32(1)
 	if deployment.Spec.Replicas != nil {
 		replicas = *deployment.Spec.Replicas
@@ -450,11 +437,15 @@ func computeDeploymentReplicasMinAvailableCondition(deployment *appsv1.Deploymen
 		maxUnavailable = 1
 	}
 	if int(deployment.Status.AvailableReplicas) < int(replicas)-maxUnavailable {
+		message := fmt.Sprintf("%d/%d of replicas are available, max unavailable is %d", deployment.Status.AvailableReplicas, replicas, maxUnavailable)
+		if err := checkPodsScheduledForDeployment(deployment, pods); err != nil {
+			message = fmt.Sprintf("%s: %s", message, err.Error())
+		}
 		return operatorv1.OperatorCondition{
 			Type:    IngressControllerDeploymentReplicasMinAvailableConditionType,
 			Status:  operatorv1.ConditionFalse,
 			Reason:  "DeploymentMinimumReplicasNotMet",
-			Message: fmt.Sprintf("%d/%d of replicas are available, max unavailable is %d", deployment.Status.AvailableReplicas, replicas, maxUnavailable),
+			Message: message,
 		}
 	}
 
@@ -555,11 +546,6 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, 
 		{
 			condition: IngressControllerAdmittedConditionType,
 			status:    operatorv1.ConditionTrue,
-		},
-		{
-			condition:   IngressControllerPodsScheduledConditionType,
-			status:      operatorv1.ConditionTrue,
-			gracePeriod: time.Minute * 10,
 		},
 		{
 			condition:   IngressControllerDeploymentAvailableConditionType,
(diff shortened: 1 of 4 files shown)


Miciah:
Latest push changes from reporting the information about unscheduled/unschedulable pods from `DeploymentReplicasAllAvailable` to `DeploymentReplicasMinAvailable` because the former gets reported in `Degraded` after a 60-_minute_ grace period whereas the latter gets reported after a 60-_second_ grace period, so the cluster admin is more likely to see useful information sooner if it's in the `DeploymentReplicasMinAvailable`.

Miciah:
[e2e-aws-ovn-upgrade failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/854/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-upgrade/1588573257546600448) because deprovisioning failed.  [e2e-gcp-ovn-serial](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/854/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-ovn-serial/1588573257622097920) because etcd had an unhealthy member.  [e2e-azure-ovn failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/854/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-ovn/1588573257588543488) because terraform failed to configure the vnet.
/retest-required

pkg/operator/controller/ingress/status_test.go
```diff
@@ -203,46 +203,50 @@ func TestComputePodsScheduledCondition(t *testing.T) {
 		},
 	}
 	tests := []struct {
-		name       string
-		deployment *appsv1.Deployment
-		pods       []corev1.Pod
-		expect     operatorv1.ConditionStatus
+		name        string
+		deployment  *appsv1.Deployment
```
gcs278, line 207: I always test to ensure functions can handle nil pointers. Do you agree with testing this edge case?

Miciah: I'm ambivalent in this case and in the general case.  Adding nil checks everywhere can make code safer (less likely to panic), but it can also obscure bugs (maybe it's a bug in the caller if the argument is nil).  In this case, the caller of `checkPodsScheduledForDeployment` itself dereferences `deployment`, so it cannot be nil.  For short helper functions like this that are used in only one or two places, I would be fine with saying something like, "The caller must ensure that the deployment argument is not nil."  Would a comment in the godoc for `checkPodsScheduledForDeployment` suffice, or do you still prefer an explicit nil check in `checkPodsScheduledForDeployment`?

Miciah: Actually, `checkPodsScheduledForDeployment` returns an error value, so I'll go ahead and add a check that returns an error if `deployment` is nil.

Miciah: Added.

candita:
@Miciah I have a small concern with this.  What if SD or customers have set up their own monitoring for specific conditions, and we are changing things here.  Will there be a good release note to explain the changes to the condition set?

Miciah:
> @Miciah I have a small concern with this. What if SD or customers have set up their own monitoring for specific conditions, and we are changing things here. Will there be a good release note to explain the changes to the condition set?

We don't document the status condition, and we don't declare it in openshift/api, so we have made no promises to anyone that they can rely on it.  We could make a release note if you think that's appropriate, though I would want to state clearly in any release note that no one should be relying on undocumented status conditions anyway.  

Your question does spark a thought—we should prune the removed status condition on upgrade to make sure we don't leave a stale status condition around.

pkg/operator/controller/ingress/status.go
```diff
@@ -204,18 +203,17 @@ func computeAllowedSourceRanges(service *corev1.Service) []operatorv1.CIDR {
...
-// current PodsScheduled status condition state by inspecting the PodScheduled
-// conditions of the pods associated with the deployment.
-func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods []corev1.Pod) operatorv1.OperatorCondition {
+// checkPodsScheduledForDeployment checks whether the given deployment has a
+// valid label selector and whether all of the deployment's pods are reporting
+// PodsScheduled=True.  This function returns an error value indicating the
+// result of that check.
+func checkPodsScheduledForDeployment(deployment *appsv1.Deployment, pods []corev1.Pod) error {
+	if deployment == nil {
+		return errors.New("Deployment was nil.  This is a bug!")
```
candita, line 212:
```suggestion
		return errors.New("System or configuration error detected - deployment was nil.")
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -279,26 +289,6 @@ func TestComputeIngressDegradedCondition(t *testing.T) {
...
-		{
-			name: "pods not scheduled for >10m",
-			conditions: []operatorv1.OperatorCondition{
-				cond(IngressControllerPodsScheduledConditionType, operatorv1.ConditionFalse, "", clock.Now().Add(time.Minute*-31)),
-			},
-			expectIngressDegradedStatus: operatorv1.ConditionTrue,
-			expectRequeue:               true,
-			// Just use the one minute retry duration for this degraded condition
-			expectAfter: time.Minute,
-		},
```
candita, line 301: Do we no longer have a grace period for which pods can remain unscheduled and not yet considered unavailable?

pkg/operator/controller/ingress/status_test.go
gcs278: So I was a bit of a NIL Pointer Check absolutist until you mentioned `it can also obscure bugs`. So I agree, it's not that cut-and-dry whether every function needs a NIL pointer check. The GO community seems torn on it too based on my web searches.

So I think I don't mind either way.

gcs278:
I forgot about status pruning, good catch.

/lgtm

pkg/operator/controller/ingress/status.go
candita: Then maybe just "System error detected." and put a link to the bug reporting tool!

pkg/operator/controller/ingress/status_test.go
Miciah: There is no longer a grace period for the `PodsScheduled` status condition, because the `PodsScheduled` status condition no longer even exists after this PR.  

The point of this PR is to remove the `PodsScheduled` status condition because it was spuriously indicating a condition (some pods weren't scheduled) as a problem when really it wasn't necessarily a problem at all (the not-scheduled pods might actually be pods that are waiting to be garbage-collected).  However, the operator still reports the `DeploymentReplicasMinAvailable` status condition (which has a 60-second grace period, before and after this PR), which conveys the information that really does matter (whether there are _enough_ pods that are scheduled, running, and ready).

pkg/operator/controller/ingress/status.go
Miciah: I updated the error message as you suggested.  The Jira link is a bit unwieldy though.

pkg/operator/controller/ingress/status.go
```diff
@@ -204,18 +203,17 @@ func computeAllowedSourceRanges(service *corev1.Service) []operatorv1.CIDR {
...
-// current PodsScheduled status condition state by inspecting the PodScheduled
-// conditions of the pods associated with the deployment.
-func computeDeploymentPodsScheduledCondition(deployment *appsv1.Deployment, pods []corev1.Pod) operatorv1.OperatorCondition {
+// checkPodsScheduledForDeployment checks whether the given deployment has a
+// valid label selector and whether all of the deployment's pods are reporting
+// PodsScheduled=True.  This function returns an error value indicating the
+// result of that check.
+func checkPodsScheduledForDeployment(deployment *appsv1.Deployment, pods []corev1.Pod) error {
+	if deployment == nil {
+		return errors.New("System error detected: deployment was nil.  Please report this issue to Red Hat: https://issues.redhat.com/secure/CreateIssueDetails!init.jspa?pid=12332330&issuetype=1&components=12367900&priority=10300&customfield_12316142=26752")
```
gcs278, line 212: I have some concern this link may break in the future. Would it be better to just do `https://issues.redhat.com/`?

pkg/operator/controller/ingress/status_test.go
candita: So will this automatically continue to reconcile while there are not enough pods scheduled?

Miciah: Yeah, if there are not _enough available_ pods, then `DeploymentReplicasMinAvailable` will have status `False`, and the grace logic will queue reconciliation for the end of the condition's grace period.  This is another way in which `PodsScheduled` is superfluous:  `DeploymentReplicasMinAvailable` has a shorter grace period than `PodsScheduled` had, so if `DeploymentReplicasMinAvailable` is `False`, its grace period already would have triggered reconcilation (and still does trigger it) before `PodsScheduled`'s grace period would have triggered reconciliation.

pkg/operator/controller/ingress/status.go
Miciah: I think that that would lead to more misrouted bug reports.  

I could say "open a support case", but I don't even know the link for that, and neither would OKD users.

gcs278: :man_shrugging: this is fine, hopefully we never hit this logic

candita:
This test output for e2e-gcp-ovn-serial is unusual in that it mentions two bugs:

- https://issues.redhat.com/browse/OCPBUGS-3874
- https://issues.redhat.com/browse/OCPBUGS-3642

> ERRO[2022-12-08T03:23:26Z] Some steps failed:                           
> ...
>   * could not run steps: step e2e-gcp-ovn-serial failed: "e2e-gcp-ovn-serial" test steps failed: "e2e-gcp-ovn-serial" pod "e2e-gcp-ovn-serial-openshift-e2e-test" failed: the pod ci-op-jk9ifky1/e2e-gcp-ovn-serial-openshift-e2e-test failed after 1h31m9s (failed containers: test): ContainerFailed one or more containers exited
> Container test exited with code 1, reason Error
> ---
> 57746Z","updated_at":"2022-12-08T02:54:41.100513Z","deleted_at":null,"status":"ON_QA","last_change_time":"2022-11-28T10:13:32Z","summary":"masters repeatedly losing connection to API and going NotReady","affects_versions":["4.12","4.13"],"fix_versions":[],"components":["Machine Config Operator / platform-baremetal"],"labels":["sippy-watchlist","sippy-watchlist","trt","trt"],"url":"https://issues.redhat.com/browse/OCPBUGS-3874"},{"id":14935611,"key":"OCPBUGS-3642","created_at":"2022-11-14T18:07:08.62315Z","updated_at":"2022-12-08T02:54:40.527553Z","deleted_at":null,"status":"New","last_change_time":"2022-11-14T16:25:50Z","summary":"Readiness and Liveliness Probe errors in openshift-oauth-apiserver and test failures","affects_versions":["4.12"],"fix_versions":[],"components":["apiserver-auth"],"labels":[],"url":"https://issues.redhat.com/browse/OCPBUGS-3642"}]},{"Name":"[sig-network-edge] ns/openshift-authentication route/oauth-openshift disruption/ingress-to-oauth-server connection/new should be available throughout the test","Risk":{"Level":{"Name":"Low","Level":1},"Reasons":["This test has passed 10.53% of 19 runs on release 4.13 [amd64 gcp ha ovn serial] in the last week."]},"OpenBugs":[]},{"Name":"[sig-network-edge] ns/openshift-console route/console disruption/ingress-to-console connection/reused should be available throughout the test","Risk":{"Level":{"Name":"Low","Level":1},"Reasons":["This test has passed 21.05% of 19 runs on release 4.13 [amd64 gcp ha ovn serial] in the last week."]},"OpenBugs":[]}],"OverallRisk":{"Level":{"Name":"High","Level":10},"Reasons":["Maximum failed test risk: High"]},"OpenBugs":[]}

candita:
@Miciah There's a new frequent (but not permanent) failure in `TestUnmanagedDNSToManagedDNSInternalIngressController`.  Could this change have caused that?

PR #858, opened 2022-11-15
OPNET-133: Support remote worker

tsorya:
Adding node affinity rule to router deployement that will not allow
    router to run on remote worker node (node with custom role label)

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -592,6 +592,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		nodeSelector["node-role.kubernetes.io/master"] = ""
 	default:
 		nodeSelector["node-role.kubernetes.io/worker"] = ""
+		// Disabling running on remote workers.
+		if deployment.Spec.Template.Spec.Affinity == nil {
+			deployment.Spec.Template.Spec.Affinity = &corev1.Affinity{}
+		}
+		deployment.Spec.Template.Spec.Affinity.NodeAffinity = &corev1.NodeAffinity{RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+			NodeSelectorTerms: []corev1.NodeSelectorTerm{{
+				MatchExpressions: []corev1.NodeSelectorRequirement{
+					{
+						Key:      controller.RemoteWorkerLabel,
+						Operator: corev1.NodeSelectorOpNotIn,
+						Values:   []string{""},
+					},
+				},
+			}},
+		}}
 	}
 
 	if ci.Spec.NodePlacement != nil {
@@ -1301,6 +1316,15 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 				})
 			}
 		}
+		if affinity.NodeAffinity != nil {
+			terms := affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+			for _, term := range terms {
+				exprs := term.MatchExpressions
+				sort.Slice(exprs, func(i, j int) bool {
+					return cmpNodeMatchExpressions(exprs[i], exprs[j])
+				})
+			}
+		}
 	}
 	hashableDeployment.Spec.Template.Spec.Affinity = affinity
 	tolerations := make([]corev1.Toleration, len(deployment.Spec.Template.Spec.Tolerations))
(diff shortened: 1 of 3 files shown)


Miciah (commented):
You'll also need to update `hashableDeployment` and `deploymentConfigChanged`.  Please also add a unit test in `TestDeploymentConfigChanged`.  

Isn't having remote worker nodes already a supported feature?  Is the "node-role.kubernetes.io/remote-worker" label new?  This change implies that running router pods on nodes with that label is strictly prohibited, but if it was possible to do it before, then it would be a regression to prohibit it now.

tsorya:
It was not possible previously as we don't support remote worker at all for now

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1417,6 +1441,25 @@ func cmpMatchExpressions(a, b metav1.LabelSelectorRequirement) bool {
 	return false
 }
 
+// cmpMatchExpressions is a helper for hashableDeployment.
+func cmpNodeMatchExpressions(a, b corev1.NodeSelectorRequirement) bool {
```
Miciah, line 1445:
```suggestion
// cmpNodeMatchExpressions is a helper for hashableDeployment.
func cmpNodeMatchExpressions(a, b corev1.NodeSelectorRequirement) bool {
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -592,6 +592,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		nodeSelector["node-role.kubernetes.io/master"] = ""
 	default:
 		nodeSelector["node-role.kubernetes.io/worker"] = ""
+		// Disabling running on remote workers
```
Miciah, line 595:
```suggestion
		// Disabling running on remote workers.
```
Is there any intention to allow scheduling routers on remote workers in some future release?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1612,6 +1627,24 @@ func TestDeploymentConfigChanged(t *testing.T) {
...
+											{
+												Key:      controller.RemoteWorkerLabel,
+												Operator: corev1.NodeSelectorOpNotIn,
+												Values:   []string{""},
+											},
+											{
+												Key:      controller.ControllerDeploymentHashLabel,
+												Operator: corev1.NodeSelectorOpIn,
+												Values:   []string{"1"},
+											},
```
Miciah, line 1643: This looks weird, but I guess you added it so that the "if the deployment template -affinity node selector expressions change ordering" test case would have at least two match expressions to work with, right?  Could you add a comment explaining that this second match expression is only here to satisfy the test case?

pkg/operator/controller/names.go
```diff
@@ -40,6 +40,8 @@ const (
 	// DefaultCanaryNamespace is the default namespace for
 	// the ingress canary check resources.
 	DefaultCanaryNamespace = "openshift-ingress-canary"
+
+	RemoteWorkerLabel = "node.openshift.io/remote-worker"
```
Miciah, line 44: Would you mind adding godoc for this definition?  

Defining the label here is fine as a temporary measure, but will it eventually be defined in openshift/api or the like?

Miciah (commented):
I have some minor suggestions or questions around comments.  Looks fine overall.

pkg/operator/controller/names.go
tsorya: I hope yes, currently we want to start using it and there is no actual need to add it to openshift-api right a way. 
But yes, when we will start increasing usage, we will move it to openshift-api

Miciah:
Thanks!
/approve
/lgtm

Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/858/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1600820795783778304) on deprovisioning, which is a known problem with CI on AWS.  Because the job failed due to a known issue that is unrelated to this PR and because the e2e-azure-operator and e2e-gcp-operator jobs succeeded, I'm overriding the e2e-aws-operator job.

lihongan:
/label qe-approved
follow pre-merge testing and tested with cluster-bot (`launch openshift/cluster-ingress-operator#858`) and passed, we can see the new nodeAffinity is added to router-default deployment, see
```
oc -n openshift-ingress get deployment/router-default -oyaml
---
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node.openshift.io/remote-worker
                operator: NotIn
                values:
                - ""
```
after adding label `node.openshift.io/remote-worker=` to one of the worker and scale ingresscontroller/default to 3, then we can see one of router pods cannot be scheduled to that node
```
$ oc -n openshift-ingress get pod -owide
NAME                              READY   STATUS    RESTARTS   AGE     IP            NODE                                        NOMINATED NODE   READINESS GATES
router-default-574bd8cd79-6vzfm   0/1     Pending   0          4m40s   <none>        <none>                                      <none>           <none>
router-default-574bd8cd79-mgd9b   1/1     Running   0          49m     10.131.0.6    ip-10-0-133-12.us-east-2.compute.internal   <none>           <none>
router-default-574bd8cd79-xv992   1/1     Running   0          6m42s   10.129.2.17   ip-10-0-128-53.us-east-2.compute.internal   <none>           <none>
```

PR #534, opened 2021-01-19
Bug 1899941: Override termination grace period for the router deployment

frobware:
Adds the `unsupported.do-not-use.openshift.io/override-liveness-grace-period-seconds` annotation (set to 10s) to the router deployment. This will override `terminationGracePeriodSeconds` which is currently set to 1 hour but needs the following workaround before the unsupported annotation will take precedence: https://github.com/openshift/kubernetes/pull/527

This is a specific openshift/internal workaround; there is an upstream issue/discussion which considers a broader solution: https://github.com/kubernetes/kubernetes/issues/64715#issuecomment-756244044.

PR #863, opened 2022-12-09
OCPBUGS-4703: Replace liveness-grace-period-seconds annotation

Miciah:
Remove the "unsupported.do-not-use.openshift.io/override-liveness-grace-period-seconds" annotation from router pods, and instead set `terminationGracePeriodSeconds` on the pods' liveness probes.

Previously, cluster-ingress-operator started setting this OpenShift-specific annotation on router pods as a short-term measure to configure the liveness probe's grace period in order to fix [BZ#1899941](https://bugzilla.redhat.com/show_bug.cgi?id=1899941).  This annotation was implemented by [a carry patch](https://github.com/openshift/kubernetes/pull/527) in [openshift/kubernetes](https://github.com/openshift/kubernetes).  Since then, upstream Kubernetes has added the `terminationGracePeriodSeconds` field for configuring probes.  Using the new API field enables us to remove the carry patch.

Follow-up to #534.

* `assets/router/deployment.yaml`: Remove the unsupported annotation, and specify `terminationGracePeriodSeconds` on the liveness probe.
* `pkg/operator/controller/ingress/deployment.go` (`deploymentConfigChanged`): Remove annotations that the operator owns when those annotations are not expected to be set.
(`copyProbe`): Copy the `terminationGracePeriodSeconds` field.
* `pkg/operator/controller/ingress/deployment_test.go` (`TestDesiredRouterDeploymentSpecTemplate`): Expect the unsupported annotation not to be set.  Expect `terminationGracePeriodSeconds` to be set on the liveness probe.
(`TestDeploymentConfigChanged`): Verify that `deploymentConfigChanged` correctly sets the liveness probe's termination grace period seconds and removes the unsupported annotation.


pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1438,6 +1439,15 @@ func TestDeploymentConfigChanged(t *testing.T) {
 			},
 			expect: true,
 		},
+		{
```
gcs278, line 1442: Should we add a test case for detecting the changing of `deployment.Spec.Template.Spec.Containers[0].LivenessProbe.terminationGracePeriodSeconds`? Or maybe it could be added to one of the test cases above?

Miciah: I added a test case for it.

gcs278:
Seems like a straightforward change. Should I also do an approve on this?
/lgtm

Miciah:
I'm fine if you want to approve this since we didn't assign a second reviewer.  * grin * .

gcs278:
:ship: 
/approve

melvinjoseph86:
Using the pre merge verification the bug is verified and steps are attached in the comments of https://issues.redhat.com/browse/OCPBUGS-4703 
/label qe-approved

PR #694, opened 2022-01-14
Specify host port for host networking strategy

arjunrn:
When the load balancer strategy is hostnetworking allow the user to
specify which ports on the host the pods should use. Also fixup the
container ports.

PR #864, opened 2022-12-10
OCPBUGS-4573: Target metrics port by name in internal service

Miciah:
#### `ensureInternalIngressControllerService`: Reformat

Reformat `ensureInternalIngressControllerService` to follow the standard pattern for "ensure" functions.

* `pkg/operator/controller/ingress/controller.go` (`ensureIngressController`): Expect a Boolean return value from `ensureInternalIngressControllerService`, indicating whether the service exists.
* `pkg/operator/controller/ingress/internal_service.go` (`ensureInternalIngressControllerService`): Add a Boolean return value.  Use a `switch` statement, and add a to-do comment for handling updates.
(`currentInternalIngressControllerService`): Add a Boolean return value.

#### `ensureInternalIngressControllerService`: Updates

Add logic to `ensureInternalIngressControllerService` to update the service when an update is required.

* `pkg/operator/controller/ingress/internal_service.go` (`ensureInternalIngressControllerService`): Use the new `updateInternalService` method to update the service as needed.
(`updateInternalService`): Check whether the given ClusterIP service needs to be updated, and update it if so, using the new `internalServiceChanged` function.
(`managedInternalServiceAnnotations`): New variable with the set of annotation keys for annotations that the operator manages for its internal router services.
(`internalServiceChanged`): New function.  Check whether the current internal service needs to be updated, and update it if it does, using the new `managedInternalServiceAnnotations` variable.
* `pkg/operator/controller/ingress/internal_service_test.go`: New file.
(`Test_desiredInternalIngressControllerService`): New test.  Verify that `desiredInternalIngressControllerService` returns the expected service.
(`Test_internalServiceChanged`): New test.  Verify that `internalServiceChanged` correctly detects changes and performs updates.

#### `service-internal.yaml`: Target metrics port by name

Use the "metrics" port name instead of port number 1936 for the router internal service's metrics port's target.

Before this change, the router's internal service's metrics port always targeted port 1936 on the router pod.  However, the router pod's metrics port can be customized, and so it is not necessarily port 1936.  As a consequence, the service's metrics port didn't work when the router pod's metrics port was customized.  The router pod's metrics port always has the name "metrics", so this PR changes the service to reference the port by name to avoid breaking when the port number changes.

Follow-up to #694..

* `assets/router/service-internal.yaml`: Use the "metrics" port name instead of port 1936 for the metrics port's target.
* `pkg/operator/controller/ingress/internal_service_test.go` (`Test_desiredInternalIngressControllerService`): Expect the metrics port to reference the "metrics" target port by name.
(`Test_internalServiceChanged`): Add a test case for changing the "metrics" port's target port from an integer to a string.


pkg/operator/controller/ingress/internal_service_test.go
```diff
@@ -116,6 +116,22 @@ func Test_internalServiceChanged(t *testing.T) {
...
 			expect: true,
 		},
+		{
+			description: "if .spec.ports changes by changing the metrics port's target port from an integer to a string",
+			mutate: func(svc *corev1.Service) {
+				foundPort := false
+				for i := range svc.Spec.Ports {
+					if svc.Spec.Ports[i].Name == "metrics" {
+						svc.Spec.Ports[i].TargetPort = intstr.FromString("metrics")
+						foundPort = true
```
frobware, line 126: Can we return early here? if so, then we don't need `foundPort`. If there's no match we can just panic.

frobware:
```
{  openshift cluster install failed with cluster bootstrap}
```

/test e2e-gcp-ovn-serial

pkg/operator/controller/ingress/internal_service.go
```diff
@@ -36,7 +40,11 @@ func (r *reconciler) ensureInternalIngressControllerService(ic *operatorv1.Ingre
 		log.Info("created internal ingresscontroller service", "service", desired)
 		return r.currentInternalIngressControllerService(ic)
 	case have:
-		// TODO Handle updates.
+		if updated, err := r.updateInternalService(current, desired); err != nil {
+			return true, current, fmt.Errorf("failed to update internal service: %v", err)
```
frobware, line 44: Just confirming that this returns true because the internal service already exists and it doesn't matter if the update fails. True?

Miciah: True.  The Boolean return value only indicates whether the service exists.  I copied the pattern from other "ensure" methods, but I neglected to put that in the godoc.  I've now updated the godoc to explain this.

frobware:
```
{  1 events happened too frequently

event happened 21 times, something is wrong: node/ci-op-rz7hm0sh-9dec8-5hnrs-worker-c-bx756 - reason/ErrorReconcilingNode roles/worker [k8s.ovn.org/node-chassis-id annotation not found for node ci-op-rz7hm0sh-9dec8-5hnrs-worker-c-bx756, macAddress annotation not found for node "ci-op-rz7hm0sh-9dec8-5hnrs-worker-c-bx756" , k8s.ovn.org/l3-gateway-config annotation not found for node "ci-op-rz7hm0sh-9dec8-5hnrs-worker-c-bx756"]}
```

/retest-required

PR #865, opened 2022-12-12
OCPBUGS-3522: Include recent errors in canary checks fail

rfredette:
The ingress canary sets the status condition `CanaryChecksSucceeding` to false when 5 or more successive checks fail. With this PR, the associated status message will include the error messages from the ~~5~~ 3 most recent failures

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -43,6 +44,9 @@ const (
 	// canaryCheckFailureCount is how many successive failing canary checks should
 	// be observed before the default ingress controller goes degraded.
 	canaryCheckFailureCount = 5
+	// canaryFailingNumErrors is how many error messages to include in the
+	// CanaryChecksSucceeding status condition when checks are failing
+	canaryFailingNumErrors = 3
 
 	// CanaryRouteRotationAnnotation is an annotation on the default ingress controller
 	// that specifies whether or not the canary check loop should periodically rotate
@@ -221,6 +225,15 @@ func (r *reconciler) isCanaryRouteRotationEnabled() bool {
 	return r.enableCanaryRouteRotation
 }
 
+type timestampedError struct {
+	timestamp time.Time
+	err       error
+}
+
+func (err timestampedError) Error() string {
+	return err.err.Error()
+}
+
 func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	// Keep track of how many canary checks have passed
 	// so the route endpoint can be periodically cycled
@@ -230,6 +243,8 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	// Keep track of successive canary check failures
 	// for status reporting.
 	successiveFail := 0
+
+	errors := []timestampedError{}
 
 	go wait.Until(func() {
 		// Get the current canary route every iteration in case it has been modified
@@ -282,9 +297,10 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			log.Error(err, "error performing canary route check")
 			SetCanaryRouteReachableMetric(getRouteHost(route), false)
 			successiveFail += 1
+			errors = append(errors, timestampedError{err: err, timestamp: time.Now()})
 			// Mark the default ingress controller degraded after 5 successive canary check failures
 			if successiveFail >= canaryCheckFailureCount {
-				if err := r.setCanaryFailingStatusCondition(); err != nil {
+				if err := r.setCanaryFailingStatusCondition(errors); err != nil {
 					log.Error(err, "error updating canary status condition")
 				}
 			}
@@ -296,6 +312,7 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			log.Error(err, "error updating canary status condition")
 		}
 		successiveFail = 0
+		errors = []timestampedError{}
 		// Only increment checkCount if periodic canary route
 		// endpoint rotation is enabled to prevent unbounded
 		// integer growth.
@@ -307,15 +324,63 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	return nil
 }
 
-func (r *reconciler) setCanaryFailingStatusCondition() error {
+func (r *reconciler) setCanaryFailingStatusCondition(errors []timestampedError) error {
+	errorStrings := deduplicateErrorStrings(errors, time.Now())
+	if len(errorStrings) > canaryFailingNumErrors {
+		errorStrings = errorStrings[len(errorStrings)-canaryFailingNumErrors:]
+	}
 	cond := operatorv1.OperatorCondition{
 		Type:    ingresscontroller.IngressControllerCanaryCheckSuccessConditionType,
 		Status:  operatorv1.ConditionFalse,
 		Reason:  "CanaryChecksRepetitiveFailures",
-		Message: "Canary route checks for the default ingress controller are failing",
+		Message: fmt.Sprintf("Canary route checks for the default ingress controller are failing. Last %d error messages:\n%s", len(errorStrings), strings.Join(errorStrings, "\n")),
 	}
 
 	return r.setCanaryStatusCondition(cond)
+}
+
+type dedupCounter struct {
+	Count           int
+	FirstOccurrence time.Time
+}
+
+// deduplicateErrorStrings takes a slice of errors in chronological order, prunes the list for unique error messages
+// It returns a slice of the error message as strings, with any string seen more than once incuding how many times it
+// was seen.
+// The chronological order is preserved, based on the last time each error message was seen
+func deduplicateErrorStrings(errors []timestampedError, now time.Time) []string {
+	encountered := map[string]dedupCounter{}
+	uniqueErrors := []string{}
+	// Iterate over the list of errors from newest to oldest, keeping track of how many times each error message was
+	// seen. This iteration order makes sure we preserve order based on the newest time each message was seen.
+	for i := len(errors) - 1; i >= 0; i-- {
+		err := errors[i]
+		if errCounter := encountered[err.Error()]; errCounter.Count >= 1 {
+			errCounter.Count += 1
+			if errCounter.FirstOccurrence.After(err.timestamp) {
+				errCounter.FirstOccurrence = err.timestamp
+			}
+			encountered[err.Error()] = errCounter
+			continue
+		}
+		encountered[err.Error()] = dedupCounter{
+			Count:           1,
+			FirstOccurrence: err.timestamp,
+		}
+		uniqueErrors = append(uniqueErrors, err.Error())
+	}
+	ret, j := make([]string, len(uniqueErrors)), len(uniqueErrors)-1
+	// Now that all error messages have been ordered by their most recent occurrence, reverse the list to switch it to
+	// cronological order, and append a message indicating how often an error has occurred if it has more than one
+	// occurrence.
+	for _, err := range uniqueErrors {
+		if counter := encountered[err]; counter.Count > 1 {
+			err = fmt.Sprintf("%s (x%d over %v)", err, counter.Count, now.Sub(counter.FirstOccurrence).Round(time.Second))
+		}
+		ret[j] = err
+		j--
+	}
+	return ret
 }
 
 func (r *reconciler) setCanaryPassingStatusCondition() error {
--- a/pkg/operator/controller/canary/controller_test.go
+++ b/pkg/operator/controller/canary/controller_test.go
@@ -1,7 +1,10 @@
 package canary
 
 import (
+	"fmt"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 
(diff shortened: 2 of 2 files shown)


rfredette:
seeing a lot of install failures on the failed CI runs.

/retest

rfredette:
Sample output:
```
    Last Transition Time:  2022-12-21T19:45:22Z
    Message:               Canary route checks for the default ingress controller are failing. Last 5 errors:
error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: certificate required
error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: certificate required
error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: certificate required
error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: certificate required
error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-33z1pqk-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: certificate required
    Reason:  CanaryChecksRepetitiveFailures
    Status:  False
    Type:    CanaryChecksSucceeding
```

Is this too verbose? I chose to include the last 5 errors since that's the minimum that will cause `CanaryChecksSucceeding` to report false, but I can tune it to report fewer if it's too much

lihongan:
@rfredette can we update the check interval and threshold to make failure detection more sensitive? we often see that authentication/console already reported error for a while (almost 2 minutes) but ingress is still good. see example
```
$ oc get co authentication console ingress
NAME             VERSION                                                   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication   4.12.0-0.ci.test-2022-12-22-014836-ci-ln-h6thnl2-latest   False       False         False      113s    OAuthServerRouteEndpointAccessibleControllerAvailable: Get "https://oauth-openshift.apps.ci-ln-h6thnl2-76ef8.origin-ci-int-aws.dev.rhcloud.com/healthz": EOF
console          4.12.0-0.ci.test-2022-12-22-014836-ci-ln-h6thnl2-latest   False       False         False      106s    RouteHealthAvailable: failed to GET route (https://console-openshift-console.apps.ci-ln-h6thnl2-76ef8.origin-ci-int-aws.dev.rhcloud.com): Get "https://console-openshift-console.apps.ci-ln-h6thnl2-76ef8.origin-ci-int-aws.dev.rhcloud.com": EOF
ingress          4.12.0-0.ci.test-2022-12-22-014836-ci-ln-h6thnl2-latest   True        False         False      39m     

```

pkg/operator/controller/canary/controller.go
```diff
@@ -230,6 +230,8 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	// for status reporting.
 	successiveFail := 0
 
+	errors := []error{}
```
lihongan, line 233: maybe it's better to add comments for variable `errors` ?

lihongan:
From my understanding, the last 5 canary check errors should be same in most of the cases, so maybe we don't need to list 5 errors.
In the https://issues.redhat.com/browse/OCPBUGS-3522, the Expected results is:
```
Canary route checks for the default ingress controller are failing: ${ERROR_MESSAGE}. ${POSSIBLY_ALSO_MORE_TROUBLESHOOTING_IDEAS?}
```
so maybe we should focus on how to improve the detailed error message (which can help troubleshooting).

gcs278:
@rfredette what's the status of this PR? Should I review or are you going to revisit this solution?

rfredette:
@gcs278 It shouldn't change much in terms of the logic, but I'm planning to reduce the number of error messages shown (I think the last 3 should be reasonable), and include some message about checking the canary logs for more info.

I'd say don't worry about the review yet; I'll ping you once I make that change, which should be in the next day or two

rfredette:
> can we update the check interval and threshold to make failure detection more sensitive? we often see that authentication/console already reported error for a while (almost 2 minutes) but ingress is still good.

@lihongan I think that's a reasonable ask, but it's out of scope for this particular issue. I think the right way to go about that is to create a story for it in the Maintenance & Debugability feature on Jira, but let me verify that.

rfredette:
The latest change makes the error message more concise in 2 ways. First, it deduplicates the error messages and includes the total number of times a given error was seen, and second, it only includes 3 unique error messages, instead of 5. Because of how I was able to test this, I only got one error message, but here's an example: 
```
  - lastTransitionTime: "2023-01-11T01:20:40Z"
    message: |-
      Canary route checks for the default ingress controller are failing. Last 1 error messages:
      [Seen 7 times] error sending canary HTTP request to "canary-openshift-ingress-canary.apps.ci-ln-h5z9pc2-72292.origin-ci-int-gce.dev.rhcloud.com": Get "https://canary-openshift-ingress-canary.apps.ci-ln-h5z9pc2-72292.origin-ci-int-gce.dev.rhcloud.com": remote error: tls: handshake failure
    reason: CanaryChecksRepetitiveFailures
    status: "False"
    type: CanaryChecksSucceeding
```

Unfortunately I can't think of a way to generate some useful comment in addition to the error message, so I think this will have to do for increased debuggability.
@gcs278 I think this is ready for review, so when you get a chance, please take a look

rfredette:
several jobs got a 503 error during the build; retrying.
/retest

rfredette:
Test failures look unrelated.

/retest-required

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +311,45 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
 		Status:  operatorv1.ConditionFalse,
 		Reason:  "CanaryChecksRepetitiveFailures",
-		Message: "Canary route checks for the default ingress controller are failing",
+		Message: fmt.Sprintf("Canary route checks for the default ingress controller are failing. Last %d error messages:\n%s", len(errorStrings), strings.Join(errorStrings, "\n")),
 	}
 
 	return r.setCanaryStatusCondition(cond)
 }
 
+func deduplicateErrorStrings(errors []error) []string {
```
gcs278, line 329: is this worth a unit test?

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +311,45 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
+		}
+		encountered[errors[i].Error()] = 1
+		ret = append(ret, errors[i].Error())
+	}
+	// Now that all error messages have been tallied, add a comment to each
+	// message that was seen more than once, so the reader can tell how often
+	// each error has occurred
+	for i := range ret {
+		if encountered[ret[i]] > 1 {
+			ret[i] = fmt.Sprintf("[Seen %d times] %s", encountered[ret[i]], ret[i])
```
gcs278, line 347: Is `[Seen %d times]` used in other places? I looked around for other examples of this, and Kubernetes uses:
`(x3 over 70s)` in their describe status.

I think `[Seen %d times]` is pretty descriptive, and trying to track timestamps is a bit much (though would be pretty neat), but just was curious if you found any other examples of this?

gcs278 (commented):
No major issues, just some questions.

pkg/operator/controller/canary/controller.go
rfredette: I haven't seen the `[Seen %d times]` format used elsewhere; I was thinking of the `(x3 over 70s)` style when I implemented it, but I guess my memory was slightly off :)

I don't think we keep timestamps of the errors at the moment, and I'm not sure how feasible it would be to implement in this bug fix. Let me take another look, though. If we do have timestamps (or if it would be easy to add), it wouldn't be too hard to track count, the first timestamp, and the last timestamp as it iterates over the list, and that would be enough for the "times over duration" style

pkg/operator/controller/canary/controller.go
rfredette: I'm not sure; this is a private function at the moment, and I know we were talking about the philosophy of "only test public functions" recently (although I don't recall if we came to any conclusion on whether to follow that or not)

I do think this may be useful enough to try to move it to somewhere like library-go if we wanted to; When I started on this I thought there might already be a deduplication function somewhere like that I could leverage, but I didn't find anything.

If we did move it or make it public, I would definitely want a unit test, though.

gcs278: Yea I agree, I truthfully don't know what I'm doing when it comes to unit testing :man_shrugging: 

The problem is the `Reconcile` function isn't being tested here either, and it may be a  difficult one to ever right a unit test for since it attempts to curl a URL. Though I guess you could stand up a GO echo server, hack the canary route to point to this localhost GO Server, and simulate failures.

That'd be worth doing, but that's a backlog thing. Not going to hold up this PR over that.

pkg/operator/controller/canary/controller.go
gcs278: Yea let me know, i don't really care all that much about the exact message, but if it's reasonable to do, that'd be a bit more consistent and kind of neat.

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +311,45 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 	return nil
 }
 
-func (r *reconciler) setCanaryFailingStatusCondition() error {
+func (r *reconciler) setCanaryFailingStatusCondition(errors []error) error {
+	errorStrings := deduplicateErrorStrings(errors)
+	if len(errorStrings) > 3 {
```
alebedev87, line 316: Why 3 errors? I thought we wanted to show up to 5 deduplicated error messages. At least the PR description says so.

pkg/operator/controller/canary/controller.go
alebedev87: We have a lot of cases when only `Reconcile` function is public in a package. I agree that the reconcile functions are indeed often not very approachable. Also they have plenty of the test variations which makes the unit testing of `Reconcile` even more difficult.
If we follow strictly the "only test public functions" philosophy, I think that we risk to either stay without the unit tests due to the unnecessary complexity of Reconcile testing for smaller changes or with the minimal coverage.

The way I see it: if the test of the public function is too complex for a smaller change or if it doesn't exist at all - better to have some testing than no testing at all.

rfredette: I think that's fair. I'll add a test for this

pkg/operator/controller/canary/controller.go
rfredette: Initially I had done 5 (since it takes 5 failures for canary to go degraded anyway), but I think it's too much to put in the status condition. I think 3 messages is more manageable, and still gives you some indication of if it's one particular error or multiple that's causing issues.

That said, I will update the description to match the current intended behavior.

pkg/operator/controller/canary/controller.go
rfredette: I took another look and while I think it would be possible to track timestamps, I don't want to over-complicate this. For now I'd prefer to leave this as-is.

rfredette:
Writing the unit test made me realize that `DeduplicateErrorStrings` was ordering its results based on which string was seen first, when we really want it to order the results based on which errors have been seen most recently. This latest commit should fix that.

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +311,48 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
+	ret := []string{}
+	// Iterate over the list of errors from newest to oldest, keeping track of how many times each error message was
+	// seen
+	for i := len(errors) - 1; i >= 0; i-- {
+		if encountered[errors[i].Error()] >= 1 {
+			encountered[errors[i].Error()] += 1
+			continue
+		}
+		encountered[errors[i].Error()] = 1
+		ret = append([]string{errors[i].Error()}, ret...)
```
rfredette, line 344: I had to iterate over the input list backwards to keep the most recent errors at the end of the list, but I'm not sure which is better: (a) building the list by "pushing" new items onto the front of the list (the way it's implemented here), or (b) building it from newest to oldest, then reverse it just before returning.

pkg/operator/controller/canary/controller.go
alebedev87: What do you think about putting this into a constant (similar to `canaryCheckFailureCount`)?

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +311,48 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
 	}
 
 	return r.setCanaryStatusCondition(cond)
 }
 
+// DeduplicateErrorStrings takes a slice of errors in chronological order, prunes the list for unique error messages
+// It returns a slice of the error message as strings, with any string seen more than once incuding how many times it
+// was seen
+// The chronological order is preserved, based on the last time each error message was seen
+func DeduplicateErrorStrings(errors []error) []string {
```
alebedev87, line 333: Since it's now a public function and quite a generic one, it may deserve to be put into `pkg/util` package. That would justify the recently added unit test even more :)

gcs278: +1 to Andrey, but what's the motivation behind making this public now? Is it in anticipation for re-use? I suppose I assumed functions should be private unless they are used by other packages.

pkg/operator/controller/canary/controller_test.go
```diff
@@ -145,3 +147,74 @@ func TestCycleServicePort(t *testing.T) {
 		}
 	}
 }
+
+func TestDeduplicateErrorStrings(t *testing.T) {
```
gcs278, line 151: *nit* I think we are trying to use `Test_[FunctionName]` in our unit tests.

pkg/operator/controller/canary/controller.go
gcs278: If we are talking efficiency, isn't `a` slightly more efficient? I think `a` is `O(n)` where as `b` is `O(2n)`, though `O(2n) == O(n)` as it scales to infinity, so it really doesn't matter a whole lot. Just trying to practice my O Notation skills :smiley:

rfredette: It depends I think on how append actually works in this case. `append(slice, newItem)` is usually O(1), but the `ret...`syntax here is more like `append(slice, ret[0], ret[1], ... , ret[n])`. If it's unoptimized, that could be an O(n) operation, and since that's potentially happening for every iteration of the for loop, this would be O(n<sup>2</sup>)

pkg/operator/controller/canary/controller.go
rfredette: I had assumed `pkg/operator/controller/canary/controller_test.go` was in the `test` package (which I believe is the idiomatic package to use in `*_test.go` files), which would require `DeduplicateErrorStrings` be public if it was to be tested. However, now that I look at it that's not the case; `controller_test.go` is in package `canary`.

I don't mind moving it to `pkg/util`, but since it's only used here at the moment, it would also be reasonable to just make it private again. @alebedev87 @gcs278 do you have a preference?

pkg/operator/controller/canary/controller_test.go
rfredette: Is that a new initiative? From what I can tell we only do the `Test_[FunctionName]` syntax for tests of private functions, but I can change it if it's preferred for all new tests.

pkg/operator/controller/canary/controller.go
rfredette: I think that's a good idea. I will include a constant for it in my next pass.

pkg/operator/controller/canary/controller_test.go
gcs278: Yea just checked - it was for all unit test functions, regardless of the scope of the function.

pkg/operator/controller/canary/controller.go
gcs278: Yea I see that now. I don't completely understand the mechanics of `append` and how it's optimized, but sounds like `b` would be more efficient, being `O(n)` for the building of the unordered slice, then another `O(n)` to reverse, so `2O(n) == O(n)`.

pkg/operator/controller/canary/controller.go
alebedev87: I'm fine to keep it where it is, private and unit tested.

gcs278: +1 for just keeping private and unit tested.

pkg/operator/controller/canary/controller.go
gcs278: Did you mean to do this in your last update? or still working on it?

rfredette: I meant to do this in the last update, but looks like I missed it. I pushed a change that should fix it just a moment ago.

pkg/operator/controller/canary/controller.go
```diff
@@ -306,17 +314,54 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
+	for i := range ret {
+		if encountered[ret[i]] > 1 {
+			ret[i] = fmt.Sprintf("[Seen %d times] %s", encountered[ret[i]], ret[i])
+		}
+	}
+	// Reverse the list so that it can be returned in oldest to newest order
+	for i, j := 0, len(ret)-1; i < j; i, j = i+1, j-1 {
+		ret[i], ret[j] = ret[j], ret[i]
+	}
+	return ret
```
alebedev87, line 362: We can add to some result slice of the same size as `ret` at the the same time we iterate over `ret`.Something like this should save us 1 run over the slice and the cost of the permutations:
```suggestion
	// Now that all error messages have been tallied, prepend a comment to each message that was seen more than once, so
	// the reader can tell how often each error has occurred
	res, j := make([]string, len(ret)), len(ret)-1
	for i := range ret {
		if encountered[ret[i]] > 1 {
			ret[i] = fmt.Sprintf("[Seen %d times] %s", encountered[ret[i]], ret[i])
		}
		res[j] = ret[i]
		j--
	}
	return res
```

pkg/operator/controller/canary/controller_test.go
```diff
@@ -145,3 +147,76 @@ func TestCycleServicePort(t *testing.T) {
...
+				fmt.Errorf("foo"),
+				fmt.Errorf("quux"),
+			},
+			ExpectedResult: []string{
+				"[Seen 2 times] bar",
+				"baz",
+				"[Seen 6 times] foo",
+				"[Seen 3 times] quux",
+			},
+		},
```
alebedev87, line 211: How about adding a test case like the one you showed me?
```
foo
bar
foo
baz
```
which should result into
```
bar
[Seen 2x] foo
baz
```
and not into:
```
[Seen 2x] foo
bar
baz
```

alebedev87 (commented):
Overall LGTM, just 2 last minor remarks.

rfredette:
In the latest commit, I added timestamp tracking to the errors and switched the `[Seen 3 times]` part of the message to the `(x3 in 35s)` style that is used elsewhere. I also reworked the deduplication code to reduce the number of times it loops over the list of errors.

gcs278:
Looks good to me. Since I'm the other reviewer, I'll approve it.
/approve
/lgtm

PR #871, opened 2022-12-21
OCPBUGS-3356: E2E test for cookie length truncation

rfredette:
This PR adds a new test, `TestCookieLen`, which verifies that cookies in `spec.logging.access.httpCaptureCookies` are truncated to the specified max length.

This tests the fix in openshift/router#436

### DIFF
--- a/test/e2e/client_tls_test.go
+++ b/test/e2e/client_tls_test.go
@@ -1408,19 +1408,6 @@ func verifyCRLs(t *testing.T, pod *corev1.Pod, expectedCRLs map[string]*x509.Rev
 	return true, nil
 }
 
-func getPods(t *testing.T, cl client.Client, deployment *appsv1.Deployment) (*corev1.PodList, error) {
-	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
-	if err != nil {
-		return nil, fmt.Errorf("deployment %s has invalid spec.selector: %w", deployment.Name, err)
-	}
-	podList := &corev1.PodList{}
-	if err := cl.List(context.TODO(), podList, client.MatchingLabelsSelector{Selector: selector}); err != nil {
-		t.Logf("failed to list pods for deployment %q: %v", deployment.Name, err)
-		return nil, err
-	}
-	return podList, nil
-}
-
 func getActiveCRLs(t *testing.T, clientPod *corev1.Pod) ([]*x509.RevocationList, error) {
 	t.Helper()
 	cmd := []string{
--- a/test/e2e/tuning_options_test.go
+++ b/test/e2e/tuning_options_test.go
@@ -4,8 +4,10 @@
 package e2e
 
 import (
+	"bufio"
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -14,9 +16,12 @@ import (
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/kubernetes"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -318,5 +323,226 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
 				t.Errorf("Expected value for \"tcp-request inspect-delay\" to be %v, got %v", []string{inspectDelayDefault, inspectDelayDefault}, values)
 			}
 		}
+	}
+}
+
+// TestCookieLen verifies that when logging is enabled and a cookie capture is specified, any captured cookies are
+// truncated at the expected max length.
+func TestCookieLen(t *testing.T) {
+	testCases := []struct {
+		name      string
+		maxLength int
+	}{
+		// 63 bytes is maximum cookie length allowed when tune.http.cookielen is unset in the HAProxy config.
+		{"63 bytes", 63},
+		// 64 bytes is the smallest number requires tune.http.cookielen to be set.
+		{"64 bytes", 64},
+		// 128 bytes is an arbitrary number between the 64 and 1024 bytes. It requires tune.http.cookielen to be set,
+		// but is between the boundaries.
+		{"128 bytes", 128},
+		// The upper bound. 1024 bytes is the maximum allowed by the OpenShift API.
+		{"1024 bytes", 1024},
+	}
+	testNamespace := "openshift-ingress"
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			maxLength := testCase.maxLength
+			t.Parallel()
+			icName := types.NamespacedName{
+				Namespace: operatorNamespace,
+				Name:      names.SimpleNameGenerator.GenerateName("test-cookielen-"),
+			}
+			domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+			ic := newPrivateController(icName, domain)
+
+			cookieName := "X-foo-bar"
+			ic.Spec.Logging = &operatorv1.IngressControllerLogging{
+				Access: &operatorv1.AccessLogging{
+					Destination: operatorv1.LoggingDestination{
+						Type: "Container",
+						Container: &operatorv1.ContainerLoggingDestinationParameters{
+							MaxLength: 2048,
+						},
+					},
+					HTTPCaptureCookies: []operatorv1.IngressControllerCaptureHTTPCookie{{
+						IngressControllerCaptureHTTPCookieUnion: operatorv1.IngressControllerCaptureHTTPCookieUnion{
+							MatchType: operatorv1.CookieMatchTypeExact,
+							Name:      cookieName,
+						},
+						MaxLength: maxLength,
+					}},
+					HttpLogFormat: "cookie: %CC",
+				},
+			}
+
+			if err := kclient.Create(context.TODO(), ic); err != nil {
+				t.Fatalf("Failed to create ingresscontroller %s: %v", icName, err)
+			}
+			defer assertIngressControllerDeleted(t, kclient, ic)
+
+			if err := waitForIngressControllerCondition(t, kclient, 2*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+				t.Fatalf("Timed out waiting for ingresscontroller %s to become available: %v", icName.Name, err)
+			}
+
+			echoPod := buildEchoPod(names.SimpleNameGenerator.GenerateName("echo-pod-"), testNamespace)
+			if err := kclient.Create(context.TODO(), echoPod); err != nil {
+				t.Fatalf("failed to create pod %s/%s: %v", echoPod.Namespace, echoPod.Name, err)
+			}
+			defer assertDeleted(t, kclient, echoPod)
+
+			echoService := buildEchoService(echoPod.Name, echoPod.Namespace, echoPod.ObjectMeta.Labels)
+			if err := kclient.Create(context.TODO(), echoService); err != nil {
+				t.Fatalf("failed to create service %s/%s: %v", echoService.Namespace, echoService.Name, err)
+			}
+			defer assertDeleted(t, kclient, echoService)
+
+			echoRoute := buildRoute(echoPod.Name, echoPod.Namespace, echoService.Name)
+			if err := kclient.Create(context.TODO(), echoRoute); err != nil {
+				t.Fatalf("failed to create route %s/%s: %v", echoRoute.Namespace, echoRoute.Name, err)
+			}
+			defer assertDeleted(t, kclient, echoRoute)
+
+			// Wait for echo pod to be ready
+			err := wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
+				if err := kclient.Get(context.TODO(), types.NamespacedName{Name: echoPod.Name, Namespace: echoPod.Namespace}, echoPod); err != nil {
+					t.Logf("failed to get client pod %q: %v", echoPod.Name, err)
+					return false, nil
+				}
+				for _, cond := range echoPod.Status.Conditions {
+					if cond.Type == corev1.PodReady {
+						return cond.Status == corev1.ConditionTrue, nil
+					}
+				}
+				return false, nil
+			})
+
+			if err != nil {
+				t.Fatalf("timed out waiting for pod %s to become ready: %v", echoPod.Name, err)
+			}
+
+			// Send a request through the router that includes the captured cookie, and verify that it's truncated as expected.
+
+			// Use the router image for the client pod since it includes curl.
+			routerDeployment := &appsv1.Deployment{}
+			routerDeploymentName := controller.RouterDeploymentName(ic)
+			if err := kclient.Get(context.TODO(), routerDeploymentName, routerDeployment); err != nil {
+				t.Fatalf("Failed to get router deployment for ingresscontroller %s: %v", icName.Name, err)
+			}
+
+			curlPodImage := ""
+			for _, container := range routerDeployment.Spec.Template.Spec.Containers {
+				if container.Name == "router" {
+					curlPodImage = container.Image
+					break
+				}
+			}
+			if curlPodImage == "" {
+				t.Fatalf("Failed to find container \"router\" in deployment %s", routerDeployment.Name)
+			}
+			curlPodName := types.NamespacedName{
+				Name:      names.SimpleNameGenerator.GenerateName("cookielen-curl-"),
+				Namespace: testNamespace,
+			}
+
+			icInternalService := &corev1.Service{}
+			icInternalServiceName := controller.InternalIngressControllerServiceName(ic)
+			if err := kclient.Get(context.TODO(), icInternalServiceName, icInternalService); err != nil {
+				t.Fatalf("failed to get service %q: %v", icInternalServiceName, err)
+			}
+
+			// Create a cookie that's one character over the max length of a captured cookie. Pad the cookie with a bunch of
+			// 'a's, and include an x at the end that should get truncated. The name of the cookie and the '=' are included in
+			// the max length, so account for that in the size of the padding.
+			padding := strings.Repeat("a", maxLength-len(cookieName)-1)
+			cookieString := fmt.Sprintf("%s=%sx", cookieName, padding)
+
+			curlArgs := []string{
+				"-b", cookieString,
+				// --resolve used this way makes curl send the request to our ingress controller's internal service address as
+				// if it was directed to it by a load balancer. This ensures the request is handled by our ingress controller
+				// without having to wait extra time for a load balancer to be provisioned.
+				"--resolve", fmt.Sprintf("%s:80:%s", echoRoute.Spec.Host, icInternalService.Spec.ClusterIP),
+			}
+
+			curlPod := buildCurlPod(curlPodName.Name, curlPodName.Namespace, curlPodImage, echoRoute.Spec.Host, "", curlArgs...)
+			if err := kclient.Create(context.TODO(), curlPod); err != nil {
+				t.Fatalf("failed to create pod %q: %v", curlPodName.Name, err)
+			}
+			defer assertDeleted(t, kclient, curlPod)
+
+			// Wait for the curl pod to complete its request
+			err = wait.PollImmediate(2*time.Second, 3*time.Minute, func() (bool, error) {
+				if err := kclient.Get(context.TODO(), curlPodName, curlPod); err != nil {
+					t.Logf("failed to get client pod %q: %v", curlPodName, err)
+					return false, nil
+				}
+				return curlPod.Status.Phase == corev1.PodSucceeded, nil
+			})
+
+			if err != nil {
+				t.Fatalf("timed out waiting for pod %q to transition to \"Succeeded\". current phase is %q. error: %v", curlPodName, curlPod.Status.Phase, err)
+			}
+
+			// Find the router pod for our ingress controller so we can check its logs. This ingress controller specifies
+			// replicas=1, so there will only be one.
+			routerPodList, err := getPods(t, kclient, routerDeployment)
+			if err != nil {
+				t.Fatalf("failed to get pods in deployment %s: %v", routerDeploymentName, err)
+			} else if len(routerPodList.Items) != 1 {
+				t.Fatalf("expected to find 1 pod in deployment %s, found %d", routerDeploymentName, len(routerPodList.Items))
+			}
+			routerPod := routerPodList.Items[0]
+
+			// Get the logs from the router pod.
+			kubeConfig, err := config.GetConfig()
+			if err != nil {
+				t.Fatalf("failed to get kube config: %v", err)
+			}
+
+			client, err := kubernetes.NewForConfig(kubeConfig)
+			if err != nil {
+				t.Fatalf("failed to create kube client: %v", err)
+			}
+
+			logsContainerName := "logs"
+			foundExpectedCookie := false
+			expectedTruncatedCookieString := fmt.Sprintf("%s=%s", cookieName, padding) // Same as cookieString, minus the 'x' at the end.
+			err = wait.PollImmediate(1*time.Second, 30*time.Second, func() (bool, error) {
+				readCloser, err := client.CoreV1().Pods(routerPod.Namespace).GetLogs(routerPod.Name, &corev1.PodLogOptions{
+					Container: logsContainerName,
+					Follow:    false,
+				}).Stream(context.TODO())
+				if err != nil {
+					t.Logf("Failed to read logs from pod %s container %s: %v", routerPod.Name, logsContainerName, err)
+					return false, nil
+				}
+
+				defer func() {
+					if err := readCloser.Close(); err != nil {
+						t.Logf("Failed to close reader for pod %s container %s: %v", routerPod.Name, logsContainerName, err)
+					}
+				}()
+
+				// Search the logs container for the specified cookie, making sure that it exactly matches the expected truncated value.
+				scanner := bufio.NewScanner(readCloser)
+				for scanner.Scan() {
+					line := strings.TrimSpace(scanner.Text())
+					if strings.Contains(line, cookieString) {
+						t.Fatalf("found untruncated cookie string: %v", cookieString)
+					}
+					if strings.Contains(line, expectedTruncatedCookieString) {
+						foundExpectedCookie = true
+					}
+				}
+				if err := scanner.Err(); err != nil {
+					t.Logf("Error while reading container logs: %v", err)
+					return false, nil
+				}
+				return foundExpectedCookie, nil
+			})
+			if !foundExpectedCookie {
+				t.Fatalf("did not find expected truncated cookie string: %v", expectedTruncatedCookieString)
+			}
+		})
 	}
 }
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -352,6 +352,19 @@ func getDeployment(t *testing.T, client client.Client, name types.NamespacedName
 	return &dep, nil
 }
 
+func getPods(t *testing.T, cl client.Client, deployment *appsv1.Deployment) (*corev1.PodList, error) {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return nil, fmt.Errorf("deployment %s has invalid spec.selector: %w", deployment.Name, err)
+	}
+	podList := &corev1.PodList{}
+	if err := cl.List(context.TODO(), podList, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		t.Logf("failed to list pods for deployment %q: %v", deployment.Name, err)
+		return nil, err
+	}
+	return podList, nil
+}
+
 func podExec(t *testing.T, pod corev1.Pod, stdout, stderr *bytes.Buffer, cmd []string) error {
 	t.Helper()
 	kubeConfig, err := config.GetConfig()
--- a/test/e2e/all_test.go
+++ b/test/e2e/all_test.go
@@ -37,6 +37,7 @@ func TestAll(t *testing.T) {
 		t.Run("TestForwardedHeaderPolicyReplace", TestForwardedHeaderPolicyReplace)
 		t.Run("TestHAProxyTimeouts", TestHAProxyTimeouts)
 		t.Run("TestHAProxyTimeoutsRejection", TestHAProxyTimeoutsRejection)
+		t.Run("TestCookieLen", TestCookieLen)
 		t.Run("TestHTTPCookieCapture", TestHTTPCookieCapture)
 		t.Run("TestHTTPHeaderBufferSize", TestHTTPHeaderBufferSize)
 		t.Run("TestHTTPHeaderCapture", TestHTTPHeaderCapture)


rfredette:
With this and the related router PR loaded on a cluster, I don't see the truncation warning, but I am seeing this message popping up every few seconds in the router logs: 
```
E1221 23:24:41.426700       1 webhook.go:90] Failed to make webhook authenticator request: the server could not find the requested resource
```
If I revert the ingress operator to the version without this change, I see the truncation warning but not this webhook message, so I don't think it's a configuration issue, but I need to dig some more to figure out what this error message even really means...

pkg/operator/controller/ingress/deployment.go
```diff
@@ -820,12 +821,29 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+			// By default, captured cookies go into a datastructure that's 64
+			// bytes, with one byte reserved, so the max length of the cookie
+			// itself is 63 bytes.  If the captured length is greater than 63
+			// bytes, we can tune that data strucutre to be larger using
+			// tune.http.cookielen (exposed through the env var
+			// ROUTER_COOKIELEN). However, this is not advised since the cookie
+			// is captured for every request and every response, and the
+			// maximum length is always allocated, so the memory impact is
+			// high.
+			if maxLength > 63 {
```
candita, line 841: `maxLength` should be evaluated prior to setting `RouterCaptureHTTPCookies`, since it uses the `maxLength`.

rfredette:
some tests failed due to an issue building the image, but since other tests didn't hit that issue, I'm assuming it was a timing thing.
/retest

rfredette:
The new test, `TestCookieLen`, won't pass in CI until openshift/router#436 merges. I ran the test on a cluster with both this PR and openshift/router#436, and it passes:
```$ TEST="TestCookieLen" make test-e2e
CGO_ENABLED=1 GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run "TestCookieLen" ./test/e2e
=== RUN   TestCookieLen                                                                                                                                                       
=== PAUSE TestCookieLen                                                                
=== CONT  TestCookieLen                                                                                                                                                       
    tuning_options_test.go:516: Deleted cookielen-curl-6qf4c                   
    tuning_options_test.go:516: deleted ingresscontroller test-cookielen-2kt9d       
--- PASS: TestCookieLen (85.55s)                                                                                                                                              
PASS                                                                                   
ok      github.com/openshift/cluster-ingress-operator/test/e2e  86.000s      
```

rfredette:
After looking at this again, it's simpler to have the router figure out when to set `tune.http.cookielen` from the existing environment variables, rather than having the operator add a new environment variable. I've updated openshift/router#436 to do that, and now this PR only includes a new e2e test. Like before, the e2e test won't pass in CI until openshift/router#436 merges, but I've run it against a cluster with openshift/router#436 deployed, and it passes:
```
$ TEST="TestCookieLen" make test-e2e
CGO_ENABLED=1 GO111MODULE=on GOFLAGS=-mod=vendor go test -timeout 1h -count 1 -v -tags e2e -run "TestCookieLen" ./test/e2e
=== RUN   TestCookieLen
=== PAUSE TestCookieLen
=== CONT  TestCookieLen
    tuning_options_test.go:523: Deleted cookielen-curl-vcmdx
    tuning_options_test.go:523: deleted ingresscontroller test-cookielen-2llcg
--- PASS: TestCookieLen (84.51s)
PASS
ok      github.com/openshift/cluster-ingress-operator/test/e2e  84.808s
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,170 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			HttpLogFormat: "cookie: %CC",
+		},
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("Failed to create ingresscontroller %s: %v", icName, err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	if err := waitForIngressControllerCondition(t, kclient, 2*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+		t.Errorf("Timed out waiting for ingresscontroller %s to become available: %v", icName.Name, err)
```
frobware, line 364: If the ingress controller doesn't become ready then this should be a call to Fatal.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,170 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+		Name:      names.SimpleNameGenerator.GenerateName("cookielen-curl-"),
+		Namespace: "default",
+	}
+	curlPod := buildExecPod(curlPodName.Name, curlPodName.Namespace, curlPodImage)
+	if err := kclient.Create(context.TODO(), curlPod); err != nil {
+		t.Fatalf("failed to create pod %q: %v", curlPodName.Name, err)
+	}
+	defer assertDeleted(t, kclient, curlPod)
+
+	// Wait for the curl pod to be available
```
frobware, line 387: nit: seems like we wait for it to be Ready.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,170 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+	// Send a request through the router that includes the captured cookie, and verify that it's truncated as expected.
+
+	// Use the router image for the client pod since it includes curl.
+	routerDeployment := &appsv1.Deployment{}
+	routerDeploymentName := controller.RouterDeploymentName(ic)
+	if err := kclient.Get(context.TODO(), routerDeploymentName, routerDeployment); err != nil {
+		t.Fatalf("Failed to get router deployment for ingresscontroller %s: %v", icName.Name, err)
+	}
+
+	curlPodImage := routerDeployment.Spec.Template.Spec.Containers[0].Image
```
frobware, line 376: We should first assert on the length of Spec.Containers == 1.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,170 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+				// It shouldn't be possible to match the regexp without the submatch, but this check will cover our
+				// bases.
+				t.Fatalf("Found regexp match but not submatch")
+			}
+			cookieValue := match[1]
+			if len(cookieValue) == maxLength {
+				foundCookie = true
+				break
+			}
+			t.Errorf("Found cookie but length did not match. Expected %d, got %d", maxLength, len(cookieValue))
```
frobware, line 490: Should we continue to iterate over the remaining lines if this error hold true? Is this actually Fatal?

frobware (changes requested):
I'd like to see tests for where the maxLength for the cookie capture is 63 and 1024. The API forces a value so the first would exercise the default, and the latter tests for the API's maximum. In both cases we should assert that cookie values that are 1 more than the specified max get truncated.

test/e2e/client_tls_test.go
```diff
@@ -1412,19 +1412,6 @@ func verifyCRLs(t *testing.T, pod *corev1.Pod, expectedCRLs map[string]*x509.Rev
 	return true, nil
 }
 
-func getPods(t *testing.T, cl client.Client, deployment *appsv1.Deployment) (*corev1.PodList, error) {
```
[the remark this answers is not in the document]
rfredette: done

test/e2e/tuning_options_test.go
rfredette: done

test/e2e/tuning_options_test.go
rfredette: done

test/e2e/tuning_options_test.go
rfredette: done

rfredette:
I updated the test to add a few test cases, one at 63 (the previous limit), one at 64 (one over the limit to check the lower bound), one at 128, and one at 1024 (the max the API allows). The 1024 byte test is currently failing, but it seems like syslog is truncating the log line, because if I shorten the log format that haproxy is using, the cookie gets truncated at a larger value.

That said, I'm not sure I can change the log format enough to get the full 1024 bytes. I'm going to see if I can tune syslog to allow longer log lines, but if not it may not be a very useful test case.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,170 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			t.Logf("failed to get client pod %q: %v", curlPodName, err)
+			return false, nil
+		}
+		for _, cond := range curlPod.Status.Conditions {
+			if cond.Type == corev1.PodReady {
+				return cond.Status == corev1.ConditionTrue, nil
+			}
+		}
+		return false, nil
+	})
```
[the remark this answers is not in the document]
frobware: Looking at https://github.com/openshift/cluster-ingress-operator/pull/871/commits/6d68259cd2fb1996c89575493ffee66be773167f I didn't see an update  where we handle this error.

test/e2e/tuning_options_test.go
rfredette: Rather than assert on `Spec.Containers = 1` (since if logging is set to "container" on the default ingresscontroller, there would be 2 containers), I updated it to loop through the containers and get the image from the one named "router" and `t.Fatalf` if that fails. Does that satisfy your concern?

rfredette:
It looks like it wasn't syslog that was truncating the logs, but instead it was haproxy. When using the "Container" logging strategy, we currently leave `ROUTER_LOG_MAX_LENGTH` unset, and the template has a default log length of 1024 characters. By setting `ROUTER_LOG_MAX_LENGTH` to 2048 when using Container logging, the new test is passing on my cluster, so I've included that commit here.

I imagine, though, that the better way to solve that is to update the template in openshift/router instead, but some discussion is warranted on what to update the default to.

rfredette:
must gather failure
/retest

rfredette:
aws-operator failed on cluster setup
/retest

rfredette:
Another must-gather failure
/retest

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+						t.Fatalf("Found regexp match but not submatch")
+					}
+					cookieValue := match[1]
+					if len(cookieValue) == maxLength {
+						return
+					}
+					t.Fatalf("Found cookie but length did not match. Expected %d, got %d", maxLength, len(cookieValue))
+				}
+			}
+			t.Errorf("Cookie %q was not found in router pod %q logs", expectedCookieString, routerPod.Name)
```
frobware, line 518:
```suggestion
			// Search the logs container for the specified
			// cookie, making sure that it exactly matches
			// the expected truncated value.
			expectedTruncatedCookieString := fmt.Sprintf("%s=%s", cookieName, padding) // this is without the 'x'
			foundExpectedCookie := false
			scanner := bufio.NewScanner(readCloser)
			for scanner.Scan() {
				line := strings.TrimSpace(scanner.Text())
				if strings.Contains(line, cookieString) {
					t.Fatalf("found untruncated cookie string: %v", cookieString)
				}
				if strings.Contains(line, expectedTruncatedCookieString) {
					foundExpectedCookie = true
				}
			}
			if !foundExpectedCookie {
				t.Fatalf("did not find expected truncated cookie string: %v", expectedTruncatedCookieString)
			}
```

frobware (changes requested):

test/e2e/tuning_options_test.go
frobware: I changed the test to also assert that we don't match the cookie that has the trailing "x" in addition to asserting that there's a match for the truncated value.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+// truncated at the expected max length.
+func TestCookieLen(t *testing.T) {
+	testCases := []struct {
+		name      string
+		maxLength int
+	}{
+		{"63 bytes", 63},
+		{"64 bytes", 64},
+		{"128 bytes", 128},
+		{"1024 bytes", 1024},
```
frobware, line 341: We need an explanation for each of the magic numbers. 

```go
	testCases := []struct {
		name      string
		maxLength int
	}{
		// 63 is one less than the default maximum length of a
		// cookie (64). We test with this length to ensure our
		// system handles values just below the default
		// maximum correctly.
		{"63 bytes", 63},

		// 64 is the default maximum length for a cookie
		// value. We test with this length to ensure our
		// system behaves correctly when the default maximum
		// length is used.
		{"64 bytes", 64},

		// 128 is significantly more than the default maximum
		// length. We test with this length to see how our
		// system behaves when a value larger than the default
		// maximum is used.
		{"128 bytes", 128},

		// 1024 is the maximum length of a cookie as defined
		// by the OpenShift Ingress API. We test with this
		// length to ensure our system can handle the largest
		// cookie value that OpenShift allows.
		{"1024 bytes", 1024},
	}
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			maxLength := testCase.maxLength
+			t.Parallel()
```
frobware, line 346: Out of consistency please can we move this to the first line in the subtest (i.e., before `maxLength`).

test/e2e/tuning_options_test.go
frobware: If it's easier to consume then here's my suggestion with the addition that we check the scanner for errors too:

```go
			// Search the logs container for the specified
			// cookie, making sure that it exactly matches
			// the expected truncated value.
			expectedTruncatedCookieString := fmt.Sprintf("%s=%s", cookieName, padding) // this is without the 'x'
			foundExpectedCookie := false
			scanner := bufio.NewScanner(readCloser)
			for scanner.Scan() {
				line := strings.TrimSpace(scanner.Text())
				if strings.Contains(line, cookieString) {
					t.Fatalf("found untruncated cookie string: %v", cookieString)
				}
				if strings.Contains(line, expectedTruncatedCookieString) {
					foundExpectedCookie = true
				}
			}
			if err := scanner.Err(); err != nil {
				t.Fatalf("Error while reading container logs: %v", err)
			}
			if !foundExpectedCookie {
				t.Fatalf("did not find expected truncated cookie string: %v", expectedTruncatedCookieString)
			}
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+				}
+			}
+			if curlPodImage == "" {
+				t.Fatalf("Failed to find container \"router\" in deployment %s", routerDeployment.Name)
+			}
+			curlPodName := types.NamespacedName{
+				Name:      names.SimpleNameGenerator.GenerateName("cookielen-curl-"),
+				Namespace: "default",
+			}
+			curlPod := buildExecPod(curlPodName.Name, curlPodName.Namespace, curlPodImage)
```
candita, line 404: Why not use `buildCurlPod`?

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			maxLength := testCase.maxLength
+			t.Parallel()
+			icName := types.NamespacedName{
+				Namespace: operatorNamespace,
+				Name:      names.SimpleNameGenerator.GenerateName("test-cookielen-"),
+			}
+			domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+			ic := newPrivateController(icName, domain)
```
candita, line 352: Is there a reason you need to create a new ic for each test case?

rfredette: The cases run in parallel, so unless I'm mistaken, they each need their own ingress controller.

test/e2e/tuning_options_test.go
rfredette: Do you care if it's in the test case name, or will a comment suffice?

test/e2e/tuning_options_test.go
rfredette: Good question. `buildCurlPod` does look like it would simplify things; I'll rework it to use that.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			if err := podExec(t, *curlPod, &stdout, &stderr, cmd); err != nil {
+				t.Fatalf("Failed running command %q: %v\nstdout:\n%s\nstderr:\n%s", cmd, err, stdout.String(), stderr.String())
+			}
+
+			// Find the router pod for our ingress controller so we can check its logs. This ingress controller specifies
+			// replicas=1, so there will only be one.
+			routerPodList, err := getPods(t, kclient, routerDeployment)
+			if err != nil {
+				t.Fatalf("failed to get pods in deployment %s: %v", routerDeploymentName, err)
+			} else if len(routerPodList.Items) == 0 {
```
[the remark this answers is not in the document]
rfredette: done

test/e2e/tuning_options_test.go
rfredette: Thanks, this does make it much more straightforward. I've incorporated it in my latest commit.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +327,195 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+
+			defer func() {
+				if err := readCloser.Close(); err != nil {
+					t.Logf("Failed to close reader for pod %s container %s: %v", routerPod.Name, logsContainerName, err)
+				}
+			}()
+
+			// Search the logs container for the specified cookie, making sure that it exactly matches the expected truncated value.
+			expectedCookieString := fmt.Sprintf("%s=%s", cookieName, padding)
+			cookieNameRegexp := regexp.MustCompile(fmt.Sprintf("cookie: (%s=a+x?)$", cookieName))
```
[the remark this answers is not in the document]
rfredette: This should be addressed with the rework in the latest commit.

test/e2e/tuning_options_test.go
candita: I was thinking - create one ic, and create one pod per test case.  That shouldn't cause any conflicts that I can think of.  I'm a little concerned that there could be resource failures or timeouts creating so many controllers.

rfredette: I don't think one ic would work; Each test case sets the ic's `spec.logging.access.httpCaptureCookies.maxLength` field to a different value, and you can only set one cookie to capture per ic.

rfredette:
After testing the latest changes, I ran into a few issues, one with the conversion to use `buildCurlPod`, and another where the canary's `spec.host` field was empty, which caused the test to try to curl `""`. Both have been fixed in the commits I just pushed.

test/e2e/tuning_options_test.go
candita: Normally we run each test in parallel, but here you are running each test case in parallel.  Did you get advice to do it this way?  I don't see the advantages, but do see the aforementioned potential disadvantage in that resource use could be compromised by running 4 private ics in parallel in ci.

Could you at least add a comment and note that it might cause an issue and point out how to fix that issue (don't run in parallel and use one private ic).

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			defer assertIngressControllerDeleted(t, kclient, ic)
+
+			if err := waitForIngressControllerCondition(t, kclient, 2*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+				t.Fatalf("Timed out waiting for ingresscontroller %s to become available: %v", icName.Name, err)
+			}
+
+			// Send a request through the router that includes the captured cookie, and verify that it's truncated as expected.
+
+			// Use the router image for the client pod since it includes curl.
+			routerDeployment := &appsv1.Deployment{}
```
candita, line 389: Could you take a quick look at how `buildCurlPod` is done in https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L3093 `TestUniqueIDHeaders`?  I think you should build your own simple image, route, and service rather than use the ones provided by router, canary and ingress.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+// truncated at the expected max length.
+func TestCookieLen(t *testing.T) {
+	testCases := []struct {
+		name      string
+		maxLength int
+	}{
+		// 63 bytes was the limit before OCPBUGS-3356 was fixed; This should pass even without the patch, so it's more
+		// of a sanity check.
+		{"63 bytes", 63},
+		// The lower bound. 64 bytes is the smallest number that would be truncated short prior to the OCPBUGS-3356 fix.
```
candita, line 340: How can both 63 and 64 be the lower limit?

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			padding := strings.Repeat("a", maxLength-len(cookieName)-1)
+			cookieString := fmt.Sprintf("%s=%sx", cookieName, padding)
+
+			curlArgs := []string{
+				"-k", "-L",
+				"-b", cookieString,
+				// --resolve used this way makes curl send the request to our ingress controller's internal service address as
+				// if it was directed to it by a load balancer. This ensures the request is handled by our ingress controller
+				// without having to wait extra time for a load balancer to be provisioned.
+				"--resolve", fmt.Sprintf("%s:443:%s", canaryRouteURL, icInternalService.Spec.ClusterIP),
```
candita, line 447: I don't think you need to do this at all. Can you explain?  It's better to use the established pattern mentioned in https://github.com/openshift/cluster-ingress-operator/pull/871/files#r1296550078, unless there is a need to do it differently.

test/e2e/util_test.go
```diff
@@ -137,7 +137,11 @@ func buildCurlPod(name, namespace, image, host, address string, extraArgs ...str
 		"--retry", "300", "--retry-delay", "1", "--max-time", "2",
 	}
 	curlArgs = append(curlArgs, extraArgs...)
-	curlArgs = append(curlArgs, "http://"+host)
+	if strings.HasPrefix(host, "http://") || strings.HasPrefix(host, "https://") {
```
candita, line 140: I think the only reason you need to do this is because you're borrowing the canary route.  Please don't borrow the canary route, as mentioned previously.

test/e2e/tuning_options_test.go
frobware: We create many ingresscontoller's in the e2e tests and creating 4 more shouldn't be a problem (until some limit proves me wrong). 

If we created 4 new discrete TestFeature functions, each creating their own private IC, to accommodate and test new features over subsequent OCP releases would you have the same concerns? 

w.r.t running each test case in parallel--that seems fine to me. It's really no different to creating 4 new TestX functions that use `t.Parallel`.

test/e2e/tuning_options_test.go
rfredette: The advantage to using the canary route is that it's guaranteed to already be deployed, which speeds up the test by not requiring additional pods and routes to be created. That said, it does make the test more confusing for a pretty small gain. I'll rework this to use an echo pod like `TestUniqueIDHeaders`.

test/e2e/tuning_options_test.go
rfredette: 63 bytes was the limit before the fix in openshift/router#436. You could remove that case and it would cover the range where behavior has changed due to the router PR, but I think it's helpful to include the 63 byte case as a sanity check. 
If it passes but all the other cases fail, there's something wrong with the way the global variable `tune.http.cookielen` was set, but if it and the other cases fail, it indicates that the failure is probably unrelated to `tune.http.cookielen`.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
 }
+
+// TestCookieLen verifies that when logging is enabled and a cookie capture is specified, any captured cookies are
+// truncated at the expected max length.
+func TestCookieLen(t *testing.T) {
+	testCases := []struct {
+		name      string
+		maxLength int
+	}{
+		// 63 bytes was the limit before OCPBUGS-3356 was fixed; This should pass even without the patch, so it's more
```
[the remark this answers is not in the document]
rfredette: "The patch" here means the fix for OCPBUGS-3356, but I'll reword it for better clarity.

test/e2e/tuning_options_test.go
rfredette: `TestUniqueIdHeader` does the same thing here: https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L3163 and there are a few other test cases that use this same behavior. It bypasses DNS and explicitly sends the request to the router pod(s) for a particular ingress controller, guaranteeing that the right router pod handles the connection, and removing any chance of slow DNS propagation causing test delays or failures.

test/e2e/tuning_options_test.go
candita: I'm going to suggest some re-wording of the comments now that I've reviewed https://github.com/openshift/router/pull/436

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+				if strings.Contains(line, cookieString) {
+					t.Fatalf("found untruncated cookie string: %v", cookieString)
+				}
+				if strings.Contains(line, expectedTruncatedCookieString) {
+					foundExpectedCookie = true
+				}
+			}
+			if err := scanner.Err(); err != nil {
+				t.Fatalf("Error while reading container logs: %v", err)
+			}
```
candita, line 520: If you found the right cookie, you probably wouldn't want this to be a Fatal error.  
```suggestion
			if err := scanner.Err(); err != nil {
				t.Errorf("Error while reading container logs: %v", err)
			}
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			if !foundExpectedCookie {
+				t.Fatalf("did not find expected truncated cookie string: %v", expectedTruncatedCookieString)
+			}
```
candita, line 523: Add a message here so that if there are some failures and some successes, you will know which succeeded.
```suggestion
			}
			t.Logf("found expected cookie truncated to %d", maxLength)
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+
+// TestCookieLen verifies that when logging is enabled and a cookie capture is specified, any captured cookies are
+// truncated at the expected max length.
+func TestCookieLen(t *testing.T) {
+	testCases := []struct {
+		name      string
+		maxLength int
+	}{
+		// 63 bytes was the limit before OCPBUGS-3356 was fixed; This should pass even without the patch, so it's more
+		// of a sanity check.
```
candita, line 338:
```suggestion
		// 63 bytes is the HAProxy default, which would be the case when no maxLength is specified
```

test/e2e/tuning_options_test.go
candita, line 340:
```suggestion
		// The lower bound. 64 bytes is the smallest number above the HAProxy default, and for which a HAProxy config change is made.
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+		{"63 bytes", 63},
+		// The lower bound. 64 bytes is the smallest number that would be truncated short prior to the OCPBUGS-3356 fix.
+		{"64 bytes", 64},
+		// 128 bytes is just a number between the upper and lower bounds. If this passes and either boundary cases fail,
+		// it's likely an off-by-one or similar boundary issue.
```
candita, line 343:
```suggestion
		// 128 bytes is an additional test to prove that the truncation length is correct.
```

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +326,201 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+		{"128 bytes", 128},
+		// The upper bound. 1024 is the maximum allowed by the API.
```
candita, line 345: What would happen for 1026?  Would it default to 1024 or 63, or break the test?

test/e2e/tuning_options_test.go
rfredette: Each test case is run with `t.Run()`, so they show up separately in the test output. To induce a failure and to test out a question you asked in another comment, I changed the 1024 byte case to 1026 bytes and ran the test. Here's a snippet of the output:
```
=== NAME  TestCookieLen/1026_bytes
    tuning_options_test.go:379: Failed to create ingresscontroller openshift-ingress-operator/test-cookielen-hv8vz: IngressController.operator.openshift.io "test-cookielen-hv8vz" is invalid: spec.logging.access.httpCaptureCookies[0].maxLength: Invalid value: 1026: spec.logging.access.httpCaptureCookies[0].maxLength in body should be less than or equal to 1024

... cut for brevity ...

=== NAME  TestCookieLen/128_bytes
    tuning_options_test.go:541: deleted ingresscontroller test-cookielen-j5lxj
=== NAME  TestCookieLen/63_bytes
    tuning_options_test.go:541: deleted ingresscontroller test-cookielen-kh5rl
--- FAIL: TestCookieLen (0.00s)
    --- FAIL: TestCookieLen/1026_bytes (0.07s)
    --- PASS: TestCookieLen/64_bytes (90.57s)
    --- PASS: TestCookieLen/128_bytes (91.78s)
    --- PASS: TestCookieLen/63_bytes (93.00s)
FAIL
FAIL    github.com/openshift/cluster-ingress-operator/test/e2e  93.475s
FAIL
```
The output does get interleaved due to the parallel execution, but I think it's clear enough which test case is which.

test/e2e/tuning_options_test.go
rfredette: Thanks for your suggestions. I did adjust the wording slightly, but I removed any reference to the specific jira issue. Let me know if you think they are in need of any additional edits.

test/e2e/util_test.go
rfredette: I think it would be good for `buildCurlPod` to allow https, but with the switch to an echo pod, it doesn't need to be done here, so I've removed this change.

rfredette:
Test failures are unrelated.
/retest

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +325,219 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+			}).Stream(context.TODO())
+			if err != nil {
+				t.Fatalf("Failed to read logs from pod %s container %s: %v", routerPod.Name, logsContainerName, err)
+			}
+
+			defer func() {
+				if err := readCloser.Close(); err != nil {
+					t.Logf("Failed to close reader for pod %s container %s: %v", routerPod.Name, logsContainerName, err)
+				}
+			}()
```
candita, line 520: I think this defer function needs to go after L525, where the readCloser is used.

test/e2e/tuning_options_test.go
```diff
@@ -320,3 +325,219 @@ func TestHAProxyTimeoutsRejection(t *testing.T) {
...
+				t.Fatalf("failed to get kube config: %v", err)
+			}
+
+			client, err := kubernetes.NewForConfig(kubeConfig)
+			if err != nil {
+				t.Fatalf("failed to create kube client: %v", err)
+			}
+
+			logsContainerName := "logs"
+			readCloser, err := client.CoreV1().Pods(routerPod.Namespace).GetLogs(routerPod.Name, &corev1.PodLogOptions{
```
candita, line 508: The pattern that is used for scanning logs usually has a poll loop, like https://github.com/openshift/cluster-ingress-operator/blame/master/test/e2e/http_header_buffer_test.go#L134-L163.  If it isn't there the first time you check, it may not have arrived yet.

test/e2e/tuning_options_test.go
rfredette: I don't think it matters, as long as the defer function is registered between `readCloser` being instantiated and the first possible return condition, since it won't get called until `TestCookieLen` returns.

test/e2e/tuning_options_test.go
candita: I was asking about adding a positive message to match the negative message: "did not find expected truncated cookie string". It's a nit.

But the error message I see above for 1026 bytes, that seems weird.  It starts with "Failed to create ingresscontroller".  Can you add a test that you expect to fail and print out the reason why it fails rather than the reason why the ingress controller doesn't get created?

rfredette: The error message in the 1026 case was the expected one. `httpCaptureCookies.maxLength` is a field of the ingresscontroller object, and the openshift API limits it to a number between 1 and 1024, so trying to set it to a number outside that range will cause the ingresscontroller to be rejected

PR #873, opened 2023-01-17
NE-1124: Add support for External platform to CIO

frobware:
https://issues.redhat.com/browse/NE-1124.

For cases where we previously handled the `NonePlatformType` also handle `ExternalPlatformType` in an identical way. If there was an existing unit test for `NonePlatformType` I carried the same test substituting None for External as appropriate.

This change also pulls in https://github.com/openshift/api/pull/1301 where the `ExternalPlatformType` was introduced.


frobware:
cc @elmiko

elmiko (commented):
i'm not overly familiar with the code base here, but the changes look like what i would expect. thanks @frobware !

/lgtm

frobware:
@elmiko I'm guessing "no" but are there any requirements for docs, release note, TE, et al?

elmiko:
we had not planned any product docs for this change, but we did have a mention that each team would evaluate adding developer docs to their repository. looking at the docs in this repository it doesn't seem like there is a discussion of platform, so i think this is good to go.

Miciah:
Thanks!
/approve

frobware:
openshift cluster install failed with cluster creation

/retest

PR #878, opened 2023-01-26
OCPBUGS-6700: updateIngressClass: Fix log message

gcs278:
updateIngressClass: Fix log message

* pkg/operator/controller/ingressclass/ingressclass.go
(updateIngressClass): Log when the update succeeds, not when it fails.


Miciah:
Thanks!
/approve
/lgtm
/retest-required

PR #879, opened 2023-01-26
OCPBUGS-6701: Avoid spurious updates for scope in IngressClass

gcs278:
Specify spec.parameters.scope on ingressclasses that the operator creates.

Before this commit, the update logic would try to revert the default value that the API set for an ingressclass's scope field.

* pkg/operator/controller/ingressclass/ingressclass.go (desiredIngressClass): Specify spec.parameters.scope with the default value: "Cluster".
* pkg/operator/controller/ingressclass/ingressclass_test.go (Test_desiredIngressClass): Update test to expect spec.parameters.scope to be set.


Miciah:
Thanks!
/approve
/lgtm
/retest-required

PR #883, opened 2023-01-27
NE-1186: Test_getRR: Fix typo: "excepted" → "expected"

gcs278:
Test_getRR: Fix typo: "excepted" → "expected"

* pkg/dns/alibaba/dns_test.go (Test_getRR): Replace "excepted" with "expected" in test case struct definition.


Miciah:
This is just a typo fix in a unit test; no docs, PX, or QE required.  
/label docs-approved
/label px-approved
/label qe-approved

PR #890, opened 2023-02-15
NE-1105: Add support for Gateway API

Miciah:
#### Add gatewayapi controller

Introduce a new "gatewayapi" controller, which creates the Gateway API CRDs when the "GatewayAPI" feature gate is enabled.

* `assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml`:
* `assets/gateway-api/gateway.networking.k8s.io_gateways.yaml`:
* `assets/gateway-api/gateway.networking.k8s.io_httproutes.yaml`:
* `assets/gateway-api/gateway.networking.k8s.io_referencegrants.yaml`: New files.
* `manifests/00-cluster-role.yaml`: Allow the operator to manage the Gateway API CRDs.
* `pkg/manifests/manifests.go` (`GatewayClassCRDAsset`, `GatewayCRDAsset`, `HTTPRouteCRDAsset`, `ReferenceGrantCRDAsset`): New consts.
(`GatewayClassCRD`, `GatewayCRD`, `HTTPRouteCRD`, `ReferenceGrantCRD`, `NewCustomResourceDefinition`): New functions.
* `pkg/manifests/manifests_test.go` (`TestManifests`): Test `GatewayClassCRD`, `GatewayCRD`, `HTTPRouteCRD`, and `ReferenceGrantCRD`.
* `pkg/operator/controller/gatewayapi/controller.go`: New file.
(`controllerName`, `featureGateName`): New consts.
(`log`): New variable.  Define a logger for the controller.
(`New`): New function.  Create an instance of the controller.
(`reconciler`): New type.  Store the state of a gatewayapi controller.
(`Reconcile`): New method.  Handle reconciliation requests for featuregates by checking for the "GatewayAPI" feature and ensuring the Gateway API CRDs exist if the feature is enabled.
(`featureIsEnabled`): New function.  Helper for `Reconcile`.
* `pkg/operator/controller/gatewayapi/controller_test.go`: New file.
(`Test`_Reconcile): New test.  Verify that `Reconcile` behaves as expected.
(`fakeCache`, `fakeClientRecorder`): New types, used in `Test_Reconcile`.
(`Get`, `List`, `Scheme`, `RESTMapper`, `Create`, `Delete`, `DeleteAllOf`, `Update`, `Patch`, `Status`): New methods for `fakeClientRecorder` to implement the controller-runtime `client.Client` interface.
* `pkg/operator/controller/gatewayapi/crds.go`: New file.
(`managedCRDs`): New variable.  Define the CRDs that this controller manages.
(`ensureCRD`): New method.  Ensure the specified CRD is present, using `currentCRD`, `createCRD`, and `updateCRD`.
(`ensureGatewayAPICRDs`): Ensure the CRDs in `managedCRDs` are present, using `ensureCRD`.
(`currentCRD`, `createCRD`, `updateCRD`): New methods.
(`crdChanged`): New function, used by `updateCRD`.
* `pkg/operator/operator.go` (`New`): Initialize the new gatewayapi controller.
* `go.mod`:
* `vendor/*`: Regenerate.

#### Add gatewayclass controller

Introduce a new "gatewayclass" controller, which watches gatewayclasses and installs the Service Mesh operator when a GatewayClass owned by OpenShift is first observed.

A GatewayClass is considered owned by OpenShift if its `spec.controllerName` field specifies "openshift.io/gateway-controller".

* `manifests/00-cluster-role.yaml`: Allow the operator to manage the Gateway API CRs, the servicemeshoperator subscription, and servicemeshcontrolplanes.
* `go.mod`: Add `github.com/maistra/istio-operator`, `github.com/operator-framework/api`, and `sigs.k8s.io/gateway-api`.
* `go.sum`:
* `pkg/manifests/bindata.go`:
* `vendor/*`: Regenerate.
* `pkg/operator/client/client.go` (`init`): Add `apiextensionsv1`, `gatewayapiv1beta1`, `operatorsv1alpha1`, `maistrav1`, and `maistrav2` to the client scheme.
* `pkg/operator/controller/gatewayapi/controller.go` (`New`): Add a config parameter.
(`Config`): New type.  Define configuration for the gatewayapi controller. For now, the only configuration is a list of dependent controllers, which for now includes only the gatewayclass controller.
(`reconciler`): Add a config field with the controller configuration.  Add a `startControllers` `sync.Once` field to ensure the gatewayapi controller only starts all dependent controllers once.
(`Reconcile`): Start dependent controllers (i.e. the gatewayclass controller) after installing the Gateway API CRDs.
* `pkg/operator/controller/gatewayapi/controller_test.go` (`Test_Reconcile`): Configure the reconciler with a fake gatewayclass controller.
(`TestReconcileOnlyStartsControllerOnce`): New test.  Verify that `Reconcile` only starts the gatewayclass controller once even if the gatewayapi controller gets multiple reconcile requests for the featuregate.
(`fakeController`): New type, used in `TestReconcileOnlyStartsControllerOnce`.
(`Reconcile`, `Watch`, `Start`, `GetLogger`): New methods for `fakeController` to implement the controller-runtime `controller.Controller` interface.
* `pkg/operator/controller/gatewayclass/controller.go`: New file.
(`controllerName`, `OpenShiftGatewayClassControllerName`, `OpenShiftDefaultGatewayClassName`): New consts.
(`NewUnmanaged`): New function.  Create a new gatewayclass controller that is not started automatically by the manager.
(`Config`): New type.  Store the configuration for a gatewayclass controller.
(`reconciler`): New type.  Store the state of a gatewayclass controller.
(`Reconcile`): New method.  Reconcile a gatewayclass by deploying OpenShift Service Mesh, using `ensureServiceMeshOperatorSubscription` and `ensureServiceMeshControlPlane`.
* `pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go` (`ensureServiceMeshControlPlane`): New method.  Ensure a servicemeshcontrolplanes is present, using `currentServiceMeshControlPlane`, `desiredServiceMeshControlPlane`, `createServiceMeshControlPlane`, and `updateServiceMeshControlPlane`.
(`desiredServiceMeshControlPlane`): New function.
(`currentServiceMeshControlPlane`, `createServiceMeshControlPlane`, `updateServiceMeshControlPlane`): New methods.
(`serviceMeshControlPlaneChanged`): New function, used by `updateServiceMeshControlPlane`.
* `pkg/operator/controller/gatewayclass/subscription.go` (`ensureServiceMeshOperatorSubscription`): New method.  Ensure a subscription for servicemeshoperator exists, using `desiredSubscription`, `currentSubscription`, `createSubscription`, and `updateSubscription`.
(`desiredSubscription`): New function.
(`currentSubscription`, `createSubscription`, `updateSubscription`): New methods.
(`subscriptionChanged`): New function, used by `updateSubscription`.
* `pkg/operator/controller/names.go` (`ServiceMeshControlPlaneName`): New function.  Return the namespaced name for a ServiceMeshControlPlane CR that is associated with a GatewayClass CR.
(`ServiceMeshSubscriptionName`): New function.  Return the namespaced name for a Subscription CR for installing OSSM.
* `pkg/operator/operator.go` (`New`): Create the gatewayclass controller and pass it to the constructor for the gatewayapi controller so that the latter controller can start the former one when Gateway API is enabled.

#### `ensureWildcardDNSRecord`: Delete `ic` parameter

Replace the `operatorv1.IngressController` parameter of `ensureWildcardDNSRecord` with parameters for the DNSRecord name, labels, owner reference, domain, and endpoint publishing strategy, and similarly for `currentWildcardDNSRecord`, `desiredWildcardDNSRecord`, and `deleteWildcardDNSRecord`.

* `pkg/operator/controller/ingress/controller.go` (`ensureIngressDeleted`): Update call to `deleteWildcardDNSRecord` with the appropriate arguments.
(`ensureIngressController`): Update to call `ensureWildcardDNSRecord` with the appropriate arguments.
* `pkg/operator/controller/ingress/dns.go` (`ensureWildcardDNSRecord`): Replace the `ic` parameter with `name`, `dnsRecordLabels`, `ownerRef`, `domain`, and `endpointPublishingStrategy` parameters.  Update calls to `desiredWildcardDNSRecord` and `currentWildcardDNSRecord` with the appropriate arguments.
(`desiredWildcardDNSRecord`): Replace the `ic` parameter with `name`, `dnsRecordLabels`, `ownerRef`, `dnsRecordDomain`, and `endpointPublishingStrategy` parameters.
(`currentWildcardDNSRecord`, `deleteWildcardDNSRecord`): Replace the `ic` parameter with a `name` parameter.
* `pkg/operator/controller/ingress/dns_test.go` (`TestDesiredWildcardDNSRecord`): Update call to `desiredWildcardDNSRecord`.

#### `ensureWildcardDNSRecord`: Delete receiver

Replace the `reconciler` receiver of `ensureWildcardDNSRecord` and `deleteWildcardDNSRecord` with a `client` parameter.

* `pkg/operator/controller/ingress/controller.go` (`ensureIngressDeleted`): Update call to `deleteWildcardDNSRecord`.
(`ensureIngressController`): Update to call `ensureWildcardDNSRecord`.
* `pkg/operator/controller/ingress/dns.go` (`ensureWildcardDNSRecord`, `currentWildcardDNSRecord`, `deleteWildcardDNSRecord`, `updateDNSRecord`): Replace receiver with `client` parameter.

#### Rename `toYaml` to `ToYaml` and move it to `pkg/util`

* `pkg/operator/controller/ingress/dns_test.go` (`toYaml`): Move from here...
* `pkg/util/yaml.go` (`ToYaml`): ...to here.  New file.
* `pkg/operator/controller/ingress/controller_test.go`:
* `pkg/operator/controller/ingress/status_test.go`: Import and call `ToYaml` using its new location.

#### dnsrecord: New package for managing DNSRecords

* `pkg/operator/controller/ingress/controller.go`: Import the new dnsrecord package and update calls to `manageDNSForDomain`, `deleteWildcardDNSRecord`, `currentWildcardDNSRecord`, and `ensureWildcardDNSRecord` to use `ManageDNSForDomain`, `DeleteWildcardDNSRecord`, `CurrentWildcardDNSRecord`, and `EnsureWildcardDNSRecord`, respectively, from the new location.
* `pkg/operator/controller/ingress/dns.go`: Move file...
* `pkg/resources/dnsrecord/dns.go`: ...here.
(`log`): New variable.  Define a logger for the new package.
(`ensureWildcardDNSRecord`): Rename function...
(`EnsureWildcardDNSRecord`): ...to this.
(`currentWildcardDNSRecord`): Rename function...
(`CurrentWildcardDNSRecord`): ...to this.
(`deleteWildcardDNSRecord`): Rename function...
(`DeleteWildcardDNSRecord`): ...to this
(`manageDNSForDomain`): Rename function...
(`ManageDNSForDomain`): ...to this.
* `pkg/operator/controller/ingress/dns_test.go`: Move file...
* `pkg/resources/dnsrecord/dns_test.go`: ...here.
(`TestManageDNSForDomain`): Update to call `ManageDNSForDomain` using its new name.

#### dns: Don't require an owning ingresscontroller

Reconcile even a dnsrecord that has no associated ingresscontroller.

* `pkg/operator/controller/dns/controller.go` (`Reconcile`): Delete check for owning ingresscontroller.

#### dns: Reconcile dnsrecords in the operand namespace

Reconcile not only dnsrecords in the operator namespace but also dnsrecords in the operand namespace.

* `pkg/operator/controller/dns/controller.go` (`Config`): Rename the `Namespace` field to `CredentialsRequestNamespace`.  Add a `DNSRecordNamespaces` field.
(`ToDNSRecords`): Check for dnsrecords in all namespaces listed in `r.config.DNSRecordNamespaces`.
* `pkg/operator/operator.go` (`New`): Specify `CredentialsRequestNamespace` with the operator namespace and `DNSRecordNamespaces` with the operator and operand namespaces.

#### Add service-dns controller

Introduce a new "service-api" controller, which watches services, dnsrecords, and gateways and creates DNSRecord CRs for services that are associated with gateways that OpenShift manages.

* `pkg/operator/controller/names.go` (`GatewayDNSRecordName`): New function.  Return the namespaced name for a DNSRecord CR associated with a Gateway.
* `pkg/operator/controller/service-dns/controller.go`: New file.
(`controllerName`, `gatewayNameLabelKey`, `managedByIstioLabelKey`): New consts.
(`log`): New variable.  Define a logger for the service-dns controller.
(`NewUnmanaged`): New function.  Create a new service-dns controller that is not started automatically by the manager.
(`gatewayListenersHostnamesChanged`): New function, used in `NewUnmanaged`.
(`Config`): New type.  Store configuration for the controller.
(`reconciler`): New type.  Store the state of a service-dns controller.
(`Reconcile`): New method.  Handle reconciliation requests for services by checking whether they are associated with OpenShift-managed gateways and creating or deleting DNSRecord CRs as appropriate, using the new `getGatewayHostnames` function and `ensureDNSRecordsForGateway` and `deleteStaleDNSRecordsForGateway` methods.
(`getGatewayHostnames`): New function.  Return the domains for a gateway listener.
(`ensureDNSRecordsForGateway`): New method.  Ensure DNSRecord CRs exist for the given domains, associated with the given gateway and service.
(`deleteStaleDNSRecordsForGateway`): New method.  Delete any DNSRecord CRs that are associated with the given gateway but have DNS names that are not in the given set of domains.
* `pkg/operator/controller/service-dns/controller_test.go`: New file.
(`Test_Reconcile`): New test.  Verify that `Reconcile` behaves as expected.
(`fakeCache`, `fakeClientRecorder`): New types, used in `Test_Reconcile`.
(`Get`, `List`, `Scheme`, `RESTMapper`, `Create`, `Delete`, `DeleteAllOf`, `Update`, `Patch`, `Status`): New methods for `fakeClientRecorder` to implement the controller-runtime `client.Client` interface.
(`Test_gatewayListenersHostnamesChanged`): New test.  Verify that `gatewayListenersHostnamesChanged` behaves as expected.
* `pkg/operator/operator.go` (`New`): Instantiate the new service-dns controller and pass it to the constructor for the gatewayapi controller so that the latter controller can start the former when Gateway API is enabled.
* `pkg/resources/dnsrecord/dns.go` (`EnsureWildcardDNSRecord`): Update calls to `CurrentWildcardDNSRecord` to use `CurrentDNSRecord` instead.
(`EnsureDNSRecord`): New function.  Create a DNS record with the specified domain without prepending a wildcard prefix, using the new `desiredDNSRecord` function.
(`desiredWildcardDNSRecord`): Factor general logic out of this function...
(`desiredDNSRecord`): ...into this.
(`CurrentWildcardDNSRecord`): Rename from this...
(`CurrentDNSRecord`): ...to this.
(`DeleteWildcardDNSRecord`): Rename from this...
(`DeleteDNSRecordand`): ...to this.
* `pkg/operator/controller/ingress/controller.go` (`ensureIngressDeleted`): Update calls to `DeleteWildcardDNSRecord` and `CurrentWildcardDNSRecord` to use `DeleteDNSRecordand` and `CurrentDNSRecord`, respectively, instead.
* `vendor/*`: Regenerate.

#### dns/aws: Store target hosted zone id in annotation

When upserting a DNS record in Route 53, store the target hosted zone id of the ELB on the associated DNSRecord CR.  If a subsequent lookup of this id fails for whatever reason, fall back to using the annotation value.

This change is necessary for Gateway API because Istio sometimes deletes the ELB before the ingress operator deletes the DNS record, and so the ELB no longer exists when the DNS record deletion logic needs to look up the target hosted zone id of the ELB.

* `pkg/dns/aws/dns.go` (`targetHostedZoneIdAnnotationKey`): New const.
(`Config`): Add `Client` field.
(`change`): Use `config.Client` and `targetHostedZoneIdAnnotationKey` to annotate the DNSRecord CR with the target hosted zone id of the ELB on upsert and look up the id on deletion.
* `pkg/operator/controller/dns/controller.go` (`Reconcile`): Get the DNSRecord CR before updating its status in case it has changed since the previous get.
(`createDNSProvider`): Provide the client to the AWS DNS provider implementation.


pkg/operator/controller/dns/controller.go
```diff
@@ -160,14 +161,6 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, nil
 	}
 
-	// Only process records associated with ingresscontrollers, because this isn't
-	// intended to be a general purpose DNS management system.
-	ingressName, ok := record.Labels[manifests.OwningIngressControllerLabel]
```
candita, line 165: I wonder if instead of removing this entirely, we could add a label for gateway DNS records that we need.  Then, we are only managing 2 known kinds of records instead of becoming a general purpose DNS management system.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -0,0 +1,136 @@
...
+	startControllers sync.Once
+}
+
+// Reconcile expects request to refer to a FeatureGate and creates or
+// reconciles the Gateway API CRDs.
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
+
+	var fg configv1.FeatureGate
+	if err := r.cache.Get(ctx, request.NamespacedName, &fg); err != nil {
```
candita, line 71: Will we allow the feature gate to be enabled and disabled?  If not, we should be able to check only once, store the value, and refer to that variable from lots of places.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -0,0 +1,136 @@
...
+	recorder         record.EventRecorder
+	startControllers sync.Once
+}
+
+// Reconcile expects request to refer to a FeatureGate and creates or
+// reconciles the Gateway API CRDs.
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
+
+	var fg configv1.FeatureGate
```
candita, line 70: The feature gate should be named "cluster", you could look it up by name.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		for _, f := range fg.Spec.FeatureGateSelection.CustomNoUpgrade.Enabled {
+			if f == feature {
+				return true
+			}
+		}
+		return false
+	}
+
+	if fs, ok := configv1.FeatureSets[fg.Spec.FeatureSet]; ok {
```
candita, line 128: This is just extra, right?

pkg/operator/controller/gatewayapi/controller_test.go
```diff
@@ -0,0 +1,342 @@
...
+				&configv1.FeatureGate{
+					ObjectMeta: metav1.ObjectMeta{Name: "cluster"},
+					Spec: configv1.FeatureGateSpec{
+						FeatureGateSelection: configv1.FeatureGateSelection{
+							FeatureSet: configv1.TechPreviewNoUpgrade,
+						},
+					},
+				},
+			},
+			// TODO Update expectCreate and expectStartCtrl when
```
candita, line 132: Not any more.

pkg/operator/controller/gatewayapi/controller_test.go
```diff
@@ -0,0 +1,342 @@
...
+				t.Fatalf("found diff between expected and actual updates: %s", diff)
+			}
+			if diff := cmp.Diff(tc.expectDelete, cl.deleted, cmpOpts...); diff != "" {
+				t.Fatalf("found diff between expected and actual deletes: %s", diff)
+			}
+		})
+	}
+}
+
+func TestReconcileOnlyStartsControllerOnce(t *testing.T) {
```
candita, line 199:
```suggestion
func Test_ReconcileOnlyStartsControllerOnce(t *testing.T) {
```

pkg/operator/controller/gatewayapi/crds.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		return r.currentCRD(ctx, name)
+	case have:
+		if updated, err := r.updateCRD(ctx, current, desired); err != nil {
+			return have, current, err
+		} else if updated {
+			return r.currentCRD(ctx, name)
+		}
+	}
+	return have, current, nil
```
candita, line 52: Is this dead code?

pkg/operator/controller/gatewayapi/crds.go
```diff
@@ -0,0 +1,128 @@
...
+		_, _, err := r.ensureCRD(ctx, crd)
+		errs = append(errs, err)
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+// currentCRD returns a Boolean indicating whether an CRD
+// exists for the IngressController with the given name, as well as the
+// CRD if it does exist and an error value.
+func (r *reconciler) currentCRD(ctx context.Context, name types.NamespacedName) (bool, *apiextensionsv1.CustomResourceDefinition, error) {
```
candita, line 73: I wonder if we should be storing and checking versions on CRDs?  And in the feature gate name.  When we move from v1beta1 to v1, will there be a chicken and egg upgrade problem?

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	c, err := controller.NewUnmanaged(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+	isOurGatewayClass := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		class := o.(*gatewayapiv1beta1.GatewayClass)
+		return class.Spec.ControllerName == OpenShiftGatewayClassControllerName
+	})
+	if err := c.Watch(&source.Kind{Type: &gatewayapiv1beta1.GatewayClass{}}, &handler.EnqueueRequestForObject{}, isOurGatewayClass); err != nil {
```
candita, line 54: Have we definitely decided that no other "custom" GatewayClasses can be created?

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -0,0 +1,218 @@
...
+func desiredServiceMeshControlPlane(name types.NamespacedName, ownerRef metav1.OwnerReference) (*maistrav2.ServiceMeshControlPlane, error) {
+	pilotContainerEnv := map[string]string{
+		"PILOT_ENABLE_GATEWAY_API":                       "true",
+		"PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER": "true",
+		"PILOT_ENABLE_GATEWAY_API_STATUS":                "true",
+		"PILOT_GATEWAY_API_CONTROLLER_NAME":              OpenShiftGatewayClassControllerName,
+	}
+	f := false
+	t := true
+	smcp := maistrav2.ServiceMeshControlPlane{
```
candita, line 71: It seems risky to hard code the SMCP values before we have attained release alignment.  Things could change on OSSM between OpenShift releases.  Is there any way to do this without hard-coding it ?

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -0,0 +1,218 @@
...
+						Container: &maistrav2.ContainerConfig{
+							Env: pilotContainerEnv,
+						},
+					},
+				},
+			},
+			Security: &maistrav2.SecurityConfig{
+				ManageNetworkPolicy: &f,
+			},
+			TechPreview: maistrav1.NewHelmValues(map[string]interface{}{
```
candita, line 148: This name is going to change, isn't it?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -0,0 +1,120 @@
...
+
+// updateSubscription updates a subscription.
+func (r *reconciler) updateSubscription(ctx context.Context, current, desired *operatorsv1alpha1.Subscription) (bool, error) {
+	changed, updated := subscriptionChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
```
candita, line 100: It doesn't look like like anything in this `Subscription` would ever change even if the version changed.  I wonder if there needs to be another field set in `desiredSubscription`, e.g. `StartingCSV` ?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -0,0 +1,120 @@
...
+			Name:      name.Name,
+		},
+		Spec: &operatorsv1alpha1.SubscriptionSpec{
+			// TODO: Use named consts if available.
+			Channel:                "stable",
+			InstallPlanApproval:    operatorsv1alpha1.ApprovalAutomatic,
+			Package:                "servicemeshoperator",
+			CatalogSource:          "redhat-operators",
+			CatalogSourceNamespace: "openshift-marketplace",
+		},
```
candita, line 66: It doesn't look like like anything in this Subscription would ever change even if the version changed. I wonder if there needs to be another field set in desiredSubscription, e.g. StartingCSV ?

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	c, err := controller.NewUnmanaged(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+	isServiceNeedingDNS := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		svc := o.(*corev1.Service)
+		return len(svc.Labels["gateway.istio.io/managed"]) != 0
+	})
+	gatewayListenerChanged := predicate.Funcs{
```
candita, line 58: Some other things could change, like ports, protocol, etc https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Listener

pkg/operator/controller/dns/controller.go
Miciah: We could, but if people wanted to abuse this controller, they could just as easily create a dnsrecord CR with whatever label we wrote the controller to check for.  Checking labels seems like unnecessary complexity with little benefit.

pkg/operator/controller/gatewayapi/controller.go
Miciah: The current logic doesn't undo anything when the feature gate is turned off.  Where would it be useful to be able to refer to a stored value?

pkg/operator/controller/gatewayapi/controller.go
Miciah: Good point.  I'll add a predicate on the watch so that we don't get reconcile requests bogus featuregates.

pkg/operator/controller/gatewayapi/controller.go
Miciah: This is in case a feature set (such as "TechPreviewNoUpgrade") includes the "GatewayAPI" feature.  It isn't needed at this time because there is no feature set that includes the "GatewayAPI" feature; it's just here for correctness and future-proofing.

pkg/operator/controller/gatewayapi/controller_test.go
Miciah: There is no "ReconcileOnlyStartsControllerOnce" function or method.

pkg/operator/controller/gatewayapi/crds.go
Miciah: We reach this line if `have` is true and no update was needed.

pkg/operator/controller/gatewayapi/crds.go
Miciah: Yeah, I'll add a to-do for updating CRDs.  I'm not sure what you mean about the feature gate name.  Keep in mind that we don't support upgrades in dev preview.

pkg/operator/controller/gatewayclass/controller.go
Miciah: I'm not sure what you mean.  We only care about gatewayclasses that specify our controller name.  The cluster admin absolutely can create gatewayclasses that specify other controller names, but then any such named controllers (not ours) are responsible for managing those gatewayclasses.

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
Miciah: What do you have in mind?  Do you envision that the ingress operator could download the SMCP configuration from some out-of-payload source?  That seems like it might be fragile and a potential attack vector, so I'd want us to design any such mechanism very carefully.  For now, I'd rather not deviate to far from the design in the EP.  If something changes in the SMCP definition, we can address it as a bug fix.

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
Miciah: Possibly with OSSM 2.4.

pkg/operator/controller/gatewayclass/subscription.go
Miciah: What is the advantage of specifying `StartingCSV` in the Subscription?  The only thing I can think of is if we need a specific minimum 2.3.z release (e.g. 2.3.2 if it has working automated deployments).

pkg/operator/controller/service-dns/controller.go
Miciah: Port and protocol shouldn't matter; the DNS record only changes if the host name changes.

pkg/operator/controller/gatewayapi/controller_test.go
Miciah: I would say "not yet"; eventually this feature will be tech preview, and then "GatewayAPI" may be added to the "TechPreviewNoUpgrade" feature set.

pkg/operator/controller/gatewayapi/controller.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/9742e1ad86e7114b21eb78fde9be83caccc596b3..d4ac3988e55abd6c35f865e03366a20245ad4256 (I did a rebase at the same time—sorry about that!—I'll try to remember to do the rebase and other changes in separate pushes in the future).

pkg/operator/controller/gatewayapi/crds.go
Miciah: I added a to-do comment in https://github.com/openshift/cluster-ingress-operator/compare/9742e1ad86e7114b21eb78fde9be83caccc596b3..d4ac3988e55abd6c35f865e03366a20245ad4256.

pkg/operator/controller/service-dns/controller.go
Miciah: I added a comment explaining this in https://github.com/openshift/cluster-ingress-operator/compare/9742e1ad86e7114b21eb78fde9be83caccc596b3..d4ac3988e55abd6c35f865e03366a20245ad4256.

Miciah:
Latest push puts the ServiceMeshControlPlane CR back in the "openshift-ingress" namespace.  Previously it was understood that the automated deployment feature didn't work if the ServiceMeshControlPlane CR and Gateway CR were in the same namespace; now it turns out that they _must_ be in the same namespace.

Miciah:
Latest push adds commit https://github.com/openshift/cluster-ingress-operator/pull/890/commits/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44 to improve robustness of DNS record cleanup when a gateway is deleted.

pkg/dns/aws/dns.go
```diff
@@ -503,7 +516,40 @@ func (m *Provider) change(record *iov1.DNSRecord, zone configv1.DNSZone, action
...
+		err = fmt.Errorf("failed to get hosted zone for load balancer target %q: %v", target, err)
+		if v, ok := record.Annotations[targetHostedZoneIdAnnotationKey]; !ok {
+			return err
+		} else {
+			log.Error(err, "falling back to the "+targetHostedZoneIdAnnotationKey+" annotation", "value", v)
+			targetHostedZoneID = v
+		}
+	}
+	// If this is an upsert, store the target hosted zone id in an
+	// annotation on the DNSRecord CR in case we later on we need the id and
```
Miciah, line 528:
```suggestion
	// annotation on the DNSRecord CR in case we later on need the id and
```

pkg/operator/controller/service-dns/controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+					},
+					ingHost("newlb.example.com"),
+				),
+				dnsrecord("example-gateway-55bcfdb97d-wildcard", "*.example.com", "oldlb.example.com"),
+			},
+			reconcileRequest: req("openshift-ingress", "example-gateway"),
+			expectCreate:     []client.Object{},
+			expectUpdate: []client.Object{
+				dnsrecord("example-gateway-55bcfdb97d-wildcard", "*.example.com", "newlb.example.com"),
+			},
```
gcs278, line 170: What if a dnsRecord already exists with the same domain name as a gateway listener specifies? Will two dnsrecords be created? Is that worth a unit test?

pkg/operator/controller/service-dns/controller_test.go
```diff
@@ -0,0 +1,267 @@
...
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	iov1.AddToScheme(scheme)
```
gcs278, line 175: *nit* deprecated. `Install()`?

pkg/resources/dnsrecord/dns.go
```diff
@@ -27,58 +32,102 @@ import (
...
+			return CurrentDNSRecord(client, name)
+		}
+	}
+
+	return haveWC, current, nil
+}
+
+// EnsureDNSRecord will create DNS records for the given LB service.  If service
+// is nil (haveLBS is false), nothing is done.
+func EnsureDNSRecord(client client.Client, name types.NamespacedName, dnsRecordLabels map[string]string, ownerRef metav1.OwnerReference, domain string, service *corev1.Service) (bool, *iov1.DNSRecord, error) {
```
gcs278, line 68: Is there a better way to structure this code opposed to having separate functions for `EnsureDNSRecord` and `EnsureWildCardDNSRecord`? It feels a little repetitive to me.

gcs278 (commented):
Looks great, overall no major issues. Given the circumstances, feel free to disagree, address later, or mark it as a todo so we can move forward.

pkg/dns/aws/dns.go
```diff
@@ -503,7 +516,40 @@ func (m *Provider) change(record *iov1.DNSRecord, zone configv1.DNSZone, action
...
+		name := types.NamespacedName{
+			Namespace: record.Namespace,
+			Name:      record.Name,
+		}
+		if err := m.config.Client.Get(context.TODO(), name, &current); err != nil {
+			// Log the error and continue.  The annotation is only
+			// needed as a fallback mechanism, and anyway we might
+			// succeed in adding it on the next upsert.
+			log.Error(err, "failed to get dnsrecord", "dnsrecord", name)
+		} else if _, ok := record.Annotations[targetHostedZoneIdAnnotationKey]; !ok {
```
Miciah, line 541:
```suggestion
		} else if _, ok := current.Annotations[targetHostedZoneIdAnnotationKey]; !ok {
```

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
```diff
@@ -0,0 +1,433 @@
+apiVersion: apiextensions.k8s.io/v1
```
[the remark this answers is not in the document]
Miciah: Good point.  OSSM 2.4's Istio started just fine without the CRD, though, whereas I remember that OSSM 2.3's Istio failed to start if I didn't create the referencepolicies, tcproutes, tlsroutes, and udproutes CRDs.  I don't know whether OSSM 2.4 creates the missing CRDs itself, whether it functions without the CRDs, or what.  Anyway, I'll go ahead and add the ReferenceGrant CRD.

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
```diff
@@ -0,0 +1,433 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1086
+    gateway.networking.k8s.io/bundle-version: v0.6.0-dev
```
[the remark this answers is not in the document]
Miciah: Good point.  https://github.com/maistra/istio/blob/maistra-2.4/go.mod#L106 (permalink: https://github.com/maistra/istio/blob/7b1d1a9b9be85e448ee8b235d1f5a52b8d09955b/go.mod#L106) has `sigs.k8s.io/gateway-api v0.5.1-0.20220921185115-ee7a83814203`.  https://github.com/kubernetes-sigs/gateway-api/compare/ee7a83814203...v0.6.1 has some improvements to API field descriptions, which would be nice to have.  However, the diff also shows that some status conditions definitions and defaults changed, some `additionalPrinterColumns` definitions changed (which affects oc/kubectl output), and the referencegrants CRD was promoted from experimental to standard.  It makes sense to use CRDs that match the Maistra version so that status conditions match.  If referencegrants work in Maistra 2.4, then it must be using the experimental CRDs from v0.5.1-0.20220921185115-ee7a83814203, so we should use those here.  

This also means we'll want to bump the vendored CRDs when OSSM bumps its vendored API definitions.  That's something we'll need to keep in mind.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -0,0 +1,147 @@
...
+	var fg configv1.FeatureGate
+	if err := r.cache.Get(ctx, request.NamespacedName, &fg); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("featuregate not found; reconciliation will be skipped", "request", request)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if !featureIsEnabled("GatewayAPI", &fg) {
```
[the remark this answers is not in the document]
Miciah: It would be cleaner; I'll make that change.

pkg/operator/controller/gatewayapi/crds.go
```diff
@@ -0,0 +1,129 @@
...
+		return have, current, err
+	}
+
+	// TODO Add update logic.
+	switch {
+	case !have:
+		if err := r.createCRD(ctx, desired); err != nil {
+			return false, nil, err
+		}
+		return r.currentCRD(ctx, name)
```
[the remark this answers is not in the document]
Miciah: Yes, the objective is consistency among `ensureX` functions.

pkg/operator/controller/gatewayapi/crds.go
```diff
@@ -0,0 +1,129 @@
...
+	name := types.NamespacedName{
+		Namespace: desired.Namespace,
+		Name:      desired.Name,
+	}
+	have, current, err := r.currentCRD(ctx, name)
+	if err != nil && !errors.IsNotFound(err) {
+		return have, current, err
+	}
+
+	// TODO Add update logic.
```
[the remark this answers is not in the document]
Miciah: Uhm...  Yeah.  I added the to-do comment in response to https://github.com/openshift/cluster-ingress-operator/pull/890#discussion_r1110391672, but you're right—the update logic is already there.

pkg/operator/controller/gatewayapi/controller_test.go
```diff
@@ -0,0 +1,362 @@
...
+	scheme := runtime.NewScheme()
+	configv1.AddToScheme(scheme)
+	apiextensionsv1.AddToScheme(scheme)
+	fg := &configv1.FeatureGate{
```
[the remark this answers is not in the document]
Miciah: You mean so that `TestReconcileOnlyStartsControllerOnce` and the "featuregates.config.openshift.io/cluster specifies CustomNoUpgrade and enables GatewayAPI" test case in `Test_Reconcile` could use the same definition?  I could, but I don't think that would really make the tests easier to read, especially since the definition is only used in two places (in separate tests, so there'd be a bit of jumping around, to this global definition and then back to the test) and all the other test cases in `Test_Reconcile` use different featuregate definitions (so the odd one out would be a bit jarring when reading through the test cases).

pkg/operator/controller/gatewayapi/controller_test.go
```diff
@@ -0,0 +1,362 @@
...
+			},
+			expectCreate:    []client.Object{},
+			expectUpdate:    []client.Object{},
+			expectDelete:    []client.Object{},
+			expectStartCtrl: false,
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	configv1.AddToScheme(scheme)
```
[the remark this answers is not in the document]
Miciah: Yeah, looks like you're right.  I don't understand the API machinery well, but the godoc for `configv1.AddToScheme` and `configv1.Install` is clear enough.

pkg/operator/controller/gatewayapi/controller_test.go
```diff
@@ -0,0 +1,362 @@
...
+	apiextensionsv1.AddToScheme(scheme)
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			fakeClient := fake.NewClientBuilder().
```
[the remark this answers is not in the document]
Miciah: Looks like we came up with a similar solution (both inspired by Andrey's E2E in in ALBO).  Let's refactor in a follow-up.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	controllerName = "gatewayclass_controller"
+
+	// OpenShiftGatewayClassControllerName is the string by which a
+	// gatewayclass identifies itself as belonging to OpenShift Istio.  If a
+	// gatewayclass's spec.controllerName field is set to this value, then
+	// the gatewayclass is ours.
+	OpenShiftGatewayClassControllerName = "openshift.io/gateway-controller"
```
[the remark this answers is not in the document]
Miciah: I thought I saw an "istio" gatewayclass on my cluster, and I thought I saw "istio" hard-coded in Istio.  Now that I check the Maistra code, I see that the default gatewayclass's name is configurable: https://github.com/maistra/istio/blob/maistra-2.4/pilot/pkg/features/pilot.go#L397-L398 (permalink: https://github.com/maistra/istio/blob/7b1d1a9b9be85e448ee8b235d1f5a52b8d09955b/pilot/pkg/features/pilot.go#L397-L398).  I don't think we can turn off creation of this default gatewayclass (the controller defined here always creates it if it doesn't already exist: https://github.com/maistra/istio/blob/maistra-2.4/pilot/pkg/config/kube/gateway/gatewayclass.go), but I can at least configure its name to be "ocp", or "openshift-default".  

I'd rather leave gatewayclass creation to the cluster admin, but in the absence of that option, I'll set `PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS=openshift-default`.  Does that seem reasonable?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -0,0 +1,120 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ensureServiceMeshOperatorSubscription attempts to ensure that a subscription
+// for servicemeshoperator is present and returns a Boolean indicating whether
+// it exists, the subscription if it exists, and an error value.
+func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context) (bool, *operatorsv1alpha1.Subscription, error) {
+	name := types.NamespacedName{
+		// TODO Use named consts for the namespace and name.
```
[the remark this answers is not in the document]
Miciah: Actually, this to-do makes sense to do now.

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -0,0 +1,220 @@
...
+				"controlPlaneMode": "ClusterScoped",
+			}),
```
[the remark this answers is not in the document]
Miciah: The current definition puts us in a place where things work, and it took us a while to get to that place, so I'm inclined to keep it this way.  (Although I believe the `techPreview` part will need to be changed at some point between now and GA, so we might need to revisit this anyway.)

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const controllerName = "service_dns_controller"
+
+var log = logf.Logger.WithName(controllerName)
+
+// NewUnmanaged creates and returns a controller that watches services that have
+// the "dns.openshift.io/domain" annotation and creates dnsrecord objects for
```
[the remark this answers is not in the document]
Miciah: Oops!  Stale comment.

pkg/operator/controller/names.go
```diff
@@ -250,3 +262,25 @@ func CanaryRouteName() types.NamespacedName {
...
+}
+
+// GatewayDNSRecordName returns the namespaced name for a DNSRecord CR
+// associated with a Gateway.  This CR is created in the Gateway's namespace and
+// is named using the Gateway's name, listener's hashed host name, and the
+// suffix "-wildcard".
+func GatewayDNSRecordName(gateway *gatewayapiv1beta1.Gateway, host string) types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: gateway.Namespace,
+		Name:      fmt.Sprintf("%s-%s-wildcard", gateway.Name, util.Hash(host)),
```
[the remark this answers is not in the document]
Miciah: In that case, the create should fail.  We have a lot of places like that (e.g. the names of resources related to ingresscontrollers also usually embed the associated ingresscontroller's name).

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,230 @@
...
+		cache:    mgr.GetCache(),
+		recorder: mgr.GetEventRecorderFor(controllerName),
+	}
+	c, err := controller.NewUnmanaged(controllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+	isServiceNeedingDNS := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		svc := o.(*corev1.Service)
+		return len(svc.Labels["gateway.istio.io/managed"]) != 0
```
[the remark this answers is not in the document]
Miciah: I haven't looked at that.  Is there a reasonable label to key off of for `istio-ingressgateway`?  I guess the SMCP could explicitly tell OSSM to add the `gateway.istio.io/managed` label.

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,230 @@
...
+			return false
+		},
+	}
+	isInOperandNamespace := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetNamespace() == config.OperandNamespace
+	})
+	gatewayToService := func(o client.Object) []reconcile.Request {
+		var services corev1.ServiceList
+		listOpts := []client.ListOption{
+			client.MatchingLabels{"istio.io/gateway-name": o.GetName()},
```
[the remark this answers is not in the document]
Miciah: Actually, I can use https://github.com/maistra/istio/blob/maistra-2.4/pilot/pkg/config/kube/gateway/conversion.go#L1453 (permalink: https://github.com/maistra/istio/blob/7b1d1a9b9be85e448ee8b235d1f5a52b8d09955b/pilot/pkg/config/kube/gateway/conversion.go#L1453).

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	// but don't match any of the gateway's hostnames.  Such DNSRecord CRs
+	// may exist if a hostname was modified or deleted on the gateway.
+	listOpts := []client.ListOption{
+		client.MatchingLabels{"istio.io/gateway-name": gatewayName.Name},
+		client.InNamespace(r.config.OperandNamespace),
+	}
+	var dnsrecords iov1.DNSRecordList
+	// Use the client rather than the cache to make sure we don't use stale
+	// data and fail to clean up stale dnsrecords.
+	if err := r.client.List(ctx, &dnsrecords, listOpts...); err != nil {
```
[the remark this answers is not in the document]
Miciah: Yeah, this `Reconcile` definition has gradually grown rather long.  I'll refactor.

pkg/operator/controller/service-dns/controller_test.go
Miciah: If it has the same host name, then it hashes to the same name for the CR, so the controller will update the existing CR rather than creating a new one.

Miciah: Or do you mean if two gateways specify the same host name?

pkg/resources/dnsrecord/dns.go
Miciah: I could move the wildcard prefixing logic to the place that calls `EnsureWildCardDNSRecord`, but my goal was to avoid changing that code more than necessary.

pkg/operator/controller/gatewayapi/crds.go
Miciah: I removed the to-do comment; see https://github.com/openshift/cluster-ingress-operator/pull/890#discussion_r1113254640.

pkg/dns/aws/dns.go
Miciah: Fixed.

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayapi/controller.go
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayapi/crds.go
```diff
@@ -0,0 +1,129 @@
...
+// ensureCRD attempts to ensure that the specified CRD exists and returns a
+// Boolean indicating whether it exists, the CRD if it does exist, and an error
+// value.
+func (r *reconciler) ensureCRD(ctx context.Context, desired *apiextensionsv1.CustomResourceDefinition) (bool, *apiextensionsv1.CustomResourceDefinition, error) {
+	name := types.NamespacedName{
+		Namespace: desired.Namespace,
+		Name:      desired.Name,
+	}
+	have, current, err := r.currentCRD(ctx, name)
+	if err != nil && !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayapi/crds.go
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayapi/controller_test.go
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayclass/controller.go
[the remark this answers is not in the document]
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.  Side-effect:  Now you need to specify `gatewayClassName: openshift-default` instead of `gatewayClassName: istio` on your gatewayclass.  :tada:!

pkg/operator/controller/gatewayclass/subscription.go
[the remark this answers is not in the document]
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/service-dns/controller.go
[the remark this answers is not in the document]
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/service-dns/controller.go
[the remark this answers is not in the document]
Miciah: I decided against adding an import on github.com/maistra/istio, so I just defined some consts in this file.  Done in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/service-dns/controller.go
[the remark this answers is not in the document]
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.  Also, I found a problem in the `gatewayListenersChanged` predicate; I refactored the predicate, rewrote it to fix the problem, and added a test.

pkg/operator/controller/service-dns/controller_test.go
Miciah: Fixed.

pkg/dns/aws/dns.go
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/c1c4ef672c6a88f64ad61f33ffaa30d83b993a44..503e0364c54c9f869b0ea8aaa7175b7bf458dca8.

pkg/operator/controller/gatewayapi/controller_test.go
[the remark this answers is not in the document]
gcs278: Ack, makes sense.

pkg/operator/controller/gatewayapi/controller_test.go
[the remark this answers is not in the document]
gcs278: Ack, sounds good.

pkg/operator/controller/service-dns/controller_test.go
gcs278: I meant if you have an outside actor (not the cluster-ingress-operator) creating dnsrecords, one with the same hostname as what you later create a Gateway Listener to have, what's the outcome? Now that I am saying that out loud, I remember you mentioning that admins are not supposed to create DNS Records by hand, so this is probably an unsupported situation.

pkg/resources/dnsrecord/dns.go
gcs278: Makes sense, we can refactor later to prevent more churn.

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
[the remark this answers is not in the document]
gcs278: Ack, I'm good with that.

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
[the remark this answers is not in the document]
gcs278: Okay make sense, and the fact reference grant will still be in under the v1alpha2 API should be in our release documentation (though documenting the version of Gateway API also might suffice).

So, at the moment, we are still using `v0.6.0-dev`. Is intentional? Is it better to align with the version that OSSM 2.4 is currently vendoring?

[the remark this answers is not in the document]
Miciah: Oh, facepalm, I updated manifests but not the vendored gateway-api.  I don't know that it matters, but I'll fix that just in case.

pkg/operator/controller/service-dns/controller_test.go
Miciah: Yeah, the only thing that should ever create or update dnsrecords is cluster-ingress-operator itself.  (Nothing else should read dnsrecords either, other than must-gather.)

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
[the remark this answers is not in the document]
Miciah: https://github.com/openshift/cluster-ingress-operator/compare/503e0364c54c9f869b0ea8aaa7175b7bf458dca8..53c61b464207eeeee5709eab343757c1d3a10f3f fixes the vendored gateway-api package.

pkg/operator/controller/service-dns/controller.go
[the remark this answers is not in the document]
gcs278: Hmm. Good question. Here's what I have for `istio-ingressgateway` service:
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{"maistra.io/mesh-generation":"development-1"},"labels":{"app":"istio-ingressgateway","app.kubernetes.io/component":"istio-ingress","app.kubernetes.io/instance":"openshift-ingress","app.kubernetes.io/managed-by":"maistra-istio-operator","app.kubernetes.io/name":"istio-ingress","app.kubernetes.io/part-of":"istio","app.kubernetes.io/version":"development-1","istio":"ingressgateway","istio.io/rev":"openshift-gateway","maistra-version":"2.4.0","maistra.io/owner":"openshift-ingress","maistra.io/owner-name":"openshift-gateway","release":"istio"},"name":"istio-ingressgateway","namespace":"openshift-ingress","ownerReferences":[{"apiVersion":"maistra.io/v2","blockOwnerDeletion":true,"controller":true,"kind":"ServiceMeshControlPlane","name":"openshift-gateway","uid":"74d05a19-ffab-4bef-a25f-e1c0d0e2237f"}]},"spec":{"ports":[{"name":"status-port","port":15021,"protocol":"TCP","targetPort":15021},{"name":"http2","port":80,"protocol":"TCP","targetPort":8080},{"name":"https","port":443,"protocol":"TCP","targetPort":8443}],"selector":{"app":"istio-ingressgateway","istio":"ingressgateway"},"type":"LoadBalancer"}}
    maistra.io/mesh-generation: development-1
  creationTimestamp: "2023-02-22T14:53:27Z"
  finalizers:
  - service.kubernetes.io/load-balancer-cleanup
  labels:
    app: istio-ingressgateway
    app.kubernetes.io/component: istio-ingress
    app.kubernetes.io/instance: openshift-ingress
    app.kubernetes.io/managed-by: maistra-istio-operator
    app.kubernetes.io/name: istio-ingress
    app.kubernetes.io/part-of: istio
    app.kubernetes.io/version: development-1
    istio: ingressgateway
    istio.io/rev: openshift-gateway
    maistra-version: 2.4.0
    maistra.io/owner: openshift-ingress
    maistra.io/owner-name: openshift-gateway
    release: istio
  name: istio-ingressgateway
  namespace: openshift-ingress
  ownerReferences:
  - apiVersion: maistra.io/v2
    blockOwnerDeletion: true
    controller: true
    kind: ServiceMeshControlPlane
    name: openshift-gateway
    uid: 74d05a19-ffab-4bef-a25f-e1c0d0e2237f
  resourceVersion: "100493"
  uid: 1d5b3120-ce7d-492a-9603-76be1144aad7
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 172.30.32.183
  clusterIPs:
  - 172.30.32.183
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: status-port
    nodePort: 30200
    port: 15021
    protocol: TCP
    targetPort: 15021
  - name: http2
    nodePort: 31149
    port: 80
    protocol: TCP
    targetPort: 8080
  - name: https
    nodePort: 31413
    port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    app: istio-ingressgateway
    istio: ingressgateway
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - hostname: a1d5b3120ce7d492a960376be1144aad-291825524.us-east-1.elb.amazonaws.com
```

> I guess the SMCP could explicitly tell OSSM to add the gateway.istio.io/managed label.

I'm guessing you are referring to `smcp.spec.gateways.ingress.service.metadata.labels`. That seems preferable, since that's a value we can control. It doesn't look like `smcp.spec.gateways.ingress.service.metadata.labels` applies to auto-deployment services just as an FYI.

My personal gut feeling is that we should support the default `istio-ingressgateway` in the same way we support auto-deployments, otherwise it's a bit of a second class citizen. I think this would be pretty trivial right? just add `smcp.spec.gateways.ingress.service.metadata.labels` and it should work?

assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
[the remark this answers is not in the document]
gcs278: Oh I actually forgot about `go.mod`'s version Gateway API, but good catch. I'm actually talking about the version `gateway.networking.k8s.io/bundle-version: v0.6.0-dev` that is still in the manifest. That isn't correct right? Still says `v0.6.0-dev`.

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	controllerName = "service_dns_controller"
```
gcs278, line 37: *nit* you named this service-dns controller, but in reality, it's a Gateway API DNS controller at the moment, since it only reconciles services belonging to a Gateway auto deployment. 

Is the grand plan to ever expand this past Gateway API? If not, would it be better to call this controller `gatewayapi-dns` to better reflect what it's accomplishing?

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// ensureDNSRecordsForGateway ensures that a DNSRecord CR exists, associated
+// with the given gateway and service, for each of the given domains.  It
+// returns a list of any errors that result from ensuring those DNSRecord CRs.
+func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *gatewayapiv1beta1.Gateway, service *corev1.Service, domains []string) []error {
+	labels := map[string]string{
+		gatewayNameLabelKey: gateway.Name,
+	}
+	for k, v := range service.Labels {
```
gcs278, line 210: What's the motivation behind taking all of the service labels and placing them on the dns record?

pkg/operator/controller/service-dns/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return errs
+}
+
+// deleteStaleDNSRecordsForGateway deletes any DNSRecord CRs that are associated
+// with the given gateway but specify a DNS name that is not in the given set of
+// domains.  Such DNSRecord CRs may exist if a hostname was modified or deleted
+// on the gateway.  deleteStaleDNSRecordsForGateway returns a list of any errors
+// that result from deleting those DNSRecord CRs.
+func (r *reconciler) deleteStaleDNSRecordsForGateway(ctx context.Context, gateway *gatewayapiv1beta1.Gateway, service *corev1.Service, domains sets.String) []error {
```
gcs278, line 233: *nit* `sets.String` is supposedly deprecated. Goland says you should use a generic set instead.

pkg/operator/controller/service-dns/controller.go
Miciah: I figured the service labels likely have useful information for understanding how these resources relate to a given gateway or manager/controller.  I can drop it if it isn't useful.

gcs278: If it doesn't hurt anything, I don't mind. Just was curious, maybe it would help us in the future.

pkg/operator/controller/service-dns/controller.go
gcs278: Either is fine with me, though, I was thinking of having `gatewayapi-...` in the name to help associate specifically to Gateway API,  also keeping in mind your other controller is called `gatewayapi`. But yea, technically it is `gateway-service-dns`, since it's doing DNS for Gateways.

I have no strong opinions. Whatever you think would help describe the duty of the controller.

pkg/operator/controller/service-dns/controller.go
gcs278: Sounds good, we can do it later.

pkg/operator/controller/service-dns/controller.go
Miciah: Come to think of it, I might have been inspired by Istio:

> Annotations and labels on the `Gateway` will be copied to the `Service` and `Deployment`.

https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment

pkg/operator/controller/service-dns/controller.go
Miciah: https://github.com/openshift/cluster-ingress-operator/compare/53c61b464207eeeee5709eab343757c1d3a10f3f..e8f3f45aec4ee7d1dc602bdd3ba8a3a841ddc336 renames "service-dns" to "gateway-service-dns".

gcs278:
Thanks for all of the hard work in getting this done in a limited amount of time!

lihongan:
/label qe-approved
did pre-merge testing and passed ( see https://issues.redhat.com/browse/NE-1226 )

Miciah:
I'm adding the "docs-approved" and "px-approved" labels myself because this is a dev preview feature that does not require product documentation or tech enablement.  

/label docs-approved
/label px-approved

PR #891, opened 2023-02-15
OCPBUGS-6661: Handle mTLS CRLs in the router

rfredette:
CRL lifecycle management is being moved to the router pod to avoid hitting the configmap max size (1MB).

This relies on openshift/router#447 to function.


pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -1,312 +0,0 @@
...
-			Namespace: crlConfigmapName.Namespace,
-		},
-		Data: map[string]string{
-			"crl.pem": crlData,
-		},
-	}
-	crlConfigmap.SetOwnerReferences([]metav1.OwnerReference{ownerRef})
-
-	return true, &crlConfigmap, context.WithValue(ctx, "nextCRLUpdate", nextCRLUpdate), nil
-}
```
Miciah, line 224: I think you should keep the controller for now so that it can delete any stale configmap.
```suggestion
func desiredCRLConfigMap(ctx context.Context, ic *operatorv1.IngressController, ownerRef metav1.OwnerReference, clientCAData []byte, crls map[string]*pkix.CertificateList) (bool, *corev1.ConfigMap, context.Context, error) {
	// The ingress operator no longer needs to create a CRL configmap;
	// the router now downloads CRLs itself (see OCPBUGS-whatever).
	return false, nil, ctx, nil
}
```

Miciah: We should also have a to-do comment to remove the controller after 4.13 though.

rfredette: Good point. I will make that change. Do we need to consider EUS-to-EUS upgrades? If so, it may need to stay there to delete stale configmaps until after 4.14

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -23,6 +23,8 @@ func (r *reconciler) ensureCRLConfigmap(ctx context.Context, ic *operatorv1.Ingr
 		return false, nil, ctx, err
 	}
 
+	// The CRL management code has been moved into the router, so the CRL configmap is no longer necessary.
+	// TODO: Remove this whole controller after 4.13
```
frobware, line 27: Let's squash this commit.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1024,68 +1022,6 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 			clientAuthCAPath := filepath.Join(clientCAVolumeMount.MountPath, clientCABundleFilename)
```
gcs278, line 1022: https://github.com/openshift/cluster-ingress-operator/pull/891/files#diff-03760f03eeac1174e38e1ec5fecadcab4cb38d05e00509f201d7ae64dc1c7c36L90 `RouterClientAuthCRL    = "ROUTER_MUTUAL_TLS_AUTH_CRL"` is unused

gcs278 (commented):
No major issues from me.

pkg/operator/controller/crl/crl_configmap.go
gcs278, line 1: I feel like we need E2E testing for CRLs, especially since this is a large update. I see no E2E test. Is this the time to do it?

rfredette: Yes, I will add e2e tests for this

rfredette:
I've pushed a temporary commit that makes the router image the latest commit in openshift/router#447 so that the new tests work. Once that router PR merges, I will revert the router image change.

/hold

gcs278 (commented):
still reviewing, but adding a couple of comments

test/e2e/client_tls_test.go
```diff
@@ -473,3 +931,60 @@ func encodeKey(key *rsa.PrivateKey) string {
 		Bytes: x509.MarshalPKCS1PrivateKey(key),
 	}))
 }
+
+// curl execs a Curl command in the test client pod and returns an error value.  The Curl command uses the specified
+// certificate from the client certificates configmap, sends a request for the canary route via the router's internal
+// service. Returns the HTTP status code returned from curl, and an error either if there is an HTTP error, or if
+// there's another error in running the command. If the error was not an HTTP error, the HTTP status code returned will
+// be -1.
+func curl(t *testing.T, clientPod *corev1.Pod, certName, endpoint, ingressControllerIP string, verbose bool) (int64, error) {
```
gcs278, line 940: *nit* should the name of this function reflect that it's getting a status code? i.e. `curlHttpStatus` or `getHttpStatusCode`?

go.mod
```diff
@@ -22,7 +22,8 @@ require (
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
 	github.com/openshift/client-go v0.0.0-20230120202327-72f107311084
 	github.com/openshift/library-go v0.0.0-20230209193239-2e9167362f31
-	github.com/operator-framework/api v0.3.7-0.20200528122852-759ca0d84007
+	github.com/operator-framework/api v0.15.0
```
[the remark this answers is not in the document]
rfredette: The `github.com/operator-framework/api` bump came automatically with adding `github.com/operator-framework/operator-lib`, which is needed for the `ReadProxyVarsFromEnv` function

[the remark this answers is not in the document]
gcs278: Ack thanks.

test/e2e/client_tls_test.go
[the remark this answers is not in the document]
rfredette: Yes, that makes sense. I will add a test for that.

test/e2e/client_tls_test.go
gcs278: Ehh sorry I see you didn't write this, but adapted it. Nevermind this.

test/e2e/util_test.go
```diff
@@ -79,6 +79,20 @@ func buildEchoPod(name, namespace string) *corev1.Pod {
 	}
 }
 
+// returns a SecurityContext with the minimum possible privileges that satisfy restricted pod security requirements
```
gcs278, line 82: *nit* I think it is a go doc standard to start with the function name
```suggestion
// generateUnprivilegedSecurityContext returns a SecurityContext with the minimum possible privileges that satisfy restricted pod security requirements
```

test/e2e/client_tls_test.go
rfredette: I think it's still reasonable to try to convey what's being returned. I'm not sure I love either of those names, but I'll see if I can come up with something that will work

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+		// Name is the name of the test case.
+		Name string
+		// CreateCerts generates the certificates for the test case. Certificates and CRLs must not have expired at the
+		// time of the run, so they must be generated at runtime.
+		CreateCerts func() TestMTLSWithCRLsCerts
+	}{
+		{
+			// This test case has CA certificates including a CRL distribution point (CDP) for the CRL that they
+			// generate and sign. This is the default way to distribute CRLs according to RFC-5280
+			Name: "subject-crl",
```
gcs278, line 352: [from meeting] - better name.

i'm not so sure I understand why "subject" is in the name here, but I could see something like: `crls-included-in-ca-certificates`.

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+						},
+					},
+				}
+			},
+		},
+		{
+			// This test case has certificates including the CRL distribution point of their signer (i.e. intermediate
+			// CA is signed by root CA, and includes the URL for root's CRL). In this case, neither of the certificates
+			// in the CA bundle include the intermediate CRL, so connections that rely on it will be rejected.
+			Name: "signer-crl",
```
gcs278, line 431: I understand this name better, but could be more explicit `crls-included-in-signer-certs` or something.

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+							"revoked-by-root":         revokedByRoot,
+							"revoked-by-intermediate": revokedByIntermediate,
+							"self-signed":             selfSigned,
+						},
+					},
+				}
+			},
+		},
+		{
+			// This test case has certificates including the CRL distribution point of their signer. In this case, a
```
gcs278, line 507: Is it worth referencing the bug in this comment? For additional supporting info.

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("Failed to create client certificate: %v", err)
+				}
+				revokedByIntermediate.CertPem = strings.Join([]string{revokedByIntermediate.CertPem, intermediateCA.CertPem, rootCA.CertPem}, "\n")
+				selfSigned, err := CreateTLSKeyCert("self signed cert", time.Now(), time.Now().Add(24*time.Hour), false, nil, nil)
+				if err != nil {
+					t.Fatalf("Failed to create client certificate: %v", err)
+				}
+
+				_, rootCRLPem, err := CreateCRL(nil, rootCA, time.Now(), time.Now().Add(10*time.Minute), RevokeCertificates(time.Now(), revokedByRoot))
```
gcs278, line 392: Anything significant about this 10 min next update value? Is there any chance that these tests could take 10 mins and something would break? Should it be like an hour to be safe? Does it matter?

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+					t.Errorf("CA Bundle:\n%s\nClient Certificate:\n%s\nCRLs:\n%s", tcCerts.CABundle, tcCerts.ClientCerts.Accepted[certName].CertPem, crlString)
+				}
+			}
+			for certName := range tcCerts.ClientCerts.Rejected {
+				if httpStatusCode, err := curl(t, clientPod, certName, route.Spec.Host, service.Spec.ClusterIP, false); err != nil {
+					if httpStatusCode == 0 {
+						// TLS/SSL verification failures result in a 0 http status code (no connection is made to the backend, so no http status code is returned).
+						continue
+					}
+					t.Errorf("Unexpected error from curl for cert %q: %v", certName, err)
```
gcs278, line 831: Would it be beneficial to add logging of the `httpStatusCode` for future debugging?

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+		// Rejected is a map containing the client keys and certificates that should NOT be able to connect to backends,
+		// indexed by a unique name.
+		Rejected map[string]KeyCert
+	}
+}
+
+// TestMTLSWithCRLs verifies that mTLS works when the client auth chain includes certificate revocation lists (CRLs).
+func TestMTLSWithCRLs(t *testing.T) {
+	t.Parallel()
+	namespaceName := "mtls-with-crls"
```
gcs278, line 337: We've started to use unique namespace names. Here's an example: https://github.com/openshift/cluster-ingress-operator/blob/992b43b3cf3e1784bfe8d3083229c7ecb410e7e3/test/e2e/route_metrics_test.go#L165

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+						return cond.Status == corev1.ConditionTrue, nil
+					}
+				}
+				return false, nil
+			})
+			if err != nil {
+				t.Fatalf("timed out waiting for pod %q to become ready: %v", clientPodName, err)
+			}
+
+			// We need an IP address to which to send requests.  The test client
```
gcs278, line 791: *nit* I struggled a bit on this comment. But I realized it's because the curl function doesn't directly use the DNS name. Is there a better way we can describe this?

```suggestion
			// We need an IP address to which to send requests (in lieu of using a DNS domain name). The test client
```

test/e2e/client_tls_test.go
```diff
@@ -384,6 +313,535 @@ func TestClientTLS(t *testing.T) {
...
+			}
+			if err := kclient.Create(context.TODO(), ic); err != nil {
+				t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+			}
+			defer assertIngressControllerDeleted(t, kclient, ic)
+
+			if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+				t.Fatalf("failed to observe expected conditions: %v", err)
+			}
+
```
gcs278, line 720: *nit* it's a nit pick, but probably worth a block comment here describing clientCerts are going to do, I tend to try to read comments first when deciphering code, feels more effcient:
```suggestion
// Create client certificate bundle as a configmap which we will mount into our test client pod
```

gcs278 (commented):
Generally, I agree with these changes. Most comments are documentation related, but I'm being a bit more nit-picky than usual, since this is a complex E2E.

rfredette:
Test failures seem unrelated; Even the operator suite failure had `TestMTLSWithCRLs` pass.

/retest

rfredette:
more unrelated failures.

/retest

pkg/operator/controller/crl/crl_configmap.go
```diff
@@ -44,32 +23,9 @@ func (r *reconciler) ensureCRLConfigmap(ctx context.Context, ic *operatorv1.Ingr
...
-			clientCAData = []byte(data)
-		}
-	}
-
-	wantCM, desired, ctx, err := desiredCRLConfigMap(ctx, ic, ownerRef, clientCAData, oldCRLs)
-	if err != nil {
-		return false, nil, ctx, fmt.Errorf("failed to build configmap: %w", err)
-	}
+	// The CRL management code has been moved into the router, so the CRL configmap is no longer necessary.
+	// TODO: Remove this whole controller after 4.13
```
Miciah, line 27:
```suggestion
	// TODO: Remove this whole controller after 4.14.
```

rfredette:
Closing this PR in favor of the combined fix in #930

PR #896, opened 2023-03-06
OCPBUGS-8366: manifests: create target namespace earlier

vrutkovs:
In order to speed up initial install target namespace definitions need to be included in the payload, as these need to have SCCs assigned and its benefitial to do this earlier on startup


vrutkovs:
/hold

This is a very hacky way to speed up cluster install. It would be best to improve scc controller / installation proceedure

vrutkovs:
/hold

This is a very hacky way to speed up cluster install. It would be best to improve scc controller / installation procedure

candita:
/close

Per @deads2k: please don't merge.  The CVO manages the operator.  The operator manages to the operand.

PR #900, opened 2023-03-19
RFE-3765 - Allow Ingress to Modify the HAProxy Log Length when using a Sidecar

dafsjr:
A code to implement the [RFE-3765](https://issues.redhat.com//browse/RFE-3765), I also modified the maxLength of the Syslog section to be the same of the Container section.


candita:
@dafsjr Thank you for your contribution.  This PR requires a change in openshift/api, not just edits in the vendor files.  Please make the change in the API and @candita me when you have completed it.

dafsjr:
@candita if I understood correctly, it is done

candita:
> @candita if I understood correctly, it is done

@dafsjr I replied to your message in Slack with an example.

Dockerfile
candita, line 1: You should not need to change the Dockerfile, please revert.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -53,6 +53,8 @@ const (
 	RouterSyslogFacilityEnvName  = "ROUTER_LOG_FACILITY"
 	RouterSyslogMaxLengthEnvName = "ROUTER_LOG_MAX_LENGTH"
 
+	RouterContainerMaxLengthEnvName = "ROUTER_LOG_MAX_LENGTH"
```
candita, line 56: You cannot use the same user-visible value for two different environment variables.

Dockerfile
dafsjr: sorry, I do not have access to the images in Dockerfile, I needed to use the origin image to compile.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -408,8 +408,10 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 	ic.Spec.Logging = &operatorv1.IngressControllerLogging{
 		Access: &operatorv1.AccessLogging{
 			Destination: operatorv1.LoggingDestination{
-				Type:      operatorv1.ContainerLoggingDestinationType,
-				Container: &operatorv1.ContainerLoggingDestinationParameters{},
+				Type: operatorv1.ContainerLoggingDestinationType,
```
candita, line 411: You can add this later, but this section should be to test the default.

candita: This comment is for L519, will repost there.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -514,7 +516,7 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 		{"ROUTER_H1_CASE_ADJUST", true, "Host,Cache-Control"},
 		{"ROUTER_CANONICAL_HOSTNAME", true, "router-" + ic.Name + "." + ic.Status.Domain},
 		{"ROUTER_LOG_FACILITY", false, ""},
-		{"ROUTER_LOG_MAX_LENGTH", false, ""},
+		{"ROUTER_LOG_MAX_LENGTH", true, "8192"},
```
candita, line 519: See https://github.com/openshift/cluster-ingress-operator/pull/900/files/e965950e1d1bb700c455d019389a28104a75c964#r1175768966

pkg/operator/controller/ingress/deployment_test.go
candita: But it does apply to the addition of MaxLength in this section.  This should only be the default, not a changed value in MaxLength.

dafsjr: ok, understood

pkg/operator/controller/ingress/deployment.go
```diff
@@ -47,11 +47,11 @@ const (
...
-	RouterLogLevelEnvName        = "ROUTER_LOG_LEVEL"
-	RouterSyslogAddressEnvName   = "ROUTER_SYSLOG_ADDRESS"
-	RouterSyslogFormatEnvName    = "ROUTER_SYSLOG_FORMAT"
-	RouterSyslogFacilityEnvName  = "ROUTER_LOG_FACILITY"
-	RouterSyslogMaxLengthEnvName = "ROUTER_LOG_MAX_LENGTH"
+	RouterLogLevelEnvName       = "ROUTER_LOG_LEVEL"
+	RouterLogMaxLengthEnvName   = "ROUTER_LOG_MAX_LENGTH"
+	RouterSyslogAddressEnvName  = "ROUTER_SYSLOG_ADDRESS"
+	RouterSyslogFormatEnvName   = "ROUTER_SYSLOG_FORMAT"
+	RouterSyslogFacilityEnvName = "ROUTER_LOG_FACILITY"
```
candita, line 54: If you haven't actually changed anything here, please revert the reordering.  It makes backports more difficult over time.

manifests/00-custom-resource-definition.yaml
candita, line 1: I suspect this needs to be regenerated.

candita:
You will need to squash the commits and add a few unit tests to test your changes to deployment.go.

manifests/00-custom-resource-definition.yaml
dafsjr: The file is exactly vendor/github.com/openshift/api/operator/v1/0000_50_ingress-operator_00-ingresscontroller.crd.yaml, should it be different?

pkg/operator/controller/ingress/deployment.go
candita: I thought we discussed having two environment parameters, one for syslog length and one for sidecar log length?

dafsjr: yeap, the modification I did is exactly for that, the RouterLogMaxLengthEnvName will be the name of variable for both cases, Syslog and Container thus we will have just one var

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1113,10 +1120,18 @@ func accessLoggingForIngressController(ic *operatorv1.IngressController) *operat
...
+		} else {
+			ContainerLoggingParameters = &operatorv1.ContainerLoggingDestinationParameters{
+				MaxLength: ic.Spec.Logging.Access.Destination.Container.MaxLength,
+			}
+		}
 		return &operatorv1.AccessLogging{
 			Destination: operatorv1.LoggingDestination{
 				Type:      operatorv1.ContainerLoggingDestinationType,
-				Container: &operatorv1.ContainerLoggingDestinationParameters{},
+				Container: ContainerLoggingParameters,
```
Miciah, line 1134: Change `ContainerLoggingParameters` to `containerLoggingParameters`.  You can simplify the logic as follows:
```suggestion
		var containerLoggingParameters operatorv1.ContainerLoggingDestinationParameters
		if ic.Spec.Logging.Access.Destination.Container != nil {
			containerLoggingParameters.MaxLength = ic.Spec.Logging.Access.Destination.Container.MaxLength
		}
		return &operatorv1.AccessLogging{
			Destination: operatorv1.LoggingDestination{
				Type:      operatorv1.ContainerLoggingDestinationType,
				Container: &containerLoggingParameters,
```

Miciah (commented):
Since https://github.com/openshift/api/pull/1439 merged, you can use the following commands to use it:
```
go mod edit -replace github.com/openshift/api=github.com/openshift/api@bdd8865676216305be6065ade196f36752dfd015 &&
go mod tidy && go mod vendor && go mod tidy &&
make generate verify test
```
Are you able to squash the code changes into one commit and put the openshift/api vendor bump in a separate commit?

pkg/operator/controller/ingress/deployment.go
candita: Now we have two different inputs for syslog length and sidecar length, and two different places  a single environment variable is set: https://github.com/openshift/cluster-ingress-operator/pull/900/files#diff-03760f03eeac1174e38e1ec5fecadcab4cb38d05e00509f201d7ae64dc1c7c36R767-R790

It's possible that one of these settings will overwrite the other if they are different values.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -777,7 +784,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 			}
 			if accessLogging.Destination.Syslog.MaxLength > 0 {
 				env = append(env, corev1.EnvVar{
-					Name:  RouterSyslogMaxLengthEnvName,
+					Name:  RouterLogMaxLengthEnvName,
```
candita, line 787: If we only have one environment variable for log length, then we need to have either the container or the syslog set it, not both.  E.g. if someone had syslog.maxlength set to 256, and then decided to use container logging, and set that to 300, then the change would be overwritten here.

I think you need to do something like:

```suggestion
if accessLogging.Destination.Container.MaxLength > 0 {
		env = append(env, corev1.EnvVar{
				Name:  RouterLogMaxLengthEnvName,
				Value: fmt.Sprintf("%d", accessLogging.Destination.Container.MaxLength),
		})
} else if if accessLogging.Destination.Syslog.MaxLength > 0 {
....
```

pkg/operator/controller/ingress/deployment_test.go
candita: This is setting up the default for testing.  Don't add MaxLength here.

pkg/operator/controller/ingress/deployment_test.go
candita: E.g. don't do this here.  Add a new test elsewhere, which tests the change from default to a new value.

pkg/operator/controller/ingress/deployment.go
candita: My mistake - I didn't see the `case` statement.  Please disregard.   If you make other changes, can you add a blank line before the `case` statement here?

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,141 @@ func TestContainerLogging(t *testing.T) {
...
+	ic.Spec.Logging = &operatorv1.IngressControllerLogging{
+		Access: &operatorv1.AccessLogging{
+			Destination: operatorv1.LoggingDestination{
+				Type: operatorv1.ContainerLoggingDestinationType,
+				Container: &operatorv1.ContainerLoggingDestinationParameters{
+					MaxLength: int32(8192),
+				},
+			},
+			// String with about 8000 characters
+			HttpLogFormat: strings.Repeat("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@@@@@@@=", 110) + "8192",
```
candita, line 2156: This log format doesn't make sense, and isn't needed.  `HttpLogFormat` defines how to format the log message.

pkg/operator/controller/ingress/deployment_test.go
candita: See `TestDeploymentConfigChanged` for examples.

pkg/operator/controller/ingress/deployment.go
Miciah: The two assignments in https://github.com/openshift/cluster-ingress-operator/pull/900/files#diff-03760f03eeac1174e38e1ec5fecadcab4cb38d05e00509f201d7ae64dc1c7c36R767-R790 are for two different `case` statements in the same `switch` statement:
```go
switch {
case accessLogging.Destination.Type == operatorv1.ContainerLoggingDestinationType:
	// [...]
	// [Set ROUTER_LOG_MAX_LENGTH to accessLogging.Destination.Container.MaxLength.]
	// [...]
case accessLogging.Destination.Type == operatorv1.SyslogLoggingDestinationType:
	// [...]
	// [Set ROUTER_LOG_MAX_LENGTH to accessLogging.Destination.Syslog.MaxLength.]
	// [...]
}
```
I think it should be fine.  Am I missing anything?

candita: No, it was me.  I overlooked the case statement - https://github.com/openshift/cluster-ingress-operator/pull/900#discussion_r1230061006

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,144 @@ func TestContainerLogging(t *testing.T) {
...
+				continue
+			}
+			data, _ := ioutil.ReadAll(readCloser)
+			scanner := bufio.NewScanner(bytes.NewBuffer(data))
+			var found bool
+			for scanner.Scan() {
+				var length int
+				line := scanner.Text()
+				length = len(line)
+				// Due to some haproxy overhead in the log message, the value 8192 generates a log message with 8158 chars
```
candita, line 2258: Unfortunately, if the limit is 8158 then you can't advertise a limit of 8192.  Are you quite sure haproxy is truncating the length shorter in every case or is 8158 the limit?
/hold

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,280 @@ func TestContainerLogging(t *testing.T) {
...
+			data, _ := ioutil.ReadAll(readCloser)
+			scanner := bufio.NewScanner(bytes.NewBuffer(data))
+			var found bool
+			for scanner.Scan() {
+				var length int
+				line := scanner.Text()
+				length = len(line)
+				// HAProxy generate a little overhead in the logline, for this test, based on the name of the router replicaset,
+				// the overhead can be of 61 or 62 chars
+				if strings.Contains(line, "8192abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@@@@@@@=") && (length == 8253 || length == 8254) {
```
candita, line 2258: @dafsjr as I mentioned in our chat, it would be better if you could find out if the content part of the log line is actually 8192 characters, not counting the timestamp and other information in the log line.  I know you responded that this is indeterminate in syslogging and appears to not align with the max length you provided, so I will invite @Miciah to take a look and see what he thinks about how you're verifying the max length here.

Miciah: We know that factors that we cannot predict (such as the replicaset name) affect the length of the log line.  Therefore, doing an exact length comparison is too fragile (e.g. if Kubernetes 1.30 were to change the length of the hash string at the end of the replicaset name, this test would break).  

The goal of the test is to verify that the option has taken effect, not to verify that the log line is truncated to some exact length (which, again, we cannot predict).  My opinion is that these tests should use some values for `MaxLength` that are far apart and verify that the log lines are truncated at approximately the right length.  

You already use `maxLength: 480` in `TestContainerLoggingMinLength` and `maxLength: 8192` here in `TestContainerLoggingMaxLength`, and both of these values are far apart from the default value of 1024, so I think it the following would suffice to verify that the setting is taking effect:

* In `TestContainerLoggingMaxLength` with `maxLength: 8192`, send a line that is >10000 characters, and verify that the log line is truncated to some length that is >8000 and <10000 characters.
* In `TestContainerLoggingMinLength` with `maxLength: 480`, verify that the log line is truncated to some length that is >200 and <700 characters.

This would verify that the setting has taken effect and that HAProxy is truncating to approximately the specified length, allowing a couple hundred characters extra for whatever information (replicaset/pod name, time stamps, and so on) HAProxy or rsyslog may add.

dafsjr: Ok, thanks, I will do this modification

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,282 @@ func TestContainerLogging(t *testing.T) {
...
+		for _, pod := range podList.Items {
+			readCloser, err := client.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
+				Container: "logs",
+				Follow:    false,
+			}).Stream(context.TODO())
+			if err != nil {
+				t.Errorf("failed to read logs from pod %s: %v", pod.Name, err)
+				continue
+			}
+			data, _ := ioutil.ReadAll(readCloser)
```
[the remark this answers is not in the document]
dafsjr: done

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,282 @@ func TestContainerLogging(t *testing.T) {
...
```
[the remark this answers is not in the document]
dafsjr: done

pkg/operator/controller/ingress/rsyslog_configmap.go
```diff
@@ -18,9 +18,13 @@ import (
 
 const (
 	// rsyslogConfiguration is the contents for rsyslog.conf.
+	// the parameter "$MaxMessageSize 10k" is needed to allow messages with
+	// 8192 characters, the default value allows messages with maximum 8096
+	// characters
```
candita, line 23: nit
```suggestion
	// The parameter "$MaxMessageSize 10k" is needed to allow messages with
	// 8192 characters, because the default value only allows messages with maximum 
	// 8096 characters.
```

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,290 @@ func TestContainerLogging(t *testing.T) {
...
+					t.Logf("found log message for maxLength and the length is equals to 8192 chars in pod %s logs: %s", pod.Name, line)
+					found = true
+					break
+				}
+			}
+			if err := readCloser.Close(); err != nil {
+				t.Errorf("failed to close logs reader for pod %s: %v", pod.Name, err)
+			}
+			if !found {
+				t.Logf("failed to find output:\n\n%s", string(data))
```
candita, line 2273: nit: Since the data can be very long, it is better not to print it out as it will fill up the test logs.
```suggestion
				t.Logf("failed to find log message for maxLength test")
```

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,290 @@ func TestContainerLogging(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+	client, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		t.Fatalf("failed to create kube client: %v", err)
+	}
+
+	err = wait.PollImmediate(1*time.Second, 3*time.Minute, func() (bool, error) {
+		for _, pod := range podList.Items {
```
candita, line 2240: We only expect one item in the podList so it doesn't make sense to iterate and then call `continue` if there is a failure.  Remove the `for` loop and whenever there is an error, you need to return `false, nil` if it's okay to try again, or `false, err` if it's not okay to try again.  Instead of using `continue` when there is an error.

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,290 @@ func TestContainerLogging(t *testing.T) {
...
```
candita, line 2382: Same comment as above to remove the `for` loop and return the error instead of using `continue`.

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,290 @@ func TestContainerLogging(t *testing.T) {
...
+					t.Logf("found log message for minLength and the length is equals to 480 chars in pod %s logs: %s", pod.Name, line)
+					found = true
+					break
+				}
+			}
+			if err := readCloser.Close(); err != nil {
+				t.Errorf("failed to close logs reader for pod %s: %v", pod.Name, err)
+			}
+			if !found {
+				t.Logf("failed to find output:\n\n%s", string(data))
```
candita, line 2415: nit : same as above, we don't need to see the data here

pkg/operator/controller/ingress/rsyslog_configmap.go
dafsjr: done

test/e2e/operator_test.go
dafsjr: done

test/e2e/operator_test.go
dafsjr: done

test/e2e/operator_test.go
dafsjr: I think it is done

manifests/00-custom-resource-definition.yaml
```diff
@@ -351,7 +359,7 @@ spec:
                                   type: string
                                 maxLength:
                                   default: 1024
-                                  description: "maxLength is the maximum length of the syslog message \n If this field is empty, the maxLength is set to \"1024\"."
+                                  description: "maxLength is the maximum length of the log message. \n Valid values are integers in the range 480 to 4096, inclusive. \n When omitted, the default value is 1024."
```
candita, line 362: @dafsjr why is the maximum different for syslog?

dafsjr: this file is the same of vendor/github.com/openshift/api/operator/v1/0000_50_ingress-operator_00-ingresscontroller.crd.yaml, it is autogenerated/updated, Miciah asked me to modify the message for syslog too in the PR for openshift/api.

candita: But don't you want the same numeric maximum for syslog and sidecar log?

dafsjr: sorry, I understood a total different thing, I will reply in slack

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,286 @@ func TestContainerLogging(t *testing.T) {
...
+
+	err = wait.PollImmediate(1*time.Second, 3*time.Minute, func() (bool, error) {
+		pod := podList.Items[0]
+		readCloser, err := client.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{
+			Container: "logs",
+			Follow:    false,
+		}).Stream(context.TODO())
+		if err != nil {
+			t.Errorf("failed to read logs from pod %s: %v", pod.Name, err)
+			return false, err
```
candita, line 2247: I think it would be ok to `return false, nil` here, which would indicate we want to try again until the timer expires.  @Miciah or @frobware can you verify?

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,286 @@ func TestContainerLogging(t *testing.T) {
...
+		}
+		data, err := ioutil.ReadAll(readCloser)
+		if err != nil {
+			t.Errorf("failed to read logs from pod %s: %v", pod.Name, err)
+			return false, err
```
candita, line 2252: Same as above, if I'm correct.
```suggestion
			return false, nil
```

test/e2e/operator_test.go
Miciah: Yes, please log the error with `t.Logf` instead of `t.Errorf`, and then return `false, nil` so that the polling loop retries after logging the error.

manifests/00-custom-resource-definition.yaml
candita: I checked with @Miciah and this should be fine.

test/e2e/operator_test.go
```diff
@@ -2139,6 +2139,286 @@ func TestContainerLogging(t *testing.T) {
...
+			// set a fixed value for the test
+			if strings.Contains(line, "0480abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@@@@@@@=") && (length >= 450 && length <= 650) {
+				t.Logf("found log message for minLength and the length is equals to 480 chars in pod %s logs", pod.Name)
+				found = true
+				break
+			}
+		}
+		if err := readCloser.Close(); err != nil {
+			t.Errorf("failed to close logs reader for pod %s: %v", pod.Name, err)
+		}
```
candita, line 2411: As mentioned, please remove this and put in a defer function before starting `scanner.Scan`, here and in the other function you added.

PR #901, opened 2023-03-22
OCPBUGS-10714: gatewayclass: Update for OSSM 2.4 API change

Miciah:
Update for an API change in ServiceMeshControlPlane in OSSM 2.4.

https://github.com/maistra/istio-operator/commit/ecd25d550dba4ee4d6477fda4c7889329b4ad1d6 replaced the `spec.techPreview.controlPlaneMode` field with a new `spec.mode` field.  Specifying `spec.techPreview.controlPlaneMode` now elicits the following error message:

    failed to create ServiceMeshControlPlane openshift-ingress/openshift-gateway: admission webhook "smcp.validation.maistra.io" denied the request: the spec.techPreview.controlPlaneMode field is not supported in version 2.4+; use spec.mode

To resolve this issue, this PR bumps the vendored Maistra API version and changes the gatewayclass controller to use the new `spec.mode` field in the ServiceMeshControlPlane CR.

* `go.mod`: Bump `github.com/maistra/istio-operator` to `v0.0.0-20230322122339-793794762e67`.
* `go.sum`:
* `vendor/*`: Regenerate.
* `pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go` (`desiredServiceMeshControlPlane`): Update to use the new `spec.mode` field instead of `spec.techPreview.controlPlanemode`.

PR #904, opened 2023-04-04
OCPBUGS-10846: Fix TestClientTLS flakes

rfredette:
Wait for old router pods to be cleaned up before testing new mTLS config.

After updating the ingresscontroller configuration, the router deployment reports when all new pods are ready, but sometimes the pods from the older generation still haven't terminated. If those older generation pods are still marked ready when `TestClientTLS` curls an endpoint, sometimes the connections are handled by the older generation router, and that test case will fail.

This PR makes the test wait until the older generation pod(s) are completely terminated before running curl, ensuring that only the correct router pods are used.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -3581,17 +3581,15 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
 
 // waitForDeploymentCompleteWithCleanup waits for a deployment to complete its rollout, then waits for the old
 // generation's pods to finish terminating.
-func waitForDeploymentCompleteWithCleanup(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
+func waitForDeploymentCompleteWithOldPodTermination(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
 	t.Helper()
 	deployment := &appsv1.Deployment{}
 	if err := cl.Get(context.TODO(), deploymentName, deployment); err != nil {
 		return fmt.Errorf("failed to get deployment %s: %w", deploymentName.Name, err)
 	}
 
-	if deployment.Generation != deployment.Status.ObservedGeneration {
-		if err := waitForDeploymentComplete(t, cl, deployment, timeout); err != nil {
-			return fmt.Errorf("timed out waiting for deployment %s to complete rollout: %w", deploymentName.Name, err)
-		}
+	if err := waitForDeploymentComplete(t, cl, deployment, timeout); err != nil {
+		return fmt.Errorf("timed out waiting for deployment %s to complete rollout: %w", deploymentName.Name, err)
 	}
 
 	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
@@ -3599,13 +3597,26 @@ func waitForDeploymentCompleteWithCleanup(t *testing.T, cl client.Client, deploy
 		return fmt.Errorf("deployment %s has invalid spec.selector: %w", deploymentName.Name, err)
 	}
 
+	expectedReplicas := 1
+	if deployment.Spec.Replicas != nil && int(*deployment.Spec.Replicas) != 0 {
+		expectedReplicas = int(*deployment.Spec.Replicas)
+	}
+
 	return wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
-		pods := &corev1.PodList{}
-		if err := cl.List(context.TODO(), pods, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		podList := &corev1.PodList{}
+		if err := cl.List(context.TODO(), podList, client.MatchingLabelsSelector{Selector: selector}); err != nil {
 			t.Logf("failed to list pods for deployment %q: %v", deploymentName.Name, err)
 			return false, nil
 		}
-		return len(pods.Items) == int(*deployment.Spec.Replicas), nil
+		readyPods := 0
+		for _, pod := range podList.Items {
+			for _, condition := range pod.Status.Conditions {
+				if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+					readyPods++
+				}
+			}
+		}
+		return readyPods == expectedReplicas, nil
 	})
 }
 
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -3579,6 +3579,36 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
 	return err
 }
 
+// waitForDeploymentCompleteWithCleanup waits for a deployment to complete its rollout, then waits for the old
+// generation's pods to finish terminating.
+func waitForDeploymentCompleteWithCleanup(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
+	t.Helper()
+	deployment := &appsv1.Deployment{}
+	if err := cl.Get(context.TODO(), deploymentName, deployment); err != nil {
+		return fmt.Errorf("failed to get deployment %s: %w", deploymentName.Name, err)
+	}
+
+	if deployment.Generation != deployment.Status.ObservedGeneration {
+		if err := waitForDeploymentComplete(t, cl, deployment, timeout); err != nil {
+			return fmt.Errorf("timed out waiting for deployment %s to complete rollout: %w", deploymentName.Name, err)
+		}
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("deployment %s has invalid spec.selector: %w", deploymentName.Name, err)
+	}
+
+	return wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
+		pods := &corev1.PodList{}
+		if err := cl.List(context.TODO(), pods, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+			t.Logf("failed to list pods for deployment %q: %v", deploymentName.Name, err)
+			return false, nil
+		}
+		return len(pods.Items) == int(*deployment.Spec.Replicas), nil
+	})
+}
+
 func clusterOperatorConditionMap(conditions ...configv1.ClusterOperatorStatusCondition) map[string]string {
 	conds := map[string]string{}
 	for _, cond := range conditions {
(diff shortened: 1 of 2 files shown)


rfredette:
It appears that terminating pods are not counted in status.replicas, although I didn't observe that field directly. Before making this change, I ran the test and had a second terminal open running `oc get pods -w -n openshift-ingress`, and noticed there was still an old openshift-router pod that was terminating but ready when I saw the output from curl resume at line 377 (https://github.com/openshift/cluster-ingress-operator/pull/904/files#diff-29d1d3601438a24c0a4a49463c65a350beed9bc9f0a3fad47734b193fb7d0c1bR377)

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,36 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
...
+		return fmt.Errorf("deployment %s has invalid spec.selector: %w", deploymentName.Name, err)
+	}
+
+	return wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
+		pods := &corev1.PodList{}
+		if err := cl.List(context.TODO(), pods, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+			t.Logf("failed to list pods for deployment %q: %v", deploymentName.Name, err)
+			return false, nil
+		}
+		return len(pods.Items) == int(*deployment.Spec.Replicas), nil
```
frobware, line 3608: Is deployment.Spec.Replicas always non-nil?

```suggestion
return len(pods.Items) == int(pointer.Int32Deref(deployment.Spec.Replicas, -1)), nil
```

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,36 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
 	return err
 }
 
+// waitForDeploymentCompleteWithCleanup waits for a deployment to complete its rollout, then waits for the old
+// generation's pods to finish terminating.
+func waitForDeploymentCompleteWithCleanup(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
```
frobware, line 3584: The "Cleanup" part was not immediately obvious for me. How about `waitForDeploymentRolloutAndOldPodsTermination`?

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,36 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
...
+		}
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("deployment %s has invalid spec.selector: %w", deploymentName.Name, err)
+	}
+
+	return wait.PollImmediate(2*time.Second, timeout, func() (bool, error) {
+		pods := &corev1.PodList{}
```
frobware, line 3603: s/pods/pod.

frobware (changes requested):

test/e2e/operator_test.go
rfredette: `pods` is a list of possibly multiple pods, so I think plural makes more sense. That said, if it's unclear as is, I do think `podList` as a name would work.

gcs278:
I'll take a look at this too, since it seems we are blocking on this on PRs such as https://github.com/openshift/cluster-ingress-operator/pull/901
/assign

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,36 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
...
+// waitForDeploymentCompleteWithCleanup waits for a deployment to complete its rollout, then waits for the old
+// generation's pods to finish terminating.
+func waitForDeploymentCompleteWithCleanup(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
+	t.Helper()
+	deployment := &appsv1.Deployment{}
+	if err := cl.Get(context.TODO(), deploymentName, deployment); err != nil {
+		return fmt.Errorf("failed to get deployment %s: %w", deploymentName.Name, err)
+	}
+
+	if deployment.Generation != deployment.Status.ObservedGeneration {
```
gcs278, line 3591: Would it be cleaner/safer to skip this check and always do `waitForDeploymentComplete`? It does this check inside of it, along with the other replica count checks?

test/e2e/operator_test.go
gcs278, line 3608: Isn't this the same as doing`spec.replicas == status.replicas`, which is what `waitForDeploymentComplete` already does? You are counting the pods manually, but I don't see how this is filtering out terminating pods.

Shouldn't you check for no `metadata.deletionTimestamp` on each of the pods before comparing the pod count to `spec.replicas`?

gcs278: I'm not so sure.  `deployment` may always be non-nil, but `deployment.Spec.Replicas` could very well be nil. Am I missing something? I think it needs to be initialized like in `waitForDeploymentComplete` https://github.com/openshift/cluster-ingress-operator/blob/6660fb0f4fc7eac1f3e8153d1d239c36d063c020/test/e2e/operator_test.go#L3536

gcs278: Alternatively, a better solution might be instead to ensure there are no terminating pods that are also ready. This saves about 15 seconds of blocking when there is 1 terminating router that is not ready (it's not going to get requests), and we have our two other router pods up.

gcs278 (commented):
I experimented with rolling out a new router pod, and I can very much confirm what @rfredette is seeing: `status.replicas` counts terminating replicas and the terminating replicas can still be ready for a solid moment.

I provided an alternative solution if you'd like to consider it, but I could be missing something.

test/e2e/operator_test.go
rfredette: I think the reason we're seeing this issue is that `status.replicas` seems to either be pruning pods that are terminating or it's only including pods in the current generation. Either way, I've found that `spec.replicas == status.replicas` doesn't say if there are still terminating pods, while manually listing the pods does.

I do like the idea of checking for pods that are terminating but still ready. I'll try that and if it works, I'll add it to the PR.

rfredette: It turns out that "terminating" isn't actually a pod state; oc/kubectl will show that, but it's not a part of the pod status.

However, I did update the function to only count the number of ready pods, so it should return once the old pods become not ready.

rfredette:
aws and gcp operator suites failed, but TestClientTLS passed. Other suites also failed, but since this PR only touches the operator suite, those don't seem related.

/retest

rfredette:
More test flakes in unrelated tests
/retest

gcs278:
Not related.
/retest-required

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,47 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
...
+		return fmt.Errorf("timed out waiting for deployment %s to complete rollout: %w", deploymentName.Name, err)
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("deployment %s has invalid spec.selector: %w", deploymentName.Name, err)
+	}
+
+	expectedReplicas := 1
+	if deployment.Spec.Replicas != nil && int(*deployment.Spec.Replicas) != 0 {
```
gcs278, line 3601: The `int(*deployment.Spec.Replicas) != 0` seems a little funny to me. `replicas: 0` is indeed an allowed value in an deployment and if you did have a deployment with it, this function would fail since it overrides it to 1.

I know it's a highly unlikely use-case. Did you have another reason for ignoring `replicas: 0` and overriding it to 1?

ShudiLi:
/label qe-approved
thanks

test/e2e/operator_test.go
frobware: We could change the Spec.Replicas deref to:

```go
return len(pods.Items) == int(pointer.Int32Deref(deployment.Spec.Replicas, -1)), nil
```

test/e2e/operator_test.go
frobware: In my head I read this as a `podList`, but `Items` brings the plurality.

test/e2e/operator_test.go
```diff
@@ -3579,6 +3579,47 @@ func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.
 	return err
 }
 
+// waitForDeploymentCompleteWithCleanup waits for a deployment to complete its rollout, then waits for the old
+// generation's pods to finish terminating.
+func waitForDeploymentCompleteWithOldPodTermination(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
```
Miciah, line 3584:
```suggestion
// waitForDeploymentCompleteWithOldPodTermination waits for a deployment to complete its rollout, then waits for the old
// generation's pods to finish terminating.
func waitForDeploymentCompleteWithOldPodTermination(t *testing.T, cl client.Client, deploymentName types.NamespacedName, timeout time.Duration) error {
```

Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/pull/904/commits/39c6aef16c999bc4e05e948238454439e8c21b3d.

test/e2e/operator_test.go
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/pull/904/commits/39c6aef16c999bc4e05e948238454439e8c21b3d.

test/e2e/operator_test.go
Miciah: https://github.com/openshift/cluster-ingress-operator/pull/904/commits/39c6aef16c999bc4e05e948238454439e8c21b3d changes this to use `Int32Deref`.  I kept the explicit `expectedReplicas` variable so that we aren't calling `Int32Deref` repeatedly; I hope that's fine.

gcs278:
Ship it.
/lgtm

gcs278:
/test e2e-aws-operator
`TestUnmanagedDNSToManagedDNSInternalIngressController` failure

gcs278:
`TestInternalLoadBalancer` failure. I haven't seen that before, but this change only affects `TestClientTLS`

/test e2e-gcp-operator

gcs278:
`TestUnmanagedDNSToManagedDNSInternalIngressController` failure...
/test e2e-gcp-operator

Miciah:
The test is also failing for 4.13 and 4.12 CI.  
/cherry-pick release-4.13

PR #908, opened 2023-04-11
CCO-318: Read feature gates for future usage

deads2k:
This is needed for the AzureWorkloadIdentity work (see https://github.com/openshift/cluster-ingress-operator/pull/906), so adding the core featuregate access here.

When we eventually read featuregates from the cluster itself, this golang construct will remain the same.

/cc @abutcher @jstuever 
/assign @Miciah


Miciah:
/retitle CCO-318: Read feature gates for future usage

manifests/00-cluster-role.yaml
```diff
@@ -101,8 +101,11 @@ rules:
   - featuregates
   - apiservers
   - networks
+  - clusterversions
   verbs:
   - get
+  - list
+  - watch
```
Miciah, line 108: I suppose this permission would be needed in the future for versioned featuregates, but why is it needed today?

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -84,16 +79,7 @@ type reconciler struct {
...
-	if err := r.cache.Get(ctx, request.NamespacedName, &fg); err != nil {
-		if apierrors.IsNotFound(err) {
-			log.Info("featuregate not found; reconciliation will be skipped", "request", request)
-			return reconcile.Result{}, nil
-		}
-		return reconcile.Result{}, err
-	}
-
-	if !featureIsEnabled(featureGateName, &fg) {
+	if !r.config.GatewayAPIEnabled {
```
Miciah, line 82: With this change, there isn't really a good reason for this controller to watch featuregates at all.  However, it does make sense to have a control loop to retry creating a CRD when a create fails.  In addition, we really do want to have the controller but have it reconcile customresourcedefinitions if they change (see <https://issues.redhat.com/browse/NE-1273>), so let's leave it as a followup to modify the controller to watch and reconcile customresourcedefinitions.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -114,40 +100,3 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-//
-//   - First, if the featuregate's spec.featureGateSelection.featureSet field is
-//     set to "CustomNoUpgrade", then the feature is enabled if, and only if, it
-//     is specified in spec.featureGateSelection.customNoUpgrade.enabled.
-//
-//   - Second, if spec.featureGateSelection.featureSet is set to a value that
-//     isn't defined in configv1.FeatureSets, then the feature is *not* enabled.
-//
-//   - Finally, the feature is enabled if, and only if, the feature is specified
-//     in configv1.FeatureSets[spec.featureGateSelection.featureSet].enabled.
```
Miciah, line 131: Could this comment be preserved somewhere?  It took me a bit of effort to  work out from the features API documentation and Slack conversations what the logic is for enabling a feature.

pkg/operator/operator.go
```diff
@@ -73,7 +78,60 @@ type Operator struct {
...
+
+	configClient, err := configclient.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	desiredVersion := config.OperatorReleaseVersion
+	missingVersion := statuscontroller.UnknownVersionValue
+
+	// by default, this will exit(0) the process if the featuregates ever change to a different set of values.
```
Miciah, line 110: I usually ask people to write full sentences in comments.
```suggestion
	// By default, this will exit(0) the process if the featuregates ever change to a different set of values.
```

pkg/operator/operator.go
```diff
@@ -73,7 +78,60 @@ type Operator struct {
...
+	)
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(config.Stop)
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		enabled, disabled, _ := featureGateAccessor.CurrentFeatureGates()
+		klog.Infof("FeatureGates initialized: enabled=%v  disabled=%v", enabled, disabled)
+	case <-time.After(1 * time.Minute):
+		klog.Errorf("timed out waiting for FeatureGate detection")
```
Miciah, line 124: Is there a reason to use klog?
```suggestion
		log.Infof("FeatureGates initialized", "enabled", enabled, "disabled", disabled)
	case <-time.After(1 * time.Minute):
		log.Error(nil, "timed out waiting for FeatureGate detection")
```

Miciah (commented):
Would you mind mentioning the Jira issue in the Git commit message?

manifests/00-cluster-role.yaml
deads2k: > I suppose this permission would be needed in the future for versioned featuregates, but why is it needed today?

Because the informer is started to be code compatible with the update comes.

pkg/operator/controller/gatewayapi/controller.go
deads2k: > Could this comment be preserved somewhere? It took me a bit of effort to work out from the features API documentation and Slack conversations what the logic is for enabling a feature.

Spoke on slack.  With the just-today-merged FeatureGate.Status, this comment is no longer pertinent.  We'll update in two stages so ingress isn't a guinea pig, but we won't preserve this comment.

deads2k:
comments updated to be full sentences, switch to log, updated commit message.

Miciah:
Looks like you need a `go mod tidy` to remove the addition of `k8s.io/klog/v2` to `go.mod` etc. that your latest changes obviated.

pkg/operator/controller/gatewayapi/controller.go
candita: @miciah I use `featureIsEnabled` in my E2E tests and can preserve it there.

deads2k:
> For this to be usable by #906, it will need to be updated to be compatible with a newer openshift/api to include the AzureWorkloadIdentity [commit](https://github.com/openshift/api/commit/f0e87a8c3bad757312c281f5437827d492ec41ee). As it stands now, #906 cannot use the newer api due to the [addition](https://github.com/openshift/api/commit/f8202f4581421c8eb8473fe266c3aacd9bc46a69) of the featureGateName type. As a result, the changes in this PR are no longer compatible.

updated.  The kcm-o and kas-o have already switched.  The code remained the same, but the library-go changed underneath it.

pkg/operator/operator.go
```diff
@@ -73,7 +77,61 @@ type Operator struct {
...
+		return nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+
+	enabled, _, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return nil, err
+	}
+	// example of future featuregate read and usage to set a variable to pass to a controller
+	_ = sets.New[configv1.FeatureGateName](enabled...).Has("AzureWorkloadIdentity")
+	gatewayAPIEnabled := sets.New[configv1.FeatureGateName](enabled...).Has("GatewayAPI")
```
Miciah, line 133: Ooh, [generics](https://go.dev/blog/intro-generics)!

deads2k:
requires https://github.com/openshift/cluster-config-operator/pull/301

deads2k:
infra Error: creating EC2 Instance: InsufficientInstanceCapacity: We currently do not have sufficient m6a.xlarge capacity in t

/retest

deads2k:
related to the external CCM flow, so labeling so it'll merge.

PR #915, opened 2023-04-27
OCPBUGS-12579: Address CVE-2022-41723

gcs278:
Address CVE-2022-41723 via

```
go get golang.org/x/net@v0.7.0
go mod tidy
go mod vendor
```


Miciah:
You also need to run `go mod tidy && go mod vendor`.

frobware:
> You also need to run `go mod tidy && go mod vendor`.

The order should really be:
- go mod vendor
- go mod tidy

We have this ordering encoded in [hack/verify-deps.sh](https://github.com/openshift/cluster-ingress-operator/blob/master/hack/verify-deps.sh).

frobware:
332 tests failed.

/test e2e-aws-ovn-single-node

frobware:
> > You also need to run `go mod tidy && go mod vendor`.
> 
> The order should really be:
> 
> * go mod vendor
> * go mod tidy
> 
> We have this ordering encoded in [hack/verify-deps.sh](https://github.com/openshift/cluster-ingress-operator/blob/master/hack/verify-deps.sh).

So reading around a little it seems the preferred order is tidy, then vendor. In which case we should update our `./hacks/verify-deps.sh` script. I have relied on the wisdom in our script for too long. Opinions sought. 

(I distinctly recall getting caught out when we switched to Go 1.18, but that could have been because of the existing ordering we had in `./hacks/verify-deps.sh`.)

PR #925, opened 2023-05-04
OCPBUGS-5478: add UBI based Dockerfile

alebedev87:
`Dockerfile.ubi` added to enable the builds for the community.


Dockerfile.ubi
```diff
@@ -0,0 +1,13 @@
...
+RUN make build
+
+FROM registry.access.redhat.com/ubi8/ubi-micro:latest
+COPY --from=builder /ingress-operator/ingress-operator /usr/bin/
+COPY manifests /manifests
+ENTRYPOINT ["/usr/bin/ingress-operator"]
+LABEL io.openshift.release.operator="true"
+LABEL io.k8s.display-name="OpenShift ingress-operator" \
+      io.k8s.description="This is a component of OpenShift Container Platform and manages the lifecycle of ingress controller components." \
+      maintainer="SukaDan Mace <dmace@redhat.com>"
```
Miciah, line 13: "SukaDan Mace"?

hack/release-local.sh
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 
 REPO="${REPO:-}"
 MANIFESTS="${MANIFESTS:-}"
+DOCKERFILE="${DOCKERFILE:-Dockerfile}"
```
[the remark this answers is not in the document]
alebedev87: Done.

hack/release-local.sh
```diff
@@ -19,10 +20,10 @@ REV=$(git rev-parse --short HEAD)
 TAG="${TAG:-$REV}"
 
 if [[ -z "${DOCKER+1}" ]] && command -v buildah >& /dev/null; then
-  buildah bud -t $REPO:$TAG .
+  buildah bud -t $REPO:$TAG -f "${DOCKERFILE}" .
   buildah push $REPO:$TAG docker://$REPO:$TAG
 else
-  docker build -t $REPO:$TAG .
+  docker build -t $REPO:$TAG -f "${DOCKERFILE}" .
```
rfredette, line 26: I think we want to prefer podman over docker when possible. Since you're changing this, would this be a good opportunity to add an `elif` block to use podman if it's present?

rfredette:
/approve

Overall looks good. I had one nit-pick; once that's been addressed or rejected, I'll tag this lgtm as well.

hack/release-local.sh
frobware: > I think we want to prefer podman over docker when possible. Since you're changing this, would this be a good opportunity to add an `elif` block to use podman if it's present?

Or add: 

```
: "${PODMAN:=podman}"
```

and you can override PODMAN on invocation. By default we'd just use podman though.
You'd obviously have to replace all the plain invocations of `podman` with `$PODMAN`.

frobware: > I thought of making a smooth change without any impact on the potential users of the script

That's a fair point. If don't want to perturb what's already there we could invert my change to DOCKER=docker which would allow the status quo but also make it easy if you only have podman installed.

rfredette: This wasn't clear from my original comment (sorry :pensive:), but I was thinking that we would prefer podman over docker, but leave buildah as the default. I know podman is an acceptable substitute to docker since that's what I use locally ("docker" on my machine is a symlink to podman), but I don't know if there's anything we'd be losing by preferring it over buildah.

alebedev87: Done. The order is:`buildah`, `podman`, `docker`.

PR #930, opened 2023-05-11
OCPBUGS-9464: Verify CRLs are updated correctly when they expire

rfredette:
This PR contains one new test with multiple test cases for the fix in openshift/router#472. 

The new test creates an ingress controller with mTLS enabled and client CA certificates that include CRLs. It verifies that the CRLs that are downloaded on startup match the CRLs specified in the CA bundle, and waits until all CRLs have expired once, making sure that each update only updates the required CRLs, and that all updates download the correct CRLs.

PR #937, opened 2023-05-24
Stop propagating proxy env to router pod

enxebre:
This was merged in https://github.com/openshift/cluster-ingress-operator/pull/930. In hypershift the operator pod runs in a different cluster than the router pod, and so they might have different proxy configs. We set ootb proxy settings for the operator to go through konnectivity. I think because this got introduced in this PR https://github.com/openshift/cluster-ingress-operator/pull/930/files\#diff-03760f03eeac1174e38e1ec5fecadcab4cb38d05e00509f201d7ae64dc1c7c36R1044-R1048 now those proxy settings are being propagated down to the router pod and so breaking hypershift.

See an example on how we diferenciate in the cluster network operator https://github.com/openshift/hypershift/blob/f03113aec2545aa08f85ab0e2fa0a790f709aacb/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go\#L341-L347

https://github.com/openshift/cluster-network-operator/blob/7df32ec5213e2abfcff18cd22a63e7f20a5cccdc/pkg/network/cloud_network.go\#L98

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -17,8 +17,6 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"github.com/operator-framework/operator-lib/proxy"
-
 	operatorv1 "github.com/openshift/api/operator/v1"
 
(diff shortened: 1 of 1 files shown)


dgoodwin:
Pursuing a full revert in https://github.com/openshift/cluster-ingress-operator/pull/938

/close

PR #946, opened 2023-06-07
OCPBUGS-14396: Set controller-runtime logger to a null logger for E2E

gcs278:
Due to controller-runtime v0.15.0 bump and changes made in https://github.com/kubernetes-sigs/controller-runtime/pull/2317, you must set a logger for controller-runtime and client.go was not initializing the logger.

`test/e2e/operator_test.go`: Initialize controller-runtime logger in init() function

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -38,8 +38,11 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 
+	"github.com/go-logr/logr"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -113,6 +116,13 @@ var (
 	defaultName       = types.NamespacedName{Namespace: operatorNamespace, Name: manifests.DefaultIngressControllerName}
 	clusterConfigName = types.NamespacedName{Namespace: operatorNamespace, Name: manifests.ClusterIngressConfigName}
 )
+
+func init() {
+	// This is required because controller-runtime expects its consumers to
+	// set a logger through log.SetLogger within 30 seconds of the program's
+	// initalization.
+	ctrlruntimelog.SetLogger(logr.New(ctrlruntimelog.NullLogSink{}))
+}
 
 func TestMain(m *testing.M) {
 	if os.Getenv("E2E_TEST_MAIN_SKIP_SETUP") == "1" {


alebedev87:
/lgtm

To test the approver's lgtm.

gcs278:
E2E test [logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/946/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1666548851550457856/build-log.txt) don't print backtrace error and ingress-operator [logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/946/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1666548851550457856/artifacts/e2e-aws-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-549568c844-nxvg2_ingress-operator.log) seem unaffected, as they should be.

pkg/operator/client/client.go
```diff
@@ -64,6 +67,11 @@ func init() {
 	if err := maistrav2.SchemeBuilder.AddToScheme(scheme); err != nil {
 		panic(err)
 	}
+
+	// This is required because controller-runtime expects its consumers to
+	// set a logger through log.SetLogger within 30 seconds of the program's
+	// initalization.
+	ctrlruntimelog.SetLogger(logr.New(ctrlruntimelog.NullLogSink{}))
```
alebedev87, line 74: `pkg/operator/client` package is used in the operator and e2e test binaries. The e2e binary was missing the logger so this change makes the required migration for it. However the operator binary already sets the controller runtime logger [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/operator.go#L62) so this change doesn't cause any conflict (because client's init runs first) but is redundant.
The alternative would be to add `SetLogger` somewhere in `test/e2e`.

gcs278: I don't see where `pkg/operator/client` is used by the operator, hence it seemed safe putting here. Am I misunderstanding the usage of `pkg/operator/client`?

alebedev87: `pkg/operator/client` is imported in [pkg/operator/operator.go](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/operator.go#L24), that is, its `init` function was invoked.

gcs278: Oh whoops, you are right. I didn't give Goland enough time to finish loading all of the references...

Yea, I see, it's not-problematic, but redundant. I'm going to find a different place for it in `test/e2e`.

test/e2e/all_test.go
```diff
@@ -16,6 +18,11 @@ import "testing"
 // ingressconfig object and the defer logic for cleanup is still
 // runnng when the new test starts.
 func TestAll(t *testing.T) {
+	// This is required because controller-runtime expects its consumers to
+	// set a logger through log.SetLogger within 30 seconds of the program's
+	// initalization.
+	ctrlruntimelog.SetLogger(logr.New(ctrlruntimelog.NullLogSink{}))
```
alebedev87, line 24: I think that the better place for this is `operator_test.go`, in a `init()` function. Because here the logger will be set at the test runtime. Some setup may take longer (`TestMain`) or some other test can be run before `TestAll` (I believe it's alphabetically ordered but not guaranteed).

gcs278: Yea that's fair, I didn't realize `TestAll` was really just a test itself, it wasn't as special as I thought. I also goofed up the imports.

Good suggestion, seems reasonable. I put it in an `init()` function in `operator_test.go`.

alebedev87:
LGTM, just the title and the description need an update.

gcs278:
Lots of E2E failures, but none look related. I confirmed we don't get the SetLogger error in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/946/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1674059716902260736/build-log.txt

/retest

PR #948, opened 2023-06-15
OCPBUGS-14994: Don't add clientca-configmap finalizer if deleting

Miciah:
Don't add the clientca-configmap finalizer to an IngressController if it is marked for deletion.

Before this change, the clientca-configmap controller would attempt to add (or re-add) its finalizer to an IngressController that had been marked for deletion.  This attempt resulted in the following error from the API:

    IngressController.operator.openshift.io "test-client-ca-configmap" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{"ingresscontroller.operator.openshift.io/finalizer-clientca-configmap"}

The controller would retry adding the finalizer repeatedly, causing the same error to be repeatedly logged until the deletion completed.

* `pkg/operator/controller/clientca-configmap/controller.go` (`Reconcile`): Only attempt to add the finalizer if the IngressController is not marked for deletion.

### DIFF
--- a/pkg/operator/controller/clientca-configmap/controller.go
+++ b/pkg/operator/controller/clientca-configmap/controller.go
@@ -31,7 +31,9 @@ const (
 var log = logf.Logger.WithName(controllerName)
 
 // New creates a new controller that syncs client CA configmaps between the
-// config and operand namespaces.
+// config and operand namespaces.  This controller also adds a finalizer to the
+// IngressController so that the controller can delete the configmap from the
+// operand namespace when the IngressController is marked for deletion.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	operatorCache := mgr.GetCache()
 	reconciler := &reconciler{
@@ -191,7 +193,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 
 	const finalizer = "ingresscontroller.operator.openshift.io/finalizer-clientca-configmap"
-	if len(ic.Spec.ClientTLS.ClientCA.Name) != 0 && !slice.ContainsString(ic.Finalizers, finalizer) {
+	if len(ic.Spec.ClientTLS.ClientCA.Name) != 0 && ic.DeletionTimestamp == nil && !slice.ContainsString(ic.Finalizers, finalizer) {
 		// Ensure the ingresscontroller has a finalizer so we get a
 		// chance to delete the configmap when the ingresscontroller is
 		// deleted.
--- a/pkg/operator/controller/clientca-configmap/controller.go
+++ b/pkg/operator/controller/clientca-configmap/controller.go
@@ -167,9 +167,8 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 
 // Config holds all the things necessary for the controller to run.
 type Config struct {
-	OperatorNamespace string
-	SourceNamespace   string
-	TargetNamespace   string
+	SourceNamespace string
+	TargetNamespace string
 }
 
 type reconciler struct {
--- a/pkg/operator/controller/clientca-configmap/controller_test.go
+++ b/pkg/operator/controller/clientca-configmap/controller_test.go
@@ -1,0 +1,320 @@
+package clientcaconfigmap
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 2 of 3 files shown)


pkg/operator/controller/clientca-configmap/controller.go
```diff
@@ -191,7 +191,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 
 	const finalizer = "ingresscontroller.operator.openshift.io/finalizer-clientca-configmap"
-	if len(ic.Spec.ClientTLS.ClientCA.Name) != 0 && !slice.ContainsString(ic.Finalizers, finalizer) {
+	if len(ic.Spec.ClientTLS.ClientCA.Name) != 0 && ic.DeletionTimestamp == nil && !slice.ContainsString(ic.Finalizers, finalizer) {
```
candita, line 194: Could you add a unit test for this change?

Miciah: https://github.com/openshift/cluster-ingress-operator/compare/d6ec5434cf35d6046264ceb5d652a5cc22cae4f3..b3ea6755c73f9e0a0a4980ae2deb838e072fda95 adds a unit test.

pkg/operator/controller/clientca-configmap/controller_test.go
```diff
@@ -0,0 +1,304 @@
...
+			ts = &metav1.Time{Time: exampleTime}
+		}
+		return &operatorv1.IngressController{
+			ObjectMeta: metav1.ObjectMeta{
+				DeletionTimestamp: ts,
+				Finalizers:        append([]string{manifests.IngressControllerFinalizer}, additionalFinalizers...),
+				Namespace:         "openshift-ingress-operator",
+				Name:              "test",
+			},
+			Spec: operatorv1.IngressControllerSpec{
```
candita, line 47: Why is the configmap being examined in these tests?

Miciah: The purpose of this controller is to copy the user-provided configmap, which is specified in the IngressController, from the "openshift-config" namespace into the "openshift-ingress" namespace, so the test is written to give the controller an IngressController that specifies a configmap and verify that the controller copies the configmap if it exists.  

Did I understand your question correctly and answer it?  Should I add a code comment with the above explanation?

pkg/operator/controller/clientca-configmap/controller_test.go
```diff
@@ -0,0 +1,304 @@
...
+			name: "no configmap specified",
+			existingObjects: []runtime.Object{
+				ic(false, ""),
+			},
+			expectCreate: []client.Object{},
+			expectUpdate: []client.Object{},
+			expectDelete: []client.Object{},
+		},
+		{
+			name: "source configmap absent, target configmap absent, finalizer absent",
```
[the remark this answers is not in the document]
Miciah: In this test case, the IngressController exists and specifies the "ca-bundle" configmap, but the configmap doesn't exist, so the controller can add the finalizer but cannot copy the configmap.

pkg/operator/controller/clientca-configmap/controller_test.go
```diff
@@ -0,0 +1,304 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			fakeClient := fake.NewClientBuilder().
+				WithScheme(scheme).
+				WithRuntimeObjects(tc.existingObjects...).
+				Build()
+			cl := &fakeClientRecorder{fakeClient, t, []client.Object{}, []client.Object{}, []client.Object{}}
+			reconciler := &reconciler{
+				client: cl,
+				config: Config{
+					OperatorNamespace: "openshift-ingress-operator",
```
frobware, line 248: As far as the tests are concerned, it seems we could drop setting the `OperatorNamespace`.

frobware (commented):
LGTM, just a comment regarding the necessity of the OperatorNamespace in Test_Reconcile.

pkg/operator/controller/clientca-configmap/controller_test.go
Miciah: I pushed https://github.com/openshift/cluster-ingress-operator/pull/948/commits/3a3fb20502db4bb46db3de1ffef89faa5f84cc6a to remove the field.

pkg/operator/controller/clientca-configmap/controller_test.go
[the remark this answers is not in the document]
candita: If it's not too much trouble, it would be helpful to specify in the test `name` what the outcome of the test is supposed to be.  So here, because there is no `cm` in the `existingObjects`, there is no `cm` in the `expectUpdate`?  And, it is ok for the configmap name to be there in the `ic` even if there is no `cm` object? And we only add a finalizer if there is a configmap name, even if there is no matching `cm` object?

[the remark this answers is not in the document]
Miciah: > If it's not too much trouble, it would be helpful to specify in the test name what the outcome of the test is supposed to be. 

Done in https://github.com/openshift/cluster-ingress-operator/compare/3a3fb20502db4bb46db3de1ffef89faa5f84cc6a..e9ceaf5969c681a0ba773d5f97bc661703dfbf17.

> So here, because there is no cm in the existingObjects, there is no cm in the expectUpdate? 

Correct.  There is no source configmap, so the controller cannot make a copy of it.

> And, it is ok for the configmap name to be there in the ic even if there is no cm object? 

It can happen, so the controller needs to handle the case.  The router pods cannot start until the configmap exists though because the router pods have a volume mount for the configmap.  

> And we only add a finalizer if there is a configmap name, even if there is no matching cm object?

Correct.

pkg/operator/controller/clientca-configmap/controller_test.go
Miciah: The controller was missing test coverage entirely, so I added a comprehensive test, including test cases for the specific issue in OCPBUGS-14994.  Would it be better if I split it into two commits: one to add the test and one to add the specific test case that is related to OCPBUGS-14994?

candita: No, this is fine.

PR #949, opened 2023-06-16
OCPBUGS-15100: Create valid DNS names for Gateway API on GCP

candita:
Before this change, in `ensureDNSRecordsForGateway` we didn't add a trailing dot to the domain before creating the dnsrecord, causing the DNS name to fail to be published on a GCP platform.

Add a trailing dot if one doesn't exist, add test to check that only one dot is added, and change expectations of other tests to check for a trailing dot in the domain name.

### DIFF
--- a/pkg/operator/controller/gateway-service-dns/controller.go
+++ b/pkg/operator/controller/gateway-service-dns/controller.go
@@ -3,6 +3,7 @@ package gateway_service_dns
 import (
 	"context"
 	"reflect"
+	"strings"
 
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
@@ -222,6 +223,10 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 	var errs []error
 	for _, domain := range domains {
 		name := operatorcontroller.GatewayDNSRecordName(gateway, domain)
+		// If domain doesn't have a trailing dot, add it
+		if !strings.HasSuffix(domain, ".") {
+			domain = domain + "."
+		}
 		_, _, err := dnsrecord.EnsureDNSRecord(r.client, name, labels, ownerRef, domain, service)
 		errs = append(errs, err)
 	}
--- a/pkg/operator/controller/gateway-service-dns/controller_test.go
+++ b/pkg/operator/controller/gateway-service-dns/controller_test.go
@@ -138,13 +138,13 @@ func Test_Reconcile(t *testing.T) {
 			},
 			reconcileRequest: req("openshift-ingress", "example-gateway"),
 			expectCreate: []client.Object{
-				dnsrecord("example-gateway-5bfc88bc87-wildcard", "*.prod.example.com", "lb.example.com"),
-				dnsrecord("example-gateway-57b76476b6-wildcard", "*.stage.example.com", "lb.example.com"),
+				dnsrecord("example-gateway-5bfc88bc87-wildcard", "*.prod.example.com.", "lb.example.com"),
+				dnsrecord("example-gateway-57b76476b6-wildcard", "*.stage.example.com.", "lb.example.com"),
 			},
 			expectUpdate: []client.Object{},
 		},
 		{
-			name: "gateway with two listeners and one dnsrecord with a stale target",
+			name: "gateway with two listeners and one dnsrecord with a stale target, hostname already has trailing dot",
 			existingObjects: []runtime.Object{
 				gw(
 					"example-gateway",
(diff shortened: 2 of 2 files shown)


candita:
> level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: platform.azure.region: Invalid value: "centralus": region "centralus" is not valid or not available for this account
> environment: line 63: ret: unbound variable

/test e2e-azure-operator

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -222,6 +223,10 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 	var errs []error
 	for _, domain := range domains {
 		name := operatorcontroller.GatewayDNSRecordName(gateway, domain)
+		// If domain doesn't have a trailing dot, add it
```
Miciah, line 226:
```suggestion
		// If domain doesn't have a trailing dot, add it.
```
I would have added the trailing dot in `desiredDNSRecord`, but maybe this makes more sense.

Miciah (commented):
The code change looks good.  Can you make sure the commit has a proper commit message?

candita:
> The code change looks good. Can you make sure the commit has a proper commit message?

Done

candita:
[OCPBUGS-12479](https://issues.redhat.com/browse/OCPBUGS-12479) exists for the azure-ovn test issue:

> level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: platform.azure.region: Invalid value: "centralus": region "centralus" is not valid or not available for this account

candita:
> {Failed  === RUN   TestNodePool/NodePool_Tests_Group/TestNodepoolMachineconfigGetsRolledout/EnsureNoCrashingPods
>     util.go:457: Container hosted-cluster-config-operator in pod hosted-cluster-config-operator-55bdff74b4-dnxtm has a restartCount > 0 (1)
>             --- FAIL: TestNodePool/NodePool_Tests_Group/TestNodepoolMachineconfigGetsRolledout/EnsureNoCrashingPods (0.08s)
> }

/test e2e-hypershift

candita:
> > > The code change looks good. Can you make sure the commit has a proper commit message?
> > 
> > 
> > Done
> 
> I still only see one line in the commit message of the most recent commit that you pushed for this PR:
> 
> ```
> % git --no-pager show -s 9341a7f19d912ba1a84b6044a222638a682d75f9
> commit 9341a7f19d912ba1a84b6044a222638a682d75f9
> Author: candita <cholman@redhat.com>
> Date:   2023-06-16 14:19:53 -0400
> 
>     OCPBUGS-15100: Fix invalid DNS name formatting for GCP
> % 
> ```
> 
> Did you mean to add more? The description of this PR is pretty descriptive, so I figured you would use that as the commit message (with minor reformatting, and with "for Gateway API" in the commit message's body as you have it in the PR's title).

I never add long commit messages like you do.  I can start doing it but may need some reminders.

Miciah:
> I never add long commit messages like you do. I can start doing it but may need some reminders.

Thanks!  I might go overboard sometimes, so I haven't been insisting on the level of detail that I usually provide, and what you have now is sufficient.  However, commit messages (not PR descriptions) are what go into the Git history, which folks rely on to understand the reason and context for a change, for example when fixing a bug or when trying to determine which feature a change was a part of, and so having clear and reasonably detailed commit messages is critical.  

For single-commit PRs, GitHub uses the commit message as the initial PR description.  I've been assuming that folks have been putting the same information in their commit messages that they put into their PR descriptions, and I often don't look at the individual commits, but I'll start being more careful when reviewing PRs.  * grin *.  

/approve
/lgtm

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -222,6 +223,10 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
+		if !strings.HasSuffix(domain, ".") {
+			domain = domain + "."
+		}
```
Miciah, line 229: You need to account for the trailing dot in `deleteStaleDNSRecordsForGateway` too, or else it immediately deletes the DNSRecord CR that `ensureDNSRecordsForGateway` created.  Better yet, add the trailing dot in `getGatewayHostnames`:
```diff
diff --git a/pkg/operator/controller/gateway-service-dns/controller.go b/pkg/operator/controller/gateway-service-dns/controller.go
index 1084d65c..919b8d32 100644
--- a/pkg/operator/controller/gateway-service-dns/controller.go
+++ b/pkg/operator/controller/gateway-service-dns/controller.go
@@ -199,7 +199,12 @@ func getGatewayHostnames(gateway *gatewayapiv1beta1.Gateway) sets.String {
 		if listener.Hostname == nil || len(*listener.Hostname) == 0 {
 			continue
 		}
-		domains.Insert(string(*listener.Hostname))
+		domain := string(*listener.Hostname)
+		// If domain doesn't have a trailing dot, add it.
+		if !strings.HasSuffix(domain, ".") {
+			domain = domain + "."
+		}
+		domains.Insert(domain)
 	}
 	return domains
 }
@@ -223,10 +228,6 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 	var errs []error
 	for _, domain := range domains {
 		name := operatorcontroller.GatewayDNSRecordName(gateway, domain)
-		// If domain doesn't have a trailing dot, add it
-		if !strings.HasSuffix(domain, ".") {
-			domain = domain + "."
-		}
 		_, _, err := dnsrecord.EnsureDNSRecord(r.client, name, labels, ownerRef, domain, service)
 		errs = append(errs, err)
 	}
```
This also has the advantage that "foo.com" without a trailing "." and and "foo.com." with a trailing "." will get the same hash.

candita: Fixed in https://github.com/openshift/cluster-ingress-operator/pull/956

PR #952, opened 2023-06-25
Fix typr

xujielong:


candita:
@xujielong Thank you for your keen eye and for this PR submission.  The typo `iff` you point out is actually a developer abbreviation for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if.)  Since this godoc is not user-facing, we will not be changing it at this point in time.

/close

PR #957, opened 2023-06-29
 OCPBUGS-15100: [release-4.13] Fix invalid DNS name formatting for GCP     

candita:
Before this change, in ensureDNSRecordsForGateway we didn't add a trailing dot to the domain
before creating the dnsrecord, causing the DNS name to fail to be published on a GCP platform.
    
Add a trailing dot if one doesn't exist, add test to check that only one dot is added, and
change expectations of other tests to check for a trailing dot in the domain name.


candita:
Mistaken attempt at manual cherry-pick.
/close

PR #961, opened 2023-07-10
OCPBUGS-9037: If mTLS is required and the canary receives a "certificate required" TLS error, consider the canary probe successful

rfredette:
Without an API to provide the canary with a valid client certificate, the canary probes are guaranteed to fail when mTLS is required by the default ingress controller. However, if the failure is determined to be because the canary is missing a client certificate, it's reasonable to assume that the ingress controller is functional.

This change checks specifically for the TLS alert "certificate required", and if the default ingress controller also has mTLS required, the probe function returns nil, indicating a successful probe.

### DIFF
--- a/pkg/operator/controller/canary/http.go
+++ b/pkg/operator/controller/canary/http.go
@@ -22,7 +22,7 @@ const (
 
 // probeRouteEndpoint probes the given route's host
 // and returns an error when applicable.
-func probeRouteEndpoint(route *routev1.Route) error {
+func (r *reconciler) probeRouteEndpoint(route *routev1.Route) error {
 	routeHost := getRouteHost(route)
 	if len(routeHost) == 0 {
 		return fmt.Errorf("route host is empty, cannot test route")
@@ -48,7 +48,7 @@ func probeRouteEndpoint(route *routev1.Route) error {
 	request = request.WithContext(ctx)
 
 	// Send the HTTP request
-	timeout, _ := time.ParseDuration("10s")
+	timeout := 10 * time.Second
 	client := &http.Client{
 		Timeout: timeout,
 		// The canary route uses edge termination and the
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/canary/http.go
```diff
@@ -68,18 +68,40 @@ func probeRouteEndpoint(route *routev1.Route) error {
...
+				// router is working as intended.
+				if opErr.Err.Error() == "tls: certificate required" {
+					if mtlsRequired, mtlsErr := r.isMTLSRequired(); mtlsErr != nil {
+						// Log the error from isMTLSRequired(), but continue the function so the actual request error is
+						// returned.
+						log.Error(mtlsErr, "Failed to verify mTLS status of default ingress controller")
+					} else if mtlsRequired {
+						// Since mTLS is required in the ingress config and our probe was rejected due to a missing
+						// certificate, consider this a successful probe.
+						return nil
```
frobware, line 100: The early return here will mean the defer block further down doesn't run.

pkg/operator/controller/canary/http.go
```diff
@@ -68,18 +68,40 @@ func probeRouteEndpoint(route *routev1.Route) error {
...
+		// Check if err is a TLS alert.
+		opErr := &net.OpError{}
+		if errors.As(err, &opErr) {
+			if opErr.Op == "remote error" {
+				// If the operation is "remote error," it may be a TLS alert, as described in RFC 8446 section 6.
+				// crypto/tls doesn't expose its TLS alert type, but we can check the error string. If the alert is
+				// "certificate required", it means the canary lacks the client certificate necessary to complete its
+				// check. In that case, verify that the router is configured to require mTLS, and if so, assume the
+				// router is working as intended.
+				if opErr.Err.Error() == "tls: certificate required" {
```
frobware, line 92: Is this message the same in a FIPS-enabled cluster?

frobware (changes requested):

pkg/operator/controller/canary/http.go
rfredette: `client.Do()` guarantees that `response.body` is already closed when it returns a non-nil error, so the defer call isn't needed if this block is hit.

pkg/operator/controller/canary/http.go
rfredette: This message is the same on FIPS, although as we discussed out-of-band, I need to see if the message is the same in all TLS versions since it was added in TLS 1.3.

test/e2e/canary_test.go
```diff
@@ -160,3 +163,98 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+		},
+	}
+	if err := kclient.Update(context.TODO(), defaultIC); err != nil {
+		t.Fatalf("Failed to enable mTLS on default ingress controller: %v", err)
+	}
+
+	defer func() {
+		// Reset client TLS.
+		if err := kclient.Get(context.TODO(), defaultName, defaultIC); err != nil {
+			panic(fmt.Errorf("failed to get default ingresscontroller: %v", err))
```
frobware, line 212: I don't think we need to panic here. We can just call t.Fatal.

test/e2e/canary_test.go
```diff
@@ -160,3 +163,98 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+		}
+		defaultIC.Spec.ClientTLS = originalClientTLS
+		if err := kclient.Update(context.TODO(), defaultIC); err != nil {
+			panic(fmt.Errorf("Failed to restore default ingress configuration: %w", err))
```
frobware, line 216: I don't think we need to panic here. We can just call t.Fatal.

pkg/operator/controller/canary/http.go
frobware: Could we summarise the testing you've already done with different TLS profiles?

frobware:
One concern that I have is:

> This change checks specifically for the TLS alert "certificate required", and if the default ingress controller also has mTLS required, the probe function returns nil, indicating a successful probe.

is this really enough to assert status-code==200?

pkg/operator/controller/canary/http.go
rfredette: Yes, to summarize the testing with TLS profiles: 
We only allow users to set the minimum TLS version, not the maximum. The canary checks are between the router and the ingress-operator pods, both of which support TLS 1.3, so regardless of what the minimum is set to, the connection gets negotiated to v1.3, so we get teh expected "certificate required" error.

rfredette:
Closing in favor of #978

PR #880, opened 2023-01-27
NE-1183: Rename unit tests for specific functions

gcs278:
For each unit test "TestFoo" that specifically tests a function name "Foo"
or "foo", rename "TestFoo" to "Test_Foo" or "Test_foo", respectively.

* pkg/dns/alibaba/dns_test.go (TestGetRR): Rename from this...
(Test_getRR): ...to this.
(TestParseZone): Rename from this...
(Test_parseZone): ...to this.
* pkg/dns/alibaba/util/numbers_test.go (TestClamp): Rename from this...
(Test_Clamp): ...to this.
* pkg/dns/aws/dns_test.go (TestZoneMatchesTags): Rename from this...
(Test_zoneMatchesTags): ...to this.
(TestZoneIDFromResource): Rename from this...
(Test_zoneIDFromResource): ...to this.
* pkg/dns/azure/dns_test.go (TestEnsureDNS): Rename from this...
(Test_Ensure): ...to this.
(TestDeleteDNS): Rename from this...
(Test_Delete): ...to this.
* pkg/dns/ibm/private/dnssvcs_provider_test.go (TestDelete): Rename from
this...
(Test_Delete): ...to this.
(TestCreateOrUpdate): Rename from this...
(Test_createOrUpdateDNSRecord): ...to this.
* pkg/dns/ibm/public/cis_provider_test.go (TestDelete): Rename from this...
(Test_Delete): ...to this.
(TestCreateOrUpdate): Rename from this...
(Test_createOrUpdateDNSRecord): ...to this.
* pkg/operator/controller/canary/controller_test.go
(TestCycleServicePort): Rename from this...
(Test_cycleServicePort): ...to this.
* pkg/operator/controller/canary/daemonset_test.go
(TestDesiredCanaryService): Rename from this...
(Test_desiredCanaryService): ...to this.
* pkg/operator/controller/canary/namespace_test.go
(TestCanaryNamespaceChanged): Rename from this...
(Test_canaryNamespaceChanged): ...to this.
* pkg/operator/controller/canary/route_test.go (TestDesiredCanaryRoute):
Rename from this...
(Test_desiredCanaryRoute): ...to this.
(TestCanaryRouteChanged): Rename from this...
(Test_canaryRouteChanged): ...to this.
* pkg/operator/controller/canary/service_test.go
(TestDesiredCanaryService): Rename from this...
(Test_desiredCanaryService): ...to this.
* pkg/operator/controller/certificate-publisher/publish_certs_test.go
[...]

PR #965, opened 2023-07-20
OCPBUGS-16089: Set spec.subdomain on the canary route

Miciah:
#### `Test_desiredCanaryRoute`: Fix owner reference check

Change `Test_desiredCanaryRoute` to check both the route's and the service's owner references, and print the correct error message if either the route or the service has an unexpected value.

Before this change, `Test_desiredCanaryRoute` checked the route's owner references but erroneously referred to the service in the failure message if the route had an unexpected value.

* `pkg/operator/controller/canary/route_test.go` (`Test_desiredCanaryRoute`): Fix checks for owner references.

#### `Test_desiredCanaryRoute`: Use testify

* `pkg/operator/controller/canary/route_test.go` (`Test_desiredCanaryRoute`): Use `assert.Equal` and `assert.True` from the testify package.

#### Set `spec.subdomain` on the canary route

Set `spec.subdomain` on the canary route so that the route is exposed using the respective domain of any shard that exposes the route.

Before this change, the canary route specified neither `spec.subdomain` nor `spec.host`, and so the API server would set a default value for `spec.host` using the cluster ingress domain.  If the cluster ingress domain didn't match the default IngressController's domain, then this would cause canary checks to fail.

For example, the cluster-admin could end up in this situation by first using the cluster ingress config's `appsDomain` option to set a custom domain different from the domain of the default IngressController and then deleting the canary route so that it would be recreated with the custom domain.  This change ensures that the canary route continues to work in this example.

* `pkg/operator/controller/canary/controller.go` (`startCanaryRoutePolling`):
* `pkg/operator/controller/canary/http.go` (`probeRouteEndpoint`): Use the new `getRouteHost` helper function.
* `pkg/operator/controller/canary/route.go` (`canaryRouteChanged`): Check whether `spec.host` or `spec.subdomain` have changed, and update them if they have.
(`desiredCanaryService`): Specify `spec.subdomain`.
(`getRouteHost`): New function.  Return the host name of the route for the default IngressController.
* `pkg/operator/controller/canary/route_test.go` (`Test_desiredCanaryRoute`): Verify that the route has the expected value for `spec.subdomain`.
(`Test_canaryRouteChanged`): Verify that `canaryRouteChanged` checks and updates the `spec.host` and `spec.subdomain` fields.
(`Test_getRouteHost`): New test.  Verify that `getRouteHost` behaves correctly.
* `test/e2e/canary_test.go` (`TestCanaryRoute`):
* `test/e2e/client_tls_test.go` (`TestClientTLS`, `TestMTLSWithCRLs`): 
* `test/e2e/operator_test.go` (`TestHTTPHeaderCapture`, `TestHTTPCookieCapture`): 
* `test/e2e/router_compression_test.go` (`TestRouterCompressionOperation`)
(`testCompressionPolicy`, `getHttpHeaders`): Use the new `getRouteHost` test helper function.
* `test/e2e/util_test.go` (`getRouteHost`): New test helper function.  Return the host name of the route for the default IngressController.

---

~/hold~
I'll need to rebase after #880 merges.

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -280,7 +280,7 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 		err = probeRouteEndpoint(route)
 		if err != nil {
 			log.Error(err, "error performing canary route check")
-			SetCanaryRouteReachableMetric(route.Spec.Host, false)
+			SetCanaryRouteReachableMetric(getRouteHost(route), false)
 			successiveFail += 1
 			// Mark the default ingress controller degraded after 5 successive canary check failures
 			if successiveFail >= canaryCheckFailureCount {
@@ -291,7 +291,7 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			return
 		}
 
-		SetCanaryRouteReachableMetric(route.Spec.Host, true)
+		SetCanaryRouteReachableMetric(getRouteHost(route), true)
 		if err := r.setCanaryPassingStatusCondition(); err != nil {
 			log.Error(err, "error updating canary status condition")
 		}
--- a/pkg/operator/controller/canary/route.go
+++ b/pkg/operator/controller/canary/route.go
@@ -101,6 +101,16 @@ func canaryRouteChanged(current, expected *routev1.Route) (bool, *routev1.Route)
 	changed := false
 	updated := current.DeepCopy()
 
+	if current.Spec.Host != expected.Spec.Host {
+		updated.Spec.Host = expected.Spec.Host
+		changed = true
+	}
+
+	if current.Spec.Subdomain != expected.Spec.Subdomain {
+		updated.Spec.Subdomain = expected.Spec.Subdomain
+		changed = true
+	}
+
 	if !cmp.Equal(current.Spec.Port, expected.Spec.Port, cmpopts.EquateEmpty()) {
 		updated.Spec.Port = expected.Spec.Port
 		changed = true
@@ -131,6 +141,7 @@ func desiredCanaryRoute(service *corev1.Service) (*routev1.Route, error) {
 
 	route.Namespace = name.Namespace
 	route.Name = name.Name
+	route.Spec.Subdomain = fmt.Sprintf("%s-%s", route.Name, route.Namespace)
 
 	if service == nil {
 		return route, fmt.Errorf("expected non-nil canary service for canary route %s/%s", route.Namespace, route.Name)
@@ -174,4 +185,24 @@ func checkRouteAdmitted(route *routev1.Route) bool {
 	}
 
 	return false
+}
+
+// getRouteHost returns the host name of the route for the default
+// IngressController.  If the default IngressController has not added its host
+// name to the route's status, this function returns the empty string.  For
+// simplicity, this function does not check the "Admitted" status condition, so
+// it will return the host name if it is set even if the IngressController has
+// rejected the route.
+func getRouteHost(route *routev1.Route) string {
+	if route == nil {
+		return ""
+	}
+
+	for _, ingress := range route.Status.Ingress {
+		if ingress.RouterName == manifests.DefaultIngressControllerName {
+			return ingress.Host
+		}
+	}
+
+	return ""
 }
--- a/test/e2e/router_compression_test.go
+++ b/test/e2e/router_compression_test.go
@@ -198,13 +198,15 @@ func TestRouterCompressionOperation(t *testing.T) {
 		t.Fatalf("failed to get route: %v", err)
 	}
 
+	routeHost := getRouteHost(t, r, ic.Name)
+
 	// curl to canary, without the Accept-Encoding header set to gzip
-	if err := testContentEncoding(t, client, r, false, ""); err != nil {
+	if err := testContentEncoding(t, client, routeHost, false, ""); err != nil {
 		t.Error(err)
 	}
 
 	// curl to canary, WITH the Accept-Encoding header set to gzip
-	if err := testContentEncoding(t, client, r, true, "gzip"); err != nil {
+	if err := testContentEncoding(t, client, routeHost, true, "gzip"); err != nil {
 		t.Error(err)
 	}
 }
@@ -212,24 +214,24 @@ func TestRouterCompressionOperation(t *testing.T) {
 // testContentEncoding makes a call to the provided route, adds a gzip content header if addHeader is true, and
 // compares the returned Content-Encoding header to the given expectedContentEncoding.  If expectedContentEncoding
 // is the same as the returned Content-Encoding header, then the test succeeds.  Otherwise it fails.
-func testContentEncoding(t *testing.T, client *http.Client, route *routev1.Route, addHeader bool, expectedContentEncoding string) error {
+func testContentEncoding(t *testing.T, client *http.Client, routeHost string, addHeader bool, expectedContentEncoding string) error {
 	t.Helper()
 
 	if err := wait.PollImmediate(2*time.Second, 5*time.Minute, func() (bool, error) {
-		header, code, err := getHttpHeaders(client, route, addHeader)
+		header, code, err := getHttpHeaders(client, routeHost, addHeader)
 
 		if err != nil {
-			t.Logf("GET %s failed: %v, retrying...", route.Spec.Host, err)
+			t.Logf("GET %s failed: %v, retrying...", routeHost, err)
 			return false, nil
 		}
 		if code != http.StatusOK {
-			t.Logf("GET %s failed: status %v, expected %v, retrying...", route.Spec.Host, code, http.StatusOK)
+			t.Logf("GET %s failed: status %v, expected %v, retrying...", routeHost, code, http.StatusOK)
 			return false, nil // retry on 503 as pod/service may not be ready
 		}
 
 		contentEncoding := header.Get("Content-Encoding")
 		if contentEncoding != expectedContentEncoding {
-			return false, fmt.Errorf("compression error: expected %q, got %q for %s route", expectedContentEncoding, contentEncoding, route.Name)
+			return false, fmt.Errorf("compression error: expected %q, got %q for %s route", expectedContentEncoding, contentEncoding, routeHost)
 		}
 		return true, nil
 	}); err != nil {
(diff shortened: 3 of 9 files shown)


Miciah:
[e2e-aws-ovn-serial failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-serial/1682135046242177024) because AWS is full:
```
 level=error msg=Error: creating EC2 Instance: InsufficientInstanceCapacity: We currently do not have sufficient m6a.xlarge capacity in the Availability Zone you requested (us-east-1d). Our system will be working on provisioning additional capacity. You can currently get m6a.xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1a, us-east-1b, us-east-1c, us-east-1f.
level=error msg=	status code: 500, request id: 91fc067a-7acd-4a15-a8f3-00cef8eda440 
```
/test e2e-aws-ovn-serial

[e2e-azure-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1682135046707744768) because the "eastus" region went missing:
```
 level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: platform.azure.region: Invalid value: "eastus": region "eastus" is not valid or not available for this account 
```
From Slack discussion, it seems like this sort of thing might be a "hiccup".  
/test e2e-azure-operator

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/3f326ea54a6fba8cd5dc4e34925c9d573d23638b..7315ba0b63dfd9150e6294d681f5f936a5f02d5a fixes some places in the code that used `spec.host` instead of checking the canary route's status.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/7315ba0b63dfd9150e6294d681f5f936a5f02d5a..8ac3ec1e5504c69ed960973a8cd18ee7b53855cc fixes some E2E tests that used `spec.host` instead of the route status.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/8ac3ec1e5504c69ed960973a8cd18ee7b53855cc..05464949e91c275a80f65c107d9649da0ad6c5f5 revises the previous fix for E2E tests to take into account that some E2E tests do not use the default IngressController.

Miciah:
[e2e-azure-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1682464318869737472) because `TestHAProxyTimeouts` failed:
```text
TestAll/parallel/TestHAProxyTimeouts 
=== RUN   TestAll/parallel/TestHAProxyTimeouts
=== PAUSE TestAll/parallel/TestHAProxyTimeouts
=== CONT  TestAll/parallel/TestHAProxyTimeouts
=== NAME  TestAll/parallel/TestHAProxyTimeouts
    haproxy_timeouts_test.go:169: Error executing grep -oP tcp-request\s+inspect-delay\s+\K2147483647ms /var/lib/haproxy/conf/haproxy.config: etcdserver: leader changed
    haproxy_timeouts_test.go:170: stderr: 
=== NAME  TestAll/parallel/TestHAProxyTimeouts
    haproxy_timeouts_test.go:181: deleted ingresscontroller haproxy-timeout
```
`TestHAProxyTimeouts` only checks global timeout settings in `haproxy.config`; the test doesn't use or check routes at all.  This fact and the  "etcdserver" error lead me to believe that this is a novel flake caused by cluster instability unrelated to ingress.
/test e2e-azure-operator

[e2e-gcp-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1682464318936846336) because `TestUnmanagedDNSToManagedDNSInternalIngressController` failed.  This is probably the same issue as [OCPBUGS-13106](https://issues.redhat.com/browse/OCPBUGS-13106).
/test e2e-gcp-operator

[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1682464318710353920) because must-gather failed.
/test e2e-aws-operator

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/05464949e91c275a80f65c107d9649da0ad6c5f5..649435d3a9157535243c51d5bbe4d6a6a15040bc rebases for #880.

pkg/operator/controller/canary/route.go
gcs278, line 1: Should https://github.com/openshift/cluster-ingress-operator/pull/965/files#diff-aabcb3468eb7e3e07cce919ec78adc2f6012ed230991bc7f40f62868ada392d7R100 `canaryRouteChanged` now reconcile `spec.subdomain`?

I notice I can change the subdomain field and it won't get reconciled back.

test/e2e/router_compression_test.go
```diff
@@ -198,38 +198,40 @@ func TestRouterCompressionOperation(t *testing.T) {
...
 		if code != http.StatusOK {
-			t.Logf("GET %s failed: status %v, expected %v, retrying...", route.Spec.Host, code, http.StatusOK)
+			t.Logf("GET %s failed: status %v, expected %v, retrying...", routeHost, code, http.StatusOK)
 			return false, nil // retry on 503 as pod/service may not be ready
 		}
 
 		contentEncoding := header.Get("Content-Encoding")
 		if contentEncoding != expectedContentEncoding {
-			return false, fmt.Errorf("compression error: expected %q, got %q for %s route", expectedContentEncoding, contentEncoding, route.Name)
+			return false, fmt.Errorf("compression error: expected %q, got %q for %s route", expectedContentEncoding, contentEncoding, routeHost)
```
gcs278, line 234: I think this is fine, but curious if you meant to change `route.Name` to the hostname of the route?

test/e2e/util_test.go
```diff
@@ -688,3 +688,27 @@ func assertDeletedWaitForCleanup(t *testing.T, cl client.Client, thing client.Ob
...
 	}
 }
+
+// getRouteHost returns the host name of the given route for the named
+// IngressController.  If the IngressController has not added its host name to
+// the route's status, this function causes a fatal error.  For simplicity, this
+// function does not check the "Admitted" status condition, so it will return
+// the host name if it is set even if the IngressController has rejected the
+// route.
+func getRouteHost(t *testing.T, route *routev1.Route, router string) string {
```
gcs278, line 698: Would switching [canary.getRouteHost ](https://github.com/openshift/cluster-ingress-operator/pull/965/files#diff-aabcb3468eb7e3e07cce919ec78adc2f6012ed230991bc7f40f62868ada392d7R186) to public be a better option than duplicating?

Or is this intentional because it's just E2E test code?

pkg/operator/controller/canary/controller.go
gcs278, line 1: Any concerns with downgrading? I think `route.spec.subdomain` was introduced in 4.11, but I'm not sure.

gcs278 (commented):
Mainly just question, but generally looks good to me.

pkg/operator/controller/canary/route.go
Miciah: I'm ambivalent about that since (a) we don't check `spec.host`, (b) leaving `spec.host` and `spec.subdomain` on their existing values on upgrade seems lowest risk, and (c) it isn't necessary to change `canaryRouteChanged` to fix the reported bug (which concerns the behavior when someone deletes the route).  However, reconciling `spec.host` and `spec.subdomain` would be more consistent with the guiding principals for the platform.  I can go ahead and make the change.

test/e2e/router_compression_test.go
Miciah: It was an intentional change; I needed `routeHost` elsewhere in the function, and it seemed simpler to use it in this log message rather than pass in both `routeHost` and `route`.

test/e2e/util_test.go
Miciah: The function is simple and yet specialized enough that it seemed simpler to me to duplicate a few lines rather than convert the other `getRouteHost` into a public interface.

pkg/operator/controller/canary/controller.go
Miciah: Right, `spec.subdomain` was implemented in 4.11.  On downgrade from 4.14 to 4.13, the route could have `spec.subdomain` set and `spec.host` empty, and this would be fine because 4.13 doesn't reconcile either of those fields and both are implemented in 4.13.

pkg/operator/controller/canary/route.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/649435d3a9157535243c51d5bbe4d6a6a15040bc..530d326013c9169d1efee77387f33c9ec13108d4.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/649435d3a9157535243c51d5bbe4d6a6a15040bc..530d326013c9169d1efee77387f33c9ec13108d4 updates `canaryRouteChanged` and `Test_canaryRouteChanged` to check and update the `spec.host` and `spec.subdomain` fields.

test/e2e/router_compression_test.go
gcs278: Ack, sounds good to me.

gcs278:
Thanks for the detailed responses. Looks good to me.
/approve
/lgtm

Miciah:
[e2e-hypershift failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-ingress-operator/965/pull-ci-openshift-cluster-ingress-operator-master-e2e-hypershift/1688614120124846080) because a bunch of tests failed: `TestCreateClusterProxy/ValidateMetricsAreExposed`, `TestNodePool/NodePool_Tests_Group/TestNTOMachineConfigGetsRolledOut`, `TestNodePool/NodePool_Tests_Group/TestNodepoolMachineconfigGetsRolledout`, `TestNodePool/NodePool_Tests_Group/TestNodePoolReplaceUpgrade`, `TestNodePool/NodePool_Tests_Group/TestNodePoolAutoRepair`, `TestCreateCluster/ValidateMetricsAreExposed`, `TestCreateClusterCustomConfig`, `TestAutoscaling/ValidateMetricsAreExposed`, and `TestCreateClusterPrivate/ValidateMetricsAreExposed`.
/test e2e-hypershift

PR #641, opened 2021-08-05
Bug 1942657:  Ingress operator stays degraded after privateZone fixed in DNS

lmzuccarelli:
This bugfix addresses the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1942657

Verification Steps

1. Create cluster via cluster-bot version 4.8.0

2. Start local dev - ```make run-local```

3. Verify ingress is working
```
$oc get clusteroperators/ingress

NAME      VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
ingress   4.8.0     True        False         False      20m
```

4. Change tags.name (prefix with lmz)
```
$oc edit dns.config.openshift.io cluster

apiVersion: config.openshift.io/v1
  7 kind: DNS
  8 metadata:
  9   creationTimestamp: "2021-08-05T10:49:47Z"
 10   generation: 1
 11   name: cluster
 12   resourceVersion: "653"
 13   uid: d47dab3f-1955-418b-942b-ca002e7a3f99
 14 spec:
 15   baseDomain: ci-ln-tdn0hi2-d5d6b.origin-ci-int-aws.dev.rhcloud.com
 16   privateZone:
 17     tags:
 18       Name: lmz-ci-ln-tdn0hi2-d5d6b-kth95-int
 19       kubernetes.io/cluster/ci-ln-tdn0hi2-d5d6b-kth95: owned
 20   publicZone:
 21     id: Z2GYOLTZHS5VK
 22 status: {}

dns.config.openshift.io/cluster edited
```

5. Check ingress status has degraded
```
$ oc get clusteroperators/ingress
NAME      VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
ingress   4.8.0     False       True          True       4m12s
```

6. Check logs 

[...]

PR #967, opened 2023-07-20
OCPBUGS-15978: Check public DNS zone when reporting status

Miciah:
When reporting the "DNSReady" status condition on an IngressController, check the status conditions for both the public zone and the private zone of the associated DNSRecord.

Before this change, only the status condition for the private zone was used to compute the "DNSReady" status condition.  If the operator failed to publish a DNS record in the public zone, this failure was not reported in the IngressController's status conditions or the "ingress" clusteroperator status conditions.

Follow-up to #641.

* `pkg/operator/controller/ingress/status.go` (`checkZoneInConfig`): Use the new `zonesMatch` helper function to check both the public zone as well as the private zone.
(`zonesMatch`): New function.  Return a Boolean value indicating whether two DNS zones match, based on their respective ID or "Name" tags.
* `pkg/operator/controller/ingress/status_test.go` (`Test_checkZoneInConfig`): Verify that `checkZoneInConfig` checks the public zone as well as the private zone.  Delete "[PrivateZone]" from test-case descriptions as each test case is now tested as both a public zone and a private zone.

---

~/hold~
Rebase after #880 merges.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -1059,19 +1059,23 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 
 // checkZoneInConfig - private utility to check for a zone in the current config
 func checkZoneInConfig(dnsConfig *configv1.DNS, zone configv1.DNSZone) bool {
-	// check PrivateZone settings only
-	// check for private zone ID
-	if dnsConfig.Spec.PrivateZone != nil && dnsConfig.Spec.PrivateZone.ID != "" && zone.ID != "" {
-		if dnsConfig.Spec.PrivateZone.ID == zone.ID {
-			return true
-		}
+	return zonesMatch(&zone, dnsConfig.Spec.PublicZone) || zonesMatch(&zone, dnsConfig.Spec.PrivateZone)
+}
+
+// zonesMatch returns a Boolean value indicating whether two DNS zones have the
+// matching ID or "Name" tag.  If either or both zones are nil, this function
+// returns false.
+func zonesMatch(a, b *configv1.DNSZone) bool {
+	if a == nil || b == nil {
+		return false
 	}
 
-	// check for private zone Tags
-	if dnsConfig.Spec.PrivateZone != nil && dnsConfig.Spec.PrivateZone.Tags["Name"] != "" && zone.Tags["Name"] != "" {
-		if dnsConfig.Spec.PrivateZone.Tags["Name"] == zone.Tags["Name"] {
-			return true
-		}
+	if a.ID != "" && b.ID != "" && a.ID == b.ID {
+		return true
+	}
+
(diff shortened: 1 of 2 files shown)


Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/9f9e4efeabb38fb31aae32f3d54313e00662b906..04e74bd30f2438729d7ed8d22f63b1b8ce25d01b rebases for #880.
/hold cancel

pkg/operator/controller/ingress/status_test.go
candita, line 1: Now that you're testing the public zone too, maybe you should adjust the test description prefixes to something other than "[PrivateZone]".

pkg/operator/controller/ingress/status.go
```diff
@@ -1059,19 +1059,23 @@ func computeDNSStatus(ic *operatorv1.IngressController, wildcardRecord *iov1.DNS
 
 // checkZoneInConfig - private utility to check for a zone in the current config
 func checkZoneInConfig(dnsConfig *configv1.DNS, zone configv1.DNSZone) bool {
-	// check PrivateZone settings only
-	// check for private zone ID
-	if dnsConfig.Spec.PrivateZone != nil && dnsConfig.Spec.PrivateZone.ID != "" && zone.ID != "" {
-		if dnsConfig.Spec.PrivateZone.ID == zone.ID {
-			return true
-		}
+	return zonesMatch(&zone, dnsConfig.Spec.PublicZone) || zonesMatch(&zone, dnsConfig.Spec.PrivateZone)
```
candita, line 1062: Is this guaranteed to run the second half of the conjunction if the first half is true? In other words, will it just stop evaluating if the public zone matches?

candita: I get it now.  You can disregard this.

candita:
It just needs a new description for the unit tests.

/approve

candita:
Noticed quite alot of ssh failures, try again.
> core@10.0.141.24: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
> Failed to ssh to AWS instance, dumping ssh configuration
> Attempt #1: ssh -F /tmp/tmp.KScJF3KjtJ -n machine sudo journalctl > /tmp/journal.log && gzip -f /tmp/journal.log

/test e2e-hypershift

pkg/operator/controller/ingress/status_test.go
Miciah: Thanks!  Fixed in https://github.com/openshift/cluster-ingress-operator/compare/04e74bd30f2438729d7ed8d22f63b1b8ce25d01b..55780444031714fc931d90af298a4b193888977a.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/04e74bd30f2438729d7ed8d22f63b1b8ce25d01b..55780444031714fc931d90af298a4b193888977a deletes "[PrivateZone]" from test-case descriptions as each test case is now tested as both a public zone and a private zone.

PR #973, opened 2023-08-28
OCPBUGS-15900: TestMTLSWithCRLs: only try to parse HTTP status code from curl output when stdout is long enough.

rfredette:
HTTP status codes will always be 3 digits if curl gets one, so only try parsing the status code if stdout from curl is 3 characters or longer.


rfredette:
Unrelated failure
/test e2e-aws-operator

PR #978, opened 2023-09-19
OCPBUGS-9037: Change Canary to use passthrough route

rfredette:
When the default ingress controller is configured to use mTLS, connecting to edge or reencrypt routes requires the client to have a valid certificate/key pair. There is no way for a user to provide a client certificate or key to the ingress operator, so canary checks using edge encryption fail. This PR changes the canary route to be passthrough and has the canary host handle the TLS handshake, allowing it to function even if mTLS is otherwise required.

### DIFF
--- a/pkg/operator/controller/canary/controller.go
+++ b/pkg/operator/controller/canary/controller.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
@@ -19,9 +20,11 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -110,12 +113,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 			if !ok {
 				return false
 			}
-			// if Spec.Port has changed, do not trigger a reconcile
-			return cmp.Equal(oldRoute.Spec.Port, newRoute.Spec.Port)
+
+			// Check if the port in newRoute is one of the ports exposed by the canary service.
+			isValidPort := false
+			for _, portNum := range CanaryPorts {
+				switch newRoute.Spec.Port.TargetPort.Type {
+				case intstr.Int:
+					if newRoute.Spec.Port.TargetPort.IntVal == portNum {
+						isValidPort = true
+					}
+				case intstr.String:
+					// Not currently supported; for simplicity, the canary check doesn't do name to port translations,
+					// so if the canary's target port is a port name, we want to treat it as an invalid port and
+					// reconcile the route.
+				}
+				if isValidPort {
+					break
+				}
+			}
+
+			// If newRoute's port isn't one of the canary service ports, reconcile the route.
+			if !isValidPort {
+				return true
+			}
+
+			// If newRoute's port is one of the canary service ports, the change may just be the ingress operator
+			// rotating canary ports. If the *only* change between oldRoute and newRoute is the port change, ignore it.
+			return !cmp.Equal(oldRoute.Spec, newRoute.Spec, cmpopts.IgnoreFields(routev1.RouteSpec{}, "Port"))
 		},
 	}
 
 	if err := c.Watch(source.Kind(operatorCache, &routev1.Route{}), enqueueRequestForDefaultIngressController(config.Namespace), canaryRoutePredicate, updateFilter); err != nil {
+		return nil, err
+	}
+	canaryDaemonSetPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		canaryDaemonSet := operatorcontroller.CanaryDaemonSetName()
+		return o.GetNamespace() == canaryDaemonSet.Namespace && o.GetName() == canaryDaemonSet.Name
+	})
+	if err := c.Watch(source.Kind(operatorCache, &appsv1.DaemonSet{}), enqueueRequestForDefaultIngressController(config.Namespace), canaryDaemonSetPredicate); err != nil {
+		return nil, err
+	}
+	canaryServicePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		canaryService := operatorcontroller.CanaryServiceName()
+		return o.GetNamespace() == canaryService.Namespace && o.GetName() == canaryService.Name
+	})
+	if err := c.Watch(source.Kind(operatorCache, &corev1.Service{}), enqueueRequestForDefaultIngressController(config.Namespace), canaryServicePredicate); err != nil {
 		return nil, err
 	}
 
@@ -184,12 +226,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return result, fmt.Errorf("failed to get ingress controller %s: %v", request.NamespacedName.Name, err)
 	}
 
-	if val, ok := ic.Annotations[CanaryRouteRotationAnnotation]; ok {
-		v, _ := strconv.ParseBool(val)
-		r.mu.Lock()
-		r.enableCanaryRouteRotation = v
-		r.mu.Unlock()
-	}
+	val, ok := ic.Annotations[CanaryRouteRotationAnnotation]
+	v, _ := strconv.ParseBool(val)
+	r.mu.Lock()
+	r.enableCanaryRouteRotation = ok && v
+	r.mu.Unlock()
 
 	// Start probing the canary route.
 	routeProbeRunner.Do(func() {
@@ -246,7 +287,8 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 
 	errors := []timestampedError{}
 
-	go wait.Until(func() {
+	// using wait.NonSlidingUntil so that the canary runs every canaryCheckFrequency, regardless of how long the function takes
+	go wait.NonSlidingUntil(func() {
 		// Get the current canary route every iteration in case it has been modified
 		haveRoute, route, err := r.currentCanaryRoute()
 		if err != nil {
@@ -265,30 +307,6 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 			if err := r.setCanaryNotAdmittedStatusCondition(); err != nil {
 				log.Error(err, "error updating canary status condition")
 			}
-			return
-		}
-
-		// Check if canary route rotations are enabled every iteration.
-		rotationEnabled := r.isCanaryRouteRotationEnabled()
-
-		// Periodically rotate the canary route endpoint if
-		// rotationEnabled is true.
-		if rotationEnabled && checkCount > canaryCheckCycleCount {
-			haveService, service, err := r.currentCanaryService()
-			if err != nil {
-				log.Error(err, "failed to get canary service")
-				return
-			} else if !haveService {
-				log.Info("canary check service does not exist")
-				return
-			}
-			route, err = r.rotateRouteEndpoint(service, route)
-			if err != nil {
-				log.Error(err, "failed to rotate canary route endpoint")
-				return
-			}
-			checkCount = 0
-			// Give the router time to reload by returning here.
 			return
 		}
 
@@ -313,11 +331,28 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
 		}
 		successiveFail = 0
 		errors = []timestampedError{}
-		// Only increment checkCount if periodic canary route
-		// endpoint rotation is enabled to prevent unbounded
-		// integer growth.
+
+		// Check if canary route rotations are enabled every iteration.
+		rotationEnabled := r.isCanaryRouteRotationEnabled()
+		// Increment checkCount and periodically rotate the canary route endpoint if canary route rotation is enabled.
 		if rotationEnabled {
 			checkCount++
+			if checkCount >= canaryCheckCycleCount {
+				haveService, service, err := r.currentCanaryService()
+				if err != nil {
+					log.Error(err, "failed to get canary service")
+					return
+				} else if !haveService {
+					log.Info("canary check service does not exist")
+					return
+				}
+				route, err = r.rotateRouteEndpoint(service, route)
+				if err != nil {
+					log.Error(err, "failed to rotate canary route endpoint")
+					return
+				}
+				checkCount = 0
+			}
 		}
 	}, canaryCheckFrequency, stop)
 
--- a/pkg/operator/controller/canary/service.go
+++ b/pkg/operator/controller/canary/service.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 
@@ -12,23 +14,71 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// ensureCanaryService ensures the ingress canary service exists
+// CanaryPorts is the list of ports exposed by the canary service
+var CanaryPorts = []int32{8443, 8888}
+
+// ensureCanaryService ensures the ingress canary service exists.
 func (r *reconciler) ensureCanaryService(daemonsetRef metav1.OwnerReference) (bool, *corev1.Service, error) {
 	desired := desiredCanaryService(daemonsetRef)
 	haveService, current, err := r.currentCanaryService()
 	if err != nil {
 		return false, nil, err
 	}
-	if haveService {
-		return true, current, nil
+	switch {
+	case haveService:
+		if updated, err := r.updateCanaryService(current, desired); err != nil {
+			return true, current, err
+		} else if updated {
+			return r.currentCanaryService()
+		}
+	case !haveService:
+		if err := r.createCanaryService(desired); err != nil {
+			return false, nil, err
+		}
+		return true, desired, nil
 	}
-	if err := r.createCanaryService(desired); err != nil {
-		return false, nil, err
-	}
-	return true, desired, nil
+	return true, current, nil
 }
 
-// currentCanaryService gets the current ingress canary service resource
+// updateCanaryService updates the canary service if an appropriate change is detected.
+func (r *reconciler) updateCanaryService(current, desired *corev1.Service) (bool, error) {
+	changed, updated := canaryServiceChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return false, fmt.Errorf("failed to update canary service %s/%s: %w", updated.Namespace, updated.Name, err)
+	}
+	log.Info("updated canary service", "namespace", updated.Namespace, "name", updated.Name, "diff", diff)
+	return true, nil
+}
+
+// canaryServiceChanged returns true if the canary service's ports or annotations have changed. If a change occurred, it
+// also returns a copy of the service with the relevant changes merged.
+func canaryServiceChanged(current, desired *corev1.Service) (bool, *corev1.Service) {
+	changed := false
+	updated := current.DeepCopy()
+
+	if !cmp.Equal(current.Spec.Ports, desired.Spec.Ports, cmpopts.EquateEmpty()) {
+		updated.Spec.Ports = desired.Spec.Ports
+		changed = true
+	}
+	if !cmp.Equal(current.Annotations, desired.Annotations, cmpopts.EquateEmpty()) {
+		updated.Annotations = desired.Annotations
+		changed = true
+	}
+
+	if !changed {
+		return false, nil
+	}
+
+	return true, updated
+}
+
+// currentCanaryService gets the current ingress canary service resource.
 func (r *reconciler) currentCanaryService() (bool, *corev1.Service, error) {
 	current := &corev1.Service{}
 	err := r.client.Get(context.TODO(), controller.CanaryServiceName(), current)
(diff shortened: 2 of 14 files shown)


rfredette:
Looking through the logs from the latest `e2e-aws-ovn` run, I only see a single canary health check failure, which isn't enough to trigger the operator to change its `CanaryChecksSucceeding` status condition, so I think it's unlikely that that failure or the other `e2e-*-ovn` are related to these changes. I've rebased in case there were relevant changes my branch is missing, but mostly I think it just needs a retest.

rfredette:
On further investigation, there is a canary related failure in `e2e-*-ovn`, for test `[sig-arch] Managed cluster [It] should expose cluster services outside the cluster [apigroup:route.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/parallel]`. It expects the canary to return an HTTP code 200, which it seems to from this part of the test output:
> ```  Oct  3 18:29:54.955: INFO: stdout: "{\"test\":0,\"rc\":0,\"curl\":{\"code\":200},\"error\":\"\",\"body\":\"SGVhbHRoY2hlY2sgcmVxdWVzdGVkCg==\",\"headers\":\"HTTP/2 200 \\r\\nx-request-port: 8080\\r\\ncontent-type: text/plain; charset=utf-8\\r\\ncontent-length: 22\\r\\ndate: Tue, 03 Oct 2023 18:29:54 GMT\\r\\n\\r\\n\"}\n" ```

Despite getting what should be a valid response, the test fails.

rfredette:
the ovn failures require the fix in https://github.com/openshift/origin/pull/28301 or something similar, but `e2e-hypershift`'s failure seems unrelated.
/test e2e-hypershift

rfredette:
The origin test fix has merged
/retest

frobware:
Given the change from re-encrypt to passthrough why do we need to do the things you hihglighted in commit https://github.com/openshift/cluster-ingress-operator/commit/c350a7e8f0dca3f76fb882d7502018827c8f836e? What breaks, given the switch to passhthrough, if you don't do the additional reconciliation logic?

rfredette:
>Given the change from re-encrypt to passthrough why do we need to do the things you highlight in commit https://github.com/openshift/cluster-ingress-operator/commit/c350a7e8f0dca3f76fb882d7502018827c8f836e? What breaks, given the switch to passhthrough, if you don't do the additional reconciliation logic?

The tl;dr is that upgrades break without the reconciliation changes. On upgrade, the route would be updated to passthrough, but the daemonset wouldn't get updated to serve https.

I used the `service.beta.openshift.io/serving-cert-secret-name` annotation to auto-generate the serving certs for canary to use, which is an annotation on the service, and the daemonset needs to include a volume mount that serving cert, so both the service and daemonset need updates for it to work.

The daemonset reconciliation changes are pretty minor; when reconciling it, the operator now needs to merge the env variables, volumes, and volume mounts from the desired config into the updated version.

The service reconciliation changes are more involved. Prior to this change, I wouldn't say the canary service was reconciled; we just made sure it existed. But at the very least, we now need to make sure that `service.beta.openshift.io/serving-cert-secret-name` annotation is present. I added the logic to check if the current and desired versions differ, and merge the two if they do. In the current change, I made sure annotations and ports are synced between the current and desired versions, but I guess it could be limited to just syncing the annotations. Previously I added an additional port to the service for serving https, so it was necessary to sync the ports, but it's not strictly necessary as it is now.

I also added a watch on the canary daemonset and service. Those probably aren't actually necessary; The canary components should get reconciled on the operator's startup, and with the changes I've already described, all the canary components should get updated. That said, it seems like an oversight to only update the canary if the route is changed, since the daemonset, service, and route are all interconnected, so I think the watches are worth including.

frobware (commented):
The commit message in efb3bc113284886a5f12b1f2f016187ff75f1ca9 could do with some of the rationale from https://github.com/openshift/cluster-ingress-operator/pull/978#issuecomment-1775592791.

test/e2e/router_compression_test.go
```diff
@@ -140,9 +143,7 @@ func TestRouterCompressionOperation(t *testing.T) {
 		t.Fatalf("failed to get ingress controller: %v", err)
 	}
 
-	// Configure the default ingress controller to apply compression to the canary route's
-	// MIME type, which is "text/plain; charset=utf-8"
-	mimeType := []operatorv1.CompressionMIMEType{"text/plain; charset=utf-8"}
+	mimeType := []operatorv1.CompressionMIMEType{"text/html; charset=utf-8"}
```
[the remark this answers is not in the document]
rfredette: As configured, the new backend pod responds with the `text/html; charset=utf-8` MIME type. Maybe there's something I could change to get it to respond with `text/plain`, but I didn't think it was worth the effort when the test just needs the MIME type in the ingress controller to match the one coming from the backend.

rfredette:
Test failures look unrelated.
/retest

frobware:
TestAll/parallel/TestMTLSWithCRLs failed.

/test e2e-gcp-operator

frobware:
> Test failures look unrelated.

/test e2e-aws-operator

frobware:
It seem that TestInternalLoadBalancerGlobalAccessGCP is failing without this change. Let's see if the TestMTLSWithCRLs fails again.

/test e2e-gcp-operator

frobware:
The changes LGTM but we need to dig into the e2e-gcp-operator tests failures.

rfredette:
I ran all the failing tests multiple times on a GCP cluster and was unable to reproduce any failures.
/retest

pkg/operator/controller/canary/controller.go
```diff
@@ -114,6 +115,20 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind(operatorCache, &routev1.Route{}), enqueueRequestForDefaultIngressController(config.Namespace), canaryRoutePredicate, updateFilter); err != nil {
```
Miciah, line 115: It seems like the logic in `updateFilter` is incorrect; the watch should ignore the update if _only_ the port changed, and specifically _only_ if the port changed between "8443-tcp" and "8888-tcp".  If the port changes between "8080-tcp" and "8443-tcp", the watch shouldn't ignore that update.  

For upgrades, this issue is probably masked by all the other watches; _something_ will trigger reconcilation during an upgrade.  

Did you test this with the `ingress.operator.openshift.io/rotate-canary-route=true` IngressController annotation set to make sure that that still works properly?

pkg/operator/controller/canary/service.go
```diff
@@ -12,20 +14,65 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// ensureCanaryService ensures the ingress canary service exists
+// ensureCanaryService ensures the ingress canary service exists.
```
Miciah, line 17: /me grins.

pkg/operator/controller/canary/service.go
```diff
@@ -12,20 +14,65 @@ import (
...
+func (r *reconciler) updateCanaryService(current, desired *corev1.Service) (bool, error) {
+	changed, updated := canaryServiceChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return false, fmt.Errorf("failed to update canary service %s/%s: %v", updated.Namespace, updated.Name, err)
```
Miciah, line 50:
```suggestion
		return false, fmt.Errorf("failed to update canary service %s/%s: %w", updated.Namespace, updated.Name, err)
```
(I suspect that `return false, err` would be good enough, but please use `%w` if you want to decorate the client error value.)

pkg/operator/controller/canary/service_test.go
```diff
@@ -50,4 +52,62 @@ func Test_desiredCanaryService(t *testing.T) {
...
+				Name:       "foo",
+				UID:        "bar",
+				Controller: &trueVar,
+			})
+			mutated := original.DeepCopy()
+			tc.mutate(mutated)
+			changed, _ := canaryServiceChanged(original, mutated)
+			if changed != tc.expected {
+				t.Errorf("expected canaryServiceChanged to be %v, got %v", tc.expected, changed)
+			}
```
Miciah, line 110: It's useful to check the returned service to make sure that `canaryServiceChanged` both detects that something changed *and* updates the copy it returns.  We do this with other `Test_fooChanged` tests, and it catches a lot of mistakes.  
```suggestion
			if changed, updated := canaryServiceChanged(original, mutated); changed != tc.expected {
				t.Errorf("expected canaryServiceChanged to be %v, got %v", tc.expected, changed)
			} else if changed {
				if changedAgain, _ := canaryServiceChanged(mutated, updated); changedAgain {
					t.Error("canaryServiceChanged does not behave as a fixed-point function")
				}
			}
```

Miciah (commented):
Looks good overall.  No major issues; just some comments around update edge-cases, possible races in test code, and stylistic things.

pkg/operator/controller/canary/controller.go
rfredette: I didn't test with `ingress.operator.openshift.io/rotate-canary-route=true` earlier, but I've tested with it now and made some changes to `updateFilter` to only ignore updates if _only_ the port changes, and I've verified that rotating ports doesn't trigger reconciliation.

That said, I don't think it adequately covers the case you mentioned:
> If the port changes between "8080-tcp" and "8443-tcp", the watch shouldn't ignore that update.

I'm not 100% clear on the reasoning. If the route was updated from port 8080 to 8443, and assuming that was the only change, I think we would be fine to ignore that update, since 8443 is a valid port in the canary daemonset. However, if it was changed from 8443 to 8080, we would want to reconcile, since 8080 isn't one of the ports that the daemonset exposes.

Or to put it another way, I think we want to ignore updates to the route if only the port changes, and the new port is one of the ones that is exposed by the desired daemonset. Does that make sense, or am I missing something?

lihongan:
In the bug https://issues.redhat.com//browse/OCPBUGS-9037 it mentioned both ingress canary check and console health check issue with mTLS, looks this PR just to fix ingress canary check part, how about console route?

pkg/operator/controller/canary/controller.go
Miciah: > Or to put it another way, I think we want to ignore updates to the route if only the port changes, and the new port is one of the ones that is exposed by the desired daemonset. Does that make sense, or am I missing something?

That makes sense.  I'm fine with that logic.  It does mean that the watch predicate and the canary route are a little more tightly coupled; what do you think about adding a package variable with the set of ports that the canary route uses, using that new variable in the predicate, and adding an assertion in `Test_desiredCanaryService` that the service has the same set of ports?

rfredette: Yeah, I think that's a good way to go about it. I'll get started on that.

rfredette: @Miciah I think I have this implemented in a way that I'm happy with, but I wanted your opinion on part of it. I've included the minimum viable implementation in 28050d6174272fa88de21881744f00c9488218c5, but if we're going to have a package variable specifying the canary ports, I think it makes more sense to actually use that as the source of truth, so I also pushed ba50ba211efed705e83247597b831145387643f2 that uses the `CanaryPorts` variable to generate the desired ports at runtime, and uses `CanaryPorts` in `ListenAndServerTLS` instead of hardcoding which ports to serve on in the canary pod. 
Let me know what you think of that; if you agree that it's the way to go, I'll squash it into the previous commit, but I wanted to leave it separate in case there's a reason to avoid making that change.

frobware: What's supposed to happen when I [now] set:

```
  annotations:
    ingress.operator.openshift.io/rotate-canary-route: "true"
```

from the ingress-operator I see: 


```
2024-03-28T13:29:47.934Z        INFO    operator.route_metrics_controller       handler/enqueue_mapped.go:81    queueing ingresscontroller      {"name": "default"}
2024-03-28T13:29:47.934Z        INFO    operator.route_metrics_controller       handler/enqueue_mapped.go:81    queueing ingresscontroller      {"name": "default"}
2024-03-28T13:29:47.934Z        INFO    operator.route_metrics_controller       controller/controller.go:119    reconciling     {"request": {"name":"default","namespace":"openshift-ingress-operator"}}
2024-03-28T13:29:48.457Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T13:29:48.965Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
```

frobware:
Given that this change switches the canary route to use passthrough we have a couple of references to edge that we should correct:

```console
% ag 'canary.*edge.*termination'
pkg/operator/controller/canary/http.go
32:     // Use https now that the canary route uses edge termination.
54:             // The canary route uses edge termination and the
```

pkg/operator/controller/canary/controller.go
frobware: I appear to have a new canary pods now listening on 8443/8888:

```sh
% oc get pods -n openshift-ingress-canary
NAME                   READY   STATUS    RESTARTS   AGE
ingress-canary-bdpr6   1/1     Running   0          36m
ingress-canary-hzn9q   1/1     Running   0          36m
ingress-canary-rfxtc   1/1     Running   0          36m

% oc rsh -n openshift-ingress-canary ingress-canary-rfxtc
sh-5.1$ lsof -n -P -p 1
COMMAND   PID       USER   FD      TYPE DEVICE SIZE/OFF      NODE NAME
ingress-o   1 1000630000  cwd       DIR  0,116       28 186649044 /
ingress-o   1 1000630000  rtd       DIR  0,116       28 186649044 /
ingress-o   1 1000630000  txt       REG  0,116 96616736 161484348 /usr/bin/ingress-operator
ingress-o   1 1000630000  mem       REG    8,4          161484348 /usr/bin/ingress-operator (path dev=0,116)
ingress-o   1 1000630000  mem       REG    8,4          180430339 /usr/lib64/libz.so.1.2.11 (path dev=0,116)
ingress-o   1 1000630000  mem       REG    8,4           31474507 /usr/lib64/libcrypto.so.3.0.7 (path dev=0,116)
ingress-o   1 1000630000  mem       REG    8,4           31474502 /usr/lib64/libc.so.6 (path dev=0,116)
ingress-o   1 1000630000  mem       REG    8,4           31474566 /usr/lib64/libresolv.so.2 (path dev=0,116)
ingress-o   1 1000630000  mem       REG    8,4           31474498 /usr/lib64/ld-linux-x86-64.so.2 (path dev=0,116)
ingress-o   1 1000630000    0u      CHR    1,3      0t0         6 /dev/null
ingress-o   1 1000630000    1w     FIFO   0,13      0t0    353731 pipe
ingress-o   1 1000630000    2w     FIFO   0,13      0t0    353732 pipe
ingress-o   1 1000630000    3u     IPv6 354868      0t0       TCP *:8443 (LISTEN)
ingress-o   1 1000630000    4u  a_inode   0,14        0        62 [eventpoll:3,5,8]
ingress-o   1 1000630000    5r     FIFO   0,13      0t0    353765 pipe
ingress-o   1 1000630000    6w     FIFO   0,13      0t0    353765 pipe
ingress-o   1 1000630000    8u     IPv6 353135      0t0       TCP *:8888 (LISTEN)
```

And the response header appears to be correct: 

```
sh-5.1$ curl -v -k  https://localhost:8443 2>&1 | grep x-request
< x-request-port: 8443
sh-5.1$ curl -v -k  https://localhost:8888 2>&1 | grep x-request
< x-request-port: 8888
sh-5.1$ curl -v -k  https://localhost:8443 2>&1 | grep x-request
< x-request-port: 8443
sh-5.1$ curl -v -k  https://localhost:8888 2>&1 | grep x-request
< x-request-port: 8888
```

but I do repeatedly see the following message from the ingress-operator: 

```
2024-03-28T14:03:28.949Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:29.464Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:33.584Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8443, but route specifies 8888"}
2024-03-28T14:03:34.100Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8443, but route specifies 8888"}
2024-03-28T14:03:38.234Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:38.756Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:39.271Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:43.388Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8443, but route specifies 8888"}
2024-03-28T14:03:43.915Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8443, but route specifies 8888"}
2024-03-28T14:03:44.445Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8443, but route specifies 8888"}
2024-03-28T14:03:48.607Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
2024-03-28T14:03:49.116Z        ERROR   operator.canary_controller      wait/backoff.go:226     error performing canary route check     {"error": "canary request received on port 8888, but route specifies 8443"}
```

Perhaps this error has been present for a while but given the changes in the PR we should understand why this logging an error.

rfredette: I can get this error to happen, but not nearly as consistently as you saw it. Here's what I think is happening: the ingress operator rotates the canary route's port, then very soon after it pings the canary. If the ping is fast enough, it could get to the router before the haproxy config is written. In that case, haproxy would send it to the old canary port when the ingress operator expects it to hit the new one, causing this error.
Looking at the code, when port rotation is enabled, the port update does happen almost immediately before the canary probe. I think we could reverse that order and it would work essentially the same, but you'd have much more time for the haproxy config to settle.

frobware (changes requested):
I didn't get around to finishing my review but we might as well clear up the one issue I found - and I can continue again tomorrow.

pkg/operator/controller/canary/controller.go
```diff
@@ -110,14 +112,50 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+			// Check if the port in newRoute is one of the ports exposed by the canary service.
+			isValidPort := false
+			for _, portNum := range CanaryPorts {
+				switch newRoute.Spec.Port.TargetPort.Type {
+				case intstr.Int:
+					if newRoute.Spec.Port.TargetPort.IntVal == portNum {
+						isValidPort = true
+						break
+					}
+				case intstr.String:
```
frobware, line 125: Drop this if it isn't used. Or add a comment as to why this is here.

test/e2e/canary_test.go
```diff
@@ -160,3 +163,143 @@ func buildCanaryCurlPod(name, namespace, image, host string) *corev1.Pod {
...
+		}
+		t.Log("Restored clientTLS config for default ingresscontroller.")
+	})
+
+	// Watch for the RouterClientAuthCA env variable to be non-empty, indicating that the router update has begun to
+	// roll out.
+	routerDeployment := &appsv1.Deployment{}
+	routerDeploymentName := controller.RouterDeploymentName(defaultIC)
+
+	err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
```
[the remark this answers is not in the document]
rfredette: Fixed

pkg/operator/controller/canary/controller.go
rfredette: Comment added

rfredette:
test failures look unrelated.
/retest

pkg/operator/controller/canary/controller.go
```diff
@@ -189,6 +230,10 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		r.mu.Lock()
 		r.enableCanaryRouteRotation = v
 		r.mu.Unlock()
+	} else {
+		r.mu.Lock()
+		r.enableCanaryRouteRotation = false
+		r.mu.Unlock()
 	}
```
frobware, line 237: Both the if and else branch could be consolidated to: 

```suggestion
	val, ok := ic.Annotations[CanaryRouteRotationAnnotation]
	v, _ := strconv.ParseBool(val)
	r.mu.Lock()
	r.enableCanaryRouteRotation = ok && v
	r.mu.Unlock()
```

Or using defer if you prefer.

pkg/operator/controller/canary/controller.go
```diff
@@ -313,11 +335,28 @@ func (r *reconciler) startCanaryRoutePolling(stop <-chan struct{}) error {
...
-		// Only increment checkCount if periodic canary route
-		// endpoint rotation is enabled to prevent unbounded
-		// integer growth.
+
+		// Check if canary route rotations are enabled every iteration.
+		rotationEnabled := r.isCanaryRouteRotationEnabled()
+		// Increment checkCount and periodically rotate the canary route endpoint if canary route rotation is enabled.
 		if rotationEnabled {
 			checkCount++
+			if checkCount > canaryCheckCycleCount {
```
frobware, line 344: If you set `canaryCheckFrequency` to 8s and `canaryCheckCycleCount=` then you have to wait 16s.

```suggestion
                        if checkCount >= canaryCheckCycleCount {
```

frobware (changes requested):
I tested by watching for port rotations and with canaryCheckFrequency=8s and canaryCheckCycleCount=1 based on the conditional check you have to wait 16s for the port to rotate. Is that intentional or, based on my parameters, should it rotate every 8s?

frobware:
TestAllowedSourceRangesStatus - could be fixed by https://github.com/openshift/cluster-ingress-operator/pull/1087.

pkg/operator/controller/canary/controller.go
```diff
@@ -110,14 +112,53 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
-			return cmp.Equal(oldRoute.Spec.Port, newRoute.Spec.Port)
+
+			// Check if the port in newRoute is one of the ports exposed by the canary service.
+			isValidPort := false
+			for _, portNum := range CanaryPorts {
+				switch newRoute.Spec.Port.TargetPort.Type {
+				case intstr.Int:
+					if newRoute.Spec.Port.TargetPort.IntVal == portNum {
+						isValidPort = true
+						break
```
Miciah, line 123: This break doesn't do what you probably think it does—it breaks out of the `switch`, not out of the `for` loop.

pkg/operator/controller/canary/controller.go
```diff
@@ -110,14 +112,53 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+
+			// If newRoute's port isn't one of the canary service ports, reconcile the route.
+			if !isValidPort {
+				return true
+			}
+
+			// If newRoute's port is one of the canary service ports, the change may just be the ingress operator
+			// rotating canary ports. If the *only* change between oldRoute and newRoute is the port change, skip
+			// reconciling.
+			oldRoute.Spec.Port = newRoute.Spec.Port
```
Miciah, line 140: Ah, the assignment is so that the `cmp.Equal` effectively ignores a difference in `Spec.Port`, right?  Using something like `cmp.Equal(oldRoute.Spec, newRoute.Spec, cmpopts.IgnoreFields(routev1.RouteSpec{}, "Port"))` would be a little more obvious.

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1087 has merged.
/test e2e-azure-operator

Miciah:
Thanks!
/lgtm
/approve

rfredette:
e2e-aws-operator failed during startup:
```
{  openshift cluster install failed with infrastructure setup}
```
/test e2e-aws-operator

PR #981, opened 2023-10-11
OCPBUGS-20192: Require non-readonly filesystem in router container

rfredette:
Setting the filesystem to read only in the router container prevents it from rendering the haproxy config, and causes haproxy to be unable to start. Adding the security context constraint `readOnlyRootFilesystem: false` makes sure the security profile it is assigned does not require a read only filesystem.

### DIFF
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -16,6 +16,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -76,8 +77,11 @@ func checkRouterContainerSecurityContext(t *testing.T, deployment *appsv1.Deploy
 
 	for _, container := range deployment.Spec.Template.Spec.Containers {
 		if container.Name == routerContainerName {
-			if v := container.SecurityContext.AllowPrivilegeEscalation; v == nil || *v != true {
+			if allowPrivEsc := container.SecurityContext.AllowPrivilegeEscalation; allowPrivEsc == nil || !*allowPrivEsc {
 				t.Errorf("%s container does not have securityContext.allowPrivilegeEscalation: true", routerContainerName)
+			}
+			if readOnlyFS := container.SecurityContext.ReadOnlyRootFilesystem; readOnlyFS == nil || *readOnlyFS {
+				t.Errorf("%s container does not have securityContext.readOnlyRootFilesystem: false", routerContainerName)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


rfredette:
`e2e-aws-ovn` failed the infrastructure step, which likely resulted in other failures.
/test e2e-aws-ovn

Miciah (commented):
Please add a test case in `Test_deploymentConfigChanged` and add the bug reference and details from the PR description to the commit message.  Otherwise this looks fine.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -79,6 +79,9 @@ func checkRouterContainerSecurityContext(t *testing.T, deployment *appsv1.Deploy
 			if v := container.SecurityContext.AllowPrivilegeEscalation; v == nil || *v != true {
 				t.Errorf("%s container does not have securityContext.allowPrivilegeEscalation: true", routerContainerName)
 			}
+			if v := container.SecurityContext.ReadOnlyRootFilesystem; v == nil || *v != false {
```
frobware, line 82:
```suggestion
			if v := container.SecurityContext.ReadOnlyRootFilesystem; v == nil || !*v {
```

rfredette: You got it backwards, it should be `... || *v` (which probably proves the point that the condition is written poorly), but I get what you're aiming at.  I've made the appropriate change.

rfredette:
I've made the suggested changes. @Miciah and @frobware, PTAL when you have a chance.

rfredette:
Job failure looks unrelated.
/retest

pkg/operator/controller/ingress/deployment_test.go
frobware: Touché. Something other than `v` would also be more informative.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1568,13 +1571,18 @@ func Test_deploymentConfigChanged(t *testing.T) {
...
-			description: "if the router container security context changes",
+			description: "if the router container .securityContext.allowPrivilegeEscalation changes",
 			mutate: func(deployment *appsv1.Deployment) {
-				v := true
-				sc := &corev1.SecurityContext{
-					AllowPrivilegeEscalation: &v,
-				}
-				deployment.Spec.Template.Spec.Containers[0].SecurityContext = sc
+				boolFalse := false
+				deployment.Spec.Template.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = &boolFalse
```
frobware, line 1577:
```suggestion
			deployment.Spec.Template.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = pointer.Bool(false)
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1568,13 +1571,18 @@ func Test_deploymentConfigChanged(t *testing.T) {
...
+			},
+			expect: true,
+		},
+		{
+			description: "if the router container .securityContext.readOnlyRootFilesystem changes",
+			mutate: func(deployment *appsv1.Deployment) {
+				boolTrue := true
+				deployment.Spec.Template.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem = &boolTrue
```
frobware, line 1585:
```suggestion				deployment.Spec.Template.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem = pointer.Bool(true)
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1762,6 +1772,10 @@ func Test_deploymentConfigChanged(t *testing.T) {
 											Protocol:      corev1.ProtocolTCP,
 										},
 									},
+									SecurityContext: &corev1.SecurityContext{
+										AllowPrivilegeEscalation: &boolTrue,
+										ReadOnlyRootFilesystem:   &boolFalse,
```
frobware, line 1777: Use `pointer.Bool` here too.

pkg/operator/controller/ingress/deployment_test.go
rfredette: True. I've updated the variables to be more informative in my latest push.

Miciah:
/approve
/lgtm
/hold
for @frobware.

rfredette:
test failures look to be unrelated
/test e2e-azure-operator
/test e2e-gcp-operator

PR #982, opened 2023-10-13
OCPBUGS-16762: Revert "OCPBUGS-16762: Bump openshift/api for container.maxLength fix"

Miciah:
Reverts openshift/cluster-ingress-operator#979

---

The 4.15.0 fix didn't get verified in time to backport it to 4.14.0, so I am reverting the changes in 4.15.0 and will close the 4.14.0 bug report as "Can't Fix".

PR #990, opened 2023-10-25
NE-1402: Add service endpoint override capability to IBM DNS provider

marcsello:
This PR adds the ability for custom service endpoints to be configured and used by the IBM DNS provider. If no overrides configured, it falls back to the default endpoints it used before.

The overrides are read from the cluster's `Infrastrucutre` configuration (Specifically from `.status.platformStatus.ibmcloud.serviceEndpoints`): https://github.com/openshift/cluster-ingress-operator/blob/dev-add-service-overrides/vendor/github.com/openshift/api/config/v1/types_infrastructure.go#L1218

Since PowerVS uses IBM DNS provider, and uses the same configuration format, this capability is implemented for it as well.

Jira reference: https://issues.redhat.com/browse/NE-1402


cjschaef:
/retitle NE-1402: Add service endpoint override capability to IBM DNS provider

pkg/dns/ibm/common.go
```diff
@@ -27,6 +68,8 @@ type Config struct {
 	UserAgent string
 	// Zones is a list of DNS zones in which the DNS provider manages DNS records.
 	Zones []string
+	// ServiceEndpointOverrides stores overrides for certain services
```
Miciah, line 71:
```suggestion
	// ServiceEndpointOverrides stores overrides for certain services.
```

pkg/operator/controller/dns/controller.go
```diff
@@ -718,14 +718,28 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 			return &dns.FakeProvider{}, nil
 		}
 
+		// Read custom service endpoints from the infra status and set when configured
+		// When none configured, the provider will use the default values
```
Miciah, line 722:
```suggestion
		// Read custom service endpoints from the infra status and set when configured.
		// When no custom endpoint is configured for a given service, the provider will use the default value.
```

pkg/operator/controller/dns/controller.go
```diff
@@ -718,14 +718,28 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
...
+		var serviceEndpointOverrides ibm.ServiceEndpointOverrides
+		for _, endpointConfig := range infraStatus.PlatformStatus.IBMCloud.ServiceEndpoints {
+			switch endpointConfig.Name {
+			case "cis":
+				serviceEndpointOverrides.CIS = endpointConfig.URL
+			case "dnsservices":
+				serviceEndpointOverrides.DNS = endpointConfig.URL
+			case "iam":
+				serviceEndpointOverrides.IAM = endpointConfig.URL
```
Miciah, line 731: Are there named constants for "cis", "dnsservices", or "iam" that we could use?  For example, for AWS, we use named constants from aws-sdk-go.  If the IBM Cloud SDK doesn't define named constants for these service names, maybe we should define them in openshift/api.

pkg/operator/controller/dns/controller.go
```diff
@@ -735,14 +749,29 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
...
+		for _, endpointConfig := range infraStatus.PlatformStatus.PowerVS.ServiceEndpoints {
+			switch endpointConfig.Name {
+			case "cis":
+				serviceEndpointOverrides.CIS = endpointConfig.URL
+			case "dnsservices":
+				serviceEndpointOverrides.DNS = endpointConfig.URL
+			case "iam":
+				serviceEndpointOverrides.IAM = endpointConfig.URL
+			}
+		}
```
Miciah, line 765: I was going to suggest writing a helper using Go generics, but it turns out that generics isn't flexible enough for this case (see golang/go#48522).  Given that, I'll just note that the comment here should be updated per my earlier suggestion.

Miciah (commented):
Please copy the description from the PR to the commit message, and add a reference to https://issues.redhat.com/browse/NE-1402 in commit message as well.  

I have some other comments inline, mostly trivial suggestions, but we should codify the service endpoint names in openshift/api if they aren't already defined somewhere.

It's unfortunate that we don't have E2E for IBM Cloud or Power VS, but we can at least make sure verify, unit tests, etc. pass and that nothing breaks on other platforms.
/ok-to-test

pkg/operator/controller/dns/controller.go
cjschaef: While I may not have a complete list of these required service names (`cis`, `iam`, etc.) yet, as we work to close on all of the Operator changes necessary for these service endpoint overrides, I have opened a PR to get these service names defined, if they can be merged and pulled in in time for FF.
https://github.com/openshift/library-go/pull/1612

cjschaef:
Incorporating the changes to the `installer`, `CCM`, and `MAPI`, in a Disconnected environment (no Internet access), I tested these changes successfully. The Ingress operator successfully deployed and created the two router pods, interacting with IBM Cloud Services, via private endpoints (overrides).
```
# oc --kubeconfig cluster-deploys/us-east-splat-1097-26/auth/kubeconfig get co ingress; curl -m3 https://iam.cloud.ibm.com
NAME      VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.15.0-0.nightly-2023-11-06-142702   True        False         False      14m     
curl: (28) Failed to connect to iam.cloud.ibm.com port 443 after 2955 ms: Connection timed out
```
```
# oc --kubeconfig cluster-deploys/us-east-splat-1097-26/auth/kubeconfig get po -n openshift-ingress
NAME                              READY   STATUS    RESTARTS   AGE
router-default-6f88d989dc-8sq92   1/1     Running   0          32m
router-default-6f88d989dc-tlmb8   1/1     Running   0          32m
```
```
# oc --kubeconfig cluster-deploys/us-east-splat-1097-26/auth/kubeconfig get infrastructure/cluster -oyaml | yq .status.platformStatus.ibmcloud.serviceEndpoints
- name: iam
  url: https://private.us-east.iam.cloud.ibm.com
- name: vpc
  url: https://us-east.private.iaas.cloud.ibm.com/v1
- name: resourcecontroller
  url: https://private.resource-controller.cloud.ibm.com
- name: resourcemanager
  url: https://private.resource-controller.cloud.ibm.com
- name: dnsservices
  url: https://api.private.dns-svcs.cloud.ibm.com/v1
- name: cos
  url: https://s3.direct.us-east.cloud-object-storage.appdomain.cloud
```

I will run some regression tests for Public, Private, and BYON with these changes and report back this week.

marcsello:
Thank you for all comments! I've fixed most of them. One exception is the named constants, I'll wait for what the outcome of @cjschaef's PR https://github.com/openshift/library-go/pull/1612 with that.

Regarding the failing test, I can not see any relation to my changes, I'll try re-running the test to see if it was an intermittent issue.

/retest

marcsello:
This time different, unrelated tests have failed.

/retest

cjschaef:
Looks like this revert has caused us to lose the necessary `openshift/api` change we need (for `ServiceEndpoints`)
https://github.com/openshift/cluster-ingress-operator/pull/982

So we no longer have access to the `IBMCloudServiceEndpoint`
https://github.com/openshift/api/commit/2f259e8fcccdc7eaf21a2c17709cdd6bbaefdb6a

marcsello:
I have rebased my commit off master, and added a new commit to update the `openshift/api` dependency to the first version that includes `IBMCloudServiceEndpoint`.

marcsello:
It seems like unrelated tests have failed again.

/retest

pkg/operator/controller/dns/controller.go
```diff
@@ -735,14 +749,29 @@ func (r *reconciler) createDNSProvider(dnsConfig *configv1.DNS, platformStatus *
 		}
 	case configv1.PowerVSPlatformType:
 		// Power VS platform will use the ibm dns implementation
+
+		// Read custom service endpoints from the infra status and set when configured
+		// When none configured, the provider will use the default values
```
Miciah, line 754: Please update this comment as you did the corresponding comment in the `case configv1.IBMCloudPlatformType:` branch.

Miciah (commented):
A few additional comments:

* Please put the openshift/api bump commit first and the implementation commit second.  The goal is that every commit should build and pass tests (this is helpful to anyone using `git bisect` in the future, for example).  
* Please mention in the bump commit message that the purpose of the bump is to get the IBM Cloud service override API. 
* Would you mind adding the `go get` commands that you used to bump the vendored packages to the commit message for the openshift/api bump?  See https://github.com/openshift/cluster-ingress-operator/commit/e50157740b1d35d5a4bd110397a8c72cc4e2f902 for an example.  (Sorry for adding this chore; this is a relatively new expectation regarding commit messages that the team has agreed on.)

marcsello:
Thanks for the feedback @Miciah! I'll try to sort all this out by today.

Miciah:
The bot must be having a bad case of the Mondays!
/joke
/approve
/lgtm

Miciah:
Other than the vendored packages, all of the changes in this PR are specific to IBM Cloud.  Moreover, `TestUnmanagedDNSToManagedDNSInternalIngressController` is failing on other PRs, so I am confident that the failures are not caused by changes in this PR.  Let's override the failing test.  
/override ci/prow/e2e-gcp-operator
(In the meantime, there is an ongoing investigation into why `TestUnmanagedDNSToManagedDNSInternalIngressController` is failing on various PRs.)

PR #994, opened 2023-11-09
Align minTLSVersion samples with enum values

martin-aders:
The minTLSVersion sample values proposed in this CRD's documentation are leading to an error because the allowed enum values differ from the syntax of the config samples.

Symptoms observed upon applying an IngressController with the suggested values were: 'error: ingresscontrollers.operator.openshift.io "<name>" is invalid'


candita:
Hi @martin-aders thanks for your contribution.  This is not where we would fix this kind of issue, it would be fixed in the openshift/api, probably in and around https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L170. Would you mind redirecting your efforts there?

PR #999, opened 2023-11-20
NE-1400: Add Gateway API v1.0.0 CRDs

candita:
Add the v1.0.0. Gateway API CRDs and vendor the changes


candita:
> error: failed to create/update secret: failed to update secret: etcdserver: request timed out

/test e2e-aws-operator

candita:
Seeing `runc` program error, `value out of range` in e2e-aws-ovn-single-node  results:

{Pod e2e-test-oc-templates-bppl6 template-type-precision 84b727e8-afb0-4067-bd5a-ac7968be2954 v1 181129 spec.containers{test}} Failed Error: container create failed: time=\"2023-12-21T20:58:19Z\" level=error msg=\**"runc create failed: annotation org.systemd.property.TimeoutStopUSec=uint64 9223372036854775807000000 value parse error: strconv.ParseUint: parsing \\\"9223372036854775807000000\\\": value out of range\**"\n {kubelet ip-10-0-44-134.ec2.internal} 2023-12-21 20:58:20 +0000 UTC 2023-12-21 20:58:20 +0000 UTC 1 Warning 0001-01-01 00:00:00 +0000 UTC nil  nil kubelet ip-10-0-44-134.ec2.internal}"

candita:
Separate this PR into a go bump + a CRD bump. Add a task to update to OSSM 2.5 and test if there are breaking changes.

/hold

candita:
>  Starting pod/ip-10-0-13-226ec2internal-debug-j4ntd ...
> ...
>   Error from server (Forbidden): nodes "ip-10-0-13-226.ec2.internal" is forbidden: is not allowed to modify labels: node-role.kubernetes.io/etcd1
> 

/test e2e-aws-ovn-single-node

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -63,12 +60,11 @@ func (r *reconciler) ensureServiceMeshControlPlane(ctx context.Context, gatewayc
...
-		"PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER":   "true",
-		"PILOT_ENABLE_GATEWAY_API_STATUS":                  "true",
-		"PILOT_GATEWAY_API_CONTROLLER_NAME":                OpenShiftGatewayClassControllerName,
-		"PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS":           OpenShiftDefaultGatewayClassName,
-		"PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER": "false",
+		"PILOT_ENABLE_GATEWAY_CONTROLLER_MODE":   "true",
+		"PILOT_GATEWAY_API_CONTROLLER_NAME":      OpenShiftGatewayClassControllerName,
+		"PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS": OpenShiftDefaultGatewayClassName,
+		// Istio will only create the default gateway class if this is true.
+		"PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER": "true",
```
Miciah, line 67: We don't want Istio to create the default gatewayclass, according to [the workflow that is described in the enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/gateway-api-with-cluster-ingress-operator.md#workflow-description).  Are we changing the design, or is this change only for testing?

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -99,17 +95,17 @@ func desiredServiceMeshControlPlane(name types.NamespacedName, ownerRef metav1.O
...
-								ServiceSpec: corev1.ServiceSpec{
-									Type: corev1.ServiceTypeLoadBalancer,
-								},
+								Enabled: &f,
 							},
+							//Service: maistrav2.GatewayServiceConfig{
+							//	ServiceSpec: corev1.ServiceSpec{
+							//		Type: corev1.ServiceTypeLoadBalancer,
+							//	},
+							//},
```
Miciah, line 108: I am very much in favor of getting rid of the istio-ingressgateway service.  Have we recorded the decision to eliminate it anywhere?  Perhaps a revision to the existing EP, or a follow-up EP?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -58,7 +58,7 @@ func desiredSubscription(name types.NamespacedName) (*operatorsv1alpha1.Subscrip
 			Channel:                "stable",
 			InstallPlanApproval:    operatorsv1alpha1.ApprovalAutomatic,
 			Package:                "servicemeshoperator",
-			CatalogSource:          "redhat-operators",
+			CatalogSource:          "quay-iib",
```
Miciah, line 61: Is this just for testing?

Miciah (commented):
I think you need to squash the first two commits or else `make verify` will fail on the first one.

pkg/operator/controller/gatewayclass/subscription.go
candita: Yes, it's the only way we can get 2.5 images right now.

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
candita: Let me look into this some more.  Note that when `PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER` is false then istiod also doesn't reconcile the GatewayClass at all.  Jacek found this and discussed in https://redhat-internal.slack.com/archives/C01R9E3SFEV/p1706222283240929?thread_ts=1705959981.087229&cid=C01R9E3SFEV  The default gatewayclass that is created would have  name: openshift-default and controllerName: openshift.io/gateway-controller.

candita:
Hold for OSSM 2.6
/hold

candita:
This is clutter until OSSM 2.6 is ready.
/close

PR #1000, opened 2023-11-21
OCPBUGS-23999: Set owner for default ingress configmap

vrutkovs:
This adds annotation which identifies the component used to file issues with this certificate. The annotation is verified by [the test](https://github.com/openshift/origin/blob/master/test/extended/operators/certs.go?rgh-link-date=2023-11-21T09%3A53%3A38Z#L130), so that all certificates would be registered and have owners assigned. More metadata would be added later.

See https://github.com/openshift/enhancements/pull/1502 for proposed workflow with TLS registry

### DIFF
--- a/pkg/operator/controller/certificate-publisher/publish_ca.go
+++ b/pkg/operator/controller/certificate-publisher/publish_ca.go
@@ -3,7 +3,9 @@ package certificatepublisher
 import (
 	"context"
 	"fmt"
+	"reflect"
 
+	"github.com/openshift/api/annotations"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 
 	corev1 "k8s.io/api/core/v1"
@@ -19,6 +21,9 @@ func (r *reconciler) ensureDefaultIngressCertConfigMap(caBundle string) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name.Name,
 			Namespace: name.Namespace,
+			Annotations: map[string]string{
+				annotations.OpenShiftComponent: "Networking / router",
+			},
 		},
 		Data: map[string]string{
 			"ca-bundle.crt": caBundle,
@@ -91,6 +96,7 @@ func (r *reconciler) updateRouterCAConfigMap(current, desired *corev1.ConfigMap)
 	}
 	updated := current.DeepCopy()
 	updated.Data = desired.Data
+	updated.Annotations = desired.Annotations
 	if err := r.client.Update(context.TODO(), updated); err != nil {
 		return false, err
 	}
@@ -111,6 +117,9 @@ func (r *reconciler) deleteRouterCAConfigMap(cm *corev1.ConfigMap) (bool, error)
 
 // routerCAConfigMapsEqual compares two router CA configmaps.
 func routerCAConfigMapsEqual(a, b *corev1.ConfigMap) bool {
+	if !reflect.DeepEqual(a.Annotations, b.Annotations) {
+		return false
+	}
 	if a.Data["ca-bundle.crt"] != b.Data["ca-bundle.crt"] {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


Miciah:
@vrutkovs, this PR needs to be rebased since #990 merged.  Also, the PR needs a Jira issue.  Do you intend to get this change in OpenShift 4.15?

pkg/operator/controller/certificate-publisher/publish_ca.go
```diff
@@ -19,6 +20,9 @@ func (r *reconciler) ensureDefaultIngressCertConfigMap(caBundle string) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name.Name,
 			Namespace: name.Namespace,
+			Annotations: map[string]string{
+				annotations.OpenShiftComponent: "Networking / router",
+			},
```
Miciah, line 25: You'll need to update `updateRouterCAConfigMap` and `deleteRouterCAConfigMap` if you want the operator to add this annotation on upgrade.

pkg/operator/controller/certificate/ca.go
```diff
@@ -125,6 +126,9 @@ func desiredRouterCASecret(namespace string) (*corev1.Secret, error) {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name.Name,
 			Namespace: name.Namespace,
+			Annotations: map[string]string{
+				annotations.OpenShiftComponent: "Networking / router",
+			},
```
Miciah, line 131: You'll need to add update logic to `ensureRouterCASecret` if you want the operator to add this annotation on upgrade.

lihongan:
cc @melvinjoseph86

candita:
Since you've started this...
/assign @Miciah

candita:
@vrutkovs should we close this one, or will you pursue the changes that Miciah mentioned?

pkg/operator/controller/certificate-publisher/publish_ca.go
vrutkovs: Updated `routerCAConfigMapsEqual` to return false when annotations are not set or do not match the desired. This will trigger `Update` call in `updateRouterCAConfigMap`

pkg/operator/controller/certificate/ca.go
vrutkovs: I don't think its required, `if !reflect.DeepEqual` in `ensureRouterCASecret` would start an update if annotations mismatch

vrutkovs:
Verified that annotations are being set on clean install and in the upgrade job:
```
"metadata": {
                "annotations": {
                    "openshift.io/owning-component": "Networking / router"
                },
                "creationTimestamp": "2024-05-16T07:07:04Z",
                "name": "default-ingress-cert",
                "namespace": "openshift-config-managed",
                "resourceVersion": "7828",
                "uid": "04ae0ff8-064a-43fd-9ebf-012b8c3f093f"
            }
```

gcs278:
FYI: Need https://github.com/openshift/cluster-ingress-operator/pull/1054 to fixed operator E2E which is permafailing.

pkg/operator/controller/certificate/ca.go
```diff
@@ -41,6 +43,13 @@ func (r *reconciler) ensureRouterCASecret() (*corev1.Secret, error) {
 		r.recorder.Event(new, "Normal", "CreatedWildcardCACert", "Created a default wildcard CA certificate")
 		return new, nil
 
+	} else if !reflect.DeepEqual(current.Annotations, desired.Annotations) {
+		updated := current.DeepCopy()
+		updated.Data = desired.Data
+		updated.Annotations = desired.Annotations
```
candita, line 49: You may need to insert/merge the annotations rather than overwriting all of them.  Adding a unit test for `addition of a new annotation to existing annotations` would be helpful.

vrutkovs: I'd expect `desired` annotations to be an full list of annotation, not just the required minimum. Is there an example of optional annotation we want to keep in these secrets?

PR #1016, opened 2024-01-11
CORS-2317: Add Ingress LB IPs to Infra CR and set DNS unmanaged when BYO DNS is enabled

sadasu:
- Bring in the latest updates to Infrastructure CR from openshift/api
- Update GCPPlatformStatus with the Ingress LB IPs when `ClusterHosted` DNS is enabled
- Put DNS in Unmanaged state because an in-cluster DNS solution would be in use in place of the platform default DNS.

Based on enhancement: https://github.com/openshift/enhancements/pull/1468


sadasu:
@Miciah could PTAL?

pkg/operator/controller/ingress/controller.go
```diff
@@ -1131,6 +1131,10 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	SetIngressControllerNLBMetric(ci)
 
+	if err := r.updateInfraIngressLoadBalancerIPs(lbService, platformStatus); err != nil {
+		errs = append(errs, fmt.Errorf("failed to Ingress Load Balancer IPs to Infrastructure PlatformStatus: %v", err))
+	}
+
```
Miciah, line 1146: `lbService` could be nil, and `updateInfraIngressLoadBalancerIPs` doesn't have a nil check.  You need to check `haveLB` here or add a nil check in `updateInfraIngressLoadBalancerIPs`.  I prefer the former because we prefer to use Booleans instead of nil checks in this operator, and we happen to have a Boolean here.
```suggestion
	if haveLB {
		if err := r.updateInfraIngressLoadBalancerIPs(lbService, platformStatus); err != nil {
			errs = append(errs, fmt.Errorf("failed to Ingress Load Balancer IPs to Infrastructure PlatformStatus: %v", err))
		}
	}
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -1205,3 +1209,30 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
 
 	return true, nil
 }
+
+// updateInfraIngressLoadBalancerIPs updates GCP's PlatformStatus with Ingress LB IPs when the DNSType is `ClusterHosted`.
+func (r *reconciler) updateInfraIngressLoadBalancerIPs(service *corev1.Service, platformStatus *configv1.PlatformStatus) error {
+	if platformStatus != nil && platformStatus.Type == configv1.GCPPlatformType {
+		switch platformStatus.GCP.CloudLoadBalancerConfig.DNSType {
```
Miciah, line 1225: I think it's safe to assume that `platformStatus.GCP` is not nil, but don't we need a nil check for `platformStatus.GCP.CloudLoadBalancerConfig`?

pkg/operator/controller/ingress/controller.go
```diff
@@ -1205,3 +1209,30 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+			for _, ingress := range ingresses {
+				if len(ingress.IP) > 0 {
+					ingressLBIPs = append(ingressLBIPs, configv1.IP(ingress.IP))
+				}
+			}
+			if platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted == nil {
+				platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted = &configv1.CloudLoadBalancerIPs{}
+			}
+			platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs = ingressLBIPs
+			return nil
```
Miciah, line 1241: Are you missing a call to `r.client.Status().Update(...)` here?  And I think you'll need to add a permission to https://github.com/openshift/cluster-ingress-operator/blob/master/manifests/00-cluster-role.yaml:
```yaml
- apiGroups:
  - config.openshift.io
  resources:
  - infrastructures/status
  verbs:
  - update
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -448,6 +448,15 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 			effectiveStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
 		}
 
+		// When the platform's default DNS solution cannot be used, set the DNSManagementPolicy
+		// accordingly. This feature is currently being implemented first for GCP. Will be
+		// extended to AWS and Azure platforms later.
+		if platformStatus.Type == configv1.GCPPlatformType && platformStatus.GCP != nil {
+			if platformStatus.GCP.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
```
Miciah, line 455: You need a nil check for `platformStatus.GCP.CloudLoadBalancerConfig`.

Miciah (commented):
Would you mind combining the first three commits?

* "Updating openshift/api to bring in changes for BYO DNS for GCP"
* "Update vendoring as result of openshift/api and k8s.io version bump"
* "Update IngressController CRD with the openshift/api version bump"

The goal here is that each individual commit should pass `make test verify`.  I do appreciate that you separated the vendor bump from the implementation commits!

pkg/operator/controller/ingress/controller.go
Miciah: Oh, also, I think it would be appropriate to check whether `platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs` is already equal to `ingressLBIPs` and avoid doing a spurious update, assuming you do mean to call `r.client.Status().Update(...)` here.  I think the order doesn't matter, in which case you should do a comparison after sorting the slice, for example something like this:
```go
			ipCmpOpts := []cmp.Option{
				cmpopts.EquateEmpty(),
				cmpopts.SortSlices(func(a, b configv1.IP) bool {
					return a < b
				}),
			}
			if !cmp.Equal(platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, ingressLBIPs, ipCmpOpts...) {
				return r.Status().Update(...)
			}
```
(You'll need to import `"github.com/google/go-cmp/cmp"` and `"github.com/google/go-cmp/cmp/cmpopts"`.)

gcs278 (commented):
No issues from me other than a nit pick and what @Miciah identified. I think a unit test for `updateInfraIngressLoadBalancerIPs` would be a nice-to-have, but I don't feel strongly about it and I won't hold up the PR for it.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1131,6 +1140,10 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	SetIngressControllerNLBMetric(ci)
 
+	if err := r.updateInfraIngressLoadBalancerIPs(lbService, platformStatus); err != nil {
+		errs = append(errs, fmt.Errorf("failed to Ingress Load Balancer IPs to Infrastructure PlatformStatus: %v", err))
```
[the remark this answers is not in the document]
Miciah: Also, please use `%w` instead of `%v` when wrapping an error value with `fmt.Errorf`.

pkg/operator/controller/ingress/controller.go
sadasu: Done.

sadasu:
@Miciah and @gcs278, thanks for the awesome comments! Afaik, I have addressed them all.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1131,6 +1143,19 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	SetIngressControllerNLBMetric(ci)
 
+	// If the lbService exists, then update Infra CR's PlatformStatus with the Ingress LB IPs
+	if haveLB {
+		if updated, err := updateInfraIngressLoadBalancerIPs(lbService, infraConfig); err != nil {
+			errs = append(errs, fmt.Errorf("failed to update Infrastructure PlatformStatus: %w", err))
+		} else if updated {
+			if err := r.client.Status().Update(context.TODO(), infraConfig); err != nil {
+				return fmt.Errorf("failed to update Infra Status with Ingress LB IPs: %v", err)
```
gcs278, line 1152: I think this might be a mistake.

sadasu: Yes. Nice catch!. Fixed it.

gcs278:
/lgtm
@Miciah could you take a look and approve?

pkg/operator/controller/ingress/controller.go
```diff
@@ -1131,6 +1134,18 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	SetIngressControllerNLBMetric(ci)
 
+	// If the lbService exists, then update Infra CR's PlatformStatus with the Ingress LB IPs
+	if haveLB {
+		if updated, err := updateInfraIngressLoadBalancerIPs(lbService, infraConfig); err != nil {
+			errs = append(errs, fmt.Errorf("failed to update Infrastructure PlatformStatus: %w", err))
+		} else if updated {
+			if err := r.client.Status().Update(context.TODO(), infraConfig); err != nil {
```
Miciah, line 1151: This code is a little surprising to me because usually in this operator `updateFoo()` performs the update itself.  Could you move the `r.client.Status().Update()` call to `updateInfraIngressLoadBalancerIPs` itself for consistency with other code?

For completeness, this is the pattern we use in this operator:
<details>

```go
func (r *reconciler) ensureFoo(...) (bool, *foov1.Foo, error) {
	haveFoo, current, err := r.currentFoo()
	if err != nil {
		return false, nil, err
	}

	wantFoo, desired, err := desiredFoo()
	if err != nil {
		return false, nil, err
	}

	switch {
	case !wantFoo && !haveFoo:
		return false, nil, nil
	case !wantFoo && haveFoo:
		// Delete current.
	case wantFoo && !haveFoo:
		// Create desired.
	case wantFoo && haveFoo:
		// Update current to match desired.
		if updated, err := r.updateFoo(current, desired); err != nil {
			return true, current, err
		} else if updated {
			return r.currentFoo()
		}
	}

	return true, current, nil
}

func desiredFoo() (bool, *foov1.Foo, error) {
	foo := &corev1.Foo{
		// ...
	}

	return true, foo, nil
}

func (r *reconciler) updateFoo(current, desired *corev1.Foo) (bool, error) {
	changed, updated := fooChanged(current, desired)
	if !changed {
		return false, nil
	}

	// Diff before updating because the client may mutate the object.
	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
	if err := r.client.Update(context.TODO(), updated); err != nil {
		return false, err
	}
	log.Info("updated foo", "namespace", updated.Namespace, "name", updated.Name, "diff", diff)

	return true, nil
}

func fooChanged(current, expected *corev1.Foo) (bool, *corev1.Foo) {
	changed := false
	fooCmpOpts := []cmp.Option{
		// ...
	}
	if !cmp.Equal(current.Spec, expected.Spec, fooCmpOpts...) {
		changed = true
	}

	if !changed {
		return false, nil
	}

	updated := current.DeepCopy()
	// ...

	return true, updated
}
```
</details>

We usually have unit tests for `fooChanged` (and sometimes for `desiredFoo` as well).

pkg/operator/controller/ingress/controller.go
```diff
@@ -1205,3 +1220,42 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+	platformStatus := infraConfig.Status.PlatformStatus
+	if platformStatus != nil {
+		switch platformStatus.Type {
+		case configv1.GCPPlatformType:
+			if platformStatus.GCP != nil && platformStatus.GCP.CloudLoadBalancerConfig != nil && platformStatus.GCP.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
+				// The cluster has to run its own CoreDNS pod for DNS. Update Infra CR
+				// with the Ingress LB IPs. These values are used to configure the
+				// in-cluster DNS to provide resolution for *.apps.
+				if infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted == nil {
+					infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted = &configv1.CloudLoadBalancerIPs{}
```
Miciah, line 1244: Any reason not to use `platformStatus` status here?
```suggestion
				if platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted == nil {
					platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted = &configv1.CloudLoadBalancerIPs{}
```
This makes the lines a little shorter, but more importantly, it makes it more obvious that you aren't potentially dereferencing a nil pointer since you checked `platformStatus.GCP != nil && platformStatus.GCP.CloudLoadBalancerConfig` above.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1205,3 +1220,42 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+					}
+				}
+				ipCmpOpts := []cmp.Option{
+					cmpopts.EquateEmpty(),
+					cmpopts.SortSlices(func(a, b configv1.IP) bool {
+						return a < b
+					}),
+				}
+				if !cmp.Equal(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, ingressLBIPs, ipCmpOpts...) {
+					infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs = ingressLBIPs
```
Miciah, line 1260:
```suggestion
				if !cmp.Equal(platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, ingressLBIPs, ipCmpOpts...) {
					platformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs = ingressLBIPs
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -1205,3 +1220,42 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+					return true, nil
+				}
+			}
+			return false, nil
+		default:
+			return false, nil
+		}
+	}
+	return false, fmt.Errorf("invalid PlatformStatus within Infrastructure config")
```
Miciah, line 1269: Instead of the following:
```go
	platformStatus := infraConfig.Status.PlatformStatus
	if platformStatus != nil {
		switch platformStatus.Type {
		case configv1.GCPPlatformType:
			// ...
			return false, nil
		default:
			return false, nil
		}
	}
	return false, fmt.Errorf("invalid PlatformStatus within Infrastructure config")
```
Please do the following:
```go
	platformStatus := infraConfig.Status.PlatformStatus
	if platformStatus == nil {
		return false, fmt.Errorf("invalid PlatformStatus within Infrastructure config")
	}
	switch platformStatus.Type {
	case configv1.GCPPlatformType:
		// ...
		return false, nil
	default:
		return false, nil
	}
```
This puts the error condition (`platformStatus == nil`) with the error value (`fmt.Errorf("invalid PlatformStatus within Infrastructure config")`).  As a nice side-effect, it reduces the function's main logic.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1551,3 +1553,135 @@ func Test_IsProxyProtocolNeeded(t *testing.T) {
...
+		},
+		{
+			description:   "gcp platform with DNSType and no LB IP",
+			platform:      &gcpPlatformWithDNSType,
+			ingresses:     []corev1.LoadBalancerIngress{},
+			expectUpdated: false,
+			expectError:   false,
+		},
+		{
+			description:   "gcp platform with DNSType and no LB IP",
```
Miciah, line 1632: This test case has the same description as the previous one.  
```suggestion
			description:   "gcp platform with DNSType and no LB IP in infra config, service has 1 IP",
```

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1551,3 +1553,135 @@ func Test_IsProxyProtocolNeeded(t *testing.T) {
...
+				ingressLBs := service.Status.LoadBalancer.Ingress
+				for _, ingress := range ingressLBs {
+					if len(ingress.IP) > 0 {
+						if !slices.Contains(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, configv1.IP(ingress.IP)) {
+							t.Errorf("expected Infra CR to contain %s", ingress.IP)
+						}
+					}
+				}
+			}
+
```
Miciah, line 1684: You have an extra blank line here.

pkg/operator/controller/ingress/controller.go
sadasu: Done

pkg/operator/controller/ingress/controller.go
sadasu: I followed the pattern seen in https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/controller.go#L1119-L1130

Miciah: `syncIngressControllerStatus` actually does an update using `r.client.Status().Update(context.TODO(), updated)`, and then `ensureIngressController` does a `r.client.Get(...)` so it has the newly updated object.  

My point is that with `syncIngressControllerStatus` (in your example) and `updateFoo` (in my example), the name suggests that the function updates the API object in etcd, and the function does exactly that.  However, this new `updateInfraIngressLoadBalancerIPs` returns a pointer to a `configv1.Infrastructure` value without actually updating object in etcd.  

So I'd prefer this new `updateInfraIngressLoadBalancerIPs` function either call `r.client.Status().Update(context.TODO(), infraConfig)` itself, or that `updateInfraIngressLoadBalancerIPs` be named something different, such as `computeUpdatedInfraFromService`.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1131,6 +1133,18 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	SetIngressControllerNLBMetric(ci)
 
+	// If the lbService exists, then update Infra CR's PlatformStatus with the Ingress LB IPs.
+	if haveLB {
+		if updated, err := updateInfraIngressLoadBalancerIPs(lbService, infraConfig); err != nil {
+			errs = append(errs, fmt.Errorf("failed to update Infrastructure PlatformStatus: %w", err))
+		} else if updated {
+			if err := r.client.Status().Update(context.TODO(), infraConfig); err != nil {
```
Miciah, line 1150: I think this code should only update the infra config if this is the service for the "default" IngressController, right?  That is, you need something like `if haveLB && ci.Name == manifests.DefaultIngressControllerName`.  Otherwise, if you have a custom IngressController that uses `spec.endpointPublishingStrategy.type: LoadBalancerService`, the operator may keep updating the infra config status with the default IngressController's ingress IP addresses, then the custom IngressController's ingress IP addresses, then the default's, then the custom's, and so on.  (Sorry I didn't think of this during earlier reviews.)

sadasu: Done. 
Will the custom Ingress Controller's LB service use the cloud's default DNS then?

Miciah: Yeah, the cluster-admin will need to configure DNS records for the custom IngressController manually.  I think that's fine—the cluster-admin is expected to configure DNS records anyway, post-install, for the default IngressController, right?  And cluster-ingress-operator only needs to update the infra config so that the platform can provide DNS name resolution for internal clients connecting to platform routes (console, oauth) so that installation can complete?

sadasu: > And cluster-ingress-operator only needs to update the infra config so that the platform can provide DNS name resolution for internal clients connecting to platform routes (console, oauth) so that installation can complete?

Yes, that is correct.

> Yeah, the cluster-admin will need to configure DNS records for the custom IngressController manually.
That is fine. I just want to make sure we are not automatically configuring the cloud default DNS for the custom Ingress Controller case.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1063,10 +1074,10 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 
 	var lbService *corev1.Service
```
gcs278, line 1075: I think this is unnecessary now since you have:
`haveLB, lbService, err := r.ensureLoadBalancerService(ci, deploymentRef, platformStatus)`
which re-declares `lbService`

sadasu: Duh! That is correct. Updated.

go.mod
```diff
@@ -26,27 +26,28 @@ require (
...
-	google.golang.org/grpc v1.54.0
+	google.golang.org/api v0.126.0
+	google.golang.org/grpc v1.58.3
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.28.3
-	k8s.io/apiextensions-apiserver v0.28.3
-	k8s.io/apimachinery v0.28.3
-	k8s.io/apiserver v0.28.3
+	k8s.io/api v0.29.0
```
gcs278, line 39: Could we go ahead and take 0.29.1 for these K8S packages?

Miciah: This PR only bumps the k8s.io/* packages to v0.29.0 because the openshift/api bump requires doing so.  I expect we'll bump to k8s.io/api v0.29.1 or later in a future PR (for example, for a controller-runtime bump or another openshift/api bump).

Miciah:
/approve
/lgtm
/hold
for @gcs278's comments and additional review if needed.  @gcs278, feel free to release the hold if your comments have been satisfactorily addressed.

go.mod
gcs278: That's fine if we want do it later. It was more of a "why not use the latest if we are bumping now" sort of thing, but we can just plan on doing it later.

gcs278:
Thanks for the quick responses @sadasu 
/unhold
/lgtm

PR #1017, opened 2024-01-23
CFE-987: Use RouterExternalCertificate feature gate for adding ROUTER_ENABLE_EXTERNAL_CERTIFICATE env var to the router pods

arkadeepsen:
Use `RouterExternalCertificate` feature gate and add `ROUTER_ENABLE_EXTERNAL_CERTIFICATE` env var to the router pod based on the feature gate according to the EP https://github.com/openshift/enhancements/pull/1307.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -119,7 +119,7 @@ func (r *reconciler) ensureRouterDeployment(ci *operatorv1.IngressController, in
 	if err != nil {
 		return false, nil, fmt.Errorf("failed to determine if proxy protocol is needed for ingresscontroller %s/%s: %v", ci.Namespace, ci.Name, err)
 	}
-	desired, err := desiredRouterDeployment(ci, r.config.IngressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig)
+	desired, err := desiredRouterDeployment(ci, r.config.IngressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig, r.config.RouteExternalCertificateEnabled)
 	if err != nil {
 		return haveDepl, current, fmt.Errorf("failed to build router deployment: %v", err)
 	}
@@ -241,7 +241,7 @@ func headerValues(values []operatorv1.IngressControllerHTTPHeader) string {
 }
 
 // desiredRouterDeployment returns the desired router deployment.
-func desiredRouterDeployment(ci *operatorv1.IngressController, ingressControllerImage string, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, apiConfig *configv1.APIServer, networkConfig *configv1.Network, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy) (*appsv1.Deployment, error) {
+func desiredRouterDeployment(ci *operatorv1.IngressController, ingressControllerImage string, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, apiConfig *configv1.APIServer, networkConfig *configv1.Network, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy, routeExternalCertificateEnabled bool) (*appsv1.Deployment, error) {
 	deployment := manifests.RouterDeployment()
 	name := controller.RouterDeploymentName(ci)
 	deployment.Name = name.Name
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -1123,6 +1123,17 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+	// Add router external certificate environment variable.
+	if routeExternalCertificateEnabled {
+		env = append(env,
+			corev1.EnvVar{Name: "ROUTER_EXTERNAL_CERTIFICATE", Value: "true"},
+		)
+	} else {
+		env = append(env,
+			corev1.EnvVar{Name: "ROUTER_EXTERNAL_CERTIFICATE", Value: "false"},
+		)
+	}
```
chiragkyal, line 1135: I think it's better to only add the env variable, if routeExternalCertificateEnabled. If feature-gate is not enabled, then do nothing. At the router side we will check `ROUTER_EXTERNAL_CERTIFICATE` exists or not, and take actions accordingly. 

In that case, we can skip the else statement

arkadeepsen: Done

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1123,6 +1123,13 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		)
 	}
 
+	// Add router external certificate environment variable.
+	if routeExternalCertificateEnabled {
+		env = append(env,
+			corev1.EnvVar{Name: "ROUTER_EXTERNAL_CERTIFICATE", Value: "true"},
```
alebedev87, line 1129: The naming in `ROUTER_EXTERNAL_CERTIFICATE` doesn't seem to reflect the meaning passed to the router. It looks like we try to pass the name or some sort of reference to an external certificate. While what we do is we try to enable the handling of the external certificate field.
```suggestion
			corev1.EnvVar{Name: "ROUTER_EXTERNAL_CERTIFICATE_ENABLED", Value: "true"},
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2265,3 +2265,37 @@ func TestDesiredRouterDeploymentDefaultPlacement(t *testing.T) {
...
+		t.Fatalf("invalid router Deployment: %v", err)
+	}
+
+	// Verify that router external certificate env var is not added.
+	expectedEnv := []envData{
+		{"ROUTER_EXTERNAL_CERTIFICATE", false, ""},
+	}
+
+	if err := checkDeploymentEnvironment(t, deployment, expectedEnv); err != nil {
+		t.Errorf("empty configv1.Proxy: %v", err)
```
alebedev87, line 2283: Proxy?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2265,3 +2265,37 @@ func TestDesiredRouterDeploymentDefaultPlacement(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("invalid router Deployment: %v", err)
+	}
+
+	// Verify that router external certificate env var is not added.
+	expectedEnv := []envData{
+		{"ROUTER_EXTERNAL_CERTIFICATE", false, ""},
+	}
+
+	if err := checkDeploymentEnvironment(t, deployment, expectedEnv); err != nil {
```
alebedev87, line 2282: This part could be integrated into the existing `Test_desiredRouterDeployment` just by adding `[]envData{{"ROUTER_EXTERNAL_CERTIFICATE", false, ""}}` to the existing set of env data. That would save us some lines of code and some CPU cycles. But I'm not insisting on this one.

arkadeepsen: I didn't want to tinker with the existing `Test_desiredRouterDeployment()` function as there are other `TestDesiredRouterDeployment__()` functions and I thought each function was designed for testing specific parameters.

pkg/operator/controller/ingress/deployment.go
alebedev87: Yeah, it's hard to focus on the right name in the future code during the EP review.  I think it's fine to tweak it in the EP if we find a better name down the implementation line.
I looked at the other variables from the router deployment and found [ROUTER_ENABLE_COMPRESSION](https://github.com/openshift/cluster-ingress-operator/blob/d31546e05cc00021e5fbd441ba6dc6739c1bf274/pkg/operator/controller/ingress/deployment.go#L96) or [ROUTER_DISABLE_HTTP2](https://github.com/openshift/cluster-ingress-operator/blob/d31546e05cc00021e5fbd441ba6dc6739c1bf274/pkg/operator/controller/ingress/deployment.go#L75). I think we should follow the same pattern:
```suggestion
			corev1.EnvVar{Name: "ROUTER_ENABLE_EXTERNAL_CERTIFICATE", Value: "true"},
```

pkg/operator/controller/ingress/deployment_test.go
alebedev87: > I thought each function was designed for testing specific parameters.

Yes. I'm not a fan of how it's done currently but that is a subject for the future refactoring, out of the scope of this PR.

pkg/operator/controller/ingress/deployment.go
arkadeepsen: Sure. @chiragkyal can you please take a note of this change in the EP? This would also need updating the Router implementation.

pkg/operator/controller/ingress/deployment_test.go
arkadeepsen: I'll then add the new test function. Whenever a future refactoring is taken up then it probably can be combined with other env var test functions.

alebedev87: > I'll then add the new test function.

You already did. I'm fine with it.

arkadeepsen:
> @arkadeepsen : can you please update the title and the description of the PR according to the latest changes?

Done

PR #1018, opened 2024-01-24
NE-1400: Bump to OSSM 2.5 and Gateway API v0.6.2 CRDs

candita:
OSSM 2.5 supports Istio 1.18 and therefore only Gateway API v0.6.2 CRDS. 

 - Pull the new CRD yaml from upstream and update pkg/manifests/assets/gateway-api
- Update the pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go to reflect changes required by OSSM 2.5.

### DIFF
--- a/pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
+++ b/pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
@@ -2,9 +2,9 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1086
-    gateway.networking.k8s.io/bundle-version: v0.6.0-dev
-    gateway.networking.k8s.io/channel: experimental
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1538
+    gateway.networking.k8s.io/bundle-version: v0.6.2
+    gateway.networking.k8s.io/channel: standard
   creationTimestamp: null
   name: gatewayclasses.gateway.networking.k8s.io
 spec:
@@ -34,6 +34,9 @@ spec:
       name: Description
       priority: 1
       type: string
+    deprecated: true
+    deprecationWarning: The v1alpha2 version of GatewayClass has been deprecated and
+      will be removed in a future release of the API. Please upgrade to v1beta1.
     name: v1alpha2
     schema:
       openAPIV3Schema:
@@ -87,7 +90,7 @@ spec:
                   resource, i.e. ConfigMap, or an implementation-specific custom resource.
                   The resource can be cluster-scoped or namespace-scoped. \n If the
                   referent cannot be found, the GatewayClass's \"InvalidParameters\"
-                  status condition will be true. \n Support: Custom"
+                  status condition will be true. \n Support: Implementation-specific"
                 properties:
                   group:
                     description: Group is the group of the referent.
@@ -135,7 +138,7 @@ spec:
                 default:
                 - lastTransitionTime: "1970-01-01T00:00:00Z"
                   message: Waiting for controller
-                  reason: Waiting
+                  reason: Pending
                   status: Unknown
                   type: Accepted
                 description: "Conditions is the current status from the controller
@@ -145,13 +148,14 @@ spec:
                   description: "Condition contains details for one aspect of the current
                     state of this API Resource. --- This struct is intended for direct
                     use as an array at the field path .status.conditions.  For example,
-                    type FooStatus struct{     // Represents the observations of a
-                    foo's current state.     // Known .status.conditions.type are:
-                    \"Available\", \"Progressing\", and \"Degraded\"     // +patchMergeKey=type
-                    \    // +patchStrategy=merge     // +listType=map     // +listMapKey=type
-                    \    Conditions []metav1.Condition `json:\"conditions,omitempty\"
-                    patchStrategy:\"merge\" patchMergeKey:\"type\" protobuf:\"bytes,1,rep,name=conditions\"`
-                    \n     // other fields }"
+                    \n \ttype FooStatus struct{ \t    // Represents the observations
+                    of a foo's current state. \t    // Known .status.conditions.type
+                    are: \"Available\", \"Progressing\", and \"Degraded\" \t    //
+                    +patchMergeKey=type \t    // +patchStrategy=merge \t    // +listType=map
+                    \t    // +listMapKey=type \t    Conditions []metav1.Condition
+                    `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\"
+                    protobuf:\"bytes,1,rep,name=conditions\"` \n \t    // other fields
+                    \t}"
                   properties:
                     lastTransitionTime:
                       description: lastTransitionTime is the last time the condition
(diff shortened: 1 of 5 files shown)


candita:
> NAME  TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector
>     route_metrics_test.go:386: failed to update route: Operation cannot be fulfilled on routes.route.openshift.io "route-rs-foo-label": the object has been modified; please apply your changes to the latest version and try again
> NAME  TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector
>     panic.go:522: deleted ingresscontroller ic-rs-metrics-test

/test e2e-azure-operator

candita:
Installation failure on an optional test
/test e2e-aws-ovn-single-node

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -63,12 +61,11 @@ func (r *reconciler) ensureServiceMeshControlPlane(ctx context.Context, gatewayc
...
-		"PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER":   "true",
-		"PILOT_ENABLE_GATEWAY_API_STATUS":                  "true",
-		"PILOT_GATEWAY_API_CONTROLLER_NAME":                OpenShiftGatewayClassControllerName,
-		"PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS":           OpenShiftDefaultGatewayClassName,
-		"PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER": "false",
+		"PILOT_ENABLE_GATEWAY_CONTROLLER_MODE":   "true",
+		"PILOT_GATEWAY_API_CONTROLLER_NAME":      OpenShiftGatewayClassControllerName,
+		"PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS": OpenShiftDefaultGatewayClassName,
+		// Istio will only create the default gateway class if this is true.
+		"PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER": "true",
```
Miciah, line 68: In [the workflow from the EP](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/gateway-api-with-cluster-ingress-operator.md#workflow-description), the cluster-admin creates the gatewayclass, not Istio.

candita: I can update that to say "Istio will only allow ...."

candita: For reference: https://github.com/maistra/istio/blob/d8f16bb9d7e2671cf9bf70b1cf4cd314ced01ed0/pilot/pkg/features/pilot.go#L422

Miciah: Hm, so OSSM doesn't reconcile gatewayclasses _at all_ without `PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER=true`?  Is that a regression from OSSM 2.4?  Then it seems like are options are (1) no Gateway API support at all or (2) cluster-admin and OSSM clash over ownership of the default gatewayclass object.  Option 2 is better if we have no other option.

candita: Yes, it is a regression from OSSM 2.4.  I could ask OSSM to try to fix it in a 2.5 patch, but they don't consider this a bug.  I would like to get this into the next release if possible, if only to let dev preview users play with OSSM 2.5 as-is, but it doesn't add much to the mix in terms of Gateway API.

candita:
Was this an issue?
> Querying https://api.integration.openshift.com/api/upgrades_info/graph?channel=candidate-4.16&arch=amd64 for version 4.16.0-0.ci.test-2024-04-19-184019-ci-op-q4j54j1m-latest
> Version 4.16.0-0.ci.test-2024-04-19-184019-ci-op-q4j54j1m-latest is not available in candidate-4.16
> Cluster cannot query OpenShift Update Service (OSUS/Cincinnati)

/test e2e-gcp-operator

candita:
> level=error msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failure applying terraform for "cluster" stage: error applying Terraform configs: failed to apply Terraform: exit status 1
> level=error
> level=error msg=Error: waiting for EC2 Instance (i-071625d5fd2353bd2) create: couldn't find resource (21 retries)
> 
/test e2e-aws-operator

pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
gcs278, line 1: I tested this with our current DevPreview blog instructions: https://github.com/openshift/network-edge-tools/blob/main/docs/blogs/EnhancedDevPreviewGatewayAPI/GettingStarted.md.

And generally, everything looks good. However, the some blog instructions are now broken after this PR, for example:
- [OSSM will create two new Istio deployments in the openshift-ingress namespace](https://github.com/openshift/network-edge-tools/blob/main/docs/blogs/EnhancedDevPreviewGatewayAPI/GettingStarted.md?plain=1#L73) we don't create the default `istio-ingressgateway` deployment, so this command is incorrect now.
- The automated deployment and service name changed from `demo-gateway` to `demo-gateway-openshift-default`, some [commands](https://github.com/openshift/network-edge-tools/blob/main/docs/blogs/EnhancedDevPreviewGatewayAPI/GettingStarted.md?plain=1#L131) will fail
- The ready condition doesn't exist, so the `Wait for the gateway deployment to be ready:` [command](https://github.com/openshift/network-edge-tools/blob/main/docs/blogs/EnhancedDevPreviewGatewayAPI/GettingStarted.md?plain=1#L234) will fail

Should we maintain our DevPreview blog instructions? Or let them go stale? Either way, I'm good with proceeding with this bump.

gcs278:
Looks good to me.
/approve

pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
candita: I noticed a couple of these too and made some notes on them.  The DevPreview instructions are specific to a release.  Hopefully we make enough progress with this release that we can issue an update in OpenShift docs.

candita:
Failures in e2e-hypershift don't seem related to this change:

> --- FAIL: TestCreateCluster (2454.22s)
>     --- PASS: TestCreateCluster/ValidateHostedCluster (1137.50s)
>         --- PASS: TestCreateCluster/ValidateHostedCluster/EnsureNodeCountMatchesNodePoolReplicas (0.09s)
>         --- PASS: TestCreateCluster/ValidateHostedCluster/EnsureNoCrashingPods (0.03s)
>         --- PASS: TestCreateCluster/ValidateHostedCluster/EnsureGuestWebhooksValidated (5.06s)
>         --- PASS: TestCreateCluster/ValidateHostedCluster/EnsureNodeCommunication (0.13s)
>     --- FAIL: TestCreateCluster/Main (765.27s)
>         --- FAIL: TestCreateCluster/Main/break-glass-credentials (600.41s)
>             --- FAIL: TestCreateCluster/Main/break-glass-credentials/independent_signers (600.41s)

/retest-required

pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml
candita: @gcs278 Yes, that should be probably be updated if there's a way to do it.

PR #1021, opened 2024-01-31
Remove http port 80 from LB service

Miciah:
* `pkg/manifests/assets/router/service-cloud.yaml`: Remove http port.

---

This is just a test PR to see what breaks.

### DIFF
--- a/pkg/manifests/assets/router/service-cloud.yaml
+++ b/pkg/manifests/assets/router/service-cloud.yaml
@@ -16,10 +16,6 @@ spec:
   externalTrafficPolicy: Local
   internalTrafficPolicy: Cluster
   ports:
-  - name: http
-    protocol: TCP
-    port: 80
-    targetPort: http
   - name: https
     protocol: TCP
     port: 443


Miciah:
[e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1021/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1752692999373983744) because `TestAWSELBConnectionIdleTimeout`, `TestManagedDNSToUnmanagedDNSIngressController`, `TestUnmanagedDNSToManagedDNSInternalIngressController`, `TestUnmanagedDNSToManagedDNSIngressController`, and `TestCanaryRoute` failed.  

* `TestAWSELBConnectionIdleTimeout`, `TestManagedDNSToUnmanagedDNSIngressController`, `TestUnmanagedDNSToManagedDNSInternalIngressController`, and `TestUnmanagedDNSToManagedDNSIngressController` use HTTP requests to connect to test routes and servers that the tests themselves set up, but the tests could just as well use edge-terminated routes and HTTPS requests.    
*  `TestCanaryRoute` likewise uses an HTTP request, because the canary route originally used an insecure route but was later changed to use an edge-terminated TLS route with `insecureEdgeTerminationPolicy: Redirect` (see <https://github.com/openshift/cluster-ingress-operator/pull/556>).  Although the canary route is intended for internal diagnostics only, there is a risk that someone is using it for unintended purposes and expects it to remain available via HTTP request from outside the cluster.  For that reason, it makes sense to test that it is reachable using HTTP when HTTP ingress is enabled.  However, it is perfectly reasonable to say that one can block HTTP but must then accept the consequence that the canary route will only be available by HTTPS.  

(On an unrelated note, the helper function `verifyInternalIngressController` is missing a call to `t.Helper()`, which makes the test output a little confusing when one tries to determine the point of the test failure.)

Miciah:
[e2e-aws-ovn failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1021/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn/1752694707080663040) because the following 4 tests failed:

* `[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should be able to connect to a service that is idled because a GET on the route will unidle it [Skipped:Disconnected] [Suite:openshift/conformance/parallel/minimal]`
* `[sig-network][Feature:Router][apigroup:operator.openshift.io] The HAProxy router should serve routes that were created from an ingress [apigroup:route.openshift.io] [Skipped:Disconnected] [Suite:openshift/conformance/parallel]`
* `[sig-network][Feature:Router][apigroup:operator.openshift.io] The HAProxy router should set Forwarded headers appropriately [Skipped:Disconnected] [Suite:openshift/conformance/parallel]`
* `[sig-network][Feature:Router][apigroup:operator.openshift.io] The HAProxy router should respond with 503 to unrecognized hosts [Skipped:Disconnected] [Suite:openshift/conformance/parallel]`

The idling test and unrecognized-hosts test could be rewritten to use HTTPS requests instead of HTTP.  The "Forwarded headers" test probably should be using HTTPS requests anyway to ensure that using TLS doesn't prevent the router from putting the actual client IP address in the `forwarded` header value (possible test gap).  The routes-created-from-ingress test highlights that blocking HTTP affects the Ingress API as well as the Route API; this isn't a surprising fact, but it may be worth noting calling it out explicitly in documentation.

PR #1036, opened 2024-03-26
NE-1688: Enable Azure MSI authentication

csrwng:
When config does not include a federated token file path, assume that we want to use managed service identity (MSI) which uses the cloud metadata endpoint to obtain a token for the operator.


bryan-cox:
/close
Closing since #1119 merged

PR #1034, opened 2024-03-24
NE-1674 : Adds logic to add service annotation `service.beta.kubernetes.io/aws-load-balancer-eip-allocations`

miheer:
Adds logic to add service annotation service.beta.kubernetes.io/aws-load-balancer-eip-allocations to the load balancer service.

PR #1046, opened 2024-04-18
NE-1531: AWS Subnet Selection

gcs278:
Adds implementation for the new IngressController AWS subnet API. It takes the subnets specified on the IngressController and propagates them to the service.beta.kubernetes.io/aws-load-balancer-subnets annotation on the LB-type Service.

This design requires a cluster admin to manually delete the LB-type Service in order to effectuate the subnet update. Once the subnets are updated in the IngressController spec, a LoadBalancerProgressing=True condition will be added to notify the cluster admin to delete the Service to effectuate the subnet change.

This change is being introduced under the Tech Preview IngressControllerLBSubnetsAWS feature gate and will be later promoted to GA.

Epic: https://issues.redhat.com/browse/NE-705
EP: https://github.com/openshift/enhancements/pull/1595
API: https://github.com/openshift/api/pull/1841


gcs278:
TODO need to fix verify job

gcs278:
Appears like infrastructure issues:
/retest-required

gcs278:
TestClientTLS related errors
/test e2e-aws-operator

gcs278:
cluster install failure

gcs278:
We will most likely need to vendor an API CEL fix https://github.com/openshift/api/pull/1961
/hold

gcs278:
/unhold
https://github.com/openshift/api/pull/1961 has been vendored

gcs278:
Andrey offered to help review
/assign @alebedev87

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+	}
+
+	// If they both are non-nil, compare the length first, then do a more detailed comparison if needed.
+	if len(subnets1.Names) != len(subnets2.Names) || len(subnets1.IDs) != len(subnets2.IDs) {
+		return false
+	}
+
+	// Create maps to track the IDs from each subnet object for comparison.
+	subnetIDMap1 := make(map[operatorv1.AWSSubnetID]bool)
+	subnetIDMap2 := make(map[operatorv1.AWSSubnetID]bool)
```
alebedev87, line 969: To save some bytes:
```suggestion
	subnetIDMap1 := make(map[operatorv1.AWSSubnetID]struct{})
	subnetIDMap2 := make(map[operatorv1.AWSSubnetID]struct{})
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,17 +776,88 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+			}
+			paramsFieldName = "classicLoadBalancer"
+		}
+		if !compareAWSSubnets(wantSubnets, haveSubnets) {
+			// Generate JSON for the oc patch command as well as "pretty" json
+			// that will be used for a more human-readable error message.
+			haveSubnetsPatchJson := convertAWSSubnetListToPatchJson(haveSubnets, "null", "null")
+			haveSubnetsPrettyJson := convertAWSSubnetListToPatchJson(haveSubnets, "{}", "[]")
+			wantSubnetsPrettyJson := convertAWSSubnetListToPatchJson(wantSubnets, "{}", "[]")
+			err := fmt.Errorf("The IngressController subnets were changed from %q to %q.", haveSubnetsPrettyJson, wantSubnetsPrettyJson)
```
alebedev87, line 811: Can this be a simple string (similar to the patch command) just to avoid the reset of `err` which may look like an override?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,17 +776,88 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+	namesJSON, idsJSON := emptySubnetSliceValue, emptySubnetSliceValue
+
+	// Marshal names and IDs separately.
+	if len(subnets.Names) > 0 {
+		var err error
+		namesJSONBytes, err := json.Marshal(subnets.Names)
+		if err != nil {
+			log.Error(err, "error marshaling subnet names")
+			return ""
+		}
```
alebedev87, line 843:
```suggestion
		if namesJSONBytes, err := json.Marshal(subnets.Names); err != nil {
			log.Error(err, "error marshaling subnet names")
			return ""
		}
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,17 +776,88 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+		if err != nil {
+			log.Error(err, "error marshaling subnet IDs")
+			return ""
+		}
+		idsJSON = string(idsJSONBytes)
+	}
+
+	// Combine the JSON of subnet names and IDs.
+	resultJSON := fmt.Sprintf("{\"ids\":%s,\"names\":%s}", idsJSON, namesJSON)
+	return resultJSON
```
alebedev87, line 858:
```suggestion
	return fmt.Sprintf("{\"ids\":%s,\"names\":%s}", idsJSON, namesJSON)
```

pkg/operator/controller/ingress/status.go
```diff
@@ -925,6 +941,35 @@ func computeLoadBalancerProgressingStatus(ic *operatorv1.IngressController, serv
...
 }
 
+// updateIngressControllerAWSSubnetStatus mutates the provided IngressController object to
+// sync its status to the effective subnets on the LoadBalancer-type service.
+func updateIngressControllerAWSSubnetStatus(ic *operatorv1.IngressController, service *corev1.Service) {
+	// Set the subnets status based on the actual service annotation and the load balancer type,
+	// as NLBs and CLBs have separate subnet configuration fields.
+	switch getAWSLoadBalancerTypeInStatus(ic) {
+	case operatorv1.AWSNetworkLoadBalancer:
+		// NetworkLoadBalancerParameters will be initialized by setDefaultPublishingStrategy
```
alebedev87, line 951: Then they cannot be `nil`, no? Do we need the `if` above?

pkg/operator/controller/ingress/status.go
```diff
@@ -799,6 +802,19 @@ func IngressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
 		if !reflect.DeepEqual(a.EndpointPublishingStrategy.LoadBalancer.AllowedSourceRanges, b.EndpointPublishingStrategy.LoadBalancer.AllowedSourceRanges) {
 			return false
 		}
+		if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+			a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+			if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters != nil {
+				if !reflect.DeepEqual(a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters.Subnets, b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters.Subnets) {
```
alebedev87, line 808: Why not use your new comparison function?

alebedev87 (commented):
First round, no tests were reviewed. Just nit picks so far.

pkg/operator/controller/ingress/load_balancer_service.go
miheer: or may be add  errs = append(errs, err) here as well?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,17 +776,88 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+		var err error
+		namesJSONBytes, err := json.Marshal(subnets.Names)
+		if err != nil {
+			log.Error(err, "error marshaling subnet names")
+			return ""
+		}
+		namesJSON = string(namesJSONBytes)
+	}
+	if len(subnets.IDs) > 0 {
+		var err error
```
miheer, line 847: why do need to declare var err because we already have initialised it down err := ?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,17 +776,88 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+		idsJSONBytes, err := json.Marshal(subnets.IDs)
+		if err != nil {
+			log.Error(err, "error marshaling subnet IDs")
+			return ""
+		}
```
miheer, line 852: idsJSONBytes, err := json.Marshal(subnets.IDs); err != nil {
			log.Error(err, "error marshaling subnet IDs")
			return ""
		}

miheer (changes requested):

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+// JoinAWSSubnets joins an AWS Subnets object into a string seperated by sep.
+func JoinAWSSubnets(subnets *operatorv1.AWSSubnets, sep string) string {
+	if subnets == nil {
+		return ""
+	}
+	joinedSubnets := ""
+	subnetCount := 0
+	for _, subnet := range subnets.IDs {
+		if subnetCount > 0 {
+			joinedSubnets += sep
```
miheer, line 1015: Can't we use `strings.Join` function ?  
joinedSubnetID = strings.Join(subnets.IDs, sep)
joinedSubnetNames = strings.Join(subnets.Names, sep)
joinedSubnets = joinedSubnetID + joinedSubnetNames

miheer (changes requested):

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+	return joinedSubnets
+}
+
+// getAWSLoadBalancerTypeInStatus gets the AWS Load Balancer Type reported in the status.
+func getAWSLoadBalancerTypeInStatus(ic *operatorv1.IngressController) operatorv1.AWSLoadBalancerType {
+	if ic.Status.EndpointPublishingStrategy != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+		return ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type
```
miheer, line 1036: Type is [required](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L536) so may be we don't need to check if it is nil `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type != nil` ?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+}
+
+// getAWSNetworkLoadBalancerParametersInSpec gets the NetworkLoadBalancerParameter struct
+// defined in the IngressController spec.
+func getAWSNetworkLoadBalancerParametersInSpec(ic *operatorv1.IngressController) *operatorv1.AWSNetworkLoadBalancerParameters {
+	if ic.Spec.EndpointPublishingStrategy != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+		return ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters
```
miheer, line 1060: I think you need to check nil for `ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters != nil` this field is [optional](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L548) and is a pointer .

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+}
+
+// getAWSClassicLoadBalancerParametersInStatus gets the ClassicLoadBalancerParameter struct
+// reported in the IngressController status.
+func getAWSClassicLoadBalancerParametersInStatus(ic *operatorv1.IngressController) *operatorv1.AWSClassicLoadBalancerParameters {
+	if ic.Status.EndpointPublishingStrategy != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+		return ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters
```
miheer, line 1072: I think you need to check nil for ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil this field is [optional](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L542) and is a pointer .

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+}
+
+// getAWSNetworkLoadBalancerParametersInStatus gets the NetworkLoadBalancerParameter struct
+// reported in the IngressController status.
+func getAWSNetworkLoadBalancerParametersInStatus(ic *operatorv1.IngressController) *operatorv1.AWSNetworkLoadBalancerParameters {
+	if ic.Status.EndpointPublishingStrategy != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+		return ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters
```
miheer, line 1084: I think you need to check nil for ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters != nil this field is [optional](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L548) and is a pointer .

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -812,3 +908,180 @@ func loadBalancerSourceRangesMatch(ic *operatorv1.IngressController, current *co
...
+}
+
+// getAWSClassicLoadBalancerParametersInSpec gets the ClassicLoadBalancerParameter struct
+// defined in the IngressController spec.
+func getAWSClassicLoadBalancerParametersInSpec(ic *operatorv1.IngressController) *operatorv1.AWSClassicLoadBalancerParameters {
+	if ic.Spec.EndpointPublishingStrategy != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil {
+		return ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters
```
miheer, line 1048: I think you need to check nil for ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil this field is [optional](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L542) and is a pointer .

pkg/operator/controller/ingress/status.go
miheer: `Type` is `required` but NetworkLoadBalancerParameters is `optional` and of type pointer so i think it is better to check nil for it from my understanding.

miheer (changes requested):

gcs278 (commented):
Thanks @miheer and @alebedev87 for the code reviews.

pkg/operator/controller/ingress/status.go
```diff
@@ -925,6 +941,37 @@ func computeLoadBalancerProgressingStatus(ic *operatorv1.IngressController, serv
...
+		}
+	case operatorv1.AWSClassicLoadBalancer:
+		// ClassicLoadBalancerParameters should be initialized by setDefaultPublishingStrategy
+		// when an IngressController is admitted, so we don't need to initialize here.
+		if clbParams != nil {
+			clbParams.Subnets = getSubnetsFromServiceAnnotation(service)
+		}
+		// Clear NLB status as the IngressController is now using a CLB.
+		if nlbParams != nil {
+			nlbParams.Subnets = nil
```
miheer, line 970: I think we shall check for nlbParams.Subnets != nil

pkg/operator/controller/ingress/status.go
```diff
@@ -925,6 +941,37 @@ func computeLoadBalancerProgressingStatus(ic *operatorv1.IngressController, serv
...
+	switch getAWSLoadBalancerTypeInStatus(ic) {
+	case operatorv1.AWSNetworkLoadBalancer:
+		// NetworkLoadBalancerParameters should be initialized by setDefaultPublishingStrategy
+		// when an IngressController is admitted, so we don't need to initialize here.
+		if nlbParams != nil {
+			nlbParams.Subnets = getSubnetsFromServiceAnnotation(service)
+		}
+		// Clear CLB status as the IngressController is now using a NLB.
+		if clbParams != nil {
+			clbParams.Subnets = nil
```
miheer, line 960: I think we shall check for clbParams.Subnets != nil

miheer (changes requested):

pkg/operator/controller/ingress/controller.go
```diff
@@ -566,6 +567,11 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 						changed = true
 					}
 				}
+				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+					if statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters == nil {
+						statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters = &operatorv1.AWSNetworkLoadBalancerParameters{}
+					}
```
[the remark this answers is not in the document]
alebedev87: I think I agree with the status being updated for the state of the managed objects instead of the spec. Also, I agree that it may easily become messy and the source of truth may be confused.

This made me think of `updateIngressControllerAWSSubnetStatus` function from `status.go`. It relies on the LB type read from the status which in turn was set from the spec (`setDefaultPublishingStrategy`) .  May this cause a problem taking into account that we clean the other type's params (CLB's if NLB is set and vice versa)? I suppose that we can turn out into a situation when the type changed in the status but the service object failed to be updated - in that case we can prematurely clean up the LB params. Should we read the service type from the annotation too?

gcs278:
/hold
Found a bug with switching LB type. `loadBalancerServiceChanged` doesn't change the subnets annotation if it's already set, which is good for our delayed effecutation pattern, but bad when we switch LB types because we want it to effectuate immediately. I need some time to figure this one out. Either:
1. Special logic in `loadBalancerServiceChanged` to effectuate the service annotation when LB Type changes. This is probably the right answer because changing LB type allows for immediate effectuation of subnets. It feel messy though.
2. Delay effectuation when switching LB Type because subnets are also change.

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	awsLBSubnetAnnotation = "service.beta.kubernetes.io/aws-load-balancer-subnets"
+)
+
+// TestAWSLBSubnets creates an IngressController with both the public and private cluster subnets for its LB-type
```
miheer, line 33: Have we added a test to check what happens when an invalid or non-exisiting subnet is provided in the IC CR ?

miheer (changes requested):

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+		return field.Required(field.NewPath("providerSpec", "value"), "a value must be provided")
+	}
+
+	if err := json.Unmarshal(m.Spec.Template.Spec.ProviderSpec.Value.Raw, &providerSpec); err != nil {
+		return field.Invalid(field.NewPath("providerSpec", "value"), providerSpec, err.Error())
+	}
+	return nil
+}
+
+// effectuateIngressControllerSubnets manually effectuates updated IngressController subnets by
```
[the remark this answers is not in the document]
gcs278: > why are manually deleting ?

Because that's something a customer would do after following the condition instructions, and I think should test End-to-End, i.e. make sure that after we observe `LoadBalancerProgressing`, we delete the service, ensure it is populated with the new subnets, and make sure the IngressController status is accurate.

> Should not we just check the progressing status to true and check the message in the IC status which mentions to delete manually or revert back using the patch command ?

We do check for `LoadBalancerProgressing`. I suppose we could try to parse the message to be absolutely certain. I did consider that, but:
1. Felt a little messy/fragile to try to parse the message 
2. We are only changing the subnet, so it's _pretty_ reasonable to assume `LoadBalancerProgressing=true` is because of this subnet update.
3. `TestScopeChange` test doesn't parse the message either, I was modeling off of that E2E test.

So I'm a bit indifferent. I'm open to parsing the message if you feel strongly we should, just want to make sure we are on the same page before refactoring more.

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+		t.Fatalf("failed to effectuate the ingresscontroller subnets: %v", err)
+	}
+
+	// Verify we can still reach the CLB with the provided public subnets.
+	t.Logf("verifying external connectivity for ingresscontroller %q using an CLB with specified public subnets", ic.Name)
+	if err := verifyExternalIngressController(t, externalTestPodName, testHostname, testHostname, 10*time.Minute); err != nil {
+		t.Fatalf("failed to verify connectivity with workload with hostname %s using external client: %v", testHostname, err)
+	}
+
+	// Now, update the IngressController to change to use the private subnets, so it can't connect externally.
```
alebedev87, line 124: Does it test something additionally to what has already tested when we passed from "IngressController with no subnets" to "IngressController with public subnets"? To me it seems to be the same code path, no?

I tried to think of a potential way to shorten this e2e. Would the below cover all the cases?

- Create IngressController with public subnets, wait for the LB to be provisioned, verifyExternal
- Update the IngressController from CLB to NLB removing 1 public subnet (I suppose we still will have `>=1`), verifyExternal
- Remove subnets with auto delete and verify the annotation is removed (just as now)

Also, these scenarios will test the update of the subnets without the need to wait for "not connect" to private subnets. Therefore `verifyExternalIngressController` can be left unchanged.

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("error updating the %q service: %v", controller.LoadBalancerServiceName(ic), err)
+	}
+	return nil
+}
+
+// verifyIngressControllerSubnetStatus verifies that the IngressController subnet status fields are
+// equal to the appropriate subnet spec fields when an IngressController is in a Progressing=False state.
```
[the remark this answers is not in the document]
gcs278: We aren't, I listed this just as a function "pre-condition", i.e. this is a restriction or scenario to use this function.

But it's a good point, we can enforce Progressing=False just to be absolutely sure. The function would fail anyways, but might be helpful to fail quicker. I'll add.

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+					Subnets: publicSubnets,
+				},
+			},
+		}
+	}); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+
+	// Effectuate the updated IngressController subnets.
+	if err := effectuateIngressControllerSubnets(t, ic, publicSubnets); err != nil {
```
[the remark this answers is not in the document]
gcs278: See https://github.com/openshift/cluster-ingress-operator/pull/1046#discussion_r1688356584, we do check for `LoadBalancerProgressing`.

alebedev87 (commented):
Another round for the tests.

test/e2e/lb_subnets_test.go
gcs278: We have a test which provides a valid private subnet to a public IngressController, which breaks the IngressController, but as far as non-existing subnets, we don't.

It's a good idea and pretty easy to do, I'll add one.

test/e2e/lb_subnets_test.go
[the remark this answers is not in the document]
miheer: OK in original PR I did this : https://github.com/openshift/cluster-ingress-operator/pull/1034/commits/86f94a54e77f46496565b6ec0b38526ab08731e0#diff-160d46f77f47e8e0d35737b019c27a6e4a6f7e7eae4f6e85c0dcdb43b20a2379R317-R365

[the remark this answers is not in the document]
miheer: make sure that after we observe LoadBalancerProgressing -- sounds good

test/e2e/lb_subnets_test.go
miheer: in my orig PR 
invalid EIP : https://github.com/openshift/cluster-ingress-operator/pull/1034/commits/86f94a54e77f46496565b6ec0b38526ab08731e0#diff-160d46f77f47e8e0d35737b019c27a6e4a6f7e7eae4f6e85c0dcdb43b20a2379R445
non-existent EIP: https://github.com/openshift/cluster-ingress-operator/pull/1034/commits/86f94a54e77f46496565b6ec0b38526ab08731e0#diff-160d46f77f47e8e0d35737b019c27a6e4a6f7e7eae4f6e85c0dcdb43b20a2379R558

gcs278 (commented):
Thanks @alebedev87 and @miheer for the detailed reviews.

The major update here, is the fix for when switching LBType. Before, this fix switching the LBType erroneously kept the same subnets from the previous LBType, and subsequently messed up the IngressController subnet status, and state of the IngressController got a bit wacky with Status not matching spec correctly. 

Now, with a LBType change, the subnet annotation will also be auto-effectuated.This has been tested (and E2E updated to reflect). AWS is already recreating the LBs, so it can safely handle new subnets. This is a design choice I made that should be reviewed. See change in `loadBalancerServiceChanged`.

The alternative, don't auto-effectuate a LBType change with Subnets that also change, felt like it would be really quite messy to implement. The way I look at it, LBType updates are already an invasive change, so we are just "piggybacking" off of that. Comments welcome.

gcs278:
/hold cancel
https://github.com/openshift/cluster-ingress-operator/pull/1046#issuecomment-2244030050 has been resolved as described in https://github.com/openshift/cluster-ingress-operator/pull/1046#pullrequestreview-2194592829

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
gcs278: > May this cause a problem taking into account that we clean the other type's params (CLB's if NLB is set and vice versa)? 

Okay, so this comment led me to attempt to enumerate what would happen, which made me realize there was a bug when I tried it out. We did have E2E for this, but E2E was passing "falsely" for this scenario. I have now fixed this issue and documented here: https://github.com/openshift/cluster-ingress-operator/pull/1046#issuecomment-2244030050 and https://github.com/openshift/cluster-ingress-operator/pull/1046#pullrequestreview-2194592829. This might be more than what you asked for.

This is *now* what happens when you change LB Type and NLB/CLB have different subnet values:
1. The IngressController triggers a reconcile loop, [readmits](https://github.com/openshift/cluster-ingress-operator/blob/a180755f2f4d082b3236ef88278cf360c52ea32d/pkg/operator/controller/ingress/controller.go#L288) the IngressController which updates the status to only what is set in `setDefaultPublishingStrategy` which "cleans" the status like you said.
2. It requeues, and runs reconcile again, which now runs `ensureIngressController`.
3. Now the `ensureLoadBalancerService` will change the LB Type, but also change the subnets as well per my [most recent update](https://github.com/openshift/cluster-ingress-operator/pull/1046/files#diff-5ee46ab285ecdbf9373a451d224299f443d5821af04eb5d0a7cee19ec0a78ed8R668-R672). Before, it didn't change the subnets, which caused many problems.
4. Then `syncIngressControllerStatus` runs, and updates the status with the correct subnets.

Hope that answers your question. Just in writing this out, I found another minor issue: the call to `computeLoadBalancerProgressingStatus` doesn't use the `updated` status, so it actually adds a `LoadBalancerProgressing` status momentarily because the subnet status isn't updated until `syncIngressControllerStatus` is finished:
https://github.com/openshift/cluster-ingress-operator/blob/a180755f2f4d082b3236ef88278cf360c52ea32d/pkg/operator/controller/ingress/status.go#L86

That might be okay, as it's temporary, but I need to look into if it's easily fixable.

> I suppose that we can turn out into a situation when the type changed in the status but the service object failed to be updated - in that case we can prematurely clean up the LB params. Should we read the service type from the annotation too?

Hmm. It's a valid point. `status.endpointPublishingStrategy.loadBalancer.providerParameters.aws.type` will be [updated regardless](https://github.com/openshift/cluster-ingress-operator/blob/3dc5456f0492e5a9bd7ec1de974813e58059101c/pkg/operator/controller/ingress/controller.go#L546) if the service failed to update. However, other logic will such as [IsProxyProtocolNeeded](https://github.com/openshift/cluster-ingress-operator/blob/a180755f2f4d082b3236ef88278cf360c52ea32d/pkg/operator/controller/ingress/controller.go#L1183) will break too if that happens.

I personally think the LB Type _should_ use the annotation for the real value, however, given it's an outstanding issue, I think we should solve it in another PR/bug/solution.

gcs278:
/hold
I need to fix `gateway_api_test` after the rebase and, after some testing, I realized one major drawback to auto-effectuating subnets on LB type change: though it can indeed take new subnets, and AWS reprovisions the LB behind-the-scenes, it doesn't remove/update the service's `status.loadBalancer.ingress[].hostname` field.

What this means is: if you switch LB Type, and the subnets for the new LB type are invalid, you will NOT get any signal/condition that they are invalid. This is explained in the current [enhancement PR](https://github.com/openshift/enhancements/pull/1595/files#diff-f8f6da2366b6e9f9eede942d6a878c8043690acb03727fad6bbf71d37c95fd92R289). if they 

That's a -1 to auto-effectuating on LB Type change. Maybe it's something we can live with for this edge case, but I want to explore the alternative of requiring recreation of the service upon LB Type change WITH subnets changing. That feels safer to me, but may be difficult to do.

gcs278:
Some notes:

I've explored what it would take to delay the effectuation of a LB Type change WITH a subnet change. Unfortunately, it's not pretty. If we don't automatically effectuate the LB Type change, then a couple pieces of CIO logic start to break down:
- The `status.endpointPublishingStrategy.loadBalancer.providerParameters.aws.type` field enters a new "scenario"  where the status doesn't match the actual LB Type on the service.
   - On it's own, this seems somewhat problematic. This produces a real scenario where the status != the actual state.
- This starts to cause problems for: 1. The code in this PR that relies on the status LB Type to be accurate and 2. `IsProxyProtocolNeeded`.
-  It can mostly be fixed by changing the logic to NOT examine the status LB Type, but instead using a helper function to examine the _actual_ Service Annotation.
- However, while trying to refactor `IsProxyProtocolNeeded` results in requiring to pass in a service so it can examine the annotations, I got stuck because `ensureRouterDeployment` calls `IsProxyProtocolNeeded` and needs the LB service object, but `ensureRouterDeployment` runs before `ensureLoadBalancerService` because `ensureLoadBalancerService` needs the deployment ref from `ensureRouterDeployment`. I couldn't reasonable figure out an easy fix to creating/getting the service first so `ensureRouterDeployment` could pass it down to `IsProxyProtocolNeeded`
   - It's a chicken-and-the-egg-type of problem.
- Well, what if we just set the LB Type status field to the _actual_ value? That starts to cause other problems because `desiredLoadBalancerService` relies on the IC status set from `setDefaultPublishingStrategy`, which is called _before_ `desiredLoadBalancerService`. Setting the LB Type would happen in `syncIngressControllerStatus` which happens _after_ `desiredLoadBalancerService`.
  - This could be overcome by refactoring to examining the LB Type in the spec.
  - However, it still doesn't solve the `IsProxyProtocolNeeded` chicken-and-the-egg problem

_All this to say:_ it's unfortunate the CIO's status, specifically LB Type, doesn't currently reflect the actual LB Type on the service (it's hardcoded from spec) and that the CIO has built it's design around using the status before it has a chance to reflect the actual values. Refactoring that feels much bigger & more risky than what I've signed up for here. Could be considered technical debt.

In my personal opinion, to keep on track with this feature, we keep my current implementation of auto-effectuation on LB Type change, accept the risk that switching LB Types might cause invalid subnets to break an IngressController.

The somewhat good news is: when you switch LB Type to invalid subnets, your LB will _totally break_ (that's a valid signal right?). For example NLB (with valid subnets) to CLB (with invalid subnets) looks like:
```
$ curl a6c93ca72d23d4b11849717b1cc08ebc-2c13aacd3f005001.elb.us-east-2.amazonaws.com
curl: (56) Recv failure: Connection reset by peer
```

And CLB (with valid subnets) to NLB (with invalid subnets) looks like:
```
$ curl a36743f6198b843fd9d9c67c181d892d-1989938852.us-east-2.elb.amazonaws.com
<html><body><h1>400 Bad request</h1>
Your browser sent an invalid request.
</body></html>
```

Presumably because the LB didn't go ready after transitioning. This is different from an "inline" subnet replacement (no LB Type change). The AWS CCM will just refuse to reconcile the LB and leave it in an unmanaged state until you fix the subnets.

gcs278:
Rebase has been fixed, and I feel pretty confident in the design of current revision of handling a LB Type & Subnet change, as it seems the simplest.

/unhold

gcs278:
`TestCanaryRouteClearsSpecHost` [failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1046/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1816212118429503488), need to be investigated 
/test e2e-aws-operator

`TestMTLSWithCRLs` and `TestUnmanagedDNSToManagedDNSInternalIngressController` TestUnmanagedDNSToManagedDNSInternalIngressController, need to be investigated
/test e2e-gcp-operator

test/e2e/util_test.go
```diff
@@ -939,3 +937,56 @@ func createNamespace(t *testing.T, name string) *corev1.Namespace {
 
 	return ns
 }
+
+// isFeatureGateEnabled returns a boolean for it the provided feature gate is enabled
+// and an error if unable to get feature gate.
+func isFeatureGateEnabled(featureGateName configv1.FeatureGateName) (bool, error) {
```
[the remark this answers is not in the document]
alebedev87: > This helps when a caller wants it check the type of error.

For this case, I agree with a simple `error`. But if there is no such a need, I think  passing `testing.T` brings the native integration and turns an "all purpose" util function into a a test helper. Plus we can save some lines for the error checking. At the end, `testing.T.Helper()` was added to the standard library not without a reason.

> Side note: this is also duplicated in https://github.com/openshift/cluster-ingress-operator/pull/1023, so whichever PR merges first, adds this function.

Ok, I'm not going to insist on this one. It's a nit pick anyway.

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -83,17 +123,19 @@ func Test_desiredLoadBalancerService(t *testing.T) {
 	}
 	testCases := []struct {
 		description                   string
-		strategy                      *operatorv1.EndpointPublishingStrategy
+		stategySpec                   *operatorv1.EndpointPublishingStrategy
+		strategyStatus                *operatorv1.EndpointPublishingStrategy
```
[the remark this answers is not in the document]
gcs278: Right. Things are getting weird in the Ingress Operator.

test/e2e/util_test.go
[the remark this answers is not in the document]
gcs278: > At the end, testing.T.Helper() was added to the standard library not without a reason.

True. "All Purpose" vs. "Test Helper" is a good way to qualify it. I guess everything doesn't have to be "All Purpose".

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+// easily accessible in any K8S API objects. However, by using the installer-created subnet
+// naming convention, we can generate the names of the public and private subnets.
+func getClusterSubnets() (public *operatorv1.AWSSubnets, private *operatorv1.AWSSubnets, err error) {
+	// Subnet Naming Convention: <clusterName>-subnet-[public|private]-<availabilityZone>
+	machineSets := &machinev1.MachineSetList{}
+	listOpts := []client.ListOption{
+		client.InNamespace("openshift-machine-api"),
+	}
+	if err := kclient.List(context.Background(), machineSets, listOpts...); err != nil {
+		return nil, nil, fmt.Errorf("failed to get machineSets: %v", err)
```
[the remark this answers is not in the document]
alebedev87: > I usually only put polls where I am waiting for an expected state or condition of that object.

Agree. I'm wondering about the featuregates and clusterversion now. Why do we need to poll there?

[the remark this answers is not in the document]
gcs278: Ah yea I see that actually, that's redundant. That code was copied around a couple PRs and I missed seeing that. I could remove, but eh, it's already merged and doesn't hurt anything.

gcs278:
I'm seeing a couple failures for `TestMTLSWithCRLs` [link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1046/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1816265773342003200). I do see another failure of this type in another PR, so I don't think it's related to my PR. I will try to look into it.
/test e2e-gcp-operator

gcs278:
Opened https://github.com/openshift/cluster-ingress-operator/pull/1110 to fix `TestMTLSWithCRLs` flakes. Simple mistake.

gcs278:
Last [push](https://github.com/openshift/cluster-ingress-operator/compare/fd6597b05a72bd9c24e7bc76e2c0ba9a5aee0935..cd850f370eb9b7dcfa1d2ab85abc5e60af4860b3) was some minor E2E refactoring.

gcs278:
`TestMTLSWithCRLs` failure

cluster-wide failure
/test e2e-gcp-operator

unrelated failure
/test e2e-gcp-ovn

gcs278:
`unmanaged_dns_test.go:175: DNSRecord zone expected to have status=Unknown but got status=True` failure. I will try to look into this, but unrelated:

`e2e-gcp-operator-ipi-deprovision-deprovision` failure:
/test e2e-gcp-operator

gcs278:
[Update](https://github.com/openshift/cluster-ingress-operator/compare/cd850f370eb9b7dcfa1d2ab85abc5e60af4860b3..36be87ad990e0fb2aa45965cbe8f698ffed3002e): Moved functions from my E2E test file to `util_test.go`, since Miheer was using them in his PR https://github.com/openshift/cluster-ingress-operator/pull/1109

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,358 @@
...
+
+const (
+	awsLBSubnetAnnotation = "service.beta.kubernetes.io/aws-load-balancer-subnets"
+)
+
+// TestAWSLBSubnets creates an IngressController with various subnets in AWS.
+// The test verifies the provisioning of the LB-type Service, confirms ingress connectivity, and
+// confirms that the service.beta.kubernetes.io/aws-load-balancer-subnets as well as the IngressController status is
+// correctly configured.
+func TestAWSLBSubnets(t *testing.T) {
```
miheer, line 37: @gcs278 I don't see the e2e test for unmanaged subnet annotation part.

miheer (changes requested):

test/e2e/util_test.go
```diff
@@ -939,3 +937,56 @@ func createNamespace(t *testing.T, name string) *corev1.Namespace {
...
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to get cluster feature gate: %v", err)
+	}
+
+	// Check if provided feature gate is enabled.
+	for _, fg := range clusterFeatureGate.Status.FeatureGates {
+		if fg.Version != desiredVersion {
+			continue
+		}
```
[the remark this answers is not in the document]
alebedev87: > I'm sure it's probably an edge case that David had in mind that we may never need, but better safe than sorry

Do you mind adding a link to David's code as the source for inspiration?

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,419 @@
...
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.ClassicLoadBalancerParameters.Subnets != nil
+}
+
+// getClusterSubnets populates public and private AWSSubnets objects with a list of
+// the public and private cluster subnets names. Unfortunately these subnets aren't
+// easily accessible in any K8S API objects. However, by using the installer-created subnet
+// naming convention, we can generate the names of the public and private subnets.
```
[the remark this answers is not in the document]
alebedev87: Overall I feel like there is quite some platform specific data which we could query from the OpenShift workload but it's not present. AWS VPC is another missing bit which may be helpful for some Red Hat operators.

alebedev87:
@gcs278:

> Found a bug with switching LB type. loadBalancerServiceChanged doesn't change the subnets annotation if it's already set, which is good for our delayed effecutation pattern, but bad when we switch LB types because we want it to effectuate immediately. I need some time to figure this one out

It took me some time but I acknowledged the problem. The two updates have different application patterns: the update corresponding to a managed annotation needs to be applied asap while the update corresponding to the loadbalancer recreation needs a manual deletion of the Kube service. I suppose that a similar problem existed before, for the simultaneous scope (`External`/`Internal`) and the LB type update. I think we don't update the Kube service with the new scope, while the LB type gets changed. Plus the status should be updated with instructions to recreate the service.

> The somewhat good news is: when you switch LB Type to invalid subnets, your LB will totally break (that's a valid signal right?). For example NLB (with valid subnets) to CLB (with invalid subnets) looks like:
> 
> ```
> $ curl a6c93ca72d23d4b11849717b1cc08ebc-2c13aacd3f005001.elb.us-east-2.amazonaws.com
> curl: (56) Recv failure: Connection reset by peer
> ```
> 
> And CLB (with valid subnets) to NLB (with invalid subnets) looks like:
> 
> ```
> $ curl a36743f6198b843fd9d9c67c181d892d-1989938852.us-east-2.elb.amazonaws.com
> <html><body><h1>400 Bad request</h1>
> Your browser sent an invalid request.
> </body></html>
> ```

I didn't manage to see this behavior with unmanaged (ingress operator scaled down) `router-default` service. 

Here is what I saw going from an active CLB with valid subnets to new NLB with invalid subnets. The CLB remained active and the service's ingress stayed untouched while the NLB was not provisioned:
```
# subnet "subnet-0bacbe7de35736bb0" does not exist
$ oc -n openshift-ingress get svc router-default -o yaml | yq .metadata.annotations
service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold: "2"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-interval: "5"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout: "4"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold: "2"
service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
service.beta.kubernetes.io/aws-load-balancer-subnets: subnet-0bacbe7de35736bb0
service.beta.kubernetes.io/aws-load-balancer-type: nlb
traffic-policy.network.alpha.openshift.io/local-with-fallback: ""

# note the CLB domain: elb.amazon.com
$ oc -n openshift-ingress get svc router-default -o yaml | yq .status
loadBalancer:
  ingress:
    - hostname: a853f70ef40f34f21a91967bb1d91c51-706519054.us-east-2.elb.amazonaws.com

$ curl -v a853f70ef40f34f21a91967bb1d91c51-706519054.us-east-2.elb.amazonaws.com
* Rebuilt URL to: a853f70ef40f34f21a91967bb1d91c51-706519054.us-east-2.elb.amazonaws.com/
* TCP_NODELAY set
* Connected to a853f70ef40f34f21a91967bb1d91c51-706519054.us-east-2.elb.amazonaws.com port 80 (#0)
> GET / HTTP/1.1
> Host: a853f70ef40f34f21a91967bb1d91c51-706519054.us-east-2.elb.amazonaws.com
> User-Agent: curl/7.61.1
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 503 Service Unavailable
< pragma: no-cache
< cache-control: private, max-age=0, no-cache, no-store
< content-type: text/html
```

Similar for going from an active NLB with valid subnets to a new CLB with invalid subnets. The CLB is not provisioned but the service remains with the old working ingress:
```
# subnet "subnet-0bacbe7de35736bb0" does not exist
# type is not set which means CLB
$ oc -n openshift-ingress get svc router-default -o yaml | yq .metadata.annotations | grep '^service.beta'
service.beta.kubernetes.io/aws-load-balancer-healthcheck-healthy-threshold: "2"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-interval: "10"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout: "4"
service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold: "2"
service.beta.kubernetes.io/aws-load-balancer-subnets: subnet-0bacbe7de35736bb0

# note the NLB domain: elb.<region>.amazonaws.com
$ oc -n openshift-ingress get svc router-default -o yaml | yq .status
loadBalancer:
  ingress:
    - hostname: ad02477fa382943289f0d234ce0734b8-ad502a60b6125b03.elb.us-east-2.amazonaws.com

$ curl -v ad02477fa382943289f0d234ce0734b8-ad502a60b6125b03.elb.us-east-2.amazonaws.com
* Rebuilt URL to: ad02477fa382943289f0d234ce0734b8-ad502a60b6125b03.elb.us-east-2.amazonaws.com/
* TCP_NODELAY set
* Connected to ad02477fa382943289f0d234ce0734b8-ad502a60b6125b03.elb.us-east-2.amazonaws.com port 80 (#0)
> GET / HTTP/1.1
> Host: ad02477fa382943289f0d234ce0734b8-ad502a60b6125b03.elb.us-east-2.amazonaws.com
> User-Agent: curl/7.61.1
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 503 Service Unavailable
< pragma: no-cache
< cache-control: private, max-age=0, no-cache, no-store
< content-type: text/html
```

Should we do another round of tests for the change of the LB type?

I suppose that if my findings can be confirmed we have a minimal impact on the incoming traffic while we can keep the behavior similar to other cases (LB scope + LB type).  WDYT?

test/e2e/util_test.go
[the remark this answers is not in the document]
gcs278: Sure, will do in next push.

gcs278:
@alebedev87 
> I think we don't update the Kube service with the new scope, while the LB type gets changed. Plus the status should be updated with instructions to recreate the service.

After our slack discussion, this [update](https://github.com/openshift/cluster-ingress-operator/compare/36be87ad990e0fb2aa45965cbe8f698ffed3002e..ad2944957adcd2c573656fdb18a90d511650793f) changes back to the "simpler" solution of leaving LB Type as auto effectuate, and also maintaining that changing subnets always require manual effectuation (i.e. need to delete the service).

That means changing both LB Type and Subnets at the same time results in:
1. CIO updates the LB Type immediately, but without changing the subnets (results in the subnets in the opposite LB type being used...temporarily)
2. CIO sets Progressing=True because subnets in status (actual) don't match the subnets in the spec (desired)
3. Cluster admin recreates the service to effectuate the subnet update (or reverts according to the status).

Ready for review again.

test/e2e/lb_subnets_test.go
gcs278: True, I did say I would test that in the EP. Added another test called `TestUnmanagedAWSLBSubnets`

alebedev87:
/lgtm
/approve

/hold
For @miheer to have a final look.

test/e2e/lb_subnets_test.go
```diff
@@ -0,0 +1,450 @@
...
+	t.Logf("updating ingresscontroller %q to use an NLB and public subnets", ic.Name)
+	if err = updateIngressControllerWithRetryOnConflict(t, icName, 5*time.Minute, func(ic *operatorv1.IngressController) {
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.Type = operatorv1.AWSNetworkLoadBalancer
+		ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters = &operatorv1.AWSNetworkLoadBalancerParameters{
+			Subnets: publicSubnets,
+		}
+	}); err != nil {
+		t.Fatalf("failed to update ingresscontroller: %v", err)
+	}
+	if err = waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, loadBalancerProgressingTrue); err != nil {
```
miheer, line 144: do we need this as you are already checking Progressing=True in effectuateIngressControllerSubnets function ?

miheer (changes requested):

test/e2e/lb_subnets_test.go
gcs278: Good catch! No we don't. Removed.

gcs278:
@miheer has gotten a chance to review
Needs one more LGTM
/unhold

gcs278:
Sorry, missed a cleanup. Need LGTM again.

miheer:
@gcs278 

ci/prow/verify is failing:
      
hack/verify-e2e-test-all-presence.sh
# github.com/openshift/cluster-ingress-operator/test/e2e [github.com/openshift/cluster-ingress-operator/test/e2e.test]
./lb_subnets_test.go:134:80: undefined: loadBalancerProgressingTrue

gcs278:
`gateway_api_test.go:76: failed to observe successful status of one or more gateway object/s: gateway class openshift-default not Accepted, last recorded status message: Waiting for controller`
This needs to be followed up.
The others are cluster install failures
/retest

test/e2e/util_test.go
```diff
@@ -983,12 +984,81 @@ func isFeatureGateEnabled(featureGateName configv1.FeatureGateName) (bool, error
...
+
+// networkLoadBalancerParametersStatusExist checks if NetworkLoadBalancerParameters exist in the IngressController
+// status.
+func networkLoadBalancerParametersStatusExist(ic *operatorv1.IngressController) bool {
+	return ic.Status.EndpointPublishingStrategy != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters != nil &&
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters.Subnets != nil
```
miheer, line 1052: do you need to check this condition for Subnets ?

miheer (changes requested):

gcs278:
/hold

minor e2e bug @miheer pointed out.

gcs278:
/unhold

resolved the e2e bug. Hopefully E2E still works correctly.

test/e2e/util_test.go
gcs278: Whoops, no, thanks! Fixed.

gcs278:
Note: It slipped my mind that `e2e-aws-operator-techpreview` is optional, so we needed to intentionally hold the PR until we see green CI. That's my fault. E2E wasn't actually passing due to a bug in the E2E logic. Follow up fixes are in https://github.com/openshift/cluster-ingress-operator/pull/1114, and that PR will be held until `e2e-aws-operator-techpreview` passes.

PR #927, opened 2023-05-05
OCPBUGS-13190: Avoid spurious updates for internalTrafficPolicy

Miciah:
#### Avoid spurious updates for internalTrafficPolicy

Specify `spec.internalTrafficPolicy` on NodePort- and ClusterIP-type services that the operator manages.  Also, ignore updates to the `spec.ipFamilies` and `spec.ipFamilyPolicy` fields.

Before this PR, the update logic for NodePort- and ClusterIP-type services would try to revert the default values that the API set for these fields.

* `assets/router/service-cloud.yaml`:
* `assets/router/service-internal.yaml`: Specify `internalTrafficPolicy: Cluster`.
* `pkg/manifests/bindata.go`: Regenerate.
* `pkg/operator/controller/ingress/internal_service.go` (`internalServiceChanged`): Ignore `spec.ipFamilies` and `spec.ipFamilyPolicy`.
* `pkg/operator/controller/ingress/internal_service_test.go` (`Test_desiredInternalIngressControllerService`): Verify that `spec.internalServiceChanged` is set to "Cluster".
(`Test_internalServiceChanged`): Verify that changes to `spec.internalTrafficPolicy` are detected and that changes to `spec.ipFamilies` and `spec.ipFamilyPolicy` are ignored.
* `pkg/operator/controller/ingress/nodeport_service.go` (`desiredNodePortService`): Set `spec.internalTrafficPolicy` to "Cluster".
(`nodePortServiceChanged`): Ignore `spec.ipFamilies` and `spec.ipFamilyPolicy`.
* `pkg/operator/controller/ingress/nodeport_service_test.go` (`TestDesiredNodePortService`): Verify that `spec.internalTrafficPolicy` is set to "Cluster".
(`TestNodePortServiceChanged`): Verify that changes to `spec.internalTrafficPolicy` are detected and that changes to `spec.ipFamilies` and `spec.ipFamilyPolicy` are ignored.

#### Ignore updates for null versus empty ennotations

Ignore updates to annotations when the update is from null to empty or vice versa.

* `pkg/operator/controller/ingress/internal_service.go` (`internalServiceChanged`): Use `EquateEmpty` when comparing annotations.
* `pkg/operator/controller/ingress/internal_service_test.go` (`TestInternalServiceChangedEmptyAnnotations`): New test to verify that `internalServiceChanged` treats empty and null annotations as equal.
[...]

PR #567, opened 2021-03-06
Bug 1936030: Fix spurious reconciliation of NodePort services

Miciah:
#### Log diffs when reconciling resources

This change improves log output when reconciling managed resources to include the diff between the current and updated resources when updating resources.

* `pkg/operator/controller/canary/route.go` (`updateCanaryRoute`): Log the diff between the current and updated resources when updating the route.
* `pkg/operator/controller/ingress/cluster-role.go` (`ensureClusterRole`): Move logging of updates from here...
(`updateClusterRole`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/dns.go` (`ensureWildcardDNSRecord`): Move logging of updates from here...
(`updateDNSRecord`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/load_balancer_service.go` (`ensureLoadBalancerService`): Move logging of updates from here...
(`updateLoadBalancerService`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/monitoring.go` (`ensureServiceMonitor`): Move logging of updates from here...
(`updateServiceMonitor`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/nodeport_service.go` (`ensureNodePortService`): Move logging of updates from here...
(`updateNodePortService`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/poddisruptionbudget.go` (`ensureRouterPodDisruptionBudget`): Move logging of updates from here...
(`updateRouterPodDisruptionBudget`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/rsyslog_configmap.go` (`ensureRsyslogConfigMap`): Move logging of updates from here...
(`updateRouterPodDisruptionBudget`): ...to here.  Log diff.
* `pkg/operator/controller/ingress/serviceca_configmap.go` (`ensureServiceCAConfigMap`): Move logging of updates from here...
(`updateServiceCAConfigMap`): ...to here.  Log diff.

#### `nodePortServiceChanged`: Fix for `clusterIPs`

Ignore the `clusterIPs` field when comparing NodePort services to determine whether an update is required.

Before this change, the update logic would keep trying to revert the default value that the API set for the service's `clusterIPs` field.

The `clusterIPs` field is new in Kubernetes 1.20 (OpenShift 4.7).

* `pkg/operator/controller/ingress/nodeport_service.go` (`nodePortServiceChanged`): Ignore the service's `clusterIPs` field.
[...]

PR #1052, opened 2024-04-30
OCPBUGS-23221: internalServiceChanged: Fix target port logic

Miciah:
#### `internalServiceChanged:` Fix target port logic

Fix `internalServiceChanged` to copy target ports from the expected service rather than preserving target ports from the current service.    Only copy `NodePort` and other fields that we want to preserve.

Before this PR, the `internalServiceChanged` function explicitly preserved the current service's ports' target ports when reconciling the service.  However, the function was intended to update the target ports if they changed.

In particular, if the current service specifies a numeric target port for the metrics port, then the operator should replace the port number with a port name so that the service uses the correct port even if the pod container's port number changes.  Without this change, updating an IngressController's `spec.endpointPublishingStrategy.hostNetwork.statsPort` API field can break metrics as the service's metrics port's numeric target port may no longer match the pod container's metrics port.

Follow-up to #864.

#### `Test_*Changed`: Verify that some update was made

This is a follow-up to the previous change for consistency.

#### `crdChanged`, `nodePortServiceChanged`: Fix preserve logic

Fix the update logic in `crdChanged` and `nodePortServiceChanged` to preserve fields that are ignored when checking whether an update is necessary.

Follow-up to #567, #890, and #927.

#### Set SessionAffinity to avoid confusing diffs

If a service does not specify `spec.sessionAffinity`, the API sets a default value of "None".  This means that when creating or updating a service, the empty value and "None" are equivalent, and so when reconciling a service, cluster-ingress-operator treats these values as equal.  This results in the correct update behavior:  The operator ignores the API defaulting, but if someone sets any value other than the empty value or "None", the operator reverts the update; and if the operator leaves `spec.sessionAffinity` empty when reverting the update, the API again defaults it to "None".

However, even the updating behavior is correct, it can cause confusing log output when the operator logs the diff with the changes:  If the operator has the empty value in the updated object but the current value is actually "None", this shows up in the diff, even though the actual update does not change the field value when taking defaulting into account.

This change adds an explicit `spec.sessionAffinity: None` setting to the services that the operator manages to avoid having the API defaulting show up in these diffs.

Follow-up to #407.

#### `Test_desiredLoadBalancerService`: Add assertions

Add assertions for service `type`, `externalTrafficPolicy`, `internalTrafficPolicy`, and `ports`, for consistency with `Test_desiredInternalIngressControllerService`.

#### `Test_desiredInternalIngressControllerService`: Fix godoc

Fix capitalization of the test function name in the godoc.

#### `TestHostNetworkPortBinding`: Use `t.Log`, `t.Cleanup`

Update `TestHostNetworkPortBinding` to comply with current conventions.  Convert some comments into `t.Log` statements, replace `defer` with `t.Cleanup`, and tweak some error messages.

#### `TestHostNetworkPortBinding`: Check service ports

Update `TestHostNetworkPortBinding` to verify that the internal service has the expected ports.  In particular, the metrics port should have its target port set to the name "metrics" rather than a port number.

#### `Test_deploymentConfigChanged`: Use k8s.io/utils/ptr

Replace use of the deprecated "k8s.io/utils/pointer" package by using "k8s.io/utils/ptr" instead.

#### Set "service-ca-bundle" volume's default mode

Set an explicit value of 420 decimal (0644 octal) for the router deployment's "service-ca-bundle" volume's default mode.

Previously, the router deployment yaml manifest had `defaultMode` indented incorrectly, and so the operator did not actually set the intended value for the volume's default mode.  Because the intended value is equal to the default value, the volume ultimately ends up with the intended default mode anyway.  However, the fact that the operator did not specify the default mode made log output show a bogus diff for the volume when the operator updated the deployment.

After this change, the yaml is correctly indented, and the log output for a deployment update has a less noisy diff.

Follow-up to #399.

#### Specify default mode on configmap/secret volumes

Set an explicit value of 420 decimal (0644 octal) for the default mode of all configmap or secret volumes in the router deployment that are defined in Go code (as opposed to being defined in the yaml manifest): client-ca, default-certificate, error-pages, metrics-certs, rsyslog-config, service-ca-bundle, and stats-auth.

After this change, the log output for a deployment update has a less noisy diff.

#### Specify parameter values for router probes

Set explicit values for the liveness, readiness, and startup probes in the router deployment's yaml manifest.

Previously, the router deployment yaml manifest omitted these values when the intended values were equal to the default values.  This meant that the API server would set the default values for these parameters, and so the probes ultimately ended up with the intended parameter values.  However, omitting these values in the yaml manifest also had the unintended side-effect of making the operator log bogus diffs for these parameters whenever it updated the deployment.

After this change, the log output for a deployment update has a less noisy diff.

####  `copyProbe`: Refactor to group copying of defaults

Group together the assignments for the case that `copyDefaultValues` is true.

### DIFF
--- a/pkg/manifests/assets/router/deployment.yaml
+++ b/pkg/manifests/assets/router/deployment.yaml
@@ -31,20 +31,33 @@ spec:
           - name: DEFAULT_DESTINATION_CA_PATH
             value: /var/run/configmaps/service-ca/service-ca.crt
           livenessProbe:
+            failureThreshold: 3
             httpGet:
               path: /healthz
               port: 1936
+              scheme: HTTP
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
             terminationGracePeriodSeconds: 10
           readinessProbe:
+            failureThreshold: 3
             httpGet:
               path: /healthz/ready
               port: 1936
+              scheme: HTTP
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
           startupProbe:
             failureThreshold: 120
             httpGet:
               path: /healthz/ready
               port: 1936
+              scheme: HTTP
             periodSeconds: 1
+            successThreshold: 1
+            timeoutSeconds: 1
           resources:
             requests:
               cpu: 100m
--- a/pkg/manifests/assets/router/deployment.yaml
+++ b/pkg/manifests/assets/router/deployment.yaml
@@ -63,9 +63,9 @@ spec:
           # SecretName is set at run-time.
       - name: service-ca-bundle
         configMap:
+          defaultMode: 420
           items:
           - key: service-ca.crt
             path: service-ca.crt
           name: service-ca-bundle
           optional: false
-        defaultMode: 420
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1520,7 +1520,7 @@ func hashableProbe(probe *corev1.Probe) *corev1.Probe {
 
 	var hashableProbe corev1.Probe
 
-	copyProbe(probe, &hashableProbe)
+	copyProbe(probe, &hashableProbe, false)
 
 	return &hashableProbe
 }
@@ -1621,9 +1621,9 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 	updated.Spec.Template.Spec.Containers[0].SecurityContext = expected.Spec.Template.Spec.Containers[0].SecurityContext
 	updated.Spec.Template.Spec.Containers[0].Env = expected.Spec.Template.Spec.Containers[0].Env
 	updated.Spec.Template.Spec.Containers[0].Image = expected.Spec.Template.Spec.Containers[0].Image
-	copyProbe(expected.Spec.Template.Spec.Containers[0].LivenessProbe, updated.Spec.Template.Spec.Containers[0].LivenessProbe)
-	copyProbe(expected.Spec.Template.Spec.Containers[0].ReadinessProbe, updated.Spec.Template.Spec.Containers[0].ReadinessProbe)
-	copyProbe(expected.Spec.Template.Spec.Containers[0].StartupProbe, updated.Spec.Template.Spec.Containers[0].StartupProbe)
+	copyProbe(expected.Spec.Template.Spec.Containers[0].LivenessProbe, updated.Spec.Template.Spec.Containers[0].LivenessProbe, true)
+	copyProbe(expected.Spec.Template.Spec.Containers[0].ReadinessProbe, updated.Spec.Template.Spec.Containers[0].ReadinessProbe, true)
+	copyProbe(expected.Spec.Template.Spec.Containers[0].StartupProbe, updated.Spec.Template.Spec.Containers[0].StartupProbe, true)
 	updated.Spec.Template.Spec.Containers[0].VolumeMounts = expected.Spec.Template.Spec.Containers[0].VolumeMounts
 	updated.Spec.Template.Spec.Containers[0].Ports = expected.Spec.Template.Spec.Containers[0].Ports
 	updated.Spec.Template.Spec.Tolerations = expected.Spec.Template.Spec.Tolerations
@@ -1639,8 +1639,9 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 }
 
 // copyProbe copies probe parameters that the operator manages from probe a to
-// probe b.
-func copyProbe(a, b *corev1.Probe) {
+// probe b.  If a field in probe a has the default value, then the value is
+// copied to probe b only if copyDefaultValues is true.
+func copyProbe(a, b *corev1.Probe, copyDefaultValues bool) {
 	if a == nil || b == nil {
 		return
 	}
@@ -1651,7 +1652,7 @@ func copyProbe(a, b *corev1.Probe) {
 			Port: a.ProbeHandler.HTTPGet.Port,
 			Host: a.ProbeHandler.HTTPGet.Host,
 		}
-		if a.ProbeHandler.HTTPGet.Scheme != "HTTP" {
+		if copyDefaultValues || a.ProbeHandler.HTTPGet.Scheme != "HTTP" {
 			b.ProbeHandler.HTTPGet.Scheme = a.ProbeHandler.HTTPGet.Scheme
 		}
 	}
@@ -1661,14 +1662,14 @@ func copyProbe(a, b *corev1.Probe) {
 
 	// Users are permitted to modify the timeout, so *don't* copy it.
 
-	// Don't copy default values that the API set.
-	if a.PeriodSeconds != int32(10) {
+	// Conditionally copy default values that the API sets.
+	if copyDefaultValues || a.PeriodSeconds != int32(10) {
 		b.PeriodSeconds = a.PeriodSeconds
 	}
-	if a.SuccessThreshold != int32(1) {
+	if copyDefaultValues || a.SuccessThreshold != int32(1) {
 		b.SuccessThreshold = a.SuccessThreshold
 	}
-	if a.FailureThreshold != int32(3) {
+	if copyDefaultValues || a.FailureThreshold != int32(3) {
 		b.FailureThreshold = a.FailureThreshold
 	}
 }
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -32,6 +32,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/ptr"
 
 	configv1 "github.com/openshift/api/config/v1"
 )
@@ -432,7 +433,8 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		Name: statsVolumeName,
 		VolumeSource: corev1.VolumeSource{
 			Secret: &corev1.SecretVolumeSource{
-				SecretName: statsSecretName,
+				DefaultMode: ptr.To[int32](0644),
+				SecretName:  statsSecretName,
 			},
 		},
 	}
@@ -459,7 +461,8 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		Name: certsVolumeName,
 		VolumeSource: corev1.VolumeSource{
 			Secret: &corev1.SecretVolumeSource{
-				SecretName: certsSecretName,
+				DefaultMode: ptr.To[int32](0644),
+				SecretName:  certsSecretName,
 			},
 		},
 	}
@@ -478,6 +481,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 			Name: "error-pages",
 			VolumeSource: corev1.VolumeSource{
 				ConfigMap: &corev1.ConfigMapVolumeSource{
+					DefaultMode: ptr.To[int32](0644),
 					LocalObjectReference: corev1.LocalObjectReference{
 						Name: configmapName.Name,
 					},
@@ -770,6 +774,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 				Name: "rsyslog-config",
 				VolumeSource: corev1.VolumeSource{
 					ConfigMap: &corev1.ConfigMapVolumeSource{
+						DefaultMode: ptr.To[int32](0644),
 						LocalObjectReference: corev1.LocalObjectReference{
 							Name: controller.RsyslogConfigMapName(ci).Name,
 						},
@@ -1064,6 +1069,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 				Name: clientCAVolumeName,
 				VolumeSource: corev1.VolumeSource{
 					ConfigMap: &corev1.ConfigMapVolumeSource{
+						DefaultMode: ptr.To[int32](0644),
 						LocalObjectReference: corev1.LocalObjectReference{
 							Name: clientCAConfigmapName.Name,
 						},
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1646,13 +1646,16 @@ func copyProbe(a, b *corev1.Probe, copyDefaultValues bool) {
 		return
 	}
 
+	// Always copy values that differ from the default values, as well as
+	// values where the default value is the same as the zero value.
+
 	if a.ProbeHandler.HTTPGet != nil {
 		b.ProbeHandler.HTTPGet = &corev1.HTTPGetAction{
 			Path: a.ProbeHandler.HTTPGet.Path,
 			Port: a.ProbeHandler.HTTPGet.Port,
 			Host: a.ProbeHandler.HTTPGet.Host,
 		}
-		if copyDefaultValues || a.ProbeHandler.HTTPGet.Scheme != "HTTP" {
+		if a.ProbeHandler.HTTPGet.Scheme != "HTTP" {
 			b.ProbeHandler.HTTPGet.Scheme = a.ProbeHandler.HTTPGet.Scheme
 		}
 	}
@@ -1662,14 +1665,24 @@ func copyProbe(a, b *corev1.Probe, copyDefaultValues bool) {
 
 	// Users are permitted to modify the timeout, so *don't* copy it.
 
-	// Conditionally copy default values that the API sets.
-	if copyDefaultValues || a.PeriodSeconds != int32(10) {
+	if a.PeriodSeconds != int32(10) {
 		b.PeriodSeconds = a.PeriodSeconds
 	}
-	if copyDefaultValues || a.SuccessThreshold != int32(1) {
+	if a.SuccessThreshold != int32(1) {
 		b.SuccessThreshold = a.SuccessThreshold
 	}
-	if copyDefaultValues || a.FailureThreshold != int32(3) {
+	if a.FailureThreshold != int32(3) {
+		b.FailureThreshold = a.FailureThreshold
+	}
+
+	// If copyDefaultValues is true, copy values even if they are the
+	// default values that the API sets.
+	if copyDefaultValues {
+		if a.ProbeHandler.HTTPGet != nil {
+			b.ProbeHandler.HTTPGet.Scheme = a.ProbeHandler.HTTPGet.Scheme
+		}
+		b.PeriodSeconds = a.PeriodSeconds
+		b.SuccessThreshold = a.SuccessThreshold
 		b.FailureThreshold = a.FailureThreshold
 	}
 }
--- a/pkg/operator/controller/ingress/internal_service.go
+++ b/pkg/operator/controller/ingress/internal_service.go
@@ -118,7 +118,8 @@ func internalServiceChanged(current, expected *corev1.Service) (bool, *corev1.Se
 
 	serviceCmpOpts := []cmp.Option{
 		// Ignore fields that the API, other controllers, or user may
-		// have modified.
+		// have modified.  Note: This list must be kept consistent with
+		// the updated.Spec.Foo = current.Spec.Foo assignments below!
 		cmpopts.IgnoreFields(corev1.ServicePort{}, "NodePort"),
 		cmpopts.IgnoreFields(corev1.ServiceSpec{},
 			"ClusterIP", "ClusterIPs",
@@ -163,14 +164,18 @@ func internalServiceChanged(current, expected *corev1.Service) (bool, *corev1.Se
 		}
 	}
 	// Preserve fields that the API, other controllers, or user may have
-	// modified.
+	// modified.  Note: This list must be kept consistent with
+	// serviceCmpOpts above!
 	updated.Spec.ClusterIP = current.Spec.ClusterIP
+	updated.Spec.ClusterIPs = current.Spec.ClusterIPs
 	updated.Spec.ExternalIPs = current.Spec.ExternalIPs
 	updated.Spec.HealthCheckNodePort = current.Spec.HealthCheckNodePort
+	updated.Spec.IPFamilies = current.Spec.IPFamilies
+	updated.Spec.IPFamilyPolicy = current.Spec.IPFamilyPolicy
 	for i, updatedPort := range updated.Spec.Ports {
 		for _, currentPort := range current.Spec.Ports {
 			if currentPort.Name == updatedPort.Name {
-				updated.Spec.Ports[i].TargetPort = currentPort.TargetPort
+				updated.Spec.Ports[i].NodePort = currentPort.NodePort
 			}
 		}
 	}
--- a/pkg/operator/controller/ingress/internal_service_test.go
+++ b/pkg/operator/controller/ingress/internal_service_test.go
@@ -126,6 +126,14 @@ func Test_internalServiceChanged(t *testing.T) {
 			expect: true,
 		},
 		{
+			description: "if .spec.ports[*].nodePort changes",
+			mutate: func(svc *corev1.Service) {
+				svc.Spec.Ports[0].NodePort = int32(33337)
+				svc.Spec.Ports[1].NodePort = int32(33338)
+			},
+			expect: false,
+		},
+		{
 			description: "if .spec.ports changes by changing the metrics port's target port from an integer to a string",
 			mutate: func(svc *corev1.Service) {
 				for i := range svc.Spec.Ports {
@@ -227,6 +235,9 @@ func Test_internalServiceChanged(t *testing.T) {
 			if changed, updated := internalServiceChanged(&original, mutated); changed != tc.expect {
 				t.Errorf("expected internalServiceChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
+				if updatedChanged, _ := internalServiceChanged(&original, updated); !updatedChanged {
+					t.Error("internalServiceChanged reported changes but did not make any update")
+				}
 				if changedAgain, _ := internalServiceChanged(mutated, updated); changedAgain {
 					t.Error("internalServiceChanged does not behave as a fixed point function")
 				}
--- a/pkg/operator/controller/ingress/internal_service_test.go
+++ b/pkg/operator/controller/ingress/internal_service_test.go
@@ -13,7 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-// test_desiredInternalIngressControllerService verifies that
+// Test_desiredInternalIngressControllerService verifies that
 // desiredInternalIngressControllerService returns the expected service.
 func Test_desiredInternalIngressControllerService(t *testing.T) {
 	ic := &operatorv1.IngressController{
--- a/pkg/operator/controller/ingress/internal_service_test.go
+++ b/pkg/operator/controller/ingress/internal_service_test.go
@@ -53,6 +53,7 @@ func Test_desiredInternalIngressControllerService(t *testing.T) {
 		Port:       int32(1936),
 		TargetPort: intstr.FromString("metrics"),
 	}}, svc.Spec.Ports)
+	assert.Equal(t, "None", string(svc.Spec.SessionAffinity))
 }
 
 // Test_internalServiceChanged verifies that internalServiceChanged properly
--- a/pkg/operator/controller/ingress/monitoring_test.go
+++ b/pkg/operator/controller/ingress/monitoring_test.go
@@ -41,7 +41,12 @@ func Test_serviceMonitorChanged(t *testing.T) {
 	}
 	if changed, sm3 := serviceMonitorChanged(sm1, sm2); !changed {
 		t.Fatal("expected changed to be true after clearing servicemonitor's selector")
-	} else if changedAgain, _ := serviceMonitorChanged(sm2, sm3); changedAgain {
-		t.Fatal("serviceMonitorChanged does not behave as a fixed-point function")
+	} else {
+		if updatedChanged, _ := serviceMonitorChanged(sm1, sm3); !updatedChanged {
+			t.Error("serviceMonitorChanged reported changes but did not make any update")
+		}
+		if changedAgain, _ := serviceMonitorChanged(sm2, sm3); changedAgain {
+			t.Fatal("serviceMonitorChanged does not behave as a fixed-point function")
+		}
 	}
 }
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -40,6 +40,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 
 	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -1008,6 +1009,30 @@ func TestHostNetworkPortBinding(t *testing.T) {
 		t.Errorf("pod %s is not running on the host's network", pod1.Name)
 	}
 
+	t.Log("verifying that the first ingresscontroller's internal service has the expected ports")
+	ing1ServiceName := controller.InternalIngressControllerServiceName(ing1)
+	ing1Service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), ing1ServiceName, ing1Service); err != nil {
+		t.Fatal(err)
+	}
+	expectedInternalServicePorts := []corev1.ServicePort{{
+		Name:       "http",
+		Protocol:   "TCP",
+		Port:       int32(80),
+		TargetPort: intstr.FromString("http"),
+	}, {
+		Name:       "https",
+		Protocol:   "TCP",
+		Port:       int32(443),
+		TargetPort: intstr.FromString("https"),
+	}, {
+		Name:       "metrics",
+		Protocol:   "TCP",
+		Port:       int32(1936),
+		TargetPort: intstr.FromString("metrics"),
+	}}
+	assert.Equal(t, expectedInternalServicePorts, ing1Service.Spec.Ports)
+
 	t.Log("creating a second ingresscontroller on the same node but with different port bindings")
 	name2 := types.NamespacedName{Namespace: operatorNamespace, Name: "samehost"}
 	strategy := &operatorv1.HostNetworkStrategy{
@@ -1059,6 +1084,14 @@ func TestHostNetworkPortBinding(t *testing.T) {
 	if !pod2.Spec.HostNetwork {
 		t.Errorf("pod %s is not running on the host's network", pod2.Name)
 	}
+
+	t.Log("verifying that the second ingresscontroller's internal service has the expected ports")
+	ing2ServiceName := controller.InternalIngressControllerServiceName(ing2)
+	ing2Service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), ing2ServiceName, ing2Service); err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, expectedInternalServicePorts, ing2Service.Spec.Ports)
 }
 
 func assertContainerHasPort(t *testing.T, container corev1.Container, name string, port int32) {
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -55,6 +55,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
@@ -4178,4 +4179,64 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
 	}
 
 	return secret, cl.Create(context.TODO(), secret)
+}
+
+// TestReconcileInternalService verifies that the operator reverts changes to
+// the router-internal service.
+func TestReconcileInternalService(t *testing.T) {
+	t.Parallel()
+	icName := types.NamespacedName{
+		Namespace: operatorNamespace,
+		Name:      "reconcile-internal-service",
+	}
+	domain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ic) })
+
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, availableConditionsForPrivateIngressController...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	serviceName := controller.InternalIngressControllerServiceName(ic)
+	service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+		t.Fatal(err)
+	}
+
+	findPort := func(name string, service *corev1.Service) *corev1.ServicePort {
+		for i := range service.Spec.Ports {
+			if service.Spec.Ports[i].Name == name {
+				return &service.Spec.Ports[i]
+			}
+		}
+		return nil
+	}
+
+	metricsPort := findPort("metrics", service)
+	if metricsPort == nil {
+		t.Fatalf(`"metrics" port was not found among the service's ports: %v`, service.Spec.Ports)
+	}
+	originalMetricsPortTargetPort := metricsPort.TargetPort
+	metricsPort.TargetPort = intstr.FromInt(12345)
+	if err := kclient.Update(context.TODO(), service); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := wait.PollImmediate(1*time.Second, time.Minute, func() (bool, error) {
+		if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+			t.Log(err)
+			return false, nil
+		}
+		if metricsPort := findPort("metrics", service); metricsPort == nil {
+			t.Fatalf(`"metrics" port was removed from the service's ports: %v`, service.Spec.Ports)
+		} else if metricsPort.TargetPort == originalMetricsPortTargetPort {
+			return true, nil
+		}
+		return false, nil
+	}); err != nil {
+		t.Errorf("failed to observe expected conditions: %v", err)
+	}
 }
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -971,23 +971,23 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
 // TestHostNetworkPortBinding creates two ingresscontrollers on the same node
 // with different port bindings and verifies that both routers are available.
 func TestHostNetworkPortBinding(t *testing.T) {
-	// deploy first ingresscontroller with the default port bindings
+	t.Log("creating an ingresscontroller with the default port bindings")
 	name1 := types.NamespacedName{Namespace: operatorNamespace, Name: "hostnetworkportbinding"}
 	ing1 := newHostNetworkController(name1, name1.Name+"."+dnsConfig.Spec.BaseDomain)
 	if err := kclient.Create(context.TODO(), ing1); err != nil {
 		t.Fatalf("failed to create the first ingresscontroller: %v", err)
 	}
-	defer assertIngressControllerDeleted(t, kclient, ing1)
+	t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ing1) })
 
 	err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name1, availableConditionsForIngressControllerWithHostNetwork...)
 	if err != nil {
 		t.Errorf("failed to observe expected conditions for the first ingresscontroller: %v", err)
 	}
 
-	// get first router's single replica
+	t.Log("getting pod list for the first ingresscontroller")
 	pods := &corev1.PodList{}
 	if err := kclient.List(context.TODO(), pods, client.InNamespace(operandNamespace)); err != nil {
-		t.Fatalf("failed to list the first ingresscontroller's PODs: %v", err)
+		t.Fatalf("failed to list the first ingresscontroller's pods: %v", err)
 	}
 	var pod1 *corev1.Pod
 	for _, p := range pods.Items {
@@ -997,7 +997,7 @@ func TestHostNetworkPortBinding(t *testing.T) {
 		}
 	}
 	if pod1 == nil {
-		t.Fatal("failed to find the first ingresscontroller's POD")
+		t.Fatal("failed to find the first ingresscontroller's pod")
 	}
 
 	routerContainer := pod1.Spec.Containers[0]
@@ -1008,14 +1008,16 @@ func TestHostNetworkPortBinding(t *testing.T) {
 		t.Errorf("pod %s is not running on the host's network", pod1.Name)
 	}
 
-	// create second ingresscontroller on the same node but with different port bindings
+	t.Log("creating a second ingresscontroller on the same node but with different port bindings")
 	name2 := types.NamespacedName{Namespace: operatorNamespace, Name: "samehost"}
 	strategy := &operatorv1.HostNetworkStrategy{
 		HTTPPort:  9080,
 		HTTPSPort: 9443,
 		StatsPort: 9936,
 	}
-	// take the node placement of the first router
+	// Take the node name of the first ingresscontroller and configure a
+	// node selector on the second ingresscontroller to force the router
+	// pods all to land on the same node host.
 	placement := &operatorv1.NodePlacement{
 		Tolerations: pod1.Spec.Tolerations,
 		NodeSelector: &metav1.LabelSelector{
@@ -1029,15 +1031,16 @@ func TestHostNetworkPortBinding(t *testing.T) {
 	if err := kclient.Create(context.TODO(), ing2); err != nil {
 		t.Fatalf("failed to create the second ingresscontroller: %v", err)
 	}
-	defer assertIngressControllerDeleted(t, kclient, ing2)
+	t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ing2) })
 
 	err = waitForIngressControllerCondition(t, kclient, 5*time.Minute, name2, availableConditionsForIngressControllerWithHostNetwork...)
 	if err != nil {
 		t.Errorf("failed to observe expected conditions for the second ingresscontroller: %v", err)
 	}
 
+	t.Log("getting pod list for the second ingresscontroller")
 	if err := kclient.List(context.TODO(), pods, client.InNamespace(operandNamespace)); err != nil {
-		t.Fatalf("failed to list the first ingresscontroller's PODs: %v", err)
+		t.Fatalf("failed to list the first ingresscontroller's pods: %v", err)
 	}
 	var pod2 *corev1.Pod
 	for _, p := range pods.Items {
@@ -1047,7 +1050,7 @@ func TestHostNetworkPortBinding(t *testing.T) {
 		}
 	}
 	if pod2 == nil {
-		t.Fatalf("failed to find the second ingresscontroller's POD")
+		t.Fatalf("failed to find the second ingresscontroller's pod")
 	}
 	routerContainer = pod2.Spec.Containers[0]
 	assertContainerHasPort(t, routerContainer, "http", 9080)
--- a/pkg/manifests/assets/router/service-cloud.yaml
+++ b/pkg/manifests/assets/router/service-cloud.yaml
@@ -24,3 +24,4 @@ spec:
     protocol: TCP
     port: 443
     targetPort: https
+  sessionAffinity: None
--- a/pkg/manifests/assets/router/service-internal.yaml
+++ b/pkg/manifests/assets/router/service-internal.yaml
@@ -19,3 +19,4 @@ spec:
     port: 1936
     protocol: TCP
     targetPort: metrics
+  sessionAffinity: None
--- a/pkg/operator/controller/canary/daemonset_test.go
+++ b/pkg/operator/controller/canary/daemonset_test.go
@@ -200,6 +200,9 @@ func Test_canaryDaemonsetChanged(t *testing.T) {
 			if changed, updated := canaryDaemonSetChanged(original, mutated); changed != tc.expect {
 				t.Errorf("expect canaryDaemonSetChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
+				if updatedChanged, _ := canaryDaemonSetChanged(original, updated); !updatedChanged {
+					t.Error("canaryDaemonSetChanged reported changes but did not make any update")
+				}
 				if changedAgain, _ := canaryDaemonSetChanged(mutated, updated); changedAgain {
 					t.Error("canaryDaemonSetChanged does not behave as a fixed point function")
 				}
--- a/pkg/operator/controller/canary/namespace_test.go
+++ b/pkg/operator/controller/canary/namespace_test.go
@@ -38,6 +38,9 @@ func Test_canaryNamespaceChanged(t *testing.T) {
 			if changed, updated := canaryNamespaceChanged(original, mutated); changed != tc.expect {
 				t.Errorf("expect canaryNamespaceChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
+				if updatedChanged, _ := canaryNamespaceChanged(original, updated); !updatedChanged {
+					t.Error("canaryNamespaceChanged reported changes but did not make any update")
+				}
 				if changedAgain, _ := canaryNamespaceChanged(mutated, updated); changedAgain {
 					t.Error("canaryNamespaceChanged does not behave as a fixed point function")
 				}
--- a/pkg/operator/controller/canary/route_test.go
+++ b/pkg/operator/controller/canary/route_test.go
@@ -134,6 +134,9 @@ func Test_canaryRouteChanged(t *testing.T) {
 			if changed, updated := canaryRouteChanged(original, mutated); changed != tc.expect {
 				t.Errorf("expected canaryRouteChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
+				if updatedChanged, _ := canaryRouteChanged(original, updated); !updatedChanged {
+					t.Error("canaryRouteChanged reported changes but did not make any update")
+				}
 				if changedAgain, _ := canaryRouteChanged(mutated, updated); changedAgain {
 					t.Error("canaryRouteChanged does not behave as a fixed point function")
 				}
--- a/pkg/operator/controller/gatewayapi/crds.go
+++ b/pkg/operator/controller/gatewayapi/crds.go
@@ -114,7 +114,9 @@ func (r *reconciler) updateCRD(ctx context.Context, current, desired *apiextensi
 // the expected spec and if not returns an updated one.
 func crdChanged(current, expected *apiextensionsv1.CustomResourceDefinition) (bool, *apiextensionsv1.CustomResourceDefinition) {
 	crdCmpOpts := []cmp.Option{
-		// Ignore fields that the API may have modified.
+		// Ignore fields that the API may have modified.  Note: This
+		// list must be kept consistent with the updated.Spec.Foo =
+		// current.Spec.Foo assignments below!
 		cmpopts.IgnoreFields(apiextensionsv1.CustomResourceDefinitionSpec{}, "Conversion"),
 		cmpopts.EquateEmpty(),
 	}
@@ -124,6 +126,10 @@ func crdChanged(current, expected *apiextensionsv1.CustomResourceDefinition) (bo
 
 	updated := current.DeepCopy()
 	updated.Spec = expected.Spec
+	// Preserve fields that the API, other controllers, or user may have
+	// modified.  Note: This list must be kept consistent with crdCmpOpts
+	// above!
+	updated.Spec.Conversion = current.Spec.Conversion
 
 	return true, updated
 }
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -1869,6 +1869,9 @@ func Test_deploymentConfigChanged(t *testing.T) {
 			if changed, updated := deploymentConfigChanged(&original, mutated); changed != tc.expect {
 				t.Errorf("expect deploymentConfigChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
+				if updatedChanged, _ := deploymentConfigChanged(&original, updated); !updatedChanged {
+					t.Error("deploymentConfigChanged reported changes but did not make any update")
+				}
 				if changedAgain, _ := deploymentConfigChanged(mutated, updated); changedAgain {
 					t.Error("deploymentConfigChanged does not behave as a fixed point function")
 				}
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -470,6 +470,21 @@ func Test_desiredRouterDeployment(t *testing.T) {
 	checkDeploymentHasEnvSorted(t, deployment)
 }
 
+// assertHasVolumes asserts that the given slice of volumes has all of the
+// named volumes.
+func assertHasVolumes(t *testing.T, volumes []corev1.Volume, expectedVolumeNames []string) {
+	t.Helper()
+
+	actualVolumeNames := []string{}
+	for _, volume := range volumes {
+		actualVolumeNames = append(actualVolumeNames, volume.Name)
+	}
+	assert.Equal(t, len(expectedVolumeNames), len(actualVolumeNames))
+	for _, volume := range expectedVolumeNames {
+		assert.Contains(t, actualVolumeNames, volume)
+	}
+}
+
 // assertVolumeHasDefaultMode asserts that the given int32 pointer points to a
 // value that is equal to the given int32 value.
 func assertVolumeHasDefaultMode(t *testing.T, expected int32, actual *int32, volumeName string) {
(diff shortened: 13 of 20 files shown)


Miciah:
[ci/prow/e2e-aws-ovn-single-node failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-single-node/1785472707073150976) with the following error:
```text
 Updating install-config.yaml to a single m5d.2xlarge control plane node and 0 workers
/bin/bash: line 19: pip3: command not found 
```
I filed [OCPBUGS-33162](https://issues.redhat.com/browse/OCPBUGS-33162) for this error.

Miciah:
[ci/prow/e2e-aws-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1785472706871824384) and [ci/prow/e2e-azure-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1785472707215757312) both failed because `TestRouteAdmissionPolicy` failed, which is a known issue that is being tracked as part of [OCPBUGS-26498](https://issues.redhat.com/browse/OCPBUGS-26498).

candita:
@Miciah will this also need to be backported to 4.12, like https://github.com/openshift/cluster-ingress-operator/pull/864#issuecomment-1520468390 ?

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/53baea6f04f80b40128edd4d55cae58e7dd13dc3..584ef427468eac9d8a5a1948e11ae26f673d3114 updates the logic to preserve the same fields that we ignore when checking whether an update is needed, and also adds an E2E test.

Miciah:
> @Miciah will this also need to be backported to 4.12, like [#864 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/864#issuecomment-1520468390) ?

Yeah, if anyone requests a backport and PM agrees, it makes sense to backport this PR.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/584ef427468eac9d8a5a1948e11ae26f673d3114..600ed28e6f713ec4fe0d2e94b37061f96bf61852 adds comments about keeping cmp options and `updated.Spec.Foo = current.Spec.Foo` assignments consistent and adds some missing assignments.

Miciah:
[ci/prow/e2e-azure-operator failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1785815240303185920) because the installer timed out waiting for the API server respond after bootstrap.

[ci/prow/e2e-aws-operator failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1785815227724468224) because `TestRouteAdmissionPolicy` failed.

pkg/operator/controller/ingress/internal_service.go
```diff
@@ -163,14 +164,18 @@ func internalServiceChanged(current, expected *corev1.Service) (bool, *corev1.Se
...
+	updated.Spec.ClusterIPs = current.Spec.ClusterIPs
 	updated.Spec.ExternalIPs = current.Spec.ExternalIPs
 	updated.Spec.HealthCheckNodePort = current.Spec.HealthCheckNodePort
+	updated.Spec.IPFamilies = current.Spec.IPFamilies
+	updated.Spec.IPFamilyPolicy = current.Spec.IPFamilyPolicy
 	for i, updatedPort := range updated.Spec.Ports {
 		for _, currentPort := range current.Spec.Ports {
 			if currentPort.Name == updatedPort.Name {
-				updated.Spec.Ports[i].TargetPort = currentPort.TargetPort
+				updated.Spec.Ports[i].NodePort = currentPort.NodePort
```
candita, line 178: Doesn't this mean that we are allowing the spec.ports.nodePorts to be changed from what we expect, to what is set in current?

candita: If something/someone changed the nodePort, and only the nodePort, then L147-148 would return "unchanged" because we ignore it.  However, if something changed the nodePort, and also changed something we don't ignore, and we get to here, then we change it.  Am I missing something?

Miciah: > Doesn't this mean that we are allowing the spec.ports.nodePorts to be changed from what we expect, to what is set in current?

It does.  This is explained in the comments: https://github.com/openshift/cluster-ingress-operator/blob/4a306cb33d26c30a6bda1a2bef92a6c564b3c3b5/pkg/operator/controller/ingress/internal_service.go#L120-L122 https://github.com/openshift/cluster-ingress-operator/blob/4a306cb33d26c30a6bda1a2bef92a6c564b3c3b5/pkg/operator/controller/ingress/internal_service.go#L166-L168

In particular, `NodePort` is set by the API, and so `NodePort` must be ignored in the comparison and preserved during the update so that the operator does not try to revert the value that the API set.  

> If something/someone changed the nodePort, and only the nodePort, then L147-148 would return "unchanged" because we ignore it. However, if something changed the nodePort, and also changed something we don't ignore, and we get to here, then we change it. Am I missing something?

`updated.Spec.Ports[i].NodePort = currentPort.NodePort` copies the value that was observed (`currentPort.NodePort`) to the final version of the service that we will use for the update (`updated`).  That is, the assignment preserves the current `NodePort` value for updates.  

Note that we initialize `updated` by copying the `spec` field from `expected`: https://github.com/openshift/cluster-ingress-operator/blob/4a306cb33d26c30a6bda1a2bef92a6c564b3c3b5/pkg/operator/controller/ingress/internal_service.go#L151-L152

Then if we want to _preserve_ the current value for some subfield, we need to copy the value from `current`: https://github.com/openshift/cluster-ingress-operator/blob/4a306cb33d26c30a6bda1a2bef92a6c564b3c3b5/pkg/operator/controller/ingress/internal_service.go#L169-L178

At the end of all this, `updated` is the result of merging `current` and `expected`, copying the field values that the operator should update from `expected` and copying the field values that the operator should preserve (that is, leave alone) from `current`.

Miciah:
[e2e-gcp-operator succeeded](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1785815245374099456), and [the operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1785815245374099456/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-58469b5f8b-6dqzj_ingress-operator.log) have the expected output:
```text
% curl -so ingress-operator.log https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1785815245374099456/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-58469b5f8b-6dqzj_ingress-operator.log
% grep -e 'created internal ingresscontroller service' -c -- ingress-operator.log                                                                                                                                       
83
% grep -e 'updated internal service' -- ingress-operator.log                                                                                      
2024-05-02T00:36:39.000Z        INFO    operator.ingress_controller     ingress/internal_service.go:44  updated internal service        {"namespace": "openshift-ingress", "name": "router-internal-reconcile-internal-service", "diff": "  &v1.Service{\n  \tTypeMeta:   {},\n  \tObjectMeta: {Name: \"router-internal-reconcile-internal-service\", Namespace: \"openshift-ingress\", UID: \"5ba7f8c8-9af0-4da4-82dd-22dbf88c5f94\", ResourceVersion: \"48575\", ...},\n  \tSpec: v1.ServiceSpec{\n  \t\tPorts: []v1.ServicePort{\n  \t\t\t{Name: \"http\", Protocol: \"TCP\", Port: 80, TargetPort: {Type: 1, StrVal: \"http\"}, ...},\n  \t\t\t{Name: \"https\", Protocol: \"TCP\", Port: 443, TargetPort: {Type: 1, StrVal: \"https\"}, ...},\n  \t\t\t{\n  \t\t\t\t... // 2 identical fields\n  \t\t\t\tAppProtocol: nil,\n  \t\t\t\tPort:        1936,\n  \t\t\t\tTargetPort: intstr.IntOrString{\n- \t\t\t\t\tType:   0,\n+ \t\t\t\t\tType:   1,\n- \t\t\t\t\tIntVal: 12345,\n+ \t\t\t\t\tIntVal: 0,\n- \t\t\t\t\tStrVal: \"\",\n+ \t\t\t\t\tStrVal: \"metrics\",\n  \t\t\t\t},\n  \t\t\t\tNodePort: 0,\n  \t\t\t},\n  \t\t},\n  \t\tSelector:                 {\"ingresscontroller.operator.openshift.io/deployment-ingresscontroller\": \"reconcile-internal-service\"},\n  \t\tClusterIP:                \"172.30.106.3\",\n  \t\tClusterIPs:               {\"172.30.106.3\"},\n  \t\tType:                     \"ClusterIP\",\n  \t\tExternalIPs:              nil,\n- \t\tSessionAffinity:          \"None\",\n+ \t\tSessionAffinity:          \"\",\n  \t\tLoadBalancerIP:           \"\",\n  \t\tLoadBalancerSourceRanges: nil,\n  \t\t... // 10 identical fields\n  \t},\n  \tStatus: {},\n  }\n"}
% grep -e 'updated internal service' -- ingress-operator.log | sed -e 's/^[^{]*//' | jq -r .diff
  &v1.Service{
        TypeMeta:   {},
        ObjectMeta: {Name: "router-internal-reconcile-internal-service", Namespace: "openshift-ingress", UID: "5ba7f8c8-9af0-4da4-82dd-22dbf88c5f94", ResourceVersion: "48575", ...},
        Spec: v1.ServiceSpec{
                Ports: []v1.ServicePort{
                        {Name: "http", Protocol: "TCP", Port: 80, TargetPort: {Type: 1, StrVal: "http"}, ...},
                        {Name: "https", Protocol: "TCP", Port: 443, TargetPort: {Type: 1, StrVal: "https"}, ...},
                        {
                                ... // 2 identical fields
                                AppProtocol: nil,
                                Port:        1936,
                                TargetPort: intstr.IntOrString{
-                                       Type:   0,
+                                       Type:   1,
-                                       IntVal: 12345,
+                                       IntVal: 0,
-                                       StrVal: "",
+                                       StrVal: "metrics",
                                },
                                NodePort: 0,
                        },
                },
                Selector:                 {"ingresscontroller.operator.openshift.io/deployment-ingresscontroller": "reconcile-internal-service"},
                ClusterIP:                "172.30.106.3",
                ClusterIPs:               {"172.30.106.3"},
                Type:                     "ClusterIP",
                ExternalIPs:              nil,
-               SessionAffinity:          "None",
+               SessionAffinity:          "",
                LoadBalancerIP:           "",
                LoadBalancerSourceRanges: nil,
                ... // 10 identical fields
        },
        Status: {},
  }

% 
```
`SessionAffinity` is an unusual case; "None" and empty are equivalent, and the API defaults the value to "None", so it shows up in the diff.  I realize the update logic is confusing, and seeing `SessionAffinity` in the diff may add to the confusion, so I'll push a commit to prevent it.

Miciah:
Multiple CI jobs failed because `Clone the correct source code into an image and tag it as src` failed with `0/26 nodes are available`.  This is an issue with the CI cluster; the tests didn't even actually run.  
/retest-required

pkg/operator/controller/ingress/internal_service.go
candita: Still, 

> If something/someone changed the nodePort, and only the nodePort, then L147-148 would return "unchanged" because we ignore it. 

Does that mean the change to nodePort won't get saved?

Miciah: The `return false, nil` statement on line 147/148 means that `updateInternalService` won't do an update.  In that case, there's no need to preserve `NodePort`'s value—we're not doing an update anyway.

candita: Okay.  If we never expect current to hold any values that are different for the ignored fields, then why do we have to preserve them later?

Miciah: The operator preserves the ignored fields so that it doesn't inadvertently update fields that we intend for it to ignore.  

The operator ignores these fields because it isn't supposed to be managing them.  In particular, the API sets `NodePort`, so the operator should never try to update `NodePort`, even if `expected.Spec.Ports[i].NodePort` differs from `current.Spec.Ports[i].NodePort`.  

Maybe the word "expected" is misleading.  `expected` has some fields where the operator manages the field and actually expects the field to have a certain value, as well as some fields where the operator is intended to ignore changes to the field and preserve the current value,, whatever that value may be.  

Ultimately, the operator should update the service if a managed field (for example, `spec.internalTrafficPolicy`) does not have the expected value, and the operator should update that particular field.  However, the operator should not update the service if _only_ ignored fields (for example, `nodePort`) changed, and the operator should not update that field if the operator has to update the service because some managed field value needs an update.

test/e2e/operator_test.go
```diff
@@ -4179,3 +4180,63 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
...
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	serviceName := controller.InternalIngressControllerServiceName(ic)
+	service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+		t.Fatal(err)
+	}
+
+	findPort := func(name string, service *corev1.Service) *corev1.ServicePort {
```
candita, line 4209: Shouldn't this e2e test find/update the port on the IngressController, and check to make sure it is updated on the Service?

Miciah: `TestHostNetworkPortBinding` verifies that updating `IngressController.spec.endpointPublishingStrategy.hostNetwork.statsPort` updates the deployment.  `TestReconcileInternalService` verifies that the "metrics" port exists, and that the operator is managing it.  :thinking:... Would it make sense to add a check to `TestHostNetworkPortBinding` to verify that the service uses "metrics" (and not a number) as the target port?

Miciah: https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/3a56b8dbad66a5ca58fdd7d140985872e119c173 adds an assertion on the expected ports in `TestHostNetworkPortBinding`.

pkg/operator/controller/ingress/monitoring_test.go
```diff
@@ -41,7 +41,12 @@ func Test_serviceMonitorChanged(t *testing.T) {
 	}
 	if changed, sm3 := serviceMonitorChanged(sm1, sm2); !changed {
 		t.Fatal("expected changed to be true after clearing servicemonitor's selector")
-	} else if changedAgain, _ := serviceMonitorChanged(sm2, sm3); changedAgain {
-		t.Fatal("serviceMonitorChanged does not behave as a fixed-point function")
+	} else {
+		if updatedChanged, _ := serviceMonitorChanged(sm1, sm3); !updatedChanged {
+			t.Error("serviceMonitorChangedChanged reported changes but did not make any update")
```
candita, line 46: nit:
```suggestion
			t.Error("serviceMonitorChanged reported changes but did not make any update")
```

Miciah: Thanks!  Fixed in https://github.com/openshift/cluster-ingress-operator/compare/3a56b8dbad66a5ca58fdd7d140985872e119c173..3d06500a64b02ae95d533bebc03fd4bee019e05b.

pkg/operator/controller/ingress/internal_service_test.go
candita, line 1: nit: you could add a test case for NodePort changes.

Miciah: Added in https://github.com/openshift/cluster-ingress-operator/compare/3d06500a64b02ae95d533bebc03fd4bee019e05b..cfba4d9ad778d01a6ab874682ad0c2d95f81eab0.

test/e2e/operator_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
candita, line 27: This import needs to be reordered alphabetically.

test/e2e/operator_test.go
```diff
@@ -970,23 +972,23 @@ func TestHostNetworkEndpointPublishingStrategy(t *testing.T) {
...
-	defer assertIngressControllerDeleted(t, kclient, ing1)
+	t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ing1) })
 
 	err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name1, availableConditionsForIngressControllerWithHostNetwork...)
 	if err != nil {
 		t.Errorf("failed to observe expected conditions for the first ingresscontroller: %v", err)
 	}
 
-	// get first router's single replica
+	t.Log("getting first router's single replica")
```
candita, line 988: I believe this is more accurate:
```suggestion
	t.Log("getting first router's pod list")
```

test/e2e/operator_test.go
Miciah: It's already alphabetical (and gofmt is happy with it).  Do you mean it needs to be regrouped?  I shied away from that given that the current grouping is inconsistent and we already have plans to fix it throughout our codebase, but I can fix up this file now if you like.  

Would you like me to regroup as follows?
```go
import (
	"bufio"
	"bytes"
	"context"
	"crypto/tls"
	"crypto/x509"
	"encoding/json"
	"errors"
	"fmt"
	"io"
	"io/ioutil"
	"net"
	"net/http"
	"os"
	"reflect"
	"sort"
	"strconv"
	"strings"
	"testing"
	"time"

	"github.com/aws/aws-sdk-go/aws/endpoints"
	"github.com/go-logr/logr"
	routev1 "github.com/openshift/api/route/v1"
	"github.com/stretchr/testify/assert"
	"gopkg.in/yaml.v2"
	appsv1 "k8s.io/api/apps/v1"
	corev1 "k8s.io/api/core/v1"
	networkingv1 "k8s.io/api/networking/v1"
	policyv1 "k8s.io/api/policy/v1"
	apierrors "k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/apimachinery/pkg/util/intstr"
	"k8s.io/apimachinery/pkg/util/sets"
	"k8s.io/apiserver/pkg/storage/names"
	discocache "k8s.io/client-go/discovery/cached"
	"k8s.io/client-go/dynamic"
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/restmapper"
	"k8s.io/client-go/scale"
	"k8s.io/utils/pointer"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/client/config"
	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"

	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
)
```

test/e2e/operator_test.go
candita: `TestHostNetworkPortBinding` would have prevented the original bug if it was testing the same thing.  It is testing `assertContainerHasPort(t, routerContainer, "metrics", 9936)`.  Doesn't it need to assert that the service now has the port?

test/e2e/operator_test.go
candita: I hadn't noticed how unordered the whole list was, only that the new import did not appear to be in the right place.  

I would be happy if you just added the new import after:
"github.com/go-logr/logr"

But if you don't want to do that without fixing all the others, I get it.

candita: The regroup isn't grouping all the github.com/openshift together, so definitely not that.

test/e2e/operator_test.go
candita: Or, maybe the new assertion for expected ports should have the new port numbers too?  https://github.com/openshift/cluster-ingress-operator/pull/1052/files/cfba4d9ad778d01a6ab874682ad0c2d95f81eab0#diff-cf4b4e5424070a666a364d1d1b04011478d888d6d3d65c943ca7783333b7a6e4R1018-R1034

candita: At least during the second check on https://github.com/openshift/cluster-ingress-operator/pull/1052/files/cfba4d9ad778d01a6ab874682ad0c2d95f81eab0#diff-cf4b4e5424070a666a364d1d1b04011478d888d6d3d65c943ca7783333b7a6e4R1093

test/e2e/operator_test.go
Miciah: Do you want me to group `"github.com/go-logr/logr"` with `"gopkg.in/yaml.v2"`, or just move `"github.com/stretchr/testify/assert"`?

test/e2e/operator_test.go
Miciah: > `TestHostNetworkPortBinding` would have prevented the original bug if it was testing the same thing as the bug report. It is testing `assertContainerHasPort(t, routerContainer, "metrics", 9936)`. Doesn't it need to assert that the service now has the port?

Is the assertion that I added what you expected?

> Or, maybe the new assertion for expected ports should have the new port numbers too? https://github.com/openshift/cluster-ingress-operator/pull/1052/files/cfba4d9ad778d01a6ab874682ad0c2d95f81eab0#diff-cf4b4e5424070a666a364d1d1b04011478d888d6d3d65c943ca7783333b7a6e4R1018-R1034

The port changes on the pod, but the port doesn't need to change on the service if the service just references the target port by name instead of by number.  That's the idea behind #864.  The fault is that the update logic didn't actually work properly, and so the service would still have port 1936 instead of port "metrics".

Miciah:
[e2e-gcp-operator failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1786519823082065920) because the API was unresponsive after bootstrap.  

[e2e-aws-operator passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1786519804631322624), but must-gather failed so I cannot check the ingress-operator logs.  However, this success does indicate that `TestHostNetworkPortBinding` is passing with the added assertions.

Miciah:
The ingress-operator logs from [the last e2e-azure-operator run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1786519818074066944) look good:
```text
% grep -e 'updated internal service' -- ingress-operator.log
2024-05-03T23:24:40.288Z        INFO    operator.ingress_controller     ingress/internal_service.go:44  updated internal service        {"namespace": "openshift-ingress", "name": "router-internal-reconcile-internal-service", "diff": "  &v1.Service{\n  \tTypeMeta:   {},\n  \tObjectMeta: {Name: \"router-internal-reconcile-internal-service\", Namespace: \"openshift-ingress\", UID: \"aaa07625-e508-435c-b8e8-97891c17cd17\", ResourceVersion: \"38668\", ...},\n  \tSpec: v1.ServiceSpec{\n  \t\tPorts: []v1.ServicePort{\n  \t\t\t{Name: \"http\", Protocol: \"TCP\", Port: 80, TargetPort: {Type: 1, StrVal: \"http\"}, ...},\n  \t\t\t{Name: \"https\", Protocol: \"TCP\", Port: 443, TargetPort: {Type: 1, StrVal: \"https\"}, ...},\n  \t\t\t{\n  \t\t\t\t... // 2 identical fields\n  \t\t\t\tAppProtocol: nil,\n  \t\t\t\tPort:        1936,\n  \t\t\t\tTargetPort: intstr.IntOrString{\n- \t\t\t\t\tType:   0,\n+ \t\t\t\t\tType:   1,\n- \t\t\t\t\tIntVal: 12345,\n+ \t\t\t\t\tIntVal: 0,\n- \t\t\t\t\tStrVal: \"\",\n+ \t\t\t\t\tStrVal: \"metrics\",\n  \t\t\t\t},\n  \t\t\t\tNodePort: 0,\n  \t\t\t},\n  \t\t},\n  \t\tSelector:  {\"ingresscontroller.operator.openshift.io/deployment-ingresscontroller\": \"reconcile-internal-service\"},\n  \t\tClusterIP: \"172.30.104.79\",\n  \t\t... // 16 identical fields\n  \t},\n  \tStatus: {},\n  }\n"}
% grep -e 'updated internal service' -- ingress-operator.log | sed -e 's/^[^{]*//' | jq -r .diff
  &v1.Service{
        TypeMeta:   {},
        ObjectMeta: {Name: "router-internal-reconcile-internal-service", Namespace: "openshift-ingress", UID: "aaa07625-e508-435c-b8e8-97891c17cd17", ResourceVersion: "38668", ...},
        Spec: v1.ServiceSpec{
                Ports: []v1.ServicePort{
                        {Name: "http", Protocol: "TCP", Port: 80, TargetPort: {Type: 1, StrVal: "http"}, ...},
                        {Name: "https", Protocol: "TCP", Port: 443, TargetPort: {Type: 1, StrVal: "https"}, ...},
                        {
                                ... // 2 identical fields
                                AppProtocol: nil,
                                Port:        1936,
                                TargetPort: intstr.IntOrString{
-                                       Type:   0,
+                                       Type:   1,
-                                       IntVal: 12345,
+                                       IntVal: 0,
-                                       StrVal: "",
+                                       StrVal: "metrics",
                                },
                                NodePort: 0,
                        },
                },
                Selector:  {"ingresscontroller.operator.openshift.io/deployment-ingresscontroller": "reconcile-internal-service"},
                ClusterIP: "172.30.104.79",
                ... // 16 identical fields
        },
        Status: {},
  }

% 
```
No spurious updates, and the diff for the expected update is now cleaner.

test/e2e/operator_test.go
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/cfba4d9ad778d01a6ab874682ad0c2d95f81eab0..7599a46d8ecfee4b398bd5f228aceb1040c2c165.

test/e2e/operator_test.go
Miciah: https://github.com/openshift/cluster-ingress-operator/compare/cfba4d9ad778d01a6ab874682ad0c2d95f81eab0..7599a46d8ecfee4b398bd5f228aceb1040c2c165 moves `"github.com/stretchr/testify/assert"` to be with `"github.com/go-logr/logr"`.

ShudiLi:
tested it with 4.16.0-0.ci.test-2024-05-06-015928-ci-ln-ktjxhtk-latest

```
1.
% oc get clusterversion
version   4.16.0-0.ci.test-2024-05-06-015928-ci-ln-ktjxhtk-latest   True        False         48m     Cluster version is 4.16.0-0.ci.test-2024-05-06-015928-ci-ln-ktjxhtk-latest

2.
% oc -n openshift-ingress-operator get ingresscontroller int1 -o=jsonpath="{.spec.endpointPublishingStrategy}" 
{"hostNetwork":{"httpPort":23080,"httpsPort":23443,"statsPort":23936},"type":"HostNetwork"}%  

3.
%  oc -n openshift-ingress get svc router-internal-int1 -o=jsonpath="{.spec.sessionAffinity}"
None%                               

4.
% oc -n openshift-ingress get svc router-internal-int1                        
NAME                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)                   AGE
router-internal-int1   ClusterIP   172.30.126.119   <none>        80/TCP,443/TCP,1936/TCP   6m11s

5.
% oc -n openshift-ingress get ep router-internal-int1 -oyaml | grep -A9 ports:
  ports:
  - name: metrics
    port: 23936
    protocol: TCP
  - name: https
    port: 23443
    protocol: TCP
  - name: http
    port: 23080
    protocol: TCP

6.
% oc -n openshift-ingress get pods -owide
NAME                              READY   STATUS    RESTARTS   AGE   IP             NODE                                       NOMINATED NODE   READINESS GATES
router-default-6cb744b766-7chnx   1/1     Running   0          69m   10.38.134.67   ci-ln-ktjxhtk-c1627-29sp9-worker-0-h8kgr   <none>           <none>
router-default-6cb744b766-t8ztx   1/1     Running   0          69m   10.38.134.26   ci-ln-ktjxhtk-c1627-29sp9-worker-0-s45kl   <none>           <none>
router-int1-5b599bf97d-bprrw      1/1     Running   0          32m   10.38.134.26   ci-ln-ktjxhtk-c1627-29sp9-worker-0-s45kl   <none>           <none>
router-int1-5b599bf97d-wnnvn      1/1     Running   0          32m   10.38.134.67   ci-ln-ktjxhtk-c1627-29sp9-worker-0-h8kgr   <none>           <none>

7.
% oc debug node/ci-ln-ktjxhtk-c1627-29sp9-worker-0-s45kl
Starting pod/ci-ln-ktjxhtk-c1627-29sp9-worker-0-s45kl-debug-vldm5 ...
To use host binaries, run `chroot /host`
chroot /hostPod IP: 10.38.134.26
If you don't see a command prompt, try pressing enter.
sh-5.1# chroot /host
sh-5.1# netstat -ant | grep 10.38.134.26:23936
tcp6       0      0 10.38.134.26:23936      10.38.134.12:46452      ESTABLISHED
tcp6       0      0 10.38.134.26:23936      10.131.0.15:53944       ESTABLISHED
sh-5.1#

8. update the ports
% oc -n openshift-ingress get ep router-internal-int1 -oyaml | grep -A9 ports:
  ports:
  - name: https
    port: 24443
    protocol: TCP
  - name: http
    port: 24080
    protocol: TCP
  - name: metrics
    port: 24936
    protocol: TCP

10. create two route
% oc get route
NAME            HOST/PORT                                                                           PATH   SERVICES        PORT   TERMINATION   WILDCARD
edge1           edge1.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com ... 1 more                  unsec-server3   http   edge          None
unsec-server3   unsec-server3.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com ... 1 more          unsec-server3   http                 None

11. curl the two routes
sh-4.4# curl http://unsec-server3.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com --resolve unsec-server3.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com:80:10.38.134.26
this a test!
sh-4.4# 
sh-4.4# curl https://edge1.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com  --resolve edge1.int1.ci-ln-ktjxhtk-c1627.vmc-ci.devcluster.openshift.com:443:10.38.134.26 -k
this a test!
sh-4.4#
```

ShudiLi:
gcp cluster:
% oc -n openshift-ingress get svc router-default  -o=jsonpath="{.spec.sessionAffinity}"
None%                           

/label qe-approved
thanks

Miciah:
@candita, are there still any changes needed for this PR?  I believe I addressed https://github.com/openshift/cluster-ingress-operator/pull/1052#discussion_r1589770740 and https://github.com/openshift/cluster-ingress-operator/pull/1052#discussion_r1589741825, but please let me know if further changes are needed.

candita:
Failure to reach cache-openshift-api-reused-connections:

/test e2e-aws-ovn

> time="2024-05-04T02:30:03Z" level=error msg="disruption sample failed: error running request: 503 Service Unavailable: error trying to reach service: dial tcp 10.129.0.53:8443: connect: connection refused\n" auditID=4ac891b4-567c-4463-a32c-5931731920b1 backend=cache-openshift-api-reused-connections this-instance="{Disruption map[backend-disruption-name:cache-openshift-api-reused-connections connection:reused disruption:openshift-tests]}" type=reused
> time="2024-05-04T02:30:03Z" level=error msg="disruption sample failed: error running request: 503 Service Unavailable: error trying to reach service: dial tcp 10.129.0.53:8443: connect: connection refused\n" auditID=fc918ec4-fea3-4023-b3ac-7867922a3e8b backend=openshift-api-reused-connections this-instance="{Disruption map[backend-
disruption-name:openshift-api-reused-connections connection:reused disruption:openshift-tests]}" type=reused

candita:
Maybe this particular test should have been skipped on SNO?

> time="2024-05-04T03:15:47Z" level=error msg="  Finished CollectData for [Jira:\"Networking / router\"] monitor test service-type-load-balancer-availability collection with not-supported error" error="not supported: insufficient nodes for service load balancers"

/test e2e-aws-ovn-single-node

test/e2e/operator_test.go
```diff
@@ -4179,3 +4216,63 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
...
+		if err := kclient.Get(context.TODO(), serviceName, service); err != nil {
+			t.Log(err)
+			return false, nil
+		}
+		if metricsPort := findPort("metrics", service); metricsPort == nil {
+			t.Fatalf(`"metrics" port was removed from the service's ports: %v`, service.Spec.Ports)
+		} else if metricsPort.TargetPort == originalMetricsPortTargetPort {
+			return true, nil
+		}
+		return false, nil
```
candita, line 4274: nit: Instead of nil, return an error here that indicates the target port changed when it wasn't expected to?

candita:
> @candita, are there still any changes needed for this PR? I believe I addressed [#1052 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/1052#discussion_r1589770740) and [#1052 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/1052#discussion_r1589741825), but please let me know if further changes are needed.

Just one more nit and waiting for CI.  (Although I don't really think it's the best idea to sync the name between IC and service, but allow the port numbers to be unsynced because they are not used if a name is present.)

test/e2e/operator_test.go
Miciah: The main reason to use a polling loop here is because the test is verifying that the operator reverts the update: https://github.com/openshift/cluster-ingress-operator/blob/add17d757c103eab60879424b60b42a9448cd401/test/e2e/operator_test.go#L4222-L4224

If the target port is set to `originalMetricsPortTargetPort`, then the operator presumably has reverted the update; else the test most likely needs to wait a little longer: https://github.com/openshift/cluster-ingress-operator/blob/add17d757c103eab60879424b60b42a9448cd401/test/e2e/operator_test.go#L4235-L4238

Returning a non-nil error here would make the test flaky; if the polling loop gets the updated service before the operator manages to revert the update, we should retry, not return an error.

Miciah:
> (Although I don't really think it's the best idea to sync the name between IC and service, but allow the port numbers to be unsynced because they are not used if a name is present.)

Can you explain in more detail what you mean?  What do you mean by "allow the port numbers to be unsynced"?  I think the current approach is the simplest possible approach:

- The router container has [a port named "metrics"](https://github.com/openshift/cluster-ingress-operator/blob/5b4f6283f48046b85f60bf9d75fa5c9222329cd1/pkg/operator/controller/ingress/deployment.go#L690-L694).  This enables the container port to be targeted by name.  
- The operator [sets the container's "metrics" port's container port number](https://github.com/openshift/cluster-ingress-operator/blob/5b4f6283f48046b85f60bf9d75fa5c9222329cd1/pkg/operator/controller/ingress/deployment.go#L721-L729) to the value of `IngressController.spec.endpointPublishingStrategy.hostNetwork.statsPort`; it also [sets the "metrics" port's host port to match](https://github.com/openshift/cluster-ingress-operator/blob/5b4f6283f48046b85f60bf9d75fa5c9222329cd1/pkg/operator/controller/ingress/deployment.go#L736).  This is a hard requirement to implement the API and enable multiple host-network router pods to co-exist on a node, but it means that the port number can change.  
- The operator [creates](https://github.com/openshift/cluster-ingress-operator/blob/5b4f6283f48046b85f60bf9d75fa5c9222329cd1/pkg/manifests/assets/router/service-internal.yaml#L18-L21) or (with this PR) [updates](https://github.com/openshift/cluster-ingress-operator/blob/add17d757c103eab60879424b60b42a9448cd401/pkg/operator/controller/ingress/internal_service.go#L152) the service to have a "metrics" port that targets the container's "metrics" port by name.  Using the port name (rather than port number) avoids the need to update the port number on the service when the container's "metrics" port number changes.  
- The [servicemonitor targets the service's "metrics" port](https://github.com/openshift/cluster-ingress-operator/blob/5b4f6283f48046b85f60bf9d75fa5c9222329cd1/pkg/operator/controller/ingress/monitoring.go#L88C1-L88C36) by name.  Again, because the servicemonitor uses the port _name_, the servicemonitor does not need to be updated if the port _number_ changes.

I sketched out a diagram to depict the relationships between these resources, taking into account the changes in this PR (this might not be the best type of diagram, and I may have made some errors, but I hope this diagram is helpful):
```mermaid
erDiagram
    IngressController ||--|| Deployment : "creates/updates/deletes"
    IngressController ||--|| Service : "creates/updates/deletes"
    IngressController ||--|| ServiceMonitor : "creates/updates/deletes"
    Deployment ||--o{ Pod : "creates/updates/deletes"
    Service ||--o{ Pod : "targets"
    ServiceMonitor ||--|| Service : "targets"
    IngressController {
        int32 statsPort "1936"
    }
    Pod {
        string name "metrics"
        int32 containerPort "1936"
        int32 hostPort "1936"
    }
    Service {
        string name "metrics"
        int32 port "1936"
        string targetPort "metrics"
    }
    ServiceMonitor {
        string port "metrics"
    }
```

candita:
My concern is not major, so I will tag. It is, exactly as you said:
> Using the port name (rather than port number) avoids the need to update the port number on the service when the container's "metrics" port number changes.
So the port number can look different on the service, and fool a human observer into thinking there might be an error.

/lgtm
/approve

candita:
The severity is critical on this bug.

Miciah:
> So the port number can look different on the service, and fool a human observer into thinking there might be an error.

Ah, are you suggesting that someone might confuse the service port number (_not_ the target port) and the container port number, or falsely believe that they need to match?  And so you're suggesting that the operator should update the service port number to match the container port number?  

I think making that change would be a bad idea for a couple reasons.  First, it would only re-enforce the confusion, or even create confusion where it didn't exist before.  After all, someone might think, why is the operator setting the service's port number to match the container's port number if it isn't necessary to do so?  Second, the change would incur an extra update every time `statsPort` were updated.  Rather than degrading the product to skirt around the confusion, it would be better to address the confusion directly by referring people to [the Kubernetes API documentation for service ports](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#serviceport-v1-core) or [Kubernetes documentation for services](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service).

candita:
Some new failures:
 
 --- FAIL: TestAll/parallel/TestRouteMetricsControllerRouteAndNamespaceSelector (0.03s)
 --- FAIL: TestAll/parallel/TestRouteMetricsControllerOnlyNamespaceSelector (0.02s)
 --- FAIL: TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector (0.02s)
 
All based on a missing prometheus service account - that's new.

=== CONT  TestAll/parallel/TestRouteMetricsControllerRouteAndNamespaceSelector
    route_metrics_test.go:86: prometheus service account not found
=== CONT  TestAll/parallel/TestRouteMetricsControllerOnlyNamespaceSelector
    route_metrics_test.go:86: prometheus service account not found
=== CONT  TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector
    route_metrics_test.go:86: prometheus service account not found

Miciah:
> Some new failures:
> 
> --- FAIL: TestAll/parallel/TestRouteMetricsControllerRouteAndNamespaceSelector (0.03s) --- FAIL: TestAll/parallel/TestRouteMetricsControllerOnlyNamespaceSelector (0.02s) --- FAIL: TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector (0.02s)
> 
> All based on a missing prometheus service account - that's new.
> 
> === CONT TestAll/parallel/TestRouteMetricsControllerRouteAndNamespaceSelector route_metrics_test.go:86: prometheus service account not found === CONT TestAll/parallel/TestRouteMetricsControllerOnlyNamespaceSelector route_metrics_test.go:86: prometheus service account not found === CONT TestAll/parallel/TestRouteMetricsControllerOnlyRouteSelector route_metrics_test.go:86: prometheus service account not found

Yup, [the e2e-aws-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/1790830697687027712), [e2e-azure-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1790830697724776448), and [e2e-gcp-operator jobs](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/978/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1790760775913050112) all failed with those errors.  [search.ci](https://search.dptools.openshift.org/?search=FAIL%3A+TestAll%2Fparallel%2FTestRouteMetricsControllerRouteAndNamespaceSelector&maxAge=48h&context=1&type=build-log&name=cluster-ingress-operator&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) shows failures for this PR as well as [an e2e-gcp-operator failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/978/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1790760775913050112) for #978 and a rehearse job ([rehearse-52060-pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview](https://prow.ci.openshift.org/job-history/test-platform-results/pr-logs/directory/rehearse-52060-pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview)).  This needs to be investigated.

gcs278:
FYI I'm testing out a bump to openshift/library-go https://github.com/openshift/cluster-ingress-operator/pull/1054 to see if that fixes things. :crossed_fingers: 

The bump incorporates https://github.com/openshift/library-go/pull/1697 which switches the Prometheus API client from the legacy service API to the TokenRequest API. I'm not familiar with these APIs, but it's seems like a suspect.

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1054 fixes the current ingress operator E2E jobs related to the Prometheus service account. More info in  https://issues.redhat.com/browse/OCPBUGS-33792

candita:
@Miciah while this isn't related to internal service - I'm also seeing some fields change in router deployments and node ports that maybe shouldn't.  I'll add to the bug and if you'd rather I opened a new one, I can do that instead.

Miciah:
> @Miciah while this isn't related to internal service - I'm also seeing some fields change in router deployments and node ports that maybe shouldn't. I'll add to the bug and if you'd rather I opened a new one, I can do that instead.

I looked at [the last run of the e2e-gcp-operator job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1790830697762525184).  The NodePort service diff appears as expected:
```text
% curl -s 'https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1790830697762525184/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-784b4dcb5-k8kvt_ingress-operator.log' -o ingress-operator.log                                
% sed -ne '/updated NodePort service/ s/^[^{]*//p' -- ingress-operator.log | jq -r .diff       
  &v1.Service{
        TypeMeta: {},
        ObjectMeta: v1.ObjectMeta{
                ... // 9 identical fields
                DeletionGracePeriodSeconds: nil,
                Labels:                     {"app": "router", "ingresscontroller.operator.openshift.io/owning-ingresscontroller": "local-with-fallback", "router": "router-nodeport-local-with-fallback"},
-               Annotations:                map[string]string{"traffic-policy.network.alpha.openshift.io/local-with-fallback": ""},
+               Annotations:                map[string]string{},
                OwnerReferences:            {{APIVersion: "apps/v1", Kind: "Deployment", Name: "router-local-with-fallback", UID: "32fc3ffb-a2b1-478c-82ca-c156aa6081ba", ...}},
                Finalizers:                 nil,
                ManagedFields:              {{Manager: "ingress-operator", Operation: "Update", APIVersion: "v1", Time: s"2024-05-15 20:35:02 +0000 UTC", ...}},
        },
        Spec:   {Ports: {{Name: "http", Protocol: "TCP", Port: 80, TargetPort: {Type: 1, StrVal: "http"}, ...}, {Name: "https", Protocol: "TCP", Port: 443, TargetPort: {Type: 1, StrVal: "https"}, ...}, {Name: "metrics", Protocol: "TCP", Port: 1936, TargetPort: {Type: 1, StrVal: "metrics"}, ...}}, Selector: {"ingresscontroller.operator.openshift.io/deployment-ingresscontroller": "local-with-fallback"}, ClusterIP: "172.30.69.206", ClusterIPs: {"172.30.69.206"}, ...},
        Status: {},
  }

% 
```
Are you seeing something else?

For deployments, the diffs have some noise for the volume `DefaultModes` parameter and for various probe parameters.  I can push a commit like https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/1badfc6baeec7156e35e21223961a40211f499a3 to fix those if you like.

Miciah:
> For deployments, the diffs have some noise for the volume `DefaultModes` parameter and for various probe parameters. I can push a commit like [1badfc6](https://github.com/openshift/cluster-ingress-operator/commit/1badfc6baeec7156e35e21223961a40211f499a3) to fix those if you like.

As this PR is blocked in #1054 anyway, I went ahead and pushed commits to fix these bogus diffs.  I'll check the diffs in the ingress-operator logs when we get back the results from a new CI run.

Miciah:
[The ingress-operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1791303594599780352/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-5866497cdc-tbbt6_ingress-operator.log) from [the last run of the e2e-gcp-operator job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1791303594599780352) still has some noise from `httpGet.scheme`.  https://github.com/openshift/cluster-ingress-operator/compare/b591332b65d3c51df6ff34e26b2cf01f2b7cce09..4d0f52090379819b1b50af60d060d167a1629ba1 modifies the deployment update logic's `copyProbe` function to eliminate this noise.

Miciah:
I realized that I was still missing `defaultMode` for some configmap and secret volumes, and also that the unit tests were rather lax (and in one case misleading, checking for a nonexistent "error-pages" volume).  https://github.com/openshift/cluster-ingress-operator/compare/4d0f52090379819b1b50af60d060d167a1629ba1..2f0e7872151af6cd62bc88ad26403d166f1259c9 addresses `defaultMode` for some additional volumes and improves test coverage.

candita:
> > For deployments, the diffs have some noise for the volume `DefaultModes` parameter and for various probe parameters. I can push a commit like [1badfc6](https://github.com/openshift/cluster-ingress-operator/commit/1badfc6baeec7156e35e21223961a40211f499a3) to fix those if you like.
> 
> As this PR is blocked on #1054 anyway, I went ahead and pushed commits to fix these bogus diffs. I'll check the diffs in the ingress-operator logs when we get back the results from a new CI run.

I noticed your changes.  I think this is causing an issue in a 4.12 bug that I'm looking into.  Posted in slack.

candita:
NodePort diffs I am seeing in 4.12:

> - \t\tClusterIPs:               []string{\"10.20.134.56\"},\n
> + \t\tClusterIPs:               nil,\n  
> 
> - \t\tSessionAffinity:          \"None\",\n
> + \t\tSessionAffinity:          \"\",\n  
> 
> - \t\tIPFamilies:                    []v1.IPFamily{\"IPv4\"},\n
> + \t\tIPFamilies:                    nil,\n
> 
> - \t\tIPFamilyPolicy:                &\"SingleStack\",\n
> + \t\tIPFamilyPolicy:                nil,\n  
> 
> - \t\tInternalTrafficPolicy:         &\"Cluster\",\n
> + \t\tInternalTrafficPolicy:         nil,\n

candita:
Looks like nodes never became available:

> level=error msg=ReadyIngressNodesAvailable: Authentication requires functional ingress which requires at least one schedulable and ready node. Got 0 worker nodes, 3 master nodes, 0 custom target nodes (none are schedulable or ready for ingress pods).

/test e2e-gcp-ovn

candita:
Regarding this failure in e2e-hypershift:

> {"level":"error","ts":"2024-05-17T10:20:44Z","msg":"failed to setup shared OIDC provider","error":"failed to create OIDC provider: LimitExceeded: Cannot exceed quota for OpenIdConnectProvidersPerAccount: 100\n\tstatus code: 409, request id: 9f217247-4e4b-456b-9ab8-a8484f6ccb08","stacktrace":"github.com/openshift/hypershift/test/e2e.main\n\t/go/src/github.com/openshift/hypershift/test/e2e/e2e_test.go:170\ngithub.com/openshift/hypershift/test/e2e.TestMain\n\t/go/src/github.com/openshift/hypershift/test/e2e/e2e_test.go:136\nmain.main\n\t_testmain.go:69\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:267"}

Slack thread discussing quota related issues for Hypershift: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1715956957922319

Miciah:
> > > For deployments, the diffs have some noise for the volume `DefaultModes` parameter and for various probe parameters. I can push a commit like [1badfc6](https://github.com/openshift/cluster-ingress-operator/commit/1badfc6baeec7156e35e21223961a40211f499a3) to fix those if you like.
> > 
> > 
> > As this PR is blocked on #1054 anyway, I went ahead and pushed commits to fix these bogus diffs. I'll check the diffs in the ingress-operator logs when we get back the results from a new CI run.
> 
> I noticed your changes. I think this is causing an issue in a 4.12 bug that I'm looking into. Posted in slack.
> 
> Update: we may need to backport to 4.12.

Right, it makes sense to backport at least https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/add17d757c103eab60879424b60b42a9448cd401 to 4.12 as #864 was backported to 4.12.

> Internal service and NodePort diffs I am seeing in 4.12:
> 
> ```
> - \t\tClusterIPs:               []string{\"10.20.134.56\"},\n
> + \t\tClusterIPs:               nil,\n  
> 
> - \t\tSessionAffinity:          \"None\",\n
> + \t\tSessionAffinity:          \"\",\n  
> 
> - \t\tIPFamilies:                    []v1.IPFamily{\"IPv4\"},\n
> + \t\tIPFamilies:                    nil,\n
> 
> - \t\tIPFamilyPolicy:                &\"SingleStack\",\n
> + \t\tIPFamilyPolicy:                nil,\n  
> ```

Right, these should be fixed by https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/d21abc8322a311d920d0be97dd553854826e7b5e and https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/1badfc6baeec7156e35e21223961a40211f499a3, so we can backport those too.  

> ```
> - \t\tInternalTrafficPolicy:         &\"Cluster\",\n
> + \t\tInternalTrafficPolicy:         nil,\n  
> ```

This should have been fixed by https://github.com/openshift/cluster-ingress-operator/pull/927/commits/5579aa1b3573a4d9479a924a1ae78688e2105c5c, which is proposed for backport to 4.13.  If you're seeing [OCPBUGS-13190](https://issues.redhat.com/browse/OCPBUGS-13190) on 4.12, please add 4.12 to OCPBUGS-13190's "Affects Versions", and we can propose backporting #927 to 4.12 as well.  

---

The diffs in [the ingress-operator logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1791470779821461504/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-57f7fb8f95-8p27t_ingress-operator.log) from [the latest e2e-gcp-operator run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1791470779821461504) look really clean now.  They are too large to paste in a comment, but you can review them using the following commands:
```text
% curl -s 'https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1052/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1791470779821461504/artifacts/e2e-gcp-operator/gather-extra/artifacts/pods/openshift-ingress-operator_ingress-operator-57f7fb8f95-8p27t_ingress-operator.log' -o ingress-operator.log
% sed < ingress-operator.log -n -e $'/diff/ {s/\\\\n/\\\n/g;s/\\\\t/\t/g;p;}'
```

However, three anomalies jumped out at me.  First, the LoadBalancer-type service for `TestAllowedSourceRanges` has an inexplicable, delayed second update:
```text
2024-05-17T15:07:39.258Z        INFO    operator.ingress_controller     ingress/load_balancer_service.go:321    updated load balancer service   {"namespace": "openshift-ingress", "name": "router-sourcerange", "diff": "  &v1.Service{
        TypeMeta:   {},
        ObjectMeta: {Name: \"router-sourcerange\", Namespace: \"openshift-ingress\", UID: \"bf35e2d3-bb10-498e-b771-0659315fc65e\", ResourceVersion: \"44908\", ...},
        Spec: v1.ServiceSpec{
                ... // 6 identical fields
                SessionAffinity:          \"None\",
                LoadBalancerIP:           \"\",
-               LoadBalancerSourceRanges: []string{\"127.0.0.0/8\"},
+               LoadBalancerSourceRanges: []string{\"10.0.0.0/8\"},
                ExternalName:             \"\",
                ExternalTrafficPolicy:    \"Local\",
                ... // 8 identical fields
        },
        Status: {LoadBalancer: {Ingress: {{IP: \"34.29.9.29\"}}}},
  }
"}
2024-05-17T15:25:51.134Z        INFO    operator.ingress_controller     ingress/load_balancer_service.go:321    updated load balancer service   {"namespace": "openshift-ingress", "name": "router-sourcerange", "diff": "  &v1.Service{
        TypeMeta:   {},
        ObjectMeta: {Name: \"router-sourcerange\", Namespace: \"openshift-ingress\", UID: \"021a0f6b-e316-4d0a-9aaa-15189dded4f6\", ResourceVersion: \"58305\", ...},
        Spec: v1.ServiceSpec{
                ... // 6 identical fields
                SessionAffinity:          \"None\",
                LoadBalancerIP:           \"\",
-               LoadBalancerSourceRanges: []string{\"127.0.0.0/8\"},
+               LoadBalancerSourceRanges: []string{\"10.0.0.0/8\"},
                ExternalName:             \"\",
                ExternalTrafficPolicy:    \"Local\",
                ... // 8 identical fields
        },
        Status: {LoadBalancer: {Ingress: {{IP: \"34.16.7.141\"}}}},
  }
"}
```
The second update happens about 18 minutes after the first update.  It also happens after the serial tests appear to have finished.  I suspect it has something to do with the fact that `TestAllowedSourceRanges` uses `t.Cleanup` to delete its IngressController; maybe the cleanup is deleted for that reason, though that doesn't fully explain why the service gets updated a second time (it should just get deleted when the cleanup happens).  

The second anomaly that I noticed was this update from the `TestRouteHardStopAfterEnableOnIngressConfig` test:
```text
                                                        {
                                                                Name:      \"ROUTER_DISABLE_HTTP2\",
-                                                               Value:     \"false\",
+                                                               Value:     \"true\",
                                                                ValueFrom: nil,
                                                        },
                                                        {Name: \"ROUTER_DISABLE_NAMESPACE_OWNERSHIP_CHECK\", Value: \"false\"},
                                                        {Name: \"ROUTER_DOMAIN\", Value: \"apps.ci-op-tg7xmq00-76b3b.origin-ci-int-gce.dev.rhcloud.com\"},
+                                                       {Name: \"ROUTER_HARD_STOP_AFTER\", Value: \"3h20m0s\"},
```
The "3h20m" value has to be from this update: https://github.com/openshift/cluster-ingress-operator/blob/487fe81f1a6b70e17e304b0ede0de849cb7bfac0/test/e2e/hard_stop_after_test.go#L49

However, `TestRouteHardStopAfterEnableOnIngressConfig` is a serial test, and so are `TestRouteHTTP2EnableAndDisableIngressController` and `TestRouteHTTP2EnableAndDisableIngressConfig`, so it isn't clear to me why a single update would modify both `ROUTER_DISABLE_HTTP2` and `ROUTER_HARD_STOP_AFTER`.  @frobware, any thoughts on this?

The third anomaly I noticed was this:
```text
% grep -e 'successfully updated Infra CR with Ingress Load Balancer IPs' -m 1 -- ingress-operator.log       
2024-05-17T14:46:01.434Z	INFO	operator.ingress_controller	ingress/controller.go:326	successfully updated Infra CR with Ingress Load Balancer IPs
% grep -e 'successfully updated Infra CR with Ingress Load Balancer IPs' -c -- ingress-operator.log 
142
```
#1016 has a logic error, which causes the operator to log this message even when it didn't do an update: https://github.com/openshift/cluster-ingress-operator/blob/009644a6b197b67f074cc34a07868ef01db31510/pkg/operator/controller/ingress/controller.go#L1135-L1145
(I've actually noticed this before, but I never got around to filing a bug report...). #1016 was backported to 4.15, so it would be nice to fix it and backport the fix to 4.15.  It's just log noise, though, so it might not be justifiable to backport on its own.  It is rather noisy, though, and it's trivial to fix, so if you like, I can bundle a fix for this log message into this PR.

ShudiLi:
/label qe-approved

tested it 4.16.0-0.ci.test-2024-05-20-061242-ci-ln-qyfl0pt-latest

1.
% oc get clusterversion
version   4.16.0-0.ci.test-2024-05-20-061242-ci-ln-qyfl0pt-latest   True        False         10m     Cluster version is 4.16.0-0.ci.test-2024-05-20-061242-ci-ln-qyfl0pt-latest

2.
% oc create -f ingress-hostnetwork.yaml 
ingresscontroller.operator.openshift.io/int1 created
% cat ingress-hostnetwork.yaml                                                         
kind: IngressController
metadata:
  name: int1
spec:
  endpointPublishingStrategy:
    hostNetwork:
      httpPort: 23080
      httpsPort: 23443
      statsPort: 23936
    type: HostNetwork
  domain: int1.ci-ln-qyfl0pt-c1627.vmc-ci.devcluster.openshift.com

3.
NAME                             READY   STATUS    RESTARTS   AGE
router-default-74f5cbcf6-qgtqx   1/1     Running   0          38m
router-default-74f5cbcf6-t742h   1/1     Running   0          38m
router-int1-c56c7dd96-4b72b      1/1     Running   0          11m
router-int1-c56c7dd96-4nvb5      1/1     Running   0          11m

4.
% oc -n openshift-ingress get svc router-internal-int1 -o=jsonpath="{.spec.ports}"
[{"name":"http","port":80,"protocol":"TCP","targetPort":"http"},{"name":"https","port":443,"protocol":"TCP","targetPort":"https"},{"name":"metrics","port":1936,"protocol":"TCP","targetPort":"metrics"}]% 

5.
 % oc -n openshift-ingress get ep router-internal-int1 -o=jsonpath="{.subsets[0].ports}"
[{"name":"metrics","port":23936,"protocol":"TCP"},{"name":"https","port":23443,"protocol":"TCP"},{"name":"http","port":23080,"protocol":"TCP"}]% 

6.
% oc -n openshift-ingress get deployment router-int1 -oyaml | grep -E -A9 "ivenessProbe:|readinessProbe:|startupProbe:"
        livenessProbe:
          failureThreshold: 3
          httpGet:
            host: localhost
            path: /healthz
            port: 23936
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          terminationGracePeriodSeconds: 10
--
        readinessProbe:
          failureThreshold: 3
          httpGet:
            host: localhost
            path: /healthz/ready
            port: 23936
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
--
        startupProbe:
          failureThreshold: 120
          httpGet:
            host: localhost
            path: /healthz/ready
            port: 23936
            scheme: HTTP
          periodSeconds: 1
          successThreshold: 1
          timeoutSeconds: 1

7.
% oc -n openshift-ingress get deployment router-int1 -oyaml | grep -A6 " - configMap:"
      - configMap:
          defaultMode: 420
          items:
          - key: service-ca.crt
            path: service-ca.crt
          name: service-ca-bundle
          optional: false

8.
% oc -n openshift-ingress get svc router-internal-int1 -o=jsonpath="{.spec.sessionAffinity}" 
None%                    

9.  create a load balancer ingress controller and check its service
% print $(oc -n openshift-ingress get svc router-int2 -o=jsonpath="{.spec.type}\n{.spec.sessionAffinity}")
LoadBalancer
None

candita:
@Miciah thanks for the thorough analysis and the bug numbers we should plan to backport to fix all the spurious updates seen in 4.12.40.  At the moment I think we can safely ignore the logs from CI testing.  I have a summary of all the spurious updates seen in 4.12.40 and will coordinate the backports of this bug and any other new one needed, including one to address https://github.com/openshift/cluster-ingress-operator/pull/1016

pkg/manifests/assets/router/deployment.yaml
candita, line 1: Thanks! That is a lot of overlooked deployment updates.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1514,7 +1520,7 @@ func hashableProbe(probe *corev1.Probe) *corev1.Probe {
 
 	var hashableProbe corev1.Probe
 
-	copyProbe(probe, &hashableProbe)
+	copyProbe(probe, &hashableProbe, false)
```
candita, line 1523: This change will make a big difference.

Miciah:
> we should plan to backport to fix all the spurious updates seen in 4.12.40. 

There have been a number of fixes for spurious updates since 4.12 (though not all of these have been confirmed to affect 4.12 generally or 4.12.40 specifically):

- #879 to fix [OCPBUGS-6701](https://issues.redhat.com/browse/OCPBUGS-6701).  
- #927 to fix [OCPBUGS-13190](https://issues.redhat.com/browse/OCPBUGS-13190).  
- #947 to fix [OCPBUGS-14995](https://issues.redhat.com/browse/OCPBUGS-14995).
- #948 to fix [OCPBUGS-14994](https://issues.redhat.com/browse/OCPBUGS-14994).
- #1052 (this PR) to fix [OCPBUGS-23221](https://issues.redhat.com//browse/OCPBUGS-23221) and other issues.  

> At the moment I think we can safely ignore the logs from CI testing.

I don't understand what you mean by that.  I think we should always be looking at the logs from CI testing, especially when we are modifying update logic, and most especially when we are specifically trying to address spurious updates.  Do you mean that we should ignore the three issues that I mentioned in https://github.com/openshift/cluster-ingress-operator/pull/1052#issuecomment-2118548732 for now and not try to fix the "successfully updated Infra CR with Ingress Load Balancer IPs" log message or diagnose the other anomalies?  

> I have a summary of all the spurious updates seen in 4.12.40 and will coordinate the backports of this bug and any other new one needed, including one to address #1016

All right, I will hold off on fixing the "successfully updated Infra CR with Ingress Load Balancer IPs" issue until a separate bug report is filed.  

In my opinion, we should backport in chronological order (#879 first, then #927, then #947, and so on) in order to keep the history consistent among branches and avoid conflicts as much as possible, but we will encounter merge conflicts along the way in any case with some of these PRs (certainly with #1052).

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1645,7 +1652,7 @@ func copyProbe(a, b *corev1.Probe) {
 			Port: a.ProbeHandler.HTTPGet.Port,
 			Host: a.ProbeHandler.HTTPGet.Host,
```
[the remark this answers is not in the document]
Miciah: `Port` is required, and the default value for `Host` is the empty string, so I believe we can copy them unconditionally.

candita:
> I don't understand what you mean by that. I think we should always be looking at the logs from CI testing, especially when we are modifying update logic, and most especially when we are specifically trying to address spurious updates. Do you mean that we should ignore the three issues that I mentioned in [#1052 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/1052#issuecomment-2118548732) for now and not try to fix the "successfully updated Infra CR with Ingress Load Balancer IPs" log message or diagnose the other anomalies?

Yes, that is what I meant.

candita:
Three bugs created for the anomalies listed in  https://github.com/openshift/cluster-ingress-operator/pull/1052#issuecomment-2118548732 :

- https://issues.redhat.com/browse/OCPBUGS-34411
- https://issues.redhat.com/browse/OCPBUGS-34412
- https://issues.redhat.com/browse/OCPBUGS-34413

pkg/operator/controller/ingress/deployment.go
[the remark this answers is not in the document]
candita: In that case, if you wouldn't mind, can you add a comment like `Values that are always updated` and group them, and another comment `Values that are only updated when requested` and group them all under a
`if copyDefaultValues {`.  I feel like this may help/force future developers to make a decision about what to do with values in `copyProbe`.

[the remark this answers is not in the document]
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/pull/1052/commits/3faf391d2d43c63526b256e5f83b163673d8ef9a.

candita:
E2E test failure:

> == NAME  TestAll/parallel/TestIngressControllerNamespaceSelectorUpdateShouldClearRouteStatus
>     router_status_test.go:248: failed to update ingresscontroller: Operation cannot be fulfilled on ingresscontrollers.operator.openshift.io "ic-namespace-selector-test": the object has been modified; please apply your changes to the latest version and try again

/test e2e-aws-operator

PR #1053, opened 2024-05-13
OCPBUGS-29894: Add test verifying that routers without the required CRLs are marked not ready

rfredette:
The new test should fail until https://github.com/openshift/router/pull/595 is merged.
This is part of the fix for OCPBUGS-29894

### DIFF
--- a/test/e2e/client_tls_test.go
+++ b/test/e2e/client_tls_test.go
@@ -1383,6 +1383,221 @@ func TestCRLUpdate(t *testing.T) {
 	}
 }
 
+// TestRouterWaitsForCRLs verifies that router pods only report ready once all CRLs are downloaded.
+func TestRouterWaitsForCRLs(t *testing.T) {
+	t.Parallel()
+	namespaceName := names.SimpleNameGenerator.GenerateName("router-waits-for-crls-")
+	namespace := createNamespace(t, namespaceName)
+
+	// Create certs.
+	crlHostName := types.NamespacedName{
+		Name:      "crl-host",
+		Namespace: namespaceName,
+	}
+	crlHostServiceName := "crl-host-service"
+	crlHostURL := crlHostServiceName + "." + crlHostName.Namespace + ".svc"
+
+	rootCDP := "http://" + crlHostURL + "/root.crl"
+	intermediateCDP := "http://" + crlHostURL + "/intermediate.crl"
+
+	rootCA := MustCreateTLSKeyCert("testing root CA", time.Now(), time.Now().Add(24*time.Hour), true, []string{rootCDP}, nil)
+	intermediateCA := MustCreateTLSKeyCert("testing intermediate CA", time.Now(), time.Now().Add(24*time.Hour), true, []string{intermediateCDP}, &rootCA)
+
+	rootCRL, rootCRLPem := MustCreateCRL(nil, rootCA, time.Now(), time.Now().Add(1*time.Hour), nil)
+	intermediateCRL, intermediateCRLPem := MustCreateCRL(nil, intermediateCA, time.Now(), time.Now().Add(1*time.Hour), nil)
+
+	// Create an ingress controller with mTLS enabled.
+	clientCAConfigMapName := "client-ca-cm"
+	clientCAConfigMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clientCAConfigMapName,
+			Namespace: "openshift-config",
+		},
+		Data: map[string]string{
+			"ca-bundle.pem": strings.Join([]string{intermediateCA.CertPem, rootCA.CertPem}, "\n"),
+		},
+	}
+	assertCreated(t, kclient, &clientCAConfigMap)
+	t.Cleanup(func() { assertDeleted(t, kclient, &clientCAConfigMap) })
+	icName := types.NamespacedName{
+		Name:      "router-waits-for-crls",
+		Namespace: operatorNamespace,
+	}
+	icDomain := icName.Name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, icDomain)
+	ic.Spec.ClientTLS = operatorv1.ClientTLS{
+		ClientCA: configv1.ConfigMapNameReference{
+			Name: clientCAConfigMapName,
+		},
+		ClientCertificatePolicy: operatorv1.ClientCertificatePolicyRequired,
+	}
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s/%s: %v", icName.Namespace, icName.Name, err)
+	}
+	t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ic) })
+
+	// Verify that the router is not ready, and that the reason includes that the "crls-updated" health check failed.
+	routerPod := corev1.Pod{}
+	err := wait.PollImmediate(10*time.Second, 5*time.Minute, func() (bool, error) {
+		podList := &corev1.PodList{}
+		labels := map[string]string{
+			controller.ControllerDeploymentLabel: icName.Name,
+		}
+		if err := kclient.List(context.TODO(), podList, client.InNamespace("openshift-ingress"), client.MatchingLabels(labels)); err != nil {
+			t.Logf("failed to list pods for ingress controllers %s: %v", ic.Name, err)
+			return false, nil
+		}
+		if len(podList.Items) == 0 {
+			t.Logf("no router pods found for ingresscontroller %s", ic.Name)
+			return false, nil
+		}
+		routerPod = podList.Items[0]
+		routerPodName := types.NamespacedName{
+			Name:      routerPod.Name,
+			Namespace: routerPod.Namespace,
+		}
+		if err := kclient.Get(context.TODO(), routerPodName, &routerPod); err != nil {
+			t.Logf("failed to get pod %s: %v", routerPod.Name, err)
+			return false, nil
+		}
+		for _, condition := range routerPod.Status.Conditions {
+			if condition.Type == corev1.PodReady {
+				if condition.Status != corev1.ConditionFalse {
+					t.Fatalf("expected %s %q condition to be False, got %s", routerPod.Name, corev1.PodReady, condition.Status)
+				}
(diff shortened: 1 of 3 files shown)


Miciah:
As noted in https://github.com/openshift/router/pull/595#pullrequestreview-2307848821, I think we need an alert.  Would it make sense to include that in this PR?

test/e2e/client_tls_test.go
```diff
@@ -889,8 +876,8 @@ func TestMTLSWithCRLs(t *testing.T) {
 				}
 				return false, nil
 			})
-			// Create CA cert bundle.
-			clientCAConfigmapName := "client-ca-cm-" + namespace.Name
+			// Create CA cert bundle
```
Miciah, line 879: Please keep the punctuation.  (This comment could also be a `t.Log` instead, but we can leave that change for another day.)

test/e2e/client_tls_test.go
```diff
@@ -905,7 +892,7 @@ func TestMTLSWithCRLs(t *testing.T) {
 			}
 			t.Cleanup(func() { assertDeleted(t, kclient, &clientCAConfigmap) })
 			icName := types.NamespacedName{
-				Name:      "mtls-with-crls",
+				Name:      names.SimpleNameGenerator.GenerateName("mtls-with-crls"),
```
Miciah, line 895: Slightly conspicuous that this prefix doesn't have a trailing "-"; is that intentional?

test/e2e/client_tls_test.go
```diff
@@ -774,10 +766,19 @@ func TestMTLSWithCRLs(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.Name, func(t *testing.T) {
-			tcCerts := tc.CreateCerts()
+			t.Parallel()
+			crlHostName := types.NamespacedName{
+				Name:      names.SimpleNameGenerator.GenerateName("crl-host-"),
+				Namespace: namespaceName,
+			}
+			// When generating certificates, the CRL distribution points need to be specified by URL
+			crlHostServiceName := names.SimpleNameGenerator.GenerateName("crl-host-service-")
```
Miciah, line 779: Is it useful to use a new generated name here, as opposed to using `crlHostName.Name + "-service"` or even just `crlHostName.Name`?  Similarly, I think `crlConfigMapName` below could be generated as `fmt.Sprintf("%s-%s", crlHostName.Name, name)`, and `clientCAConfigmapName` and `icName` could use the same prefix.  I imagine tracing the test or diagnosing failures would be easier if it used the same prefix when naming resources related to the same test case.  

Then there are `clientCertsConfigmap`, `podName`, `echoPod`, etc., but maybe it makes more sense for some of those to have distinct prefixes.  That is, the echo pod, echo route, and echo service can (and do) share a prefix, but a different prefix from `crlConfigMapName` etc.

If I'm missing something or you find the current approach easier to follow, feel free to ignore this suggestion.

test/e2e/client_tls_test.go
```diff
@@ -1222,15 +1191,33 @@ func TestCRLUpdate(t *testing.T) {
...
+			// When generating certificates, the CRL distribution points need to be specified by URL.
+			crlHostServiceName := names.SimpleNameGenerator.GenerateName("crl-host-service")
+			crlHostURL := crlHostServiceName + "." + crlHostName.Namespace + ".svc"
+			caCerts := tc.CreateCerts(crlHostURL)
 			// Generate CRLs. Offset the expiration times by 1 minute each so that we can verify that only the correct CRLs get updated at each expiration.
 			currentCRLs := map[string]*x509.RevocationList{}
 			crlPems := map[string]string{}
 			caBundle := []string{}
 			validTime := 3 * time.Minute
-			//expirations := []time.Time{}
```
Miciah, line 1233: Debug code?

test/e2e/client_tls_test.go
```diff
@@ -1222,15 +1191,33 @@ func TestCRLUpdate(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}
+	if err := kclient.Create(context.TODO(), &namespace); err != nil {
+		t.Fatalf("Failed to create namespace %q: %v", namespace.Name, err)
+	}
+	t.Cleanup(func() { assertDeleted(t, kclient, &namespace) })
+	for i := range testCases {
+		tc := testCases[i]
```
Miciah, line 1205: What's the point of this change?  Did you mean to use `tc := &testCases[i]` to avoid copying the whole test case?  Even that would be an unnecessary micro-optimization test code, in my opinion, but I don't understand the purpose of `tc := testCases[i]` at all.

test/e2e/client_tls_test.go
```diff
@@ -1379,6 +1379,225 @@ func TestCRLUpdate(t *testing.T) {
 	}
 }
 
+// TestRouterWaitsForCRLs verifies that router pods only report ready once all CRLs are downloaded.
+func TestRouterWaitsForCRLs(t *testing.T) {
+	namespaceName := names.SimpleNameGenerator.GenerateName("router-waits-for-crls-")
```
Miciah, line 1384: Missing `t.Parallel()`.

test/e2e/client_tls_test.go
```diff
@@ -1670,3 +1615,57 @@ func curlGetStatusCode(t *testing.T, clientPod *corev1.Pod, certName, endpoint,
...
+					ContainerPort: 8080,
+					Name:          "http-svc",
+				}},
+				VolumeMounts: []corev1.VolumeMount{{
+					Name:      "data",
+					MountPath: "/var/www/html",
+					ReadOnly:  true,
+				}},
+				SecurityContext: generateUnprivilegedSecurityContext(),
+			}},
```
Miciah, line 1859: This is subtle.  You factored this function out of `TestCRLUpdate`, but you also changed the image pullspec from `quay.io/centos7/httpd-24-centos7:centos7` to `quay.io/openshifttest/httpd-24:multi` and removed the readiness probe.  Were those changes intentional?  If they were, can you use separate commits, one adding `buildCRLHostPod` and one changing image pullspec and probe?  It also helps if the commit message explains the changes.

test/e2e/util_test.go
```diff
@@ -796,6 +796,12 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
 	}
 }
 
+func assertCreated(t *testing.T, cl client.Client, thing client.Object) {
+	if err := cl.Create(context.TODO(), thing); err != nil {
+		t.Fatalf("Failed to create %s: %v", thing.GetName(), err)
+	}
+}
```
Miciah, line 803: Seems like this was supposed to be in the other commit.

test/e2e/client_tls_test.go
```diff
@@ -1222,15 +1191,33 @@ func TestCRLUpdate(t *testing.T) {
...
+		tc := testCases[i]
 		t.Run(tc.Name, func(t *testing.T) {
-			caCerts := tc.CreateCerts()
+			t.Parallel()
+			crlHostName := types.NamespacedName{
+				Name:      names.SimpleNameGenerator.GenerateName("crl-host-"),
+				Namespace: namespaceName,
+			}
+			// When generating certificates, the CRL distribution points need to be specified by URL.
+			crlHostServiceName := names.SimpleNameGenerator.GenerateName("crl-host-service")
```
Miciah, line 1213: Did you mean to have a trailing "-"?

test/e2e/client_tls_test.go
```diff
@@ -1441,7 +1385,7 @@ func verifyCRLs(t *testing.T, pod *corev1.Pod, expectedCRLs map[string]*x509.Rev
 	if err != nil {
 		return false, err
 	}
-	if len(activeCRLs) == 0 {
+	if len(activeCRLs) == 0 && len(expectedCRLs) != 0 {
```
Miciah, line 1607: Was this change necessary for any of the test cases?

test/e2e/client_tls_test.go
rfredette: I took the brute force approach to make sure there were no name conflicts when the subtests are run in parallel, but I agree, it'd be better be able to identify which resources are a part of the same subtest.

test/e2e/client_tls_test.go
rfredette: `for _, tc := range testCases` isn't thread safe. I initially had it that way, but once `t.Parallel()` was called, `tc` in all threads converged on the last test case in the list.

test/e2e/client_tls_test.go
rfredette: Good catch; that wasn't intentional. I had to do some manual merging when I rebased recently, so either I missed that difference, or it got dropped by git. Either way, I'll restore those changes.

test/e2e/client_tls_test.go
rfredette: Fixed

test/e2e/client_tls_test.go
rfredette: Fixed

test/e2e/client_tls_test.go
```diff
@@ -1222,15 +1191,33 @@ func TestCRLUpdate(t *testing.T) {
...
+	namespaceName := names.SimpleNameGenerator.GenerateName("crl-update-")
+	namespace := corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespaceName,
+		},
+	}
+	if err := kclient.Create(context.TODO(), &namespace); err != nil {
+		t.Fatalf("Failed to create namespace %q: %v", namespace.Name, err)
+	}
+	t.Cleanup(func() { assertDeleted(t, kclient, &namespace) })
```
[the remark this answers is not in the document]
rfredette: No reason; I've changed it to use `createNamespace`.

test/e2e/client_tls_test.go
rfredette: Should be fixed. I've re-added the readiness probe and changed it to use `quay.io/centos7/httpd-24-centos7:centos7`.

test/e2e/client_tls_test.go
rfredette: Yes. Fixed!

test/e2e/client_tls_test.go
rfredette: Looks like it's not necessary. Reverted.

PR #1064, opened 2024-05-24
OCPBUGS-34418: Allow router pods to use the "restricted" SCC

Miciah:
Give router pods permission to use the "restricted" securitycontextconstraint.  This permission is required since OpenShift 4.11 in order to use the "restricted" SCC.

[The OCP 4.11 release notes mention](https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-auth-pod-security-admission) that explicit permission is now required to use the "restricted" SCC:

    The restricted SCC is no longer available to users of new clusters, unless the access is explicitly granted. In clusters originally installed in OpenShift Container Platform 4.10 or earlier, all authenticated users can use the restricted SCC when upgrading to OpenShift Container Platform 4.11 and later.

Router pods already have permission to use the "hostnetwork" SCC as they need this SCC when they use the host network.  Because we did not give router pods permission to use the "restricted" SCC, the pods always used the "hostnetwork" SCC, even when they did not use the host network.  After this commit, router pods use the "hostnetwork" SCC if they use the host network, and router pods use the "restricted" SCC otherwise.

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -239,6 +239,7 @@ rules:
   - use
   resourceNames:
   - hostnetwork
+  - restricted
 
 # Mirrored from assets/router/cluster-role.yaml
 - apiGroups:
--- a/pkg/manifests/assets/router/cluster-role.yaml
+++ b/pkg/manifests/assets/router/cluster-role.yaml
@@ -51,6 +51,7 @@ rules:
   - use
   resourceNames:
   - hostnetwork
+  - restricted
 
 - apiGroups:
   - discovery.k8s.io


ShudiLi:
/label qe-approved

tested it with 4.16.0-0.ci.test-2024-05-27-060132-ci-ln-nqikg9t-latest
1.
 % oc get clusterversion
version   4.16.0-0.ci.test-2024-05-27-060132-ci-ln-nqikg9t-latest   True        False         42m     Cluster version is 4.16.0-0.ci.test-2024-05-27-060132-ci-ln-nqikg9t-latest

2.
% oc get ClusterRole openshift-ingress-operator -oyaml | grep -A2 -B2 resourceNames
- apiGroups:
  - security.openshift.io
  resourceNames:
  - hostnetwork
  - restricted

3.
% oc get ClusterRole openshift-ingress-router -oyaml | grep -A2 -B2 resourceNames
- apiGroups:
  - security.openshift.io
  resourceNames:
  - hostnetwork
  - restricted

candita:
@Miciah one question about this.  The docs at https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-auth-pod-security-admission mention "restricted-v2" but this PR uses "restricted".  Is there a difference?

Miciah:
> @Miciah one question about this. The docs at https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-auth-pod-security-admission mention "restricted-v2" but this PR uses "restricted". Is there a difference?

Yes, the documentation that you linked to mentions several differences between the old and new SCCs.  Crucially, the "restricted-v2" SCC prohibits enabling `allowPrivilegeEscalation`, which router pods require.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/a4c8bc78eb2e7926dc3738eb3df70006b5467d70..85df7e5a5c3ba9325bf93508829383dee544f9ea fixes a typo in the commit message.

manifests/00-cluster-role.yaml
```diff
@@ -239,6 +239,7 @@ rules:
   - use
   resourceNames:
   - hostnetwork
+  - restricted
```
alebedev87, line 242: I'm wondering why the router pod is not applicable for to the new `restricted-v2` in case the hostnetwork is not used. And as continuation, if the scc has to be specified explicitly, why not specify the new one instead?

alebedev87: Sorry, I see now that Candace had [a similar question](https://github.com/openshift/cluster-ingress-operator/pull/1064#issuecomment-2138149989).

> Crucially, the "restricted-v2" SCC prohibits enabling allowPrivilegeEscalation, which router pods require.

Right.

PR #1047, opened 2024-04-24
OCPBUGS-32887: Allow operator to update Route spec.subdomain

Miciah:
Before this change, cluster-ingress-operator did not have permission to update `spec.host` or `spec.subdomain` on an existing route as the operator's serviceaccount did not have the necessary "routes/custom-host" permission.  #965 added logic to the operator to clear `spec.host` and instead set `spec.subdomain`, but without the required permission, the update would fail with the following error message:

    ERROR   operator.init   controller/controller.go:265    Reconciler error {"controller": "canary_controller", "object": {"name":"default","namespace":"openshift-ingress-operator"}, "namespace": "openshift-ingress-operator", "name": "default", "reconcileID": "463061e3-93a1-4067-802e-03e3f1f8cdd0", "error": "failed to ensure canary route: failed to update canary route openshift-ingress-canary/canary: Route.route.openshift.io \"canary\" is invalid: spec.subdomain: Invalid value: \"canary-openshift-ingress-canary\": field is immutable"}

This PR adds the needed permission to the clusterrole for the operator's serviceaccount so that the update can succeed.

PR #1095, opened 2024-06-27
OCPBUGS-36465: Delete and recreate canary route to clear spec.host

Miciah:
Fix the update logic for the canary route to handle clearing `spec.host`. Attempts to clear `spec.host` using a simple update may be ignored (see https://github.com/openshift/origin/commit/54c072c122b48c5d7074fe8904d493d0534733e2).  Therefore it is necessary to delete and recreate the route.  

Before this change, the operator would set `spec.subdomain`, but it did not actually clear `spec.host`, and so setting `spec.subdomain` had no effect.

After this change, the operator should clear `spec.host`, and `spec.subdomain` should be in effect.

Follow-up to #1047.

### DIFF
--- a/pkg/operator/controller/canary/route.go
+++ b/pkg/operator/controller/canary/route.go
@@ -14,6 +14,8 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ensureCanaryRoute ensures the canary route exists
@@ -75,6 +77,21 @@ func (r *reconciler) updateCanaryRoute(current, desired *routev1.Route) (bool, e
 		return false, nil
 	}
 
+	if len(updated.Spec.Host) == 0 && len(current.Spec.Host) != 0 {
+		// Attempts to clear spec.host may be ignored.  Thus, to clear
+		// spec.host, it is necessary to delete and recreate the route.
+		log.Info("deleting and recreating the canary route to clear spec.host", "namespace", current.Namespace, "name", current.Name, "old spec.host", current.Spec.Host)
+		foreground := metav1.DeletePropagationForeground
+		deleteOptions := crclient.DeleteOptions{PropagationPolicy: &foreground}
+		if _, err := r.deleteCanaryRoute(current, &deleteOptions); err != nil {
+			return false, err
+		}
+		if err := r.createCanaryRoute(desired); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
 	// Diff before updating because the client may mutate the object.
 	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
 	if err := r.client.Update(context.TODO(), updated); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/canary/route.go
```diff
@@ -75,6 +77,21 @@ func (r *reconciler) updateCanaryRoute(current, desired *routev1.Route) (bool, e
 		return false, nil
 	}
 
+	if len(updated.Spec.Host) == 0 && len(current.Spec.Host) != 0 {
+		// Attempts to clear spec.host may be ignored.  Thus, to clear
+		// spec.host, it is necessary to delete and recreate the route.
+		log.Info("deleting and recreating the canary route to clear spec.host", "namespace", current.Namespace, "name", current.Name, "old spec.host", current.Spec.Host)
+		foreground := metav1.DeletePropagationForeground
```
candita, line 84: `DeletePropagationForeground` [keeps the key-value in the store until all the dependencies are also deleted (making sure there are no orphans)](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L489-L494).  It's not perfectly clear to me that this is needed, but I don't expect it to cause any harm.

test/e2e/canary_test.go
```diff
@@ -127,6 +127,64 @@ func TestCanaryRoute(t *testing.T) {
...
+	if v := canaryRoute.Spec.Host; len(v) != 0 {
+		t.Fatalf("Expected canary route to have empty spec.host, found %q.", v)
+	}
+
+	const bogusHost = "foo.bar"
+	canaryRoute.Spec.Host = bogusHost
+	t.Log("Setting a bogus spec.host...")
+	if err := kclient.Update(context.TODO(), &canaryRoute); err != nil {
+		t.Fatal(err)
+	}
```
candita, line 164: I think we need to specifically call the function `updateCanaryRoute` to test this.

pkg/operator/controller/canary/route.go
```diff
@@ -75,6 +77,21 @@ func (r *reconciler) updateCanaryRoute(current, desired *routev1.Route) (bool, e
 		return false, nil
 	}
 
+	if len(updated.Spec.Host) == 0 && len(current.Spec.Host) != 0 {
```
candita, line 80: It looks like we also need to fix the function `canaryRouteChanged` to return `true` if the `Spec.Host` changed, otherwise we never get here.

/hold

candita: Update: I was mistaken on this.
/unhold

candita: https://github.com/openshift/cluster-ingress-operator/pull/965/files hasn't been fully backported yet and I was looking in the wrong branch.

pkg/operator/controller/canary/route.go
Miciah: The godoc for the `DeletePropagationForeground` value is getting into the implementation details, but [the godoc for the `PropagationPolicy` field](https://github.com/kubernetes/apimachinery/blob/6b362fab6df25da6dc077a5414bb1e54235a8bc7/pkg/apis/meta/v1/types.go#L530-L538) is more to the point: 
```go
	// Whether and how garbage collection will be performed.
	// Either this field or OrphanDependents may be set, but not both.
	// The default policy is decided by the existing finalizer set in the
	// metadata.finalizers and the resource-specific default policy.
	// Acceptable values are: 'Orphan' - orphan the dependents; 'Background' -
	// allow the garbage collector to delete the dependents in the background;
	// 'Foreground' - a cascading policy that deletes all dependents in the
	// foreground.
```
The important thing is that deletion happens in the foreground so that the operator can immediately recreate the route.

test/e2e/canary_test.go
Miciah: This is an end-to-end test, not a unit test.  All it needs to do is the following:

1. Get the canary route.
2. Update its `spec.host` field.
3. Verify that the operator clears the field.

The fact that clearing the field requires deleting and recreating the route is irrelevant for this test.  If the operator succeeds in clearing `spec.host`, it's good.  Is there some case that this test fails to cover?

PR #1099, opened 2024-07-02
[release-4.16] OCPBUGS-32887: Delete and recreate canary route to clear spec.host

Miciah:
Fix the update logic for the canary route to handle clearing `spec.host`. Attempts to clear `spec.host` using a simple update may be ignored (see https://github.com/openshift/origin/commit/54c072c122b48c5d7074fe8904d493d0534733e2).  Therefore it is necessary to delete and recreate the route.  

Before this change, the operator would set `spec.subdomain`, but it did not actually clear `spec.host`, and so setting `spec.subdomain` had no effect.

After this change, the operator should clear `spec.host`, and `spec.subdomain` should be in effect.

Follow-up to #1047.

---

This is manual cherry-pick of #1095.  #978 added the `TestCanaryWithMTLS` test to `test/e2e/all_test.go` in the release-4.17 branch, and this test is not in the release-4.16 branch, so it caused a conflict for this backport that required manual resolution.

PR #1100, opened 2024-07-03
[release-4.15] OCPBUGS-36466: Allow operator to update Route spec.subdomain

Miciah:
#### Allow operator to update Route `spec.subdomain`

Before this change, cluster-ingress-operator did not have permission to update `spec.host` or `spec.subdomain` on an existing route as the operator's serviceaccount did not have the necessary "routes/custom-host" permission.  #965 added logic to the operator to clear `spec.host` and instead set `spec.subdomain`, but without the required permission, the update would fail with the following error message:

    ERROR   operator.init   controller/controller.go:265    Reconciler error {"controller": "canary_controller", "object": {"name":"default","namespace":"openshift-ingress-operator"}, "namespace": "openshift-ingress-operator", "name": "default", "reconcileID": "463061e3-93a1-4067-802e-03e3f1f8cdd0", "error": "failed to ensure canary route: failed to update canary route openshift-ingress-canary/canary: Route.route.openshift.io \"canary\" is invalid: spec.subdomain: Invalid value: \"canary-openshift-ingress-canary\": field is immutable"}

This PR adds the needed permission to the clusterrole for the operator's serviceaccount so that the update can succeed.

#### Delete and recreate canary route to clear `spec.host`

Fix the update logic for the canary route to handle clearing `spec.host`. Attempts to clear `spec.host` using a simple update may be ignored (see https://github.com/openshift/origin/commit/54c072c122b48c5d7074fe8904d493d0534733e2).  Therefore it is necessary to delete and recreate the route.  

Before this change, the operator would set `spec.subdomain`, but it did not actually clear `spec.host`, and so setting `spec.subdomain` had no effect.

After this change, the operator should clear `spec.host`, and `spec.subdomain` should be in effect.

---

This is manual cherry-pick of #1047 and #1099.

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -217,6 +217,7 @@ rules:
   - route.openshift.io
   resources:
   - routes
+  - routes/custom-host
   verbs:
   - '*'
 
--- a/pkg/operator/controller/canary/route.go
+++ b/pkg/operator/controller/canary/route.go
@@ -14,6 +14,8 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ensureCanaryRoute ensures the canary route exists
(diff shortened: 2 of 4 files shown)


Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/d19536ff9fb53fc72fb6f362e81567eabf45fb9d..357c1a7942e5a194687febab169cce3624f35ddc adds a missing `fmt` import.

melvinjoseph86:
Verified the fix using clusterbot

    create a build using this PR and upgrade a 4.13 cluster using that build.
    Didn't see any error messages in the ingress operator logs.

hence marking the bug are verified
/label qe-approved

candita:
Dependency on #1099

candita:
Backport is required because the original bug [OCPBUGS-36465](https://issues.redhat.com//browse/OCPBUGS-36465) was found in version 4.14 and disrupts upgrades from 4.13 to 4.14. Any risk introduced by this fix is balanced against the improvement in the success of 4.14 upgrades.

/label backport-risk-assessed

PR #1106, opened 2024-07-12
NE-1273: Add a watch to the ingress operator so it will recreate the gwapi crds

anirudhAgniRedhat:
Added watch for Gateway API CRDs to recreate CRDs if they get deleted

### DIFF
--- a/pkg/operator/controller/gatewayapi/controller.go
+++ b/pkg/operator/controller/gatewayapi/controller.go
@@ -15,6 +15,7 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -50,6 +51,24 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	})
 	if err := c.Watch(source.Kind(operatorCache, &configv1.FeatureGate{}), &handler.EnqueueRequestForObject{}, clusterNamePredicate); err != nil {
 		return nil, err
+	}
+
+	toFeatureGate := func(ctx context.Context, _ client.Object) []reconcile.Request {
+		return []reconcile.Request{{
+			NamespacedName: operatorcontroller.FeatureGateClusterConfigName(),
+		}}
+	}
+
+	// watch for CRDs
+	for i := range managedCRDs {
+		if err = c.Watch(source.Kind(operatorCache, managedCRDs[i]), handler.EnqueueRequestsFromMapFunc(toFeatureGate), predicate.Funcs{
+			CreateFunc:  func(e event.CreateEvent) bool { return false },
+			DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+			UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+			GenericFunc: func(e event.GenericEvent) bool { return false },
+		}); err != nil {
+			return nil, err
+		}
 	}
 	return c, nil
 }
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -80,9 +80,16 @@ func TestGatewayAPI(t *testing.T) {
 // testGatewayAPIResources tests that Gateway API Custom Resource Definitions are available.
 // It specifically verifies that when the GatewayAPI feature gate is enabled, that the Gateway API
 // CRDs are created.
+// It also deletes and ensure the CRDs are recreated.
 func testGatewayAPIResources(t *testing.T) {
 	t.Helper()
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
+	ensureCRDs(t)
+
+	// Deleting CRDs to ensure they gets recreated again
+	deleteCRDs(t)
+
+	// Make sure all the *.gateway.networking.k8s.io CRDs are available since they should be recreated after manual deletion.
 	ensureCRDs(t)
 }
 
@@ -143,6 +150,17 @@ func ensureCRDs(t *testing.T) {
 			t.Fatalf("failed to find crd %s: %v", crdName, err)
 		}
 		t.Logf("found crd %s at version %s", crdName, crdVersion)
+	}
+}
+
+// deleteCRDs deletes Gateway API custom resource definitions.
+func deleteCRDs(t *testing.T) {
+	t.Helper()
+	for _, crdName := range crdNames {
+		err := deleteExistingCRD(t, crdName)
+		if err != nil {
+			t.Errorf("failed to delete crd %s: %v", crdName, err)
+		}
 	}
 }
 
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -90,7 +90,7 @@ func testGatewayAPIResources(t *testing.T) {
 // the following installation operations complete automatically and successfully:
 // - the required Subscription and CatalogSource are created.
 // - the OSSM Istio operator is installed successfully and has status Running and Ready. e.g. istio-operator-9f5c88857-2xfrr  -n openshift-operators
-// - the Istiod proxy is installed successfully and has status Running and Ready.  e.g istiod-openshift-gateway-867bb8d5c7-4z6mp -n openshift-ingress
+// - Istiod is installed successfully and has status Running and Ready.  e.g istiod-openshift-gateway-867bb8d5c7-4z6mp -n openshift-ingress
 // - the SMCP is created successfully (OSSM 2.x).
 func testGatewayAPIIstioInstallation(t *testing.T) {
 	t.Helper()
@@ -99,13 +99,13 @@ func testGatewayAPIIstioInstallation(t *testing.T) {
 		t.Fatalf("failed to find expected Subscription %s: %v", expectedSubscriptionName, err)
 	}
 	if err := assertCatalogSource(t, expectedCatalogSourceNamespace, expectedCatalogSourceName); err != nil {
-		t.Fatalf("failed to find expected CatalogSource %s", expectedCatalogSourceName)
+		t.Fatalf("failed to find expected CatalogSource %s: %v", expectedCatalogSourceName, err)
 	}
 	if err := assertOSSMOperator(t); err != nil {
-		t.Fatal("failed to find expected Istio operator")
+		t.Fatalf("failed to find expected Istio operator: %v", err)
 	}
 	if err := assertIstiodControlPlane(t); err != nil {
-		t.Fatal("failed to find expected Istiod control plane")
+		t.Fatalf("failed to find expected Istiod control plane: %v", err)
 	}
 	// TODO - In OSSM 3.x the configuration object to check will be different.
 	if err := assertSMCP(t); err != nil {
@@ -163,7 +163,7 @@ func ensureGatewayObjectCreation(ns *corev1.Namespace) error {
 	if err != nil {
 		return fmt.Errorf("feature gate was enabled, but gateway object could not be created: %v", err)
 	}
-	// The gateway is cleaned up in testGatewayAPIObjects.
+	// The gateway is cleaned up in TestGatewayAPI.
 
 	hostname := names.SimpleNameGenerator.GenerateName("test-hostname-")
 	defaultRoutename = hostname + "." + domain
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -39,7 +39,7 @@ const (
 	openshiftOperatorsNamespace = "openshift-operators"
 	// openshiftIstioOperatorDeploymentName holds the expected istio-operator deployment name.
 	openshiftIstioOperatorDeploymentName = "istio-operator"
-	// openshiftIstiodDeploymentName holds the expected istiod proxy deployment name
+	// openshiftIstiodDeploymentName holds the expected istiod deployment name
 	openshiftIstiodDeploymentName = "istiod-openshift-gateway"
 	// openshiftSMCPName holds the expected OSSM ServiceMeshControlPlane name
 	openshiftSMCPName = "openshift-gateway"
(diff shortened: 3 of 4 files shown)


candita:
@anirudhAgniRedhat please add an e2e test to this so we can see how it works.

test/e2e/gateway_api_test.go
```diff
@@ -80,17 +80,25 @@ func TestGatewayAPI(t *testing.T) {
...
 	t.Helper()
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
 	ensureCRDs(t)
+
+	// Deleting CRDs to ensure they gets recreated again
+	deleteCRDs(t)
+
+	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled and if deleted
+	// manually.
+	ensureCRDs(t)
```
gcs278, line 94: Thanks for taking on the PR review!

Since they were two different efforts/tasks, would you mind breaking apart the E2E fixes from the Candace's PR code review and your new additional E2E changes for NE-1273. It will help with clarity in documentation and traceability, and future reverts (if ever needed).

And in your commit message of the E2E fixes commit, mind putting a link in for https://github.com/openshift/cluster-ingress-operator/pull/1023 to reference the PR where the code review changes are being addressed from?

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,51 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
...
+		return err
+	}
+	return nil
+}
+
 // createHttpRoute checks if the HTTPRoute can be created.
 // If it can't an error is returned.
-func createHttpRoute(namespace, routename, parentnamespace, hostname, backendrefname string, gateway *gwapi.Gateway) (*gwapi.HTTPRoute, error) {
+func createHttpRoute(namespace, routeName, parentNamespace, hostname, backendRefname string, gateway *gwapi.Gateway) (*gwapi.HTTPRoute, error) {
 	// Just in case gateway creation failed, supply a fake gateway name.
```
gcs278, line 141: Should remove comment too.

```suggestion
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -125,10 +157,10 @@ func createHttpRoute(namespace, routename, parentnamespace, hostname, backendref
 		return nil, fmt.Errorf("failed to create service %s/%s: %v", echoService.Namespace, echoService.Name, err)
 	}
 
-	httpRoute := buildHTTPRoute(routename, namespace, name, parentnamespace, hostname, backendrefname)
+	httpRoute := buildHTTPRoute(routeName, namespace, name, parentNamespace, hostname, backendRefname)
```
gcs278, line 160: I think the `name` variable is making the code more ambiguous and harder to read, I'd just get rid of it.

```suggestion
	httpRoute := buildHTTPRoute(routeName, namespace, gateway.Name, parentNamespace, hostname, backendRefname)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,51 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
...
-func createHttpRoute(namespace, routename, parentnamespace, hostname, backendrefname string, gateway *gwapi.Gateway) (*gwapi.HTTPRoute, error) {
+func createHttpRoute(namespace, routeName, parentNamespace, hostname, backendRefname string, gateway *gwapi.Gateway) (*gwapi.HTTPRoute, error) {
 	// Just in case gateway creation failed, supply a fake gateway name.
-	name := "NONE"
-	if gateway != nil {
-		name = gateway.Name
+	if gateway == nil {
+		return nil, errors.New("unable to create httpRoute, no gateway available")
 	}
+	name := gateway.Name
```
gcs278, line 145: Related to my comment below.
```suggestion
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,51 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
 	return crdVersion, err
 }
 
+// deleteExistingCRDs deletes if the CRD of the given name exists and returns an error if not.
+func deleteExistingCRDs(t *testing.T, crdname string) error {
```
gcs278, line 107: Function is deleting a single CRD, name should reflect:

```suggestion
func deleteExistingCRD(t *testing.T, crdname string) error {
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,51 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
...
+			t.Logf("failed to get crd %s: %v", name, err)
+			return false, nil
+		}
+		crdConditions := crd.Status.Conditions
+		for _, c := range crdConditions {
+			if c.Type == apiextensionsv1.Established && c.Status == apiextensionsv1.ConditionTrue {
+				return true, nil
+			}
+		}
+		t.Logf("failed to find crd %s to be Established", name)
```
gcs278, line 123: Is there a reason we need to check if the CRD is `Established` before deleting it?

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,51 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
...
+	})
+	if err != nil {
+		t.Errorf("failed to get crd %s: %v", name, err)
+		return err
+	}
+	err = kclient.Delete(context.Background(), crd)
+	if err != nil {
+		t.Errorf("failed to delete crd %s: %v", name, err)
+		return err
+	}
```
gcs278, line 134: I believe this will mark the CRD for deletion, not-block, and continue on. Deletion can take a couple of seconds. A potential issue is that the CRD is still available, but being deleted, and the `EnsureCRDs` function finds them without being delete, and passes because they still look good even though they are in the process of being deleted.

To be safe, I'd consider blocking here. One example of blocking or waiting for deletion is [deleteIngressController](https://github.com/openshift/cluster-ingress-operator/blob/dc4e7f9a511ae028db8149609a1f4b80f7b8a500/test/e2e/operator_test.go#L4263-L4272).

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -2,6 +2,8 @@ package gatewayapi
 
 import (
 	"context"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
gcs278, line 5: *nit*
We should move this into it's own "group" below `"k8s.io/apimachinery/pkg/types"` I suppose.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -2,6 +2,8 @@ package gatewayapi
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
```
gcs278, line 6: Should be moved into the `sigs.k8s.io` group below.

gcs278 (commented):
Thanks for taking on the PR review comments. Just a quick review while I'm here.

test/e2e/util_gatewayapi_test.go
anirudhAgniRedhat: I added it for error handling if you think we can remove it.

gcs278: Hmm, yes it's extra error handling, but it seems to me it's already been confirmed as established, and from the perspective of this helper function's role to delete the CRD, I don't think it matters if it's established.

I'd vote to remove the the logic to keep things simple, and aligned with the roles of the functions (`deleteExistingCRD` just deletes, and `assertCrdExists` does the check for established). I'll leave it up to you though, its not a big concern to me.

test/e2e/util_gatewayapi_test.go
anirudhAgniRedhat: @gcs278 I have added the poll because the IngressController has multiple operands, which means it takes some time to delete. However, CRDs are deleted quite quickly. When I receive my first poll CRD, it gives me a new CRD with a different UUID, and it ends up stuck in a loop searching for a new CRD.
```
// if new CRD got recreated while the poll ensures the CRD is deleted.
if newCRD != nil && newCRD.UID != crd.UID {
	return true, nil
}
```

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -51,6 +54,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind(operatorCache, &configv1.FeatureGate{}), &handler.EnqueueRequestForObject{}, clusterNamePredicate); err != nil {
 		return nil, err
 	}
+
+	// watch for CRDs
+	if err = c.Watch(source.Kind(operatorCache, &apiextensionsv1.CustomResourceDefinition{}), &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
```
gcs278, line 61: So this reconciles this control loop whenever there is any delete in an object informed by the `operatorCache`. I suppose that's not too bad, but generally, I think we try to be as precise as possible, to avoid the growing pains that come with unnecessary Operator reconcile cycles. Never know what crazy deletes/creates/updates a cluster at scale might be doing.

Have you thought about using the `managedCRDs` object and filtering the `DeleteEvent` by only when the CRDs we care about are deleted?

test/e2e/gateway_api_test.go
```diff
@@ -80,17 +80,25 @@ func TestGatewayAPI(t *testing.T) {
...
 func testGatewayAPIResources(t *testing.T) {
 	t.Helper()
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
 	ensureCRDs(t)
+
+	// Deleting CRDs to ensure they gets recreated again
+	deleteCRDs(t)
+
+	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled and if deleted
+	// manually.
```
gcs278, line 93: *nit* a bit of an confusing sentence.

```suggestion
	// Make sure all the *.gateway.networking.k8s.io CRDs are available since they should be recreated after manual deletion.
```

test/e2e/util_gatewayapi_test.go
gcs278: > When I receive my first poll CRD, it gives me a new CRD with a different UUID

Sorry, I might be confused, isn't this a good thing? Not sure why it'd get stuck in this loop.

Looks like the test passed, are you saying it's not working correctly still or did you figure it out?

gcs278 (commented):
Thanks for the responses, looking good.

test/e2e/util_gatewayapi_test.go
anirudhAgniRedhat: The problem here I am trying to address is that if I remove the code section for 
```
// if new CRD got recreated while the poll ensures the CRD is deleted.
if newCRD != nil && newCRD.UID != crd.UID {
	return true, nil
}
```
The problem will be the reconcile loop is creating the CRD again pretty quickly and when I am trying get the CRD in the poll section it is already been recreated, in that case the `if kerrors.IsNotFound(err)` condition will never satisfied and tests will end-up failing due to context deadline exceeded.
This is the reason Why I have validated it with old CRD's UID.

pkg/operator/controller/gatewayapi/controller.go
anirudhAgniRedhat: @gcs278 The `managedCRDs` function already handles the case where if the CRDs are already present it will not do anything.
However I  added it for only delete events because that is what only required in this case, I don't want to add any extra events if they are not required at this moment.

test/e2e/gateway_api_test.go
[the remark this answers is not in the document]
anirudhAgniRedhat: Added!

test/e2e/util_gatewayapi_test.go
```diff
@@ -103,19 +103,63 @@ func assertCrdExists(t *testing.T, crdname string) (string, error) {
...
+		t.Errorf("failed to get crd %s: %v", name, err)
+		return err
+	}
+	// deleting CRD.
+	err = kclient.Delete(context.Background(), crd)
+	if err != nil {
+		t.Errorf("failed to delete crd %s: %v", name, err)
+		return err
+	}
+	err = wait.PollImmediate(100*time.Millisecond, 30*time.Second, func() (bool, error) {
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Added suggested changes!!

pkg/operator/controller/gatewayapi/controller.go
gcs278: > @gcs278 The managedCRDs function already handles the case where if the CRDs are already present it will not do anything.

Correct, `ensureGatewayAPICRDs` won't do anything incorrectly if it gets extra reconciles, everything will work fine, but it's doing more work than it needs, and logging extra reconcile loops for other-than-gateway-api CRDs. e.g. when a `IngressController` is deleted in openshift-ingress-operator namespace, this logic will run reconcile, but we know deleting an `IngressController` won't cause any impact. 

I'm suggesting to be more specific from watching all CRDs delete events to only the managedCRDs we care about, maybe this suggestion will help to make sense (I haven't tested it):
```
	for i := range managedCRDs {
		if err = c.Watch(source.Kind(operatorCache, managedCRDs[i]), &handler.EnqueueRequestForObject{}, predicate.Funcs{
			CreateFunc:  func(e event.CreateEvent) bool { return false },
			DeleteFunc:  func(e event.DeleteEvent) bool { return true },
			UpdateFunc:  func(e event.UpdateEvent) bool { return false },
			GenericFunc: func(e event.GenericEvent) bool { return false },
		}); err != nil {
			return nil, err
		}
	}
```

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -51,6 +53,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind(operatorCache, &configv1.FeatureGate{}), &handler.EnqueueRequestForObject{}, clusterNamePredicate); err != nil {
 		return nil, err
 	}
+
+	// watch for CRDs
+	if err = c.Watch(source.Kind(operatorCache, &apiextensionsv1.CustomResourceDefinition{}), &handler.EnqueueRequestForObject{}, predicate.Funcs{
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Updated as per your suggestion!

anirudhAgniRedhat:
@rfredette @gcs278 Added Suggested Changes!
Do you guys have any more suggestions in this, else I can get a tag on this!

test/e2e/util_gatewayapi_test.go
gcs278: Oh yes, that sounds very reasonable, better to be precise to avoid race conditions.

gcs278:
Looks good to me! Thanks for the code review responses.
Since this is a minimal change and it's still in tech preview, I'll label myself.
/approve
/lgtm

PR #1110, opened 2024-07-26
OCPBUGS-37627: Fix getRouteHost error handling

gcs278:
Some E2E tests were calling `getRouteHost` and expected it not to call `t.Fatal` when the hostname couldn't be found. However, `getRouteHost` did call `t.Fatal`, causing E2E test flakes. This fix moves the `t.Fatal` call outside of `getRouteHost`, allowing the function callers to handle the error as they wish.

### DIFF
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -832,15 +832,12 @@ func assertDeletedWaitForCleanup(t *testing.T, cl client.Client, thing client.Ob
 
 // getRouteHost returns the host name of the given route for the named
 // IngressController.  If the IngressController has not added its host name to
-// the route's status, this function causes a fatal error.  For simplicity, this
+// the route's status, this function returns an empty string.  For simplicity, this
 // function does not check the "Admitted" status condition, so it will return
 // the host name if it is set even if the IngressController has rejected the
 // route.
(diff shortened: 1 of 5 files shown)


test/e2e/util_test.go
```diff
@@ -850,7 +850,6 @@ func getRouteHost(t *testing.T, route *routev1.Route, router string) string {
 		}
 	}
 
-	t.Fatalf("failed to find host name for default router in route: %#v", route)
```
frobware, line 853: Any reason not to drop the other call to fatal in this function? Would that perturb other test behaviour in a bad way? If the other call can be dropped we can drop the call to t.Helper() too.

gcs278:
Not failures related to this E2E fix:
/retest

test/e2e/util_test.go
gcs278: I don't think it would perturb the test behavior, removed.

gcs278:
`received unexpected HTTP status: 500 Internal Server Error` infra failure of some sort.
/retest

gcs278:
`Throttling: Rate exceeded` failures in AWS ([I asked about this in slack](https://redhat-internal.slack.com/archives/CBN38N3MW/p1722287253954199)) and other cluster provisioning failures.
/retest

PR #1112, opened 2024-07-29
NE-1907: Manage OSSM operator subscription manually to ensure a compatible version is installed

rfredette:
Istio and the Gateway API CRDs need to be in sync to work. The CRDs are baked into a particular openshift release, so this change updates the ingress operator to install a compatible version with the CRDs it already installs.

### DIFF
--- a/cmd/ingress-operator/start.go
+++ b/cmd/ingress-operator/start.go
@@ -32,7 +32,9 @@ import (
 const (
 	// defaultTrustedCABundle is the fully qualified path of the trusted CA bundle
 	// that is mounted from configmap openshift-ingress-operator/trusted-ca.
-	defaultTrustedCABundle = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	defaultTrustedCABundle           = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	defaultGatewayAPIOperatorChannel = "stable"
+	defaultGatewayAPIOperatorVersion = "servicemeshoperator.v2.6.2"
 )
 
 type StartOptions struct {
@@ -51,6 +53,10 @@ type StartOptions struct {
 	CanaryImage string
 	// ReleaseVersion is the cluster version which the operator will converge to.
 	ReleaseVersion string
+	// GatewayAPIOperatorChannel is the release channel of the Gateway API implementation to install.
+	GatewayAPIOperatorChannel string
+	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
+	GatewayAPIOperatorVersion string
 }
 
 func NewStartCommand() *cobra.Command {
@@ -74,6 +80,8 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&options.ReleaseVersion, "release-version", "", statuscontroller.UnknownVersionValue, "the release version the operator should converge to (required)")
 	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", "127.0.0.1:60000", "metrics endpoint listen address (required)")
 	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "if provided, shut down the operator when this file changes")
+	cmd.Flags().StringVarP(&options.GatewayAPIOperatorChannel, "gateway-api-operator-channel", "", defaultGatewayAPIOperatorChannel, "release channel of the Gateway API implementation to install")
+	cmd.Flags().StringVarP(&options.GatewayAPIOperatorVersion, "gateway-api-operator-version", "", defaultGatewayAPIOperatorVersion, "name and release of the Gateway API implementation to install")
 
 	if err := cmd.MarkFlagRequired("namespace"); err != nil {
 		panic(err)
@@ -118,10 +126,12 @@ func start(opts *StartOptions) error {
 	defer cancel()
 
 	operatorConfig := operatorconfig.Config{
-		OperatorReleaseVersion: opts.ReleaseVersion,
-		Namespace:              opts.OperatorNamespace,
-		IngressControllerImage: opts.IngressControllerImage,
-		CanaryImage:            opts.CanaryImage,
+		OperatorReleaseVersion:    opts.ReleaseVersion,
+		Namespace:                 opts.OperatorNamespace,
+		IngressControllerImage:    opts.IngressControllerImage,
+		CanaryImage:               opts.CanaryImage,
+		GatewayAPIOperatorChannel: opts.GatewayAPIOperatorChannel,
+		GatewayAPIOperatorVersion: opts.GatewayAPIOperatorVersion,
 	}
 
 	// Start operator metrics.
--- a/manifests/02-deployment-ibm-cloud-managed.yaml
+++ b/manifests/02-deployment-ibm-cloud-managed.yaml
@@ -35,6 +35,10 @@ spec:
         - $(CANARY_IMAGE)
         - --release-version
         - $(RELEASE_VERSION)
+        - --gateway-api-operator-channel
+        - $(GATEWAY_API_OPERATOR_CHANNEL)
+        - --gateway-api-operator-version
+        - $(GATEWAY_API_OPERATOR_VERSION)
         env:
         - name: RELEASE_VERSION
           value: 0.0.1-snapshot
@@ -46,6 +50,10 @@ spec:
           value: openshift/origin-haproxy-router:v4.0
         - name: CANARY_IMAGE
           value: openshift/origin-cluster-ingress-operator:latest
+        - name: GATEWAY_API_OPERATOR_CHANNEL
+          value: stable
+        - name: GATEWAY_API_OPERATOR_VERSION
+          value: servicemeshoperator.v2.6.2
         image: openshift/origin-cluster-ingress-operator:latest
         imagePullPolicy: IfNotPresent
         name: ingress-operator
--- a/manifests/02-deployment.yaml
+++ b/manifests/02-deployment.yaml
@@ -64,6 +64,10 @@ spec:
           - "$(CANARY_IMAGE)"
           - --release-version
           - "$(RELEASE_VERSION)"
+          - --gateway-api-operator-channel
+          - "$(GATEWAY_API_OPERATOR_CHANNEL)"
+          - --gateway-api-operator-version
+          - "$(GATEWAY_API_OPERATOR_VERSION)"
           env:
             - name: RELEASE_VERSION
               value: "0.0.1-snapshot"
@@ -75,6 +79,10 @@ spec:
               value: openshift/origin-haproxy-router:v4.0
             - name: CANARY_IMAGE
               value: openshift/origin-cluster-ingress-operator:latest
+            - name: GATEWAY_API_OPERATOR_CHANNEL
+              value: stable
+            - name: GATEWAY_API_OPERATOR_VERSION
+              value: servicemeshoperator.v2.6.2
           resources:
             requests:
               cpu: 10m
--- a/pkg/operator/config/config.go
+++ b/pkg/operator/config/config.go
@@ -15,5 +15,11 @@ type Config struct {
 	// CanaryImage is the ingress operator image, which runs a canary command.
 	CanaryImage string
 
+	// GatewayAPIOperatorChannel is the release channel of the Gateway API implementation to install.
+	GatewayAPIOperatorChannel string
+
+	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
+	GatewayAPIOperatorVersion string
+
 	Stop chan struct{}
 }
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -77,6 +77,28 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 		enqueueRequestForDefaultGatewayClassController(config.OperandNamespace), isServiceMeshSubscription)); err != nil {
 		return nil, err
 	}
+
+	isOurInstallPlan := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		installPlan := o.(*operatorsv1alpha1.InstallPlan)
+		if len(installPlan.Spec.ClusterServiceVersionNames) > 0 {
+			for _, csv := range installPlan.Spec.ClusterServiceVersionNames {
+				if csv == config.GatewayAPIOperatorVersion {
+					return true
+				}
+			}
+		}
+		return false
+	})
+	// Check the approved status of an InstallPlan. The ingress operator only needs to potentially move install plans
+	// from Approved=false to Approved=true, so we can filter out all approved plans at the Watch() level.
+	isInstallPlanApproved := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		installPlan := o.(*operatorsv1alpha1.InstallPlan)
+		return installPlan.Spec.Approved
+	})
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.InstallPlan{}, enqueueRequestForDefaultGatewayClassController(config.OperatorNamespace), isOurInstallPlan, predicate.Not(isInstallPlanApproved))); err != nil {
+		return nil, err
+	}
+
 	gatewayClassController = c
 	return c, nil
 }
@@ -89,6 +111,10 @@ type Config struct {
 	OperatorNamespace string
 	// OperandNamespace is the namespace in which Istio should be deployed.
 	OperandNamespace string
+	// GatewayAPIOperatorChannel is the release channel of the Gateway API implementation to install.
+	GatewayAPIOperatorChannel string
+	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
+	GatewayAPIOperatorVersion string
 }
 
 // reconciler reconciles gatewayclasses.
@@ -130,6 +156,9 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	var errs []error
 	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
+	}
+	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx); err != nil {
+		errs = append(errs, err)
 	}
 	if _, _, err := r.ensureServiceMeshControlPlane(ctx, &gatewayclass); err != nil {
 		errs = append(errs, err)
--- a/pkg/operator/controller/gatewayclass/subscription.go
+++ b/pkg/operator/controller/gatewayclass/subscription.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
@@ -26,7 +27,7 @@ func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context) 
 		return false, nil, err
 	}
 
-	desired, err := desiredSubscription(name)
+	desired, err := desiredSubscription(name, r.config.GatewayAPIOperatorChannel, r.config.GatewayAPIOperatorVersion)
 	if err != nil {
 		return have, current, err
 	}
@@ -48,18 +49,19 @@ func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context) 
 }
 
 // desiredSubscription returns the desired subscription.
-func desiredSubscription(name types.NamespacedName) (*operatorsv1alpha1.Subscription, error) {
+func desiredSubscription(name types.NamespacedName, gwapiOperatorChannel, gwapiOperatorVersion string) (*operatorsv1alpha1.Subscription, error) {
 	subscription := operatorsv1alpha1.Subscription{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: name.Namespace,
 			Name:      name.Name,
 		},
 		Spec: &operatorsv1alpha1.SubscriptionSpec{
-			Channel:                "stable",
-			InstallPlanApproval:    operatorsv1alpha1.ApprovalAutomatic,
+			Channel:                gwapiOperatorChannel,
+			InstallPlanApproval:    operatorsv1alpha1.ApprovalManual,
 			Package:                "servicemeshoperator",
 			CatalogSource:          "redhat-operators",
 			CatalogSourceNamespace: "openshift-marketplace",
+			StartingCSV:            gwapiOperatorVersion,
 		},
 	}
 	return &subscription, nil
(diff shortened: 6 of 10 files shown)


rfredette:
test failed due to some test infra issue: 
``` 
error: unable to read image registry.build02.ci.openshift.org/ci-op-hx0k37ig/stable@sha256:e3f1d39aa298ba7049f7d28ecf4ab8d4d49c54918355e8fb37bd94a2050434fa: unable to retrieve source image registry.build02.ci.openshift.org/ci-op-hx0k37ig/stable@sha256:e3f1d39aa298ba7049f7d28ecf4ab8d4d49c54918355e8fb37bd94a2050434fa manifest #1 from manifest list: received unexpected HTTP status: 504 Gateway Time-out ```

/retest

**Comment by rfredette** [comment]
single node test failure is unrelated.
/test e2e-aws-ovn-single-node

File: pkg/operator/controller/gatewayclass/controller.go [hunk 1]
```diff
@@ -61,6 +62,22 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.GatewayClass{}, &handler.EnqueueRequestForObject{}, isOurGatewayClass, notIstioGatewayClass)); err != nil {
 		return nil, err
 	}
+	isOurInstallPlan := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		installPlan := o.(*operatorsv1alpha1.InstallPlan)
+		for _, csv := range installPlan.Spec.ClusterServiceVersionNames {
```
candita, line 67: Should you check to make sure `len(ClusterServiceVersionNames) > 0` first?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+	currentInstallPlan, err := r.getCurrentInstallPlan(ctx)
+	if err != nil {
+		return false, nil, err
+	} else if currentInstallPlan == nil {
+		return false, nil, nil
+	}
+	if !currentInstallPlan.Spec.Approved {
+		currentInstallPlan.Spec.Approved = true
+		if err := r.client.Update(ctx, currentInstallPlan); err != nil {
+			return false, nil, fmt.Errorf("Failed to update %s/%s: %w", currentInstallPlan.Namespace, currentInstallPlan.Name, err)
```
candita, line 138:
```suggestion
			return false, nil, fmt.Errorf("Failed to update %s/%s: %v", currentInstallPlan.Namespace, currentInstallPlan.Name, err)
```

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
 
 	return true, updated
 }
+
+// ensureServiceMeshOperatorInstallPlan attempts to ensure that the install plan for the appropriate OSSM operator
+// version is approved.
+func (r *reconciler) ensureServiceMeshOperatorInstallPlan(ctx context.Context) (bool, *operatorsv1alpha1.InstallPlan, error) {
```
candita, line 128: We never use the first two return values (bool, *operatorsv1alpha1.InstallPlan), so maybe just return the error.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+		return nil, nil
+	}
+	for _, InstallPlan := range InstallPlans.Items {
+		for _, CSVName := range InstallPlan.Spec.ClusterServiceVersionNames {
+			if CSVName == serviceMeshOperatorDesiredVersion {
+				return &InstallPlan, nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("No InstallPlan with cluster service version %s found", serviceMeshOperatorDesiredVersion)
```
candita, line 160:
```suggestion
	return nil, fmt.Errorf("no InstallPlan with cluster service version %s found in namespace %s", serviceMeshOperatorDesiredVersion, serviceMeshOperatorNamespace)
```

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+		currentInstallPlan.Spec.Approved = true
+		if err := r.client.Update(ctx, currentInstallPlan); err != nil {
+			return false, nil, fmt.Errorf("Failed to update %s/%s: %w", currentInstallPlan.Namespace, currentInstallPlan.Name, err)
+		}
+		return true, currentInstallPlan, nil
+	}
+	return false, currentInstallPlan, nil
+}
+
+func (r *reconciler) getCurrentInstallPlan(ctx context.Context) (*operatorsv1alpha1.InstallPlan, error) {
```
candita, line 145: Might be better to use the function name we normally use.
```suggestion
func (r *reconciler) currentInstallPlan(ctx context.Context) (*operatorsv1alpha1.InstallPlan, error) {
```

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+// ensureServiceMeshOperatorInstallPlan attempts to ensure that the install plan for the appropriate OSSM operator
+// version is approved.
+func (r *reconciler) ensureServiceMeshOperatorInstallPlan(ctx context.Context) (bool, *operatorsv1alpha1.InstallPlan, error) {
+	currentInstallPlan, err := r.getCurrentInstallPlan(ctx)
+	if err != nil {
+		return false, nil, err
+	} else if currentInstallPlan == nil {
+		return false, nil, nil
+	}
+	if !currentInstallPlan.Spec.Approved {
```
candita, line 135: The idiomatic way to write an "ensure" function is to get the current object, get the desired object, compare them, and use the "have" and "want" values to determine whether to create or  update.  Can you update this to use the idiomatic form?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +122,40 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+			return false, nil, fmt.Errorf("Failed to update %s/%s: %w", currentInstallPlan.Namespace, currentInstallPlan.Name, err)
+		}
+		return true, currentInstallPlan, nil
+	}
+	return false, currentInstallPlan, nil
+}
+
+func (r *reconciler) getCurrentInstallPlan(ctx context.Context) (*operatorsv1alpha1.InstallPlan, error) {
+	InstallPlans := &operatorsv1alpha1.InstallPlanList{}
+	if err := r.client.List(ctx, InstallPlans, client.InNamespace(serviceMeshOperatorNamespace)); err != nil {
```
candita, line 147: So the installPlan doesn't have a unique name?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -16,6 +17,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var (
+	serviceMeshOperatorDesiredVersion = "servicemeshoperator.v2.6.2"
```
candita, line 21: I'm not sure this is going to work.  If we change the version in a future release and need to backport a different change in the same file, won't this get overwritten?  I think normally version information is recorded in a manifest yaml.  Can you find a way to do this without hardcoding the version?

candita:
@rfredette can you add a unit test for the ensure function?

pkg/operator/controller/gatewayclass/subscription.go
Miciah: I'm not too worried about merge conflicts, but I wouldn't object to putting the version string in a separate `.go` file, or putting it in a yaml manifest, maybe as an environment variable, à la `WATCH_NAMESPACE` etc., which would also make it easier to override for testing and development.  If we do define it in Go code, it should be a `const`.

pkg/operator/controller/gatewayclass/subscription.go
rfredette: Unfortunately no; they appear to all just be named "install-<hash>".

pkg/operator/controller/gatewayclass/subscription.go
rfredette: fixed

pkg/operator/controller/gatewayclass/subscription.go
rfredette: fixed

pkg/operator/controller/gatewayclass/controller.go
rfredette: It's probably not necessary; From what I can tell, an InstallPlan is primarily a list of ClusterServiceVersions, so having one with 0 CSVs would be weird, but better safe than sorry, so I've included it in the latest commit

cmd/ingress-operator/start.go
```diff
@@ -74,6 +80,8 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&options.ReleaseVersion, "release-version", "", statuscontroller.UnknownVersionValue, "the release version the operator should converge to (required)")
 	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", "127.0.0.1:60000", "metrics endpoint listen address (required)")
 	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "if provided, shut down the operator when this file changes")
+	cmd.Flags().StringVarP(&options.GatewayAPIOperatorChannel, "gateway-api-operator-channel", "", defaultGatewayAPIOperatorVersion, "release channel of the Gateway API implementation to install if Gateway API is enabled")
+	cmd.Flags().StringVarP(&options.GatewayAPIOperatorVersion, "gateway-api-operator-version", "", defaultGatewayAPIOperatorVersion, "name and release of the Gateway API implementation to install if Gateway API is enabled")
```
candita, line 84: nit: 
You could remove the "if Gateway API is enabled" phrase from the flag text (and comments).  Within this release we'll reach a point when it will always be enabled.  Please remove it if you make any other changes.

manifests/02-deployment-ibm-cloud-managed.yaml
```diff
@@ -35,6 +35,10 @@ spec:
         - $(CANARY_IMAGE)
         - --release-version
         - $(RELEASE_VERSION)
+        - --gateway-api-operator-channel
+        - $(GATEWAY_API_OPERATOR_CHANNEL)
+        - --gateway-api-operator-version
+        - $(GATEWAY_API_OPERATOR_VERSION)
```
candita, line 41: To keep consistency with 02-deployment.yaml, it might be better quote these, and fix the other $(VALUE) values to be quoted as well, "$(VALUE)".

pkg/operator/config/config.go
```diff
@@ -15,5 +15,11 @@ type Config struct {
 	// CanaryImage is the ingress operator image, which runs a canary command.
 	CanaryImage string
 
+	// GatewayAPIOperatorChannel is the release channel of the Gateway API implementation to install if Gateway API is enabled.
+	GatewayAPIOperatorChannel string
+
+	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install if Gateway API is enabled.
```
candita, line 21:
```suggestion
	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
```

manifests/02-deployment-ibm-cloud-managed.yaml
rfredette: I initially did that, but it failed the verify job. Running the `hack/update-profile-manifests.sh` script generates an unquoted version.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -61,6 +63,24 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
...
+				}
+			}
+		}
+		return false
+	})
+	isInstallPlanApproved := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		installPlan := o.(*operatorsv1alpha1.InstallPlan)
+		return installPlan.Spec.Approved
+	})
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.InstallPlan{}, enqueueRequestForDefaultGatewayClassController(config.OperatorNamespace), isOurInstallPlan, predicate.Not(isInstallPlanApproved))); err != nil {
```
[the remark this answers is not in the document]
rfredette: The ingress operator only needs to move the InstallPlan from not approved to approved for OLM to actually install things, so we don't need a reconcile event for any InstallPlans that are already approved.

Maybe we'd care about the phase of the InstallPlan if it's getting stuck somewhere and we want to, like, go degraded or put up some other status condition, but for the purposes of getting the gwapi implementation running, I don't think it's necessary.

pkg/operator/operator.go
```diff
@@ -147,6 +147,7 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 				operatorcontroller.DefaultOperandNamespace:               {},
 				operatorcontroller.DefaultCanaryNamespace:                {},
 				operatorcontroller.GlobalMachineSpecifiedConfigNamespace: {},
+				"openshift-operators":                                    {},
```
candita, line 150: Please make a global constant in the operatorcontroller package for "openshift-operators" namespace.  Just in case it changes, we can change it in one place.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -48,18 +53,19 @@ func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context)
...
 	subscription := operatorsv1alpha1.Subscription{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: name.Namespace,
 			Name:      name.Name,
 		},
 		Spec: &operatorsv1alpha1.SubscriptionSpec{
-			Channel:                "stable",
-			InstallPlanApproval:    operatorsv1alpha1.ApprovalAutomatic,
+			Channel:                r.config.GatewayAPIOperatorChannel,
+			InstallPlanApproval:    operatorsv1alpha1.ApprovalManual,
```
candita, line 64: What happens to the variable overrides that might be passed in via flags?  Do you need to look for overrides in the env variables at this stage?  Otherwise, why use flag opts?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +121,76 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+		for _, CSVName := range InstallPlan.Spec.ClusterServiceVersionNames {
+			if CSVName == r.config.GatewayAPIOperatorVersion {
+				return true, &InstallPlan, nil
+			}
+		}
+	}
+	return false, nil, fmt.Errorf("no InstallPlan with cluster service version %s found in namespace %s", r.config.GatewayAPIOperatorVersion, serviceMeshOperatorNamespace)
+}
+
+func desiredInstallPlan(current *operatorsv1alpha1.InstallPlan) (*operatorsv1alpha1.InstallPlan, error) {
```
candita, line 168: `desiredInstallPlan` would normally have all the fields set to what we want them to be.   It should not be a copy of the current install plan, even if you expect that will never change.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -16,6 +17,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	serviceMeshOperatorNamespace = "openshift-operators"
```
candita, line 21: This namespace should be declared a global default as mentioned in https://github.com/openshift/cluster-ingress-operator/pull/1112/files?diff=split&w=1#r1958739919.

pkg/operator/controller/gatewayclass/subscription.go
rfredette: I would generally agree, but I think this is a special case. The InstallPlan is really owned by the OLM operator, and the ingress operator is only managing the Approved field of this particular InstallPlan. Maybe there's a different way to write it or there's something I can write in the comments to explain why it deviates from the usual pattern, but do think this is the right logic for this particular case.

pkg/operator/controller/gatewayclass/subscription.go
rfredette: I'm not if I understand the question, but I chose to follow the same pattern as the existing env variables, like `$RELEASE_VERSION` or `$CANARY_IMAGE`. They are also defined as env variables, but actually passed to the operator through flag opts.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +121,76 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+		return false, nil, err
+	}
+	switch {
+	case !haveInstallPlan:
+		// The OLM operator creates the initial InstallPlan, so if it doesn't exist yet, do nothing.
+		return false, nil, nil
+	case haveInstallPlan:
+		desired, err := desiredInstallPlan(current)
+		if err != nil {
+			return true, current, err
```
candita, line 139: At the moment, `desiredInstallPlan` only returns a nil error.  If that doesn't change, it's probably fine to remove the returned err and don't check it.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +121,76 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+		} else if updated {
+			return r.currentInstallPlan(ctx)
+		}
+	}
+	return false, current, nil
+}
+
+func (r *reconciler) currentInstallPlan(ctx context.Context) (bool, *operatorsv1alpha1.InstallPlan, error) {
+	InstallPlans := &operatorsv1alpha1.InstallPlanList{}
```
candita, line 151: nit: `InstallPlans` should be lower case, it's local.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -97,6 +136,9 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
 		errs = append(errs, err)
 	}
+	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx); err != nil {
```
candita, line 139: Please add a unit test for this function, or the Reconcile function, using the pattern in the gatewayapi/controller_test.go if possible.  At least make sure the install plan is not created if it's missing, and that it is properly reconciled if changes are attempted that are not allowed.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +121,76 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+	if err := r.client.List(ctx, InstallPlans, client.InNamespace(serviceMeshOperatorNamespace)); err != nil {
+		return false, nil, err
+	}
+	if InstallPlans == nil || len(InstallPlans.Items) == 0 {
+		return false, nil, nil
+	}
+	for _, InstallPlan := range InstallPlans.Items {
+		for _, CSVName := range InstallPlan.Spec.ClusterServiceVersionNames {
+			if CSVName == r.config.GatewayAPIOperatorVersion {
```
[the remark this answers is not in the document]
rfredette: I don't think the phase or approved values are needed to know if it's the correct InstallPlan, but looking at this more, there is an owner reference that should point to the Subscription we create. I can change this to look at the owner ref and the CSV name (which includes the version number), which should be a more robust way to find the right InstallPlan.

cmd/ingress-operator/start.go
```diff
@@ -74,6 +80,8 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&options.ReleaseVersion, "release-version", "", statuscontroller.UnknownVersionValue, "the release version the operator should converge to (required)")
 	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", "127.0.0.1:60000", "metrics endpoint listen address (required)")
 	cmd.Flags().StringVarP(&options.ShutdownFile, "shutdown-file", "s", defaultTrustedCABundle, "if provided, shut down the operator when this file changes")
+	cmd.Flags().StringVarP(&options.GatewayAPIOperatorChannel, "gateway-api-operator-channel", "", defaultGatewayAPIOperatorVersion, "release channel of the Gateway API implementation to install if Gateway API is enabled")
```
Miciah, line 83:
```suggestion
	cmd.Flags().StringVarP(&options.GatewayAPIOperatorChannel, "gateway-api-operator-channel", "", defaultGatewayAPIOperatorChannel "release channel of the Gateway API implementation to install if Gateway API is enabled")
```

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -48,18 +53,19 @@ func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context)
 }
 
 // desiredSubscription returns the desired subscription.
-func desiredSubscription(name types.NamespacedName) (*operatorsv1alpha1.Subscription, error) {
+func (r *reconciler) desiredSubscription(name types.NamespacedName) (*operatorsv1alpha1.Subscription, error) {
```
Miciah, line 56: Better to pass the required values as explicit parameters rather than passing in the whole reconciler as a receiver.  We want the `desired` functions not to depend on the client.

rfredette:
multiple cluster bootstrap failures, probably unrelated to this change.
/retest

pkg/operator/controller/gatewayclass/controller.go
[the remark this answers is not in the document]
candita: I meant a comment in the code, since the way it works isn't understood by everyone.  Please add an excerpt from above if you need to make any other changes.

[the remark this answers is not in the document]
rfredette: Ah, I see. Sure, I'll add a clarifying comment

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
candita, line 19: Please alphabetize the package imports.

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+)
+
+// func (r *reconciler) ensureServiceMeshOperatorInstallPlan(ctx context.Context) (bool, *operatorsv1alpha1.InstallPlan, error) {
```
candita, line 22: Forgot to remove this comment.

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+						Approval: operatorsv1alpha1.ApprovalManual,
+						Approved: true,
+					},
+				},
+			},
+			expectDelete: []client.Object{},
+		},
+		{
+			// Multiple InstallPlans exist, and the desired InstallPlan is not approved. Expect InstallPlan.Spec.Approved = true only for the desired InstallPlan
+			name:    "Multiple InstallPlans",
```
candita, line 150:
```suggestion
			name:    "Multiple installPlans, none approved",
```

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+						},
+						Approval: operatorsv1alpha1.ApprovalManual,
+						Approved: true,
+					},
+				},
+			},
+			expectDelete: []client.Object{},
+		},
+		{
+			// An InstallPlan with the correct version exists, but it's owned by a different subscription. Expect no change.
```
candita, line 230: What does "owned by a different subscription" mean? I see the UIDs are different, are the UIDs are good indicator of the owership?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -115,3 +117,91 @@ func subscriptionChanged(current, expected *operatorsv1alpha1.Subscription) (boo
...
+// ensureServiceMeshOperatorInstallPlan attempts to ensure that the install plan for the appropriate OSSM operator
+// version is approved.
+func (r *reconciler) ensureServiceMeshOperatorInstallPlan(ctx context.Context) (bool, *operatorsv1alpha1.InstallPlan, error) {
+	haveInstallPlan, current, err := r.currentInstallPlan(ctx)
+	if err != nil {
+		return false, nil, err
+	}
+	switch {
+	case !haveInstallPlan:
+		// The OLM operator creates the initial InstallPlan, so if it doesn't exist yet, do nothing.
```
candita, line 130: What if it somehow got deleted?

candita:
@rfredette generally looks good.  Just a couple of questions/comments.

melvinjoseph86:
Hi @rfredette 
I tried to create a cluster using this PR and here are the results

```
mjoseph@mjoseph-mac Downloads % oc -n openshift-operators get subscription.operators.coreos.com/servicemeshoperator -oyaml 
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
 creationTimestamp: "2025-03-03T11:27:05Z"
 generation: 1
 labels:
  operators.coreos.com/servicemeshoperator.openshift-operators: ""
 name: servicemeshoperator
 namespace: openshift-operators
 resourceVersion: "84872"
 uid: 77756e5f-6294-4477-9a27-f1b922433c17
spec:
 channel: stable
 installPlanApproval: Manual
 name: servicemeshoperator
 source: redhat-operators
 sourceNamespace: openshift-marketplace
 startingCSV: servicemeshoperator.v2.6.2

mjoseph@mjoseph-mac Downloads % oc -n openshift-ingress get servicemeshcontrolplanes
NAME                READY   STATUS            PROFILES      VERSION   AGE
openshift-gateway   5/5     ComponentsReady   ["default"]   2.6.2     73s

```
for  the other cluster's it is 2.6.4. Is this intended?
Can you let me know what else needs to be tested or checked as per this PR?

candita:
> Hi @rfredette I tried to create a cluster using this PR and here are the results
> 
> ```
> mjoseph@mjoseph-mac Downloads % oc -n openshift-operators get subscription.operators.coreos.com/servicemeshoperator -oyaml 
> apiVersion: operators.coreos.com/v1alpha1
> kind: Subscription
> metadata:
>  creationTimestamp: "2025-03-03T11:27:05Z"
>  generation: 1
>  labels:
>   operators.coreos.com/servicemeshoperator.openshift-operators: ""
>  name: servicemeshoperator
>  namespace: openshift-operators
>  resourceVersion: "84872"
>  uid: 77756e5f-6294-4477-9a27-f1b922433c17
> spec:
>  channel: stable
>  installPlanApproval: Manual
>  name: servicemeshoperator
>  source: redhat-operators
>  sourceNamespace: openshift-marketplace
>  startingCSV: servicemeshoperator.v2.6.2
> 
> mjoseph@mjoseph-mac Downloads % oc -n openshift-ingress get servicemeshcontrolplanes
> NAME                READY   STATUS            PROFILES      VERSION   AGE
> openshift-gateway   5/5     ComponentsReady   ["default"]   2.6.2     73s
> ```
> 
> for the other cluster's it is 2.6.4. Is this intended? Can you let me know what else needs to be tested or checked as per this PR?

Hi @melvinjoseph86 - yes this is correct, as this PR is pinning the version rather than letting the deployment choose the latest one available.  See https://github.com/openshift/cluster-ingress-operator/pull/1112/files#diff-24ce9f24fd9e7e47b7ac69fa0b597fe0470c1ba7b8257d0fa6537c9caf1b5deeR37

pkg/operator/controller/gatewayclass/subscription_test.go
rfredette: The main concern with "owned by a different subscription" is if a user installed OSSM separately, but in theory there could also be another OLM package that includes OSSM in its install plan. When OLM creates install plans, it includes an owner reference to the subscription that it was created for.

If I'm reading the k8s docs correctly (for completeness, [short description of UIDs here](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids)), UIDs for a given resource are guaranteed to be unique, so matching the subscription's UID to the UID in an owner reference is sufficient to make sure we only approve install plans for the subscription created by the ingress operator.

pkg/operator/controller/gatewayclass/subscription.go
rfredette: If the install plan gets deleted, we rely on the OLM operator to recreate it.

(I'll update the comment to say something to that effect too)

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+		name            string
+		channel         string
+		version         string
+		existingObjects []runtime.Object
+		expectCreate    []client.Object
+		expectUpdate    []client.Object
+		expectDelete    []client.Object
+	}{
+		{
+			// No InstallPlan found; no changes expected.
```
candita, line 34: Can you add a test that would expect the install plan to be created?

pkg/operator/controller/gatewayclass/subscription_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+func Test_ensureServiceMeshOperatorInstallPlan(t *testing.T) {
+	tests := []struct {
+		name            string
+		channel         string
+		version         string
+		existingObjects []runtime.Object
+		expectCreate    []client.Object
+		expectUpdate    []client.Object
+		expectDelete    []client.Object
```
candita, line 31: We would never delete an install plan, right?  So we don't need `expectDelete`.

cmd/ingress-operator/start.go
```diff
@@ -32,7 +32,9 @@ import (
 const (
 	// defaultTrustedCABundle is the fully qualified path of the trusted CA bundle
 	// that is mounted from configmap openshift-ingress-operator/trusted-ca.
-	defaultTrustedCABundle = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	defaultTrustedCABundle           = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	defaultGatewayAPIOperatorChannel = "stable"
+	defaultGatewayAPIOperatorVersion = "servicemeshoperator.v2.6.2"
```
shaneutt, line 37: Isn't it:

```suggestion
	defaultGatewayAPIOperatorVersion = "servicemeshoperator.v3.0.0"
```

(or am I misunderstanding)?

manifests/02-deployment.yaml
```diff
@@ -75,6 +79,10 @@ spec:
               value: openshift/origin-haproxy-router:v4.0
             - name: CANARY_IMAGE
               value: openshift/origin-cluster-ingress-operator:latest
+            - name: GATEWAY_API_OPERATOR_CHANNEL
+              value: stable
+            - name: GATEWAY_API_OPERATOR_VERSION
+              value: servicemeshoperator.v2.6.2
```
shaneutt, line 85:
```suggestion
              value: servicemeshoperator.v3.0.0
```

pkg/operator/controller/gatewayclass/subscription_test.go
rfredette: The ingress operator should never create or delete an install plan. Both of those operations should always be handled by OLM.

pkg/operator/controller/gatewayclass/subscription_test.go
rfredette: We never create or delete install plans, but verifying that nothing is created or deleted is still a helpful test signal. If there are other changes that are needed, I can update the test to get rid of the `expectCreate` and `expectDelete` variables, and always check that there are no creates or deletes, but I think this accomplishes the same thing in a similarly readable way.

cmd/ingress-operator/start.go
rfredette: Yeah, this PR doesn't include any changes for OSSM 3.0 support, so if we merged it before #1152, it would break things. The subscription management logic will be the same, but we'll need to change the package, channel, and version strings.

rfredette:
Tests took too long and timed out. Probably unrelated
/retest

PR #1113, opened 2024-07-29
[WIP] NE-1788: Enable Dynamic Configuration Manager by default, limit dynamic servers to two

alebedev87:


alebedev87:
The following tests need to be checked as they fail on all the providers:
```
        --- FAIL: TestAll/parallel/TestConnectTimeout (265.86s)
        --- FAIL: TestAll/parallel/TestHeaderNameCaseAdjustment (384.03s)
        --- FAIL: TestAll/parallel/TestSetRouteResponseHeaders (325.01s)
```

alebedev87:
Disabled the blueprints by hard resetting the pool size to 0. [The pool size flag](https://github.com/openshift/router/blob/master/pkg/cmd/infra/router/template.go#L183) was used to bypass the lower bound limit of the envvar.

alebedev87:
`TestAll/parallel/TestCRLUpdate/certificate-distributes-its-own-crl` fails for `aws-operator` and `gcp-operator`.

alebedev87:
The following origin tests fail on all 3 major providers (`e2e-<provider>-ovn` jobs) and must be checked:
-  `[sig-network-edge][Conformance][Area:Networking][Feature:Router][apigroup:route.openshift.io][apigroup:config.openshift.io] The HAProxy router should pass the http2 tests`
- `[sig-network][Feature:Router] The HAProxy router should expose prometheus metrics for a route`
- `[sig-network-edge][Feature:Idling] Idling with a single service and ReplicationController should idle the service and ReplicationController properly`
- ` [sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should be able to connect to a service that is idled because a GET on the route will unidle it`

alebedev87:
_Test_: `[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should be able to connect to a service that is idled because a GET on the route will unidle it`
_Platform_: AWS
_OCP version_: `4.17.0-0.ci-2024-08-30-050930`

_Conclusion_: failures found while waiting for the idled service to get the special idled annotations (`STEP: Checking the idle targets`). At the same time, quite some successful passes too, the failure rate is about 40%, every second run gets stuck on waiting for the idle annotations. It's hard to say how this is related to the router because the idling of the service is not done by it or by any NetworkEdge component. [A search through the CI logs](https://search.dptools.openshift.org/?search=%5C%5Bgithub.com%2Fopenshift%2Forigin%2Ftest%2Fextended%2Frouter%2Fidle.go%3A49%5C%5D%3A+Interrupted+by+User&maxAge=48h&context=1&type=junit&name=4.17&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) shows that this job fails in the same place for other PRs too.

alebedev87:
_Test_: `[sig-network-edge][Conformance][Area:Networking][Feature:Router][apigroup:route.openshift.io][apigroup:config.openshift.io] The HAProxy router should pass the http2 tests`
_Platform_: GCP
_OCP version_: 4.17.0-0.ci-2024-08-30-050930

_Conclusion_: After running the test from my laptop. I found failures similar to the CI logs. The backend keeps replying with HTTP/1.1 protocol even if all the other parties use HTTP/2 (client, frontend, workload). The problem was seen only on reencrypt routes, edge and passthrough were passing ok. The root cause of this is [the missing ALPN setting on the server directive in HAProxy config](https://github.com/openshift/router/blob/47b8420185ec2eb4c7861483921d31c65b912625/images/router/haproxy/conf/haproxy-config.template#L757).

alebedev87:
https://github.com/openshift/cluster-ingress-operator/pull/1159 takes over from here.

PR #1115, opened 2024-08-01
OCPBUGS-31550: Gateway API - recreating SMCP which breaks Gateway API

anirudhAgniRedhat:
Recreating the SMCP and subscription, in case the SMCP or subscriptions are deleted manually.

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -2,13 +2,20 @@ package gatewayclass
 
 import (
 	"context"
+	"fmt"
+	"sync"
 
 	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	maistrav2 "github.com/maistra/istio-operator/pkg/apis/maistra/v2"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	"k8s.io/client-go/tools/record"
 
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
@@ -35,6 +42,7 @@ const (
 )
 
 var log = logf.Logger.WithName(controllerName)
+var gatewayClassController controller.Controller
 
 // NewUnmanaged creates and returns a controller that watches gatewayclasses and
 // installs and configures Istio.  This is an unmanaged controller, which means
@@ -61,6 +69,15 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.GatewayClass{}, &handler.EnqueueRequestForObject{}, isOurGatewayClass, notIstioGatewayClass)); err != nil {
 		return nil, err
 	}
+
+	isServiceMeshSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == operatorcontroller.ServiceMeshSubscriptionName().Name
+	})
+	if err = c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.Subscription{},
+		enqueueRequestForDefaultGatewayClassController(config.OperandNamespace), isServiceMeshSubscription)); err != nil {
+		return nil, err
+	}
+	gatewayClassController = c
 	return c, nil
 }
 
@@ -81,6 +98,23 @@ type reconciler struct {
 	client   client.Client
 	cache    cache.Cache
 	recorder record.EventRecorder
+
+	startSMCPWatch sync.Once
+}
+
+func enqueueRequestForDefaultGatewayClassController(namespace string) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, a client.Object) []reconcile.Request {
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: namespace,
+						Name:      OpenShiftDefaultGatewayClassName,
+					},
+				},
+			}
+		},
+	)
 }
 
 // Reconcile expects request to refer to a GatewayClass and creates or
@@ -95,10 +129,23 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	var errs []error
 	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
-		errs = append(errs, err)
+		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
 	}
 	if _, _, err := r.ensureServiceMeshControlPlane(ctx, &gatewayclass); err != nil {
 		errs = append(errs, err)
+	} else {
+		// The OSSM operator installs the maistra.io CRDs, Need to create the SMCP watch after the OSSM operator is installed.
+		// Using sync.Once here, to start the watch for SMCP, which should run only once.
+		r.startSMCPWatch.Do(func() {
+			isOurSMCP := predicate.NewPredicateFuncs(func(o client.Object) bool {
+				return o.GetName() == operatorcontroller.ServiceMeshControlPlaneName(r.config.OperandNamespace).Name
+			})
+			if err = gatewayClassController.Watch(source.Kind[client.Object](r.cache, &maistrav2.ServiceMeshControlPlane{}, enqueueRequestForDefaultGatewayClassController(r.config.OperandNamespace), isOurSMCP)); err != nil {
+				log.Error(err, "failed to watch ServiceMeshControlPlane", "request", request)
+				errs = append(errs, err)
+			}
+		})
 	}
+
 	return reconcile.Result{}, utilerrors.NewAggregate(errs)
 }
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -289,6 +289,51 @@ func assertSubscription(t *testing.T, namespace, subName string) error {
 	return err
 }
 
+// deleteExistingSubscription deletes if the subscription of the given name exists and returns an error if not.
+func deleteExistingSubscription(t *testing.T, namespace, subName string) error {
+	t.Helper()
+	existingSubscription := &operatorsv1alpha1.Subscription{}
+	newSubscription := &operatorsv1alpha1.Subscription{}
+	nsName := types.NamespacedName{Namespace: namespace, Name: subName}
+
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, nsName, existingSubscription); err != nil {
+			t.Logf("failed to get Subscription %s: %v", nsName.Name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("timed out getting subscription %s: %w", nsName.Name, err)
+	}
+	// deleting Subscription.
+	err = kclient.Delete(context.Background(), existingSubscription)
+	if err != nil {
+		return err
+	}
+	err = wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, nsName, newSubscription); err != nil {
+			if kerrors.IsNotFound(err) {
+				return true, nil
+			}
+			t.Logf("failed to get deleted subscription %s: %v", nsName.Name, err)
+			return false, nil
+		}
+		// if new Subscription got recreated while the poll ensures the Subscription is deleted.
+		if newSubscription != nil && newSubscription.UID != existingSubscription.UID {
+			return true, nil
+		}
+		t.Logf("Subscription %s still exists", nsName.Name)
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("timed out waiting for Subscription %s to be deleted: %v", nsName.Name, err)
+	}
+	t.Logf("deleted Subscription %s", nsName.Name)
+	return nil
+
+}
+
 // assertOSSMOperator checks if the OSSM Istio operator gets successfully installed
 // and returns an error if not.
 func assertOSSMOperator(t *testing.T) error {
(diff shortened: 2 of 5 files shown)


anirudhAgniRedhat:
/ hold

need to add watch

anirudhAgniRedhat:
/hold
need to add watch

pkg/operator/controller/gatewayclass/servicemeshcontrolplane.go
```diff
@@ -24,7 +24,7 @@ import (
 func (r *reconciler) ensureServiceMeshControlPlane(ctx context.Context, gatewayclass *gatewayapiv1beta1.GatewayClass) (bool, *maistrav2.ServiceMeshControlPlane, error) {
 	name := controller.ServiceMeshControlPlaneName(r.config.OperandNamespace)
 	have, current, err := r.currentServiceMeshControlPlane(ctx, name)
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
```
rfredette, line 27: I think we can drop this change and the equivalent in `subscription.go`; the existing code is consistent with other `ensureX` and `currentX` functions elsewhere in this repo.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -61,6 +66,46 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
...
+			NamespacedName: operatorcontroller.ServiceMeshControlPlaneName(config.OperandNamespace),
+		}}
+	}
+
+	isOurSMCP := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == operatorcontroller.ServiceMeshControlPlaneName(config.OperandNamespace).Name
+	})
+
+	if err = c.Watch(source.Kind[client.Object](operatorCache, &maistrav2.ServiceMeshControlPlane{},
+		handler.EnqueueRequestsFromMapFunc(toServiceMeshControlPlane), isOurSMCP, predicate.Funcs{
```
rfredette, line 81: This reconciler expects a gatewayclass name, not the name of an SMCP. This should either enqueue a request for the default gatewayclass (similar to [enqueueRequestForDefaultIngressController](https://github.com/openshift/cluster-ingress-operator/blob/2639954d5a10378ad4702faadac2f86806e714ad/pkg/operator/controller/canary/controller.go#L167)), or the gatewayclass that owns the SMCP (via the ownerref metadata). With the rest of the code as it is now, either would be equivalent since `isOurSMCP` selects for just the SMCP of the default gatewayclass.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -61,6 +66,46 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
...
+			NamespacedName: operatorcontroller.ServiceMeshSubscriptionName(),
+		}}
+	}
+
+	isServiceMeshSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == operatorcontroller.ServiceMeshSubscriptionName().Name &&
+			o.GetNamespace() == operatorcontroller.ServiceMeshSubscriptionName().Namespace
+	})
+	if err = c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.Subscription{},
+		handler.EnqueueRequestsFromMapFunc(toServiceMeshSubscription), isServiceMeshSubscription, predicate.Funcs{
```
rfredette, line 101: Like above, this should enqueue the default gatewayclass's name, not the subscription name.

rfredette:
Hmm, on my cluster, adding a watch for SMCPs when the operator starts is causing some issues. 
The operator logs are hitting the error that the `ServiceMeshControlPlane.maistra.io` CRD doesn't exist: 
> `2024-08-09T18:41:15.444Z        ERROR   operator.init.controller-runtime.source.EventHandler    wait/loop.go:87 if kind is a CRD, it should be installed before calling Start{"kind": "ServiceMeshControlPlane.maistra.io", "error": "failed to get restmapping: no matches for kind \"ServiceMeshControlPlane\" in group \"maistra.io\""}`

The OSSM operator installs the `maistra.io` CRDs, but because of this error, the gatewayclass controller never starts, and therefore it never creates the subscription for the OSSM operator.

We need to create the SMCP watch after the OSSM operator is installed. I messed around with the code and got a version where it seems to work. It may still need some polishing, but this is what I changed on my end: https://gist.github.com/rfredette/366a1f36ed4cf7425a5cc0f4ef270a24

candita:
@anirudhAgniRedhat please add an e2e test for this to the existing e2e-aws-gatewayapi test.

anirudhAgniRedhat:
Hey @candita Added the e2e  tests in `testGatewayAPIIstioInstallation` tests which is passing. 
I  think test `testGatewayAPIObjects` which is failing is  irrespective of my change and is failing across different PRs.
PTAL Thanks!!

gcs278:
DNS Propagation bug should be fixed, retesting gatewayapi:
/test e2e-aws-gatewayapi

anirudhAgniRedhat:
Thanks @gcs278 for the update!!!
/test e2e-azure-ovn

rhamini3:
pre-merge-verified

`1. create the gatewayclass and gateway 
oc create -f -<<EOF
apiVersion: gateway.networking.k8s.io/v1beta1
kind: GatewayClass
metadata:
  name: openshift-default
spec:
  controllerName: openshift.io/gateway-controller
EOF

oc create -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gateway
  namespace: openshift-ingress
spec:
  gatewayClassName: openshift-default
  listeners:
  - name: http
    hostname: "*.$gwapi_domain"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
  - name: https
    hostname: "*.$gwapi_domain"
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - name: gwapi-wildcard
    allowedRoutes:
      namespaces:
        from: All
EOF`

`2. Delete servicemeshcontrolplane and check if it is recreated
 % oc delete servicemeshcontrolplane openshift-gateway -n openshift-ingress
servicemeshcontrolplane.maistra.io "openshift-gateway" deleted
iamin@iamin-mac Downloads % oc get servicemeshcontrolplane --all-namespaces                         
NAMESPACE           NAME                READY   STATUS            PROFILES      VERSION   AGE
openshift-ingress   openshift-gateway   5/5     ComponentsReady   ["default"]   2.6.3     6s`

`3. create the HTTProute and check whether it is properly created and connected to the gateway
oc create -f - <<EOF                                               
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: myroute1
spec:
  parentRefs:
  - name: gateway
    namespace: openshift-ingress
  hostnames: ["test.$gwapi_domain"]
  rules:
  - backendRefs:
    - name: service-unsecure
      port: 27017
EOF
httproute.gateway.networking.k8s.io/myroute1 created

% oc get httproute myroute1 -oyaml
kind: HTTPRoute
metadata:
  creationTimestamp: "2024-12-05T17:40:06Z"
  generation: 1
  name: myroute1
  namespace: default
  resourceVersion: "56779"
  uid: 1fc69414-c4ff-43d6-9dcb-61c3e6f552fb
spec:
  hostnames:
  - test.gwapi.ci-ln-xmn87rt-76ef8.aws-2.ci.openshift.org
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: gateway
    namespace: openshift-ingress
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: service-unsecure
      port: 27017
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
status:
  parents:
  - conditions:
    - lastTransitionTime: "2024-12-05T17:40:06Z"
      message: Route was valid, bound to 2 parents
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2024-12-05T17:40:06Z"
      message: All references resolved <--
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True" <--
      type: ResolvedRefs
    controllerName: openshift.io/gateway-controller
    parentRef:
      group: gateway.networking.k8s.io
      kind: Gateway
      name: gateway
      namespace: openshift-ingress
`
Route is bound to the gateway and SMCP is recreated after controller reconciles 
Mark as verified 
/label qe-approved

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -90,15 +123,29 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	var errs []error
 	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
+		log.Error(err, "failed to ensure ServiceMeshOperatorSubscription", "request", request)
 		errs = append(errs, err)
 	}
 	if _, _, err := r.ensureServiceMeshControlPlane(ctx, &gatewayclass); err != nil {
+		log.Error(err, "failed to ensure ServiceMeshControlPlane", "request", request)
 		errs = append(errs, err)
+	} else {
+		r.startSMCPWatch.Do(func() {
```
candita, line 139: Why would we want to start a watch every time we reconcile?

anirudhAgniRedhat: Hey @candita, We are using sync.Once here, start the watch for SMCP, which should run only once. So, this should not run on every reconciliation.

Since we need to create the SMCP watch after the OSSM operator is installed, we cannot start the watch as usual.

anirudhAgniRedhat:
Hello, just a gentle reminder about this PR. It’s been open for a while now. PTAL!
/retest

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -90,15 +123,29 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	var gatewayclass gatewayapiv1.GatewayClass
 	if err := r.cache.Get(ctx, request.NamespacedName, &gatewayclass); err != nil {
+		log.Error(err, "failed to get gatewayclass", "request", request)
 		return reconcile.Result{}, err
```
alebedev87, line 127: Errors returned from `Reconcile` are logged as errors by the controller-runtime. Failure to get a resource will be clear enough straight out of the client.
```suggestion
		return reconcile.Result{}, err
```

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -90,15 +123,29 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
 	var errs []error
 	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
+		log.Error(err, "failed to ensure ServiceMeshOperatorSubscription", "request", request)
 		errs = append(errs, err)
```
alebedev87, line 133: It's enough to details the error, it will be logged automatically by controller-runtime.
```suggestion
		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -289,6 +289,53 @@ func assertSubscription(t *testing.T, namespace, subName string) error {
...
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, nsName, existingSubscription); err != nil {
+			t.Logf("failed to get Subscription %s: %v", nsName.Name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to get Subscription %s: %v", nsName.Name, err)
+		return err
```
alebedev87, line 308: You `Fatalf()` the test if `deleteExistingSubscription` returns error, this will 1) fail the test, 2) log the error:
```suggestion
		return fmt.Errorf("timed out getting subscription %s: %w", nsName.Name, err)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -289,6 +289,53 @@ func assertSubscription(t *testing.T, namespace, subName string) error {
...
+	}
+	// deleting Subscription.
+	err = kclient.Delete(context.Background(), existingSubscription)
+	if err != nil {
+		t.Errorf("failed to delete Subscription %s: %v", nsName.Name, err)
+		return err
```
alebedev87, line 314: To avoid repeating things - the error from the client should say which operation failed and for what reason:
```suggestion
		return err
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -289,6 +289,53 @@ func assertSubscription(t *testing.T, namespace, subName string) error {
...
+	}
+	err = wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, nsName, newSubscription); err != nil {
+			if kerrors.IsNotFound(err) {
+				return true, nil
+			}
+			t.Logf("failed to delete Subscription %s: %v", nsName.Name, err)
```
alebedev87, line 321:
```suggestion
			t.Logf("failed to get deleted subscription %s: %v", nsName.Name, err)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -625,6 +672,52 @@ func assertSMCP(t *testing.T) error {
...
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, nsName, existingSMCP); err != nil {
+			t.Logf("failed to get smcp %s: %v", nsName.Name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("failed to get smcp %s: %v", nsName.Name, err)
+		return err
```
alebedev87, line 691:
```suggestion
		return fmt.Errorf("timed out getting smcp %s: %w", nsName.Name, err)
```

alebedev87 (commented):
The approach seems like a reasonable way out off the dependency on the CRD existence.

pkg/operator/controller/gatewayclass/controller.go
candita: @anirudhAgniRedhat it looks like you discussed this with @rfredette in https://github.com/openshift/cluster-ingress-operator/pull/1115#issuecomment-2278598112 .  Could you please add an attribute in your git commit that Ryan helped with this?  cc @Miciah

anirudhAgniRedhat: Added!!

alebedev87: I think that we should explain the reason why the watch is added after starting the controller. With a comment or in the commit message or both (the best I think).

alebedev87 (commented):
LGTM, just 2 last remarks:
1. @anirudhAgniRedhat: can you please add the bug number to the commit message in the format `<BUG_NUMBER>: <COMMIT_MESSAGE>`?
2. See below.

anirudhAgniRedhat:
@candita @alebedev87 @rfredette Who should be docs POC for this?

PR #1116, opened 2024-08-01
OCPBUGS-36340: Retry IngressController and Route updates in E2E tests

grzpiotrowski:
This PR fixes flakes in e2e tests due to the IngressController or Route objects being updated after the e2e test retrieves a local copy, resulting in an "object has been modified" error.

This is achieved by using the existing helper function `updateIngressControllerWithRetryOnConflict`.

Also a new helper function `updateRouteWithRetryOnConflict` was added to retry updates of Route object in the same way.

Example fail to update IngressController:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1114/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-operator/1818681004299128832

Example fail to update Route:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1023/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1815818916132294656

### DIFF
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -520,6 +520,32 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
 	}
 }
 
+// updateRouteWithRetryOnConflict gets a fresh copy of the named route,
+// calls mutateRouteFn() where callers can modify fields of the route,
+// and then updates the route object. If there is a conflict error
+// on update then the complete sequence of get, mutate, and update
+// is retried until timeout is reached.
+func updateRouteWithRetryOnConflict(t *testing.T, name types.NamespacedName, timeout time.Duration, mutateRouteFn func(route *routev1.Route)) error {
+	t.Helper()
+	route := &routev1.Route{}
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, name, route); err != nil {
+			t.Logf("error getting route %v: %v, retrying...", name, err)
+			return false, nil
+		}
+		mutateRouteFn(route)
+		if err := kclient.Update(ctx, route); err != nil {
+			if errors.IsConflict(err) {
+				t.Logf("conflict when updating route %v: %v, retrying...", name, err)
+				return false, nil
+			}
+			t.Logf("error updating route %v: %v, retrying...", name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
 // updateIngressControllerWithRetryOnConflict gets a fresh copy of
 // the named ingresscontroller, calls mutateIngressControllerFn() where
 // callers can modify fields of the ingresscontroller, and then updates
--- a/pkg/operator/controller/names.go
+++ b/pkg/operator/controller/names.go
@@ -203,7 +203,6 @@ func IngressControllerDeploymentPodSelector(ic *operatorv1.IngressController) *m
 }
 
 func InternalIngressControllerServiceName(ic *operatorv1.IngressController) types.NamespacedName {
-	// TODO: remove hard-coded namespace
 	return types.NamespacedName{Namespace: DefaultOperandNamespace, Name: "router-internal-" + ic.Name}
 }
 
@@ -215,7 +214,11 @@ func IngressControllerServiceMonitorName(ic *operatorv1.IngressController) types
 }
 
 func LoadBalancerServiceName(ic *operatorv1.IngressController) types.NamespacedName {
-	return types.NamespacedName{Namespace: DefaultOperandNamespace, Name: "router-" + ic.Name}
+	return LoadBalancerServiceNameFromICName(ic.Name)
+}
+
+func LoadBalancerServiceNameFromICName(icName string) types.NamespacedName {
+	return types.NamespacedName{Namespace: DefaultOperandNamespace, Name: "router-" + icName}
 }
 
 func NodePortServiceName(ic *operatorv1.IngressController) types.NamespacedName {
--- a/test/e2e/allowed_source_ranges_test.go
+++ b/test/e2e/allowed_source_ranges_test.go
@@ -92,11 +92,9 @@ func TestAllowedSourceRanges(t *testing.T) {
 	// Update the AllowedSourceRanges and verify that LoadBalancerSourceRanges field is updated as well.
 	updatedCIDR := "10.0.0.0/8"
 	t.Log("Updating AllowedSourceRanges")
-	if err := kclient.Get(context.TODO(), name, ic); err != nil {
-		t.Fatalf("failed to get ingresscontroller %s: %v", name, err)
-	}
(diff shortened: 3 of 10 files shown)


test/e2e/util_test.go
```diff
@@ -520,6 +520,30 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
 	}
 }
 
+// updateRouteWithRetryOnConflict gets a fresh copy of the named route,
+// calls mutateRouteFn() where callers can modify fields of the route,
+// and then updates the route object. If there is a conflict error
+// on update then the complete sequence of get, mutate, and update
+// is retried until timeout is reached.
+func updateRouteWithRetryOnConflict(t *testing.T, name types.NamespacedName, timeout time.Duration, mutateRouteFn func(route *routev1.Route)) error {
+	route := routev1.Route{}
```
Miciah, line 529: Please add `t.Helper()` at the start of the helper function.

grzpiotrowski: Done. Thank you

test/e2e/util_test.go
```diff
@@ -520,6 +520,31 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
...
 }
 
+// updateRouteWithRetryOnConflict gets a fresh copy of the named route,
+// calls mutateRouteFn() where callers can modify fields of the route,
+// and then updates the route object. If there is a conflict error
+// on update then the complete sequence of get, mutate, and update
+// is retried until timeout is reached.
+func updateRouteWithRetryOnConflict(t *testing.T, name types.NamespacedName, timeout time.Duration, mutateRouteFn func(route *routev1.Route)) error {
+	t.Helper()
+	route := routev1.Route{}
```
alebedev87, line 530:
```suggestion
	route := &routev1.Route{}
```
Then there will be no need to take its address 3 times down the function.

test/e2e/util_test.go
```diff
@@ -520,6 +520,31 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
...
+	return wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
```
alebedev87, line 531: I know this function is used everywhere but [it's deprecated](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate). We better start using `PollUntilContextTimeout` ([example](https://github.com/openshift/cluster-ingress-operator/blob/76cfd8a48e4dc98dfd5787d4facaacc32616522d/test/e2e/util_test.go#L981)).

alebedev87 (commented):
LGTM, just some nit picks.

test/e2e/util_test.go
frobware: I suppose it's a matter of preference. Personally, I like to avoid using &. Why? Because when you have to take the address multiple times within a function, it's a useful signal to me that the route might be, or could be, mutated—even if it doesn't actually get mutated. Maybe that's just the C programmer in me...

test/e2e/util_test.go
```diff
@@ -520,6 +520,31 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
...
+// updateRouteWithRetryOnConflict gets a fresh copy of the named route,
+// calls mutateRouteFn() where callers can modify fields of the route,
+// and then updates the route object. If there is a conflict error
+// on update then the complete sequence of get, mutate, and update
+// is retried until timeout is reached.
+func updateRouteWithRetryOnConflict(t *testing.T, name types.NamespacedName, timeout time.Duration, mutateRouteFn func(route *routev1.Route)) error {
+	t.Helper()
+	route := &routev1.Route{}
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(context.TODO(), name, route); err != nil {
```
alebedev87, line 532: The context is already passed to the function from the caller (`PollUntilContextTimeout`):
```suggestion
		if err := kclient.Get(ctx, name, route); err != nil {
```

test/e2e/util_test.go
```diff
@@ -520,6 +520,31 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
...
+			t.Logf("error getting route %v: %v, retrying...", name, err)
+			return false, nil
+		}
+		mutateRouteFn(route)
+		if err := kclient.Update(context.TODO(), route); err != nil {
```
alebedev87, line 537:
```suggestion
		if err := kclient.Update(ctx, route); err != nil {
```

test/e2e/util_test.go
```diff
@@ -520,6 +520,31 @@ func probe(timeout, period, success, failure int) *corev1.Probe {
...
+			if errors.IsConflict(err) {
+				t.Logf("conflict when updating route %v: %v, retrying...", name, err)
+				return false, nil
+			}
+			return false, err
```
alebedev87, line 542: This flow is different from the `Get(route)` one. In the get case we log the error and return `false, nil`, while here we don't log and return `false, err`. It would be nice to align these 2 flows. I think that logging the error and returning `false, nil` is preferable here too since we are in the testing utils (logs will be helpful).

test/e2e/util_test.go
grzpiotrowski: Thank you. I opted to keep taking the address in each line there. Keeping it in line with other helpers in test/e2e/util_test.go I suppose. I hope that's alright.

test/e2e/util_test.go
grzpiotrowski: Done. Thank you

test/e2e/util_test.go
grzpiotrowski: :heavy_check_mark:

test/e2e/util_test.go
grzpiotrowski: :heavy_check_mark: 
thank you

test/e2e/util_test.go
alebedev87: I think that taking address once at initialization is more go idiomatic. In the code of this operator, we can this usage is more popular:
```
cluster-ingress-operator$ agi ' &routev1.Route{' -c | cut -d ':' -f2 | sum
26

cluster-ingress-operator$ agi ' routev1.Route{' -c | cut -d ':' -f2 | sum
6

$ which agi
alias agi='ag --ignore-dir=vendor --ignore=tags --ignore=go.sum'
	/usr/bin/ag
```

as well as in some other highly used go projects like kube's client-go, for instance (taken from `vendor` of this repository too):
```
client-go$ ag ' &v1.Pod{' -c | cut -d ':' -f2 | sum
8
client-go$ ag ' &v1.Service{' -c | cut -d ':' -f2 | sum
7

client-go$ ag ' v1.Service{' || echo "not found"
not found
client-go$ ag ' v1.Pod{' || echo "not found"
not found
```

I agree that the pointer can signal a potential modification. However, unfortunately, in golang we never know for sure whether the intention was to save some copying or to modify the object. So the caller have to make sure anyway.

grzpiotrowski: Thank you for different perspectives on this. After further consideration I put this change back in as @alebedev87 suggested.

PR #1117, opened 2024-08-02
e2e: introduce waitForLBAnnotation helper function

alebedev87:


gcs278:
@alebedev87 I support this change. Are you planning to continue to follow up or do you want me or @miheer to take this on?

alebedev87:
@gcs278 : no, it was just an example. Miheer added [waitForLBAnnotation function](https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/util_test.go#L1085). We can migrate other tests to it little by little.

gcs278:
@alebedev87 ah I see now, thanks.

PR #1118, opened 2024-08-05
NE-1798: API bump for promotion of eipAllocation from feature gates to GA.

miheer:
After promoting eipAllocation to GA, there are no longer any featuregated
fields in the IngressController CRD, so now we just have 1 copy of the
file, and so the hack scripts has to be modified.

Following command were run:

    go mod edit -replace=github.com/openshift/api=github.com/openshift/api@6b4a57ec20b0bfc77cb3909da4f1a9234d498571
    go mod tidy
    go mod vendor
    make generate
    make update


hack/update-generated-crd.sh
```diff
@@ -10,23 +10,29 @@ function install_crd {
...
 
-install_crd \
+[[ -e "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-CustomNoUpgrade.crd.yaml" ]] && install_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-CustomNoUpgrade.crd.yaml" \
   "manifests/00-custom-resource-definition-CustomNoUpgrade.yaml"
 
-install_crd \
+[[ -e "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-DevPreviewNoUpgrade.crd.yaml" ]] && install_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-DevPreviewNoUpgrade.crd.yaml" \
   "manifests/00-custom-resource-definition-DevPreviewNoUpgrade.yaml"
```
Miciah, line 26: I know I suggested adding the checks on the `install_crd` commands, but what do you think about putting the check in `install_crd` function itself?
```
function install_crd {
  local SRC="$1"
  local DST="$2"
  if [[ -e "$SRC" ]]
  then
    if [[ -e "$DST" ]]
    then
      if ! diff -Naup "$SRC" "$DST"; then
        cp "$SRC" "$DST"
        echo "updated CRD: $SRC => $DST"
      else
        echo "skipped CRD that is already up to date: $DST"
      fi
    else
      cp "$SRC" "$DST"
      echo "updated CRD: $SRC => $DST"
    fi
  else
    if [[ -e "$DST" ]]
    then
      rm "$DST"
      echo "removed CRD that not vendored: $DST"
    else
      echo "skipped CRD that is not vendored: $SRC"
    fi
  fi
}
```
Then you don't need to change the `install_crd` commands in this PR to repeat the check on each one.

hack/update-generated-crd.sh
```diff
@@ -10,23 +10,29 @@ function install_crd {
   fi
 }
 
+
 # Can't rely on associative arrays for old Bash versions (e.g. OSX)
-install_crd \
+
+[[ -e "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-Default.crd.yaml" ]] && install_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-Default.crd.yaml" \
   "manifests/00-custom-resource-definition.yaml"
```
Miciah, line 18: Does this work?  I think you need something like this to match either `0000_50_ingress_00_ingresscontrollers-Default.crd.yaml` or `0000_50_ingress_00_ingresscontrollers.crd.yaml`:
```suggestion
shopt -s extglob
verify_crd \
  vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers?(-Default).crd.yaml \
  "manifests/00-custom-resource-definition.yaml"
```

hack/verify-generated-crd.sh
```diff
@@ -10,22 +10,27 @@ function verify_crd {
...
-verify_crd \
+
+[[ -e "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-Default.crd.yaml" ]] && verify_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-Default.crd.yaml" \
   "manifests/00-custom-resource-definition.yaml"
 
-verify_crd \
+[[ -e "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-CustomNoUpgrade.crd.yaml" ]] && verify_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-CustomNoUpgrade.crd.yaml" \
   "manifests/00-custom-resource-definition-CustomNoUpgrade.yaml"
```
Miciah, line 20: Again, what do you think about changing `verify_crd` instead, as follows?
```bash
function verify_crd {
  local SRC="$1"
  local DST="$2"
  if [[ -e "$SRC" ]]
  then
    if [[ -e "$DST" ]]
    then
      if ! diff -Naup "$SRC" "$DST"; then
        echo "inconsistent CRD: $SRC => $DST"
      fi
    else
      echo "missing CRD: $SRC => $DST"
      exit 1
    fi
  else
    if [[ -e "$DST" ]]
    then
      echo "extra CRD: $DST"
      exit 1
    fi
  fi
}
```

Miciah (commented):
Please follow conventions for the commit message:
```text
Bump openshift/api to promote eipAllocation to GA

After promoting eipAllocation to GA, there are no longer any featuregated
fields in the IngressController CRD, so now we just have 1 copy of the
file, and so the hack scripts has to be modified.

Following command were run:

    go mod edit -replace=github.com/openshift/api=github.com/openshift/api@6b4a57ec20b0bfc77cb3909da4f1a9234d498571
    go mod tidy
    go mod vendor
    make generate
    make update
```

hack/update-generated-crd.sh
miheer: done.

hack/update-generated-crd.sh
miheer: yes that is a nice way to do pattern match file.

miheer: done

hack/verify-generated-crd.sh
miheer: yes looks good thanks!

miheer: done

miheer:
@Miciah PTAL thanks!

hack/update-generated-crd.sh
```diff
@@ -27,6 +47,10 @@ install_crd \
   "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers-TechPreviewNoUpgrade.crd.yaml" \
   "manifests/00-custom-resource-definition-TechPreviewNoUpgrade.yaml"
 
+install_crd \
+  "vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers.crd.yaml" \
+  "manifests/00-custom-resource-definition.yaml"
```
Miciah, line 52: You made this change for the verify script, but you should also make it for the update script:
```suggestion
shopt -s extglob
install_crd \
  vendor/github.com/openshift/api/operator/v1/zz_generated.crd-manifests/0000_50_ingress_00_ingresscontrollers?(-Default).crd.yaml \
  "manifests/00-custom-resource-definition.yaml"
```

miheer: @Miciah sorry yes done and pushed the changes.

Miciah:
Thanks!
/lgtm

PR #1126, opened 2024-08-13
OCPBUGS-38217: Clear LB Status Parameters on LB Type Change

gcs278:
When the LB type is changed, the status parameters for the previously selected type should be cleared. This fix ensures that the status remains consistent with the currently selected LB type.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -548,6 +548,7 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					changed = true
 				}
 				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSClassicLoadBalancer {
+					statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters = nil
 					if statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters == nil {
 						statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters = &operatorv1.AWSClassicLoadBalancerParameters{}
 					}
@@ -569,6 +570,7 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					}
 				}
 				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+					statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters = nil
 					if statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters == nil {
 						statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters = &operatorv1.AWSNetworkLoadBalancerParameters{}
 					}
--- a/pkg/operator/controller/ingress/controller_test.go
+++ b/pkg/operator/controller/ingress/controller_test.go
@@ -625,6 +625,20 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
 			domainMatchesBaseDomain: true,
 		},
 		{
+			name:                    "loadbalancer type changed from ELB to NLB, with old ELB parameters removal",
+			ic:                      makeIC(spec(nlb()), status(elbWithNullParameters())),
+			expectedResult:          true,
+			expectedIC:              makeIC(spec(nlb()), status(nlbWithNullParameters())),
+			domainMatchesBaseDomain: true,
+		},
+		{
+			name:                    "loadbalancer type changed from NLB to ELB, with old NLB parameters removal",
+			ic:                      makeIC(spec(elb()), status(nlbWithNullParameters())),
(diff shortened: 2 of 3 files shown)


gcs278:
CI issues with AWS DNS that is currently under investigation:
/retest

gcs278:
CI is in pretty rough shape with DNS failures, need to wait on https://github.com/openshift/cluster-ingress-operator/pull/1127

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1127 has merged
/retest

gcs278:
preemptive hold for multiple reviewers
/hold

pkg/operator/controller/ingress/controller.go
```diff
@@ -548,6 +548,7 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					changed = true
 				}
 				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSClassicLoadBalancer {
+					statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters = nil
```
miheer, line 551: statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters   does not the .NetworkLoadBalancerParameters give a panic error for nil  as it will access NetworkLoadBalancerParameters which does not exist?

Or may be we need to add a condition to check statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters for nil.

pkg/operator/controller/ingress/controller.go
```diff
@@ -569,6 +570,7 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					}
 				}
 				if statusLB.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer {
+					statusLB.ProviderParameters.AWS.ClassicLoadBalancerParameters = nil
```
miheer, line 573: same as above.

miheer (changes requested):

pkg/operator/controller/ingress/controller.go
gcs278: No need to worry about a nil pointer dereference when assigning nil to `statusLB.ProviderParameters.AWS.NetworkLoadBalancerParameters`. The [logic](https://github.com/openshift/cluster-ingress-operator/blob/7edef371026129f765a86748148c79b38fba636e/pkg/operator/controller/ingress/controller.go#L543-L545) above this PR's changes already ensures that `statusLB.ProviderParameters.AWS` is non-nil, so this assignment is safe. Assigning nil to this field is simply setting a pointer to nil (doesn't matter if it's already nil) and doesn’t cause a panic as long as the parent structure exists.

gcs278:
hold for @alebedev87 incase he wants to review
/hold

pkg/operator/controller/ingress/controller.go
alebedev87, line 551: I see that the return value of `setDefaultPublishingStrategy` is not really taken into account by the calling function (why btw) but at least the unit test does. 1. should we set `changed=true` taken into account that we reset the status? 2. I think that we should add unit tests to see the status gets reset.

gcs278: Yea I find this pattern a bit odd. `changed` is only used in unit testing. The really odd thing is: why do you even need a `changed` bool for unit testing? Our unit test `TestSetDefaultPublishingStrategyHandlesUpdates` diffs the expected IngressController with the actual, so we are measuring the "change" that way. Seems much better to detect the change through the actual diff, and not have an artificial `change` flag, which a bit meaningless, prone to error, and arbitrary (I don't think it adds any value).

I think the pattern with `changed` the original author was trying to follow was similar to [recordIngressCondition](https://github.com/openshift/router/blob/fa20a2437f013dfc51d360e3f50b43df1db590bc/pkg/router/controller/status.go#L197), where `changed` is actually used to indicate if an update is needed. However, that isn't the case with `setDefaultPublishingStrategy`, as we always update the status in `admit`.

> 1. should we set changed=true taken into account that we reset the status?

Given that:
1. These [initialization-type](https://github.com/openshift/cluster-ingress-operator/blob/7edef371026129f765a86748148c79b38fba636e/pkg/operator/controller/ingress/controller.go#L537-L545) changes don't set `change=true` when the status is actually changed (I think this PR's status "clean up" is similar to these status "initialization")
2. My point above about `changed` having little to no value and it appears like it's not being maintained properly (shouldn't `changed` be true when anything in the status changes?)

I vote to not worry about it.

> 2. I think that we should add unit tests to see the status gets reset.

Ah I completely forgot about unit testing. Done.

gcs278:
TestRouteMetricsControllerOnlyRouteSelector failed, but fix is coming in https://github.com/openshift/cluster-ingress-operator/pull/1116
/test e2e-gcp-operator

Odd deprecated API failure:
/test e2e-aws-ovn-serial

PR #1130, opened 2024-08-19
OCPBUGS-38411: Not allow eipAllocations in CLB while changing the LBType of the Ingress Controller from NLB to Classic.

miheer:
After changing the LBType of the IngressController from `NLB` to `Classic`, the eipAllocations still were attached with controller and the LB service. Howver, eipAllocations are  not supported in Classic.

This commit fixes service annotation `service.beta.kubernetes.io/aws-load-balancer-eip-allocations` value to a blank value. The CCM then removes this annotation from the service. It also sets the IngressController in Progressing Status to True with a status message to delete the load balancer service so that a new lb service is recreated for Classic LB.


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -441,6 +441,13 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 								service.Annotations[awsLBSubnetsAnnotation] = JoinAWSSubnets(clbParams.Subnets, ",")
 							}
 						}
+
+						if eipAllocationsAWSEnabled {
+							nlbParams := getAWSNetworkLoadBalancerParametersInSpec(ci)
+							if nlbParams != nil && awsEIPAllocationsExist(nlbParams.EIPAllocations) {
+								service.Annotations[awsEIPAllocationsAnnotation] = ""
+							}
+						}
```
gcs278, line 450: I don't think this do what you want, the annotation isn't set by default anyways.

And this adds the annotation with a value of `""`, it's not clearing it. You should be able to just remove this.

miheer: yes. i was already working on it. thanks for pointing out !

gcs278:
DNS Propagation bug should be fixed, retesting gatewayapi:
/test e2e-aws-gatewayapi

gcs278:
I won't be able to review this for the next couple of months due to other commitments.
/unassign

candita:
To be followed up and reassigned later.
/hold

PR #1132, opened 2024-08-20
WIP DO NOT MERGE: DNS Propagation Test

gcs278:
PR to test DNS propagation in CI. DO NOT MERGE

PR #1133, opened 2024-08-22
OCPBUGS-32776: Fix IBM Public Cloud DNS Provider Update Logic

gcs278:
The IBM Public Cloud DNS provider (`cis_provider.go`) had a bug in `createOrUpdateDNSRecord` where it checked for the existence of a DNS record by filtering both DNS name and target. If the target was updated (e.g., due to a load balancer recreation), the logic would not match the existing DNS record. As a result, the function would attempt to create a new record, but fail because a record with that name already existed, as multiple DNS records with the same
name are not allowed in IBM Cloud DNS providers.

The fix is to remove the filtering by target and rely solely on filtering by name, as the name is the only attribute that needs to be unique.

This fix also includes a subtle fix in `createOrUpdateDNSRecord` of cis_provider.go to fall through to the create logic in the case of a not-found (404) error.

Additionally, the IBM DNS logic doesn't work for multiple targets and this creates unexpected and problematic results. The logic has been refactored to only create using the first target and it warns the user when multiple targets are set. This change is low risk since the Ingress Operator will never create a DNSRecord with multiple targets in `desiredDNSRecord`.

Additionally, modify TestScopeChange to check for `Available=True` after deleting the service to add test coverage for this issue.

This PR also includes some unit test fix up and missing unit test coverage for the IBM CIS Provider.

This resolves the same DNS issues for public PowerVS cloud as it uses the same logic.

### DIFF
--- a/pkg/dns/ibm/private/dnssvcs_provider.go
+++ b/pkg/dns/ibm/private/dnssvcs_provider.go
@@ -103,6 +103,9 @@ func (p *Provider) Delete(record *iov1.DNSRecord, zone configv1.DNSZone) error {
 
 	for _, resourceRecord := range result.ResourceRecords {
 		var resourceRecordTarget string
+		if resourceRecord.ID == nil {
+			return fmt.Errorf("delete: record id is nil")
+		}
 		rData, ok := resourceRecord.Rdata.(map[string]interface{})
 		if !ok {
 			return fmt.Errorf("delete: failed to get resource data: %v", resourceRecord.Rdata)
@@ -126,7 +129,8 @@ func (p *Provider) Delete(record *iov1.DNSRecord, zone configv1.DNSZone) error {
 		default:
 			return fmt.Errorf("delete: resource data has record with unknown type: %v", *resourceRecord.Type)
 		}
-
+		// While creating DNS records with multiple targets is unsupported, we still
+		// iterate through all targets during deletion to be extra cautious.
 		for _, target := range record.Spec.Targets {
 			if *resourceRecord.Name == dnsName {
 				if resourceRecordTarget != target {
@@ -190,9 +194,14 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 		log.Info("Warning: TTL must be one of [1 60 120 300 600 900 1800 3600 7200 18000 43200]. RecordTTL set to default", "default DSNSVCS record TTL", defaultDNSSVCSRecordTTL)
 		record.Spec.RecordTTL = defaultDNSSVCSRecordTTL
 	}
+	if len(record.Spec.Targets) > 1 {
+		return fmt.Errorf("createOrUpdateDNSRecord: only one target is supported, but found %d: targets=%v", len(record.Spec.Targets), record.Spec.Targets)
+	}
 
 	listResult, response, err := p.dnsService.ListResourceRecords(listOpt)
 	if err != nil {
+		// Avoid continuing with an invalid list response, as we can't determine
+		// whether to create or update the DNS record, which may lead to further issues.
 		if response == nil || response.StatusCode != http.StatusNotFound {
 			return fmt.Errorf("createOrUpdateDNSRecord: failed to list the dns record: %w", err)
 		}
@@ -201,72 +210,74 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 		return fmt.Errorf("createOrUpdateDNSRecord: ListResourceRecords returned nil as result")
 	}
 
-	for _, target := range record.Spec.Targets {
-		updated := false
-		for _, resourceRecord := range listResult.ResourceRecords {
-			if *resourceRecord.Name == dnsName {
-				updateOpt := p.dnsService.NewUpdateResourceRecordOptions(p.config.InstanceID, zone.ID, *resourceRecord.ID)
-				updateOpt.SetName(dnsName)
-
-				if resourceRecord.Type == nil {
-					return fmt.Errorf("createOrUpdateDNSRecord: failed to get resource type, resourceRecord.Type is nil")
-				}
-
-				// TODO DNS record update should handle the case where we have an A record and want a CNAME record or vice versa
-				switch *resourceRecord.Type {
-				case string(iov1.CNAMERecordType):
-					inputRData, err := p.dnsService.NewResourceRecordUpdateInputRdataRdataCnameRecord(target)
-					if err != nil {
-						return fmt.Errorf("createOrUpdateDNSRecord: failed to create CNAME inputRData for the dns record: %w", err)
-					}
-					updateOpt.SetRdata(inputRData)
-				case string(iov1.ARecordType):
-					inputRData, err := p.dnsService.NewResourceRecordUpdateInputRdataRdataARecord(target)
-					if err != nil {
-						return fmt.Errorf("createOrUpdateDNSRecord: failed to create A inputRData for the dns record: %w", err)
-					}
-					updateOpt.SetRdata(inputRData)
-				default:
-					return fmt.Errorf("createOrUpdateDNSRecord: resource data has record with unknown type: %v", *resourceRecord.Type)
-				}
-				updateOpt.SetTTL(record.Spec.RecordTTL)
-				_, _, err := p.dnsService.UpdateResourceRecord(updateOpt)
-				if err != nil {
-					return fmt.Errorf("createOrUpdateDNSRecord: failed to update the dns record: %w", err)
-				}
-				updated = true
-				log.Info("updated DNS record", "record", record.Spec, "zone", zone, "target", target)
+	target := record.Spec.Targets[0]
+	updated := false
+	for _, resourceRecord := range listResult.ResourceRecords {
+		if *resourceRecord.Name == dnsName {
+			if resourceRecord.ID == nil {
+				return fmt.Errorf("createOrUpdateDNSRecord: record id is nil")
 			}
-		}
-		if !updated {
-			createOpt := p.dnsService.NewCreateResourceRecordOptions(p.config.InstanceID, zone.ID)
-			createOpt.SetName(dnsName)
-			createOpt.SetType(string(record.Spec.RecordType))
+			updateOpt := p.dnsService.NewUpdateResourceRecordOptions(p.config.InstanceID, zone.ID, *resourceRecord.ID)
+			updateOpt.SetName(dnsName)
 
-			switch record.Spec.RecordType {
-			case iov1.CNAMERecordType:
-				inputRData, err := p.dnsService.NewResourceRecordInputRdataRdataCnameRecord(target)
+			if resourceRecord.Type == nil {
+				return fmt.Errorf("createOrUpdateDNSRecord: failed to get resource type, resourceRecord.Type is nil")
+			}
+
+			// TODO DNS record update should handle the case where we have an A record and want a CNAME record or vice versa
+			switch *resourceRecord.Type {
+			case string(iov1.CNAMERecordType):
+				inputRData, err := p.dnsService.NewResourceRecordUpdateInputRdataRdataCnameRecord(target)
 				if err != nil {
 					return fmt.Errorf("createOrUpdateDNSRecord: failed to create CNAME inputRData for the dns record: %w", err)
 				}
-				createOpt.SetRdata(inputRData)
-			case iov1.ARecordType:
-				inputRData, err := p.dnsService.NewResourceRecordInputRdataRdataARecord(target)
+				updateOpt.SetRdata(inputRData)
+			case string(iov1.ARecordType):
+				inputRData, err := p.dnsService.NewResourceRecordUpdateInputRdataRdataARecord(target)
 				if err != nil {
 					return fmt.Errorf("createOrUpdateDNSRecord: failed to create A inputRData for the dns record: %w", err)
 				}
-				createOpt.SetRdata(inputRData)
+				updateOpt.SetRdata(inputRData)
 			default:
-				return fmt.Errorf("createOrUpdateDNSRecord: resource data has record with unknown type: %v", record.Spec.RecordType)
-
+				return fmt.Errorf("createOrUpdateDNSRecord: resource data has record with unknown type: %v", *resourceRecord.Type)
 			}
-			createOpt.SetTTL(record.Spec.RecordTTL)
-			_, _, err := p.dnsService.CreateResourceRecord(createOpt)
+			updateOpt.SetTTL(record.Spec.RecordTTL)
+			_, _, err := p.dnsService.UpdateResourceRecord(updateOpt)
 			if err != nil {
-				return fmt.Errorf("createOrUpdateDNSRecord: failed to create the dns record: %w", err)
+				return fmt.Errorf("createOrUpdateDNSRecord: failed to update the dns record: %w", err)
 			}
-			log.Info("created DNS record", "record", record.Spec, "zone", zone, "target", target)
+			updated = true
+			log.Info("updated DNS record", "record", record.Spec, "zone", zone, "target", target)
 		}
+	}
+	if !updated {
+		createOpt := p.dnsService.NewCreateResourceRecordOptions(p.config.InstanceID, zone.ID)
+		createOpt.SetName(dnsName)
+		createOpt.SetType(string(record.Spec.RecordType))
+
+		switch record.Spec.RecordType {
+		case iov1.CNAMERecordType:
+			inputRData, err := p.dnsService.NewResourceRecordInputRdataRdataCnameRecord(target)
+			if err != nil {
+				return fmt.Errorf("createOrUpdateDNSRecord: failed to create CNAME inputRData for the dns record: %w", err)
+			}
+			createOpt.SetRdata(inputRData)
+		case iov1.ARecordType:
+			inputRData, err := p.dnsService.NewResourceRecordInputRdataRdataARecord(target)
+			if err != nil {
+				return fmt.Errorf("createOrUpdateDNSRecord: failed to create A inputRData for the dns record: %w", err)
+			}
+			createOpt.SetRdata(inputRData)
+		default:
+			return fmt.Errorf("createOrUpdateDNSRecord: resource data has record with unknown type: %v", record.Spec.RecordType)
+
+		}
+		createOpt.SetTTL(record.Spec.RecordTTL)
+		_, _, err := p.dnsService.CreateResourceRecord(createOpt)
+		if err != nil {
+			return fmt.Errorf("createOrUpdateDNSRecord: failed to create the dns record: %w", err)
+		}
+		log.Info("created DNS record", "record", record.Spec, "zone", zone, "target", target)
 	}
 	return nil
 }
--- a/pkg/dns/ibm/private/dnssvcs_provider_test.go
+++ b/pkg/dns/ibm/private/dnssvcs_provider_test.go
@@ -65,18 +65,12 @@ func Test_Delete(t *testing.T) {
 			},
 		},
 		{
-			desc:         "listFailError",
-			recordedCall: "DELETE",
-			DNSName:      "testDelete",
-			target:       "11.22.33.44",
+			desc:    "listFailError",
+			DNSName: "testDelete",
+			target:  "11.22.33.44",
 			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
 				OutputError:      errors.New("error in ListAllDnsRecords"),
 				OutputStatusCode: http.StatusRequestTimeout,
-			},
-			deleteDnsRecordInputOutput: dnsclient.DeleteDnsRecordInputOutput{
-				InputId:          "testDelete",
-				OutputError:      nil,
-				OutputStatusCode: http.StatusOK,
 			},
 			expectErrorContains: "error in ListAllDnsRecords",
 		},
@@ -121,6 +115,7 @@ func Test_Delete(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
+			dnsService.ClearCallHistory()
 
 			record := iov1.DNSRecord{
 				Spec: iov1.DNSRecordSpec{
@@ -209,10 +204,9 @@ func Test_createOrUpdateDNSRecord(t *testing.T) {
 			},
 		},
 		{
-			desc:         "listFailError",
-			DNSName:      "testUpdate",
-			target:       "11.22.33.44",
-			recordedCall: "PUT",
+			desc:    "listFailError",
+			DNSName: "testUpdate",
+			target:  "11.22.33.44",
 			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
 				OutputError:      errors.New("error in ListAllDnsRecords"),
 				OutputStatusCode: http.StatusRequestTimeout,
@@ -245,6 +239,7 @@ func Test_createOrUpdateDNSRecord(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
+			dnsService.ClearCallHistory()
 
 			record := iov1.DNSRecord{
 				Spec: iov1.DNSRecordSpec{
--- a/pkg/dns/ibm/private/dnssvcs_provider_test.go
+++ b/pkg/dns/ibm/private/dnssvcs_provider_test.go
@@ -6,10 +6,16 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/IBM/go-sdk-core/v5/core"
+	"github.com/IBM/networking-go-sdk/dnssvcsv1"
+
 	configv1 "github.com/openshift/api/config/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 	dnsclient "github.com/openshift/cluster-ingress-operator/pkg/dns/ibm/private/client"
+
 	"github.com/stretchr/testify/assert"
+
+	"k8s.io/utils/pointer"
 )
 
 func Test_Delete(t *testing.T) {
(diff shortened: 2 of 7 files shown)


gcs278:
I missed one detail of the bug, missing instructions in the `Progressing` status condition for PowerVS when you change scope. Looking into it.
/hold

gcs278:
@lihongan Added the missing scope change instructions for the PowerVS type.

/unhold

gcs278:
infra failures
/retest

gcs278:
Infra issues
/retest

lihongan:
pre-merge tested on OpenStack and looks good now
```
$ oc -n openshift-ingress-operator patch ingresscontroller/intlb --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"type":"LoadBalancerService", "loadBalancer": {"scope":"External"}}}}'
ingresscontroller.operator.openshift.io/intlb patched

$ oc get co/ingress
NAME      VERSION                                                   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.17.0-0.ci.test-2024-08-28-040831-ci-ln-13rxbgt-latest   True        True          False      23m     ingresscontroller "intlb" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: LoadBalancerProgressing=True (OperandsProgressing: One or more managed resources are progressing: The IngressController scope was changed from "Internal" to "External".  To effectuate this change, you must delete the service: `oc -n openshift-ingress delete svc/router-intlb`; the service load-balancer will then be deprovisioned and a new one created.  This will most likely cause the new load-balancer to have a different host name and IP address from the old one's.  Alternatively, you can revert the change to the IngressController: `oc -n openshift-ingress-operator patch ingresscontrollers/intlb --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"loadBalancer":{"scope":"Internal"}}}}'`).

$ oc -n openshift-ingress delete svc/router-intlb
service "router-intlb" deleted

$ oc -n openshift-ingress get svc/router-intlb
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)                      AGE
router-intlb   LoadBalancer   172.30.110.89   38.102.83.53   80:32114/TCP,443:30834/TCP   4m23s
```
will test on IBMCloud later

lihongan:
pre-merge tested on IBMCloud and also looks good
```
$ oc -n openshift-ingress get svc/router-intlb
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP                         PORT(S)                      AGE
router-intlb   LoadBalancer   172.30.111.21   53084b0d-eu-de.lb.appdomain.cloud   80:31026/TCP,443:32398/TCP   78s

$ oc -n openshift-ingress-operator patch ingresscontroller/intlb --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"type":"LoadBalancerService", "loadBalancer": {"scope":"External"}}}}'
ingresscontroller.operator.openshift.io/intlb patched

$ oc -n openshift-ingress delete svc/router-intlb
service "router-intlb" deleted

$ oc -n openshift-ingress get svc/router-intlb
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP                         PORT(S)                      AGE
router-intlb   LoadBalancer   172.30.242.68   72736508-eu-de.lb.appdomain.cloud   80:32041/TCP,443:30579/TCP   65s

$ oc -n openshift-ingress-operator get ingresscontroller/intlb -oyaml
<......>
  - lastTransitionTime: "2024-08-28T09:24:02Z"
    reason: NoFailedZones
    status: "True"
    type: DNSReady
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -794,7 +794,7 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
 	if wantScope != haveScope {
 		err := fmt.Errorf("The IngressController scope was changed from %q to %q.", haveScope, wantScope)
 		switch platform.Type {
-		case configv1.AWSPlatformType, configv1.IBMCloudPlatformType:
+		case configv1.AWSPlatformType, configv1.IBMCloudPlatformType, configv1.PowerVSPlatformType, configv1.AlibabaCloudPlatformType, configv1.OpenStackPlatformType:
```
Miciah, line 797: Can this logic use `platformsWithMutableScope` so that the logic can never become inconsistent?

gcs278: Ah yes good idea. Let me look into that.

Miciah:
@SzucsAti, this is a follow-up to #796.  Are you available to review the changes to the IBM DNS provider?

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Done.

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -161,7 +161,6 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 	dnsName := strings.TrimSuffix(record.Spec.DNSName, ".")
 	listOpt.SetName(dnsName)
 	for _, target := range record.Spec.Targets {
-		listOpt.SetContent(target)
 		result, response, err := dnsService.ListAllDnsRecords(listOpt)
```
candita, line 164: Did you consider, if the target never changes, do we still need to loop on targets?  We definitely only need to make the call to `ListAllDnsRecords` once, outside the loop.

candita:
Please squash commits where possible.

gcs278:
> Please squash commits where possible.

Ack, but I try to make my commits "one concern" or "one thing" as it makes reviews easier, it encourages specific and descriptive commit messages, it allows for more specific code reverts, and enables selective (custom) backports.

But I do agree with your expressed feeling that this PR is overloaded. How about I work with @lihongan to split up OCPBUGS-32776 into a "missing Progressing condition instructions" bug and a "IBM Public DNS Provider fix" bug, so we can have two separate PRs with these two separate concerns? I'm happy to do that to keep things cleaner, and that might help keeping backports clean.

They after all, really are separate and mostly unrelated concerns, I was just trying to conform my fixes to the current concern of the bug.

gcs278:
/hold

I have some work to do per @candita [comment](https://github.com/openshift/cluster-ingress-operator/pull/1133#discussion_r1735181050), and I'd like to split this PR up, that seems like the right thing to do.

gcs278:
Scope Change Instructions bug has been split into https://issues.redhat.com/browse/OCPBUGS-39151 and https://github.com/openshift/cluster-ingress-operator/pull/1135

Still working on addressing https://github.com/openshift/cluster-ingress-operator/pull/1133#discussion_r1735181050.

lihongan:
/label qe-approved
pre-merged tested and passed
```
$ oc -n openshift-ingress-operator patch ingresscontroller/intlb --type=merge --patch='{"spec":{"endpointPublishingStrategy": {"type":"LoadBalancerService", "loadBalancer": {"scope":"External"}}}}'
ingresscontroller.operator.openshift.io/intlb patched

$ oc -n openshift-ingress delete svc/router-intlb
service "router-intlb" deleted

$ oc -n openshift-ingress-operator get ingresscontroller/intlb -oyaml
  - lastTransitionTime: "2024-09-04T05:35:16Z"
    reason: NoFailedZones
    status: "True"
    type: DNSReady

$ oc get co/ingress
NAME      VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.18.0-0.test-2024-09-04-030153-ci-ln-41w8p6k-latest   True        False         False      94m     
```

gcs278:
Sorry, I should mark WIP for now. Uncovering more unit test fixes that are needed as a result of recent updates.
/wip

gcs278:
infra failures
/retest

gcs278:
Removed WIP, sorry for the delay, I am open to reviews again @candita @Miciah @SzucsAti 

**Updates since last review:**
- **I added more unit test coverage.**
  - I felt it necessary to fix existing bugs in the unit tests and add more coverage as it provides confidence that I didn't perturb the DNS provider logic in an unintentional way.
  - The unit tests being in separate commits is intentional and provides a valuable diff before & after the *actual* bug fix here. You can review the last commit and see there were a few required unit test updates which are results of the bug fixes I added. 
- **I removed *create/update* support for multiple targets in DNSRecords for IBM Cloud.**
  - First, the code was misleading because IBM DNS doesn't support it, and therefore it doesn't work (it hot loops and overwrites records in IBM private cloud).
  - Second, since the bug fix actually impacts the way that multi-targeted DNSRecords would get created in IBM Public Cloud, it seemed logical to fix this right now, otherwise, we'd create a future headache for maintaining compatibility (if we were ever concerned about compatibility here)

gcs278:
Keeping the hold while I wait to see what the results of adding a IBM Cloud job are https://github.com/openshift/release/pull/56785

/hold

gcs278:
Now that https://github.com/openshift/release/pull/56785 has merged, though it will most likely fail due to DNS Propagation issues (being tested in https://github.com/openshift/cluster-ingress-operator/pull/1132), I'm curious to see if we see anything interesting:
/test e2e-ibmcloud-operator

gcs278:
I don't think it's enabled, but just curious:
/testwith e2e-ibmcloud-operator https://github.com/openshift/cluster-ingress-operator/pull/1164

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-ibmcloud-operator openshift/cluster-ingress-operator#1164

gcs278:
After introducing and running e2e-ibmcloud-operator on different PRs, I realized it was passing successfully without this fix. I was mistaken that our E2E would have caught this bug. However, it looks like adding a `Availabile=True` condition check to `TestScopeChange` after deleting the service ensures test coverage for bugs like this. Updated this PR to add that.

Now let's test to make sure everything works:
/testwith openshift/cluster-ingress-operator/master/e2e-ibmcloud-operator https://github.com/openshift/cluster-ingress-operator/pull/1164

gcs278:
Forgot to push...
/testwith openshift/cluster-ingress-operator/master/e2e-ibmcloud-operator https://github.com/openshift/cluster-ingress-operator/pull/1164

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-ibmcloud-operator openshift/cluster-ingress-operator/pull#1164

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-ibmcloud-operator https://github.com/openshift/cluster-ingress-operator#1164

gcs278:
[multi-pr-openshift-cluster-ingress-operator-1133-openshift-cluster-ingress-operator-1164](https://github.com/openshift/cluster-ingress-operator/pull/1133/checks?check_run_id=32837360017) passed which has E2E test coverage for this bug, proving it resolved the problem.

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -160,41 +166,43 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
...
-		result, response, err := dnsService.ListAllDnsRecords(listOpt)
+
+	result, response, err := dnsService.ListAllDnsRecords(listOpt)
+	if err != nil {
+		if response == nil || response.StatusCode != http.StatusNotFound {
+			return fmt.Errorf("createOrUpdateDNSRecord: failed to list the dns record: %w", err)
+		}
+	}
+	if result == nil || result.Result == nil {
+		return fmt.Errorf("createOrUpdateDNSRecord: ListAllDnsRecords returned nil as result")
```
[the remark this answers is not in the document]
gcs278: Hmm, that's a good question. I found that `result` and `result.Result` are both initialized when creating the DNS record for the first time. The results gets filtered for the `record.Spec.DNSName`, so whether it's the very first DNS record in the hosted zone isn't a factor.

I agree it's a bit odd to check for `result.Result`, but the original author wrote this logic, I haven't update it as works as intended so far, and it's possible there is a subtle reason why it's here.

gcs278 (commented):
@candita thanks for the detailed review. Should be ready for your review again.

candita:
@gcs278 I didn't re-review this because it was on hold.  I assume that was an oversight.
/cancel hold

gcs278:
> @gcs278 I didn't re-review this because it was on hold. I assume that was an oversight. /cancel hold

@candita Yea I forgot to take the hold off this time, though I often will put holds to ensure every gets a chance to review, or I'm waiting for final CI results, but still want code reviews. I could be wrong. I use WIP as an in indicator it's not ready for review.

/hold cancel

pkg/dns/ibm/private/dnssvcs_provider.go
```diff
@@ -190,9 +194,15 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 		log.Info("Warning: TTL must be one of [1 60 120 300 600 900 1800 3600 7200 18000 43200]. RecordTTL set to default", "default DSNSVCS record TTL", defaultDNSSVCSRecordTTL)
 		record.Spec.RecordTTL = defaultDNSSVCSRecordTTL
 	}
+	// We only support one target, warn the user.
+	if len(record.Spec.Targets) > 1 {
```
candita, line 198: Why not add this to [ValidateInputDNSData](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/dns/ibm/common.go#L88-L90)?  Shouldn't this error stop the addition of the record, rather than just logging it?

pkg/dns/ibm/private/dnssvcs_provider.go
```diff
@@ -201,72 +209,74 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
...
-
-				if resourceRecord.Type == nil {
-					return fmt.Errorf("createOrUpdateDNSRecord: failed to get resource type, resourceRecord.Type is nil")
-				}
+	target := record.Spec.Targets[0]
+	updated := false
+	for _, resourceRecord := range listResult.ResourceRecords {
+		if *resourceRecord.Name == dnsName {
+			if resourceRecord.ID == nil {
+				return fmt.Errorf("createOrUpdateDNSRecord: record id is nil")
```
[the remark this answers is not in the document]
candita: But why wouldn't we want to continue iterating over the `listResult.ResourceRecords` to find a match to the `dnsName`?

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -152,6 +154,10 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 		log.Info("Warning: TTL must be between 120 and 2,147,483,647 seconds, or 1 for Automatic. RecordTTL set to default", "default CIS record TTL", defaultCISRecordTTL)
 		record.Spec.RecordTTL = defaultCISRecordTTL
 	}
+	// We only support one target, warn the user.
```
candita, line 157: Same comment as https://github.com/openshift/cluster-ingress-operator/pull/1133/files#r1919281693

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -160,41 +166,45 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
 	// "." when it creates a wildcard DNS record.
 	dnsName := strings.TrimSuffix(record.Spec.DNSName, ".")
 	listOpt.SetName(dnsName)
-	for _, target := range record.Spec.Targets {
-		listOpt.SetContent(target)
```
candita, line 164: Wow, how did anything work with this here?  But do we want to add just one target to the listOpt, or is that the issue you're fixing - we don't want to match on name + target?

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -160,41 +166,43 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
...
-		result, response, err := dnsService.ListAllDnsRecords(listOpt)
+
+	result, response, err := dnsService.ListAllDnsRecords(listOpt)
+	if err != nil {
+		if response == nil || response.StatusCode != http.StatusNotFound {
```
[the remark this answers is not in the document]
candita: I meant, if response is nil, wouldn't that be a signal to create the record?

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -160,41 +166,45 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
...
-				return fmt.Errorf("createOrUpdateDNSRecord: failed to list the dns record: %w", err)
-			}
-			continue
+			return fmt.Errorf("createOrUpdateDNSRecord: failed to create the dns record: %w", err)
 		}
-		if result == nil || result.Result == nil {
-			return fmt.Errorf("createOrUpdateDNSRecord: ListAllDnsRecords returned nil as result")
+		log.Info("created DNS record", "record", record.Spec, "zone", zone, "target", target)
+	} else {
+		if result.Result[0].ID == nil {
```
candita, line 195: If it's possible for `len(result.Result) > 1`, then you'd need to iterate over the results, wouldn't you?

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +42,133 @@ func Test_Delete(t *testing.T) {
...
-				OutputError:      nil,
-				OutputStatusCode: http.StatusOK,
+				OutputError:    errors.New("error in ListAllDnsRecords"),
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusRequestTimeout},
 			},
 			expectErrorContains: "error in ListAllDnsRecords",
 		},
 		{
-			desc:         "deleteRecordNotFound",
+			desc:         "delete failure with 404",
```
candita, line 86:
```suggestion
			desc:         "delete failure with 404 (record disappears)",
```

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +42,133 @@ func Test_Delete(t *testing.T) {
...
+				OutputResult: &dnsrecordsv1.ListDnsrecordsResp{
+					Result: []dnsrecordsv1.DnsrecordDetails{{}},
+				},
 			},
 		},
 		{
-			desc:         "listFailError",
-			DNSName:      "testDelete",
-			recordedCall: "DELETE",
+			desc:    "list failure (not 404)",
```
candita, line 77:
```suggestion
			desc:    "list failure (fake timeout)",
```

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +42,133 @@ func Test_Delete(t *testing.T) {
...
-				OutputError:      errors.New("error in DeleteDnsRecord"),
-				OutputStatusCode: http.StatusNotFound,
+				InputId:        "testDelete",
+				OutputError:    errors.New("error in DeleteDnsRecord"),
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusNotFound},
 			},
 		},
 		{
-			desc:         "deleteError",
+			desc:         "delete failure (not 404)",
```
candita, line 107:
```suggestion
			desc:         "delete failure (fake timeout)",
```

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -171,61 +233,156 @@ func Test_createOrUpdateDNSRecord(t *testing.T) {
...
-				OutputError:      errors.New("error in ListAllDnsRecords"),
-				OutputStatusCode: http.StatusRequestTimeout,
+				OutputError:    errors.New("error in ListAllDnsRecords"),
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusRequestTimeout},
 			},
 			expectErrorContains: "error in ListAllDnsRecords",
 		},
 		{
-			desc:         "updateError",
+			desc:         "update error",
```
candita, line 287:
```suggestion
			desc:         "update error (fake timeout)",
```

pkg/dns/ibm/private/dnssvcs_provider_test.go
```diff
@@ -35,82 +41,199 @@ func Test_Delete(t *testing.T) {
...
+					ResourceRecords: []dnssvcsv1.ResourceRecord{},
+				},
 			},
-			deleteDnsRecordInputOutput: dnsclient.DeleteDnsRecordInputOutput{
-				InputId:          "testDelete",
-				OutputError:      nil,
-				OutputStatusCode: http.StatusOK,
+		},
+		{
+			desc:    "list failure (not 404)",
```
candita, line 102:
```suggestion
			desc:    "list failure (fake timeout)",
```

pkg/dns/ibm/private/dnssvcs_provider_test.go
```diff
@@ -35,82 +41,199 @@ func Test_Delete(t *testing.T) {
...
-				OutputError:      errors.New("error in DeleteDnsRecord"),
-				OutputStatusCode: http.StatusNotFound,
+				InputId:        "testDelete",
+				OutputError:    errors.New("error in DeleteDnsRecord"),
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusNotFound},
 			},
 		},
 		{
-			desc:         "deleteError",
+			desc:         "delete failure (not 404)",
```
candita, line 152:
```suggestion
			desc:         "delete failure (fake timeout)",
```

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -171,61 +233,156 @@ func Test_createOrUpdateDNSRecord(t *testing.T) {
...
 	}{
 		{
 			desc:         "happy path",
 			DNSName:      "testUpdate",
 			recordedCall: "PUT",
 			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
-				OutputError:      nil,
-				OutputStatusCode: http.StatusOK,
+				OutputError:    nil,
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusOK},
```
candita, line 245: This test's desc should be "happy path - update" or something similar.

gcs278 (commented):
Thanks for another code review @candita! Comments should be addressed.

gcs278:
e2e-aws-ovn-serial has issue that https://github.com/openshift/origin/pull/29447 is attempting to fix

pkg/dns/ibm/private/dnssvcs_provider.go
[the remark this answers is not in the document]
candita: Sounds reasonable.  I was concerned that there could be a backward compatibility issue if it was possible at any time to have a missing ID.

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +41,133 @@ func Test_Delete(t *testing.T) {
...
+				OutputResult: &dnsrecordsv1.ListDnsrecordsResp{
+					Result: []dnsrecordsv1.DnsrecordDetails{{
+						ID:      pointer.String("testDelete"),
+						Name:    pointer.String("testDelete"),
+						Content: pointer.String("11.22.33.44"),
+					}},
```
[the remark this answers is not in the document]
candita: Maybe `Expected...` but we can live with for now.

pkg/dns/ibm/public/cis_provider.go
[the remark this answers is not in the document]
gcs278: Sorry for resolving this prematurely.

The response type of [DetailedResponse](https://github.com/openshift/cluster-ingress-operator/blob/10966557ef8592858fe068065e526a87c37315c5/vendor/github.com/IBM/go-sdk-core/v5/core/detailed_response.go#L27) has a `StatusCode` field. That should always be set on a valid list API call, so `response` should be not nil here.

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +42,133 @@ func Test_Delete(t *testing.T) {
...
-				OutputStatusCode: http.StatusOK,
+				InputId:        "testDelete",
+				OutputError:    nil,
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusOK},
 			},
 		},
 		{
-			desc:         "listFailNotFound",
-			DNSName:      "testDelete",
-			recordedCall: "DELETE",
```
candita, line 56: nit: why was this removed?

candita:
I think the image stream failure was fixed, so let's try:
/test e2e-aws-ovn-serial

pkg/dns/ibm/public/cis_provider_test.go
gcs278: Sorry that link isn't working for me, github is glitching trying to take me to the right line.

So `recordedCall` was removed here specifically because I changed how this test works. I saw this `listFailNotFound` as a duplicate, unnecessary, and confusing. Here it is before modification for reference: https://github.com/openshift/cluster-ingress-operator/blob/0b9e89b829ea2eb76e193a6efe9e23e303e384c8/pkg/dns/ibm/public/cis_provider_test.go#L53-L66

So it's a nil `OutputError`, meaning the API call worked, but the HTTP status code is 404 (StatusNotFound). The way the original author wrote the logic, if the `err` is nil, it doesn't even look at the status code: https://github.com/openshift/cluster-ingress-operator/blob/21971b42cc86089c823988ba62312c41491d0f2c/pkg/dns/ibm/private/dnssvcs_provider.go#L95-L99

Therefore, it's really a duplicate of the "happy path" unit test case above: https://github.com/openshift/cluster-ingress-operator/blob/0b9e89b829ea2eb76e193a6efe9e23e303e384c8/pkg/dns/ibm/public/cis_provider_test.go#L39-L52

That's why I adjusted the test to `list failure with 404`, so it would actually test a different code path. I agree that this one is really nuanced to review, and the diffs are painful to look at.

As for the other removals of `recordedCall`, they are organized within the commit: 
https://github.com/openshift/cluster-ingress-operator/pull/1133/commits/0b9e89b829ea2eb76e193a6efe9e23e303e384c8. Because they were actually incorrect. The `CallHistory` was using the `recordedCall` of the unit test previously ran (not good!). Once fixing that, a bunch of the `recordedCall`s were removed to make the unit test accurate again.

candita:
/lgtm
/approve
Holding for @Miciah, who would like to look too.
/hold

gcs278:
Releasing the hold due to no new reviews/comments.

/hold cancel

gcs278:
While maybe not absolutely critical, I consider this fix a major fix as users are prevented from recreating IngressController's  service IBMCloud platforms (something we explicitly tell them to do when they change scope). I also have other commitments that prevent me from waiting another month to merge this.

pkg/dns/ibm/private/dnssvcs_provider_test.go
```diff
@@ -108,6 +108,23 @@ func Test_Delete(t *testing.T) {
 			},
 			expectErrorContains: "error in ListAllDnsRecords",
 		},
+		{
+			desc:    "list success but missing ID",
```
Miciah, line 112: Seems like this should be "list failure (fake timeout) and missing ID".  However, it really doesn't matter what the status is if the id is missing or mismatched, so maybe the description should be "list failure (missing ID)"?

pkg/dns/ibm/private/dnssvcs_provider_test.go
```diff
@@ -35,76 +41,182 @@ func Test_Delete(t *testing.T) {
...
+		},
+		{
+			desc:    "list success but nil dns record list",
+			DNSName: "testDelete",
+			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
+				OutputError:    nil,
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusOK},
+				OutputResult:   &dnssvcsv1.ListResourceRecords{},
+			},
+		},
```
Miciah, line 226: The looks the same as "create happy path": https://github.com/openshift/cluster-ingress-operator/blob/c5d48ec7eeafdfa58cf005ff50d822228db65ca6/pkg/dns/ibm/private/dnssvcs_provider_test.go#L385-L400

Ah, but "list success but nil dns record list" is missing `target`.  Should it rather be called "list success but _no matching entry_ in dns record list"?

pkg/dns/ibm/private/dnssvcs_provider_test.go
gcs278: Good catch, a bit of a copypasta. The response code doesn't get checked if the error is nil, but I meant to just have it be a success to be a less confusing test.

Miciah: Ah, I see, the HTTP error is ignored because `ListResourceRecords` returns a nil error value.  I guess "list success" refers to the error value, not the HTTP code.  It's kind of confusing to have a test case where the HTTP code indicates an error but `ListResourceRecords` reports no error.  I wouldn't have said anything if you had `http.StatusOK` instead of `http.StatusRequestTimeout`.  I guess it doesn't really matter, and the conspicuous test input did prompt me to look more closely at the logic, which is a good thing.  ¯\\\_(ツ)\_/¯

pkg/dns/ibm/private/dnssvcs_provider_test.go
Miciah: Well, I guess the "create happy path" is the same as "no matching entry in dns record list".  In both cases, the list happens to be empty, but in this case, the target is an empty string.  

It's weird that specifying an empty target doesn't cause an error, and it isn't clear that this test was intended to test that scenario.  Given that that's the current behavior, I don't mind having a test case.  However, the test case should be named clearly (something like "create with an empty target"), and we might follow-up with a fix to reject records with empty target and update the test case accordingly to expect an error.

gcs278: I think you misplaced your comment on the delete unit test, but I think you meant to comment on the one for create (same named test cases).

Yea I actually think your first thought is correct, it feels redundant. The target does not impact anything in this test. Right, I was going to say, maybe it provides value by encoding/protecting a specific behavior, but that wasn't the intention.

I'll at least update the name.

gcs278:
Working through some comments
/hold

pkg/dns/ibm/public/cis_provider_test.go
Miciah: > The way the original author wrote the logic, if the `err` is nil, it doesn't even look at the status code:
[...]
> Therefore, it's really a duplicate of the "happy path" unit test case above: 
[...]
> That's why I adjusted the test to `list failure with 404`, so it would actually test a different code path.

In the spirit of black-box testing, I think it's fine to have both test cases.  Even if an examination of the logic that is being tested reveals that it uses the same code path for both cases, they are still two meaningfully different inputs from an outside perspective.  If the logic were modified, we would benefit from having both test cases to verify that both inputs continued to result in the same behavior.  

That said, don't worry about adding back the test case at this point (unless you really want to!).  I just wanted to respond to the implication that a test case that follows the same code path as another test case is necessarily redundant.

pkg/dns/ibm/public/cis_provider_test.go
```diff
@@ -37,77 +41,133 @@ func Test_Delete(t *testing.T) {
...
 			listAllDnsRecordsInputOutput: dnsclient.ListAllDnsRecordsInputOutput{
-				OutputError:      errors.New("error in ListAllDnsRecords"),
-				OutputStatusCode: http.StatusRequestTimeout,
-			},
-			deleteDnsRecordInputOutput: dnsclient.DeleteDnsRecordInputOutput{
-				InputId:          "testDelete",
-				OutputError:      nil,
-				OutputStatusCode: http.StatusOK,
+				OutputError:    errors.New("error in ListAllDnsRecords"),
+				OutputResponse: &core.DetailedResponse{StatusCode: http.StatusRequestTimeout},
```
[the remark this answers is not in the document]
Miciah: > This function is making sure we don't execute a delete command if there is a timeout.

Crucially, the test case also verifies that the DNS provider logic returns an error value.  This error value is what lets the DNS controller know that it needs to retry the whole operation.

pkg/dns/ibm/private/dnssvcs_provider.go
Miciah: We could add a Boolean argument to `ValidateInputDNSData` for allowing/rejecting multiple targets.  That said, `ValidateInputDNSData` always felt weird to me anyway; it seems like the validation should be moved into the controller, or even enforced by API validation.  The approach in this PR seems fine to me.  Eventually, it would be nice to follow up with some rethinking and refactoring of the validation logic.

pkg/dns/ibm/public/cis_provider.go
```diff
@@ -160,41 +165,45 @@ func (p *Provider) createOrUpdateDNSRecord(record *iov1.DNSRecord, zone configv1
...
+		createOpt.SetName(record.Spec.DNSName)
+		createOpt.SetType(string(record.Spec.RecordType))
+		createOpt.SetContent(target)
+		createOpt.SetTTL(record.Spec.RecordTTL)
+		_, _, err := dnsService.CreateDnsRecord(createOpt)
 		if err != nil {
-			if response != nil && response.StatusCode != http.StatusNotFound {
-				return fmt.Errorf("createOrUpdateDNSRecord: failed to list the dns record: %w", err)
-			}
-			continue
```
Miciah, line 170: This change is really subtle, and it isn't clear to me how it relates to the description of the changes in the commit message.  (Indeed, it would be easy not to notice this change at all, but for the change to the unit tests!)

Of course, you cannot keep the literal `continue` statement since this logic is no longer inside a `for` loop.  However, you could have replaced `continue` with `return`.  Presumably, the reason you instead simply drop the `continue` is that the old logic was wrong:  It was skipping the create/update in the case of a not-found error, but the correct behavior _is_ to create in the case of a not-found error.  Am I correctly understanding the old behavior and the intent of your change?  

The commit message mentions two changes to this logic:

- "remove the filtering by target and rely solely on filtering by name" and
- "only create using the first target".

I think it makes sense to mention "fall through to the create logic in the case of a not-found error" as a distinct change as well.

Miciah:
I've finished reviewing.  Sorry for the last-minute comments!  There's nothing absolutely critical, so
/lgtm
but I can re-/lgtm if you do have time to push changes for my minor suggestions.

pkg/dns/ibm/public/cis_provider.go
gcs278: > Presumably, the reason you instead simply drop the continue is that the old logic was wrong: It was skipping the create/update in the case of a not-found error, but the correct behavior is to create in the case of a not-found error. Am I correctly understanding the old behavior and the intent of your change?

Exactly, that was my thought. One other point:
- It was inconsistent with the createOrUpdateDNSRecord in dnsvcs_provider.go for no obvious reason. Based on the reason you explained, I dropped the continue/return, reference: https://github.com/openshift/cluster-ingress-operator/blob/b0b91bb9ea12c4e09676551a0c19df4c3ce878c3/pkg/dns/ibm/private/dnssvcs_provider.go#L201-L209

I'll update the commit message.

gcs278:
@miciah ready for review/lgtm.

Miciah:
Thanks!
/lgtm

candita:
Another case where acquiring an AWS lease took 30 minutes and the test timed out after 4 hours.
/test e2e-aws-ovn-serial

gcs278:
I forgot to remove the hold

/hold cancel

Installation failures and unrelated failures (nothing related to ibmcloud)
/retest

PR #1134, opened 2024-08-27
OCPBUGS-38871: ingress: deployment: explicitly set DeploymentStrategy in SingleReplica case

sjenning:
Currently, the code does not set `deployment.Spec.Strategy` before breaking out in the SingleReplica case. This results in the kube-apiserver defaulting the `Strategy`. On the next reconcile pass, this code will attempt to unset the defaulted values leading to a hot reconciliation loop.

To avoid this, explicitly set the `Strategy` to the kube default before the `break` out of the switch.

xref https://github.com/openshift/cluster-ingress-operator/pull/810

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -309,7 +309,16 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		// because we need the rule on multi-node clusters, even if
 		// desiredReplicas is 1, so that a rolling update schedules
 		// the new pod to the same node as the old pod.
+		pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
 		if singleReplica(ingressConfig, infraConfig) {
+			// We must explicitly set deployment strategy to the kube defaults to avoid a reconcile loop
+			deployment.Spec.Strategy = appsv1.DeploymentStrategy{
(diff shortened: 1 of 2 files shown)


sjenning:
cc @michael-topchiev

Miciah:
We've been trying to be more consistent in commit messages and follow [upstream guidelines](https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines). Would you mind adding a reference to https://issues.redhat.com//browse/OCPBUGS-38871 to the commit messages and making sure lines are wrapped (50 chars for the subject, 72 for the body)?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -309,7 +309,16 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		// because we need the rule on multi-node clusters, even if
 		// desiredReplicas is 1, so that a rolling update schedules
 		// the new pod to the same node as the old pod.
+		pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
```
Miciah, line 312: We should change this to `ptr.To[intstr.IntOrString]`, but not in this PR as that would complicate backports and as we have several places where we should make this change.

Miciah:
/approve
@sjenning, I can /lgtm as soon as you update the commit message per my earlier comment.

Miciah:
Thanks!
/lgtm

PR #1144, opened 2024-09-12
OCPBUGS-42004: Set the MI client ID for the ARO HCP override

bryan-cox:
Set the managed identity client ID for the ARO HCP override to use managed identity.


bryan-cox:
I think we may need to set the client ID after all. Opening this as draft for now until I confirm that.

Miciah:
Would you mind adding a reference to https://issues.redhat.com//browse/OCPBUGS-42004 to the commit message?

bryan-cox:
> Would you mind adding a reference to https://issues.redhat.com//browse/OCPBUGS-42004 to the commit message?

@Miciah yeah no problem, it's been added.

PR #1145, opened 2024-09-12
OCPBUGS-31521: Don't allow DNS records for gateway to be created if they clash with any existing ingress controller.

rfredette:
If a gateway attempts to create a listener for a domain that's already served by an ingress controller, don't create a DNSRecord for that domain, and log an error.

In the future, some more visible method of error reporting should be added, such as an alert or a status condition on a relevant object, but that is intentionally left for a future change.

### DIFF
--- a/pkg/operator/controller/gateway-service-dns/controller.go
+++ b/pkg/operator/controller/gateway-service-dns/controller.go
@@ -17,6 +17,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -241,8 +242,18 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 		Name:       service.Name,
 		UID:        service.UID,
 	}
+	ics := operatorv1.IngressControllerList{}
+	if err := r.cache.List(ctx, &ics); err != nil {
+		log.Error(err, "failed to list Ingress Controllers")
+		return []error{err}
+	}
 	var errs []error
 	for _, domain := range domains {
+		// Check if the domain matches one from an existing ingress controller. If so, log an error and ignore the invalid domain.
+		if err := domainClashes(domain, ics); err != nil {
+			log.Error(fmt.Errorf("error creating DNS record for gateway %s: %w", gateway.Name, err), "ignoring invalid gateway domain")
+			continue
+		}
 		name := operatorcontroller.GatewayDNSRecordName(gateway, domain)
 		dnsPolicy := iov1.UnmanagedDNS
 		if dnsrecord.ManageDNSForDomain(domain, infraConfig.Status.PlatformStatus, dnsConfig) {
@@ -252,6 +263,20 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 		errs = append(errs, err)
 	}
 	return errs
+}
+
+// domainClashes checks if domain clashes with any of the wildcard records of the ingress controllers in icList. If
+// there is a clash, an error is returned describing which ingress controller already uses the specified domain name.
+func domainClashes(domain string, icList operatorv1.IngressControllerList) error {
+	for _, ic := range icList.Items {
+		if len(ic.Status.Domain) == 0 {
+			continue
+		}
+		if strings.Compare(domain, "*."+ic.Status.Domain+".") == 0 {
+			return fmt.Errorf("listener hostname %q conflicts with the domain from ingress controller %s.", domain, ic.Name)
+		}
+	}
+	return nil
 }
 
 // deleteStaleDNSRecordsForGateway deletes any DNSRecord CRs that are associated
--- a/pkg/operator/controller/gateway-service-dns/controller_test.go
+++ b/pkg/operator/controller/gateway-service-dns/controller_test.go
@@ -11,6 +11,7 @@ import (
 	gatewayapiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 
 	corev1 "k8s.io/api/core/v1"
@@ -40,6 +41,17 @@ func Test_Reconcile(t *testing.T) {
 				Type: configv1.AWSPlatformType,
 			},
 		},
+	}
+	ic := func(name, domain string) *operatorv1.IngressController {
+		return &operatorv1.IngressController{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: "openshift-ingress-operator",
+				Name:      name,
+			},
+			Status: operatorv1.IngressControllerStatus{
+				Domain: domain,
+			},
+		}
 	}
 	gw := func(name string, listeners ...gatewayapiv1beta1.Listener) *gatewayapiv1beta1.Gateway {
 		return &gatewayapiv1beta1.Gateway{
@@ -127,6 +139,7 @@ func Test_Reconcile(t *testing.T) {
 				infraConfig,
 				gw("example-gateway", l("stage-http", "*.stage.example.com", 80)),
 				svc("example-gateway", gatewayManagedLabel, exampleGatewayLabel, ingHost("lb.example.com")),
+				ic("default", "apps.example.com"),
 			},
 			reconcileRequest: req("openshift-ingress", "example-gateway"),
 			expectCreate:     []client.Object{},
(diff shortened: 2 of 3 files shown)


rfredette:
Hit some 500 internal server errors during initial setup on several tests
/retest

candita:
Lots of disruption failures.
/retest

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -241,8 +242,28 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
+	}
 	var errs []error
 	for _, domain := range domains {
+		// Check if the domain matches one from an existing ingress controller. If so, log an error and ignore the invalid domain.
+		domainClashes := false
+		for _, ic := range ics.Items {
+			if len(ic.Status.Domain) == 0 {
+				continue
+			}
+			if domainsEqual(domain, "*."+ic.Status.Domain) {
```
candita, line 258: It's possible for a gateway hostname to be non-wildcard.  Is there a case where you had to prepend `*.`  to the ingress controller domain to make this work?

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -241,8 +242,28 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
+				log.Error(fmt.Errorf("gateway %s domain %q matches existing domain from ingress controller %s.", gateway.Name, domain, ic.Name), "ignoring invalid gateway domain")
```
candita, line 259:
```suggestion
				log.Error(fmt.Errorf("error creating DNS record for gateway %s with listener hostname %q because it conflicts with the domain from ingress controller %s.", gateway.Name, domain, ic.Name), "ignoring invalid gateway domain")
```

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -241,8 +242,28 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
 	}
+	ics := operatorv1.IngressControllerList{}
+	if err := r.cache.List(ctx, &ics); err != nil {
+		log.Error(err, "failed to list Ingress Controllers")
+		return []error{err}
+	}
 	var errs []error
 	for _, domain := range domains {
+		// Check if the domain matches one from an existing ingress controller. If so, log an error and ignore the invalid domain.
+		domainClashes := false
```
candita, line 253: You don't really need this boolean, just break from the loop when you see the match.

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -241,8 +242,28 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
+		for _, ic := range ics.Items {
+			if len(ic.Status.Domain) == 0 {
+				continue
+			}
+			if domainsEqual(domain, "*."+ic.Status.Domain) {
+				log.Error(fmt.Errorf("gateway %s domain %q matches existing domain from ingress controller %s.", gateway.Name, domain, ic.Name), "ignoring invalid gateway domain")
+				domainClashes = true
+				break
```
candita, line 261:
```suggestion
				continue
```

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -254,6 +275,18 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 	return errs
 }
 
+// domainsEqual compares 2 DNS domains and returns if they are equivalent. Domains will be compared as if they were
+// fully qualified (ending with ".").
```
candita, line 279: The trailing dot should already be on the gateway hostname.  https://github.com/rfredette/cluster-ingress-operator/blob/master/pkg/operator/controller/gateway-service-dns/controller.go#L219-L221

pkg/operator/controller/gateway-service-dns/controller.go
candita: It would be a little faster to just do `strings.Compare(domain, ic.Status.Domain+".") == 0`

pkg/operator/controller/gateway-service-dns/controller_test.go
```diff
@@ -246,12 +266,32 @@ func Test_Reconcile(t *testing.T) {
...
+				svc("example-gateway", gatewayManagedLabel, exampleGatewayLabel, ingHost("lb.example.com")),
+				ic("default", "apps.example.com"),
+			},
+			reconcileRequest: req("openshift-ingress", "example-gateway"),
+			expectCreate: []client.Object{
+				dnsrecord("example-gateway-644bf77744-wildcard", "*.stage.apps.example.com.", iov1.ManagedDNS, exampleGatewayLabel, "lb.example.com"),
+			},
+			expectUpdate: []client.Object{},
+			expectDelete: []client.Object{},
+		},
```
candita, line 287: Add another test with the listener hostname not a wildcard, but would end up clashing with the ingress controller domain that produces a dnsrecord with a wildcard. So, the listener hostname "apps.example.com" too.

pkg/operator/controller/gateway-service-dns/controller.go
rfredette: While the gateway hostname could be either a wildcard or not, `ic.Status.Domain` is always non-wildcard, but the DNS record created for it is always a wildcard (`desiredWildcardDNSRecord` always adds `*.` before and `.` after the domain, shown here: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/resources/dnsrecord/dns.go#L108). We need to compare the domain names on the DNS records, so `ic.Status.Domain` does need `*.` prepended.

pkg/operator/controller/gateway-service-dns/controller.go
rfredette: As written, this boolean is needed because you can't `continue` or `break` out of both the inner and outer `for` loop from inside the inner one. However, I reorganized the new code here to not need the boolean and I think it's more readable the new way.

rfredette:
Test failures include underlying cluster instability
/retest

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -241,8 +242,18 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
...
+	ics := operatorv1.IngressControllerList{}
+	if err := r.cache.List(ctx, &ics); err != nil {
+		log.Error(err, "failed to list Ingress Controllers")
+		return []error{err}
+	}
 	var errs []error
 	for _, domain := range domains {
+		// Check if the domain matches one from an existing ingress controller. If so, log an error and ignore the invalid domain.
+		if err := domainClashes(domain, ics); err != nil {
+			log.Error(fmt.Errorf("error creating DNS record for gateway %s: %w", gateway.Name, err), "ignoring invalid gateway domain")
```
candita, line 254: Please use %v instead of %w for errors.  Also, since a Gateway can have several Listeners/domains, you have to mention the domain here.
```suggestion
			log.Error(fmt.Errorf("error creating DNS record for gateway %s, domain %q: %v", gateway.Name, domain, err), "ignoring invalid gateway domain")
```

pkg/operator/controller/gateway-service-dns/controller_test.go
```diff
@@ -246,12 +266,54 @@ func Test_Reconcile(t *testing.T) {
...
+		{
+			name: "gateway with two unique host names, neither of which clashes with an existing ingress controller",
+			existingObjects: []runtime.Object{
+				dnsConfig, infraConfig,
+				gw(
+					"example-gateway",
+					l("stage-https", "*.stage.apps.example.com", 443),
+					// apps.example.com looks like it'll clash with the default ic below, but the ic creates a record
+					// for *.apps.example.com, which doesn't actually clash with apps.example.com
+					l("apps-https", "apps.example.com", 443),
```
candita, line 297: This should clash though, because the DNSRecord for the listener will prefix it with `*.` too.

pkg/operator/controller/names.go
```diff
@@ -287,8 +288,12 @@ func ServiceMeshSubscriptionName() types.NamespacedName {
...
 // suffix "-wildcard".
 func GatewayDNSRecordName(gateway *gatewayapiv1beta1.Gateway, host string) types.NamespacedName {
+	name := fmt.Sprintf("%s-%s", gateway.Name, util.Hash(host))
+	if strings.HasPrefix(host, "*.") {
+		name = name + "-wildcard"
+	}
 	return types.NamespacedName{
 		Namespace: gateway.Namespace,
-		Name:      fmt.Sprintf("%s-%s-wildcard", gateway.Name, util.Hash(host)),
+		Name:      name,
```
candita, line 297: We can't change this, this is a standard way of naming the DNSRecord, and it will always be a wildcard.

pkg/operator/controller/gateway-service-dns/controller_test.go
candita: Yes, this is actually a different problem.  I don't know what happens if a Gateway listener tries to claim a name (abc.example.com) that would match a *.apps.example.com wildcard. Will add a test in the e2e tests for this and take it from there.

pkg/operator/controller/gateway-service-dns/controller.go
candita: Ignore this.  I have since learned that this is not preferred.  Or no longer preferred.

rfredette:
Closing because this is superceded by https://github.com/openshift/cluster-ingress-operator/pull/1229 .

PR #1048, opened 2024-04-24
Empty commit for testing CI failures

Miciah:

PR #1147, opened 2024-09-16
RFE-6242: WIP - openstack: support setting LB IP

EmilienM:
Add support for setting the (deprecated) Service field: `loadBalancerIP` on platform OpenStack when the LB is not internal.


Miciah (commented):
We should set `effectiveStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS` when the platform is OpenStack.  You can add a check around the existing `if !domainMatchesBaseDomain` and `if platformStatus.GCP.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType` checks near the top of `setDefaultPublishingStrategy`.  

At some point, we should probably refactor this logic because `setDefaultPublishingStrategy` is already quite large, and the logic for determining the value of `DNSManagementPolicy` somewhat couples the operator's ingress and dns controllers.  We can do that refactoring in a follow-up.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -398,6 +398,15 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
 			}
+
+			// Set the Load Balancer IP for external load balancers on OpenStack only
+			if platform.Type == configv1.OpenStackPlatformType {
+				if lb != nil && lb.ProviderParameters != nil &&
+					lb.ProviderParameters.Type == operatorv1.OpenStackLoadBalancerProvider &&
+					lb.ProviderParameters.OpenStack != nil {
+					service.Spec.LoadBalancerIP = lb.ProviderParameters.OpenStack.LoadBalancerIP
+				}
+			}
```
Miciah, line 409: The API definition doesn't mention that `loadBalancerIP` is only valid for external scope: https://github.com/openshift/cluster-ingress-operator/blob/0e0f3e3f26bfe8bc3737999886c20e3afd1b3c9c/vendor/github.com/openshift/api/operator/v1/types_ingress.go#L681-L683

Can you add that limitation to the API field godoc?  I expect we could even add some CEL validation rules to prevent specifying `loadBalancerIP` with `scope: Internal`.  

Alternatively, if that limitation doesn't in fact exist, then the `if` block that you are adding here should be moved outside the `if isInternal`/`else`.  

Actually, even if using `loadBalancerIP` is limited to external scope, I think the code would be easier to read if you moved the OpenStack logic to a new `case` in the `switch` below.  The `if isInternal`/`else` blocks are really more for setting the platform-specific scope annotations.

EmilienM: @mdbooth do you think we should restrict this to the external scope?
I'm trying to find a reason why a FIP would be attached to a internally scoped LB.

EmilienM: actually, reading CPO makes me think that this is only for external LBs: https://github.com/kubernetes/cloud-provider-openstack/blob/333a126f63038d52fad1e306eff6ee4ef7f53948/pkg/openstack/loadbalancer.go#L624-L625

EmilienM: @Miciah I'm going to update the API as requested: https://github.com/openshift/api/pull/2051
And move that code into the switch later. I confirm we only want the floating IP to be for external LB.

EmilienM:
WIP for now because the API is being changed: https://github.com/openshift/api/pull/2051

EmilienM:
> Could you also add some test cases to `TestSetDefaultPublishingStrategySetsPlatformDefaults` and `TestSetDefaultPublishingStrategyHandlesUpdates` in `pkg/operator/controller/ingress/controller_test.go` and to `Test_desiredLoadBalancerService` in `pkg/operator/controller/ingress/load_balancer_service_test.go`?

For `TestSetDefaultPublishingStrategySetsPlatformDefaults` I think it's not possible because the default is not an LB but a HostNetwork endpoint strategy. But I'll add tests for others.

candita:
@gcs278 assigning to you for visibility, since this impacts one of your pending PRs.

/assign @gcs278

pkg/operator/controller/ingress/controller.go
```diff
@@ -617,6 +617,27 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					statusLB.ProviderParameters.IBM.Protocol = specProtocol
 					changed = true
 				}
+			case operatorv1.OpenStackLoadBalancerProvider:
```
EmilienM, line 620: @Miciah and reviewers: I didn't add a check here whether the scope is external (we have CEL validation) I wonder if we want it.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -494,6 +494,16 @@ func TestSetDefaultPublishingStrategyHandlesUpdates(t *testing.T) {
 			}
 			return eps
 		}
+		openstackLB = func(floatingIP string) *operatorv1.EndpointPublishingStrategy {
+			eps := eps(lbs(operatorv1.ExternalLoadBalancer, &managedDNS))
+			eps.LoadBalancer.ProviderParameters = &operatorv1.ProviderLoadBalancerParameters{
+				Type: operatorv1.OpenStackLoadBalancerProvider,
+				OpenStack: &operatorv1.OpenStackLoadBalancerParameters{
+					FloatingIP: floatingIP,
```
EmilienM, line 502: same here we don't have a check, do we want one?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -475,6 +475,15 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 			if !isInternal {
 				service.Annotations[alibabaCloudLBAddressTypeAnnotation] = alibabaCloudLBAddressTypeInternet
 			}
+		case configv1.OpenStackPlatformType:
+			// Set a floating IP only if the load balancer scope is external.
+			if !isInternal {
```
EmilienM, line 480: here we have one though.

EmilienM:
/test e2e-hypershift e2e-aws-ovn-single-node

The openstack job never worked, I'm looking at it via another PR.

pkg/operator/controller/ingress/controller.go
```diff
@@ -453,6 +453,11 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 			effectiveStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
 		}
 
+		// OpenStack platform does not support managed DNS for load balancers.
+		if platformStatus.Type == configv1.OpenStackPlatformType && platformStatus.OpenStack != nil {
```
gcs278, line 457: I think this should be unit tested in `TestSetDefaultPublishingStrategySetsPlatformDefaults`. I looked at the test, be careful with `makePlatformStatus`, it doesn't set `platformStatus.OpenStack` so this logic doesn't get exercised.

pkg/operator/controller/ingress/load_balancer_service_test.go
gcs278, line 1: `Test_loadBalancerServiceChanged` should test `spec.loadBalancerIP`. One second thought, I see that `loadBalancerServiceChanged` won't actually detect changes to `spec.loadBalancerIP`. This appears to say you can't actually up the Floating IP after the service is updated, have you tried that?

Can you update the FloatingIP without recreating the service? Does OpenStack reconcile that change properly?

pkg/operator/controller/ingress/controller.go
```diff
@@ -617,6 +622,27 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 					statusLB.ProviderParameters.IBM.Protocol = specProtocol
 					changed = true
 				}
+			case operatorv1.OpenStackLoadBalancerProvider:
+				// The only provider parameter that is supported
+				// for OpenStack is the FloatingIP parameter.
+				var statusFloatingIP string
```
gcs278, line 628: So I hate to bring this up, but I have been pushing for new APIs to have their status counterpart reflect the *actual* state. You have followed the traditional path of `status` representing the desired state of the IngressController (user desire from`spec` + defaulting).

For `Subnets`, `EIPAllocations`, `AllowedSourceRanges`, we have started to pave this new path for status to reflect actual state. But I have mixed feelings about whether we did this prematurely because:
1. `status.endpointPublishingStrategy...` has mixed purpose status (some actual, some desired). That's probably bad and confusing UX (we probably should have forked into new status structure....)
2. The Ingress Operator spec & status related code is splitting into two logic "code flows" (complexity is :chart_with_upwards_trend:)

And we need to consider the priority of your RFE here. But when we start using the new `FloatingIP` status field, the API Team has said a change in "behavior" of status is incompatible (i.e. we are locked in). 

I'm not against maintaining the traditional `status` patterns here given the circumstances. Rewriting this code would be quite painful and time consuming.  But I'd like to have consensus from our team on what our plan is. I'm thinking we should continue having `status.endpointPublishingStrategy` traditionally as *desired* as you've done here, but we should have a backlog task to add a new `status` struct for representing *actual* values to keep things organized. @Miciah thoughts?

gcs278 (commented):
Thanks for the PR! Couple of comments, feel free to reach out.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -475,6 +475,15 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
+				if lb != nil && lb.ProviderParameters != nil &&
+					lb.ProviderParameters.Type == operatorv1.OpenStackLoadBalancerProvider &&
+					lb.ProviderParameters.OpenStack != nil {
+					service.Spec.LoadBalancerIP = lb.ProviderParameters.OpenStack.FloatingIP
```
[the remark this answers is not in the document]
EmilienM: the last time I checked it didn't but I'll investigate again and report my findings here.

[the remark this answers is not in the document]
mdbooth: I also checked, and it did not as of about a month ago.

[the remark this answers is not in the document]
gcs278: Okay, understood.

I was a bit mislead, I was thinking a future rebase would remove `spec.loadBalancerIP` field, but that would require a rev of the service API (service V2). I'm betting that would be a massive effort, and that's not going to happen anytime soon, so I think my concern is mostly a minor one.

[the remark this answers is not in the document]
EmilienM: :sweat_smile: I'm glad we sort this out.
I'll look at other comments this week. Thanks for all your inputs, I'll take them in consideration one by one and address the comments.

pkg/operator/controller/ingress/controller.go
EmilienM: oh thanks, I missed that.

pkg/operator/controller/ingress/load_balancer_service_test.go
EmilienM: I'll give it a try in my env this week but I'm nearly sure that updating `Service.Spec.LoadBalancerIP` will effectively update the floating IP used for the Octavia load balancer that was created. I'm nearly sure CCM/cloud-provider-openstack will unassign the FIP and assign the new one if it's technically feasible. And the Service wouldn't be recreated. I'll test that though and update it here.

pkg/operator/controller/ingress/controller.go
EmilienM: I'll let you decide how to want to deal with that. I'm happy either way, I can leave it like this or patch `IngressStatusesEqual` to add this new field. And I can also create that task and I could be assigned to it for the next release if you want.

pkg/operator/controller/ingress/load_balancer_service_test.go
EmilienM: ok I was wrong. I updated the spec and nothing happened. So if we wanted to update the FIP, we'll need to re-create a new Service which makes sense. So I guess we don't need another test?

gcs278: > So if we wanted to update the FIP, we'll need to re-create a new Service which makes sense. 

Okay, so to be clear, you are saying that OpenStack CCM does not support updating `spec.loadBalancerIP` after creation? 

If it doesn't, that's perfectly fine, but we do have a [process](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/lb-subnet-selection-aws.md#effectuating-subnet-updates) in place that provides the cluster admin a `Progressing` condition with a descriptive message.

Check out existing patterns in `loadBalancerServiceIsProgressing` in `load_balancer_service.go`. You'll need to add a condition to detect when the desired != actual. That function is getting a little confusing, so be careful. As I mentioned in the other comment, the EIPAllocation and Subnets `status` fields represent the actual value, while your pattern of `status` should be more like the scope change detection in that function. You should add to that function, along with the appropriate unit testing.

EmilienM: >  you are saying that OpenStack CCM does not support updating spec.loadBalancerIP after creation?

Indeed, the code that use `loadBalancerIP`:
https://github.com/search?q=repo%3Aopenshift%2Fcloud-provider-openstack%20loadBalancerIP&type=code
https://github.com/openshift/cloud-provider-openstack/blob/f917b21a8a748ac0d34fb2e6bfcc1b8cf30179b5/pkg/openstack/loadbalancer.go#L267-L271
which is part of this function: `createOctaviaLoadBalancer`.
This function is only called from `ensureOctaviaLoadBalancer` if the LB doesn't exist yet:
https://github.com/openshift/cloud-provider-openstack/blob/f917b21a8a748ac0d34fb2e6bfcc1b8cf30179b5/pkg/openstack/loadbalancer.go#L1694-L1710

So indeed, the `loadBalancerIP` isn't taken in account when an update on the Service occurs. I haven't seen any complaint or bug report so far about it, so I guess this is just like this now :man_shrugging:.

pkg/operator/controller/ingress/controller.go
gcs278: @EmilienM I had a long conversation with @Miciah about IngressController status today, with this PR in mind. We arrived at the conclusion that we should continue the pattern for status set forth in syncIngressControllerStatus: https://github.com/openshift/cluster-ingress-operator/blob/7fef44643ba04946a23b591a312cb46dbc25aac5/pkg/operator/controller/ingress/status.go#L77-L82
 
By this, I mean status should reflect the *actual* state, not the desired state. Mind trying this out? This mean a couple of things will change in your PR:
1. You shouldn't be making any changes to `setDefaultPublishingStrategy` or `setDefaultProviderParameters` (i.e. your current changes should be remove)
2. You'll need to add some logic in `syncIngressControllerStatus`. I recommend following the existing patterns set forth by Subnets & EIPAllocations.
3. You'll need to update any references that consume status as the desired state, and ensure you consume status as the actual state (e.g. your have & want logic in `loadBalancerServiceIsProgressing`)
4. Don't forget about updating `IngressStatusesEqual` too
5. I'd like to put a note in the docs of `setDefaultPublishingStrategy` about this pattern we are changing, that way we can communicate our desired pattern to future developers.

I might be forgetting something, but hope this helps with the transition. Sorry for churn.

EmilienM: 1. :heavy_check_mark: 
2. :heavy_check_mark: 
3. :heavy_check_mark: but no tests yet, I want to make sure I'm on the right path first.
4. :heavy_check_mark: but no tests yet, I want to make sure I'm on the right path first.
5. I'll need your help for this one or maybe we can do it via its own PR, please.

pkg/operator/controller/ingress/status.go
```diff
@@ -967,7 +975,6 @@ func updateIngressControllerAWSSubnetStatus(ic *operatorv1.IngressController, se
 // updateIngressControllerAWSEIPAllocationStatus mutates the provided IngressController object to
 // sync its status to the effective eipAllocations on the LoadBalancer-type service.
 func updateIngressControllerAWSEIPAllocationStatus(ic *operatorv1.IngressController, service *corev1.Service) {
-	// Set the eipAllocations status based on the actual service annotation and on the load balancer type `NLB`.
```
gcs278, line 970: Did you mean to delete this comment?

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -836,6 +847,22 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
 			err := fmt.Errorf("%[1]s  To effectuate this change, you must delete the service: `oc -n %[2]s delete svc/%[3]s`; the service load-balancer will then be deprovisioned and a new one created. This will most likely cause the new load-balancer to have a different host name and IP address and cause disruption. To return to the previous state, you can revert the change to the IngressController: `%[4]s`", changedMsg, service.Namespace, service.Name, ocPatchRevertCmd)
 			errs = append(errs, err)
 		}
+
+		if platform.Type == configv1.OpenStackPlatformType {
```
gcs278, line 851: you have two conditionals here, I think you forgot to clean up one of them.

pkg/operator/controller/ingress/status.go
```diff
@@ -979,6 +986,14 @@ func updateIngressControllerAWSEIPAllocationStatus(ic *operatorv1.IngressControl
 	}
 }
 
+// updateIngressControllerFloatingIPOpenStackStatus mutates the provided IngressController object to
+// sync its status to the effective floatingIP on the LoadBalancer-type service.
+func updateIngressControllerFloatingIPOpenStackStatus(ic *operatorv1.IngressController, service *corev1.Service) {
+	if fip := getOpenStackFloatingIPInStatus(ic); fip != "" {
```
gcs278, line 992: I think this doesn't do much, it sets status to status. You should be taking the spec value from the Service, and placing it into the status. You'll need a new function similar to `getSubnetsFromServiceAnnotation` but for getting loadBalancerIP.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -480,6 +480,17 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
 			}
+		case configv1.OpenStackPlatformType:
+			// Set a floating IP only if the load balancer scope is external.
+			if !isInternal {
+				if lb != nil && lb.ProviderParameters != nil &&
+					lb.ProviderParameters.Type == operatorv1.OpenStackLoadBalancerProvider &&
+					lb.ProviderParameters.OpenStack != nil {
+					// We know that LoadBalancerIP is deprecated but in this current version of Service, this won't be removed.
+					// It'll probably be removed in Service v2 which is not going to happen in the near future.
+					service.Spec.LoadBalancerIP = lb.ProviderParameters.OpenStack.FloatingIP
```
gcs278, line 491: This logic needs to now use only the IngressController spec to determine the desired FloatingIP, not status (which is the `lb` variable here).

pkg/operator/controller/ingress/status.go
EmilienM: oops

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -370,8 +370,9 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 
 	service.Spec.Selector = controller.IngressControllerDeploymentPodSelector(ci).MatchLabels
 
-	lb := ci.Status.EndpointPublishingStrategy.LoadBalancer
-	isInternal := lb != nil && lb.Scope == operatorv1.InternalLoadBalancer
+	lbSpec := ci.Spec.EndpointPublishingStrategy.LoadBalancer
```
gcs278, line 373: double check for me, but I think this could be a NIL pointer exception if you don't have `spec.EndpointPublishingStrategy` set. I don't think `spec` gets "defaulted" anywhere unlike status.

pkg/operator/controller/ingress/status.go
```diff
@@ -979,6 +987,16 @@ func updateIngressControllerAWSEIPAllocationStatus(ic *operatorv1.IngressControl
 	}
 }
 
+// updateIngressControllerFloatingIPOpenStackStatus mutates the provided IngressController object to
+// sync its status to the effective floatingIP on the LoadBalancer-type service.
+func updateIngressControllerFloatingIPOpenStackStatus(ic *operatorv1.IngressController, service *corev1.Service) {
+	// FloatingIP should be initialized by setDefaultPublishingStrategy
+	// when an IngressController is admitted, so we don't need to initialize here.
+	if fip := getFloatingIPFromSpec(service); fip != "" {
```
gcs278, line 995: Not quite, you'll want to get this from the service, the actual value. So you'll need a new function called `getLoadBalancerIPFromService`. Or you might not really even need the function, just read the value of `service.spec.loadBalancerIP`.

pkg/operator/controller/ingress/load_balancer_service.go
EmilienM: I've added a check.

pkg/operator/controller/ingress/status.go
EmilienM: I think I've addressed the comment.

EmilienM:
just a rebase on master.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -877,6 +904,12 @@ func Test_computeLoadBalancerProgressingStatus(t *testing.T) {
 			platformStatus: awsPlatformStatus,
 			expectStatus:   operatorv1.ConditionFalse,
 		},
+		// TODO: openstack tests
+		{
+			name:         "LoadBalancerService, OpenStack floating IP spec and status are empty",
+			ic:           loadBalancerIngressControllerWithFloatingIP("", ""),
+			expectStatus: operatorv1.ConditionTrue,
```
gcs278, line 911: I'm a bit confused why this is True. Shouldn't progressing be False if both spec and status are empty?

Also, we should add more test cases here, like when it's not empty, but that's probably what you mean by the TODO.

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -674,6 +702,18 @@ func Test_desiredLoadBalancerService(t *testing.T) {
 				localWithFallbackAnnotation:   {true, ""},
 			},
 		},
+		{
+			description:                   "internal load balancer for openstack platform with floating IP",
+			platformStatus:                platformStatus(configv1.OpenStackPlatformType),
+			strategyStatus:                openstackLbWithFloatingIP(operatorv1.InternalLoadBalancer, "1.2.3.4"),
```
gcs278, line 708: Shouldn't this be `strategySpec` since we now use spec as desired, and status is actual? It's a valid test, but probably makes more sense to use spec here.

gcs278 (commented):
Thanks for the updates.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -861,6 +876,23 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
...
+		)
+		if fip := getOpenStackFloatingIPInSpec(ic); fip != "" {
+			wantFloatingIP = fip
+		}
+		if fip := getOpenStackFloatingIPInStatus(ic); fip != "" {
+			haveFloatingIP = fip
+		}
+		// OpenStack CCM does not support updating Service.Spec.LoadBalancerIP after creation, the load balancer will never be updated.
+		if wantFloatingIP != haveFloatingIP {
+			err := fmt.Errorf("The IngressController floatingIP was changed from %q to %q.", haveFloatingIP, wantFloatingIP)
```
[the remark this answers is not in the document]
EmilienM: I'll add the instructions in this PR and a TODO to use your new function once it has merged.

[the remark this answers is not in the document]
gcs278: That's fine, but I expect to land that PR in the next couple weeks, but we aren't gated by FF date for that bug. If you'd like this PR by FF, then you might want to avoid the dependency on that PR.

pkg/operator/controller/ingress/status.go
```diff
@@ -821,6 +824,14 @@ func IngressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
 				}
 			}
 		}
+		if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+			a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil {
+			if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil {
```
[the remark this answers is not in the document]
EmilienM: oops, that's what happens when I do it late :smile:

pkg/operator/controller/ingress/status_test.go
EmilienM: A bad test. Fixing it and adding more of course.

EmilienM:
> Thanks for the updates.

Thanks for the great feedback, I addressed all the comments.

pkg/operator/controller/ingress/load_balancer_service.go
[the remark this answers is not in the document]
EmilienM: Would it be possible to address that comment after FF, since your PR is a bugfix, so I can address that via a bug post FF as well? I want to avoid copying your unmerged code and create complexity. I'll come back to this for sure once your PR has merged.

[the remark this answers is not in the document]
gcs278: I'd prefer to add the message in this PR if possible, as we've considered missing condition messages a bug in the past.

No worries on pulling in my code, instead follow the existing patterns for subnet and EIPs, and I'll clean it up later when I introduce the function. Shouldn't be too bad to implement, but let me know if you have issues.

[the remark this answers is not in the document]
EmilienM: good, I've added the message following subnet and EIPs's patterns.

pkg/operator/controller/ingress/status.go
```diff
@@ -979,6 +988,16 @@ func updateIngressControllerAWSEIPAllocationStatus(ic *operatorv1.IngressControl
...
 }
 
+// updateIngressControllerFloatingIPOpenStackStatus mutates the provided IngressController object to
+// sync its status to the effective floatingIP on the LoadBalancer-type service.
+func updateIngressControllerFloatingIPOpenStackStatus(ic *operatorv1.IngressController, service *corev1.Service) {
+	// ProviderParameters should be initialized by setDefaultProviderParameters
+	// when an IngressController is admitted, so we don't need to initialize here.
+	if fip := getLoadBalancerIPFromService(service); fip != "" {
+		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack.FloatingIP = fip
+	}
```
gcs278, line 998: Now that I think about it more, I think we should have a NIL check here as we do with `updateIngressControllerAWSEIPAllocationStatus` and `updateIngressControllerAWSSubnetStatus`. `setDefaultProviderParameters` does initialize `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack`, but that only happens on first admission, and a user can remove it later via `oc apply`, then this would be a nil pointer exception. We are trying to fix those in the code.

Additionally, I think you don't really need to check if `fip != ""` because string `"" ` and `nil` are the same thing since it's not a pointer.

```suggestion
if ic.Status.EndpointPublishingStrategy != nil &&
		ic.Status.EndpointPublishingStrategy.LoadBalancer != nil &&
		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil {
		ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack.FloatingIP = getLoadBalancerIPFromService(service)
}
```

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -779,6 +820,9 @@ func Test_desiredLoadBalancerService(t *testing.T) {
 				TargetPort: intstr.FromString("https"),
 			}}, svc.Spec.Ports)
 			assert.Equal(t, "None", string(svc.Spec.SessionAffinity))
+			if tc.expectedFloatingIP != "" {
+				assert.Equal(t, tc.expectedFloatingIP, svc.Spec.LoadBalancerIP)
+			}
```
gcs278, line 825: No need to filter out for empty floating IP, that removes coverage for ensuring it doesn't get added for internal load balancers.

```suggestion
assert.Equal(t, tc.expectedFloatingIP, svc.Spec.LoadBalancerIP)
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -468,6 +468,11 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 			effectiveStrategy.LoadBalancer.DNSManagementPolicy = operatorv1.UnmanagedLoadBalancerDNS
 		}
 
+		// OpenStack platform does not support managed DNS for load balancers.
+		if platformStatus.Type == configv1.OpenStackPlatformType && platformStatus.OpenStack != nil {
```
gcs278, line 472: Do we need to check `platformStatus.OpenStack != nil` here? I think it's safer to just check for the Type since we aren't using any fields below it, in the rare scenario that there aren't any `status.openStack` values set.

```suggestion
if platformStatus.Type == configv1.OpenStackPlatformType {
```

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -779,6 +820,7 @@ func Test_desiredLoadBalancerService(t *testing.T) {
 				TargetPort: intstr.FromString("https"),
 			}}, svc.Spec.Ports)
 			assert.Equal(t, "None", string(svc.Spec.SessionAffinity))
+			assert.Equal(t, tc.expectedFloatingIP, svc.Spec.LoadBalancerIP)
 		})
```
gcs278, line 824: Mind adding a test case to `Test_loadBalancerServiceChanged` to cover the fact that updates to `spec.LoadBalancerIP` should NOT cause a change? (now that we are using `spec.LoadBalancerIP`.

```
		{
			description: "if .spec.loadBalancerIP changes",
			mutate: func(svc *corev1.Service) {
				svc.Spec.LoadBalancerIP = "3.4.5.6"
			},
			expect: false,
		},
```

gcs278:
@EmilienM  I appreciate the co-author, but don't feel like you have to give me credit, you did all of the hard work :smile_cat: If you want to keep it, mind using my redhat email? gspence@redhat.com

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -861,6 +876,19 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
 		}
 	}
 
+	if platform.Type == configv1.OpenStackPlatformType {
+		wantFloatingIP := getOpenStackFloatingIPInSpec(ic)
+		haveFloatingIP := getOpenStackFloatingIPInStatus(ic)
+		// OpenStack CCM does not support updating Service.Spec.LoadBalancerIP after creation, the load balancer will never be updated.
+		if wantFloatingIP != haveFloatingIP {
+			// TODO(emilien) rewrite this once PR for OCPBUGS-16728 has been merged with a function that standardizes the revert message.
```
gcs278, line 884: *nit* I think it's okay to drop this. https://github.com/openshift/cluster-ingress-operator/pull/1142 is still under review and may change, so this more of a prediction of the future rather than a concrete TODO, and might end up going stale.

gcs278 (commented):
Whoops, one major thing I forgot, we need to enable `ingress.operator.openshift.io/auto-delete-load-balancer`.

EmilienM:
I don't think the CI failures are related. Before retesting these jobs, I'll let maintainers reviewing it.

Miciah:
Please squash https://github.com/openshift/cluster-ingress-operator/pull/1147/commits/90c5aa185ba65d177a02e3944a41fb5d440aef9f and https://github.com/openshift/cluster-ingress-operator/pull/1147/commits/66d181e6bc420927bb586f6e5e51a48319046d0b so that `make verify test` succeeds on each commit.  

Please add a link to https://issues.redhat.com/browse/OSASINFRA-3642 in the commit message for https://github.com/openshift/cluster-ingress-operator/pull/1147/commits/80bf760ba715300c97fdc5d6188150125154097b.  You can also copy the PR description and paste it into the commit message to provide more context.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -370,8 +370,12 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
 	service.Spec.Selector = controller.IngressControllerDeploymentPodSelector(ci).MatchLabels
 
-	lb := ci.Status.EndpointPublishingStrategy.LoadBalancer
-	isInternal := lb != nil && lb.Scope == operatorv1.InternalLoadBalancer
+	var lbSpec *operatorv1.LoadBalancerStrategy
+	if ci.Spec.EndpointPublishingStrategy != nil {
+		lbSpec = ci.Spec.EndpointPublishingStrategy.LoadBalancer
+	}
+	lbStatus := ci.Status.EndpointPublishingStrategy.LoadBalancer
+	isInternal := lbStatus != nil && lbStatus.Scope == operatorv1.InternalLoadBalancer
```
Miciah, line 378: Thanks for fixing the potential nil-pointer dereference!  Normally, I'd ask for this change to be moved to a separate commit as it is logically separate from the other changes, and renaming variables adds a lot of delta that distracts from the changes that are directly related to implementing the feature.  However, in the interest of time, we can keep this in the same commit if that's easier for you.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -666,6 +681,9 @@ func shouldRecreateLoadBalancer(current, desired *corev1.Service, platform *conf
 	if platform.Type == configv1.AWSPlatformType && !serviceEIPAllocationsEqual(current, desired) {
 		return true, "its eipAllocations changed"
 	}
+	if platform.Type == configv1.OpenStackPlatformType && !serviceLoadBalancerIPsEqual(current, desired) {
```
Miciah, line 684: Do `serviceLoadBalancerIPsEqual` and `getLoadBalancerIPFromService` make the logic any clearer?  To me, they only seem to obfuscate the logic.  
```suggestion
	if platform.Type == configv1.OpenStackPlatformType && current.Spec.LoadBalancerIP != desired.Spec.LoadBalancerIP {
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -480,6 +484,17 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
 			}
+		case configv1.OpenStackPlatformType:
+			// Set a floating IP only if the load balancer scope is external.
+			if !isInternal {
+				if lbSpec != nil && lbSpec.ProviderParameters != nil &&
+					lbSpec.ProviderParameters.Type == operatorv1.OpenStackLoadBalancerProvider &&
+					lbSpec.ProviderParameters.OpenStack != nil {
+					// We know that LoadBalancerIP is deprecated but in this current version of Service, this won't be removed.
+					// It'll probably be removed in Service v2 which is not going to happen in the near future.
+					service.Spec.LoadBalancerIP = lbSpec.ProviderParameters.OpenStack.FloatingIP
```
Miciah, line 495: I went to check whether validation for `service.Spec.LoadBalancerIP` was stricter than validation for `lbSpec.ProviderParameters.OpenStack.FloatingIP` (that is to say, whether `FloatingIP` could have some value that would be invalid for `LoadBalancerIP`), but I cannot find any validation for `LoadBalancerIP` whatsoever!  But I suppose that's exactly the reason why it's being deprecated.  * grin *.  
https://github.com/openshift/cluster-ingress-operator/blob/80bf760ba715300c97fdc5d6188150125154097b/vendor/k8s.io/api/core/v1/types.go#L5421

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -861,6 +879,18 @@ func loadBalancerServiceIsProgressing(ic *operatorv1.IngressController, service
 		}
 	}
 
+	if platform.Type == configv1.OpenStackPlatformType {
+		wantFloatingIP := getOpenStackFloatingIPInSpec(ic)
+		haveFloatingIP := getOpenStackFloatingIPInStatus(ic)
+		// OpenStack CCM does not support updating Service.Spec.LoadBalancerIP after creation, the load balancer will never be updated.
+		if wantFloatingIP != haveFloatingIP {
+			changeMsg := fmt.Sprintf("The IngressController floatingIP was changed from %q to %q.", haveFloatingIP, wantFloatingIP)
+			ocPatchRevertCmd := fmt.Sprintf("oc -n openshift-ingress-operator patch ingresscontrollers/%[1]s --type=merge --patch='{\"spec\":{\"endpointPublishingStrategy\":{\"type\":\"LoadBalancerService\",\"loadBalancer\":{\"providerParameters\":{\"type\":\"OpenStack\",\"openstack\":{\"floatingIP\":\"%[2]s\"}}}}}}'", ic.Name, haveFloatingIP)
```
Miciah, line 888:
```suggestion
			ocPatchRevertCmd := fmt.Sprintf(`oc -n %[1]s patch ingresscontrollers/%[2]s --type=merge --patch='{"spec":{"endpointPublishingStrategy":{"type":"LoadBalancerService","loadBalancer":{"providerParameters":{"type":"OpenStack","openstack":{"floatingIP":"%[3]s"}}}}}}'`, ic.Namespace, ic.Name, haveFloatingIP)
```

pkg/operator/controller/ingress/status.go
```diff
@@ -979,6 +990,16 @@ func updateIngressControllerAWSEIPAllocationStatus(ic *operatorv1.IngressControl
 	}
 }
 
+// updateIngressControllerFloatingIPOpenStackStatus mutates the provided IngressController object to
+// sync its status to the effective floatingIP on the LoadBalancer-type service.
+func updateIngressControllerFloatingIPOpenStackStatus(ic *operatorv1.IngressController, service *corev1.Service) {
+	// FloatingIP should be initialized by setDefaultPublishingStrategy
+	// when an IngressController is admitted, so we don't need to initialize here.
```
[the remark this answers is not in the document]
Miciah: This comment is still present, and it is confusing to me.  

If `setDefaultProviderParameters` initializes `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack`, then why do we need all the nil checks here?

In fact, I believe `setDefaultProviderParameters` does not initialize `ic.Status.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack` unless `ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.Type` set, right?  So this comment is misleading; I believe it should be removed.

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -674,6 +702,19 @@ func Test_desiredLoadBalancerService(t *testing.T) {
 				localWithFallbackAnnotation:   {true, ""},
 			},
 		},
+		{
+			description:                   "internal load balancer for openstack platform with floating IP",
+			platformStatus:                platformStatus(configv1.OpenStackPlatformType),
+			strategySpec:                  openstackLbWithFloatingIP(operatorv1.InternalLoadBalancer, "1.2.3.4"),
+			strategyStatus:                lbs(operatorv1.InternalLoadBalancer),
```
Miciah, line 709: Maybe add a comment to say that floatingip is ignored for internal scope.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -696,6 +696,33 @@ func Test_computeLoadBalancerProgressingStatus(t *testing.T) {
 		return ic
 	}
 
+	loadBalancerIngressControllerWithFloatingIP := func(floatingIPSpec string, floatingIPStatus string) *operatorv1.IngressController {
```
Miciah, line 699:
```suggestion
	loadBalancerIngressControllerWithFloatingIP := func(floatingIPSpec, floatingIPStatus string) *operatorv1.IngressController {
```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -877,6 +914,27 @@ func Test_computeLoadBalancerProgressingStatus(t *testing.T) {
...
+			platformStatus: openstackPlatformStatus,
+			expectStatus:   operatorv1.ConditionFalse,
+		},
+		{
+			name:           "LoadBalancerService, OpenStack floating IP spec and status are not the same",
+			ic:             loadBalancerIngressControllerWithFloatingIP("192.168.0.1", ""),
+			service:        lbServiceLoadBalancerIP("192.168.0.1"),
+			platformStatus: openstackPlatformStatus,
+			expectStatus:   operatorv1.ConditionTrue,
+		},
```
Miciah, line 937: Why do you need `lbServiceLoadBalancerIP`?  `computeLoadBalancerProgressingStatus` doesn't look at the service's `spec.loadBalancerIP`.  You can substitute `lbService` in these test cases, and they still pass.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -666,6 +681,9 @@ func shouldRecreateLoadBalancer(current, desired *corev1.Service, platform *conf
 	if platform.Type == configv1.AWSPlatformType && !serviceEIPAllocationsEqual(current, desired) {
 		return true, "its eipAllocations changed"
 	}
+	if platform.Type == configv1.OpenStackPlatformType && !serviceLoadBalancerIPsEqual(current, desired) {
+		return true, "its load balancer IP changed"
+	}
```
Miciah, line 686: We have not test coverage for `shouldRecreateLoadBalancer`.  As a follow-up, we should add a unit test.

pkg/operator/controller/ingress/load_balancer_service.go
EmilienM: The nil-pointer dereference is linked to the change introduced to this commit since we didn't have the lbSpec before and lbSpec is only used by OpenStack now.
I'm happy to make it a separate commit but when I tried just now, it didn't make much sense to me since this block was really changed for the purpose of this commit.

pkg/operator/controller/ingress/load_balancer_service.go
EmilienM: fine, I just did like other platforms which leverage a function. I agree in this case this is not really needed. I'll remove it.

pkg/operator/controller/ingress/load_balancer_service.go
EmilienM: yeah not ideal :cry:

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Yea I can take that action, but I'm removing it in my PR https://github.com/openshift/cluster-ingress-operator/pull/1142/commits/c5ecad99f07ee972bed670cb3a67400196f0e6f9 actually.

EmilienM: I've added it in an extra commit :+1:

Miciah:
> I don't think the CI failures are related. Before retesting these jobs, I'll let maintainers reviewing it.

[e2e-azure-ovn failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1147/pull-ci-openshift-cluster-ingress-operator-master-e2e-azure-ovn/1857099470299205632) because `[sig-arch][Late] operators should not create watch channels very often` failed:
```text
{  fail [github.com/openshift/origin/test/extended/apiserver/api_requests.go:406]: Expected
    <[]string | len:1, cap:1>: [
        "Operator \"cluster-samples-operator\" produces more watch requests than expected: watchrequestcount=151, upperbound=118, ratio=1.28",
    ]
to be empty
Ginkgo exit error 1: exit with code 1}
```
I have filed [OCPBUGS-44581 CI fails on "[sig-arch][Late] operators should not create watch channels very often"](https://issues.redhat.com/browse/OCPBUGS-44581) for this issue.  

[e2e-aws-ovn-techpreview failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1147/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/1857099469703614464) because `[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info` failed:
```text
{  KubeAPIErrorBudgetBurn was at or above info for at least 1m58s on platformidentification.JobType{Release:"4.18", FromRelease:"", Platform:"aws", Architecture:"amd64", Network:"ovn", Topology:"ha"} (maxAllowed=0s): pending for 16m8s, firing for 1m58s:

Nov 14 17:26:14.852 - 118s  E namespace/openshift-kube-apiserver alert/KubeAPIErrorBudgetBurn alertstate/firing severity/critical ALERTS{alertname="KubeAPIErrorBudgetBurn", alertstate="firing", long="6h", namespace="openshift-kube-apiserver", prometheus="openshift-monitoring/k8s", severity="critical", short="30m"}}
```
I believe [OCPBUGS-42083](https://issues.redhat.com/browse/OCPBUGS-42083) is tracking this issue.

[e2e-hypershift failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1147/pull-ci-openshift-cluster-ingress-operator-master-e2e-hypershift/1857099476942983168) because `TestCreateClusterV2/Main/break-glass-credentials/independent_signers` failed:
```text
{Failed  === RUN   TestCreateClusterV2/Main/break-glass-credentials/independent_signers
    control_plane_pki_operator.go:92: generating new break-glass credentials for more than one signer
    pki.go:75: loading certificate/key pair from disk for signer customer-break-glass, use $REGENERATE_PKI to generate new ones
    control_plane_pki_operator.go:201: creating CSR "o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083" for signer "customer-break-glass", requesting client auth usages
    control_plane_pki_operator.go:211: creating CSRA e2e-clusters-f8jng-example-989zr/o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083 to trigger automatic approval of the CSR
    control_plane_pki_operator.go:218: Successfully waited for CSR "o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083" to be approved and signed in 1s
    control_plane_pki_operator.go:130: validating that the client certificate provides the appropriate access
    control_plane_pki_operator.go:116: amending the existing kubeconfig to use break-glass client certificate credentials
    control_plane_pki_operator.go:133: issuing SSR to identify the subject we are given using the client certificate
    control_plane_pki_operator.go:153: ensuring that the SSR identifies the client certificate as having system:masters power and correct username
    pki.go:75: loading certificate/key pair from disk for signer sre-break-glass, use $REGENERATE_PKI to generate new ones
    control_plane_pki_operator.go:201: creating CSR "zpxsfoj4y5ltot4gswx1k5t01p50a3bcz3um2rtgbi" for signer "sre-break-glass", requesting client auth usages
    control_plane_pki_operator.go:211: creating CSRA e2e-clusters-f8jng-example-989zr/zpxsfoj4y5ltot4gswx1k5t01p50a3bcz3um2rtgbi to trigger automatic approval of the CSR
    control_plane_pki_operator.go:218: Successfully waited for CSR "zpxsfoj4y5ltot4gswx1k5t01p50a3bcz3um2rtgbi" to be approved and signed in 1s
    control_plane_pki_operator.go:130: validating that the client certificate provides the appropriate access
    control_plane_pki_operator.go:116: amending the existing kubeconfig to use break-glass client certificate credentials
    control_plane_pki_operator.go:133: issuing SSR to identify the subject we are given using the client certificate
    control_plane_pki_operator.go:153: ensuring that the SSR identifies the client certificate as having system:masters power and correct username
    control_plane_pki_operator.go:96: revoking the "customer-break-glass" signer
    pki.go:75: loading certificate/key pair from disk for signer customer-break-glass, use $REGENERATE_PKI to generate new ones
    control_plane_pki_operator.go:253: creating CRR e2e-clusters-f8jng-example-989zr/o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083 to trigger signer certificate revocation
    eventually.go:100: Failed to get *v1alpha1.CertificateRevocationRequest: context deadline exceeded
    control_plane_pki_operator.go:260: Failed to wait for CRR e2e-clusters-f8jng-example-989zr/o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083 to complete in 10m0s: context deadline exceeded
    eventually.go:220: observed *v1alpha1.CertificateRevocationRequest e2e-clusters-f8jng-example-989zr/o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083 invalid at RV 78205 after 10m0s: incorrect condition: wanted PreviousCertificatesRevoked=True, got PreviousCertificatesRevoked=False: WaitingForAvailable(Previous signer certificate not yet revoked.)
    control_plane_pki_operator.go:260: *v1alpha1.CertificateRevocationRequest e2e-clusters-f8jng-example-989zr/o7ult40rg5ngulsfotl96hvl0hz0lm1qu7g9x7pz083 conditions:
    control_plane_pki_operator.go:260: LeafCertificatesRegenerated=True: AsExpected(All leaf certificates are re-generated.)
    control_plane_pki_operator.go:260: PreviousCertificatesRevoked=False: WaitingForAvailable(Previous signer certificate not yet revoked.)
    control_plane_pki_operator.go:260: NewCertificatesTrusted=True: AsExpected(New signer certificate e2e-clusters-f8jng-example-989zr/customer-system-admin-signer trusted.)
    control_plane_pki_operator.go:260: RootCertificatesRegenerated=True: AsExpected(Signer certificate e2e-clusters-f8jng-example-989zr/customer-system-admin-signer regenerated.)
    control_plane_pki_operator.go:260: SignerClassValid=True: AsExpected(Signer class "customer-break-glass" known.)
            --- FAIL: TestCreateClusterV2/Main/break-glass-credentials/independent_signers (602.07s)
}
```
I have filed [OCPBUGS-44582](https://issues.redhat.com/browse/OCPBUGS-44582) for this issue.

pkg/operator/controller/ingress/load_balancer_service.go
Miciah: Ah, I see, `lb` was renamed to `lbStatus`, `lbSpec` is new, and the check is only for ``lbSpec`.  It may seem silly, but I still prefer to put small changes such as renaming `lb` to `lbStatus` in a separate commit.  I don't insist on it though.

EmilienM: done !

pkg/operator/controller/ingress/status.go
```diff
@@ -821,6 +824,12 @@ func IngressStatusesEqual(a, b operatorv1.IngressControllerStatus) bool {
 				}
 			}
 		}
+		if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters != nil &&
+			a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil && b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack != nil {
+			if a.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack.FloatingIP != b.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.OpenStack.FloatingIP {
+				return false
```
Miciah, line 830: Hm, this logic does not seem correct.  Try adding the following test case in `Test_IngressStatusesEqual`:
```go
		{
			description: "OpenStack providerParameters present for a but nil for b",
			expected:    false,
			a:           icStatusWithFloatingIP("1.2.3.4"),
			b: operatorv1.IngressControllerStatus{
						ProviderParameters: nil,
					},
				},
			},
		},
```
This test case should pass, but it fails.

pkg/operator/controller/ingress/status.go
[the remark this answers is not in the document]
EmilienM: removed

pkg/operator/controller/ingress/status.go
EmilienM: indeed, I fixed it.

Miciah:
Thanks!
/approve
/lgtm

Miciah:
Thanks again!
/lgtm

EmilienM:
> /test e2e-openstack-operator

the job will fail. I plan to work on that early 2025.

Miciah:
> > /test e2e-openstack-operator
> 
> the job will fail. I plan to work on that early 2025.

It's still worth checking to make sure it fails as expected (same tests failing with similar errors).

Miciah:
https://github.com/openshift/cluster-ingress-operator/pull/1048#issuecomment-2477797570 will give us a baseline for e2e-openstack-operator.

EmilienM:
> [#1048 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/1048#issuecomment-2477797570) will give us a baseline for e2e-openstack-operator.

that or https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1157/pull-ci-openshift-cluster-ingress-operator-master-e2e-openstack-operator/1846256306713792512

lihongan:
@EmilienM Is that possible to do pre-merge test with the PR? If yes, then any requirements to setup the cluster on OSP (Octavia required?), any configuration and function should be checked? Thanks

EmilienM:
> @EmilienM Is that possible to do pre-merge test with the PR? If yes, then any requirements to setup the cluster on OSP (Octavia required?), any configuration and function should be checked? Thanks

Sorry, this just merged. I'll do some testing again today and let people know if something's wrong.

PR #1148, opened 2024-09-23
CFE-1134: Watch infrastructure and update AWS tags

chiragkyal:
The PR introduces the following changes:

- The ingress controller now watches for the `Infrastructure` object changes. This ensures that any modifications to user-defined tags (`platform.AWS.ResourceTags`) trigger an update of the load balancer service.

- Consider the `awsLBAdditionalResourceTags` annotation as a managed annotation. Any changes to user-defined tags in the Infrastructure object will be reflected in this annotation, prompting an update to the load balancer service.

Implements: [CFE-1134](https://issues.redhat.com//browse/CFE-1134)

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -49,6 +49,8 @@ import (
 
 const (
 	controllerName = "ingress_controller"
+	// clusterInfrastructureName is the name of the 'cluster' infrastructure object.
+	clusterInfrastructureName = "cluster"
 )
 
 // TODO: consider moving these to openshift/api
@@ -134,6 +136,12 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.Proxy{}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController))); err != nil {
 		return nil, err
 	}
+	// Watch for changes to infrastructure config to update user defined tags.
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.Infrastructure{}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController),
+		predicate.NewPredicateFuncs(hasName(clusterInfrastructureName)),
+	)); err != nil {
+		return nil, err
+	}
 	return c, nil
 }
 
@@ -185,6 +193,13 @@ func enqueueRequestForOwningIngressController(namespace string) handler.EventHan
 				return []reconcile.Request{}
 			}
 		})
+}
+
+// hasName returns a predicate which checks whether an object has the given name.
+func hasName(name string) func(o client.Object) bool {
+	return func(o client.Object) bool {
+		return o.GetName() == name
+	}
 }
 
 // Config holds all the things necessary for the controller to run.
--- a/pkg/operator/controller/ingress/load_balancer_service.go
+++ b/pkg/operator/controller/ingress/load_balancer_service.go
@@ -254,6 +254,11 @@ var (
 			//
 			// https://cloud.ibm.com/docs/containers?topic=containers-vpc-lbaas
 			iksLBEnableFeaturesAnnotation,
+			// awsLBAdditionalResourceTags annotation is populated
+			// by user tags present in
+			// Status.PlatformStatus.AWS.ResourceTags in the
+			// infrastructure config.
+			awsLBAdditionalResourceTags,
 		)
 
 		// Azure and GCP support switching between internal and external
@@ -751,12 +756,6 @@ func IsServiceInternal(service *corev1.Service) bool {
 	return false
 }
 
-// loadBalancerServiceTagsModified verifies that none of the managedAnnotations have been changed and also the AWS tags annotation
-func loadBalancerServiceTagsModified(current, expected *corev1.Service) (bool, *corev1.Service) {
-	ignoredAnnotations := managedLoadBalancerServiceAnnotations.Union(sets.NewString(awsLBAdditionalResourceTags))
-	return loadBalancerServiceAnnotationsChanged(current, expected, ignoredAnnotations)
-}
-
 // loadBalancerServiceIsUpgradeable returns an error value indicating if the
 // load balancer service is safe to upgrade.  In particular, if the current
 // service matches the desired service, then the service is upgradeable, and the
@@ -773,7 +772,11 @@ func loadBalancerServiceIsUpgradeable(ic *operatorv1.IngressController, deployme
 		return nil
 	}
 
-	changed, updated := loadBalancerServiceTagsModified(current, desired)
+	// Verify that none of the managedAnnotations have been changed by something or someone.
+	// Since the status logic runs after the controller sets the annotations, it checks for
+	// any discrepancy (in case modified) between the desired annotation values of the controller
+	// and the current annotation values.
+	changed, updated := loadBalancerServiceAnnotationsChanged(current, desired, managedLoadBalancerServiceAnnotations)
 	if changed {
 		diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
 		return fmt.Errorf("load balancer service has been modified; changes must be reverted before upgrading: %s", diff)
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -12,7 +12,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/aws/aws-sdk-go/service/ec2"
 	"io"
 	"io/ioutil"
 	"net"
@@ -32,6 +31,7 @@ import (
 	iov1 "github.com/openshift/api/operatoringress/v1"
 	routev1 "github.com/openshift/api/route/v1"
 
+	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
@@ -39,6 +39,7 @@ import (
 	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
@@ -116,6 +117,7 @@ var (
 
 var (
 	kclient           client.Client
+	configClient      *configclientset.Clientset
 	dnsConfig         configv1.DNS
 	infraConfig       configv1.Infrastructure
 	operatorNamespace = operatorcontroller.DefaultOperatorNamespace
@@ -152,6 +154,12 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 	kclient = kubeClient
+
+	configClient, err = configclientset.NewForConfig(kubeConfig)
+	if err != nil {
+		fmt.Printf("failed to create config client: %s\n", err)
+		os.Exit(1)
+	}
 
 	if err := kclient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig); err != nil {
 		fmt.Printf("failed to get DNS config: %v\n", err)
@@ -1289,6 +1297,83 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
 	if err != nil {
 		t.Errorf("failed to observe expected annotations on load balancer service %s: %v", controller.LoadBalancerServiceName(ic), err)
 	}
+}
+
+// TestAWSResourceTagsChanged tests the functionality of updating AWS resource tags
+// in the infrastructure configuration and validates that the expected
+// awsLBAdditionalResourceTags is set correctly on the
+// loadBalancer service associated with the default Ingress Controller.
+//
+// This test is a serial test because it modifies the cluster infrastructure config and
+// therefore should not run in parallel with other tests.
+func TestAWSResourceTagsChanged(t *testing.T) {
+	if infraConfig.Status.Platform != "AWS" {
+		t.Skipf("test skipped on platform %q", infraConfig.Status.Platform)
+	}
+	if err := waitForIngressControllerCondition(t, kclient, 10*time.Second, defaultName, defaultAvailableConditions...); err != nil {
+		t.Errorf("did not get expected conditions: %v", err)
+	}
+	defaultIC := &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: defaultName.Namespace,
+			Name:      defaultName.Name,
+		},
+	}
+	awsLBAdditionalResourceTags := "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"
+
+	// Save a copy of the original infraConfig, to revert changes before exiting.
+	originalInfra := infraConfig.DeepCopy()
+	t.Cleanup(func() {
+		err := updateInfrastructureConfigStatusWithRetryOnConflict(t, 5*time.Minute, configClient, func(infra *configv1.Infrastructure) *configv1.Infrastructure {
+			infra.Status = originalInfra.Status
+			return infra
+		})
+		if err != nil {
+			t.Logf("Unable to remove changes from the infraConfig, possible corruption of test environment: %v", err)
+		}
+	})
+
+	initialTags := []configv1.AWSResourceTag{
+		{Key: "Key1", Value: "Value1"},
+		{Key: "Key2", Value: "Value2"},
+	}
+	t.Logf("Updating AWS ResourceTags in the cluster infrastructure config: %v", initialTags)
+	err := updateInfrastructureConfigStatusWithRetryOnConflict(t, 5*time.Minute, configClient, func(infra *configv1.Infrastructure) *configv1.Infrastructure {
+		if infra.Status.PlatformStatus == nil {
+			infra.Status.PlatformStatus = &configv1.PlatformStatus{}
+		}
+		if infra.Status.PlatformStatus.AWS == nil {
+			infra.Status.PlatformStatus.AWS = &configv1.AWSPlatformStatus{}
+		}
+		infra.Status.PlatformStatus.AWS.ResourceTags = initialTags
+		return infra
+	})
+	if err != nil {
+		t.Errorf("failed to update infrastructure status: %v", err)
+	}
+
+	// Check awsLBAdditionalResourceTags annotation with initial tags.
+	expectedTags := "Key1=Value1,Key2=Value2"
+	t.Logf("Validating the %s annotation for the load balancer service of the default ingresscontroller", awsLBAdditionalResourceTags)
+	assertLoadBalancerServiceAnnotationWithPollImmediate(t, kclient, defaultIC, awsLBAdditionalResourceTags, expectedTags)
+
+	// Update the status again, removing one tag.
+	updatedTags := []configv1.AWSResourceTag{
+		{Key: "Key1", Value: "Value1"},
+	}
+	t.Logf("Updating AWS ResourceTags in the cluster infrastructure config: %v", updatedTags)
+	err = updateInfrastructureConfigStatusWithRetryOnConflict(t, 5*time.Minute, configClient, func(infra *configv1.Infrastructure) *configv1.Infrastructure {
+		infra.Status.PlatformStatus.AWS.ResourceTags = updatedTags
+		return infra
+	})
+	if err != nil {
+		t.Errorf("failed to update infrastructure status: %v", err)
+	}
+
+	// Check awsLBAdditionalResourceTags annotation with updated tags.
+	expectedTags = "Key1=Value1"
+	t.Logf("Validating the %s annotation for the load balancer service of the default ingresscontroller", awsLBAdditionalResourceTags)
+	assertLoadBalancerServiceAnnotationWithPollImmediate(t, kclient, defaultIC, awsLBAdditionalResourceTags, expectedTags)
 }
 
 func TestAWSLBTypeChange(t *testing.T) {
@@ -4320,6 +4405,29 @@ func assertServiceAnnotation(t *testing.T, serviceName types.NamespacedName, ann
 		t.Errorf("service %s is missing expected %s=%s annotation: %v", serviceName, annotationKey, expectedValue, service.Annotations)
 	case expected && expectedValue != actualValue:
 		t.Errorf("expected service %[1]s to have annotation %[2]s=%[3]s, found %[2]s=%[4]s", serviceName, annotationKey, expectedValue, actualValue)
+	}
+}
+
+// assertLoadBalancerServiceAnnotationWithPollImmediate checks if the specified annotation on the
+// LoadBalancer Service of the given IngressController matches the expected value.
+func assertLoadBalancerServiceAnnotationWithPollImmediate(t *testing.T, kclient client.Client, ic *operatorv1.IngressController, annotationKey, expectedValue string) {
+	err := wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 5*time.Minute, true, func(ctx context.Context) (bool, error) {
+		service := &corev1.Service{}
+		if err := kclient.Get(ctx, controller.LoadBalancerServiceName(ic), service); err != nil {
+			t.Logf("failed to get service %s: %v, retrying...", controller.LoadBalancerServiceName(ic), err)
+			return false, nil
+		}
+		if actualValue, ok := service.Annotations[annotationKey]; !ok {
+			t.Logf("load balancer has no %q annotation yet: %v, retrying...", annotationKey, service.Annotations)
+			return false, nil
+		} else if actualValue != expectedValue {
+			t.Logf("expected %s, found %s", expectedValue, actualValue)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("timed out waiting for the %s annotation to be updated: %v", annotationKey, err)
 	}
 }
 
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -17,6 +17,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	configclientset "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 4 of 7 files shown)


Miciah (commented):
Could you add an E2E test?  (I don't know whether an E2E test can update the `ResourceTags` in the infrastructure config status.)

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,6 +136,12 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.Proxy{}, handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController))); err != nil {
 		return nil, err
 	}
+	// Watch for changes to infrastructure config to update user defined tags
```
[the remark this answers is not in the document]
chiragkyal: Updated, thanks!

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -254,6 +254,8 @@ var (
 			//
 			// https://cloud.ibm.com/docs/containers?topic=containers-vpc-lbaas
 			iksLBEnableFeaturesAnnotation,
+			// awsLBAdditionalResourceTags annotation is populated by user tags present in `platform.AWS.ResourceTags`
```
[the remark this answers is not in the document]
chiragkyal: Updated, thanks.

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -751,9 +753,10 @@ func IsServiceInternal(service *corev1.Service) bool {
 	return false
 }
 
-// loadBalancerServiceTagsModified verifies that none of the managedAnnotations have been changed and also the AWS tags annotation
+// loadBalancerServiceTagsModified verifies that none of the managedAnnotations except awsLBAdditionalResourceTags have been changed
 func loadBalancerServiceTagsModified(current, expected *corev1.Service) (bool, *corev1.Service) {
-	ignoredAnnotations := managedLoadBalancerServiceAnnotations.Union(sets.NewString(awsLBAdditionalResourceTags))
+	ignoredAnnotations := managedLoadBalancerServiceAnnotations.Clone()
+	ignoredAnnotations.Delete(awsLBAdditionalResourceTags)
```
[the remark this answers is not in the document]
chiragkyal: Thank you for the detailed explanation on how the status logic works and how it sets `Upgradeable=False`, as well as the exception that existed with `awsLBAdditionalResourceTags` before this PR. Earlier, I was under the impression that the status logic would still set `Upgradeable=False` even if `awsLBAdditionalResourceTags` was updated by the controller.


> So now that you are making the awsLBAdditionalResourceTags annotation a managed annotation, don't we still want to set Upgradeable=False if the annotation value doesn't match the expected value?

Since `awsLBAdditionalResourceTags` will now be managed by the controller, and we still want to set `Upgradeable=False` if it’s updated by something other than the ingress controller, it does indeed make sense to use `managedLoadBalancerServiceAnnotations` directly in this logic. This way, the status logic will behave consistently for managed annotations when any discrepancy is observed.

I've removed the `loadBalancerServiceTagsModified()` function and used `loadBalancerServiceAnnotationsChanged()` directly inside `loadBalancerServiceIsUpgradeable()` and also added some comments for clearer understanding of the flow.

chiragkyal:
> Could you add an E2E test? (I don't know whether an E2E test can update the ResourceTags in the infrastructure config status.)

I need to try it to see if updating the infrastructure config status is possible through E2E. I used  `kubectl edit-status infrastructure cluster` command to manually update the status while doing some testing. Need to check if similar thing can be done through E2E. 

Having said that, do you think we should get QE sign-off for this PR?

Miciah:
> I need to try it to see if updating the infrastructure config status is possible through E2E. I used `kubectl edit-status infrastructure cluster` command to manually update the status while doing some testing. Need to check if similar thing can be done through E2E.

The tests in `test/e2e/configurable_route_test.go` update the _ingress_ cluster config status; maybe that's useful as a reference or precedent for adding a similar E2E test to this PR.  

What I'm wondering is whether the test can update the infrastructure config status without some other controller stomping the changes, and whether there could be other reasons specific to the infrastructures resource or `resourceTags` API field why an E2E test should not or cannot update it.  

> Having said that, do you think we should get QE sign-off for this PR?

As a general matter, we should have QE sign-off for this PR.  QE might prefer to do pre-merge testing as well.  

Is day2 tags support being handled by a specific group of QA engineers, or are the the QA engineers for each affected component responsible for testing the feature?  Cc: @lihongan.

lihongan:
Did pre-merge test on standalone OCP cluster and it can add new tags key/value pair and update existing tags value, but cannot delete the user added tags. see
```
$ oc get clusterversion
version   4.18.0-0.test-2024-10-18-013151-ci-ln-4jn9mbt-latest   True        False         59m     Cluster version is 4.18.0-0.test-2024-10-18-013151-ci-ln-4jn9mbt-latest

$ kubectl edit infrastructure cluster --subresource='status'

$ oc get infrastructures.config.openshift.io cluster -oyaml | yq .status.platformStatus
aws:
  region: us-east-2
  resourceTags:
    - key: Owner
      value: QE
    - key: CaseID
      value: OCP-76984
type: AWS

$ aws elb describe-tags --load-balancer-name a8a32335a6697415e9d55bafce2e6060 --output yaml
TagDescriptions:
- LoadBalancerName: a8a32335a6697415e9d55bafce2e6060
  Tags:
  - Key: kubernetes.io/service-name
    Value: openshift-ingress/router-default
  - Key: Owner
    Value: QE
  - Key: CaseID
    Value: OCP-76984
  - Key: kubernetes.io/cluster/hongli-tags-z7s5b
    Value: owned

// edit status and remove one tags pair and update one tag value
$ oc get infrastructures.config.openshift.io cluster -oyaml | yq .status.platformStatus
aws:
  region: us-east-2
  resourceTags:
    - key: Owner
      value: None
type: AWS

$ aws elb describe-tags --load-balancer-name a8a32335a6697415e9d55bafce2e6060 --output yaml
TagDescriptions:
- LoadBalancerName: a8a32335a6697415e9d55bafce2e6060
  Tags:
  - Key: kubernetes.io/service-name
    Value: openshift-ingress/router-default
  - Key: Owner
    Value: None
  - Key: CaseID
    Value: OCP-76984
  - Key: kubernetes.io/cluster/hongli-tags-z7s5b
    Value: owned
```
@chiragkyal please confirm if that's expected. 
And will try to check HCP later.

lihongan:
And it also works for NLB custom ingresscontroler 
```
$ oc get infrastructures.config.openshift.io cluster -oyaml | yq .status.platformStatus
aws:
  region: us-east-2
  resourceTags:
    - key: caseid
      value: ocp-88888
type: AWS

$ aws elbv2 describe-tags --resource-arns arn:aws:elasticloadbalancing:us-east-2:******:loadbalancer/net/a84a56f6f236b4f77b0032aaa037882d/9b00573907e8e952 --output yaml
TagDescriptions:
- ResourceArn: arn:aws:elasticloadbalancing:us-east-2:******:loadbalancer/net/a84a56f6f236b4f77b0032aaa037882d/9b00573907e8e952
  Tags:
  - Key: kubernetes.io/service-name
    Value: openshift-ingress/router-nlb
  - Key: caseid
    Value: ocp-88888
  - Key: kubernetes.io/cluster/hongli-tags-z7s5b
    Value: owned
```

lihongan:
looks https://github.com/kubernetes/kubernetes/pull/96939 is for NLB fix but closed

chiragkyal:
> @chiragkyal please confirm if that's expected.

It looks like the tags are getting merged with the existing AWS Resource tags. I think we can only control the `aws-load-balancer-additional-resource-tags`  annotation values; the next step depends on how this annotation is getting read by its consumer. @Miciah is there a way we can control this behavior ?

chiragkyal:
> looks https://github.com/kubernetes/kubernetes/pull/96939 is for NLB fix but closed

Okay, it looks like an existing issue then.

Miciah:
Yeah, it looks like the behavior would need to be corrected in cloud-provider-aws.  However, the upstream maintainers' usual response is to suggest using aws-load-balancer-controller instead: <https://github.com/kubernetes/cloud-provider-aws/issues/113#issuecomment-906630865>.  This has been an issue with other features that depend on cloud-provider-aws support as well.  

Absent support in cloud-provider-aws, these are the options that come to mind:

- Change OpenShift to use aws-load-balancer-controller instead of cloud-provider-aws for NLBs.  This would be a major change, and it isn't one we could make in time for 4.18.  
- Delete and recreate the NLB for tags updates.  This would disrupt traffic, so it isn't really practical.    
- Update tags directly from cluster-ingress-operator.  We would need to make sure we had the necessary permissions in the CredentialsRequest and then write the client code, but then I _think_ we could have `ensureLoadBalancerService` handle the AWS resource tags updates itself, and it shouldn't interfere with cloud-provider-aws's operation as cloud-provider-aws evidently isn't reconciling tags anyway.  I don't love this option, but it seems like the only practical one.  

Is handling updates a hard requirement?  https://github.com/openshift/enhancements/pull/1700 is ambiguous:

> If the userTags field is changed post-install, there is no guarantee about how an in-cluster operator will respond to the change. Some operators may reconcile the change and change tags on the AWS resource. Some operators may ignore the change. However, if tags are removed from userTags, the tag will not be removed from the AWS resource.

But also:

> For the resources created and managed by hosted control plane, cluster api provider for aws reconciles the user tags on AWS resources. The hosted control plane updates the `infrastructure.config.openshift.io` resource to reflect new tags in `resourceTags`. The OpenShift operators, both core and non-core (managed by RedHat), reconcile the respective AWS resources created and managed by them. 

The first blockquote was present from the earlier implementation of the feature, so maybe it was supposed to be removed.  I'll follow up on https://github.com/openshift/enhancements/pull/1700.

Miciah:
I started a discussion here:  https://github.com/openshift/enhancements/pull/1700/files#r1806975007

chiragkyal:
>Yeah, it looks like the behavior would need to be corrected in cloud-provider-aws. However, the upstream maintainers' usual response is to suggest using aws-load-balancer-controller instead: https://github.com/kubernetes/cloud-provider-aws/issues/113#issuecomment-906630865. This has been an issue with other features that depend on cloud-provider-aws support as well.

Okay. The lack of tags support for NLB in `cloud-provider-aws` is an existing issue that will require significant changes and discussions. I _don't believe_  that should be part of this feature implementation. I think we need to document this exception somewhere until it is resolved. 

/cc @TrilokGeer

lihongan:
@chiragkyal  Also did pre-mrege testing with HyperShift and got same results.

Miciah:
> Okay. The lack of tags reconciliation support for NLB in `cloud-provider-aws` is an existing issue that will require significant changes and discussions. I _don't believe_ that should be part of this feature implementation. I think we need to document this exception somewhere until it is resolved.

Do we need confirmation that this exception is acceptable?  Would the exception be documented in https://github.com/openshift/enhancements/pull/1700?  

Aside from the matter of reconciling tags for NLBs managed by cloud-controller-manager/cloud-provider-aws, is this PR ready for review?

chiragkyal:
>Do we need confirmation that this exception is acceptable? Would the exception be documented in https://github.com/openshift/enhancements/pull/1700?

I had a discussion with @TrilokGeer on this. He'll check with other stakeholders regarding this issue and once accepted we can document the exception in the EP.

>Aside from the matter of reconciling tags for NLBs managed by cloud-controller-manager/cloud-provider-aws, is this PR ready for review?

I believe so. And if I recall correctly, we need another review from @candita or @alebedev87 before final approval.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1301,74 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
 	}
 }
 
+// TestAWSResourceTagsChanged tests the functionality of updating AWS resource tags
+// in the infrastructure configuration and validates that the expected
+// awsLBAdditionalResourceTags is set correctly on the
+// loadBalancer service associated with the default Ingress Controller.
+func TestAWSResourceTagsChanged(t *testing.T) {
+	t.Parallel()
```
Miciah, line 1309: Generally, any test that modifies cluster config should be a serial test.  I think that that general rule holds true here.  We want to avoid nondeterministic test results should changing the `aws-load-balancer-additional-resource-tags` annotation somehow affects other tests.

test/e2e/operator_test.go
```diff
@@ -25,13 +24,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/client-go/util/retry"
```
Miciah, line 28: Please group `"k8s.io/client-go/util/retry"` with the other `k8s.io/client-go` imports, and please group `"github.com/aws/aws-sdk-go/service/ec2"` with `"github.com/aws/aws-sdk-go/aws/endpoints"`.

test/e2e/operator_test.go
```diff
@@ -25,13 +24,17 @@ import (
...
+
 	"gopkg.in/yaml.v2"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 	routev1 "github.com/openshift/api/route/v1"
 
+	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
```
Miciah, line 38: Hm, I suppose grouping the `github.com/openshift/client-go` import with the `github.com/openshift/cluster-ingress-operator` imports is fine for now, but at some point, we do want to formalize and enforce import grouping rules for this repository.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1301,74 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+
+	awsLBAdditionalResourceTags := "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"
+	expectedTags := "Key1=Value1,Key2=Value2"
+	defaultIC := &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: defaultName.Namespace,
+			Name:      defaultName.Name,
+		},
+	}
+	// validate awsLBAdditionalResourceTags for load balancer service of default ingress controller.
```
Miciah, line 1351: Again, if this is worth commenting, it's probably worth logging.
```suggestion
	t.Logf("Validating the %s annotation for the load balancer service of the default ingresscontroller.", awsLBAdditionalResourceTags)
```

test/e2e/operator_test.go
```diff
@@ -1291,6 +1301,74 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+			return false, nil
+		} else if actualTags != expectedTags {
+			t.Logf("expected %s, found %s", expectedTags, actualTags)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("timed out waiting for the %s annotation to be updated: %v", awsLBAdditionalResourceTags, err)
+	}
```
Miciah, line 1369: Can you use `assertServiceAnnotation` here?

chiragkyal: `assertServiceAnnotation` does not use `wait.PollImmediate`, and I think we should wait for the controller to update the annotation using the polling loop. I checked other tests like `TestInternalLoadBalancer`  and `TestAWSLBTypeChange` which are also using similar logic. If it's safe to use `assertServiceAnnotation` here, then we can update it.

test/e2e/operator_test.go
chiragkyal: Sure, I've replaced the comment with a log line.

test/e2e/operator_test.go
chiragkyal: Thanks for the suggestion. I agreed that we should run this test in serial. I've removed `t.Parallel()` and added a go-doc for the function related to this.

test/e2e/operator_test.go
chiragkyal: Thanks, updated and grouped.

test/e2e/operator_test.go
chiragkyal: Yeah, it would be better to add this as part of the `verify`  job.

test/e2e/operator_test.go
Miciah: Thanks!  You also need to update `TestAll`.

test/e2e/operator_test.go
Miciah: As we discussed on the call, I'm fine with not using `assertServiceAnnotation`.  You could add a code comment explaining why you cannot use `assertServiceAnnotation`, but I'm fine with the test in any case.

chiragkyal: Sure, I've added the following code comment explaining the reason
```go
	// Use a polling loop instead of assertServiceAnnotation since
	// the operator might not update the annotation immediately.
```

test/e2e/operator_test.go
chiragkyal: Ah! right. I've updated `TestAll` and moved `TestAWSResourceTagsChanged` into the serial section.

chiragkyal:
Looks like there was [some issue](https://redhat-internal.slack.com/archives/CFUGK0K9R/p1729867234347449) with `build05`, causing the jobs to fail. 
Moreover, the [latest change](https://github.com/openshift/cluster-ingress-operator/compare/16b0e86f8f83487bea30105a2a8f140cb75f5d0e..a5560176abe9d448630ff5f515eccc0850e70c67) does not seem to be a breaking change to me. 

/retest

Miciah:
/approve
/lgtm
/hold
for Candace.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1300,79 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+
+	if infraConfig.Status.Platform != "AWS" {
+		t.Skipf("test skipped on platform %q", infraConfig.Status.Platform)
+	}
+	if err := waitForIngressControllerCondition(t, kclient, 10*time.Second, defaultName, defaultAvailableConditions...); err != nil {
+		t.Errorf("did not get expected conditions: %v", err)
+	}
+
+	t.Log("Updating AWS ResourceTags in the cluster infrastructure config")
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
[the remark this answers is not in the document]
chiragkyal: Because we want to update the status of the Infrastructure config, instead of spec. I've moved the status update logic to a new function `updateInfrastructureConfigStatusWithRetryOnConflict` for better understanding.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1300,79 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+			{Key: "Key2", Value: "Value2"},
+		}
+		_, updateErr := configClient.ConfigV1().Infrastructures().UpdateStatus(context.TODO(), infra, metav1.UpdateOptions{})
+		return updateErr
+	})
+
+	if retryErr != nil {
+		t.Errorf("failed to update infrastructure status: %v", retryErr)
+	}
+
```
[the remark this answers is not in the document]
chiragkyal: Thanks for the suggestion. I added `t.Cleanup` to revert the infra config status, which should, in turn reset the annotation on the default ingress controller.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1300,79 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
 	}
 }
 
+// TestAWSResourceTagsChanged tests the functionality of updating AWS resource tags
```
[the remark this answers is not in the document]
chiragkyal: > like deleting a user-defined tag and adding a user-defined tag?

The test already covers adding a user-defined tag. 

However, updating the infra status again to remove certain tag is possible, which will update the annotation as well, but the tag won't be removed from the AWS resource itself, and this is an expected behaviour for cloud-provider-aws. See https://github.com/openshift/cluster-ingress-operator/pull/1148#issuecomment-2421219651 for more details. 

I've extended the test to cover this scenario of tag removal and annotation update in the latest changes. Hope it covers the acceptance criteria.

test/e2e/operator_test.go
```diff
@@ -1291,6 +1299,80 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+	defaultIC := &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: defaultName.Namespace,
+			Name:      defaultName.Name,
+		},
+	}
+	awsLBAdditionalResourceTags := "service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags"
+
+	// Revert to original status
+	originalInfraStatus := infraConfig.Status
```
candita, line 1325:
```suggestion
	// Save a copy of the original infraConfig.Status, to revert changes before exiting.
	originalInfraStatus := infraConfig.Status.DeepCopy()
```

test/e2e/operator_test.go
```diff
@@ -1291,6 +1299,80 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+	t.Cleanup(func() {
+		updateInfrastructureConfigStatusWithRetryOnConflict(configClient, func(infra *configv1.Infrastructure) *configv1.Infrastructure {
```
candita, line 1327:
```suggestion
		err := updateInfrastructureConfigStatusWithRetryOnConflict(configClient, func(infra *configv1.Infrastructure) *configv1.Infrastructure {
```

test/e2e/operator_test.go
```diff
@@ -1291,6 +1299,80 @@ func TestInternalLoadBalancerGlobalAccessGCP(t *testing.T) {
...
+			infra.Status = originalInfraStatus
+			return infra
+		})
+	})
```
candita, line 1331: Something like this:
```suggestion
	}
        if err != nil {
             t.Logf("Unable to remove changes to the infraConfig: %v, possible corruption of test environment", err)
        }
	)
```

test/e2e/operator_test.go
```diff
@@ -4323,6 +4405,29 @@ func assertServiceAnnotation(t *testing.T, serviceName types.NamespacedName, ann
 	}
 }
 
+// assertLoadBalancerServiceAnnotationWithPollImmediate checks if the specified annotation on the
+// LoadBalancer Service of the given IngressController matches the expected value.
+func assertLoadBalancerServiceAnnotationWithPollImmediate(t *testing.T, kclient client.Client, ic *operatorv1.IngressController, annotationKey, expectedValue string) {
+	err := wait.PollImmediate(5*time.Second, 5*time.Minute, func() (bool, error) {
```
candita, line 4411: We have started replacing the use of `wait.PollImmediate` with 
`wait.PollUntilContextTimeout(context.Background()`, as used in https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L3189.  Please use the updated function here, and any new code that requires polled waiting.

test/e2e/operator_test.go
```diff
@@ -4323,6 +4405,29 @@ func assertServiceAnnotation(t *testing.T, serviceName types.NamespacedName, ann
...
+		service := &corev1.Service{}
+		if err := kclient.Get(context.Background(), controller.LoadBalancerServiceName(ic), service); err != nil {
+			t.Logf("failed to get service %s: %v", controller.LoadBalancerServiceName(ic), err)
```
candita, line 4414:
```suggestion
			t.Logf("failed to get service %s: %v, retrying...", controller.LoadBalancerServiceName(ic), err)
```

test/e2e/operator_test.go
```diff
@@ -4323,6 +4405,29 @@ func assertServiceAnnotation(t *testing.T, serviceName types.NamespacedName, ann
...
+			return false, nil
+		}
+		if actualValue, ok := service.Annotations[annotationKey]; !ok {
+			t.Logf("load balancer has no %q annotation: %v", annotationKey, service.Annotations)
```
candita, line 4418:
```suggestion
			t.Logf("load balancer has no %q annotation yet: %v, retrying...", annotationKey, service.Annotations)
```

test/e2e/operator_test.go
```diff
@@ -4323,6 +4405,29 @@ func assertServiceAnnotation(t *testing.T, serviceName types.NamespacedName, ann
...
+			return false, nil
+		} else if actualValue != expectedValue {
+			t.Logf("expected %s, found %s", expectedValue, actualValue)
+			return false, nil
```
candita, line 4422: I expect that we don't want to keep trying, after we found an unexpected value.  Or would we expect it change after this?
```suggestion
			return false, fmt.Errorf("expected %s, found %s", expectedValue, actualValue)
```

test/e2e/util_test.go
```diff
@@ -595,6 +597,25 @@ func updateInfrastructureConfigSpecWithRetryOnConflict(t *testing.T, name types.
 	})
 }
 
+// updateInfrastructureStatus updates the Infrastructure status by applying
+// the given update function to the current Infrastructure object.
+func updateInfrastructureConfigStatusWithRetryOnConflict(configClient *configclientset.Clientset, updateFunc func(*configv1.Infrastructure) *configv1.Infrastructure) error {
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
candita, line 603: The `retry.DefaultRetry` is only 10ms https://pkg.go.dev/k8s.io/client-go/util/retry#pkg-variables.  Why not use `wait.PollUntilContextTimeout(context.Background()` and allow it to loop on conflict for a configured amount of time?

test/e2e/operator_test.go
chiragkyal: Thanks, addressed the suggestion.

test/e2e/operator_test.go
chiragkyal: Thanks, updated

test/e2e/operator_test.go
chiragkyal: Sure, updated.

test/e2e/operator_test.go
chiragkyal: No problem and thanks for the suggestion. I have replaced `wait.PollImmediate` with `wait.PollUntilContextTimeout` to be consistent.

test/e2e/operator_test.go
chiragkyal: Done

test/e2e/operator_test.go
chiragkyal: Updated

test/e2e/operator_test.go
chiragkyal: I think we need to keep trying here because the annotation value might not get updated immediately after the infra status is updated.

test/e2e/util_test.go
chiragkyal: Sure, we can use `wait.PollUntilContextTimeout(context.Background()` here as well. I've updated the logic in the latest changes. Thanks!

PR #1151, opened 2024-10-01
HOSTEDCP-2031: Use Client Certificate Authentication for ARO HCP deployments

bryan-cox:
Use Client Certificate Authentication for ARO HCP deployments. HyperShift will pass the needed environment variables for this authentication method: ARO_HCP_MI_CLIENT_ID, ARO_HCP_TENANT_ID, and ARO_HCP_CLIENT_CERTIFICATE_PATH.

### DIFF
--- a/pkg/dns/azure/client/auth.go
+++ b/pkg/dns/azure/client/auth.go
@@ -2,8 +2,10 @@ package client
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
@@ -11,7 +13,11 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/jongio/azidext/go/azidext"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/util/filewatcher"
 )
+
+var watchCertificateFileOnce sync.Once
 
 func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
 	var cloudConfig cloud.Configuration
@@ -70,18 +76,43 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
 	}
 
 	var cred azcore.TokenCredential
-	// Managed Identity Override for ARO HCP
+	// Managed Identity Override for ARO HCP. In ARO HCP, we ignore the values provided for AZURE_TENANT_ID and
+	// AZURE_CLIENT_ID and use ARO_HCP_TENANT_ID and ARO_HCP_MI_CLIENT_ID instead.
 	managedIdentityClientID := os.Getenv("ARO_HCP_MI_CLIENT_ID")
 	if managedIdentityClientID != "" {
-		options := azidentity.ManagedIdentityCredentialOptions{
+		options := &azidentity.ClientCertificateCredentialOptions{
 			ClientOptions: azcore.ClientOptions{
 				Cloud: cloudConfig,
 			},
-			ID: azidentity.ClientID(managedIdentityClientID),
+			SendCertificateChain: true,
 		}
 
-		var err error
-		cred, err = azidentity.NewManagedIdentityCredential(&options)
+		tenantID := os.Getenv("ARO_HCP_TENANT_ID")
+		certPath := os.Getenv("ARO_HCP_CLIENT_CERTIFICATE_PATH")
+
+		certData, err := os.ReadFile(certPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read certificate file %q: %w", certPath, err)
+		}
+
+		certs, key, err := azidentity.ParseCertificates(certData, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate data %q: %w", certPath, err)
+		}
+
+		// Watch the certificate for changes; if the certificate changes, the pod will be restarted.
+		// This starts only one occurrence of the file watcher, which watches the file, certPath.
+		var fileWatcherError error
+		watchCertificateFileOnce.Do(func() {
+			if err = filewatcher.WatchFileForChanges(certPath); err != nil {
+				fileWatcherError = err
+			}
+		})
+		if fileWatcherError != nil {
+			return nil, fmt.Errorf("failed to watch certificate file %q: %w", certPath, fileWatcherError)
+		}
+
+		cred, err = azidentity.NewClientCertificateCredential(tenantID, managedIdentityClientID, certs, key, options)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/util/filewatcher/filewatcher.go
+++ b/pkg/util/filewatcher/filewatcher.go
@@ -1,0 +1,65 @@
+package filewatcher
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+)
+
+var log = logf.Logger.WithName("filewatcher")
+
+// WatchFileForChanges watches the file, fileToWatch, for changes. If the file contents have changed, the pod this
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
+	log.Info("Starting the file change watcher")
+
+	// Update the file path to watch in case this is a symlink
+	fileToWatch, err := filepath.EvalSymlinks(fileToWatch)
+	if err != nil {
+		return err
+	}
+	log.Info("Watching file", "file", fileToWatch)
+
+	// Start the file watcher to monitor file changes
+	go func() {
+		if err := checkForFileChanges(fileToWatch); err != nil {
+			log.Error(err, "Error checking for file changes")
+		}
+	}()
+
+	return nil
+}
+
+// checkForFileChanges starts a new file watcher. If the file is changed, the pod running this function will exit.
+func checkForFileChanges(path string) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if ok && (event.Has(fsnotify.Write) || event.Has(fsnotify.Chmod) || event.Has(fsnotify.Remove)) {
+					log.Info("file was modified, exiting...", "file", event.Name)
+					os.Exit(0)
+				}
+			case err, ok := <-watcher.Errors:
+				if ok {
+					log.Error(err, "file watcher error")
+				} else {
+					log.Error(err, "failed to retrieve watcher error")
+				}
+			}
+		}
+	}()
+
+	if err = watcher.Add(path); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/aws/aws-sdk-go v1.38.49
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/florianl/go-nfqueue v1.3.2
+	github.com/fsnotify/fsnotify v1.7.0
 	github.com/go-logr/logr v1.4.2
 	github.com/go-logr/zapr v1.3.0
 	github.com/google/go-cmp v0.6.0
@@ -70,7 +71,6 @@ require (
 	github.com/evanphx/json-patch v5.7.0+incompatible // indirect
 	github.com/evanphx/json-patch/v5 v5.9.0 // indirect
 	github.com/fatih/color v1.16.0 // indirect
-	github.com/fsnotify/fsnotify v1.7.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
 	github.com/gabriel-vasile/mimetype v1.4.3 // indirect
 	github.com/go-openapi/errors v0.21.0 // indirect


pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +36,14 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
-
-	// Fallback to using client secret from env variable if not set.
-	if strings.TrimSpace(config.ClientSecret) == "" {
-		config.ClientSecret = os.Getenv("AZURE_CLIENT_SECRET")
-		// Skip validation; fallback to token (below) if env variable is also not set.
-	}
-
-	// Fallback to using federated token file from env variable if not set.
-	if strings.TrimSpace(config.FederatedTokenFile) == "" {
-		config.FederatedTokenFile = os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
```
bryan-cox, line 65: I'll need to add this back if this is the path we take

pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +38,48 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
-	if strings.TrimSpace(config.TenantID) == "" {
-		config.TenantID = os.Getenv("AZURE_TENANT_ID")
-		if strings.TrimSpace(config.TenantID) == "" {
-			return nil, errors.New("empty tenant ID")
+	// If the environment variable is set for any of the following, use the environment variable over any configuration
+	// value: AZURE_TENANT_ID, AZURE_CLIENT_ID, and AZURE_CLIENT_SECRET. The Azure SDK for Go function,
+	// NewDefaultAzureCredential, checks environment variables first.
+	if os.Getenv("AZURE_TENANT_ID") == "" {
+		err := os.Setenv("AZURE_TENANT_ID", config.TenantID)
+		if err != nil {
```
Miciah, line 46: The following is more idiomatic since you don't need the return value outside the `if` statement:
```go
		if err := os.Setenv("AZURE_TENANT_ID", config.TenantID); err != nil {
```
Same comment applies to other places where you add `err := ...; if err != nil`.

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
 	}
 	return scope
 }
+
+// azureCreds uses the Azure SDK for Go's standard default credential chain.It attempts to authenticate with each of
```
Miciah, line 106: Missing space:
```suggestion
// azureCreds uses the Azure SDK for Go's standard default credential chain.  It attempts to authenticate with each of
```

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
...
+func azureCreds(options *azidentity.DefaultAzureCredentialOptions) (*azidentity.DefaultAzureCredential, error) {
+	err := os.Setenv("AZURE_CLIENT_SEND_CERTIFICATE_CHAIN", "true")
+	if err != nil {
+		return nil, fmt.Errorf("could not set AZURE_CLIENT_SEND_CERTIFICATE_CHAIN")
+	}
+
+	if certPath := os.Getenv("AZURE_CLIENT_CERTIFICATE_PATH"); certPath != "" {
+		// Set up a watch on our config file; if it changes, we should exit -
+		// (we don't have the ability to dynamically reload config changes).
+		stopCh := make(chan struct{})
```
Miciah, line 123: Wiring up `stopCh` is the missing piece, and it's is the reason why this is still a draft PR—right?

The whole "watch for a change and terminate the process if the change is observed" pattern makes this feel like something that might fit into library-go, similar to the featuregate logic: https://github.com/openshift/cluster-ingress-operator/blob/4975cde4f58e993bf8a1a4f72951036c2715b6b4/pkg/operator/operator.go#L110-L111

Have you considered putting some of this logic in library-go, or is that not practical given time constraints or other constraints?

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
...
+		}
+	}
+
+	go func() {
+		for {
+			select {
+			case <-stopCh:
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
```
Miciah, line 175: It might be helpful to log something here.

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
...
+				}
+				klog.V(2).Infof("Configuration file %s changed, exiting...", event.Name)
+				close(stopCh)
+				return
+			case err, ok := <-watcher.Errors:
+				if !ok {
+					return
```
Miciah, line 182: Again, it might be helpful to log something here.

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
...
+				}
+				klog.V(4).Infof("fsnotify error %v", err)
```
Miciah, line 184: Any reason not to use `klog.Error`?
```suggestion
				klog.Errorf("fsnotify error: %v", err)
```

pkg/dns/azure/client/auth.go
```diff
@@ -132,3 +102,88 @@ func endpointToScope(endpoint string) string {
...
+			}
+		}
+	}()
+	return nil
+}
```
Miciah, line 189: We have some logic in OpenShift router for watching volume mounts: https://github.com/openshift/router/blob/72114ea78be0a032d5cdddcad694e0a352a00e93/pkg/router/template/router.go#L344-L410

The main difference is that instead of checking for and chasing symlinks, the router logic just assumes that there is a `..data` symlink and watches that.  

There are probably many different implementations of using fsnotify on a volume, which is another reason I'd prefer if some of this logic could be moved to library-go.

pkg/dns/azure/client/auth.go
```diff
@@ -1,27 +1,27 @@
...
 
 func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
 	var cloudConfig cloud.Configuration
 	switch config.Environment {
 	case azure.ChinaCloud:
 		cloudConfig = cloud.AzureChina
-	// GermanCloud was closed on Oct 29, 2021
-	// https://learn.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud
-	// case azure.GermanCloud:
-	// return nil, nil
```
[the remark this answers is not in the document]
r4f4: @Miciah Which other reason do you need other than `"// GermanCloud was closed on Oct 29, 2021"`. The new `azidentity` doesn't even have a config for `GermanCloud` as it does for the other clouds.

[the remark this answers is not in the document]
bryan-cox: Yeah I had just removed it since there is no German cloud and I hadn't seen a message like this in the other repos in similar code. Basically removed it as clean up. Happy to put it back if you want.

pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +38,48 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
 	}
 
-	// Fallback to using tenant ID from env variable if not set.
-	if strings.TrimSpace(config.TenantID) == "" {
-		config.TenantID = os.Getenv("AZURE_TENANT_ID")
-		if strings.TrimSpace(config.TenantID) == "" {
-			return nil, errors.New("empty tenant ID")
+	// If the environment variable is set for any of the following, use the environment variable over any configuration
+	// value: AZURE_TENANT_ID, AZURE_CLIENT_ID, and AZURE_CLIENT_SECRET. The Azure SDK for Go function,
+	// NewDefaultAzureCredential, checks environment variables first.
```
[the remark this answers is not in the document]
bryan-cox: Yeah, I can see where that is not exactly clear. I was meaning (b)

pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +38,48 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
+	if os.Getenv("AZURE_TENANT_ID") == "" {
+		err := os.Setenv("AZURE_TENANT_ID", config.TenantID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to set environment variable AZURE_TENANT_ID: %v", err)
```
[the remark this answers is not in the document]
bryan-cox: No I didn't mean to drop that. It's been added back.

pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +38,48 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
-		config.FederatedTokenFile = os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
-		if strings.TrimSpace(config.FederatedTokenFile) == "" {
-			// Default to a generic token file location.
-			config.FederatedTokenFile = "/var/run/secrets/openshift/serviceaccount/token"
+	if config.AzureWorkloadIdentityEnabled &&
+		os.Getenv("AZURE_CLIENT_SECRET") == "" &&
+		strings.TrimSpace(config.ClientSecret) == "" &&
+		os.Getenv("AZURE_FEDERATED_TOKEN_FILE") == "" &&
+		strings.TrimSpace(config.FederatedTokenFile) == "" {
+		err := os.Setenv("AZURE_FEDERATED_TOKEN_FILE", "/var/run/secrets/openshift/serviceaccount/token")
```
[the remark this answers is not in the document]
bryan-cox: That was an oversight on my part. I'll add that in.

pkg/dns/azure/client/auth.go
```diff
@@ -38,78 +38,48 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
-		options := azidentity.WorkloadIdentityCredentialOptions{
-			ClientOptions: azcore.ClientOptions{
-				Cloud: cloudConfig,
-			},
-			ClientID:      config.ClientID,
-			TenantID:      config.TenantID,
-			TokenFilePath: config.FederatedTokenFile,
-		}
-		var err error
-		cred, err = azidentity.NewWorkloadIdentityCredential(&options)
```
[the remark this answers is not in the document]
bryan-cox: Yeah essentially. Set the right environment variables and let `azidentity.NewDefaultAzureCredential` figure out what the right authentication type is.

pkg/dns/azure/client/auth.go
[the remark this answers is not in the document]
Miciah: Thanks!  Yeah, let's keep the comment then.

pkg/dns/azure/client/auth.go
bryan-cox: I believe I wired up the `stopCh` appropriately based on another OpenShift repo this code is from.

Yeah Alberto mentioned it would be nice to have this logic in library-go somewhere but given time constraints to do that, it's not practical at the moment.

bryan-cox:
I tested this PR with the following PRs and was able to deploy an Azure Hosted Cluster from an AKS management cluster:
- https://github.com/openshift/hypershift/pull/4811
- https://github.com/openshift/hypershift/pull/4877
- https://github.com/openshift/hypershift/pull/4884

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
+	var err error
+
+	// This starts only one occurrence of the file watcher, which watches the file, fileToWatch.
+	watchCertificateFileOnce.Do(func() {
+		klog.Infof("Starting the file change watcher on file, %s", fileToWatch)
+
+		// Update the file path to watch in case this is a symlink
+		fileToWatch, err = filepath.EvalSymlinks(fileToWatch)
```
gcs278, line 24: Have you explore the case where the file doesn't exist when this starts, then appears after the operator starts? I think this function will error out. Do you think it'd be more explicit to check for file existence before this function?

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if ok && (event.Has(fsnotify.Write) || event.Has(fsnotify.Chmod) || event.Has(fsnotify.Remove)) {
+					klog.Infof("file, %s, was modified, exiting...", event.Name)
+					os.Exit(0)
+				}
+			case err, ok := <-watcher.Errors:
+				if ok {
```
Miciah, line 52: Should we log an error for the `!ok` cases?

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+					klog.Errorf("file watcher error: %v", err)
+				}
+			}
+		}
+	}()
+
+	err = watcher.Add(path)
+	if err != nil {
```
Miciah, line 60:
```suggestion
	if err := watcher.Add(path); err != nil {
```

pkg/dns/azure/client/auth.go
```diff
@@ -2,15 +2,17 @@ package client
...
-	"strings"
-
+	"fmt"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"os"
+	"strings"
```
gcs278, line 12: Mind bumping these back up to the original grouping?

pkg/dns/azure/client/auth.go
```diff
@@ -2,15 +2,17 @@ package client
 
 import (
 	"errors"
-	"os"
-	"strings"
-
+	"fmt"
```
gcs278, line 5: Mind bumping this up to the previous group?

gcs278 (commented):
Looks good , but I do feel a bit mixed about shutting down the operator to get updates to the cert file. Did you consider writing logic to refresh the client if the cert gets updated instead of shutting down the operator? I'm assuming that will require more effort.

But it's scoped to HCP logic, so the risk is lower. If we need to get this feature in, I suppose it could work if other folks don't have objections.

pkg/dns/azure/client/auth.go
```diff
@@ -73,15 +75,33 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
 			ClientOptions: azcore.ClientOptions{
 				Cloud: cloudConfig,
 			},
-			ID: azidentity.ClientID(managedIdentityClientID),
+			SendCertificateChain: true,
 		}
 
-		var err error
-		cred, err = azidentity.NewManagedIdentityCredential(&options)
+		tenantID := os.Getenv("ARO_HCP_TENANT_ID")
```
Miciah, line 85: What is the relationship between the tenant id (`config.TenantID` / `AZURE_TENANT_ID`) and the HCP tenant id (`ARO_HCP_TENANT_ID`)?  `getAuthorizerForResource` returns an error if the tenant id isn't set, but it looks like the tenant id isn't ultimately used if the HCP tenant id is used.  Does HCP just set a dummy value for tenant id?  Will this be simplified in the future?  For now, can you add a comment explaining that the tenant id is ignored if the HCP tenant id is used (or clarify if I misunderstood the logic)?

Similarly, it looks like the client id is ignored when using the HCP client id.

Miciah (commented):
Please squash the commits; each commit should pass `make test verify`.

Miciah:
@gcs278, looks like we both submitted reviews at about the same time.  Apologies, Grant and Bryan, for any redundant review comments!

> Looks good , but I do feel a bit mixed about shutting down the operator to get updates to the cert file. Did you consider writing logic to refresh the client if the cert gets updated instead of shutting down the operator? I'm assuming that will require more effort.

Quoting https://github.com/openshift/enhancements/pull/1694:

> Unfortunately, the SDK today does not support hot-reloading the credential. While Microsoft works on getting that into the SDK, we could either use a library to create a generator for the credential and allow reloading in-process, or use the common OpenShift fsnotify +os.Exit() to restart the pod on config change.

Using `os.Exit()` to force a pod restart when the operator needs to reread configuration has precedent with the featuregate logic.

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+		// Update the file path to watch in case this is a symlink
+		fileToWatch, err = filepath.EvalSymlinks(fileToWatch)
+		if err != nil {
+			return
+		}
+		klog.Infof("Watching file, %s", fileToWatch)
+
+		// Start the file watcher to monitor file changes
+		go checkForFileChanges(fileToWatch)
```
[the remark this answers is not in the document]
bryan-cox: I'm my previous testing with other operators it did need to be a goroutine. 

I added error handling to it.

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog/v2"
+)
+
+var watchCertificateFileOnce sync.Once
+
+// WatchFileForChanges watches the file, fileToWatch, for changes. If the file contents have changed, the pod this
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
```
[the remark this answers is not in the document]
bryan-cox: I had. I was actually planning to put in a follow up PR in librarygo with this function and take care of it there. After that PR, I was planning to update ingress, cncc, and image registry operators with that function so that the exact same function was being utilized by all 3. 

There wasn't enough time to do that initially. We need to get this into 4.18 and backported to 4.17.z before the ARO HCP milestone P2 which is Dec 13th.

pkg/util/filewatcher/filewatcher.go
bryan-cox: I'm my previous testing with other operators it did need to be a goroutine.

I added error handling to it.

bryan-cox: I don't think we can specify the items in the volume source since the volume source is CSI.

pkg/dns/azure/client/auth.go
bryan-cox: Yes, sorry, I'm not sure how those got moved around (assuming my IDE did it).

pkg/dns/azure/client/auth.go
bryan-cox: Yes, sorry, I'm not sure how those got moved around (assuming my IDE did it).

pkg/dns/azure/client/auth.go
```diff
@@ -73,15 +75,33 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
+		certPath := os.Getenv("ARO_HCP_CLIENT_CERTIFICATE_PATH")
+
+		certData, err := os.ReadFile(certPath)
+		if err != nil {
+			return nil, fmt.Errorf(`failed to read certificate file "%s": %v`, certPath, err)
```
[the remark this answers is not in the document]
bryan-cox: Got it, thanks!

pkg/dns/azure/client/auth.go
```diff
@@ -73,15 +75,33 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
+		}
+
+		certs, key, err := azidentity.ParseCertificates(certData, []byte{})
```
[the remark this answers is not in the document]
bryan-cox: No, I thought nil would cause issues when `ParseCertificates` did a length check on password but I guess that is 0 when that happens.

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
+package filewatcher
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog/v2"
```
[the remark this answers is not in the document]
bryan-cox: No reason, that was a copy/paste thing from bringing this from another OpenShift operator. I will change it to use log.

pkg/util/filewatcher/filewatcher.go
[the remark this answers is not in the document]
bryan-cox: See discussion on https://github.com/openshift/cluster-ingress-operator/pull/1151#discussion_r1841349225

pkg/util/filewatcher/filewatcher.go
bryan-cox: I don't think it will hurt anything to do that.

pkg/dns/azure/client/auth.go
bryan-cox: Yeah we will just be passing a dummy client ID/tenant ID through the required secret - https://github.com/openshift/hypershift/pull/4587/commits/37f837af562216ce71857be86dd8e4ad2099cc6a#diff-c888020c5dc46c458d818b931db97131d6e35b90661fe1030a39ebeac8859b19R1443 

`AZURE_TENANT_ID` and `AZURE_CLIENT_ID` ultimately is not used in ARO HCP cases for the operator. We are still reconciling the needed secret since its required in ingress. The ingress controller fails if the secret is not provided. The ingress controller runs on the control plane side in HCP. In managed azure, we are overriding the Azure credentials authentication method to always use client certificate authentication. The secret is just created so that the ingress controller does not fail; the data in the secret is never used due to the aforementioned override to use client certificate authentication.

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+
+var watchCertificateFileOnce sync.Once
+
+// WatchFileForChanges watches the file, fileToWatch, for changes. If the file contents have changed, the pod this
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
+	var err error
+
+	// This starts only one occurrence of the file watcher, which watches the file, fileToWatch.
+	watchCertificateFileOnce.Do(func() {
```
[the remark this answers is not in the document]
bryan-cox: I moved the sync.Once logic to the caller.

pkg/dns/azure/client/auth.go
```diff
@@ -70,18 +76,43 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
+		certs, key, err := azidentity.ParseCertificates(certData, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate data %q: %w", certPath, err)
+		}
+
+		// Watch the certificate for changes; if the certificate changes, the pod will be restarted.
+		// This starts only one occurrence of the file watcher, which watches the file, certPath.
+		var fileWatcherError error
+		watchCertificateFileOnce.Do(func() {
+			if err = filewatcher.WatchFileForChanges(certPath); err != nil {
```
Miciah, line 107:
```suggestion
			if err := filewatcher.WatchFileForChanges(certPath); err != nil {
```

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if ok && (event.Has(fsnotify.Write) || event.Has(fsnotify.Chmod) || event.Has(fsnotify.Remove)) {
```
Miciah, line 46: Maybe log an error in the `!ok` case?
```suggestion
				if !ok {
					log.Error(nil, "fsnotify channel closed")
				} else if event.Has(fsnotify.Write) || event.Has(fsnotify.Chmod) || event.Has(fsnotify.Remove) {
```

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+					log.Info("file was modified, exiting...", "file", event.Name)
+					os.Exit(0)
+				}
+			case err, ok := <-watcher.Errors:
+				if ok {
+					log.Error(err, "file watcher error")
```
Miciah, line 52: `!ok` means the channel was closed (according to <https://go.dev/ref/spec#Receive_operator>).  In this case, `err` will be nil.  I'd suggest using a `nil` literal value to make this more explicit, and log something like "fsnotify channel closed".  
```suggestion
					log.Error(nil, "fsnotify channel closed")
```

pkg/util/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,65 @@
...
+				} else {
+					log.Error(err, "failed to retrieve watcher error")
+				}
+			}
+		}
+	}()
+
+	if err = watcher.Add(path); err != nil {
```
Miciah, line 60:
```suggestion
	if err := watcher.Add(path); err != nil {
```

pkg/dns/azure/client/auth.go
```diff
@@ -70,18 +76,43 @@ func getAuthorizerForResource(config Config) (autorest.Authorizer, error) {
...
+		certPath := os.Getenv("ARO_HCP_CLIENT_CERTIFICATE_PATH")
+
+		certData, err := os.ReadFile(certPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read certificate file %q: %w", certPath, err)
+		}
+
+		certs, key, err := azidentity.ParseCertificates(certData, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse certificate data %q: %w", certPath, err)
```
Miciah, line 100: The log message has the file path, not the file content (that is, not the certificate data), so the current wording is potentially misleading.  
```suggestion
			return nil, fmt.Errorf("failed to parse certificate data read from %q: %w", certPath, err)
```

pkg/util/filewatcher/filewatcher.go
Miciah: I'd like to understand why you need `go` in both places, but I can let it be, especially if you're planning to refactor this code eventually anyway.

Miciah:
Thanks!  I've made a few more suggestions, all for minor issues.  I'll add my `/lgtm` and leave it up to you and Grant if you want to make further changes.  

/approve
/lgtm
/hold

gcs278:
I don't have any major concerns. @Miciah's comment about catching fsnotify channel closed seems valid, but not a show stopper.
/lgtm

@bryan-cox feel free to remove the hold when you are ready.

PR #1153, opened 2024-10-10
OCPBUGS-43033: e2e/ingress_dns: support both private & public

EmilienM:
Some platforms only have public zones, some others private, some none.
Let's just run zone update tests where applicable so we support
all platforms and also cover more scenarios.

### DIFF
--- a/test/e2e/dns_ingressdegrade_test.go
+++ b/test/e2e/dns_ingressdegrade_test.go
@@ -10,6 +10,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	"k8s.io/apimachinery/pkg/types"
@@ -35,6 +36,23 @@ func TestIngressStatus(t *testing.T) {
 		t.Fatalf("failed to create kube client: %v", err)
 	}
 
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig); err != nil {
+		t.Fatalf("failed to get DNS config: %v", err)
+	}
+
+	// Run DNS Config update tests on private and public zones when
+	// they are defined in the DNS config (which depends on the platform).
+	if dnsConfig.Spec.PrivateZone != nil {
+		t.Log("Testing private zone")
+		testUpdateDNSConfig(t, kubeClient)
+	}
+	if dnsConfig.Spec.PublicZone != nil {
+		t.Log("Testing public zone")
+		testUpdateDNSConfig(t, kubeClient)
+	}
+}
+
+func testUpdateDNSConfig(t *testing.T, kubeClient client.Client) {
 	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig); err != nil {
 		t.Fatalf("failed to get DNS config: %v", err)
 	}


test/e2e/dns_ingressdegrade_test.go
```diff
@@ -86,20 +86,22 @@ func updateDNSConfig(set bool, dnsConfig *configv1.DNS) {
 	// Tested on Azure, AWS, and GCP.  Azure privateZone uses "/" notation,
 	// so prepending "error-" without "/" does not cause a failure on Azure.
 	const injectError = "/error"
-	if dnsConfig.Spec.PrivateZone.ID != "" {
-		if set {
-			dnsConfig.Spec.PrivateZone.ID = injectError + dnsConfig.Spec.PrivateZone.ID
-		} else {
-			// Remove injectError from prefix.
-			dnsConfig.Spec.PrivateZone.ID = dnsConfig.Spec.PrivateZone.ID[len(injectError):]
+	if dnsConfig.Spec.PrivateZone != nil {
```
Miciah, line 89: This will cause step 3 of `TestIngressStatus` to fail, right?  `TestIngressStatus` needs a skip when the private zone is not specified.  

It would be even better if we had tests for both the private zone and public zone, with an appropriate skip for each zone if it isn't set.  

Historical note: #641 added the original test.  At that point, I believe, putting an invalid value in the private zone would cause the clusteroperator to report a problem, but putting an invalid value in the public zone would not.  #967 subsequently modified the status reporting so that an invalid value in either public or private zone config would cause the clusteroperator to report an error.  In retrospect, we should have updated this test as part of #967.

EmilienM: Before skipping this test I want to know what scenarios we want to test here.
Do we want to test:

- Public empty
  Private set

- Public set
  Private set

- Both set

- Both empty (I guess that's not possible?)

Miciah: I think all combinations are possible:

- As far as I know, most cloud platforms have both public and private set.
- Today I learned that OpenStack has public set and private nil.  
- An internal/private cluster can have public nil and private set.
- Public and private can both be nil on, say, a bare-metal cluster; the cluster-admin must configure DNS manually.  

The simplest thing would be to split `TestIngressStatus` into two tests: one for the public zone, and one for the private zone, each with a skip rule if the zone is not specified.  Most of the current test logic can be factored into new helper function, so each test can be just a skip rule and then a call to the helper.

EmilienM: I've updated it. Let me know if you like the logic, or if you really want it split by tests that are skipped when needed.

test/e2e/dns_ingressdegrade_test.go
```diff
@@ -39,6 +40,23 @@ func TestIngressStatus(t *testing.T) {
 		t.Fatalf("failed to get DNS config: %v", err)
 	}
```
Miciah, line 42: You should be able to remove the `Get(..., ..., &dnsConfig)` above since `TestMain` initializes `dnsConfig`, but we can keep it if you prefer to keep the test logic self-contained.  

For that matter, we should use `kclient` that `TestMain` also initializes, but that can be left as a follow-up clean-up.

Miciah (commented):
Thanks!

/approve
/lgtm

/hold
since I made some comments, but feel free just to `/hold cancel` if you want to keep the PR as is.

test/e2e/dns_ingressdegrade_test.go
EmilienM: I'll try doing it here

test/e2e/dns_ingressdegrade_test.go
```diff
@@ -39,6 +40,23 @@ func TestIngressStatus(t *testing.T) {
...
+		testUpdateDNSConfig(t, kubeClient)
+	}
+	if dnsConfig.Spec.PublicZone != nil {
+		t.Log("Testing public zone")
+		testUpdateDNSConfig(t, kubeClient)
+	}
+}
+
+func testUpdateDNSConfig(t *testing.T, kubeClient client.Client) {
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig); err != nil {
```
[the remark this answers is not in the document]
EmilienM: I'm in favor of doing it as a follow-up with your help maybe. I've tried myself and I suspect I didn't get it right.

test/e2e/dns_ingressdegrade_test.go
EmilienM: I'll follow-up with a PR to address both comments. That'll help me to iterate at the same time on my other PR for LB.
Thanks

EmilienM:
OpenStack errors: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1153/pull-ci-openshift-cluster-ingress-operator-master-e2e-openstack-operator/1844725964710350848/artifacts/e2e-openstack-operator/test/build-log.txt

PR #1154, opened 2024-10-10
OCPBUGS-42480: Limit SHA-1 check to the leaf cert in default cert chain

gcs278:
Previously, the Upgradeable status was set to false if any certificates in the default certificate chain used SHA-1. However, HAProxy only rejects SHA-1 for the leaf (serving) certificate. This update ensures that the Upgradeable status is set to false only if the leaf certificate uses SHA-1, while ignoring other certificates in the chain.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -100,6 +100,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Index IngressControllers over the default certificate name so that
+	// secretToIngressController can look up IngressControllers that
+	// reference the secret.
+	if err := operatorCache.IndexField(context.Background(), &operatorv1.IngressController{}, "ingressDefaultCertificateName", func(o client.Object) []string {
+		secret := operatorcontroller.RouterEffectiveDefaultCertificateSecretName(o.(*operatorv1.IngressController), o.GetNamespace())
+		return []string{secret.Name}
+	}); err != nil {
+		return nil, fmt.Errorf("failed to create index for ingresscontroller: %w", err)
+	}
+
 	scheme := mgr.GetClient().Scheme()
 	mapper := mgr.GetClient().RESTMapper()
 	if err := c.Watch(source.Kind(operatorCache, &operatorv1.IngressController{}), &handler.EnqueueRequestForObject{}); err != nil {
@@ -134,7 +144,47 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind(operatorCache, &configv1.Proxy{}), handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
 		return nil, err
 	}
+	// Watch secrets in the operand namespace that are referenced as default certificates so that
+	// checkDefaultCertificate can determine whether changes to the default certificate affect upgradeability.
+	if err := c.Watch(source.Kind(operatorCache, &corev1.Secret{}), handler.EnqueueRequestsFromMapFunc(reconciler.secretToIngressController), predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object.GetNamespace() == operatorcontroller.DefaultOperandNamespace
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Object.GetNamespace() == operatorcontroller.DefaultOperandNamespace
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectNew.GetNamespace() == operatorcontroller.DefaultOperandNamespace
+		},
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}); err != nil {
+		return nil, err
+	}
 	return c, nil
+}
+
+// secretToIngressController maps a secret to a slice of reconcile requests,
+// one request per IngressController that references the secret as its default certificate.
+func (r *reconciler) secretToIngressController(ctx context.Context, o client.Object) []reconcile.Request {
+	var requests []reconcile.Request
+	controllers := &operatorv1.IngressControllerList{}
+	listOpts := client.MatchingFields(map[string]string{
+		"ingressDefaultCertificateName": o.GetName(),
+	})
+	if err := r.cache.List(ctx, controllers, client.InNamespace(r.config.Namespace), listOpts); err != nil {
+		log.Error(err, "failed to list ingresscontrollers for default certificate secret", "related", o.GetSelfLink())
+		return requests
+	}
+	for _, ic := range controllers.Items {
+		request := reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: ic.Namespace,
+				Name:      ic.Name,
+			},
+		}
+		requests = append(requests, request)
+	}
+	return requests
 }
 
 func (r *reconciler) ingressConfigToIngressController(ctx context.Context, o client.Object) []reconcile.Request {
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -728,9 +728,13 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
 		if cert.Subject.CommonName == domain && !foundSAN {
 			return fmt.Errorf("certificate in secret %s/%s has legacy Common Name (CN) but has no Subject Alternative Name (SAN) for domain: %s", secret.Namespace, secret.Name, domain)
 		}
-		switch cert.SignatureAlgorithm {
-		case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
-			return fmt.Errorf("certificate in secret %s/%s has weak SHA1 signature algorithm: %s (see https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-sha-haproxy-support-removed_release-notes for more details)", secret.Namespace, secret.Name, cert.SignatureAlgorithm)
+		// Prevent the upgrade only if the leaf certificate (the first certificate) uses a SHA1 signature algorithm.
+		// SHA1 can still be used by other certificates in the chain, such as the root and intermediate certificates.
+		if !cert.IsCA {
+			switch cert.SignatureAlgorithm {
+			case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
+				return fmt.Errorf("certificate in secret %s/%s has weak SHA1 signature algorithm: %s (see https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-sha-haproxy-support-removed_release-notes for more details)", secret.Namespace, secret.Name, cert.SignatureAlgorithm)
+			}
 		}
 	}
 
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -1,6 +1,7 @@
 package ingress
 
 import (
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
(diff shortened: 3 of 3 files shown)


gcs278:
/hold
Doing some sanity checks and testing.

gcs278:
/assign @Miciah 

/hold
Miciah pointed out in slack we should have test coverage for this, I forgot to add a test case to `Test_computeIngressUpgradeableCondition`

gcs278:
testing has been added
/unhold
/jira refresh

pkg/operator/controller/ingress/status_test.go
```diff
@@ -2215,51 +2216,95 @@ func Test_checkZoneInConfig(t *testing.T) {
...
 			SerialNumber:       big.NewInt(1),
 			Subject:            pkix.Name{CommonName: cn},
 			DNSNames:           sans,
 			SignatureAlgorithm: signatureAlgorithm,
 		}
-
-		var cert []byte
+		if isCA {
+			certTemplate.IsCA = true
+			certTemplate.KeyUsage = x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign
```
Miciah, line 2243: I believe you also need `certTemplate.BasicConstraintsValid = true`.

gcs278: Yup, thanks, done.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -2295,23 +2340,33 @@ func Test_computeIngressUpgradeableCondition(t *testing.T) {
...
+			secret:      makeCertificateSecret(sha1RSACert),
+			expect:      false,
+		},
+		{
+			description: "if the default leaf certificate uses RSA SHA256 included with it's root certificate that uses RSA SHA1",
+			secret:      makeCertificateSecret(sha256RSACertSignedBySha1RSA, sha1RSARootCA),
+			expect:      true,
+		},
+		{
+			description: "if the default leaf certificate uses RSA SHA1 included with it's root certificate that uses RSA SHA256",
```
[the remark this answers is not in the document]
gcs278: Done.

pkg/operator/controller/ingress/status.go
```diff
@@ -728,10 +729,15 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
 		if cert.Subject.CommonName == domain && !foundSAN {
 			return fmt.Errorf("certificate in secret %s/%s has legacy Common Name (CN) but has no Subject Alternative Name (SAN) for domain: %s", secret.Namespace, secret.Name, domain)
 		}
-		switch cert.SignatureAlgorithm {
-		case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
-			return fmt.Errorf("certificate in secret %s/%s has weak SHA1 signature algorithm: %s (see https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-sha-haproxy-support-removed_release-notes for more details)", secret.Namespace, secret.Name, cert.SignatureAlgorithm)
+		// Prevent the upgrade only if the leaf certificate (the first certificate) uses a SHA1 signature algorithm.
+		// SHA1 can still be used by other certificates in the chain, such as the root and intermediate certificates.
+		if certIndex == 0 {
```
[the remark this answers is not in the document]
gcs278: I tried to test this, but I couldn't work around this error while creating a TLS secret with improperly ordered certificate:
```
# reverse order
cat exampleca.crt example.crt > example-combo-reversed.crt
$ oc create secret -n openshift-ingress tls router-cert --key=example.key --cert=example-combo-reversed.crt --validate=false
error: tls: private key does not match public key
```

However, in case there is a way to force a CA as index 0 that I am unaware, generally it seems more resilient to do `if !cert.IsCA` to handle any possible edge case.

Thanks! Updated.

gcs278 (commented):
Thanks for the suggestions @miciah

gcs278:
I'll do one last round of clusterbot testing with the new updates.
/hold

Miciah:
Thanks!
/approve
/lgtm

Miciah:
This is a low-risk change to the operator's logic for blocking upgrades when the operator detects that the default certificate uses the problematic SHA-1 signature algorithm.  This change adds an additional condition that must be met before setting `Upgradeable=False` in order to reduce false positives that block upgrades.  

/label backport-risk-assessed

gcs278:
My sanity checks are complete:
/unhold

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,9 +134,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+func (r *reconciler) defaultCertificateSecretToIngressController(ctx context.Context, o client.Object) []reconcile.Request {
+	var requests []reconcile.Request
+	secret := o.(*corev1.Secret)
+	controllers := &operatorv1.IngressControllerList{}
+	if err := r.cache.List(ctx, controllers, client.InNamespace(r.config.Namespace)); err != nil {
+		log.Error(err, "failed to list ingresscontrollers for default certificate secret", "related", o.GetSelfLink())
+		return requests
+	}
+	for _, ic := range controllers.Items {
+		if ic.Spec.DefaultCertificate != nil && ic.Spec.DefaultCertificate.Name == secret.Name {
```
Miciah, line 169: `ic.Spec.DefaultCertificate == nil` implies that it's an operator-generated secret, and in the context of the upgradeability check, I suppose we don't really need to check the secret if it's an operator-generated secret.  However, `defaultCertificateSecretToIngressController` itself isn't obviously applicable only to custom default certificates.  Could you either add a comment to make this explicit, or make sure to cover operator-generated secrets?  For the latter option, I think using `RouterEffectiveDefaultCertificateSecretName` would be sufficient and simple.

Miciah, line 169: You might also consider using an index, à la the certificate-publisher controller's `secretToIngressController` function.

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,9 +134,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind(operatorCache, &configv1.Proxy{}), handler.EnqueueRequestsFromMapFunc(reconciler.ingressConfigToIngressController)); err != nil {
 		return nil, err
 	}
+	// Watch secrets in the operand namespace that are referenced as default certificates so that
+	// checkDefaultCertificate can determine whether changes to the default certificate affect upgradeability.
+	// This watch is NOT required for updating the router deployment, as its volume mount automatically
+	// handles updates to the default certificate.
```
Miciah, line 140: I was thinking it would make sense to state the **non**-reason for the watch:  It would be easy to assume that we need the watch in order to update router pods when the secret changes; however, the kubelet automatically updates the router pods' volume mount when the secret changes, so this watch is *not* needed for that purpose.  

On further thought, this might be unnecessary to state explicitly, and the lack of need for a watch to update router pods precedes any of this SHA-1 stuff, so feel free to disregard this suggestion.

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,9 +134,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+	if err := c.Watch(source.Kind(operatorCache, &corev1.Secret{}), handler.EnqueueRequestsFromMapFunc(reconciler.defaultCertificateSecretToIngressController), predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object.(*corev1.Secret).Namespace == operatorcontroller.DefaultOperandNamespace
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Object.(*corev1.Secret).Namespace == operatorcontroller.DefaultOperandNamespace
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectNew.(*corev1.Secret).Namespace == operatorcontroller.DefaultOperandNamespace
```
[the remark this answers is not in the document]
gcs278: Sure, done.

pkg/operator/controller/ingress/controller.go
gcs278: Ah good point, didn't think about the operator generated secret.

Since `checkDefaultCertificate` also checks the operator generated secret, it makes sense to reconcile changes to it to be complete.

pkg/operator/controller/ingress/controller.go
gcs278: I will disregard, thanks.

pkg/operator/controller/ingress/controller.go
gcs278: I thought about it, but initially didn't add because it felt a bit unnecessary, but now that you pointed out that someone already wrote it, and I felt like I couldn't say no now. Added.

gcs278:
need to do another test
/hold

pkg/operator/controller/ingress/controller.go
gcs278: Whoops, didn't realize you can't have two indexes as the same name.

I just updated the name, but do you think having two redundant indexes is a problem? I suppose I could share them, but that would create a dependency between the two control loops, which feels odd to me.

Miciah: > Whoops, didn't realize you can't have two indexes as the same name.
> 
> I just updated the name, but do you think having two redundant indexes is a problem? I suppose I could share them, but that would create a dependency between the two control loops, which feels odd to me.

It shouldn't cause any problems to have redundant indexes other than the cost of the redundancy in terms of code, memory, and computation.  However, I expect you could create the index in `pkg/operator/operator.go` and pass it to each controller that needed it.  

If you do that refactoring, please do it in a separate commit so that we can cherry-pick it to the main branch.  But maybe it's simplest, for this PR, just to live with the redundancy.

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,9 +144,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+		},
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}); err != nil {
+		return nil, err
+	}
 	return c, nil
 }
 
+// secretToIngressController maps a secret to a slice of reconcile requests,
+// one request per IngressController that references the secret as it's default certificate.
```
Miciah, line 167:
```suggestion
// one request per IngressController that references the secret as its default certificate.
```
But this isn't worth worrying about if you aren't making any other changes.

pkg/operator/controller/ingress/controller.go
```diff
@@ -100,6 +100,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
 		return nil, err
 	}
+	// Index IngressControllers over the default certificate name so that
+	// secretToIngressController can look up IngressControllers that
+	// reference the secret.
+	if err := operatorCache.IndexField(context.Background(), &operatorv1.IngressController{}, "ingressDefaultCertificateName", func(o client.Object) []string {
+		secret := operatorcontroller.RouterEffectiveDefaultCertificateSecretName(o.(*operatorv1.IngressController), operatorcontroller.DefaultOperandNamespace)
+		return []string{secret.Name}
+	}); err != nil {
+		return nil, fmt.Errorf("failed to create index for ingresscontroller: %v", err)
```
Miciah, line 110:
```suggestion
		return nil, fmt.Errorf("failed to create index for ingresscontroller: %w", err)
```
But this isn't worth worrying about if you aren't making any other changes.

pkg/operator/controller/ingress/controller.go
```diff
@@ -100,6 +100,16 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Index IngressControllers over the default certificate name so that
+	// secretToIngressController can look up IngressControllers that
+	// reference the secret.
+	if err := operatorCache.IndexField(context.Background(), &operatorv1.IngressController{}, "ingressDefaultCertificateName", func(o client.Object) []string {
+		secret := operatorcontroller.RouterEffectiveDefaultCertificateSecretName(o.(*operatorv1.IngressController), operatorcontroller.DefaultOperandNamespace)
```
Miciah, line 107: It would be good practice to use `o.GetNamespace()` just in case we ever support ingresscontrollers in other namespaces:
```suggestion
		secret := operatorcontroller.RouterEffectiveDefaultCertificateSecretName(o.(*operatorv1.IngressController), o.GetNamespace())
```
But it isn't really necessary for the release-4.15 branch.

pkg/operator/controller/ingress/controller.go
```diff
@@ -134,9 +144,51 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+	controllers := &operatorv1.IngressControllerList{}
+	listOpts := client.MatchingFields(map[string]string{
+		"ingressDefaultCertificateName": o.GetName(),
+	})
+	if err := r.cache.List(ctx, controllers, client.InNamespace(r.config.Namespace), listOpts); err != nil {
+		log.Error(err, "failed to list ingresscontrollers for default certificate secret", "related", o.GetSelfLink())
+		return requests
+	}
+	for _, ic := range controllers.Items {
+		if operatorcontroller.RouterEffectiveDefaultCertificateSecretName(&ic, operatorcontroller.DefaultOperandNamespace).Name == o.GetName() {
```
Miciah, line 179: This `if` check should be redundant now since you used an index to get the list of items.

Miciah:
I only had some minor suggestions this time.  
/approve
/lgtm

/hold
in case you decide to implement any of those suggestions.

pkg/operator/controller/ingress/controller.go
gcs278: Done.

pkg/operator/controller/ingress/controller.go
gcs278: Done.

pkg/operator/controller/ingress/controller.go
gcs278: Done.

pkg/operator/controller/ingress/controller.go
gcs278: Good point, done.

pkg/operator/controller/ingress/controller.go
gcs278: Fair point. I lean towards leaving it simple. Unless you think duplicate index makes us worse off than just not using an index at all, I'm inclined to leave it as is. We are restricted to secrets in the `openshift-ingress` namespace, and my guess is that shouldn't see a whole lot of activity.

gcs278:
Manually tested to ensure default certificate secret updates are reconciled.
/unhold

gcs278:
must gather failure
/test e2e-gcp-operator

ShudiLi:
tested it with 4.15.0-0.ci.test-2024-10-23-070517-ci-ln-z89svkb-latest

```
1.
%  oc -n openshift-ingress-operator get ingresscontroller default -oyaml
kind: IngressController
metadata:
  creationTimestamp: "2024-10-23T07:38:48Z"
  finalizers:
  - ingresscontroller.operator.openshift.io/finalizer-ingresscontroller
  generation: 2
  name: default
  resourceVersion: "42881"
  uid: 591fbdba-b77e-492c-adfb-741bf46d92b9
spec:
  clientTLS:
    clientCA:
      name: ""
    clientCertificatePolicy: ""
  defaultCertificate:
    name: router-cert
  httpCompression: {}
  httpEmptyRequestsPolicy: Respond
  httpErrorCodePages:
    name: ""
  replicas: 2
  tuningOptions:
    reloadInterval: 0s
  unsupportedConfigOverrides: null
status:
  availableReplicas: 2
  conditions:
  - lastTransitionTime: "2024-10-23T07:38:49Z"
    reason: Valid
    status: "True"
    type: Admitted
  - lastTransitionTime: "2024-10-23T07:39:50Z"
    message: Canary route checks for the default ingress controller are successful
    reason: CanaryChecksSucceeding
    status: "True"
    type: CanaryChecksSucceeding
  - lastTransitionTime: "2024-10-23T07:39:34Z"
    message: The deployment has Available status condition set to True
    reason: DeploymentAvailable
    status: "True"
    type: DeploymentAvailable
  - lastTransitionTime: "2024-10-23T07:39:34Z"
    message: Minimum replicas requirement is met
    reason: DeploymentMinimumReplicasMet
    status: "True"
    type: DeploymentReplicasMinAvailable
  - lastTransitionTime: "2024-10-23T08:02:21Z"
    message: All replicas are available
    reason: DeploymentReplicasAvailable
    status: "True"
    type: DeploymentReplicasAllAvailable
  - lastTransitionTime: "2024-10-23T08:02:21Z"
    message: Deployment is not actively rolling out
    reason: DeploymentNotRollingOut
    status: "False"
    type: DeploymentRollingOut
  - lastTransitionTime: "2024-10-23T07:38:51Z"
    message: The endpoint publishing strategy supports a managed load balancer
    reason: WantedByEndpointPublishingStrategy
    status: "True"
    type: LoadBalancerManaged
  - lastTransitionTime: "2024-10-23T07:39:23Z"
    message: The LoadBalancer service is provisioned
    reason: LoadBalancerProvisioned
    status: "True"
    type: LoadBalancerReady
  - lastTransitionTime: "2024-10-23T07:38:51Z"
    message: LoadBalancer is not progressing
    reason: LoadBalancerNotProgressing
    status: "False"
    type: LoadBalancerProgressing
  - lastTransitionTime: "2024-10-23T07:38:51Z"
    message: DNS management is supported and zones are specified in the cluster DNS
      config.
    reason: Normal
    status: "True"
    type: DNSManaged
  - lastTransitionTime: "2024-10-23T07:39:24Z"
    reason: NoFailedZones
    status: "True"
    type: DNSReady
  - lastTransitionTime: "2024-10-23T07:39:34Z"
    status: "True"
    type: Available
  - lastTransitionTime: "2024-10-23T08:02:21Z"
    status: "False"
    type: Progressing
  - lastTransitionTime: "2024-10-23T07:39:50Z"
    status: "False"
    type: Degraded
  - lastTransitionTime: "2024-10-23T07:38:51Z"
    message: IngressController is upgradeable.
    reason: Upgradeable
    status: "True"
    type: Upgradeable
  - lastTransitionTime: "2024-10-23T07:38:51Z"
    message: No evaluation condition is detected.
    reason: NoEvaluationCondition
    status: "False"
    type: EvaluationConditionsDetected
  domain: apps.ci-ln-z89svkb-72292.origin-ci-int-gce.dev.rhcloud.com
  endpointPublishingStrategy:
    loadBalancer:
      dnsManagementPolicy: Managed
      scope: External
    type: LoadBalancerService
  observedGeneration: 2
  selector: ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default
  tlsProfile:
    ciphers:
    - ECDHE-ECDSA-AES128-GCM-SHA256
    - ECDHE-RSA-AES128-GCM-SHA256
    - ECDHE-ECDSA-AES256-GCM-SHA384
    - ECDHE-RSA-AES256-GCM-SHA384
    - ECDHE-ECDSA-CHACHA20-POLY1305
    - ECDHE-RSA-CHACHA20-POLY1305
    - DHE-RSA-AES128-GCM-SHA256
    - DHE-RSA-AES256-GCM-SHA384
    - TLS_AES_128_GCM_SHA256
    - TLS_AES_256_GCM_SHA384
    - TLS_CHACHA20_POLY1305_SHA256
    minTLSVersion: VersionTLS12

2.
% oc get clusterversion
version   4.15.0-0.ci.test-2024-10-23-070517-ci-ln-z89svkb-latest   True        False         29m     Cluster version is 4.15.0-0.ci.test-2024-10-23-070517-ci-ln-z89svkb-latest
```
/label qe-approved
thanks

Miciah:
Thanks!
/approve
/lgtm

PR #1155, opened 2024-10-15
OCPBUGS-9037: Use cluster wildcard certificate for ingress canary

rfredette:
Utilize the existing ingress controller certificate management controller to also manage the certificate for the ingress canary, and use that certificate when serving the canary endpoint.


rfredette:
test failures appear unrelated.
/test e2e-gcp-operator
/test e2e-hypershift

candita (commented):
I've had these comments pending for some time.  Please let me know if I've misunderstood the assignment here, but in some places it looks like we are just reusing the default cert.  Was that the plan?

candita:
Failure in infra:
    --- FAIL: TestNodePool/HostedCluster2/Main/TestAdditionalTrustBundlePropagation (2271.02s)

/test e2e-hypershift

pkg/operator/controller/canary/daemonset.go
```diff
@@ -7,17 +7,23 @@ import (
...
+	"k8s.io/apimachinery/pkg/types"
 )
 
 // ensureCanaryDaemonSet ensures the canary daemonset exists
 func (r *reconciler) ensureCanaryDaemonSet() (bool, *appsv1.DaemonSet, error) {
-	desired := desiredCanaryDaemonSet(r.config.CanaryImage)
+	secretName, err := r.canarySecretName(controller.CanaryDaemonSetName().Namespace)
+	if err != nil {
+		return false, nil, err
+	}
```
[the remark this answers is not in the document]
Miciah: In order to enable `desiredCanaryDaemonSet` to get the secret, we would need to pass the client to `desiredCanaryDaemonSet`.  Passing the client to `desiredCanaryDaemonSet` would go against our established pattern that `ensureFoo` gets the necessary inputs for `desiredFoo` and passes those _specific_ inputs to `desiredFoo`.  Passing the client would also mean we would be adding a parameter to `desiredCanaryDaemonSet` anyway, so it wouldn't really have any benefit over the current change.

pkg/operator/controller/certificate/controller.go
```diff
@@ -105,6 +106,33 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+				daemonset := &appsv1.DaemonSet{}
+				err = r.client.Get(ctx, controller.CanaryDaemonSetName(), daemonset)
+				if err != nil {
+					if errors.IsNotFound(err) {
+						// All ingresses should have a deployment, so this one may not have been
+						// created yet. Retry after a reasonable amount of time.
+						log.Info("canary daemonset not found; will retry default cert sync")
+						result.RequeueAfter = 5 * time.Second
+					} else {
+						errs = append(errs, fmt.Errorf("failed to get daemonset: %v", err))
```
Miciah: Use `%w` when wrapping errors, on line 120 and on line 132.

pkg/operator/controller/certificate/controller.go
```diff
@@ -105,6 +107,48 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+						errs = append(errs, fmt.Errorf("failed to get daemonset: %w", err))
+					}
+				} else {
+					defaultCertName := controller.RouterEffectiveDefaultCertificateSecretName(ingress, deployment.Namespace)
+					defaultCert := &corev1.Secret{}
+					if err := r.client.Get(ctx, defaultCertName, defaultCert); err != nil {
+						errs = append(errs, fmt.Errorf("failed to get certificate for canary: %w", err))
+					}
+					canaryCert := defaultCert.DeepCopy()
+					canaryCertName := controller.RouterEffectiveDefaultCertificateSecretName(ingress, daemonset.Namespace)
```
Miciah: This should work, but is there a reason to copy the name from the effective certificate?  If you used a static name, you could get rid of `canarySecretName` in `pkg/operator/controller/canary/daemonset.go` and simplify the logic a bit.

Miciah: Well...  I wonder whether the issue referenced at https://github.com/openshift/cluster-ingress-operator/pull/1155#discussion_r2591642782 will prevent garbage collection from working properly?

PR #1156, opened 2024-10-15
OCPBUGS-43412:  Bump to k8s.io v0.31.1 (and deps)

EmilienM:
Bump the operator's vendored k8s.io/* packages to v0.31.1; bump its
vendored sigs.k8s.io/controller-runtime package to v0.19.0, the latest
release, which uses the k8s.io/* v0.31.1 packages.
This also bumps prometheus/common which needs to be updated to work with
k8s v0.31.1.

    go get k8s.io/api@v0.31.1 k8s.io/apimachinery@v0.31.1 k8s.io/apiextensions-apiserver@v0.31.1 k8s.io/apiserver@v0.31.1 k8s.io/client-go@v0.31.1 sigs.k8s.io/controller-runtime@v0.19.0 github.com/prometheus/common@v0.55.0
    go mod tidy
    go mod vendor

Update calls to route-metrics to adapt to breaking changes in
controller-runtime.  Namely, RateLimiter workqueue now uses
`NewTypedMaxOfRateLimiter` instead of ``NewMaxOfRateLimiter`.

This commit is related to [RFE-6242](https://issues.redhat.com//browse/RFE-6242).

https://issues.redhat.com/browse/RFE-6242

* go.mod: Bump.
* go.sum: Regenerate.
* pkg/operator/controller/route-metrics/controller.go: Update for the
  new controller-runtime.
* vendor/*: Regenerate.

PR #712, opened 2022-03-02
NE-585: Make ROUTER_BACKEND_CHECK_INTERVAL configurable

candita:
[NE-585](https://issues.redhat.com/browse/NE-585) Expose and make configurable ROUTER_BACKEND_CHECK_INTERVAL to customize the length of time between subsequent liveness checks on backends. Details in the [enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/haproxy-healthcheckinterval.md)

Refactor some unit tests and add e2e test.

PR #1159, opened 2024-10-22
NE-1790: Enable Dynamic Configuration Manager feature gate

alebedev87:
- Bumped openshift/api
- Enabled Dynamic Configuration Manager if corresponding feature is on.
- Added a new unsupported config override option `maxDynamicServers`, with a default value of 1, which is used only if the feature gate is on.
- Disabled DCM route blueprints by setting the pool size to 0.- Added a dedicated unit test for desiredRouterDeployemnt with DCM feature gate.


pkg/manifests/assets/router/deployment.yaml
```diff
@@ -17,6 +17,7 @@ spec:
         - name: router
           # image is set at runtime.
           imagePullPolicy: IfNotPresent
+          args: ["--blueprint-route-pool-size=0"]
```
Miciah, line 20: The operator generally uses environment variables rather than command-line switches.  Any reason not to use the `ROUTER_BLUEPRINT_ROUTE_POOL_SIZE` environment variable for consistency?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -568,12 +572,25 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 	}
 
 	dynamicConfigOverride := unsupportedConfigOverrides.DynamicConfigManager
-	if v, err := strconv.ParseBool(dynamicConfigOverride); err == nil && v {
+	if v, err := strconv.ParseBool(dynamicConfigOverride); (err == nil && v) || (err != nil && ingressControllerDCMEnabled) {
```
Miciah, line 575: This condition is a little tricky.  What do you think about simplifying it by using an intermediate variable?
```suggestion
	enableDCM := false
	if ingressControllerDCMEnabled {
		enableDCM = true
	}
	dynamicConfigOverride := unsupportedConfigOverrides.DynamicConfigManager
	if v, err := strconv.ParseBool(dynamicConfigOverride); err == nil {
		enableDCM = v
	}
	if enableDCM {
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -806,6 +809,54 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("invalid router Deployment: %v", err)
+	}
+	tests = []envData{
+		{"ROUTER_HAPROXY_CONFIG_MANAGER", false, ""},
+		{"ROUTER_MAX_DYNAMIC_SERVERS", false, "0"},
+	}
+	if err := checkDeploymentEnvironment(t, deployment, tests); err != nil {
+		t.Error(err)
+	}
```
Miciah, line 859: Long ago, we had just one gigantic `TestDesiredRouterDeployment` unit test.  This test became difficult to modify when we needed to expand test coverage, and so in https://github.com/openshift/cluster-ingress-operator/pull/712/commits/2849fcf835669b5c1f9d8ab21d6195287b0a5502, Candace started splitting that gigantic test into multiple tests.  We should continue in that direction: more, smaller, focused tests rather than fewer, larger, "all the things!" tests.  

What do you think about defining a new `TestDesiredRouterDeploymentDynamicConfigManager` test, perhaps using a list of test cases à la `TestDesiredRouterDeploymentDefaultPlacement`?

test/e2e/operator_test.go
```diff
@@ -3424,10 +3424,11 @@ func TestUnsupportedConfigOverride(t *testing.T) {
...
 
 	var tests = []struct {
-		name, unsupportedConfigOverride, env string
+		name, unsupportedConfigOverride, env, defaultValue, value string
 	}{
-		{"dynamic-config-manager", "dynamicConfigManager", "ROUTER_HAPROXY_CONFIG_MANAGER"},
-		{"contstats", "contStats", "ROUTER_HAPROXY_CONTSTATS"},
+		{"dynamic-config-manager", "dynamicConfigManager", "ROUTER_HAPROXY_CONFIG_MANAGER", "true", "false"},
+		{"contstats", "contStats", "ROUTER_HAPROXY_CONTSTATS", "", "true"},
+		{"max-dynamic-servers", "maxDynamicServers", "ROUTER_MAX_DYNAMIC_SERVERS", "2", "7"},
```
Miciah, line 3431: Should `defaultValue` for `ROUTER_MAX_DYNAMIC_SERVERS` be `"1"`?

test/e2e/operator_test.go
```diff
@@ -3424,10 +3424,11 @@ func TestUnsupportedConfigOverride(t *testing.T) {
 	t.Parallel()
 
 	var tests = []struct {
-		name, unsupportedConfigOverride, env string
+		name, unsupportedConfigOverride, env, defaultValue, value string
```
Miciah, line 3427: I think this would make the test easier to follow:  
```suggestion
		name, unsupportedConfigOverride, env, initialValue, updatedValue string
```

pkg/manifests/assets/router/deployment.yaml
alebedev87: Yes, the reason is that [`ROUTER_BLUEPRINT_ROUTE_POOL_SIZE` has a lower bound of `1`](https://github.com/openshift/cluster-ingress-operator/pull/1113#issuecomment-2289153747).

test/e2e/operator_test.go
alebedev87: Yes, it should. `ROUTER_HAPROXY_CONFIG_MANAGER` should have `false` default value too. These tests failed in CI:
```
        --- FAIL: TestAll/parallel/TestUnsupportedConfigOverride (306.33s)
            --- FAIL: TestAll/parallel/TestUnsupportedConfigOverride/dynamic-config-manager (111.43s)
            --- PASS: TestAll/parallel/TestUnsupportedConfigOverride/contstats (83.49s)
            --- FAIL: TestAll/parallel/TestUnsupportedConfigOverride/max-dynamic-servers (111.41s)
```

pkg/operator/controller/ingress/deployment_test.go
alebedev87: > What do you think about defining a new TestDesiredRouterDeploymentDynamicConfigManager test, perhaps using a list of test cases à la TestDesiredRouterDeploymentDefaultPlacement?

> Upd: This test already exists: TestUnsupportedConfigOverride.

At the end, I think the best way would be to make it a dedicated unit test indeed. Working on it.

pkg/manifests/assets/router/deployment.yaml
alebedev87: I can go ahead and modify the router's code to change the lower bound to `0`. This was done to minimize the changes however this may be inconsistent with the other configurations indeed.

pkg/operator/controller/ingress/deployment.go
alebedev87: Done.

test/e2e/operator_test.go
alebedev87: Updated.

alebedev87:
@Miciah : I made the changes you requested, can you please have another look?
Note that I will wait for https://github.com/openshift/cluster-ingress-operator/pull/1156 to merge before removing `WIP` in the title.

alebedev87:
```
        --- FAIL: TestAll/parallel/TestManagedDNSToUnmanagedDNSIngressController (222.23s)
        --- FAIL: TestAll/parallel/TestAWSLBSubnets (365.36s)
```

alebedev87:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator-techpreview openshift/origin/29180

alebedev87:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator-techpreview openshift/origin#29180

alebedev87:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-techpreview openshift/origin#29180

pkg/operator/controller/ingress/deployment_test.go
alebedev87: `TestDesiredRouterDeploymentDynamicConfigManager` unit test added.

alebedev87:
Added a commit for the k8s and openshift api bumps.

pkg/manifests/assets/router/deployment.yaml
alebedev87: Will be done in https://github.com/openshift/router/pull/635.

alebedev87: Done.

alebedev87:
https://github.com/openshift/cluster-ingress-operator/pull/1159#discussion_r1811510003 addressed.

alebedev87:
Rebased after https://github.com/openshift/cluster-ingress-operator/pull/1156 merge.

alebedev87:
/assign @Miciah 

Miciah already started the review.

ShudiLi:
tested it with 4.18.0-0.test-2024-11-15-011231-ci-ln-ffcj902-latest

```
1.
% oc get clusterversion
version   4.18.0-0.test-2024-11-15-011231-ci-ln-ffcj902-latest   True        False         3h44m   Cluster version is 4.18.0-0.test-2024-11-15-011231-ci-ln-ffcj902-latest

2.
% oc patch featuregates cluster -p '{"spec": {"featureSet": "TechPreviewNoUpgrade"}}' --type=merge
featuregate.config.openshift.io/cluster patched

3.
% go version -m ingress-operator | grep github.com/openshift/api
        dep     github.com/openshift/api        v3.9.1-0.20190924102528-32369d4db2ad+incompatible
        =>      github.com/openshift/api        v0.0.0-20241018083007-4f6053f954b0

4.
% oc -n openshift-ingress get deployment router-default -oyaml | grep -i -A1 -E "ROUTER_MAX_DYNAMIC_SERVERS|ROUTER_BLUEPRINT_ROUTE_POOL_SIZE"
        - name: ROUTER_BLUEPRINT_ROUTE_POOL_SIZE
          value: "0"
--
        - name: ROUTER_MAX_DYNAMIC_SERVERS
          value: "1"

5.
% oc -n openshift-ingress rsh router-default-85487f949f-drhnv
sh-5.1$ cat haproxy.config | grep -i  _dynamic-pod
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/lib/haproxy/router/cacerts/test:reencrypt1.pem
  server-template _dynamic-pod- 1-1 172.4.0.4:8765 check disabled
sh-5.1$

6. check the update of CRD dnsrecords
% oc get CustomResourceDefinition dnsrecords.ingress.operator.openshift.io -oyaml 
```

ShudiLi:
Ran some cases for the regression test
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/ginkgo-test/277505/consoleFull
11-15 15:29:43.371  error: 1 fail, 82 pass, 5 skip (37m13s)
reran the fail case and passed:
passed: (1m25s) 2024-11-15T07:37:13 "[sig-network-edge] Network_Edge Component_Router should ROSA-OSD_CCS-ARO-Author:hongli-High-73771-router can load secret"

/label qe-approved
thanks

alebedev87:
Merge conflict with `master` fixed.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -567,13 +573,36 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
 			Name:  RouterHAProxyConfigManager,
 			Value: "true",
+		}, corev1.EnvVar{
+			Name:  RouterHAProxyBlueprintRoutePoolSize,
+			Value: "0",
+		})
+
+		// Max dynamic servers override is only taken into account when DCM is enabled.
+		routerHAProxyMaxDynamicServersValue := RouterHAProxyMaxDynamicServersDefaultValue
+		if v, err := strconv.Atoi(unsupportedConfigOverrides.MaxDynamicServers); err == nil && v > 0 {
```
candita, line 597: It looks like this allows users to set it to a value > 1, which we don't yet support.  Should there be a validation to make sure it doesn't exceed 1?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1109,6 +1115,105 @@ func TestDesiredRouterDeploymentClientTLS(t *testing.T) {
...
+			dcmEnabled: true,
+			expectedEnv: []envData{
+				{"ROUTER_HAPROXY_CONFIG_MANAGER", true, "true"},
+				{"ROUTER_MAX_DYNAMIC_SERVERS", true, "1"},
+				{"ROUTER_BLUEPRINT_ROUTE_POOL_SIZE", true, "0"},
+			},
+		},
+		{
+			name:                       "featuregate-enabled-max-servers",
+			unsupportedConfigOverrides: `{"maxDynamicServers":"7"}`,
```
candita, line 1138: See https://github.com/openshift/cluster-ingress-operator/pull/1159/files#r1844266657

Do we want to allow more than 1 server without documenting the consequences?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1109,6 +1115,105 @@ func TestDesiredRouterDeploymentClientTLS(t *testing.T) {
...
+			},
+		},
+		{
+			name:                       "featuregate-disabled-dcm-override",
+			unsupportedConfigOverrides: `{"dynamicConfigManager":"true","maxDynamicServers":"7"}`,
+			dcmEnabled:                 false,
+			expectedEnv: []envData{
+				{"ROUTER_HAPROXY_CONFIG_MANAGER", true, "true"},
+				{"ROUTER_MAX_DYNAMIC_SERVERS", true, "7"},
+				{"ROUTER_BLUEPRINT_ROUTE_POOL_SIZE", true, "0"},
```
candita, line 1182: Are we going to allow users the over ride turn this **on** without the feature gate enabled?  I thought we were just going to allow them an over ride to turn it **off** after the feature gate was enabled.

candita:
@alebedev87 I just reviewed the enhancement proposal and it doesn't mention that 1) we would allow users to set more than 1 server (before we knew more about the consequences) - but if we did there has to be some maximum, or we could have users mistakenly setting this to 1000, for example; and 2) we would allow users to turn this feature on via override without the featuregate being enabled - I thought the point of the override was to allow users to turn it off if it caused an issue.

pkg/operator/controller/ingress/deployment.go
alebedev87: `1` offers the maximum benefit with minimal downside. However, there may be some "brave customers" who’d like to go beyond this limit. I wanted to provides some middle ground between the recommended default and experimental usage.

pkg/operator/controller/ingress/deployment_test.go
alebedev87: Right, we should document the consequences. It may be a little difficult for an unsupported option though. Let me think where I can put some words about the consequences.

pkg/operator/controller/ingress/deployment_test.go
alebedev87: I just kept it compatible with the current behavior. Currently (4.17 and before), we can enable DCM using `upsupportedConfigOverride` even though it would be not a wise decision as it has some gaps (which will be fixed in 4.18).

alebedev87:
> 1) we would allow users to set more than 1 server (before we knew more about the consequences) - but if we did there has to be some maximum, or we could have users mistakenly setting this to 1000, for example; 

Right, let me enforce an upper limit. I think `10` could be reasonable.

alebedev87:
Added the upper limit of `10` for the unsupported config option: `maxDynamicServers`.

alebedev87:
[Comment describing the impact of the usage of `maxDynamicServers`](https://github.com/openshift/cluster-ingress-operator/pull/1159/files#diff-03760f03eeac1174e38e1ec5fecadcab4cb38d05e00509f201d7ae64dc1c7c36R531-R539) is added.

alebedev87:
As discussed during the PR scrub, removed the upper bound limit for the `unsupportedConfigOverride.maxDynamicServers`. Some customers wanted to play with hundreds of dynamic slots.

alebedev87:

This PR does not introduce any change to the default featureset of OpenShift.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -121,7 +126,7 @@ func (r *reconciler) ensureRouterDeployment(ci *operatorv1.IngressController, in
 	if err != nil {
 		return false, nil, fmt.Errorf("failed to determine if proxy protocol is needed for ingresscontroller %s/%s: %v", ci.Namespace, ci.Name, err)
 	}
-	desired, err := desiredRouterDeployment(ci, r.config.IngressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig, r.config.RouteExternalCertificateEnabled)
+	desired, err := desiredRouterDeployment(ci, r.config.IngressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig, r.config.RouteExternalCertificateEnabled, r.config.IngressControllerDCMEnabled)
```
candita, line 129: nit: if there is anything else to change, I would suggest refactoring the `desiredRouterDeployment` signature to pass in `r.config` instead of three different parameters in `r.config`.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -523,6 +528,15 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
 		LoadBalancingAlgorithm string `json:"loadBalancingAlgorithm"`
 		DynamicConfigManager   string `json:"dynamicConfigManager"`
 		ContStats              string `json:"contStats"`
+		// maxDynamicServers specifies the number of dynamic server slots that will be added to each route.
+		// This setting can only be configured if dynamicConfigManager is `true`.
+		// The default value is 1, with a maximum of 10.
```
candita, line 533:
```suggestion
		// The default value is 1.
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -523,6 +528,15 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		// Dynamic server slots help to avoid reloads during endpoint scale-ups. The more dynamic servers
+		// added, the fewer reloads required when scaling up.
+		// Note, however, that dynamic servers consume memory even when not enabled.
+		// Use this analysis of the memory usage to assess the impact of different numbers of servers:
+		// https://gist.github.com/frobware/2b527ce3f040797909eff482a4776e0b
```
candita, line 538: nit: When this becomes suppported, we should add a link to an OpenShift repo.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2433,7 +2548,7 @@ func TestDesiredRouterDeploymentRouterExternalCertificate(t *testing.T) {
 		t.Error(err)
 	}
 
-	deployment, err = desiredRouterDeployment(ic, ingressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, false, false, nil, clusterProxyConfig, true)
+	deployment, err = desiredRouterDeployment(ic, ingressControllerImage, ingressConfig, infraConfig, apiConfig, networkConfig, false, false, nil, clusterProxyConfig, true, false)
```
candita, line 2551: There are no tests where the feature is `true`.  I think we need one.  Also, I'm curious what happens when `unsupportedConfigOverrides.DynamicConfigManager` is an empty string.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1109,6 +1115,115 @@ func TestDesiredRouterDeploymentClientTLS(t *testing.T) {
...
+		{
+			name:                       "featuregate-disabled-dcm-override",
+			unsupportedConfigOverrides: `{"dynamicConfigManager":"true","maxDynamicServers":"100"}`,
+			dcmEnabled:                 false,
+			expectedEnv: []envData{
+				{"ROUTER_HAPROXY_CONFIG_MANAGER", true, "true"},
+				{"ROUTER_MAX_DYNAMIC_SERVERS", true, "100"},
+				{"ROUTER_BLUEPRINT_ROUTE_POOL_SIZE", true, "0"},
+			},
+		},
```
candita, line 1194: Can you add a test to see what happens when `unsupportedConfigOverrides.DynamicConfigManager` is an empty string?

pkg/operator/controller/ingress/deployment.go
candita: It turns out we allow value > 1 in the unsupportedConfigOverride.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -523,6 +528,15 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, ingressController
...
+		MaxDynamicServers string `json:"maxDynamicServers"`
```
candita, line 539: This should probably be an int, not a string.  Users may be confused that they'd have to specify 10000 as "10000" here.

pkg/operator/controller/ingress/deployment.go
candita: Regarding converting string to int here, see https://github.com/openshift/cluster-ingress-operator/pull/1159/files#r1852667015

pkg/operator/controller/ingress/deployment_test.go
alebedev87: `featuregate-enabled-default-values` and `featuregate-enabled-max-servers` tests don't set `unsupportedConfigOverrides.dynamicConfigManager` which results into an `""` after the unmarshaling.

pkg/operator/controller/ingress/deployment.go
alebedev87: Well, I just followed the same pattern as for the other overrides. Two of them (`dynamicConfigManager`,`contStats`) could be `bool`s because `bool` is a valid JSON type.

pkg/operator/controller/ingress/deployment.go
alebedev87: Yes, this can be done in a separate PR.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -447,6 +447,8 @@ func Test_desiredRouterDeployment(t *testing.T) {
 		{"ROUTER_ERRORFILE_503", false, ""},
 		{"ROUTER_ERRORFILE_404", false, ""},
 		{"ROUTER_HAPROXY_CONFIG_MANAGER", false, ""},
+		{"ROUTER_MAX_DYNAMIC_SERVERS", false, ""},
+		{"ROUTER_BLUEPRINT_ROUTE_POOL_SIZE", false, ""},
```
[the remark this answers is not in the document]
alebedev87: For unit tests, I try to abstain from using the constants which are used in the code. This avoids repeating mistakes made in a code (typo etc.) and brings an additional check.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -84,6 +84,11 @@ const (
 
 	RouterHAProxyConfigManager = "ROUTER_HAPROXY_CONFIG_MANAGER"
 
+	RouterHAProxyMaxDynamicServers             = "ROUTER_MAX_DYNAMIC_SERVERS"
+	RouterHAProxyMaxDynamicServersDefaultValue = 1
+
+	RouterHAProxyBlueprintRoutePoolSize = "ROUTER_BLUEPRINT_ROUTE_POOL_SIZE"
```
candita, line 90: These two env variables are not found in our haproxy.config template yet and I don't see a PR that adds them.  Does that matter?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -84,6 +84,11 @@ const (
 
 	RouterHAProxyConfigManager = "ROUTER_HAPROXY_CONFIG_MANAGER"
 
+	RouterHAProxyMaxDynamicServers             = "ROUTER_MAX_DYNAMIC_SERVERS"
```
candita, line 87: Does there need to be any change for this in the flags https://github.com/openshift/router/blob/master/pkg/cmd/infra/router/template.go#L185?

pkg/operator/controller/ingress/deployment.go
alebedev87: They are not used in the config template, they are only for the configuration of the router's DCM behavior.

pkg/operator/controller/ingress/deployment.go
alebedev87: No. Why? This flag accepts an environment varible as an alternative way of using it. This one is limited to 1 as minimal value and defaults to 5. I set the envvar explicitly to 1 so no default value is used.

test/e2e/operator_test.go
```diff
@@ -3508,11 +3509,32 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+
 	var tests = []struct {
-		name, unsupportedConfigOverride, env string
+		name, unsupportedConfigOverride, env, defaultValue, updateValue, expectedValue string
 	}{
-		{"dynamic-config-manager", "dynamicConfigManager", "ROUTER_HAPROXY_CONFIG_MANAGER"},
-		{"contstats", "contStats", "ROUTER_HAPROXY_CONTSTATS"},
+		{"dynamic-config-manager", "dynamicConfigManager", "ROUTER_HAPROXY_CONFIG_MANAGER", dcmDefaultValue, dcmUpdateValue, dcmExpectedValue},
+		{"contstats", "contStats", "ROUTER_HAPROXY_CONTSTATS", "", "true", "true"},
+		{"max-dynamic-servers", "maxDynamicServers", "ROUTER_MAX_DYNAMIC_SERVERS", maxServersDefaultValue, maxServersUpdateValue, maxServersExpectedValue},
```
candita, line 3537: nit: We do have variable names for many of these hard-coded values.

test/e2e/operator_test.go
```diff
@@ -3508,11 +3509,32 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+		dcmExpectedValue        = "true"
+		maxServersDefaultValue  = ""
+		maxServersUpdateValue   = "2"
+		maxServersExpectedValue = "" // max dynamic servers cannot be set if DCM is off
+	)
+
+	if dcmEnabled, err := isFeatureGateEnabled(features.FeatureGateIngressControllerDynamicConfigurationManager); err != nil {
+		t.Fatalf("failed to get dynamic config manager feature gate: %v", err)
+	} else if dcmEnabled {
+		dcmDefaultValue = "true"
```
candita, line 3524: You may want to log something here to let us know this was run on a techpreview cluster.
```suggestion
                t.Logf("DynamicConfigurationManager feature gate is enabled for this test")
		dcmDefaultValue = "true"
```

test/e2e/operator_test.go
candita: Also, there's no test for the route blueprint default.

test/e2e/operator_test.go
```diff
@@ -3533,22 +3555,21 @@ func TestUnsupportedConfigOverride(t *testing.T) {
 			if err := kclient.Get(context.TODO(), controller.RouterDeploymentName(ic), deployment); err != nil {
 				t.Fatalf("failed to get ingresscontroller deployment: %v", err)
 			}
-			if err := waitForDeploymentEnvVar(t, kclient, deployment, 30*time.Second, tt.env, ""); err != nil {
-				t.Fatalf("expected initial deployment not to set %s=true: %v", tt.env, err)
+			if err := waitForDeploymentEnvVar(t, kclient, deployment, 30*time.Second, tt.env, tt.defaultValue); err != nil {
+				t.Fatalf("expected initial deployment not to set %s=%s: %v", tt.env, tt.defaultValue, err)
```
candita, line 3559: Even if the feature gate is enabled, we don't expect to see the max servers set to 1, the default?

pkg/operator/controller/ingress/deployment_test.go
candita: Sorry, I meant when the max dynamic servers is set to "" or something like "100K" or over the HAProxy limit for numbers.

pkg/operator/controller/ingress/deployment.go
candita: True, but I think we have precedent for string-based bools.  Choosing the string type for numbers allows malformatting, like "100K", which will fail later.

pkg/operator/controller/ingress/deployment.go
candita: Ok, but it's strange that they'd need to be env variables in that case.

pkg/operator/controller/ingress/deployment.go
candita: I noticed that you changed the minimum for BlueprintRoutePoolSize, and wondered why the default didn't need to change for MaxDynamicServers.

candita:
@alebedev87 All of my review comments can be handled in a followup PR unless you see something that has a greater impact than I assumed.

/approve

pkg/operator/controller/ingress/deployment.go
Miciah: In order to deliver an MVP in 4.18, we decided to turn off blueprint routes and pre-allocate 1 dynamic server per backend.  https://github.com/openshift/enhancements/pull/1687 explains the rationale.

pkg/operator/controller/ingress/deployment_test.go
Miciah: Unsupported config overrides should *not* be documented, at least not in product documentation for end-users.

pkg/operator/controller/ingress/deployment_test.go
Miciah: Unsupported config override is exactly what the name says.  If you use this, you're unsupported.  It's right there in the name, as explicit as we can make it without descending into absurdity (à la the `unsupported.do-not-use.openshift.io/override-liveness-grace-period-seconds` annotation).  If you use this option, you're overriding the operator and by extension the best judgment of the engineers who wrote the operator.  We're not going to second-guess you or encumber your (or ourselves!) with a bunch of validation or restrictions.  You're on your own, and you better know what you're doing if you use an unsupported config override.

pkg/operator/controller/ingress/deployment.go
Miciah: The operator usually use environment variables rather than command-line switches for openshift-router, for which reason I asked Andrey to use environment variables in this case too (https://github.com/openshift/cluster-ingress-operator/pull/1159#discussion_r1811510003).

test/e2e/operator_test.go
Miciah: Correct, `ROUTER_MAX_DYNAMIC_SERVERS` is only set if the dynamic config manager is enabled: https://github.com/openshift/cluster-ingress-operator/blob/9532e7874711548509cb73f1a394f485ce80d816/pkg/operator/controller/ingress/deployment.go#L594-L611

test/e2e/operator_test.go
alebedev87: > Also, there's no test for the route blueprint default.

Right, the test was made the way that only 1 variable is expected to be checked. I didn't have a goal to fully cover all configuration options in this e2e test, I just wanted to adapt the test to work with two new config overrides. The route blueprint configuration is well covered in the unit tests.

PR #1167, opened 2024-11-19
CORS-3755: AWS: Add Ingress LB IPs to Infra CR when in-cluster DNS is enabled

sadasu:
Add Ingress LB IPs to Infra CR and set DNS unmanaged when BYO DNS is enabled on the AWS platform.

Replicating the changes added for GCP as part of https://github.com/openshift/cluster-ingress-operator/pull/1016 for the AWS platform.


sadasu:
@gcs278, @candita could you PTAL? You had reviewed similar changes for the GCP platform.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1277,7 +1284,34 @@ func computeUpdatedInfraFromService(service *corev1.Service, infraConfig *config
 	if platformStatus == nil {
```
gcs278, line 1284: Can you update the GO Docs for `computeUpdatedInfraFromService` to reflect that it also AWS now:
```
// computeUpdatedInfraFromService updates GCP's PlatformStatus with Ingress LB IPs when the DNSType is `ClusterHosted`.
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -1277,7 +1284,34 @@ func computeUpdatedInfraFromService(service *corev1.Service, infraConfig *config
...
 		return false, fmt.Errorf("invalid PlatformStatus within Infrastructure config")
 	}
+	ingresses := service.Status.LoadBalancer.Ingress
+	ingressLBIPs := []configv1.IP{}
+	for _, ingress := range ingresses {
+		if len(ingress.IP) > 0 {
```
gcs278, line 1290: I don't think this will work as AWS doesn't set `status.loadBalancer.ingress.ip`, it only sets `status.loadBalancer.ingress.hostname`. AWS doesn't use IPs for their load balancer, only hostnames. GCP uses IP addresses.

Are you aware of this? The API design doesn't seem to accommodate hostnames for AWS https://github.com/openshift/api/pull/2046. Am I mising something?

pkg/operator/controller/ingress/controller.go
sadasu: Done.

pkg/operator/controller/ingress/controller.go
sadasu: Thanks for pointing this out. I have updated my fix to account for the fact that only the `Hostname` of the Ingress LB is available on AWS.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1271,13 +1279,47 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+			// The cluster has to run its own CoreDNS pod for DNS. Update Infra CR
+			// with the Ingress LB IPs. These values are used to configure the
+			// in-cluster DNS to provide resolution for *.apps.
+			if platformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted == nil {
+				platformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted = &configv1.CloudLoadBalancerIPs{}
+			}
+			ingresses := service.Status.LoadBalancer.Ingress
+			ingressLBIPs := []configv1.IP{}
+			for _, ingress := range ingresses {
+				ingressIPs, err := net.LookupIP(ingress.Hostname)
```
gcs278, line 1306: The AWS Load Balancer's IPs change frequently. Any time the router pods rollout, nodes scale up/down, or internal AWS maintenance. From what I saw, AWS recommends against using LB hostname's resolved IPs, as they can go stale for a number of reasons, and are not guaranteed to persist. The TTL is 60 seconds, so they could change as often as every 60 seconds, but realistically it's not that often. Even if the downstream components handle these updates, there could still be buggy behavior with gaps between IP changes  and CIO reconcilation.

Could you help me understand why we chose IPs over hostname for AWS in the infra config? If the downstream product is going to be CoreDNS serving this *.api wildcard, then can't CoreDNS serve the hostname as a CNAME record?

sadasu: @gcs278 We are aware of this limitation that the LB IP address could change during the life of the cluster. During the time of investigation, CoreDNS could not handle CNAME records. There are some untested workarounds suggested. Until we figure out a way to directly use the LB Hostname, in-cluster DNS for AWS in this form (using the IP corresponding to the LB Hostname), would be in tech-preview.

gcs278: > CoreDNS could not handle CNAME records.

Hmm yea, I see now after some testing of the CoreDNS file plugin. Regardless, the IP address needs to be stored because the CNAME is just another hostname, which needs to also be looked up. Darn, that's tricky. You might need to force the use of static EIPs to solve this (I assume you've also explored this).

> Until we figure out a way to directly use the LB Hostname, in-cluster DNS for AWS in this form (using the IP corresponding to the LB Hostname), would be in tech-preview.

So to be clear, we are going to try to figure this out for GA? And this will *hopefully* change? If so, do you mind putting a comment regard that here? We should make it clear that this is a known problem:

```suggestion
                                // Resolving the LoadBalancer's IPs is not ideal because they may change, but currently there is no better alternative.
				ingressIPs, err := net.LookupIP(ingress.Hostname)
```

sadasu: @gcs278 yes, this feature is definitely staying in TP until we figure out a better alternative. Proceeding in TP with this workaround is still useful. I still have some unanswered questions around using static EIPs. Happy to have those discussions to get to a better solution.
Yes, suggestion definitely helps.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1648,10 +1669,16 @@ func Test_computeUpdatedInfraFromService(t *testing.T) {
...
 		}
+		ingresses = []corev1.LoadBalancerIngress{
+			{IP: "196.78.125.4"},
+		}
 		ingressesWithMultipleIPs = []corev1.LoadBalancerIngress{
 			{IP: "196.78.125.4"},
 			{IP: "10.10.10.4"},
 		}
+		awsIngresses = []corev1.LoadBalancerIngress{
+			{Hostname: "196.78.125.4"},
```
gcs278, line 1680: It's probably worth a comment here so devs in the future aren't confused

```suggestion
                        // Hostname is intentionally assigned an IP address for unit testing purposes since net.LookupIP simply returns the provided IP.
			{Hostname: "196.78.125.4"},
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -476,8 +477,15 @@ func setDefaultPublishingStrategy(ic *operatorv1.IngressController, platformStat
 		// When the platform's default DNS solution cannot be used, set the DNSManagementPolicy
 		// accordingly. This feature is currently being implemented first for GCP. Will be
 		// extended to AWS and Azure platforms later.
```
gcs278, line 479:
```suggestion
		// accordingly. This feature is currently being implemented first for GCP and AWS. Will be
		// extended to the Azure platform later.
```

pkg/operator/controller/ingress/controller.go
```diff
@@ -1271,13 +1279,48 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+		cmpopts.SortSlices(func(a, b configv1.IP) bool {
+			return a < b
+		}),
+	}
 	switch platformStatus.Type {
+	case configv1.AWSPlatformType:
+		if platformStatus.AWS != nil && platformStatus.AWS.CloudLoadBalancerConfig != nil && platformStatus.AWS.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
+			// The cluster has to run its own CoreDNS pod for DNS. Update Infra CR
+			// with the Ingress LB IPs. These values are used to configure the
+			// in-cluster DNS to provide resolution for *.apps.
```
gcs278, line 1299: *nit* could we move this up to the above the switch statement, and remove the other duplicate comment in the GCP case?

```suggestion'
	// The cluster has to run its own CoreDNS pod for DNS. Update Infra CR
	// with the Ingress LB IPs. These values are used to configure the
	// in-cluster DNS to provide resolution for *.apps.
	switch platformStatus.Type {
	case configv1.AWSPlatformType:
		if platformStatus.AWS != nil && platformStatus.AWS.CloudLoadBalancerConfig != nil && platformStatus.AWS.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
```

gcs278:
@sadasu Do you mind putting a link to https://issues.redhat.com//browse/CORS-3755 in your second commit (the actual implementation)? Just a convenience for those trying to find the history.

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1710,6 +1737,34 @@ func Test_computeUpdatedInfraFromService(t *testing.T) {
...
+			expectUpdated: true,
+			expectError:   false,
+		},
+		{
+			description:   "aws platform with no change to LB IPs",
+			platform:      &awsPlatformWithLBIP,
+			ingresses:     awsIngresses,
+			expectUpdated: false,
+			expectError:   false,
+		},
```
gcs278, line 1767: To be complete, you could consider adding another case like:
```
		{
			description:   "gcp platform with DNSType and LB IP",
			platform:      &gcpPlatformWithLBIP,
			ingresses:     ingressesWithMultipleIPs,
			expectUpdated: true,
			expectError:   false,
		},
```

In which both the platform and Ingress specify IPs, but it's different changes. You just can't specify multiple Ingresses, but you can change the single IP value right?

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1710,6 +1737,34 @@ func Test_computeUpdatedInfraFromService(t *testing.T) {
...
 	}
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
```
gcs278, line 1770: Take a look at https://github.com/openshift/cluster-ingress-operator/pull/1167/commits/90693b70c8ddc4e2a8a815968d970d0817e33734#diff-bc23e1b7c8999c0b9674840022c285029528b1777b0fab5cc284974dfaa9cbebR1794, the unit test is still only checking GCP:
```
if !slices.Contains(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, configv1.IP(ingress.IP)) {
```

gcs278: Also, we are still assuming IP and not hostname [here](https://github.com/openshift/cluster-ingress-operator/pull/1167/commits/90693b70c8ddc4e2a8a815968d970d0817e33734#diff-bc23e1b7c8999c0b9674840022c285029528b1777b0fab5cc284974dfaa9cbebR1793):
```
if len(ingress.IP) > 0 {
```

pkg/operator/controller/ingress/controller_test.go
```diff
@@ -1736,8 +1807,13 @@ func Test_computeUpdatedInfraFromService(t *testing.T) {
...
-						if !slices.Contains(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, configv1.IP(ingress.IP)) {
-							t.Errorf("expected Infra CR to contain %s", ingress.IP)
+						if !reflect.DeepEqual(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, tc.expectedLBIPs) {
+							t.Errorf("expected Infra CR to contain %s but found %s", tc.expectedLBIPs, infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs)
+						}
+					}
+					if len(ingress.Hostname) > 0 {
+						if !reflect.DeepEqual(infraConfig.Status.PlatformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, tc.expectedLBIPs) {
+							t.Errorf("expected Infra CR to contain %s but found %s", tc.expectedLBIPs, infraConfig.Status.PlatformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs)
 						}
```
gcs278, line 1817: Instead of using the `status.loadBalancer.ingress.[ip|hostname]` to determine the platform, consider just using the platform provided in the test case. When we expand this to azure, this will have to be refactored because both GCP and Azure use `ingress.IP`.

Also, this code is somewhat confusing. We loop through each of the `status.loadBalancer.ingress`, but then in the `reflect.DeepEqual` we don't use the ingress variable. I don't think we need to check `if len(ingress.IP) > 0 {`, we can let the test compare if they are both `nil` too for better coverage (making sure it doesn't add an unwanted IP).

```suggestion
			if updated {
				switch tc.platform.Type {
				case configv1.AWSPlatformType:
					if !reflect.DeepEqual(infraConfig.Status.PlatformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, tc.expectedLBIPs) {
						t.Errorf("expected Infra CR to contain %s but found %s", tc.expectedLBIPs, infraConfig.Status.PlatformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs)
					}
				case configv1.GCPPlatformType:
					if !reflect.DeepEqual(infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs, tc.expectedLBIPs) {
						t.Errorf("expected Infra CR to contain %s but found %s", tc.expectedLBIPs, infraConfig.Status.PlatformStatus.GCP.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs)
					}
				}
			}
```

Let me know if that makes sense.

sadasu: Yes to both. Updated.

gcs278:
Thanks for the quick responses. I discussed with the team today, and we acknowledged the disadvantage of manually resolving the AWS load balancer hostname, but since this is restricted to tech preview, it's okay to move forward with this. Hopefully we can find a better design, but we will see how it works in tech preview.

/approve
/lgtm

sadasu:
ci/prow/verify job failed without the vendoring updates

gcs278:
Not totally sure how we usually handle go vendor pruning updates, I noticed your first commit fails `make test` because it's removing vendored files still in use.

In this scenario, I think the vendoring update should actually be co-located with the commit that removed the import for `"golang.org/x/exp/slices"`, so no commit is "broken". Feel free to fix it, but I don't think it's worth holding up the PR.

/lgtm

sadasu:
> Not totally sure how we usually handle go vendor pruning updates, I noticed your first commit fails `make test` because it's removing vendored files still in use.
> 
> In this scenario, I think the vendoring update should actually be co-located with the commit that removed the import for `"golang.org/x/exp/slices"`, so no commit is "broken". Feel free to fix it, but I don't think it's worth holding up the PR.
> 
> /lgtm

Yes, the 1st commit is changing the imports requiring the vendoring updates. I was following the installer team convention of separating out the vendoring changes into its own commit.

sadasu:
/cherry-pick release-4.18

I think this landed shortly after automatic forwarding of release-4.18 to master stopped. So, initiating a backport to release-4.18.

PR #1170, opened 2024-11-26
[WIP] OCPBUGS-31353: Minimize wildcard privileges for CRDs and namespaces

alebedev87:
- Scoped namespace permissions to read-only to enable informer watches.
- Restricted wildcard namespace permissions exclusively to operand namespaces (routers and canary).
- Scoped CRD permissions to read-only to enable informer watches.
- Restricted wildcard CRD permissions exclusively to Gateway API.

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -15,13 +15,31 @@ rules:
   - ""
   resources:
   - configmaps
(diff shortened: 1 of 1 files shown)


alebedev87:
https://github.com/openshift/cluster-ingress-operator/pull/1190 takes over from here.

PR #1171, opened 2024-11-26
[WIP] OCPBUGS-31353: Minimize wildcard privileges for secrets, configmaps and service accounts

alebedev87:
- Defined operand namespaces in the operator payload to be able to create local roles.
- Defined local roles for secrets, configmaps and service accounts in the operand namespaces.
- Defined a local role for the certificate management in openshift-config-manager namespace.


alebedev87:
https://github.com/openshift/cluster-ingress-operator/pull/1190 takes over from here.

PR #1172, opened 2024-12-04
[release-4.15] OCPBUGS-49392: Block Upgrades in release 4.15 for CA-Signed Certs Using SHA1

gcs278:
Previously, upgrades from 4.15 to 4.16 were only blocked for default leaf certs using SHA1. However, in 4.16, any SHA1 cert that is CA-signed (not self-signed) is unsupported. As a result, we were incorrectly allowing upgrades for SHA1 intermediate certificates, while blocking upgrades for self-signed SHA1 leaf certificates.

This fix blocks upgrades to 4.16 for IngressControllers with default certificates containing CA-signed SHA1 certs. Root CA certs using SHA1 are still supported.

**Additional Fix For Blocking 4.15-->4.16 Upgrades for DSA SHA1 Default Certs:**

Previously, upgrades were not being blocked when default certificates used DSA SHA1. Although DSA certificates are rejected at the router level by our validation logic, they can still be used in a default certificate. In 4.15, DSA SHA1 certificates are allowed, but in 4.16, they are rejected by OpenSSL. This fix adds an upgrade blocker in the rare case that a user was using a DSA SHA1 default certificate.

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -733,7 +733,7 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
 		// SHA1 can still be used by the root CA certificate.
 		if !isSelfSignedCert(cert) {
 			switch cert.SignatureAlgorithm {
-			case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
+			case x509.SHA1WithRSA, x509.ECDSAWithSHA1, x509.DSAWithSHA1:
 				return fmt.Errorf("a CA-signed certificate in secret %s/%s has weak SHA1 signature algorithm: %s (see https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-sha-haproxy-support-removed_release-notes for more details)", secret.Namespace, secret.Name, cert.SignatureAlgorithm)
 			}
 		}
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -1,6 +1,7 @@
 package ingress
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
 	"encoding/pem"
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -728,19 +729,63 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
...
-		// Prevent the upgrade only if the leaf certificate (the first certificate) uses a SHA1 signature algorithm.
-		// SHA1 can still be used by other certificates in the chain, such as the root and intermediate certificates.
-		if !cert.IsCA {
+		// Block the upgrade only if CA-signed certificates (not self-signed) uses a SHA1 signature algorithm.
+		// SHA1 can still be used by the root CA certificate.
+		if !isSelfSignedCert(cert) {
 			switch cert.SignatureAlgorithm {
-			case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
-				return fmt.Errorf("certificate in secret %s/%s has weak SHA1 signature algorithm: %s (see https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-sha-haproxy-support-removed_release-notes for more details)", secret.Namespace, secret.Name, cert.SignatureAlgorithm)
+			case x509.SHA1WithRSA, x509.ECDSAWithSHA1, x509.DSAWithSHA1:
```
Miciah, line 736: The commit message doesn't say anything about adding `x509.DSAWithSHA1`.  Is this PR actually fixing two separate defects (first, the use of `cert.IsCA` instead of checking the signer; and second, the missing `x509.DSAWithSHA1` case expression)?

Miciah:
Just to recap,

- #1014 added logic to check the signing algorithm on *every* certificate in the chain;
- #1154 changed that logic to check *only* the leaf certificate; and
- this PR changes that logic to checking every **CA**-signed (*not* **self**-signed) certificate in the chain (including any root, intermediates, and leaf).

Right?

pkg/operator/controller/ingress/status.go
gcs278: I added per this comment [thread](https://github.com/openshift/router/pull/642#discussion_r1911454965) for the 4.19 fix. It wasn't really fixing anything new, as DSA was already rejected by the router. I decided to add it to provide a more precise error message about SHA1 not being supported rather than a more obscure `Invalid value: "redacted key data": block PRIVATE KEY is not valid`.

For this 4.15 upgrade blocker, I carried it over for completeness/consistency, but I suppose that since it's already rejected, it's a bit pointless to block upgrades for it and may cause an unnecessary upgrade blocker for someone that already has a DSA SHA1 cert that is rejected. I don't think anyone should be using DSA SHA1 at this point, but I suppose its better to be safe.

I'll remove it.

gcs278: **Update**: Sorry for the confusion here. My answer above actually applies to https://github.com/openshift/router/pull/641 (I'll follow up over there). Here's what I should have said...

> Is this PR actually fixing two separate defects (first, the use of cert.IsCA instead of checking the signer; and second, the missing x509.DSAWithSHA1 case expression)?

Short Answer: Yes, upon further reflection, it is fixing two things, which I should mention in the commit message(s).

Long Answer: Originally, for this PR, I added DSA SHA1 without thinking much about it. However, I did some testing for DSA SHA1, and it's nuanced, but we DO actually need to block upgrades with DSA SHA1 for CIO. Here's why:
- While openshift-router (not HAProxy) rejects DSA certificates because it doesn't have [logic](https://github.com/openshift/router/blob/69f742bfaccb4ca002d3b4fbc0a2fc6ca66e1b9a/pkg/router/routeapihelpers/validation.go#L73-L88) to parse them, the CIO allows cluster admins to create DSA SHA1 default certificates
- However, you can't use the K8S TLS secret object because it can't parse the DSA key and you get an error (I guess K8S doesn't support it?) but you can still use an opaque/generic secret:
  - `oc create secret generic router-cert -n openshift-ingress --from-file=tls.key=dsa.key --from-file=tls.crt=dsa-combo.crt`
- I tested a 4.15 router image with a DSA-SHA1 default cert, and it worked
- I tested a 4.16 router image with a DSA-SHA1 default cert, and it failed

In summary, I was unaware there was a path for users to use DSA certs in the default certificate on the IngressController. I do think we need to block upgrades in CIO with DSA-SHA1 in this PR, but not in the router PR because the router has always rejected DSA certs. It's an odd situation.

gcs278:
> Just to recap,
> 
> * [[release-4.15] OCPBUGS-28928: Prevent upgrades for SHA1 default cert and SHA1 route certs  #1014](https://github.com/openshift/cluster-ingress-operator/pull/1014) added logic to check the signing algorithm on _every_ certificate in the chain;
> * [OCPBUGS-42480: Limit SHA-1 check to the leaf cert in default cert chain #1154](https://github.com/openshift/cluster-ingress-operator/pull/1154) changed that logic to check _only_ the leaf certificate; and
> * this PR changes that logic to checking every **CA**-signed (_not_ **self**-signed) certificate in the chain (including any ~root~, intermediates, and leaf, but not root because a root certificate is necessarily self-signed).
> 
> Right?

Yup exactly. An unfortunate failure(s) on my part, which convicted me to do better and actually dig in and understand what/how/why OpenSSL is doing the rejections rather than just trying to observe them from the outside. I found the HAProxy code that [throws the md too weak error](https://github.com/haproxy/haproxy/blob/8235a24782e528b9bf8ca9dd69c0a147556dfcb5/src/ssl_sock.c#L2669) and traced it into the OpenSSL library [skipping checks for self-signed certs](https://github.com/openssl/openssl/blob/master/ssl/t1_lib.c#L3648). And the fix here and in the router reflect how OpenSSL [determines](https://github.com/openssl/openssl/blob/b85e6f534906f0bf9114386d227e481d2336a0ff/crypto/x509/v3_purp.c#L558-L563) what constitutes a self-signed certificate.

pkg/operator/controller/ingress/status.go
gcs278: I split the fixes two commit so I can clearly document each.

I'm not against splitting it out into another PR and/or bug if recommended, but it feels closely related enough I thought to keep it together for now.

Miciah:
Thanks!
/approve
/lgtm

Thealisyed (commented):
What is the plan for communicating these upgrade-blocking changes to end-users to ensure they are aware and can take necessary actions?

gcs278:
> How do these changes impact the performance of the certificate validation process, given if it actually has any performance benchmarks?

Great question. We don't maintain a baseline of performance for really anything in the router, but what I have done before is create a benchmark GO test, and compared the old logic to the new logic.

I think performance is more of a concern with the router admitting routes vs. the CIO, but either way we still need to be cautious.

At one point, I did benchmarking for an iteration of this change (mainly benchmarking the `isSelfSigned` function) and the impact was minimal. The code is a couple of byte comparisons and an object comparison, all of which aren't heavy operations, so I feel pretty confident about the performance impact being negligible.

Now, at one point I was actually using the X509 validation function to validation each certificate against itself to detect if it was self-signed, and that was _not_ negligible. If that was my current implementation, I would be more cautious and provide some hard benchmarking numbers.

Thealisyed:
Timeout failures mainly, unrelated to changes
/retest

Thealisyed:
`failed to verify connectivity with workload with reqURL http://10.0.128.7/ using external client: timed out waiting for the condition`

/test e2e-gcp-operator

candita:
This is not actually a backport.  I'm trying a retitle to see if it fixes the requirement for backport-risk-assessed and cherry-pick-approved labels.
/retitle OCPBUGS-49392: Block Upgrades in release 4.15 for CA-Signed Certs Using SHA1

candita:
This doesn't require backport risk assessment or cherry-pick-approved labels because this isn't a backport.  Not clear why prow/tide is asking for it.

/label backport-risk-assessed

gcs278:
> This is not actually a backport. I'm trying a retitle to see if it fixes the requirement for backport-risk-assessed and cherry-pick-approved labels. /retitle OCPBUGS-49392: Block Upgrades in release 4.15 for CA-Signed Certs Using SHA1

Oh sorry, didn't see this comment before I renamed it back, but it is targeting release 4.15 so I thought [release-4.15] was appropriate in the title. In Jira-terms, at least the way I set it up, it is indeed a 4.15 backport of https://issues.redhat.com//browse/OCPBUGS-45290 (side note: https://github.com/openshift/router/pull/641 is a companion to this PR). 

This is the backport summary:
4.19: https://github.com/openshift/router/pull/642 (Fix SHA1 Routes Rejection)
4.18: https://github.com/openshift/router/pull/649 (Fix SHA1 Routes Rejection)
4.17: https://github.com/openshift/router/pull/650 (Fix SHA1 Routes Rejection)
4.16: https://github.com/openshift/router/pull/651 (Fix SHA1 Routes Rejection)
4.15: https://github.com/openshift/cluster-ingress-operator/pull/1172 & https://github.com/openshift/router/pull/641 (Fix SHA1 Routes Upgrade Blocker)

I used the 4.15 backport to add the upgrade blocking, because I viewed it as the same SHA1 bug, however, 4.15 needs a different code fix (block upgrades instead of reject the route). I suppose you could remove this bug's (https://issues.redhat.com/browse/OCPBUGS-49392) dependencies to not make it a backport if what I did doesn't make sense.

PR #1181, opened 2025-01-13
OCPBUGS-43745: Add support for IdleCloseTerminationPolicy (serial)

frobware:
This is the same as #1166, but the `Test_IdleConnectionTerminationPolicy` test runs in the serial suite.


frobware:
Solution moving forward in #1182.

PR #1186, opened 2025-01-24
OCPBUGS-43745: Skip Test_IdleConnectionTerminationPolicyDeferred when DCM feature gate is enabled

alebedev87:
The DCM enabled router returns 503 error after a service change on a route when idle-close-on-response option is enabled and the client reuses connections. Ref: https://issues.redhat.com/browse/OCPBUGS-48560.


alebedev87:

This PR follows up on the main PR for [OCPBUGS-43745](https://issues.redhat.com//browse/OCPBUGS-43745). Also, it doesn't impact the OCP payload in anyway.

PR #1152, opened 2024-10-04
Bump to OSSM 3.0 for Gateway API support

Miciah:
OSSM 3.0 replaces the OpenShift Service Mesh Operator [with Sail Operator](https://developers.redhat.com/articles/2023/08/30/istio-sail-operator-openshift), and the ServiceMeshControlPlane CRD is replaced by the Istio CRD.  Vendor the sail-operator API:

    go mod edit -replace github.com/imdario/mergo=github.com/imdario/mergo@v0.3.5
    go get github.com/istio-ecosystem/sail-operator/api/v1alpha1
    go mod tidy
    go mod vendor

Note that vendoring Sail Operator requires the mergo override.

Update the gatwayclass controller to create a subscription for Sail Operator instead of the Service Mesh Operator, and to create an Istio CR instead of a ServiceMeshControlPlane CR.

Currently, [OSSM 3.0 is Tech Preview](https://www.redhat.com/en/blog/red-hat-openshift-service-mesh-3-now-technology-preview), so we must configure the subscription accordingly.

PR #1192, opened 2025-02-20
NE-1953: Add Validating Admission Policy for Gateway API CRDs

alebedev87:
This PR introduces a validating admission policy that restricts modifications to CRDs from the "gateway.networking.k8s.io" group, allowing only the cluster ingress operator's service account to make changes.

This ensures that the core OpenShift retains exclusive ownership of the Gateway API implementation, preventing modifications from any add-ons (whether third-party or Red Hat-owned).

The policy and its corresponding binding will be managed by CVO.

### DIFF
--- a/manifests/01-validating-admission-policy-binding.yaml
+++ b/manifests/01-validating-admission-policy-binding.yaml
@@ -1,0 +1,13 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: openshift-ingress-operator-gatewayapi-crd-admission
+  annotations:
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: "TechPreviewNoUpgrade,CustomNoUpgrade"
+spec:
+  policyName: openshift-ingress-operator-gatewayapi-crd-admission
+  validationActions: [Deny]
--- a/manifests/01-validating-admission-policy.yaml
+++ b/manifests/01-validating-admission-policy.yaml
@@ -1,0 +1,36 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: openshift-ingress-operator-gatewayapi-crd-admission
+  annotations:
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: "TechPreviewNoUpgrade,CustomNoUpgrade"
+spec:
+  matchConstraints:
+    # Consider only requests to CRD resources.
+    resourceRules:
+    - apiGroups:   ["apiextensions.k8s.io"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE","UPDATE","DELETE"]
+      resources:   ["customresourcedefinitions"]
+  matchConditions:
+    # Consider only request to Gateway API CRDs.
+    - name: "check-only-gateway-api-crds"
+      # When the operation is DELETE, the "object" variable is null.
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
+  # Validations are evaluated in the the order of their declaration.
+  validations:
+    # Verify that the request was sent by the ingress operator's service account.
+    - expression: "has(request.userInfo.username) && (request.userInfo.username == 'system:serviceaccount:openshift-ingress-operator:ingress-operator')"
+      message: "Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be modified"
+      reason: Forbidden
+    # Verify that the request was sent from a pod. The presence of both "node" and "pod" claims implies that the token is bound to a pod object.
+    # Ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#schema-for-service-account-private-claims.
+    - expression: "has(request.userInfo.extra) && ('authentication.kubernetes.io/node-name' in request.userInfo.extra) && ('authentication.kubernetes.io/pod-name' in request.userInfo.extra)"
+      message: "this user must have both \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims"
+      reason: Forbidden
+  # Fail the admission if any validation evaluates to false.
+  failurePolicy: Fail
--- a/test/e2e/all_test.go
+++ b/test/e2e/all_test.go
@@ -83,7 +83,6 @@ func TestAll(t *testing.T) {
 		t.Run("TestSetRouteResponseHeaders", TestSetRouteResponseHeaders)
 		t.Run("TestReconcileInternalService", TestReconcileInternalService)
 		t.Run("TestConnectTimeout", TestConnectTimeout)
-		t.Run("TestGatewayAPI", TestGatewayAPI)
 		t.Run("TestAWSLBSubnets", TestAWSLBSubnets)
 		t.Run("TestUnmanagedAWSLBSubnets", TestUnmanagedAWSLBSubnets)
 		t.Run("TestAWSEIPAllocationsForNLB", TestAWSEIPAllocationsForNLB)
@@ -126,5 +125,16 @@ func TestAll(t *testing.T) {
 		t.Run("TestRouteHardStopAfterEnableOnIngressControllerHasPriorityOverIngressConfig", TestRouteHardStopAfterEnableOnIngressControllerHasPriorityOverIngressConfig)
 		t.Run("TestHostNetworkPortBinding", TestHostNetworkPortBinding)
 		t.Run("TestDashboardCreation", TestDashboardCreation)
+		// TestGatewayAPI creates a test ServiceMeshControlPlane (SMCP) resource,
+		// which triggers the creation of a mutating webhook for all pods in the cluster.
+		// This introduces a race condition where any pod creation request between
+		// the webhook's creation and the SMCP pod becoming ready can fail with:
+		//
+		// failed calling webhook "sidecar-injector.istio.io": failed to call webhook:
+		// no endpoints available for service "istiod-openshift-gateway"
+		//
+		// Serializing the test ensures it runs in isolation with other tests,
+		// preventing any impact of the mutating webhook on pod creation in the cluster
+		t.Run("TestGatewayAPI", TestGatewayAPI)
 	})
 }
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -59,8 +59,6 @@ var defaultRoutename = ""
 // feature gate is still in effect, preface the test names with "TestGatewayAPI"
 // so that they run via the openshift/release test configuration.
 func TestGatewayAPI(t *testing.T) {
-	t.Parallel()
-
 	// Skip if feature is not enabled
 	if gatewayAPIEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
 		t.Fatalf("error checking feature gate enabled status: %v", err)
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -10,14 +10,20 @@ import (
 	"testing"
 
 	"github.com/openshift/api/features"
+	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gatewayclass"
 
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/rest"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
@@ -30,6 +36,8 @@ const (
 	expectedCatalogSourceNamespace = "openshift-marketplace"
 	// The test gateway name used in multiple places.
 	testGatewayName = "test-gateway"
+	// gwapiCRDVAPName is the name of the ingress operator's Validating Admission Policy (VAP).
+	gwapiCRDVAPName = "openshift-ingress-operator-gatewayapi-crd-admission"
 )
 
 var crdNames = []string{
@@ -84,6 +92,7 @@ func TestGatewayAPI(t *testing.T) {
 	} else {
 		t.Log("Gateway API Controller not enabled, skipping testGatewayAPIObjects and testGatewayAPIIstioInstallation")
 	}
+	t.Run("testGatewayAPIResourcesProtection", testGatewayAPIResourcesProtection)
 }
 
 // testGatewayAPIResources tests that Gateway API Custom Resource Definitions are available.
@@ -167,6 +176,93 @@ func testGatewayAPIObjects(t *testing.T) {
 	}
 }
 
+// testGatewayAPIResourcesProtection verifies that the ingress operator's Validating Admission Policy
+// denies admission requests attempting to modify Gateway API CRDs on behalf of a user
+// who is not the ingress operator's service account.
+func testGatewayAPIResourcesProtection(t *testing.T) {
+	t.Helper()
+
+	// Get kube client which impersonates ingress operator's service account.
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+	kubeConfig.Impersonate = rest.ImpersonationConfig{
+		UserName: "system:serviceaccount:openshift-ingress-operator:ingress-operator",
+	}
+	kubeClient, err := operatorclient.NewClient(kubeConfig)
+	if err != nil {
+		t.Fatalf("failed to to create kube client: %v", err)
+	}
+
+	// Create test CRDs.
+	var testCRDs []*apiextensionsv1.CustomResourceDefinition
+	for _, name := range crdNames {
+		testCRDs = append(testCRDs, buildGWAPICRDFromName(name))
+	}
+
+	testCases := []struct {
+		name           string
+		kclient        client.Client
+		expectedErrMsg string
+	}{
+		{
+			name:           "Ingress operator service account required",
+			kclient:        kclient,
+			expectedErrMsg: "Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be modified",
+		},
+		{
+			name:           "Pod binding required",
+			kclient:        kubeClient,
+			expectedErrMsg: "this user must have both \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Verify that GatewayAPI CRD creation is forbidden.
+			for i := range testCRDs {
+				if err := tc.kclient.Create(context.Background(), testCRDs[i]); err != nil {
+					if !strings.Contains(err.Error(), tc.expectedErrMsg) {
+						t.Errorf("unexpected error received while creating CRD %q: %v", testCRDs[i].Name, err)
+					}
+				} else {
+					t.Errorf("admission error is expected while creating CRD %q but not received", testCRDs[i].Name)
+				}
+			}
+
+			// Verify that GatewayAPI CRD update is forbidden.
+			for i := range testCRDs {
+				crdName := types.NamespacedName{Name: testCRDs[i].Name}
+				crd := &apiextensionsv1.CustomResourceDefinition{}
+				if err := tc.kclient.Get(context.Background(), crdName, crd); err != nil {
+					t.Errorf("failed to get %q CRD: %v", crdName.Name, err)
+					continue
+				}
+				crd.Spec = testCRDs[i].Spec
+				if err := tc.kclient.Update(context.Background(), crd); err != nil {
+					if !strings.Contains(err.Error(), tc.expectedErrMsg) {
+						t.Errorf("unexpected error received while updating CRD %q: %v", testCRDs[i].Name, err)
+					}
+				} else {
+					t.Errorf("admission error is expected while updating CRD %q but not received", testCRDs[i].Name)
+				}
+			}
+
+			// Verify that GatewayAPI CRD deletion is forbidden.
+			for i := range testCRDs {
+				if err := tc.kclient.Delete(context.Background(), testCRDs[i]); err != nil {
+					if !strings.Contains(err.Error(), tc.expectedErrMsg) {
+						t.Errorf("unexpected error received while deleting CRD %q: %v", testCRDs[i].Name, err)
+					}
+				} else {
+					t.Errorf("admission error is expected while deleting CRD %q but not received", testCRDs[i].Name)
+				}
+			}
+		})
+	}
+}
+
 // ensureCRDs tests that the Gateway API custom resource definitions exist.
 func ensureCRDs(t *testing.T) {
 	t.Helper()
@@ -182,6 +278,19 @@ func ensureCRDs(t *testing.T) {
 // deleteCRDs deletes Gateway API custom resource definitions.
 func deleteCRDs(t *testing.T) {
 	t.Helper()
+
+	vm := newVAPManager(t, gwapiCRDVAPName)
+	// Remove the ingress operator's Validating Admission Policy (VAP)
+	// which prevents modifications of Gateway API CRDs
+	// by anything other than the ingress operator.
+	if err, recoverFn := vm.disable(); err != nil {
+		defer recoverFn()
+		t.Fatalf("failed to disable vap: %v", err)
+	}
+	// Put back the VAP to ensure that it does not prevent
+	// the ingress operator from managing Gateway API CRDs.
+	defer vm.enable()
+
 	for _, crdName := range crdNames {
 		err := deleteExistingCRD(t, crdName)
 		if err != nil {
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -19,6 +19,7 @@ import (
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -43,6 +44,10 @@ const (
 	openshiftIstiodDeploymentName = "istiod-openshift-gateway"
 	// openshiftSMCPName holds the expected OSSM ServiceMeshControlPlane name
 	openshiftSMCPName = "openshift-gateway"
+	// cvoNamespace is the namespace of cluster version operator (CVO).
+	cvoNamespace = "openshift-cluster-version"
+	// cvoDeploymentName is the name of cluster version operator's deployment.
+	cvoDeploymentName = "cluster-version-operator"
 )
 
 // updateIngressOperatorRole updates the ingress-operator cluster role with cluster-admin privilege.
@@ -146,6 +151,53 @@ func deleteExistingCRD(t *testing.T, crdName string) error {
 		return fmt.Errorf("timed out waiting for gatewayAPI CRD %s to be deleted: %v", crdName, err)
 	}
 	t.Logf("deleted crd %s", crdName)
+	return nil
+}
+
+// deleteExistingVAP deletes if the ValidatingAdmissionPolicy of the given name exists and returns an error if not.
+func deleteExistingVAP(t *testing.T, vapName string) error {
+	t.Helper()
+
+	vap := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+	newVAP := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+	name := types.NamespacedName{Name: vapName}
+
+	// Retrieve the object to be deleted.
+	if err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, name, vap); err != nil {
+			t.Logf("failed to get vap %q: %v, retrying ...", vapName, err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("failed to get vap %q: %w", vapName, err)
+	}
+
+	if err := kclient.Delete(context.Background(), vap); err != nil {
+		return fmt.Errorf("failed to delete vap %q: %w", vapName, err)
+	}
+
+	// Verify VAP was not recreated.
+	if err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, name, newVAP); err != nil {
+			if kerrors.IsNotFound(err) {
+				// VAP does not exist as expected.
+				return true, nil
+			}
+			t.Logf("failed to get vap %q: %v, retrying ...", vapName, err)
+			return false, nil
+		}
+		// Check if new VAP got recreated.
+		if newVAP != nil && newVAP.UID != vap.UID {
+			return true, fmt.Errorf("vap %q got recreated", vapName)
+		}
+		t.Logf("vap %q still exists, retrying ...", vapName)
+		return false, nil
+	}); err != nil {
+		return fmt.Errorf("failed to verify deletion of vap %q: %v", vapName, err)
+	}
+
+	t.Logf("deleted vap %q", vapName)
 	return nil
 }
 
@@ -270,6 +322,68 @@ func buildHTTPRoute(routeName, namespace, parentgateway, parentNamespace, hostna
 			Rules:           []gatewayapiv1.HTTPRouteRule{rule},
 		},
 	}
+}
+
+// buildGWAPICRDFromName initializes the GatewayAPI CRD deducing most of its required fields from the given name.
+func buildGWAPICRDFromName(name string) *apiextensionsv1.CustomResourceDefinition {
+	var (
+		plural   = strings.Split(name, ".")[0]
+		group, _ = strings.CutPrefix(name, plural+".")
+		scope    = apiextensionsv1.NamespaceScoped
+		// removing trailing "s"
+		singular = plural[0 : len(plural)-1]
+		versions = []map[string]bool{{"v1": true /*storage version*/}, {"v1beta1": false}}
+		kind     string
+	)
+
+	switch plural {
+	case "gatewayclasses":
+		singular = "gatewayclass"
+		kind = "GatewayClass"
+		scope = apiextensionsv1.ClusterScoped
+	case "gateways":
+		kind = "Gateway"
+	case "httproutes":
+		kind = "HTTPRoute"
+	case "referencegrants":
+		kind = "ReferenceGrant"
+		versions = []map[string]bool{{"v1beta1": true}}
+	}
+
+	crd := &apiextensionsv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: plural + "." + group,
+			Annotations: map[string]string{
+				"api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/gateway-api/pull/2466",
+			},
+		},
+		Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+			Group: group,
+			Names: apiextensionsv1.CustomResourceDefinitionNames{
+				Singular: singular,
+				Plural:   plural,
+				Kind:     kind,
+			},
+			Scope: scope,
+		},
+	}
+
+	for _, v := range versions {
+		for name, storage := range v {
+			crd.Spec.Versions = append(crd.Spec.Versions, apiextensionsv1.CustomResourceDefinitionVersion{
+				Name:    name,
+				Storage: storage,
+				Served:  true,
+				Schema: &apiextensionsv1.CustomResourceValidation{
+					OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+						Type: "object",
+					},
+				},
+			})
+		}
+	}
+
+	return crd
 }
 
 // assertSubscription checks if the Subscription of the given name exists and returns an error if not.
@@ -740,4 +854,84 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
 		return false, nil
 	})
 	return err
+}
+
+// assertVAP checks if the ValidatingAdmissionPolicy of the given name exists, and returns an error if not.
+func assertVAP(t *testing.T, name string) error {
+	t.Helper()
+	vap := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, types.NamespacedName{Name: name}, vap); err != nil {
+			t.Logf("failed to get vap %q: %v, retrying...", name, err)
+			return false, nil
+		}
+		t.Logf("found vap %q", name)
+		return true, nil
+	})
+}
+
+// scaleDeployment scales the deployment with the given name to the specified number of replicas.
+func scaleDeployment(t *testing.T, namespace, name string, replicas int32) error {
+	t.Helper()
+
+	nsName := types.NamespacedName{Namespace: namespace, Name: name}
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(context context.Context) (bool, error) {
+		depl := &appsv1.Deployment{}
+		if err := kclient.Get(context, nsName, depl); err != nil {
+			t.Logf("failed to get deployment %q: %v, retrying...", nsName, err)
+			return false, nil
+		}
+		if *depl.Spec.Replicas != replicas {
+			depl.Spec.Replicas = &replicas
+			if err := kclient.Update(context, depl); err != nil {
+				t.Logf("failed to update deployment %q: %v, retrying...", nsName, err)
+				return false, nil
+			}
+			t.Logf("scaled deployment %q to %d replica(s)", nsName, replicas)
+		}
+		if depl.Status.AvailableReplicas != replicas {
+			t.Logf("deployment %q expected to have %d available replica(s) but got %d, retrying...", nsName, replicas, depl.Status.AvailableReplicas)
+			return false, nil
+		}
+		t.Logf("deployment %q has %d available replica(s)", nsName, replicas)
+		return true, nil
+	})
+}
+
+// vapManager helps to disable the VAP resource which is managed by CVO.
+type vapManager struct {
+	t    *testing.T
+	name string
+}
+
+// newVAPManager returns a new instance of VAPManager.
+func newVAPManager(t *testing.T, vapName string) *vapManager {
+	return &vapManager{
+		t:    t,
+		name: vapName,
+	}
+}
+
+// disable scales down CVO and removes the VAP resource.
+func (m *vapManager) disable() (error, func()) {
+	if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 0); err != nil {
+		return fmt.Errorf("failed to scale down cvo: %w", err), func() { /*scale down didn't work, nothing to do*/ }
+	}
+	if err := deleteExistingVAP(m.t, m.name); err != nil {
+		return fmt.Errorf("failed to delete vap %q: %w", m.name, err), func() {
+			if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 1); err != nil {
+				m.t.Errorf("failed to scale up cvo: %v", err)
+			}
+		}
+	}
+	return nil, nil
+}
+
+// Enable scales up CVO and waits until the VAP is recreated.
+func (m *vapManager) enable() {
+	if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 1); err != nil {
+		m.t.Errorf("failed to scale up cvo: %v", err)
+	} else if err := assertVAP(m.t, m.name); err != nil {
+		m.t.Errorf("failed to find vap %q: %v", m.name, err)
+	}
 }
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e: generate
-	CGO_ENABLED=1 $(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	CGO_ENABLED=1 $(GO) test -timeout 1.5h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
 
 .PHONY: test-e2e-list
 test-e2e-list: generate


manifests/01-validating-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
+  # Validations are evaluated in the the order of their declaration.
+  validations:
+    # Verify that the request was sent by the ingress operator's service account.
+    - expression: "has(request.userInfo.username) && (request.userInfo.username == 'system:serviceaccount:openshift-ingress-operator:ingress-operator')"
+      message: "modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator"
+      reason: Forbidden
+    # Verify that the request was sent from a pod. The presence of both "node" and "pod" claims implies that the token is bound to a pod object.
+    # Ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#schema-for-service-account-private-claims.
+    - expression: "has(request.userInfo.extra) && ('authentication.kubernetes.io/node-name' in request.userInfo.extra) && ('authentication.kubernetes.io/pod-name' in request.userInfo.extra)"
```
alebedev87, line 26: Would the node claim alone be enough?

alebedev87:
/assign @JoelSpeed

To get a quick feedback on the shape of the VAP. I'm particularly interested in [this question](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1963230313).

manifests/01-validating-admission-policy.yaml
JoelSpeed: I suspect so, but I think this is extra safe and I don't see a need to remove the pod name check

JoelSpeed (commented):
At some point, you're going to want to debug why the CIO is broken. Running the CIO locally and connecting it to the cluster won't pass these checks, and will therefore fail.

What is the local development/debugging option once this is merged? Has that been considered?

Otherwise this LGTM

alebedev87:
> At some point, you're going to want to debug why the CIO is broken. Running the CIO locally and connecting it to the cluster won't pass these checks, and will therefore fail.
> 
> What is the local development/debugging option once this is merged? Has that been considered?

Right, CIO running outside the cluster won't be able to manage GWAPI CRDs after this PR is merged. Unless we scale down the CVO and remove the VAP or its binding which may be an option for debugging as we are supposed to be privileged enough for this.
I cannot talk for everybody for sure but from what I know the CIO debugging is always done inside a running cluster: we scale down the CVO and replace the CIO image with the one we test (sub-optiomal, I agree).

test/e2e/crd_admission_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/openshift/api/features"
+)
+
+// Test_GatewayAPICRDAdmissionOutsideCluster verifies that the ingress operator's ValidatingAdmissionPolicy
+// fails admission requests to modify Gateway API CRDs for a user which resides outside the cluster.
+func Test_GatewayAPICRDAdmissionOutsideCluster(t *testing.T) {
+	if gwapiEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
+		t.Fatalf("Failed to get Gateway API feature gate: %v", err)
+	} else if !gwapiEnabled {
+		t.Skipf("Skipping Test_GatewayAPICRDAdmissionOutsideCluster when GatewayAPI featuregate is not enabled")
```
shaneutt, line 24: The VAP is being deployed in the manifests alongside the CIO, right? So I wouldn't expect the feature gate to be required here: the VAP behavior should be present whether or not someone enables the GWAPI feature, as we are not intending to allow any other entity to manage these APIs. :thinking:

shaneutt (changes requested):

test/e2e/crd_admission_test.go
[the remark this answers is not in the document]
alebedev87: The life cycle of a feature gate is bound to the feature set it belongs to. There are 3 main sets (in the order of maturity): `DevPreview`, `TechPreview`, `Default`. A feature gate may belong to multiple sets but it gets installed by default on OCP cluster only when it gets into `Default` set. In that case, it becomes a part of the OCP payload but it doesn't mean that the feature gate disappears, the feature gate is still there and enabled. Since currently the gateway API (and OSSM) management is a TechPreview feature we would like VAP to be only seen on TechPreview clusters (not impacting the current OCP payload). However when `GatewayAPI` featuregate will be promoted to GA (end of 4.10 dev cycle) the management of GatewAPI CRDs will get into `Default` set however this test will still be working because the `GatewayAPI` feature gate will remain.

[OpenShift featuregates are different from Kubernetes ones](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1740512620181129?thread_ts=1740476756.655219&cid=CBWMXQJKD), they don't serve to toggle a piece of code on or off but rather aim at protecting the default OCP payload from immature code.

shaneutt (approved):

test/e2e/crd_admission_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	testCRDs := []*apiextensionsv1.CustomResourceDefinition{
+		// managed CRDs
+		makeGWAPICRD("gatewayclass", "GatewayClass"),
+		makeGWAPICRD("gateway", "Gateway"),
+		makeGWAPICRD("httproute", "HTTPRoute"),
+		makeGWAPICRD("referencegrant", "ReferenceGrant"),
+		// unmanaged CRDs
+		//makeGWAPICRD("tcproute", "TCPRoute"),
```
shaneutt, line 34: Technically we "manage" all CRDs in the group, at least to the effect that we don't allow anyone else to manage them.

Was the intention here to eventually test these other APIs? Or was it intentional to leave it as a comment to be illustrative for future code readers?

alebedev87: > Was the intention here to eventually test these other APIs?

Yes, that was the idea. Technically it's a little more complicated because I will have to disable the cluster version operator to be able to create some "unmanaged" CRDs. Also, this will not test any new "code path": the `.spec.group` check is the same as for the "managed" CRDs. So I'm not fully convinced it's worth the effort.

alebedev87:
Factorized the VAP management (removal and recovery) into a dedicated `struct`.

alebedev87:
```
=== RUN   TestAll/serial/TestCanaryRoute
    canary_test.go:129: failed to observe the expected canary response body: timed out waiting for the condition
```

I don't see a clear link of this PR to the serial `TestCanaryRoute`. Is it flaky? Let's retry.

alebedev87:
```
=== NAME  TestAll/parallel/TestHTTPCookieCapture
    operator_test.go:2887: failed to create pod openshift-ingress/cookietest: Internal error occurred: failed calling webhook "sidecar-injector.istio.io": failed to call webhook: Post "https://istiod-openshift-gateway.openshift-ingress.svc:443/inject?timeout=10s": no endpoints available for service "istiod-openshift-gateway"
. . .
=== NAME  TestAll/parallel/TestGatewayAPI/testGatewayAPIIstioInstallation
    util_gatewayapi_test.go:776: found ServiceMeshControlPlane openshift-ingress/openshift-gateway, but it isn't ready. Retrying...
=== NAME  TestAll/parallel/TestGatewayAPI/testGatewayAPIIstioInstallation
    util_gatewayapi_test.go:776: found ServiceMeshControlPlane openshift-ingress/openshift-gateway, but it isn't ready. Retrying...
=== NAME  TestAll/parallel/TestGatewayAPI/testGatewayAPIIstioInstallation
    util_gatewayapi_test.go:780: found ServiceMeshControlPlane openshift-ingress/openshift-gateway with ready components: [istio-discovery istio-egress istio-ingress mesh-config telemetry-common]
. . .
--- FAIL: TestAll/parallel/TestHTTPCookieCapture (82.67s)
```

The mutating webhook created for the openshift's SMCP was failing because the SMCP pod was not ready.

I think that it's caused by `TestGatewayAPI` which is run in parallel. SMCP can be created at any time adding the mutating webhook for all pods in the cluster. This poses a problem to any pod created by any other parallel test for the amount of time required by the SMCP pod to become ready.

lihongan:
Did some pre-merge tests and looks CRD delete/update/create operations are well managed, but the VAP can be deleted even they are re-created later. During the period I could create v1.2.1 CRDs on the cluster. 
```
$ oc delete crd gatewayclasses.gateway.networking.k8s.io
Error from server (Forbidden): customresourcedefinitions.apiextensions.k8s.io "gatewayclasses.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator

$ oc annotate crds referencegrants.gateway.networking.k8s.io gateway.networking.k8s.io/bundle-version=v1.1.1 --overwrite
Error from server (Forbidden): customresourcedefinitions.apiextensions.k8s.io "referencegrants.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator

### try to create v1.2.1 CRDs which contains grpcroutes CRD
$  kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml
<...snip...>
Error from server (Forbidden): error when creating "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml": customresourcedefinitions.apiextensions.k8s.io "grpcroutes.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator

$ oc delete validatingadmissionpolicies.admissionregistration.k8s.io openshift-ingress-operator-gatewayapi-crd-admission 
validatingadmissionpolicy.admissionregistration.k8s.io "openshift-ingress-operator-gatewayapi-crd-admission" deleted

$ oc delete validatingadmissionpolicybindings.admissionregistration.k8s.io openshift-ingress-operator-gatewayapi-crd-admission 
validatingadmissionpolicybinding.admissionregistration.k8s.io "openshift-ingress-operator-gatewayapi-crd-admission" deleted
```

alebedev87:
Increased the e2e test timeout to `1.5h` because the `TestGatewayAPI` run serially adds ~10 mins to the total execution which is too close to `1h` limit.

JoelSpeed:
> but the VAP can be deleted even they are re-created later. During the period I could create v1.2.1 CRDs on the cluster.

If users are deleting the VAPs to get around the mechanism, then that's a problem. Our users shouldn't be deleting the VAP, and I'm certain we could push back on them if we were triaging a support case and could see that the VAP had recently been recreated

Have had this discussion before in arch calls, we shouldn't need to protect our protection, users can delete MWC/VWC already today

alebedev87:
@lihongan : [As discussed in Slack](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1741150341406379), we cannot prevent privileged users from modifying GWAPI CRDs (adding unwanted kinds or changing managed versions). This puts the user into the unsupported territory which we cannot control.

alebedev87:
/test e2e-azure-operator

```
INFO[2025-03-05T09:41:34Z] Step e2e-gcp-operator-ipi-install-install failed after 2m47s.
```

knobunc (commented):
/lgtm

lihongan:
/label qe-approved
the pre-merge test results are expected

manifests/01-validating-admission-policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+      message: "this user must have a \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims"
```
danwinship, line 33: Why do we care about validating that the request came from a pod? What's the threat model? If someone can steal the service account token, they can presumably create a pod as well.

If we do need this check, the message is unhelpful (both to users and to future developers trying to debug failures.)

manifests/01-validating-admission-policy-binding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: openshift-ingress-operator-gatewayapi-crd-admission
+  annotations:
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: "TechPreviewNoUpgrade,CustomNoUpgrade"
```
candita, line 10: We don't use this annotation in other manifest yaml.  Is it temporary and will be removed before GA?

test/e2e/gateway_api_test.go
```diff
@@ -51,8 +55,6 @@ var defaultRoutename = ""
 // feature gate is still in effect, preface the test names with "TestGatewayAPI"
 // so that they run via the openshift/release test configuration.
 func TestGatewayAPI(t *testing.T) {
-	t.Parallel()
```
candita, line 54: We discussed in slack that making this serial should not be necessary.

test/e2e/gateway_api_test.go
```diff
@@ -158,6 +161,60 @@ func testGatewayAPIObjects(t *testing.T) {
...
 }
 
+// testGatewayAPIResourcesProtection verifies that the ingress operator's Validating Admission Policy
+// denies admission requests attempting to modify Gateway API CRDs on behalf of a user
+// who is not the ingress operator's service account.
+func testGatewayAPIResourcesProtection(t *testing.T) {
+	t.Helper()
+
+	var testCRDs []*apiextensionsv1.CustomResourceDefinition
+	for _, name := range crdNames {
```
candita, line 171: Need to update `crdNames` with GRPCRoute.  It would be beneficial to put the `crdNames` declaration in a shared location. cc @Thealisyed @grzpiotrowski

test/e2e/gateway_api_test.go
```diff
@@ -158,6 +161,60 @@ func testGatewayAPIObjects(t *testing.T) {
...
+		testCRDs = append(testCRDs, buildGWAPICRDFromName(name))
+	}
+	expectedErrMsg := "modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator"
+
+	// Verify that GatewayAPI CRD creation is forbidden.
+	for i := range testCRDs {
+		if err := kclient.Create(context.Background(), testCRDs[i]); err != nil {
+			if !strings.Contains(err.Error(), expectedErrMsg) {
```
candita, line 179: I'd like to see a more precise error message comparison.  It would be better for our docs and troubleshooting to add a  human-readable `messageExpression` to each of the `validations`.

test/e2e/util_gatewayapi_test.go
```diff
@@ -741,3 +858,78 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
...
+	return &vapManager{
+		t:    t,
+		name: vapName,
+	}
+}
+
+// disable scales down CVO and removes the VAP resource.
+func (m *vapManager) disable() (error, func()) {
+	if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 0); err != nil {
+		return fmt.Errorf("failed to scale down cvo: %w", err), func() { /*scale down didn't work, nothing to do*/ }
```
candita, line 916: Replace %w with %v here and elsewhere.
```suggestion
		return fmt.Errorf("failed to scale down cvo: %v", err), func() { /*scale down didn't work, nothing to do*/ }
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -272,6 +324,71 @@ func buildHTTPRoute(routeName, namespace, parentgateway, parentNamespace, hostna
...
+			Group: group,
+			Names: apiextensionsv1.CustomResourceDefinitionNames{
+				Singular: singular,
+				Plural:   plural,
+				Kind:     kind,
+			},
+			Scope: scope,
+			Versions: []apiextensionsv1.CustomResourceDefinitionVersion{
+				{
+					Name:    "v1",
```
candita, line 368: TIL that not all the standard CRDs are v1, so this isn't an accurate "fake" CRD generator.  You'll need to get the definitions from https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/standard.  `ReferenceGrants` is still v1beta1.

manifests/01-validating-admission-policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+metadata:
+  name: openshift-ingress-operator-gatewayapi-crd-admission
+  annotations:
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/feature-set: "TechPreviewNoUpgrade,CustomNoUpgrade"
+spec:
+  matchConstraints:
```
candita, line 12: The godocs on matchConstraints are not very good.  

> // The MutatingAdmissionPolicy cares about a request if it matches _all_ Constraint.
> // However, in order to prevent clusters from being put into an unstable state that cannot be recovered from via the API
> // ValidatingAdmissionPolicy cannot match ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding.

I looked into the Kubernetes reference and found in https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#type-checking, this:

> Type Checking does not apply to CRDs, including matched CRD types and reference of paramKind. The support for CRDs will come in future release.

Does it apply to us?  No issues with adding this VAP?

manifests/01-validating-admission-policy.yaml
Miciah: The purpose of the VAP is to prevent the cluster-admin or some other controller from doing an invalid update.  

Conceptually, we want these CRDs to be like core APIs.  In practice, this means (among other things) blocking the cluster-admin's attempts to delete or modify the CRDs.  The actor in this threat model isn't a malicious individual but rather a cluster-admin who doesn't know (or controller written or deployed by someone who doesn't know) that the platform owns these CRDs.  

As Joel noted in a comment, the cluster-admin can always find a way to subvert the platform's protections.

manifests/01-validating-admission-policy-binding.yaml
Miciah: No, it will be included in GA.

test/e2e/util_gatewayapi_test.go
Miciah: Why would we use `%v` instead of `%w` here?  This looks like an ordinary instance of wrapping an error value with [`Errorf`](https://pkg.go.dev/fmt#Errorf).

manifests/01-validating-admission-policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+  matchConditions:
+    # Consider only request to Gateway API CRDs.
+    - name: "check-only-gateway-api-crds"
+      # When the operation is DELETE, the "object" variable is null.
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
+  # Validations are evaluated in the the order of their declaration.
+  validations:
+    # Verify that the request was sent by the ingress operator's service account.
+    - expression: "has(request.userInfo.username) && (request.userInfo.username == 'system:serviceaccount:openshift-ingress-operator:ingress-operator')"
+      message: "modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator"
```
candita, line 28:
```suggestion
      message: "Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be customized"
```

manifests/01-validating-admission-policy.yaml
candita, line 33: nit, if we do need to check this:
```suggestion
      message: "unauthorized user"
```

Miciah:
> I think that it's caused by `TestGatewayAPI` which is run in parallel. SMCP can be created at any time adding the mutating webhook for all pods in the cluster. This poses a problem to any pod created by any other parallel test for the amount of time required by the SMCP pod to become ready.

I missed this detail in our earlier discussion.  So the `TestGatewayAPI` test causes the webhook to be initialized, the webhook's initialization takes a significant amount of time, and during this time, the webhook blocks the initialization of pods that other tests create, causing these tests to time out while waiting for the pod initialization and fail—is that what you are saying?  In that case, it makes sense to me to make the test serial.  

Please be sure to update the test's godoc to explain why it needs to be run serially.  (All tests that run serially are supposed to have such an explanatory comment.)

manifests/01-validating-admission-policy.yaml
Miciah: I think that "matched CRD types" refers to the type that is defined by the CRD, not to the CRD type itself.  That is, I expect that the VAP can block updates to, say, the gatewayclasses CRD, but it cannot at this time block updates to an arbitrary gatewayclass CR.

manifests/01-validating-admission-policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+      reason: Forbidden
+    # Verify that the request was sent from a pod. The presence of both "node" and "pod" claims implies that the token is bound to a pod object.
+    # Ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#schema-for-service-account-private-claims.
+    - expression: "has(request.userInfo.extra) && ('authentication.kubernetes.io/node-name' in request.userInfo.extra) && ('authentication.kubernetes.io/pod-name' in request.userInfo.extra)"
+      message: "this user must have a \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims"
+      reason: Forbidden
```
candita, line 34:
```suggestion
      reason: Unauthorized
```

manifests/01-validating-admission-policy-binding.yaml
Miciah: Oh, wait, I thought we could associate the manifest with a specific featuregate, which is what we really want to do.

manifests/01-validating-admission-policy.yaml
alebedev87: > Does it apply to us? No issues with adding this VAP?

The type checking will be missing for this VAP. Meaning that when the VAP will be updated and if it will contain any syntax mistakes in validation expressions (or variables) the update of VAP will be blocked and the syntax error will be reported in the VAP's status. I don't think it's a blocking gap because we will validate our VAP before it will hit any cluster (local testing, integration testing, QE testing).

alebedev87: > The godocs on matchConstraints are not very good.
> 
> >     // The MutatingAdmissionPolicy cares about a request if it matches all Constraint.
> >     // However, in order to prevent clusters from being put into an unstable state that cannot be recovered from via the API
> >     // ValidatingAdmissionPolicy cannot match ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding.

You mean the part about MutatingAdmissionPolicy? Indeed, it's unclear but should not impact us in anyway as we are adding VAP.
The otherpart about the ValidatingAdmissionPolicy is a known constraint which does not impact us either because we don't need to match VAPs or VAP bindings.

test/e2e/util_gatewayapi_test.go
```diff
@@ -272,6 +324,71 @@ func buildHTTPRoute(routeName, namespace, parentgateway, parentNamespace, hostna
...
+		singular = "gatewayclass"
+		kind = "GatewayClass"
+		scope = apiextensionsv1.ClusterScoped
+	case "gateways":
+		kind = "Gateway"
+	case "httproutes":
+		kind = "HTTPRoute"
+	case "referencegrants":
+		kind = "ReferenceGrant"
+	}
```
candita, line 349: Also `GRPCRoute`.  Is it in our best interest to implement this function, knowing it has to change in the future every time a new CRD is added or the version changed?

test/e2e/util_gatewayapi_test.go
```diff
@@ -741,3 +858,78 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
 	})
 	return err
 }
+
+// assertVAP checks if the VAP of the given name exists, and returns an error if not.
```
candita, line 862: nit:
```suggestion
// assertVAP checks if the ValidatingAdmissionPolicy of the given name exists, and returns an error if not.
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -741,3 +858,78 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
...
+			return false, nil
+		}
+		if *depl.Spec.Replicas != replicas {
+			depl.Spec.Replicas = &replicas
+			if err := kclient.Update(context, depl); err != nil {
+				t.Logf("failed to update deployment %q: %v, retrying...", nsName, err)
+				return false, nil
+			}
+		}
+		t.Logf("scaled  deployment %q to %d replica(s)", nsName, replicas)
```
candita, line 894: You may want to check that the replicas actually deployed.

alebedev87:
> > I think that it's caused by TestGatewayAPI which is run in parallel. SMCP can be created at any time adding the mutating webhook for all pods in the cluster. This poses a problem to any pod created by any other parallel test for the amount of time required by the SMCP pod to become ready.
> 
> I missed this detail in our earlier discussion. So the TestGatewayAPI test causes the webhook to be initialized, the webhook's initialization takes a significant amount of time, and during this time, the webhook blocks the initialization of pods that other tests create, causing these tests to time out while waiting for the pod initialization and fail—is that what you are saying? In that case, it makes sense to me to make the test serial.

Not exactly. The webhook gets created by `TestGatewayAPI`, yes, as a consequence of `SMCP` resource creation. This webhook forwards requests to the control plane pod (`istiod`). However the control plane pod may take some time to become ready. So during this time (between the MWC is created and the control plane pod becomes ready) any other parallel test may try to create a pod which would result into:
```
failed to create pod openshift-ingress/cookietest: Internal error occurred: failed calling webhook "sidecar-injector.istio.io": failed to call webhook: Post "[https://istiod-openshift-gateway.openshift-ingress.svc:443/inject?timeout=10s](https://istiod-openshift-gateway.openshift-ingress.svc/inject?timeout=10s)": no endpoints available for service "istiod-openshift-gateway"
```

manifests/01-validating-admission-policy-binding.yaml
Miciah: After reviewing https://github.com/openshift/enhancements/blob/master/enhancements/update/cvo-techpreview-manifests.md (and looking through the CVO code), I see that there **isn't** an option to associate a manifest with a feature*gate* rather than a feature*set*.  

It looks like the purpose of this annotation was to make it easier to introduce an entire operator as tech preview.  Our use of it, to introduce a specific resource for a tech-preview feature implemented in a GA operator, isn't exactly what the annotation was intended for.  

That said, it seems like a valid use.  However, it does mean that we will need to remove the annotation at the same time that we graduate the featuregate to GA.  (This seems to me like a major design flaw because it means that the feature cannot be atomically promoted and cannot be completely turned off once it goes into the default featureset, but ¯\\\_(ツ)\_/¯.)

manifests/01-validating-admission-policy.yaml
alebedev87: I think it's debatable. I followed the logic of the corresponding HTTP codes where `401 Unauthorized` means more `I don’t recognize you` while `403 Forbidden` means more `I know who you are, but you're not allowed here`.

manifests/01-validating-admission-policy.yaml
candita: What does that other part even mean?  I can't understand the grammar.

test/e2e/util_gatewayapi_test.go
alebedev87: Right, [the standard way to wrap errors after go1.13](https://go.dev/blog/go1.13-errors#wrapping-errors-with-w).

Miciah:
> > > I think that it's caused by TestGatewayAPI which is run in parallel. SMCP can be created at any time adding the mutating webhook for all pods in the cluster. This poses a problem to any pod created by any other parallel test for the amount of time required by the SMCP pod to become ready.
> > 
> > 
> > I missed this detail in our earlier discussion. So the TestGatewayAPI test causes the webhook to be initialized, the webhook's initialization takes a significant amount of time, and during this time, the webhook blocks the initialization of pods that other tests create, causing these tests to time out while waiting for the pod initialization and fail—is that what you are saying? In that case, it makes sense to me to make the test serial.
> 
> Not exactly. The webhook gets created by `TestGatewayAPI`, yes, as a consequence of `SMCP` resource creation. This webhook forwards requests to the control plane pod (`istiod`). However the control plane pod may take some time to become ready. So during this time (between the MWC is created and the control plane pod becomes ready) any other parallel test may try to create a pod which would result into:
> 
> ```
> failed to create pod openshift-ingress/cookietest: Internal error occurred: failed calling webhook "sidecar-injector.istio.io": failed to call webhook: Post "[https://istiod-openshift-gateway.openshift-ingress.svc:443/inject?timeout=10s](https://istiod-openshift-gateway.openshift-ingress.svc/inject?timeout=10s)": no endpoints available for service "istiod-openshift-gateway"
> ```

I'm not sure what nuance I'm missing (why "not exactly"?).  The important thing is that we understand why `TestGatewayAPI` disrupts other tests when run in parallel, the test is indeed is doing something that interferes with other tests, we understand what that something is, and that something is completely valid for the test to do.  It isn't a defect in the test itself but rather an implication of how webhooks work.  The test is doing something valid that happens to affect the global state of the cluster in a way that impacts other tests.  This is exactly the kind of scenario in which we legitimately move a test to the serial list.

Miciah:
> I'm not sure what nuance I'm missing (why "not exactly"?).

Hm, I guess the nuance is that the webhook is technically initialized, but the webhook rejects pods because Istiod isn't ready?

alebedev87:
> Hm, I guess the nuance is that the webhook is technically initialized, but the webhook rejects pods because Istiod isn't ready?

Right, plus you mentioned that other tests (which fail to create pods) time out. What I saw is that the tests just `Fatal` exit when their pods cannot be created.

manifests/01-validating-admission-policy-binding.yaml
alebedev87: Right, that's a flaw which prevents the atomic promotion and even worse it can block the promotion. If there are some tests for the feature gate in  origin repository. @JoelSpeed started [a thread](https://redhat-internal.slack.com/archives/CEGKQ43CP/p1740583567370779) for this with CVO people.

manifests/01-validating-admission-policy.yaml
alebedev87: > // ValidatingAdmissionPolicy cannot match ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding.

This part means that it's not possible to create ValidatingAdmissionPolicy which validates the admission of another ValidatingAdmissionPolicy or ValidatingAdmissionPolicyBinding.

test/e2e/gateway_api_test.go
alebedev87: `GRPCRoute` will be added as part of https://github.com/openshift/cluster-ingress-operator/pull/1152. It's not yet mentioned anywhere. However, I can do a helper function in `pkg/manifests/assets/gatewayapi/` package which would return all "desired" CRDs. This way we can use a single function in all controllers and tests.

manifests/01-validating-admission-policy-binding.yaml
JoelSpeed: As Andrey points out, this is on our todo, but for now there's a slightly awkward dance to promote the manifest to GA separately from the gate being promoted

test/e2e/gateway_api_test.go
alebedev87: > I'd like to see a more precise error message comparison

This compassion ensures that the error message contains the substring we are responsible for: VAP's message. The rest of the message is out of our control and may change depending on how the admission chain wraps the admission request failures.

From [Hongan's test](https://github.com/openshift/cluster-ingress-operator/pull/1192#issuecomment-2699788573) we can see that the error is wrapped (made of multiple other error messages):
```
Error from server (Forbidden): error when creating "https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml": customresourcedefinitions.apiextensions.k8s.io "grpcroutes.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator
```

> It would be better for our docs and troubleshooting to add a human-readable messageExpression to each of the validations.

I'm not sure I understand this part. The current message (`"modifications to Gateway API Custom Resource Definitions may only be made by the Ingress Operator"`) seems to me quite human readable.

test/e2e/gateway_api_test.go
alebedev87: > However, I can do a helper function in pkg/manifests/assets/gatewayapi/ package which would return all "desired" CRDs. This way we can use a single function in all controllers and tests.

I gave this thing another thought. I think that it's better to abstain from using the actual "running" code in the tests. We should explicitly set the list of CRDs we are testing. Otherwise we risk to duplicate (and hide) errors we have in the code.

manifests/01-validating-admission-policy.yaml
alebedev87: I took this message pattern from other cluster oeprators: [ccm](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/e938398a53735d2f448fcf924b54f34b5b93af06/pkg/cloud/azure/assets/validating-admission-policy.yaml#L16) and [mco](https://github.com/openshift/machine-config-operator/blob/5dcc44abebbfb5dbeaaa8ef2569ba211032e1c00/manifests/machineconfigdaemon/mcn-guards-validatingadmissionpolicy.yaml#L20). I don't know to which degree we can diverge in this aspect across different cluster operators but I decided to stick to the same pattern for the sake of consistency.

test/e2e/util_gatewayapi_test.go
alebedev87: The check for the actual number of available replicas is added.

test/e2e/gateway_api_test.go
```diff
@@ -158,6 +161,60 @@ func testGatewayAPIObjects(t *testing.T) {
...
+	// Verify that GatewayAPI CRD update is forbidden.
+	for i := range testCRDs {
+		crdName := types.NamespacedName{Name: testCRDs[i].Name}
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		if err := kclient.Get(context.Background(), crdName, crd); err != nil {
+			t.Errorf("failed to get %q CRD: %v", crdName.Name, err)
+			continue
+		}
+		crd.Spec = testCRDs[i].Spec
+		crd.Spec.Conversion = testCRDs[i].Spec.Conversion
```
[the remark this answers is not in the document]
alebedev87: Right, removed it.

manifests/01-validating-admission-policy.yaml
alebedev87: Changed to `"Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be modified"`

test/e2e/util_gatewayapi_test.go
alebedev87: Done.

test/e2e/gateway_api_test.go
alebedev87: https://github.com/openshift/cluster-ingress-operator/pull/1192#issuecomment-2705207947

test/e2e/all_test.go
```diff
@@ -126,5 +125,6 @@ func TestAll(t *testing.T) {
 		t.Run("TestRouteHardStopAfterEnableOnIngressControllerHasPriorityOverIngressConfig", TestRouteHardStopAfterEnableOnIngressControllerHasPriorityOverIngressConfig)
 		t.Run("TestHostNetworkPortBinding", TestHostNetworkPortBinding)
 		t.Run("TestDashboardCreation", TestDashboardCreation)
+		t.Run("TestGatewayAPI", TestGatewayAPI)
```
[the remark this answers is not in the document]
alebedev87: https://github.com/openshift/cluster-ingress-operator/pull/1192#issuecomment-2705207947

test/e2e/util_gatewayapi_test.go
alebedev87: > Is it in our best interest to implement this function, knowing it has to change in the future every time a new CRD is added or the version changed?

Well, yes, we will have to come to this function for every new type just like `crdNames` we discussed in [this comment](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1984117536). And I think it's the good thing for the test because the test should not rely on the packages used in the code to avoid duplication of erroneous logic (same reason as in [my response](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1984753210) for the before mentioned comment).

Miciah:
> > > Hm, I guess the nuance is that the webhook is technically initialized, but the webhook rejects pods because Istiod isn't ready?
> > 
> > 
> > Right, plus you mentioned that other tests (which fail to create pods) time out. What I saw was that the tests just `Fatal` exited when their pods could not be created.
> 
> If the behavior is triggered by SMCP creation, then we have to wait until we bump to OSSM 3, where there is no SMCP. If the behavior is triggered by Istio/istiod inserting labels into every pod everywhere, then we need to fix that in our configuration or we need to report this a bug to OSSM.

I'd rather make the test run serially than delay this PR.  I expect that OSSM 3 has the same issue since the design of sidecar injection hasn't changed as far as I know.  I wouldn't object to pushing on this initialization-time side-effect of installing OSSM, but it shouldn't block our work.

manifests/01-validating-admission-policy.yaml
danwinship: Right, I understand why we want to validate that the request was made with the ingress operator's service account. I was asking why it _also_ requires that the request was made from a pod. It doesn't seem like that adds any actual security, and it doesn't seem like it would prevent any accidental mistakes either.

alebedev87:
Something with the image registry tests:
```
[sig-imageregistry][Feature:ImageInfo] Image info should display information about images
```

/retest

manifests/01-validating-admission-policy.yaml
candita: I'm not sure if this is the error message that @danwinship said wasn't helpful, but to me it isn't helpful as is.  If you're reviewing the logs later, you might want to know the user name of the non-admin person trying to change CRDs.

manifests/01-validating-admission-policy.yaml
candita: It is debatable but probably not worth it.

test/e2e/all_test.go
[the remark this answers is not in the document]
candita: SMCP is going away immediately after bumping to 3.0.  If we have this same issue in 3.0, what is the game plan?

test/e2e/gateway_api_test.go
candita: I'm going to hold the rest of my reviews until after #1152 is merged.  Based on the number of changes dependent on it, I feel like I don't want to have to review this again.

test/e2e/gateway_api_test.go
candita: There are two validations, the other has the message: "this user must have a \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims".  I'd like to see a test that produces that error.

I wasn't sure if there were other "generated" error messages for other parts of the policy.  I had read that if there was no message, a message would be generated from the expression.  If that's not the case here, and the only error messages  are the ones tied to validations, we just need one more for the other validation.

test/e2e/util_gatewayapi_test.go
candita: The only error returned from scaleDeployment is from wait.PollUntilContextTimeout timing out and that's not a wrapped error.

test/e2e/util_gatewayapi_test.go
```diff
@@ -741,3 +855,83 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
...
+
+// disable scales down CVO and removes the VAP resource.
+func (m *vapManager) disable() (error, func()) {
+	if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 0); err != nil {
+		return fmt.Errorf("failed to scale down cvo: %w", err), func() { /*scale down didn't work, nothing to do*/ }
+	}
+	if err := deleteExistingVAP(m.t, m.name); err != nil {
+		return fmt.Errorf("failed to delete vap %q: %w", m.name, err), func() {
+			if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 1); err != nil {
+				m.t.Errorf("failed to scale up cvo: %v", err)
```
candita, line 923: Here it is %v.

test/e2e/util_gatewayapi_test.go
```diff
@@ -741,3 +855,83 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
...
+	return nil, nil
+}
+
+// Enable scales up CVO and waits until the VAP is recreated.
+func (m *vapManager) enable() {
+	if err := scaleDeployment(m.t, cvoNamespace, cvoDeploymentName, 1); err != nil {
+		m.t.Errorf("failed to scale up cvo: %v", err)
+	} else if err := assertVAP(m.t, m.name); err != nil {
+		m.t.Errorf("failed to find vap %q: %v", m.name, err)
+	}
```
candita, line 936: These are also %v.

test/e2e/util_gatewayapi_test.go
alebedev87: Right, because here we log the error, we don't wrap it like in [the other case](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1984125500). Here is a small example to demonstrate the difference: [link](https://go.dev/play/p/p0_vzqWkzmv).

test/e2e/all_test.go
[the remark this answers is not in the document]
alebedev87: > If we have this same issue in 3.0, what is the game plan?

For the test? Make the test serial as I did in this PR.
Overall for the feature? We should find (or implement if it doesn't exist) a knob to disable the creation of any webhook configuration related to the sidecar injection.

test/e2e/util_gatewayapi_test.go
alebedev87: > The only error returned from scaleDeployment is from wait.PollUntilContextTimeout timing out and that's not a wrapped error.

1. It may be a wrapped error, 2. We are wrapping this error into another one so that the final error is recognized as both (e.g. by `errors.Is()`) . Example: [link](https://go.dev/play/p/p0_vzqWkzmv).

manifests/01-validating-admission-policy.yaml
alebedev87: > I was asking why it also requires that the request was made from a pod. It doesn't seem like that adds any actual security, and it doesn't seem like it would prevent any accidental mistakes either.

The idea was to prevent impersonated request from bypassing VAP:
```
$ oc --as=system:serviceaccount:openshift-ingress-operator:ingress-operator apply -f pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml 
Error from server (Forbidden): error when creating "pkg/manifests/assets/gateway-api/gateway.networking.k8s.io_gatewayclasses.yaml": customresourcedefinitions.apiextensions.k8s.io "gatewayclasses.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: this user must have a "authentication.kubernetes.io/node-name" and "authentication.kubernetes.io/pod-name" claims
```

> If we do need this check, the message is unhelpful (both to users and to future developers trying to debug failures.)

I was inspired by similar messages from the [CCM](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/e938398a53735d2f448fcf924b54f34b5b93af06/pkg/cloud/azure/assets/validating-admission-policy.yaml#L16) and [MCO](https://github.com/openshift/machine-config-operator/blob/5dcc44abebbfb5dbeaaa8ef2569ba211032e1c00/manifests/machineconfigdaemon/mcn-guards-validatingadmissionpolicy.yaml#L20) operators. I think the idea was to remain opaque about the request's user since it can be impersonated - hence the wording "this user." Listing all expected claims might be a bit detailed for some users, but it ensures the exact requirements are clear.

manifests/01-validating-admission-policy-binding.yaml
alebedev87: > Adding a comment regarding the workaround would be nice.

Can you explain which workaround you mean? For the moment, CVO does not have a way to set a specific feature gate for a manifest.

test/e2e/gateway_api_test.go
alebedev87: > There are two validations, the other has the message: "this user must have a "authentication.kubernetes.io/node-name" and "authentication.kubernetes.io/pod-name" claims". I'd like to see a test that produces that error.

Right, this part was missing. Let me add a test case.

manifests/01-validating-admission-policy.yaml
alebedev87: > you might want to know the user name of the non-admin person trying to change CRDs.

Well, the user name cannot be anything but `system:serviceaccount:openshift-ingress-operator:ingress-operator` because the first validation ensures this. There can be [an impersonation involved](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1987073723) but `request.userInfo.username` would not contain the name of the actual user anyway.

test/e2e/util_gatewayapi_test.go
alebedev87: Same answer as [here](https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1985894164).

manifests/01-validating-admission-policy.yaml
Miciah: We'll go with the message as is.  We can follow up to find out whether it needs to follow a specific pattern or whether it can be made more user-friendly.

candita: Okay.  I suggest changing the grammar a little too.
```suggestion
      message: "this user must have both \"authentication.kubernetes.io/node-name\" and \"authentication.kubernetes.io/pod-name\" claims"
```

test/e2e/all_test.go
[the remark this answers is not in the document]
candita: I still believe we should investigate and resolve this sooner rather than later, but I was outvoted.  Please don't forget to follow up.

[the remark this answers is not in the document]
Miciah: https://issues.redhat.com/browse/NE-1982 tracks the follow-up.  @alebedev87, I assigned myself to follow up once the OSSM 3 bump merges.  If you want to take it, feel free to do so.

alebedev87:
`e2e-aws-operator-techpreview` failed to create a pod for HTTPRoute due to `unable to find annotation openshift.io/sa.scc.uid-range` error:
```
=== RUN   TestAll/serial/TestGatewayAPI/testGatewayAPIObjects
    gateway_api_test.go:88: Creating namespace "test-e2e-gwapi-ws7d6"...
    gateway_api_test.go:88: Waiting for ServiceAccount test-e2e-gwapi-ws7d6/default to be provisioned...
    gateway_api_test.go:88: Waiting for RoleBinding test-e2e-gwapi-ws7d6/system:image-pullers to be created...
    gateway_api_test.go:88: failed to create one or more gateway object/s: feature gate was enabled, but http route object could not be created: failed to create pod test-e2e-gwapi-ws7d6/test-gateway-openshift-default: pods "test-gateway-openshift-default" is forbidden: error fetching namespace "test-e2e-gwapi-ws7d6": unable to find annotation openshift.io/sa.scc.uid-range
```

After a search in Slack, it seems like it's a recurring issue related to some race condition in one of the controllers responsible for the UID annotations ([example](https://redhat-internal.slack.com/archives/CB48XQ4KZ/p1655306634970219)). Most of the people just (suspense...) rerun their jobs :open_mouth:.

manifests/01-validating-admission-policy.yaml
alebedev87: Done.

alebedev87:
Addressed https://github.com/openshift/cluster-ingress-operator/pull/1192#discussion_r1988087977.

Miciah:
> After a search in Slack, it seems like it's a recurring issue related to some race condition in one of the controllers responsible for the UID annotations ([example](https://redhat-internal.slack.com/archives/CB48XQ4KZ/p1655306634970219)). Most of the people just (suspense...) rerun their jobs 😮.

Did _you_ file a bug report in Jira?  *grin*.

alebedev87:
Some problems with the CI image registry:
```
 * 2025-03-11T12:32:44Z 242x kubelet: Back-off pulling image "registry.ci.openshift.org/ci/entrypoint-wrapper:latest"
* 2025-03-11T11:38:44Z 4x kubelet: Error: ImagePullBackOff 
```

/retest

alebedev87:
> Did you file a bug report in Jira? grin.

No. Will do if I will see it more than once. This time was the only one so far.

alebedev87:
```
=== NAME  TestAll/parallel/TestManagedDNSToUnmanagedDNSIngressController
    util_test.go:714: verified connectivity with workload with req http://a6bc4691b16974830b41cb0cc8598e2f-661557309.us-west-2.elb.amazonaws.com and response 200
    unmanaged_dns_test.go:147: Updating ingresscontroller managed-migrated to dnsManagementPolicy=Unmanaged
    unmanaged_dns_test.go:160: Waiting for stable conditions on ingresscontroller managed-migrated after dnsManagementPolicy=Unmanaged
    unmanaged_dns_test.go:176: verifying conditions on DNSRecord zone {ID: Tags:map[Name:ci-op-4fxxjl7i-9e7c5-s6hzp-int kubernetes.io/cluster/ci-op-4fxxjl7i-9e7c5-s6hzp:owned]}
    unmanaged_dns_test.go:176: DNSRecord zone expected to have status=Unknown but got status=True
```
Seems like a reoccurred of https://issues.redhat.com/browse/OCPBUGS-17671.

candita:
Two failures in the e2e-aws-operator-techpreview.  First one you already noted:
RUN   TestAll/serial/TestGatewayAPI/testGatewayAPIObjects
..    gateway_api_test.go:88: failed to create one or more gateway object/s: feature gate was enabled, but http route object could not be created: failed to create pod test-e2e-gwapi-t8nzn/test-gateway-openshift-default: pods "test-gateway-openshift-default" is forbidden: error fetching namespace "test-e2e-gwapi-t8nzn": unable to find annotation openshift.io/sa.scc.uid-range
  
  Was this one there before?
  === RUN   TestAll/serial/TestGatewayAPI/testGatewayAPIIstioInstallation
    util_gatewayapi_test.go:400: found subscription servicemeshoperator at installed version 
    util_gatewayapi_test.go:748: found catalogSource redhat-operators with last observed state READY
    util_gatewayapi_test.go:461: failed to get deployment openshift-operators/istio-operator, retrying...
    [repeated, trying]
    util_gatewayapi_test.go:461: failed to get deployment openshift-operators/istio-operator, retrying...
    gateway_api_test.go:89: failed to find expected Istio operator: OSSM operator failure: pod istio-operator-75bb585bb-bktzx is not running, it is Pending
    
 It's got to be some other new techpreview feature if e2e-aws-gatewayapi is fine, don't you think?

PR #1193, opened 2025-02-25
NE-1952: Pre-upgrade check for Gateway API CRD Management Succession

grzpiotrowski:
This PR adds a new check in the Cluster Ingress Operator which sets its `Upgradeable` condition to `False` until the state of the Gateway API CRDs is validated.

The check logic verifies whether any Gateway API CRDs exist. If no GWAPI CRDs are installed in the cluster, then the operator's condition `Upgradable=True` (granted that nothing else prevents the upgrade).

If Gateway API CRDs are already present in the cluster the Cluster Ingress Operator `Upgradeable` condition is `False` until:
- only the standard CRDs exist: `GatewayClass`, `Gateway`, `HTTPRoute`, `GRPCRoute` or `ReferenceGrant`, and no other APIs from the `gateway.networking.k8s.io` are present.
- the existing standard CRDs are in `v1.2.1` which is the one currently targeted for OCP 4.19.

The check logic is added only to 4.18.z clusters.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -47,6 +47,12 @@ const (
 	HTTPRouteCRDAsset      = "assets/gateway-api/gateway.networking.k8s.io_httproutes.yaml"
 	ReferenceGrantCRDAsset = "assets/gateway-api/gateway.networking.k8s.io_referencegrants.yaml"
 
+	GatewayClassCRDAsset_v1_2_1   = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_gatewayclasses.yaml"
+	GatewayCRDAsset_v1_2_1        = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_gateways.yaml"
+	GRPCRouteCRDAsset_v1_2_1      = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_grpcroutes.yaml"
+	HTTPRouteCRDAsset_v1_2_1      = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_httproutes.yaml"
+	ReferenceGrantCRDAsset_v1_2_1 = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_referencegrants.yaml"
+
 	// Annotation used to inform the certificate generation service to
 	// generate a cluster-signed certificate and populate the secret.
 	ServingCertSecretAnnotation = "service.alpha.openshift.io/serving-cert-secret-name"
@@ -281,6 +287,46 @@ func HTTPRouteCRD() *apiextensionsv1.CustomResourceDefinition {
 
 func ReferenceGrantCRD() *apiextensionsv1.CustomResourceDefinition {
 	crd, err := NewCustomResourceDefinition(MustAssetReader(ReferenceGrantCRDAsset))
+	if err != nil {
+		panic(err)
+	}
+	return crd
+}
+
+func GatewayClassCRD_v1_2_1() *apiextensionsv1.CustomResourceDefinition {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(GatewayClassCRDAsset_v1_2_1))
+	if err != nil {
+		panic(err)
+	}
+	return crd
+}
+
+func GatewayCRD_v1_2_1() *apiextensionsv1.CustomResourceDefinition {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(GatewayCRDAsset_v1_2_1))
+	if err != nil {
+		panic(err)
+	}
+	return crd
+}
+
+func GRPCRouteCRD_v1_2_1() *apiextensionsv1.CustomResourceDefinition {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(GRPCRouteCRDAsset_v1_2_1))
+	if err != nil {
+		panic(err)
+	}
+	return crd
+}
+
+func HTTPRouteCRD_v1_2_1() *apiextensionsv1.CustomResourceDefinition {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(HTTPRouteCRDAsset_v1_2_1))
+	if err != nil {
+		panic(err)
+	}
+	return crd
+}
+
+func ReferenceGrantCRD_v1_2_1() *apiextensionsv1.CustomResourceDefinition {
+	crd, err := NewCustomResourceDefinition(MustAssetReader(ReferenceGrantCRDAsset_v1_2_1))
 	if err != nil {
 		panic(err)
 	}
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -18,9 +18,11 @@ import (
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
 	oputil "github.com/openshift/cluster-ingress-operator/pkg/util"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
 	corev1 "k8s.io/api/core/v1"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -52,13 +54,21 @@ var log = logf.Logger.WithName(controllerName)
 // clock is to enable unit testing
 var clock utilclock.Clock = utilclock.RealClock{}
 
+var expectedGatewayAPICRDs = map[string]*apiextensionsv1.CustomResourceDefinition{
+	manifests.GatewayClassCRD_v1_2_1().Name:   manifests.GatewayClassCRD_v1_2_1(),
+	manifests.GatewayCRD_v1_2_1().Name:        manifests.GatewayCRD_v1_2_1(),
+	manifests.GRPCRouteCRD_v1_2_1().Name:      manifests.GRPCRouteCRD_v1_2_1(),
+	manifests.HTTPRouteCRD_v1_2_1().Name:      manifests.HTTPRouteCRD_v1_2_1(),
+	manifests.ReferenceGrantCRD_v1_2_1().Name: manifests.ReferenceGrantCRD_v1_2_1(),
+}
+
 // New creates the status controller. This is the controller that handles all
 // the logic for creating the ClusterOperator operator and updating its status.
 //
 // The controller watches IngressController resources in the manager namespace
-// and uses them to compute the operator status.  It also watches the
-// clusteroperators resource so that it reconciles the ingress clusteroperator
-// in case something else updates or deletes it.
+// and Gateway API CRDs and uses them to compute the operator status.
+// It also watches the clusteroperators resource so that it reconciles
+// the ingress clusteroperator in case something else updates or deletes it.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	operatorCache := mgr.GetCache()
 	reconciler := &reconciler{
@@ -92,6 +102,17 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.ClusterOperator{}, handler.EnqueueRequestsFromMapFunc(toDefaultIngressController), predicate.NewPredicateFuncs(isIngressClusterOperator))); err != nil {
 		return nil, err
 	}
+
+	if !config.GatewayAPIEnabled {
+		crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+			group := o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group
+			return group == "gateway.networking.k8s.io" || group == "gateway.networking.x-k8s.io"
+		})
+		if err := c.Watch(source.Kind[client.Object](operatorCache, &apiextensionsv1.CustomResourceDefinition{}, handler.EnqueueRequestsFromMapFunc(toDefaultIngressController), crdPredicate)); err != nil {
+			return nil, err
+		}
+	}
+
 	return c, nil
 }
 
@@ -101,6 +122,7 @@ type Config struct {
 	CanaryImage            string
 	OperatorReleaseVersion string
 	Namespace              string
+	GatewayAPIEnabled      bool
 }
 
 // reconciler handles the actual status reconciliation logic in response to
@@ -186,7 +208,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 			r.config.CanaryImage,
 		),
 		computeOperatorDegradedCondition(state.IngressControllers),
-		computeOperatorUpgradeableCondition(state.IngressControllers),
+		computeOperatorUpgradeableCondition(state.IngressControllers, state.GatewayAPICRDs),
 		computeOperatorEvaluationConditionsDetectedCondition(state.IngressControllers),
 	)
 
@@ -236,6 +258,7 @@ type operatorState struct {
 	CanaryNamespace    *corev1.Namespace
 	IngressControllers []operatorv1.IngressController
 	DNSRecords         []iov1.DNSRecord
+	GatewayAPICRDs     []apiextensionsv1.CustomResourceDefinition
 }
 
 // getOperatorState gets and returns the resources necessary to compute the
@@ -266,6 +289,20 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string) 
 		return state, fmt.Errorf("failed to list ingresscontrollers in %q: %v", r.config.Namespace, err)
 	} else {
 		state.IngressControllers = ingressList.Items
+	}
+
+	if !r.config.GatewayAPIEnabled {
+		// Get all CRDs from the gateway.networking.k8s.io and gateway.networking.x-k8s.io groups.
+		clusterCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+		if err := r.cache.List(context.TODO(), clusterCRDs); err != nil {
+			return state, fmt.Errorf("failed to list CRDs: %w", err)
+		}
+
+		for _, crd := range clusterCRDs.Items {
+			if crd.Spec.Group == "gateway.networking.k8s.io" || crd.Spec.Group == "gateway.networking.x-k8s.io" {
+				state.GatewayAPICRDs = append(state.GatewayAPICRDs, crd)
+			}
+		}
 	}
 
 	return state, nil
@@ -392,24 +429,66 @@ func checkIngressControllersUpgradeableCondition(ingresses []operatorv1.IngressC
 	return false, message
 }
 
+// checkGatewayAPICRDs checks if Gateway API CRDs exist and if they are compatible.
+func checkGatewayAPICRDs(gatewayAPICRDs []apiextensionsv1.CustomResourceDefinition) (bool, string) {
+	if len(gatewayAPICRDs) == 0 {
+		return true, "No Gateway API CRDs found, upgradeable by default."
+	}
+
+	incompatibleCRDs := []string{}
+
+	for _, crd := range gatewayAPICRDs {
+		expectedCRD, exists := expectedGatewayAPICRDs[crd.Name]
+		// Ignoring the .Spec.Conversion field for comparison
+		sanitizedCRD := crd.DeepCopy()
+		sanitizedCRD.Spec.Conversion = nil
+		if !exists || !equality.Semantic.DeepEqual(sanitizedCRD.Spec, expectedCRD.Spec) {
+			incompatibleCRDs = append(incompatibleCRDs, crd.Name)
+		}
+	}
+
+	if len(incompatibleCRDs) > 0 {
+		message := fmt.Sprintf("Incompatible Gateway API CRDs found: %s", strings.Join(incompatibleCRDs, ", "))
+		message += "\nSee the release notes for more information: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes"
+		return false, message
+	}
+
+	return true, ""
+}
+
 // computeOperatorUpgradeableCondition computes the operator's Upgradeable
 // status condition.
-func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressController) configv1.ClusterOperatorStatusCondition {
-	ingressControllersUpgradeable, message := checkIngressControllersUpgradeableCondition(ingresses)
+func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressController, gatewayAPICRDs []apiextensionsv1.CustomResourceDefinition) configv1.ClusterOperatorStatusCondition {
+	ingressControllersUpgradeable, ingressControllersUpgradeableMessage := checkIngressControllersUpgradeableCondition(ingresses)
+	gatewayAPICRDsCompatible, gatewayAPICRDsMessage := checkGatewayAPICRDs(gatewayAPICRDs)
 
-	if ingressControllersUpgradeable {
+	if ingressControllersUpgradeable && gatewayAPICRDsCompatible {
 		return configv1.ClusterOperatorStatusCondition{
 			Type:   configv1.OperatorUpgradeable,
 			Status: configv1.ConditionTrue,
-			Reason: "IngressControllersUpgradeable",
+			Reason: "Upgradeable",
 		}
+	}
+
+	messages := []string{}
+	reason := "ComponentsNotUpgradeable"
+
+	if !ingressControllersUpgradeable && !gatewayAPICRDsCompatible {
+		reason = "MultipleComponentsNotUpgradeable"
+		messages = append(messages, ingressControllersUpgradeableMessage, gatewayAPICRDsMessage)
+	} else if !ingressControllersUpgradeable {
+		reason = "IngressControllersNotUpgradeable"
+		messages = append(messages, ingressControllersUpgradeableMessage)
+	} else if !gatewayAPICRDsCompatible {
+		reason = "GatewayAPICRDsNotCompatible"
+		messages = append(messages, gatewayAPICRDsMessage)
 	}
 
 	return configv1.ClusterOperatorStatusCondition{
 		Type:    configv1.OperatorUpgradeable,
 		Status:  configv1.ConditionFalse,
-		Reason:  "IngressControllersNotUpgradeable",
-		Message: message,
+		Reason:  reason,
+		Message: strings.Join(messages, "\n"),
 	}
 }
 
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -361,9 +361,8 @@ func computeOperatorDegradedCondition(ingresses []operatorv1.IngressController) 
 	return degradedCondition
 }
 
-// computeOperatorUpgradeableCondition computes the operator's Upgradeable
-// status condition.
-func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressController) configv1.ClusterOperatorStatusCondition {
+// checkIngressControllersUpgradeableCondition checks if all ingress controllers are upgradeable.
+func checkIngressControllersUpgradeableCondition(ingresses []operatorv1.IngressController) (bool, string) {
 	nonUpgradeableIngresses := make(map[*operatorv1.IngressController]operatorv1.OperatorCondition)
 	for i, ingress := range ingresses {
 		for j, cond := range ingress.Status.Conditions {
@@ -373,12 +372,9 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
 		}
 	}
 	if len(nonUpgradeableIngresses) == 0 {
-		return configv1.ClusterOperatorStatusCondition{
-			Type:   configv1.OperatorUpgradeable,
-			Status: configv1.ConditionTrue,
-			Reason: "IngressControllersUpgradeable",
-		}
+		return true, ""
 	}
+
 	message := "Some ingresscontrollers are not upgradeable:"
 	// Sort keys so that the result is deterministic.
 	keys := make([]*operatorv1.IngressController, 0, len(nonUpgradeableIngresses))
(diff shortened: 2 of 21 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -392,11 +407,72 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
...
+// TODO: validate existing gatewayapi crds with crd-schema-checker
+func checkGatewayAPICRDs(gatewayAPICRDs []apiextensionsv1.CustomResourceDefinition) (bool, string) {
+	var managedGatewayAPICRDs = []*apiextensionsv1.CustomResourceDefinition{
+		manifests.GatewayClassCRD(),
+		manifests.GatewayCRD(),
+		manifests.HTTPRouteCRD(),
+		manifests.ReferenceGrantCRD(),
+	}
+
+	managedCRDNames := map[string]bool{}
```
alebedev87, line 425: Just a nit remark. The optimal (and I believe go-idiomatic too) way to do a set data structure is to use an empty `struct`, this way you don't waste bytes for storing `bool`s.
```suggestion
	managedCRDNames := map[string]struct{}{}
```

pkg/operator/controller/status/controller.go
```diff
@@ -392,11 +407,72 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
...
+	for _, crd := range managedGatewayAPICRDs {
+		managedCRDNames[crd.Name] = true
+	}
+
+	if len(gatewayAPICRDs) == 0 {
+		return true, "No Gateway API CRDs found, upgradeable by default."
+	}
```
alebedev87, line 432: This condition can be moved at the very beginning of the function. This will save some CPU cycles which you use to fill a set of managed CRD names.

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +271,22 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
...
+	var installedGatewayAPICRDs []apiextensionsv1.CustomResourceDefinition
+
+	clusterCrds := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.client.List(context.TODO(), clusterCrds); err != nil {
+		return state, fmt.Errorf("failed to list CRDs: %v", err)
+	}
+
+	for _, crd := range clusterCrds.Items {
+		if crd.Spec.Group == "gateway.networking.k8s.io" {
+			installedGatewayAPICRDs = append(installedGatewayAPICRDs, crd)
```
alebedev87, line 284: You can append to `state.gatewayAPICRDs` directly without an intermediate slice.

pkg/operator/controller/status/controller.go
```diff
@@ -392,11 +407,72 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
...
+// checkGatewayAPICRDs will check if Gateway API CRDs exist and if they are compatible
+// TODO: Consider getting the managedCRDs from the gatewayAPI controller? Currently duplicates the list from there.
+// TODO: validate existing gatewayapi crds with crd-schema-checker
+func checkGatewayAPICRDs(gatewayAPICRDs []apiextensionsv1.CustomResourceDefinition) (bool, string) {
+	var managedGatewayAPICRDs = []*apiextensionsv1.CustomResourceDefinition{
+		manifests.GatewayClassCRD(),
+		manifests.GatewayCRD(),
+		manifests.HTTPRouteCRD(),
+		manifests.ReferenceGrantCRD(),
+	}
```
alebedev87, line 423: I think it's better to move this variable into the global scope to avoid its re-initialization at each function call.

pkg/operator/controller/status/controller.go
```diff
@@ -392,11 +407,72 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
...
+		messages = append(messages, gatewayAPICRDsMessage)
+	}
+
 	return configv1.ClusterOperatorStatusCondition{
 		Type:    configv1.OperatorUpgradeable,
 		Status:  configv1.ConditionFalse,
-		Reason:  "IngressControllersNotUpgradeable",
-		Message: message,
+		Reason:  "ComponentsNotUpgradeable",
+		Message: strings.Join(messages, "\n"),
```
alebedev87, line 475: Will the end of line character be expanded or inserted verbatim into the status? Do you have an example of how the status message will look like?

pkg/operator/controller/status/controller_test.go
```diff
@@ -533,32 +535,76 @@ func Test_computeOperatorStatusVersions(t *testing.T) {
 	}
 }
 
+// TODO: Consider separating the test into two: one for ingresscontrollers, another for gatewayapi crds (?)
```
alebedev87, line 538: It looks fine, `computeOperatorUpgradeableCondition` now takes the CRDs into consideration too.

pkg/operator/controller/status/controller_test.go
```diff
@@ -533,32 +535,76 @@ func Test_computeOperatorStatusVersions(t *testing.T) {
...
 func Test_computeOperatorUpgradeableCondition(t *testing.T) {
 	testCases := []struct {
 		description                   string
 		ingresscontrollersUpgradeable []bool
+		gatewayAPICRDs                []apiextensionsv1.CustomResourceDefinition
 		expectUpgradeable             bool
 	}{
 		{
-			description:                   "no ingresscontrollers exist",
+			description:                   "no ingresscontrollers exist, no gateway api crds",
```
alebedev87, line 547: I think it's fine to keep the description the way it was as missing `gatewayAPI` field already suggests "no gateway apis".

alebedev87 (commented):
Good start, keep it up! Just some early nit comments.

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +277,18 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
 		state.IngressControllers = ingressList.Items
 	}
 
+	//Get all CRDs from the gateway.networking.k8s.io group
```
danwinship, line 280: (nit: space after `//`)

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +277,18 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
...
+	clusterCrds := &apiextensionsv1.CustomResourceDefinitionList{}
```
danwinship, line 281: nit: `clusterCRDs`. (Go never does titlecase for acronyms. Always all-caps or all-lowercase.)

pkg/operator/controller/status/controller.go
```diff
@@ -392,24 +413,61 @@ func checkIngressControllersUpgradeableCondition(ingresses []operatorv1.IngressC
...
+	ingressControllersUpgradeable, ingressControllersUpgradeableMessage := checkIngressControllersUpgradeableCondition(ingresses)
+	gatewayAPICRDsCompatible, gatewayAPICRDsMessage := checkGatewayAPICRDs(gatewayAPICRDs)
 
-	if ingressControllersUpgradeable {
+	if ingressControllersUpgradeable && gatewayAPICRDsCompatible {
 		return configv1.ClusterOperatorStatusCondition{
 			Type:   configv1.OperatorUpgradeable,
 			Status: configv1.ConditionTrue,
-			Reason: "IngressControllersUpgradeable",
+			Reason: "AllComponentsUpgradeable", // TODO: Find better description
```
danwinship, line 463: eh, `Reason` is really more for when things are wrong. You could just say `Reason: "Upgradeable"`

pkg/operator/controller/status/controller.go
```diff
@@ -392,24 +413,61 @@ func checkIngressControllersUpgradeableCondition(ingresses []operatorv1.IngressC
...
+	if !gatewayAPICRDsCompatible {
+		messages = append(messages, gatewayAPICRDsMessage)
+	}
+
 	return configv1.ClusterOperatorStatusCondition{
 		Type:    configv1.OperatorUpgradeable,
 		Status:  configv1.ConditionFalse,
-		Reason:  "IngressControllersNotUpgradeable",
-		Message: message,
+		Reason:  "ComponentsNotUpgradeable",
```
danwinship, line 478: OTOH it would be better to have distinct reasons here. `IngressControllersNotUpgradeable`, `GatewayAPICRDsNotCompatible`, `MultipleProblems`/`MultipleComponentsNotUpgradeable`.

pkg/operator/controller/status/controller.go
```diff
@@ -421,23 +421,32 @@ func checkGatewayAPICRDs(gatewayAPICRDs []apiextensionsv1.CustomResourceDefiniti
...
 
 	nonStandardCRDs := []string{}
+	incompatibleCRDs := []string{}
+
 	for _, crd := range gatewayAPICRDs {
-		if !managedCRDNames[crd.Name] {
+		expectedCRD, exists := managedCRDMap[crd.Name]
+		if !exists {
 			nonStandardCRDs = append(nonStandardCRDs, crd.Name)
+		} else if !cmp.Equal(crd, *expectedCRD, cmpopts.EquateEmpty()) {
```
danwinship, line 436: hm... I see this file is already using `cmp`, but [its docs](https://pkg.go.dev/github.com/google/go-cmp/cmp#pkg-overview) say "It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal." (Upstream k8s is in the process of getting rid of all non-test uses of `cmp`.)

The standard k8s way to do this would be `apiequality.Semantic.DeepEqual` from `"k8s.io/apimachinery/pkg/api/equality"`.

pkg/manifests/manifests.go
```diff
@@ -47,6 +47,11 @@ const (
 	HTTPRouteCRDAsset      = "assets/gateway-api/gateway.networking.k8s.io_httproutes.yaml"
 	ReferenceGrantCRDAsset = "assets/gateway-api/gateway.networking.k8s.io_referencegrants.yaml"
 
+	GatewayClassCRDAsset_v1_2_1   = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_gatewayclasses.yaml"
+	GatewayCRDAsset_v1_2_1        = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_gateways.yaml"
+	HTTPRouteCRDAsset_v1_2_1      = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_httproutes.yaml"
+	ReferenceGrantCRDAsset_v1_2_1 = "assets/gateway-api/v1_2_1/gateway.networking.k8s.io_referencegrants.yaml"
```
danwinship, line 53: What about `GRPCRoute`?

pkg/operator/controller/status/controller.go
```diff
@@ -54,10 +54,10 @@ var log = logf.Logger.WithName(controllerName)
...
 
 var managedGatewayAPICRDs = []*apiextensionsv1.CustomResourceDefinition{
-	manifests.GatewayClassCRD(),
-	manifests.GatewayCRD(),
-	manifests.HTTPRouteCRD(),
-	manifests.ReferenceGrantCRD(),
+	manifests.GatewayClassCRD_v1_2_1(),
+	manifests.GatewayCRD_v1_2_1(),
+	manifests.HTTPRouteCRD_v1_2_1(),
+	manifests.ReferenceGrantCRD_v1_2_1(),
```
danwinship, line 60: Right, so at this point, an upgrade would fail on a cluster where someone had installed the 1.2.1 CRDs "normally", because they'd have the `GRPCRoute` CRD, which `checkGatewayAPICRDs` would consider "non-standard".

Even if we're not using `GRPCRoute`, I think the plan from the enhancement assumes we install the complete set of standard Gateway API CRDs, right?

pkg/operator/controller/status/controller.go
Miciah: (Also, please punctuate with a period at the end of the sentence.)

pkg/operator/controller/status/controller.go
grzpiotrowski: Thanks for having a look.
Yes, at the time I was not sure about the case for the GRPCRoute so I left it out until clarified. I will add it to the check as part of the expected and complete set of standard CRDs.

pkg/operator/controller/status/controller_test.go
grzpiotrowski: Thanks. As suggested I reverted the descriptions of existing unit tests to their original state.

pkg/operator/controller/status/controller.go
grzpiotrowski: Done. Also renamed this to `expectedGatewayAPICRDs` as it is now an array of the Gateway API CRDs in version 1.2.1, which are the `expected` ones as opposed to the ones currently managed by the operator in 4.18. That was my reasoning behind it anyways.

pkg/operator/controller/status/controller.go
grzpiotrowski: Done.

pkg/operator/controller/status/controller.go
grzpiotrowski: Exit early, got it :heavy_check_mark:

pkg/manifests/manifests.go
grzpiotrowski: Right, added `GPRCRoute` as expected.

pkg/operator/controller/status/controller.go
grzpiotrowski: Fixed.

pkg/operator/controller/status/controller.go
grzpiotrowski: Fixed :wrench:

pkg/operator/controller/status/controller.go
grzpiotrowski: I need to check that and I'll post the output. This was inspired by what's already done in `computeOperatorProgressingCondition`.

pkg/operator/controller/status/controller.go
grzpiotrowski: This now evolved into a map of pointers to the crds.

grzpiotrowski:
/rename NE-1952: Pre-upgrade check for Gateway API CRD Management Succession

grzpiotrowski:
/retitle NE-1952: Pre-upgrade check for Gateway API CRD Management Succession

alebedev87 (commented):
Looks good, let's polish it to the mergeable state. Some nit remarks plus two more conceptual ones:
- I tend to agree with [Dan's comment](https://github.com/openshift/cluster-ingress-operator/pull/1193/files#r1983589358). I think more specific reason would be beneficial for the user.
- I think that we can do an e2e test for this code change.

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +279,18 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
 		state.IngressControllers = ingressList.Items
 	}
 
+	// Get all CRDs from the gateway.networking.k8s.io group.
+	clusterCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.client.List(context.TODO(), clusterCRDs); err != nil {
+		return state, fmt.Errorf("failed to list CRDs: %v", err)
```
Miciah, line 285:
```suggestion
		return state, fmt.Errorf("failed to list CRDs: %w", err)
```

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +279,18 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
 		state.IngressControllers = ingressList.Items
 	}
 
+	// Get all CRDs from the gateway.networking.k8s.io group.
+	clusterCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.client.List(context.TODO(), clusterCRDs); err != nil {
```
Miciah, line 284: If you can use the cache, that reduces load on the API:
```suggestion
	if err := r.cache.List(context.TODO(), clusterCRDs); err != nil {
```

lihongan:
Did pre-merge test with the PR, it works as expected if no GWAPI CRDs or non-standard CRDs exist, but co/ingress still reports `Upgradeable=False` even the standard v1.2.1 CRDs are installed.
```
## no Gateway API CRDs
$ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
lastTransitionTime: "2025-03-13T02:45:30Z"
reason: Upgradeable
status: "True"
type: Upgradeable


## with experimental v1.1.1 CRDs
$ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
lastTransitionTime: "2025-03-13T05:18:12Z"
message: 'Incompatible Gateway API CRDs found: udproutes.gateway.networking.k8s.io'
reason: GatewayAPICRDsNotCompatible
status: "False"
type: Upgradeable

## with standard v1.2.1 CRDs
$ oc create -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml

$ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
lastTransitionTime: "2025-03-13T05:29:52Z"
message: 'Incompatible Gateway API CRDs found: gatewayclasses.gateway.networking.k8s.io, referencegrants.gateway.networking.k8s.io, httproutes.gateway.networking.k8s.io, grpcroutes.gateway.networking.k8s.io, gateways.gateway.networking.k8s.io'
reason: GatewayAPICRDsNotCompatible
status: "False"
type: Upgradeable
```

pkg/operator/controller/status/controller.go
grzpiotrowski: Example message including the new line:
```bash
message: |-
  Non-standard Gateway API CRDs found: gateways.gateway.networking.k8s.io, tcproutes.gateway.networking.k8s.io
  Incompatible Gateway API CRDs found: referencegrants.gateway.networking.k8s.io
reason: GatewayAPICRDsNotCompatible
status: "False"
type: Upgradeable
```
so it does look like it is expanded alright.

In the meantime I realized that the `gateway` or `httproute` and other CRDs can also be coming from the experimental channel, which I did not account for previously and these would be in the `Incompatible` category according to the old logic I had.

The choice was to either check the CRD annotations for the channel and categorise the CRD as `Non-standard` for the message or just call every CRD that does not pass our requirements as `Incompatible`.
Since we will be providing a reference to the docs in the message, there would be little benefit of this distinction in the message itself, hence I chose to call both experimental and old versions as `Incompatible`, listing them for the user.
This keeps the check logic simpler, without additional conditions just for a more detailed message.


New message for the scenario from the first example:
```bash
message: 'Incompatible Gateway API CRDs found: gateways.gateway.networking.k8s.io, tcproutes.gateway.networking.k8s.io, referencegrants.gateway.networking.k8s.io'
reason: GatewayAPICRDsNotCompatible
status: "False"
type: Upgradeable
```

pkg/operator/controller/status/controller_test.go
```diff
@@ -559,6 +563,39 @@ func Test_computeOperatorUpgradeableCondition(t *testing.T) {
...
+			description:                   "expected version of standard gatewayapi crds installed",
+			ingresscontrollersUpgradeable: []bool{true, true},
+			gatewayAPICRDs: []apiextensionsv1.CustomResourceDefinition{
+				*manifests.GatewayClassCRD_v1_2_1(),
+				*manifests.GatewayCRD_v1_2_1(),
+				*manifests.GRPCRouteCRD_v1_2_1(),
+				*manifests.HTTPRouteCRD_v1_2_1(),
+				*manifests.ReferenceGrantCRD_v1_2_1(),
+			},
+			expectUpgradeable: true,
```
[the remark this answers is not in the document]
grzpiotrowski: Done

pkg/operator/controller/status/controller.go
grzpiotrowski: Fixed

pkg/operator/controller/status/controller.go
grzpiotrowski: Fixed

pkg/operator/controller/status/controller.go
```diff
@@ -392,11 +411,72 @@ func computeOperatorUpgradeableCondition(ingresses []operatorv1.IngressControlle
...
+// checkGatewayAPICRDs will check if Gateway API CRDs exist and if they are compatible.
+func checkGatewayAPICRDs(gatewayAPICRDs []apiextensionsv1.CustomResourceDefinition) (bool, string) {
+	if len(gatewayAPICRDs) == 0 {
+		return true, "No Gateway API CRDs found, upgradeable by default."
+	}
+
+	expectedCRDMap := make(map[string]*apiextensionsv1.CustomResourceDefinition)
+	for _, crd := range expectedGatewayAPICRDs {
+		expectedCRDMap[crd.Name] = crd
+	}
```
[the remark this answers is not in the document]
grzpiotrowski: Done :+1:

pkg/operator/controller/status/controller.go
```diff
@@ -92,6 +103,14 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.ClusterOperator{}, handler.EnqueueRequestsFromMapFunc(toDefaultIngressController), predicate.NewPredicateFuncs(isIngressClusterOperator))); err != nil {
 		return nil, err
 	}
+
+	crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group == gatewayapiv1.GroupName
```
shaneutt, line 108: Related to an ongoing Slack discussion:

```suggestion
		group := o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group
		return group == gatewayapiv1.GroupName || group == "gateway.networking.x-k8s.io"
```

(getting the group from imports instead, however)

pkg/operator/controller/status/controller.go
```diff
@@ -268,6 +288,18 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
 		state.IngressControllers = ingressList.Items
 	}
 
+	// Get all CRDs from the gateway.networking.k8s.io group.
```
shaneutt, line 291: Related to the above:

```suggestion
	// Get all CRDs from the gateway.networking.k8s.io and gateway.networking.x-k8s.io group.
```

grzpiotrowski:
> Did pre-merge test with the PR, it works as expected if no GWAPI CRDs or non-standard CRDs exist, but co/ingress still reports `Upgradeable=False` even the standard v1.2.1 CRDs are installed.
> 
> ```
> ## no Gateway API CRDs
> $ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
> lastTransitionTime: "2025-03-13T02:45:30Z"
> reason: Upgradeable
> status: "True"
> type: Upgradeable
> 
> 
> ## with experimental v1.1.1 CRDs
> $ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
> lastTransitionTime: "2025-03-13T05:18:12Z"
> message: 'Incompatible Gateway API CRDs found: udproutes.gateway.networking.k8s.io'
> reason: GatewayAPICRDsNotCompatible
> status: "False"
> type: Upgradeable
> 
> ## with standard v1.2.1 CRDs
> $ oc create -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml
> 
> $ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
> lastTransitionTime: "2025-03-13T05:29:52Z"
> message: 'Incompatible Gateway API CRDs found: gatewayclasses.gateway.networking.k8s.io, referencegrants.gateway.networking.k8s.io, httproutes.gateway.networking.k8s.io, grpcroutes.gateway.networking.k8s.io, gateways.gateway.networking.k8s.io'
> reason: GatewayAPICRDsNotCompatible
> status: "False"
> type: Upgradeable
> ```

Thank you. This is now addressed by having a more targeted comparison, which now looks at .Spec of the CRDs.
That is in line with the approach used in the [gatewayapi controller](https://github.com/openshift/cluster-ingress-operator/blob/7dc12cf28557f8c80460435649cfc393761e0285/pkg/operator/controller/gatewayapi/crds.go#L123-L125)

pkg/operator/controller/status/controller.go
```diff
@@ -392,24 +429,66 @@ func checkIngressControllersUpgradeableCondition(ingresses []operatorv1.IngressC
...
+		sanitizedCRD := crd.DeepCopy()
+		sanitizedCRD.Spec.Conversion = nil
+		if !exists || !equality.Semantic.DeepEqual(sanitizedCRD.Spec, expectedCRD.Spec) {
+			incompatibleCRDs = append(incompatibleCRDs, crd.Name)
+		}
+	}
+
+	if len(incompatibleCRDs) > 0 {
+		message := fmt.Sprintf("Incompatible Gateway API CRDs found: %s", strings.Join(incompatibleCRDs, ", "))
+		message += "\nSee the release notes for more information: https://docs.openshift.com/container-platform/4.19/release_notes/ocp-4-19-release-notes.html#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes"
```
grzpiotrowski, line 452: Added the expected docs link to the message. Example of how it appears:
```yaml
lastTransitionTime: "2025-03-13T18:08:16Z"
message: |-
  Incompatible Gateway API CRDs found: listenersets.gateway.networking.x-k8s.io, tcproutes.gateway.networking.k8s.io
  See the release notes for more information: https://docs.openshift.com/container-platform/4.19/release_notes/ocp-4-19-release-notes.html#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes
reason: GatewayAPICRDsNotCompatible
status: "False"
type: Upgradeable
```

grzpiotrowski: Updated the URL to the new docs domain in [667f6cc](https://github.com/openshift/cluster-ingress-operator/pull/1193/commits/667f6cc19e4ada3e28a8def46af61dae32a15fcd)

grzpiotrowski: Spotted a small difference in the new URLs, so I'll need to correct that to:
`https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes`

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,93 @@
...
+			if !errors.IsAlreadyExists(err) {
+				t.Fatalf("Failed to create CRD %s: %v", crd.Name, err)
+			}
+			continue
+		}
+	}
+}
+
+// createCRD attempts to create the specified CRD and returns an error value.
+func createCRD(t *testing.T, desired *apiextensionsv1.CustomResourceDefinition) error {
```
grzpiotrowski, line 79: Just noticed that I left it here while I didn't use it in the end.

pkg/operator/controller/status/controller.go
grzpiotrowski: Added the check for `gateway.networking.x-k8s.io` group.
Since I was importing the package only to have that GroupName, I switched to hard coded values to avoid that import. For the `x` groupName I'd have to vendor `sigs.k8s.io/gateway-api/apis/v1alpha1`.

Not sure if that's a valid reason, so also happy to do it from imports if need be.

pkg/operator/controller/status/controller.go
grzpiotrowski: :heavy_check_mark:

grzpiotrowski:
```
{  openshift cluster install failed with infrastructure setup}
```
/test e2e-azure-ovn

lihongan:
@grzpiotrowski  retested again and looks good now, but the result below still need your confirmation: 
```
## just one of the 5 standard v1.2.1 CRDs exist, Upgradeable=True
$ oc get crds | grep -e gateway.networking.k8s.io
gatewayclasses.gateway.networking.k8s.io                          2025-03-17T04:42:29Z

$ oc get co/ingress -oyaml | yq '.status.conditions.[] | select(.type == "Upgradeable")'
lastTransitionTime: "2025-03-17T04:43:20Z"
reason: Upgradeable
status: "True"
type: Upgradeable
```
I'm not sure if we need to ensure all 5 CRDs exist in the cluster.

lihongan:
/label qe-approved

The test results are expected according to the discussion in slack

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	if gatewayAPIEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
+		t.Fatalf("error checking feature gate enabled status: %v", err)
+	} else if gatewayAPIEnabled {
+		t.Skip("Gateway API is enabled, skipping TestGatewayAPIUpgradeable")
+	}
+
+	t.Parallel()
+
+	defer deleteExistingCRDs(t, append(expectedCRDs, incompatibleCRDs...))
```
Miciah, line 42: For future reference, new test code should use [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup) if possible rather than `defer`.  Using `defer` will work, so don't worry about it for this PR.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func createCRDs(t *testing.T, crds []*apiextensionsv1.CustomResourceDefinition) {
+	t.Helper()
+	for _, crd := range crds {
+		if err := kclient.Create(context.TODO(), crd); err != nil {
+			if !errors.IsAlreadyExists(err) {
+				t.Fatalf("Failed to create CRD %s: %v", crd.Name, err)
+			}
+			continue
```
Miciah, line 72: This `continue` is superfluous, but it doesn't hurt anything.

Miciah:
This is a low-risk change.  The impact is limited to the operator's clusteroperator status reporting logic, it is well contained, and it has good test coverage.  (Thanks for the detailed comments and commit messages!)

/label backport-risk-assessed

Miciah:
/lgtm

/hold
I think @alebedev87 was going to give this PR a last pass.

alebedev87:
LGTM
/unhold

alebedev87:
Trying to bypass the jira automation which is made for bugs only.

/label jira/valid-bug

alebedev87:
/jira refresh

Will `4.18` target version work? `4.18.z` doesn't exist (yet).

PR #1196, opened 2025-02-26
NE-1951: Pre-upgrade Admin Gate for Gateway API CRD Management Succession

Thealisyed:
This PR introduces the following:

Initialize the GatewayAPI upgradeable controller
- Added the required imports and constants.
- Defined predicates for CRD and ConfigMap events to manage the admin gate.
- Configured watches in the controller for specific changes in the ConfigMap.

Implement the reconciler logic
- Added the Reconcile function to manage the admin gate addition and removal based on conditions

Enhance the operator initialization
- Conditionally initialize the gatewayapi_upgradeable controller based on the GatewayAPI feature gate.

 End-to-End Tests
- Added the TestGatewayAPIUpgradeable end-to-end test.
- Included the new test case in the test suite.

### DIFF
--- a/pkg/operator/controller/gatewayapi-upgradeable/controller.go
+++ b/pkg/operator/controller/gatewayapi-upgradeable/controller.go
@@ -1,0 +1,164 @@
+package gatewayapi_upgradeable
+
+import (
+	"context"
+	"fmt"
+
+	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	corev1 "k8s.io/api/core/v1"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	gatewayAPIAdminKey = "ack-4.18-gateway-api-management-in-4.19"
+	gatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
+)
+
+var (
+	log = logf.Logger.WithName(controllerName)
+)
+
+// The New function initializes the controller and sets up the watch for the ConfigMap.
+func New(mgr manager.Manager) (controller.Controller, error) {
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		Reconciler: &reconciler{
+			client: mgr.GetClient(),
+			cache:  mgr.GetCache(),
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Mapping the events to the admin gate ConfigMap.
+	toAdminGatesConfigMap := func(_ context.Context, _ client.Object) []reconcile.Request {
+		return []reconcile.Request{{
+			NamespacedName: operatorcontroller.AdminGatesConfigMapName(),
+		}}
+	}
+
+	// Defining the CRD predicate.
+	crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		group := o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group
+		return group == gatewayapiv1.GroupName || group == "gateway.networking.x-k8s.io"
+	})
+
+	// Setting up a watch for CRD events.
+	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &apiextensionsv1.CustomResourceDefinition{}, handler.EnqueueRequestsFromMapFunc(toAdminGatesConfigMap), crdPredicate)); err != nil {
+		return nil, err
+	}
+
+	// A predicate filter to watch for specific changes in the ConfigMap.
+	// Verify that the ConfigMap's name and namespace match the expected values.
+	adminGatePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetNamespace() == operatorcontroller.AdminGatesConfigMapName().Namespace &&
+			o.GetName() == operatorcontroller.AdminGatesConfigMapName().Name
+	})
+
+	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &corev1.ConfigMap{}, &handler.EnqueueRequestForObject{}, adminGatePredicate)); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+// reconciler struct holds the client and cache attributes.
+type reconciler struct {
+	client client.Client
+	cache  cache.Cache
+}
+
+// Reconcile function implements the logic to check conditions and manage the admin gate.
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
+
+	adminGateConditionExists, err := r.adminGateConditionExists(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to determine if admin gate condition exists: %w", err)
+	}
+
+	if adminGateConditionExists {
+		if err := r.addAdminGate(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to add admin gate: %w", err)
+		}
+	} else {
+		if err := r.removeAdminGate(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to remove admin gate: %w", err)
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// adminGateConditionExists checks if the admin gate condition exists based on both ConfigMap and CRDs.
+func (r *reconciler) adminGateConditionExists(ctx context.Context) (bool, error) {
+	crds := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.cache.List(ctx, crds); err != nil {
+		return false, fmt.Errorf("failed to list CRDs: %w", err)
+	}
+
+	for _, crd := range crds.Items {
+		if crd.Spec.Group == gatewayapiv1.GroupName || crd.Spec.Group == "gateway.networking.x-k8s.io" {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// The addAdminGate function is responsible for adding the admin gate to the ConfigMap.
+func (r *reconciler) addAdminGate(ctx context.Context) error {
+	adminGatesConfigMap := &corev1.ConfigMap{}
+	if err := r.cache.Get(ctx, operatorcontroller.AdminGatesConfigMapName(), adminGatesConfigMap); err != nil {
+		return fmt.Errorf("failed to get configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+
+	if adminGatesConfigMap.Data == nil {
+		adminGatesConfigMap.Data = map[string]string{}
+	}
+
+	// The function checks if the admin key exists and if it is set to the expected message.
+	if val, ok := adminGatesConfigMap.Data[gatewayAPIAdminKey]; ok && val == gatewayAPIAdminMsg {
+		return nil
+	}
+	adminGatesConfigMap.Data[gatewayAPIAdminKey] = gatewayAPIAdminMsg
+
+	log.Info("Adding admin gate for Gateway API management")
+	if err := r.client.Update(ctx, adminGatesConfigMap); err != nil {
+		return fmt.Errorf("failed to update configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+	return nil
+}
+
+// The removeAdminGate function is responsible for removing the admin gate from the ConfigMap.
+func (r *reconciler) removeAdminGate(ctx context.Context) error {
+	adminGatesConfigMap := &corev1.ConfigMap{}
+	if err := r.cache.Get(ctx, operatorcontroller.AdminGatesConfigMapName(), adminGatesConfigMap); err != nil {
+		return fmt.Errorf("failed to get configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+
+	if _, ok := adminGatesConfigMap.Data[gatewayAPIAdminKey]; !ok {
+		return nil
+	}
+
+	log.Info("Removing admin gate for Gateway API management")
+	delete(adminGatesConfigMap.Data, gatewayAPIAdminKey)
+	if err := r.client.Update(ctx, adminGatesConfigMap); err != nil {
+		return fmt.Errorf("failed to update configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+	return nil
+}
--- a/pkg/operator/controller/names.go
+++ b/pkg/operator/controller/names.go
@@ -49,6 +49,13 @@ const (
 	RemoteWorkerLabel = "node.openshift.io/remote-worker"
 )
 
+func AdminGatesConfigMapName() types.NamespacedName {
+	return types.NamespacedName{
+		Name:      "admin-gates",
+		Namespace: GlobalMachineSpecifiedConfigNamespace,
+	}
+}
+
 // IngressClusterOperatorName returns the namespaced name of the ClusterOperator
 // resource for the operator.
 func IngressClusterOperatorName() types.NamespacedName {
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -35,13 +35,13 @@ import (
 	dnscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/dns"
 	gatewayservicednscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gateway-service-dns"
 	gatewayapicontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gatewayapi"
+	gatewayapi_upgradeable "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gatewayapi-upgradeable"
 	gatewayclasscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gatewayclass"
 	ingress "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
 	ingresscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingress"
 	ingressclasscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/ingressclass"
 	statuscontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/status"
 	"github.com/openshift/library-go/pkg/operator/events"
-
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -318,6 +318,13 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		},
 	}); err != nil {
 		return nil, fmt.Errorf("failed to create gatewayapi controller: %w", err)
+	}
+
+	// Add conditional setup for gateway_upgradeable controller only if FeatureGate is not enabled
+	if !gatewayAPIEnabled {
+		if _, err := gatewayapi_upgradeable.New(mgr); err != nil {
+			return nil, fmt.Errorf("failed to create gatewayapi upgradeable controller: %w", err)
+		}
 	}
 
 	return &Operator{
--- a/test/e2e/gatewayapi_upgradeable_test.go
+++ b/test/e2e/gatewayapi_upgradeable_test.go
@@ -6,15 +6,18 @@ package e2e
 import (
 	"context"
 	"testing"
-
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
 	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
 	test_crds "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/test/crds"
 
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var expectedCRDs = []*apiextensionsv1.CustomResourceDefinition{
@@ -30,21 +33,35 @@ var incompatibleCRDs = []*apiextensionsv1.CustomResourceDefinition{
 	test_crds.TCPRouteCRD_experimental_v1(),
 }
 
+// TestGatewayAPIUpgradeable verifies the operator's upgradeable condition and admin gate behavior
+// when the Gateway API feature gate is disabled.
 func TestGatewayAPIUpgradeable(t *testing.T) {
+	t.Parallel()
 	if gatewayAPIEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
 		t.Fatalf("error checking feature gate enabled status: %v", err)
 	} else if gatewayAPIEnabled {
 		t.Skip("Gateway API is enabled, skipping TestGatewayAPIUpgradeable")
 	}
 
-	t.Parallel()
-
-	defer deleteExistingCRDs(t, append(expectedCRDs, incompatibleCRDs...))
+	t.Cleanup(func() {
+		for _, crd := range append(expectedCRDs, incompatibleCRDs...) {
+			if err := kclient.Delete(context.TODO(), crd); err != nil {
+				if errors.IsNotFound(err) {
+					continue
+				}
+				t.Errorf("failed to delete crd %q: %v", crd.Name, err)
+			}
+		}
+	})
 
 	createCRDs(t, expectedCRDs)
+	testAdminGate(t, true)
 	testOperatorUpgradeableCondition(t, true)
 	createCRDs(t, incompatibleCRDs)
 	testOperatorUpgradeableCondition(t, false)
+	deleteExistingCRDs(t, append(expectedCRDs, incompatibleCRDs...))
+	testAdminGate(t, false)
+	testOperatorUpgradeableCondition(t, true)
 }
 
 func testOperatorUpgradeableCondition(t *testing.T, expectUpgradeable bool) {
@@ -69,7 +86,6 @@ func createCRDs(t *testing.T, crds []*apiextensionsv1.CustomResourceDefinition) 
 			if !errors.IsAlreadyExists(err) {
 				t.Fatalf("Failed to create CRD %s: %v", crd.Name, err)
 			}
-			continue
 		}
 	}
 }
@@ -78,5 +94,24 @@ func deleteExistingCRDs(t *testing.T, crds []*apiextensionsv1.CustomResourceDefi
 	t.Helper()
 	for _, crd := range crds {
 		deleteExistingCRD(t, crd.Name)
+	}
+}
+
+func testAdminGate(t *testing.T, shouldExist bool) {
+	t.Helper()
+	adminGatesConfigMap := &corev1.ConfigMap{}
+	if err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, client.ObjectKey{Namespace: "openshift-config-managed", Name: "admin-gates"}, adminGatesConfigMap); err != nil {
+			t.Logf("Failed to get configmap admin-gates: %v, retrying...", err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Timed out trying to get admin-gates configmap: %v", err)
+	}
+
+	_, adminGateKeyExists := adminGatesConfigMap.Data["ack-4.18-gateway-api-management-in-4.19"]
+	if adminGateKeyExists != shouldExist {
+		t.Fatalf("Expected admin gate key existence to be %v, but got %v", shouldExist, adminGateKeyExists)
 	}
 }


pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -83,6 +83,9 @@ type Config struct {
 	// resources.  The gatewayapi controller starts these controllers once
 	// the Gateway API CRDs have been created.
 	DependentControllers []controller.Controller
+
+	//AdminConsentRequired indicates
+	AdminConsentRequired bool
```
Thealisyed, line 88: Not needed, will delete!

Thealisyed:
Used Grant's [PR](https://github.com/openshift/cluster-ingress-operator/pull/1014/commits/6778ce1a619b2f58310a037694e341e446888c55#) as reference to implement admin gate in this manner.

pkg/operator/controller/names.go
```diff
@@ -49,6 +49,13 @@ const (
 	RemoteWorkerLabel = "node.openshift.io/remote-worker"
 )
 
+func AdminGatesConfigMapName() types.NamespacedName {
+	return types.NamespacedName{
+		Name:      "admin-gates",
+		Namespace: "openshift-config",
```
Thealisyed, line 55: use GlobalMachineSpecifiedConfigNamespace instead. 
OP Config managed is what the operator created

pkg/operator/operator.go
```diff
@@ -3,6 +3,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	gatewayapi_upgradeable "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/gatewayapi-upgradeable"
```
Thealisyed, line 6: group these properly, should be in the subsequent box

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	GatewayAPIAdminKey = "ack-gateway-api-management"
+	GatewayAPIAdminMsg = "You are now acknowledging that you are taking control of the Gateway API CRD Management from here on out. Please review the documentation for implications."
```
Thealisyed, line 25: Reference the correct documentation and fix the message

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+	//TODO: too many args to c.watch and not enough to source.Kind
+	if err := c.Watch(source.Kind(operatorCache, &corev1.ConfigMap{}),
+		handler.EnqueueRequestsFromMapFunc(toAdminGateConfigMap),
+		adminGateFilter); err != nil {
+		return nil, err
+	}
+
+	// Watch for CRD updates using the reconciler cache and apply the CRD predicate
+	if err := c.Watch(source.Kind(reconciler.cache, &apiextensionsv1.CustomResourceDefinition{}),
+		&handler.EnqueueRequestForObject{},
```
Thealisyed, line 89: the kinds should be swapped for the watches you have.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+
+var (
+	log = logf.Logger.WithName(controllerName)
+)
+
+// New function initializes the controller and sets up the watch for the ConfigMap
+func New(mgr manager.Manager, config Config) (controller.Controller, error) {
+	operatorCache := mgr.GetCache()
+	reconciler := &reconciler{
+		cache:  config.Cache,
```
alebedev87, line 36: The cache you passed in the operator is the manager's cache. No need to spawn many cache variables or fields, if `mgr.GetCache()` is what you need then `mgr` parameter is enough.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+			configmapNew := e.ObjectNew.(*corev1.ConfigMap)
+			if configmapOld.Name != operatorcontroller.AdminGatesConfigMapName().Name || configmapOld.Namespace != operatorcontroller.AdminGatesConfigMapName().Namespace {
+				return false
+			}
+			oldVal, oldExists := configmapOld.Data[GatewayAPIAdminKey]
+			newVal, newExists := configmapNew.Data[GatewayAPIAdminKey]
+			return oldExists != newExists || oldVal != newVal
+		},
+		GenericFunc: func(genericEvent event.GenericEvent) bool { return false }, // Ignore generic events
+	}
```
alebedev87, line 68: I think something like this should be enough:
```suggestion
	adminGatePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
	         cm := o.(*corev1.ConfigMap)
		return cm.Name ==  operatorcontroller.AdminGatesConfigMapName().Name && cm.Namespace == operatorcontroller.AdminGatesConfigMapName().Namespace
	})
```

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+		}}
+	}
+	// Define the CRD predicate
+	crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group == gatewayapiv1.GroupName
+	})
+
+	//TODO: too many args to c.watch and not enough to source.Kind
+	if err := c.Watch(source.Kind(operatorCache, &corev1.ConfigMap{}),
+		handler.EnqueueRequestsFromMapFunc(toAdminGateConfigMap),
```
alebedev87, line 82: Why it's not just `&handler.EnqueueRequestForObject{}`?  I think your controller should reconcile the admin gate configmap.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
alebedev87: What does your controller reconcile? ConfigMaps or CRDs? IMO, this watch should do the mapping to the admingate configmap. This way once you have GWAPI CRDs the admingate configmap is re-enrsured.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+// reconciler struct holds the client and cache attributes
+type reconciler struct {
+	client client.Client
+	cache  cache.Cache
+}
+
+// Config struct holds the cache information
+type Config struct {
+	//namespace referes to namespace where admin gates controller resides
+	Namespace string
```
alebedev87, line 105: I don't think you need this field (and maybe the whole config?) because the admingate configmap is already defined as a constant in `names.go`.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if adminGateConditionExists {
+		if err := r.addAdminGate(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to add admin gate: %w", err)
+		}
+	} else {
+		if err := r.removeAdminGate(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to remove admin gate: %w", err)
+		}
+	}
```
alebedev87, line 125: I suppose that the decision about adding or removing a data field to/from the admin in gate should be based on the presence of GWAPI CRDs which I don't see here.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
alebedev87, line 1: The name of the controller doesn't seem to quite reflect what the controller is supposed to do. The controller will ensure the admingate for the gatewayAPI CRDs.

alebedev87 (commented):
Good start! Keep it up! Some comments to steer the direction of the controller's implementation.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
alebedev87: Sorry, I think I see where the inspiration is coming from: `route_upgradable` in `release-4.15` branch. Ok then, it may make sense to keep this kinda consistent.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return reconcile.Result{}, nil
+}
+
+// adminGateConditionExists checks if the admin gate condition exists based on both ConfigMap and CRDs
+func (r *reconciler) adminGateConditionExists(ctx context.Context) (bool, error) {
```
Thealisyed, line 106: Function should check if the CRD's exist or not only!

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,174 @@
...
+	adminGateConfigMap := &corev1.ConfigMap{}
+	if err := r.cache.Get(ctx, operatorcontroller.AdminGatesConfigMapName(), adminGateConfigMap); err != nil {
+		return false, fmt.Errorf("failed to get configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+
+	// Check if the admin key exists and is set to the expected message
+	if val, ok := adminGateConfigMap.Data[GatewayAPIAdminKey]; ok && val == GatewayAPIAdminMsg {
+		return true, nil // Exists as expected
+	}
```
Thealisyed, line 115: Copy and paste as AddAdminGate?...

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,164 @@
+package gatewayapi_upgradeable
+
+import (
+	"context"
+	"fmt"
```
candita, line 5: Add lines between package groups imports e.g. between standard packages, openshift packages, and k8s packages.

pkg/operator/controller/names.go
```diff
@@ -49,6 +49,13 @@ const (
 	RemoteWorkerLabel = "node.openshift.io/remote-worker"
 )
 
+func AdminGatesConfigMapName() types.NamespacedName {
+	return types.NamespacedName{
+		Name:      "admin-gates",
+		Namespace: GlobalMachineSpecifiedConfigNamespace,
```
candita, line 55: Where is GlobalMachineSpecifiedConfigNamespace declared?

pkg/operator/controller/names.go
```diff
@@ -49,6 +49,13 @@ const (
 	RemoteWorkerLabel = "node.openshift.io/remote-worker"
 )
 
+func AdminGatesConfigMapName() types.NamespacedName {
+	return types.NamespacedName{
+		Name:      "admin-gates",
```
candita, line 54: We probably want to call this something like:  `openshift-ingress-admingates`

pkg/operator/operator.go
```diff
@@ -319,6 +320,11 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		return nil, fmt.Errorf("failed to create gatewayapi controller: %w", err)
 	}
 
+	// Set up the gateway api upgradeable controller
+	if _, err := gatewayapi_upgradeable.New(mgr); err != nil {
+		return nil, fmt.Errorf("failed to create gatewayapi upgradeable controller: %w", err)
```
candita, line 325:
```suggestion
		return nil, fmt.Errorf("failed to create gatewayapi upgradeable controller: %v", err)
```

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	GatewayAPIAdminKey = "ack-gateway-api-management"
+	GatewayAPIAdminMsg = "You are now acknowledging that you are taking control of the Gateway API CRD Management from here on out. Please review the documentation for implications."
```
candita, line 28: The message should include three pieces of info: why the message was generated, what the issue is, and how seriously we advise against what they're trying to do.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+var (
+	log = logf.Logger.WithName(controllerName)
+)
+
+// New function initializes the controller and sets up the watch for the ConfigMap
```
candita, line 35: It's a nit, but all comments should be complete sentences and end in a period.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Reconcile function implements the logic to check conditions and manage the admin gate
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
+
+	// Check if the admin gate should be added or removed
+	adminGateConditionExists, err := r.adminGateConditionExists(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to determine if admin gate exist: %w", err)
```
candita, line 94:
```suggestion
		return reconcile.Result{}, fmt.Errorf("failed to determine if admin gate condition exists: %v", err)
```

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+func (r *reconciler) adminGateConditionExists(ctx context.Context) (bool, error) {
+	// Check for the presence of Gateway API CRDs
+	crds := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.client.List(ctx, crds); err != nil {
+		return false, fmt.Errorf("failed to list CRDs: %w", err)
+	}
+
+	for _, crd := range crds.Items {
+		if crd.Spec.Group == gatewayapiv1.GroupName {
+			return true, nil
```
candita, line 120: Should we check the versions? If it is the same version, that should be less of a problem than if it were an earlier or later version, right?  It would impact the kind of error message we write.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return false, nil // No Gateway API CRDs found
+}
+
+// addAdminGate function adds the admin gate to the ConfigMap
+func (r *reconciler) addAdminGate(ctx context.Context) error {
+	adminGateConfigMap := &corev1.ConfigMap{}
+	if err := r.client.Get(ctx, operatorcontroller.AdminGatesConfigMapName(), adminGateConfigMap); err != nil {
+		return fmt.Errorf("failed to get configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
```
candita, line 131: What if it doesn't exist and we need to create it?  Can that happen?

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// adminGateConditionExists checks if the admin gate condition exists based on both ConfigMap and CRDs
+func (r *reconciler) adminGateConditionExists(ctx context.Context) (bool, error) {
+	// Check for the presence of Gateway API CRDs
+	crds := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := r.client.List(ctx, crds); err != nil {
```
Miciah, line 114: Use `r.cache` if you can.  Using `r.Client` puts more load on the API.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: According to the [EP](https://github.com/openshift/enhancements/pull/1756/files?short_path=e820058#diff-e82005837f3b20b3e0c690dfbe6dacac0581f41812783e633b73b1ffb030f16b:~:text=IF%20the%20CRDs,of%20CRD%20succession), the admingate is created upon the presense of the GWAPI CRD's (regardless of the versions).

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: The CVO is responsible for recreating  and managing the [ConfigMap](https://github.com/openshift/enhancements/blob/master/enhancements/update/upgrades-blocking-on-ack.md#:~:text=Another%20risk%20is,has%20been%20removed.) 
So if I understand correctly the if the ConfigMap is bogus or missing etc, the CVO will reconcile and fix it accordingly
Also in favor of following the same logic to implement this as done in this [PR](https://github.com/openshift/cluster-ingress-operator/pull/1014/commits/6778ce1a619b2f58310a037694e341e446888c55).

pkg/operator/controller/gatewayapi-upgradeable/controller.go
alebedev87: Just a note of clarification for @Thealisyed. By default the controller-runtime's client is a split client meaning that it makes read-only requests (get/list) against the downstream cache and write requests directly against the api server. However the cluster ingress operator [customizes the client](https://github.com/openshift/cluster-ingress-operator/blob/e80fa46b72ec08e6d8a7fc7f111d25842cbb80bf/pkg/operator/operator.go#L154-L165) to make all requests against the api server.

pkg/operator/controller/names.go
alebedev87: I don't think the name of the admin gate configmap can be customizable. [The EP](https://github.com/openshift/enhancements/blob/master/enhancements/update/upgrades-blocking-on-ack.md) mentions only `admin-gates` configmap.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	t.Parallel()
+
+	t.Log("Starting test: Ensure the Gateway API CRDs exist")
+	ensureCRDs(t)
```
alebedev87, line 24: Your PR targets 4.18 branch which doesn't have `GatewayAPI` enabled. So, this part will fail on most of e2e tests (except for `e2e-operator-techpreview`).

alebedev87: As we discussed, you need to create a new function which creates a test CRD from gateway group (`.spec.group`).

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	t.Log("Starting test: Deleting Gateway API CRDs")
+	deleteCRDs(t)
+
+	t.Log("Starting test: Testing if admin gate works properly after CRDs are deleted")
+	testAdminGate(t)
+
+	// Defer the cleanup of the test gateway.
+	t.Cleanup(func() {
+		t.Log("Starting cleanup: Deleting test gateway")
```
alebedev87, line 37: Deleting test gateway? You didn't create any gateway. You should check the code you copy from other tests whether it suits your need.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+}
+
+func getClient() (*kubernetes.Clientset, error) {
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	return kubernetes.NewForConfig(kubeConfig)
+}
```
alebedev87, line 57: You don't need to create another instance of the client. There is [a global variable](https://github.com/openshift/cluster-ingress-operator/blob/4db86485a0c44e8592c993a8215b5d8ad6b44947/test/e2e/operator_test.go#L120) for this.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	t.Parallel()
+
+	t.Log("Starting test: Ensure the Gateway API CRDs exist")
```
alebedev87, line 23: In CIO repository, the tests tend to be less verbose. We try to limit logs to things which fail (e.g. retries) or important to the logic of the test (featuregates, parameter changes, etc.). You can keep those logs are comments though.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+		Get(context.TODO(), "admin-gates", metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("failed to get configmap admin-gates: %v", err)
+		if errors.IsNotFound(err) {
+			t.Logf("ConfigMap 'admin-gates' not found in namespace openshift-config-managed")
+		} else {
+			t.Logf("Error getting ConfigMap: %v", err)
+		}
+		return
+	}
```
alebedev87, line 79: Can you please wrap this into a polling loop? See other tests, how they make use of `wait.PollUntilContextTimeout` function.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func testAdminGate(t *testing.T) {
+	kclient, err := getClient()
+	if err != nil {
+		t.Fatalf("failed to get kube client: %v", err)
+	}
+
+	t.Log("Ensuring CRDs exist")
+	crdsExist := true
```
alebedev87, line 66: How can this impact the check of the admingate?

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+		return
+	}
+
+	t.Log("Checking for admin gate key in the configmap")
+	adminGateKeyExists := false
+	if _, ok := configMap.Data[gatewayapi_upgradeable.GatewayAPIAdminKey]; ok {
+		adminGateKeyExists = true
+	}
+
+	t.Logf("CRDs exist: %v, Admin Gate Key exists: %v", crdsExist, adminGateKeyExists)
```
alebedev87, line 87: This is ok to log while you debug your test but, it's too verbose for a job which will be run as a pre-submit for every PR.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if crdsExist && !adminGateKeyExists {
```
alebedev87, line 89: Again, I don't quite get how `crdExist` influences the decision about the admingate.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,94 @@
...
+		t.Errorf("expected admin gate key to exist in configmap, but it does not")
+	} else if !crdsExist && adminGateKeyExists {
+		t.Errorf("expected admin gate key to be removed from configmap, but it still exists")
+	}
+}
```
alebedev87, line 94: You use the same `testAdminGate` function to validate the existence of the expected admin gate for its absence (after the crd deletion). How is it supposed to work?

alebedev87 (changes requested):
I had a look at e2e test. Good start but it needs some modifications.

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Right this was not implemented properly in this commit however overall if gwapi crd are present operator will add admin gate

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group == gatewayapiv1.GroupName
+	})
+
+	// Setting up a watch for CRD events.
+	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &apiextensionsv1.CustomResourceDefinition{}, handler.EnqueueRequestsFromMapFunc(toAdminGateConfigMap), crdPredicate)); err != nil {
+		return nil, err
+	}
+
+	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &corev1.ConfigMap{}, &handler.EnqueueRequestForObject{}, adminGatePredicate)); err != nil {
```
shaneutt, line 73: nit: you could move the definition of the `adminGatePredicate` right above here since this is the only place its used.

```suggestion
	adminGatePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
		return o.GetNamespace() == operatorcontroller.AdminGatesConfigMapName().Namespace &&
			o.GetName() == operatorcontroller.AdminGatesConfigMapName().Name
	})

	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &corev1.ConfigMap{}, &handler.EnqueueRequestForObject{}, adminGatePredicate)); err != nil {
```

shaneutt (commented):
I took a very brief initial first pass and have some small suggestions. I intend to do another pass that's thorough in the morning.

Important note: whenever you see a "nit" comment from me, please consider it resolvable at your discretion. These are not items I feel need to hold anything up.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
...
+func kubeinit() {
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		panic(err)
+	}
+	kclient, err = client.New(kubeConfig, client.Options{})
+	if err != nil {
+		panic(err)
+	}
+}
```
alebedev87, line 32: Please use [the package global kclient variable](https://github.com/openshift/cluster-ingress-operator/blob/f1b159ad1f30f2a3034e13b264081f08e67695b4/test/e2e/operator_test.go#L120) instead.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	t.Parallel()
+
+	//t.Log("Starting test: Ensure the Gateway API CRDs exist")
```
alebedev87, line 37: What I meant was:

```suggestion
	// Starting test: Ensure the Gateway API CRDs exist
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
...
+	crdsExist := checkCRDsExist(t)
+	configMap := &corev1.ConfigMap{}
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, client.ObjectKey{Namespace: "openshift-config-managed", Name: "admin-gates"}, configMap); err != nil {
+			//t.Logf("Failed to get configmap admin-gates: %v", err)
+			if errors.IsNotFound(err) {
+				//t.Logf("ConfigMap 'admin-gates' not found in namespace openshift-config-managed")
+			} else {
+				//t.Logf("Error getting ConfigMap: %v", err)
+			}
```
alebedev87, line 80: This log message, on the other hand, does make sense as it signals a problem:
```suggestion
			t.Logf("Failed to get configmap admin-gates: %v, retrying...", err)
```
And yes, you don't need to make a distinction between error types because the error will contain the needed description. E.g. if it's "NotFound" error, the description will say so.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("Failed to get configmap admin-gates after polling: %v", err)
```
alebedev87, line 86:
```suggestion
		t.Errorf("Timed out trying to get admin-gates configmap: %v", err)
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
...
+		//t.Log("Admin gate key exists in the configmap")
+	} else {
+		//	t.Log("Admin gate key does not exist in the configmap")
+	}
+	if crdsExist != shouldExist {
+		t.Errorf("Expected CRDs existence: %v, but got: %v", shouldExist, crdsExist)
+	}
+	if adminGateKeyExists != shouldExist {
+		t.Errorf("Expected admin gate key existence: %v, but got: %v", shouldExist, adminGateKeyExists)
+	}
```
alebedev87, line 102: I still don't get why you check for the existence of CRDs, you ensured their existence as the first step of `TestGatewayAPIUpgradeable` test (`createCRDs` function).
```suggestion
	_, adminGateKeyExists := configMap.Data[gatewayapi_upgradeable.GatewayAPIAdminKey]
        return adminGateKeyExists
```

alebedev87 (commented):
Another look at the e2e.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,124 @@
+//go:build e2e
```
shaneutt, line 1: Throughout this file there are several commented out log lines that would appear to need to be uncommented, or changed into regular comments?

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	// Mapping the events to the admin gate ConfigMap.
+	toAdminGateConfigMap := func(_ context.Context, _ client.Object) []reconcile.Request {
+		return []reconcile.Request{{
+			NamespacedName: operatorcontroller.AdminGatesConfigMapName(),
+		}}
+	}
+
+	// Defining the CRD predicate.
+	crdPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group == gatewayapiv1.GroupName
```
shaneutt, line 65: It occurs to me that we technically need to cover the new [gateway.networking.x-k8s.io](https://github.com/kubernetes-sigs/gateway-api/blob/2b0b9ceff08f1ac7899efa25657480d3196b1185/config/crd/experimental/gateway.networking.x-k8s.io_listenersets.yaml#L11) group here as well.

```suggestion
		group := o.(*apiextensionsv1.CustomResourceDefinition).Spec.Group
		return group == gatewayapiv1.GroupName || group == "gateway.networking.x-k8s.io"
```

(note: feel free to merge this suggestion, but don't use the suggestion verbatim switch to an import)

shaneutt (commented):
Second pass!

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: So initially, I used kclient (for whatever reason), however when I looked around other test files, I found the standard practice for retrieving the kubeconfig as follows:

```
kubeConfig, err := config.GetConfig()
	if err != nil {
		t.Fatalf("failed to get kube config: %v", err)
	}
	client, err := kubernetes.NewForConfig(kubeConfig)
	if err != nil {
		t.Fatalf("failed to create kube client: %v", err)
	}
```

To maintain consistency with the existing tests, I have adopted this approach in my code.
Does this approach work, or should I revert to using the package global kclient variable instead?

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: I removed the excessive logs and comments, just kept what is needed for simplicity

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Correct, I removed the excessive logs and comments, keeping only what is needed

Thealisyed:
/retest 

TestNodePool and unrelated to changes

Thealisyed:
/retest 
Node installer timeout

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	if gatewayAPIEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
+		t.Fatalf("error checking feature gate enabled status: %v", err)
+	} else if gatewayAPIEnabled {
+		t.Skip("Gateway API is enabled, skipping TestGatewayAPIUpgradeable")
+	}
+
+	t.Parallel()
```
alebedev87, line 43: This is better to be at the very start of the test function.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+		manifests.HTTPRouteCRD(),
+		manifests.ReferenceGrantCRD(),
+	}
+
+	for _, crd := range crds {
+		if err := kclient.Create(context.TODO(), crd); err != nil {
+			if !errors.IsAlreadyExists(err) {
+				t.Fatalf("Failed to create CRD %s: %v", crd.Name, err)
+			}
+			continue
```
alebedev87, line 64:
```suggestion
			}
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+}
+
+func testAdminGate(t *testing.T, shouldExist bool) {
+	t.Helper()
+	//	crdsExist := checkCRDsExist(t)
+	configMap := &corev1.ConfigMap{}
```
alebedev87, line 72:
```suggestion
	adminGatesConfigMap := &corev1.ConfigMap{}
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+		if err := kclient.Get(ctx, client.ObjectKey{Namespace: "openshift-config-managed", Name: "admin-gates"}, configMap); err != nil {
+			t.Logf("Failed to get configmap admin-gates: %v, retrying...", err)
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("Timed out trying to get admin-gates configmap: %v", err)
+		return
+	}
```
alebedev87, line 83:
```suggestion
	if err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
		if err := kclient.Get(ctx, client.ObjectKey{Namespace: "openshift-config-managed", Name: "admin-gates"}, configMap); err != nil {
			t.Logf("Failed to get configmap admin-gates: %v, retrying...", err)
			return false, nil
		}
		return true, nil
	}); err != nil {
		t.Fatalf("Timed out trying to get admin-gates configmap: %v", err)
	}
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	_, adminGateKeyExists := configMap.Data[gatewayapi_upgradeable.GatewayAPIAdminKey]
+	if adminGateKeyExists != shouldExist {
+		t.Errorf("Expected admin gate key existence to be %v, but got %v", shouldExist, adminGateKeyExists)
```
alebedev87, line 87:
```suggestion
		t.Fatalf("Expected admin gate key existence to be %v, but got %v", shouldExist, adminGateKeyExists)
```

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	controllerName     = "gatewayapi_upgradeable_controller"
+	GatewayAPIAdminKey = "ack-gateway-api-management"
+	GatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. Tampering with the prevention mechanism will cause the cluster to move to a 'Degraded' state. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See [link to doc pending] for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
+)
+
+var (
+	log = logf.Logger.WithName(controllerName)
+)
+
+var kclient client.Client
```
alebedev87, line 34: What is it for?! Didn't use mix this with e2e test accidentally? Your controller will use the client from the controller-runtime's manager.

pkg/operator/operator.go
```diff
@@ -319,6 +319,13 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		return nil, fmt.Errorf("failed to create gatewayapi controller: %w", err)
 	}
 
+	// Add conditional setup for gateway_upgradeable controller only if FeatureGate is enabled
```
alebedev87, line 322: The comment is telling the opposite of what is happening.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	GatewayAPIAdminKey = "ack-gateway-api-management"
+	GatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. Tampering with the prevention mechanism will cause the cluster to move to a 'Degraded' state. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See [link to doc pending] for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
```
alebedev87, line 27: We agreed on the doc link during our last call: `https://docs.openshift.com/container-platform/4.19/release_notes/ocp-4-19-release-notes.html#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes`. Can you please confirm @grzpiotrowski?

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The function checks if the admin key exists and if it is set to the expected message.
+	if val, ok := adminGateConfigMap.Data[GatewayAPIAdminKey]; ok && val == GatewayAPIAdminMsg {
+		return nil
+	}
+	adminGateConfigMap.Data[GatewayAPIAdminKey] = GatewayAPIAdminMsg
+
+	log.Info("Adding admin gate for Gateway API management")
+	if err := r.client.Patch(ctx, adminGateConfigMap, client.MergeFrom(originalAdminGateConfigMap)); err != nil {
+		return fmt.Errorf("failed to update configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
```
alebedev87, line 145: You can do `r.clientUpdate()` instead, like we do in most of our controllers. This will avoid the need for the copying of the admingates configmap. Btw your error message says `update` too.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// The addAdminGate function is responsible for adding the admin gate to the ConfigMap.
+func (r *reconciler) addAdminGate(ctx context.Context) error {
+	adminGateConfigMap := &corev1.ConfigMap{}
```
alebedev87, line 126:
```suggestion
	adminGatesConfigMap := &corev1.ConfigMap{}
```

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	log.Info("Adding admin gate for Gateway API management")
+	if err := r.client.Patch(ctx, adminGateConfigMap, client.MergeFrom(originalAdminGateConfigMap)); err != nil {
+		return fmt.Errorf("failed to update configmap %s: %w", operatorcontroller.AdminGatesConfigMapName(), err)
+	}
+	return nil
+}
+
+// The removeAdminGate function is responsible for removing the admin gate from the ConfigMap.
+func (r *reconciler) removeAdminGate(ctx context.Context) error {
+	adminGateConfigMap := &corev1.ConfigMap{}
```
alebedev87, line 152:
```suggestion
	adminGatesConfigMap := &corev1.ConfigMap{}
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Errorf("Timed out trying to get admin-gates configmap: %v", err)
+		return
+	}
+
+	_, adminGateKeyExists := configMap.Data[gatewayapi_upgradeable.GatewayAPIAdminKey]
```
alebedev87, line 85: The tests should avoid using the same variables as used in the code they test. Otherwise we risk to repeat mistakes made in the code.
```suggestion
	_, adminGateKeyExists := configMap.Data["ack-gateway-api-management"]
```

And please unexport the admin key and admin message variables from the controller package.

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+	testAdminGate(t, false)
+}
+
+func createCRDs(t *testing.T) {
+	crds := []*apiextensionsv1.CustomResourceDefinition{
+		manifests.GatewayClassCRD(),
+		manifests.GatewayCRD(),
+		manifests.HTTPRouteCRD(),
+		manifests.ReferenceGrantCRD(),
+	}
```
alebedev87, line 57: Same remark about using the code in the tests. Your controller checks an existence of any CRD from 2 API groups. For the purpose of your test it's enough to build 2 test CRDs (see [an example of a test CRD](https://github.com/openshift/cluster-ingress-operator/blob/7dc12cf28557f8c80460435649cfc393761e0285/test/e2e/util_gatewayapi_test.go#L353-L369)).

test/e2e/gatewayapi_upgradeable_test.go
alebedev87: Just remove `kubeinit()` completely, you don't need it.

alebedev87:
@Thealisyed : can you please update the PR description too?

pkg/operator/controller/gatewayapi-upgradeable/controller.go
grzpiotrowski: Since `docs.openshift.com` is decommissioned in favour of `docs.redhat.com` the doc link should most likely be:
`https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/release_notes/ocp-4-19-release-notes.html#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes`

shaneutt:
```suggestion
	GatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. Tampering with the prevention mechanism will cause the cluster to move to a 'Degraded' state. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
```

shaneutt (commented):
Did another pass. Most of my comments are covered already by existing comments :+1:

Thealisyed:
This PR introduces the following:

Initialize the GatewayAPI upgradeable controller

Added the required imports and constants.
Defined predicates for CRD and ConfigMap events to manage the admin gate.
Configured watches in the controller for specific changes in the ConfigMap.
Implement the reconciler logic

Added the Reconcile function to manage the admin gate addition and removal based on conditions
Enhance the operator initialization

Conditionally initialize the gatewayapi_upgradeable controller based on the GatewayAPI feature gate.
End-to-End Tests

Added the TestGatewayAPIUpgradeable end-to-end test.
Included the new test case in the test suite.

alebedev87 (approved):
Removing "requested changes".

Thealisyed:
/retest 

Time out failures and unrelated to changes

Thealisyed:
/retest 

Cluster Authentication, timeout and unrelated to changes failure again

Thealisyed:
/retest 

TestHeaderName test is failing unrelated to changes

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,163 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	gatewayAPIAdminKey = "ack-gateway-api-management"
+	gatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. Tampering with the prevention mechanism will cause the cluster to move to a 'Degraded' state. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
```
alebedev87, line 27: I would like to avoid mentioning any details from the development which is currently ongoing (https://issues.redhat.com/browse/NE-1969).
```suggestion
	gatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
```

Thealisyed: Makes sense, fixed it!

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,163 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	controllerName     = "gatewayapi_upgradeable_controller"
+	gatewayAPIAdminKey = "ack-gateway-api-management"
```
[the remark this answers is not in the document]
alebedev87: > The format of the admin gate should mention the version it applies to.

And maybe the version it blocks the upgrade to as well (e.g. `ack-4.18-gateway-api-management-in-4.19`). Something to be checked with @wking.

alebedev87 (changes requested):

pkg/operator/controller/gatewayapi-upgradeable/controller.go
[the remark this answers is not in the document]
alebedev87: For the reference, th EP which describes the format: https://github.com/openshift/enhancements/blob/16ec2e8ba16e8cd53e3744b9678aa02c4f3f4f32/enhancements/update/upgrades-blocking-on-ack.md.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,163 @@
...
+	gatewayAPIAdminMsg = "Gateway API CRDs have been detected. OCP fully manages the life-cycle of Gateway API CRDs. External management is unsupported and will be prevented. The cluster administrator is responsible for the safety of existing Gateway API implementations and must acknowledge their responsibilities via the admin gate to proceed with upgrades. See https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes#ocp-4-19-networking-gateway-api-crd-lifecycle_release-notes for details. Failure to read and understand the documentation for this and the implications can result in outages and data loss."
```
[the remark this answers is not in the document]
Miciah: https://github.com/openshift/openshift-docs/pull/89357 adds the 4.19 release notes outline, with a placeholder section for this link to target.  We'll need to check with the docs team as to when the docs will go live.  

(Even once it initially goes "live", that just means it'll say "We will have details here when OpenShift Container Platform 4.19 is released" instead of a 404.)

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Miciah: Still missing the blank line here.

pkg/operator/controller/gatewayapi-upgradeable/controller.go
```diff
@@ -0,0 +1,163 @@
...
+	})
+
+	if err := c.Watch(source.Kind[client.Object](mgr.GetCache(), &corev1.ConfigMap{}, &handler.EnqueueRequestForObject{}, adminGatePredicate)); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+// Reconciler struct holds the client and cache attributes.
```
Miciah, line 78: Make sure the name in the godoc matches the type's actual name.  
```suggestion
// reconciler struct holds the client and cache attributes.
```

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,90 @@
...
+func TestGatewayAPIUpgradeable(t *testing.T) {
+	t.Parallel()
+	if gatewayAPIEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPI); err != nil {
+		t.Fatalf("error checking feature gate enabled status: %v", err)
+	} else if gatewayAPIEnabled {
+		t.Skip("Gateway API is enabled, skipping TestGatewayAPIUpgradeable")
+	}
+	createCRDs(t)
+	testAdminGate(t, true)
+	deleteExistingCRD(t, "gatewayclasses.gateway.networking.k8s.io")
```
Miciah, line 31: The test helpers are inconsistent:

- `createCRDs` appears to create only 1 CRD, so the name is misleading.
- The definition of `createCRDs` hard-codes a CRD whereas `deleteExistingCRD` takes the CRD name as a parameter, which is inconsistent.  

Aside from that, the test should have a `t.Cleanup` handler to delete the CRD that `createCRDs` creates.  Suppose you ran the test twice and that `testAdminGate` failed the first time; then the second time you ran the test, `createCRDs` would fail because the test would have left behind the CRD the first time around.  Where feasible, tests should be idempotent.  

It would also be helpful to have godoc on `TestGatewayAPIUpgradeable` to explain what it does.

Miciah:
This is a low-risk change.  The impact is limited to a single new controller, similar to the one in https://github.com/openshift/cluster-ingress-operator/pull/1014/commits/6778ce1a619b2f58310a037694e341e446888c55, to update the "admin-gates" configmap.  The logic is well contained and has E2E test coverage.  

/label backport-risk-assessed

This is a low-risk change. The impact is limited to the operator's clusteroperator status reporting logic, it is well contained, and it has good test coverage. (Thanks for the detailed comments and commit messages!)

pkg/operator/controller/gatewayapi-upgradeable/controller.go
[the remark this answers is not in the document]
wking: So do y'all want to create a KCS or other document you can release independently, so folks who see this message while running a 4.18 release can onboard themselves with that context without waiting for 4.19 to GA?  They could theoretically be working through walking back any Gateway API CRD managers they run today, if they had access to those docs, and be ready to move to 4.19 on GA-day.

alebedev87:
@Thealisyed: Thanks for addressing all the comments!

/lgtm

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -80,3 +96,22 @@ func deleteExistingCRDs(t *testing.T, crds []*apiextensionsv1.CustomResourceDefi
...
+		if err := kclient.Get(ctx, client.ObjectKey{Namespace: "openshift-config-managed", Name: "admin-gates"}, adminGatesConfigMap); err != nil {
+			t.Logf("Failed to get configmap admin-gates: %v, retrying...", err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Timed out trying to get admin-gates configmap: %v", err)
+	}
+
+	_, adminGateKeyExists := adminGatesConfigMap.Data["ack-gateway-api-management"]
```
grzpiotrowski, line 113: I saw some failed tests and got curious as it seemed to be fine earlier.
Is this the culprit and should be `ack-4.18-gateway-api-management-in-4.19`?

alebedev87: > Is this the culprit and should be ack-4.18-gateway-api-management-in-4.19?

Yes. Good catch!

Thealisyed: Added that! However I can see the test still failing for some reason

Thealisyed:
```
=== NAME  TestAll/parallel/TestGatewayAPIUpgradeable
    gatewayapi_upgradeable_test.go:58: Expected admin gate key existence to be true, but got false
```

e2e-aws-operator

candita:
@Thealisyed I see many comments resolved with no answer from you. Can you please take a quick look through and add responses that explain what you did to address the comment?  Even just a "done", or thumbs-up would suffice.  Our team generally does not resolve an un-answered comment.

candita:
> == NAME  TestAll/parallel/TestAWSEIPAllocationsForNLB
>     util_test.go:1068: service openshift-ingress/openshift-ingress UID has not changed yet, retrying...
>     lb_eip_test.go:149: failed to delete and recreate service: context deadline exceeded

alebedev87 (approved):
Removing "change requested" [(change)](https://github.com/openshift/cluster-ingress-operator/compare/bd93c36050048bd025308a30d08d8373723b71ca..c22c21d56ad6f0169d72f3890d6c17ab58c7a44f).

Thealisyed:
/retest 

Static nodes not created and other unrelated to changes

alebedev87:
Build of the CIO image took too much time:
```
Build cluster-ingress-operator-amd64 succeeded after 1h4m8s 
```

Due to some problems with the internal image registry:
```
Pushing image image-registry.openshift-image-registry.svc:5000/ci-op-0s543lfi/pipeline:cluster-ingress-operator-amd64 ...
Warning: Push failed, retrying in 5s ...
Getting image source signatures
Warning: Push failed, retrying in 5s ...
Getting image source signatures
Warning: Push failed, retrying in 5s ...
Registry server Address: 
Registry server User Name: <token>
Registry server Email: 
Registry server Password: <<non-empty>>
error: build error: Failed to push image: trying to reuse blob sha256:25c75c34b2e2b68ba9245d9cddeb6b8a0887371ed30744064f85241a75704d87 at destination: unable to retrieve auth token: invalid username/password: authentication required
```

The CI job was forcibly stopped because the default 4h timeout was reached.

/retest-required

Thealisyed:
@candita Thanks for the feedback, and my apologies for missing that. I’ve been resolving the comments without adding responses to the reviewer, which is something Miciah and I had discussed earlier. I’ll go through the PR now and make sure to address the remaining comments.

Thealisyed:
/test e2e-aws-ovn-serial

Due to some problems with the internal image registry:


```
error: build error: Failed to push image: trying to reuse blob sha256:25c75c34b2e2b68ba9245d9cddeb6b8a0887371ed30744064f85241a75704d87 at destination: unable to retrieve auth token: invalid username/password: authentication required

{"component":"entrypoint","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:169","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2025-03-21T04:34:06Z"}

```
The CI job was forcibly stopped because the default 4h timeout was reached.
Nothing to suggest issues with test so will retest now

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed that, thanks for that catch @grzpiotrowski!

pkg/operator/controller/gatewayapi-upgradeable/controller.go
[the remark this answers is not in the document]
Thealisyed: Had a look at EP and yes you are correct. I fixed the formate of the admin gate

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: After rebasing [PR](https://github.com/openshift/cluster-ingress-operator/pull/1193), `createCRDs` is reflective as it creates a range of CRDs. 
`t.Cleanup` was [added](https://github.com/openshift/cluster-ingress-operator/pull/1196/files#diff-28e96b945950136ab0fee2af650b3c6b166d1ae94ea2ebc12b42269dbb054bb8R36:~:text=t.Cleanup,%7D)) to check if any CRDs are left over and removes them.
A [godoc](https://github.com/openshift/cluster-ingress-operator/pull/1196/files#diff-28e96b945950136ab0fee2af650b3c6b166d1ae94ea2ebc12b42269dbb054bb8R36:~:text=//%20TestGatewayAPIUpgradeable%20verifies%20the,gate%20is%20disabled.) was added for `TestGatewayAPIUpgrdeable`

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Latest commit addresses this

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed this in latest commit

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Makes sense and I have implemented this in the latest commit

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Right, this was not correct as it was hardcoding the value, was fixed in the latest commit

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: The logic for this is not accurate, have updated and fixed this in the latest commit

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Fixed

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Fixed

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Amended this in the latest commit

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Correct, this is redundant

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: @shaneutt Thanks for sharing this, will look into it!

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Fixed to use `r.cache` now

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Fixed

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Done

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+				t.Fatalf("Failed to create CRD %s: %v", crd.Name, err)
+			}
+			continue
+		}
+	}
+}
+
+func testAdminGate(t *testing.T, shouldExist bool) {
+	t.Helper()
+	//	crdsExist := checkCRDsExist(t)
```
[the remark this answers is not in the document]
Thealisyed: Thanks, removed now

test/e2e/gatewayapi_upgradeable_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	t.Parallel()
+
+	createCRDs(t)
+	testAdminGate(t, true)
+	deleteCRDs(t)
+	testAdminGate(t, false)
+}
+
+func createCRDs(t *testing.T) {
```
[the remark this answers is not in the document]
Thealisyed: Added helper

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Apologies, seems to be a mix up, fixed now!

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Thanks, have fixed this link

pkg/operator/controller/gatewayapi-upgradeable/controller.go
Thealisyed: Corrected this!

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed

test/e2e/gatewayapi_upgradeable_test.go
Thealisyed: Fixed after rebased with Greg's [PR](https://github.com/openshift/cluster-ingress-operator/pull/1193/commits)

Thealisyed:
/test e2e-aws-ovn-serial

auth error: failed to push image

`error: build error: Failed to push image: trying to reuse blob sha256:25c75c34b2e2b68ba9245d9cddeb6b8a0887371ed30744064f85241a75704d87 at destination: unable to retrieve auth token: invalid username/password: authentication required
`

Failed at 4 hour timeout:

`Process did not finish before 4h0m0s timeout`

It has failed before in Job history for this reason so retesting

PR #1197, opened 2025-03-05
CORS-3907: Update ingress operator to with custom endpoints

barbacbd:
Enhancement link for the feature: https://github.com/openshift/enhancements/pull/1734

** Vendor update
** Set the endpoint information in the Config for GCP
** If an endpoint exists for the DNS service use the endpoint as an option when creating the new client.

### DIFF
--- a/pkg/dns/gcp/provider.go
+++ b/pkg/dns/gcp/provider.go
@@ -31,13 +31,30 @@ type Provider struct {
 }
 
 type Config struct {
-	Project         string
-	UserAgent       string
-	CredentialsJSON []byte
+	Project                   string
+	UserAgent                 string
+	CredentialsJSON           []byte
+	Endpoints                 []configv1.GCPServiceEndpoint
+	GCPCustomEndpointsEnabled bool
 }
 
 func New(config Config) (*Provider, error) {
-	dnsService, err := gdnsv1.NewService(context.TODO(), option.WithCredentialsJSON(config.CredentialsJSON), option.WithUserAgent(config.UserAgent))
+	options := []option.ClientOption{
+		option.WithCredentialsJSON(config.CredentialsJSON),
+		option.WithUserAgent(config.UserAgent),
+	}
+	if config.GCPCustomEndpointsEnabled {
+		for _, endpoint := range config.Endpoints {
+			if endpoint.Name == configv1.GCPServiceEndpointNameDNS {
+				// There should be at most 1 endpoint override per service. If there
+				// is more than one, only use the first instance.
+				options = append(options, option.WithEndpoint(endpoint.URL))
+				break
+			}
+		}
+	}
+
+	dnsService, err := gdnsv1.NewService(context.TODO(), options...)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -114,6 +114,10 @@ type Config struct {
 	// PrivateHostedZoneAWSEnabled indicates whether the "SharedVPC" feature gate is
 	// enabled.
 	PrivateHostedZoneAWSEnabled bool
+
+	// GCPCustomEndpointsEnabled indicates whether the "GCPCustomAPIEndpoints"
+	// feature gate is enabled.
+	GCPCustomEndpointsEnabled bool
 }
 
 type reconciler struct {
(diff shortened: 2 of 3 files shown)


pkg/dns/gcp/provider.go
```diff
@@ -34,10 +34,22 @@ type Config struct {
...
+	options := []option.ClientOption{
+		option.WithCredentialsJSON(config.CredentialsJSON),
+		option.WithUserAgent(config.UserAgent),
+	}
+	for _, endpoint := range config.Endpoints {
+		if endpoint.Name == configv1.GCPServiceEndpointNameDNS {
+			options = append(options, option.WithEndpoint(endpoint.URL))
+			break
+		}
+	}
```
patrickdillon, line 50: Potentially you could wrap this for loop with the feature gate... I think that would be _technically_ correct

The installer will only lay down the config if the feature gate is enabled, so it is already _effectively_ gated. I would defer to the preferences of the owning team.

patrickdillon (commented):
Enhancement link for more context: Enhancement link for more context: https://github.com/openshift/enhancements/pull/1734

@barbacbd might be worth adding to the description

Also worth noting that this is the equivalent of the existing AWS implementation here:

https://github.com/openshift/cluster-ingress-operator/blob/69cadab177d66bb2324897db4be50ae67fab1fc7/pkg/dns/aws/dns.go#L209-L243

For AWS there is an LB and tagging client as well, but not sure why. Those should be handled by cloud-provider-aws AFAIK

This looks simple enough I feel comfortable to LGTM.

Miciah:
> For AWS there is an LB and tagging client as well, but not sure why. Those should be handled by cloud-provider-aws AFAIK

The tagging client is used to look up the hosted zone by tags, and the elb client is used to look up the target hosted zone from the ELB.

pkg/operator/controller/dns/controller.go
```diff
@@ -114,6 +114,7 @@ type Config struct {
 	// PrivateHostedZoneAWSEnabled indicates whether the "SharedVPC" feature gate is
 	// enabled.
 	PrivateHostedZoneAWSEnabled bool
+	CustomEndpointsEnabled      bool
```
Miciah, line 117: The new field should have "GCP" in the name in order to make it obvious that this is GCP-specific, and to avoid a potential conflict if we eventually add a similar featuregate for Azure (the other platforms already have APIs for spedcifying custom service endpoints, so it looks like Azure is the only potential future conflict).  The field should also have godoc.  
```suggestion
	// CustomEndpointsEnabled indicates whether the "GCPCustomAPIEndpoints"
	// featuregate is enabled.
	CustomEndpointsEnabled bool
```

Miciah (commented):
@patrickdillon, @barbacbd, since the GCP custom endpoints feature involves multiple components but is being implemented by the Installer team developers, is the Installer team QE going to verify the changes in cluster-ingress-operator, or is that going to be entirely NI&D QE's responsibility?

pkg/operator/controller/dns/controller.go
barbacbd: Thanks @Miciah I made those changes. One of the tests I believe is still failing so I need to look into that though.

Miciah:
Thanks!  The code changes look good.

Would you mind adding some details to the commit message?  Something like this:
```text
dns/gcp: Allow configuring custom endpoints

If a custom service endpoint has been configured for DNS, use it when
creating the client for the Google Cloud DNS service.

This feature is behind the "GCPCustomAPIEndpoints" featuregate.

This commit resolves CORS-3907.

https://issues.redhat.com/browse/CORS-3907

* pkg/dns/gcp/provider.go (Config): Add GCPCustomEndpointsEnabled field
to the provider config.
(New): If a custom endpoint is set and the featuregate is enabled,
configure the new client to use the endpoint.
* pkg/operator/controller/dns/controller.go (Config): Add
GCPCustomEndpointsEnabled field.
(createDNSProvider): Pass GCPCustomEndpointsEnabled from the controller
config to the GCP provider config.
* pkg/operator/operator.go (New): Set the controller config field based
on the featuregate.
```
I like to have file-by-file changes, but those are less important.  What are more important are the summary and first paragraph to describe the intent of the change, and then the reference to the Jira issue.

Miciah:
Thanks!

/approve
/lgtm

PR #1198, opened 2025-03-06
NE-1954: Implement GatewayAPIController feature gate

alebedev87:
This PR bumps `openshift/api` to get  [GatewayAPIController feature gate](https://github.com/openshift/api/pull/2219)  and implements it by skipping the start of the gatewayclass and service-dns controllers when it's disabled.


Miciah:
Besides gatewayclass_controller, service_dns_controller also should run only if the GatewayAPIController featuregate is enabled.  

(In retrospect, maybe some of the changes in dns_controller should have been put behind the gate, but I don't think it is worth making that change now.)

Another way to implement the GatewayAPIController featuregate would be to specify an empty or nil value for `DependentControllers` if the GatewayAPIController featuregate is not enabled.  The advantages of that approach would be (1)  it would avoid the overhead of starting the controllers and their watches and (2) it would make the featuregate that much more comprehensive (as much as possible, turning off the featuregate should turn off _all_ of the feature's logic).

alebedev87:
> Another way to implement the GatewayAPIController featuregate would be to specify an empty or nil value for DependentControllers if the GatewayAPIController featuregate is not enabled. The advantages of that approach would be (1) it would avoid the overhead of starting the controllers and their watches and (2) it would make the featuregate that much more comprehensive (as much as possible, turning off the featuregate should turn off all of the feature's logic).

Yeah, I was thinking about this too. I went the known path - [Gateway API controller starts even if the corresponding featuregate is off](https://github.com/openshift/cluster-ingress-operator/blob/69cadab177d66bb2324897db4be50ae67fab1fc7/pkg/operator/controller/gatewayapi/controller.go#L100-L102). Also, since [I only considered the gatewyclass controller](https://github.com/openshift/cluster-ingress-operator/pull/1198#issuecomment-2706467462) I didn't want to introduce a more complex logic of detecting which dependent controller is the one I should not start.

Miciah:
> Well, yeah, I was thinking whether the service-dns-controller should be part of the GatewayAPI Controller. My logic was - since it's possible to create all resources the service-dns controller watches: `Gateway` CR (`GatewayAPI` featuregate controls CRDs), `Service` and `DNSRecord` CR, the service-dns controller can work without OSSM. There still the labeling part though (`istio.io/gateway-name`, `gateway.istio.io/managed`) but this can be left on purpose so that the users can manage Gateway DNS without OSSM installation. I can imagine a completely hypothetical scenario - cluster admin wants to install the CRDs and manage Gateway DNS first and only after allow the controller installation.

Since the purpose of the featuregate is to prevent the feature from impacting clusters with the default featureset, it really should be a switch for as much of the feature's logic as possible (with reasonable effort).  

> Yeah, I was thinking about this too. I went the known path - [Gateway API controller starts even if the corresponding featuregate is off](https://github.com/openshift/cluster-ingress-operator/blob/69cadab177d66bb2324897db4be50ae67fab1fc7/pkg/operator/controller/gatewayapi/controller.go#L100-L102). Also, since [I only considered the gatewyclass controller](https://github.com/openshift/cluster-ingress-operator/pull/1198#issuecomment-2706467462) I didn't want to introduce a more complex logic of detecting which dependent controller is the one I should not start.

In retrospect again, we probably would have done the logic differently today.  When that controller was added, the operator didn't check featuregates during its initialization and didn't have logic to terminate when featuregates changed, so the controller itself watched the featuregate.  Now that operator initialization has that logic, it makes more sense for the operator initialization to decide whether to start a controller or not.

alebedev87:
/retitle [WIP] NE-1954: Implement GatewayAPIController feature gate

alebedev87:
/retitle NE-1954: Implement GatewayAPIController feature gate

alebedev87:
Another occurence of https://github.com/openshift/cluster-ingress-operator/pull/1192#issuecomment-2705197358:
```
=== NAME  TestAll/parallel/TestConnectTimeout
    operator_test.go:3249: failed to create pod openshift-ingress/connect-timeout-http: Internal error occurred: failed calling webhook "sidecar-injector.istio.io": failed to call webhook: Post "https://istiod-openshift-gateway.openshift-ingress.svc:443/inject?timeout=10s": no endpoints available for service "istiod-openshift-gateway"
--- FAIL: TestAll/parallel/TestConnectTimeout (0.59s)
```

Miciah:
Thanks!

/approve
/lgtm

PR #1199, opened 2025-03-10
NE-1981: Move controller test helpers to dedicated package

alebedev87:
This PR:
- extracts test helpers that were used across multiple controllers into a dedicated package: `pkg/operator/controller/test/util`. This includes `FakeCache`, `FakeController`, and `FakeClientRecorder`.
- allows `test` target to run specific unit tests using `-run` flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,13 @@ endif
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
 GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
 
-TEST ?= ^TestAll$
+ifdef TEST
+TEST_UNIT ?= $(TEST)
+TEST_E2E ?= $(TEST)
+else
+TEST_UNIT ?= ^.*$
+TEST_E2E ?= ^TestAll$
+endif
 
 .PHONY: build
 build: generate
@@ -45,7 +51,7 @@ manifests:
 
 .PHONY: test
 test: generate
-	CGO_ENABLED=1 $(GO) test ./...
+	CGO_ENABLED=1 $(GO) test -run "$(TEST_UNIT)" ./...
 
 .PHONY: release-local
 release-local:
@@ -53,7 +59,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e: generate
-	CGO_ENABLED=1 $(GO) test -timeout 1.5h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	CGO_ENABLED=1 $(GO) test -timeout 1.5h -count 1 -v -tags e2e -run "$(TEST_E2E)" ./test/e2e
 
 .PHONY: test-e2e-list
 test-e2e-list: generate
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -45,7 +45,7 @@ manifests:
 
 .PHONY: test
 test: generate
-	CGO_ENABLED=1 $(GO) test ./...
+	CGO_ENABLED=1 $(GO) test -run "$(TEST)" ./...
```
Miciah, line 48: I'm ambivalent about having this change combined with the test refactoring.  It needs to be called out in the commit message, and it logically seems like it should be a separate commit.

Miciah (commented):
I created a Jira issue for this: https://issues.redhat.com/browse/NE-1981

Please add "NE-1981" to the PR title, and add a link to the issue in the commit message.  

I feel like the `-run "$(TEST)"`change should be a separate commit.  It should at least be called out in the commit message.  

Other than that, these changes look good.  Thanks!

/approve

Makefile
alebedev87: I put this change into a separate commit. Btw, I found a problem in using the same `TEST` variable for both e2e and unit tests - the default value works only with the e2e tests, for unit tests it passes all the time (because `TestAll` doesn't exist). I added a new variable for unit tests: `TESTU`.

Miciah: What do you think about the following?
```diff
diff --git a/Makefile b/Makefile
index ff26aa24b..f43ffc4d2 100644
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,13 @@ endif
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
 GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(MAIN_PACKAGE)
 
-TEST ?= ^TestAll$
+ifeq ($(strip $(TEST)),)
+	TEST_UNIT ?= ^.*$
+	TEST_E2E ?= ^TestAll$
+else
+	TEST_UNIT ?= $(TEST)
+	TEST_E2E ?= $(TEST)
+endif
 
 .PHONY: build
 build: generate
@@ -45,7 +51,7 @@ manifests:
 
 .PHONY: test
 test: generate
-	CGO_ENABLED=1 $(GO) test ./...
+	CGO_ENABLED=1 $(GO) test -run "$(TEST_UNIT)" ./...
 
 .PHONY: release-local
 release-local:
@@ -53,7 +59,7 @@ release-local:
 
 .PHONY: test-e2e
 test-e2e: generate
-	CGO_ENABLED=1 $(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	CGO_ENABLED=1 $(GO) test -timeout 1h -count 1 -v -tags e2e -run "$(TEST_E2E)" ./test/e2e
 
 .PHONY: test-e2e-list
 test-e2e-list: generate
```
This approach is verbose but it is easy to understand and makes it easy to override the pattern for just E2E tests; just unit tests; or both, with either a single pattern for both or separate patterns.

alebedev87: Yeah, this looks good to me. `TESTU` is an ugly name.

Miciah:
Thanks!

/lgtm

PR #1200, opened 2025-03-13
NE-1953: Add experimental Gateway API group to Validating Admission Policy

alebedev87:
This PR adds `gateway.networking.x-k8s.io` to the match condition of the operator's Validating Admission Policy. This change ensures that modifications to experimental Gateway API CRDs can be made only by the ingress operator.

### DIFF
--- a/manifests/01-validating-admission-policy.yaml
+++ b/manifests/01-validating-admission-policy.yaml
@@ -20,7 +20,7 @@ spec:
     # Consider only request to Gateway API CRDs.
     - name: "check-only-gateway-api-crds"
       # When the operation is DELETE, the "object" variable is null.
-      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io' || (request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.x-k8s.io'"
   # Validations are evaluated in the the order of their declaration.
   validations:
     # Verify that the request was sent by the ingress operator's service account.
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -109,7 +109,6 @@ func TestGatewayAPI(t *testing.T) {
 // CRDs are created.
 // It also deletes and ensure the CRDs are recreated.
 func testGatewayAPIResources(t *testing.T) {
-	t.Helper()
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
 	ensureCRDs(t)
 
@@ -128,8 +127,6 @@ func testGatewayAPIResources(t *testing.T) {
 // - the SMCP is created successfully (OSSM 2.x).
 // - deletes SMCP and subscription and tests if it gets recreated
 func testGatewayAPIIstioInstallation(t *testing.T) {
-	t.Helper()
-
 	if err := assertSubscription(t, openshiftOperatorsNamespace, expectedSubscriptionName); err != nil {
 		t.Fatalf("failed to find expected Subscription %s: %v", expectedSubscriptionName, err)
 	}
@@ -166,8 +163,6 @@ func testGatewayAPIIstioInstallation(t *testing.T) {
 
 // testGatewayAPIObjects tests that Gateway API objects can be created successfully.
 func testGatewayAPIObjects(t *testing.T) {
-	t.Helper()
-
 	// Create a test namespace that cleans itself up and sets up its own service account and role binding.
 	ns := createNamespace(t, names.SimpleNameGenerator.GenerateName("test-e2e-gwapi-"))
 
@@ -189,8 +184,6 @@ func testGatewayAPIObjects(t *testing.T) {
 // denies admission requests attempting to modify Gateway API CRDs on behalf of a user
 // who is not the ingress operator's service account.
 func testGatewayAPIResourcesProtection(t *testing.T) {
-	t.Helper()
-
 	// Get kube client which impersonates ingress operator's service account.
 	kubeConfig, err := config.GetConfig()
 	if err != nil {
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -47,6 +47,10 @@ var crdNames = []string{
 	"referencegrants.gateway.networking.k8s.io",
 }
 
+var xcrdNames = []string{
+	"listenersets.gateway.networking.x-k8s.io",
+}
+
 // Global variables for testing.
 // The default route name to be constructed.
 var defaultRoutename = ""
@@ -82,6 +86,13 @@ func TestGatewayAPI(t *testing.T) {
 		}
 		// TODO: Uninstall OSSM after test is completed.
 	})
+
+	// Create test experimental CRDs for the subsequent subtests.
+	// Specifically, `testGatewayAPIResourcesProtection`, which tests VAP protection
+	// for the experimental Gateway API group, needs to check the update verb.
+	// Since an API `Get` is called before the update, the CRD must exist in the cluster,
+	// just like standard Gateway API CRDs.
+	ensureExperimentalCRDs(t)
 
 	t.Run("testGatewayAPIResources", testGatewayAPIResources)
 	if gatewayAPIControllerEnabled {
@@ -195,7 +206,7 @@ func testGatewayAPIResourcesProtection(t *testing.T) {
 
 	// Create test CRDs.
 	var testCRDs []*apiextensionsv1.CustomResourceDefinition
-	for _, name := range crdNames {
+	for _, name := range append(crdNames, xcrdNames...) {
 		testCRDs = append(testCRDs, buildGWAPICRDFromName(name))
 	}
 
@@ -293,6 +304,26 @@ func deleteCRDs(t *testing.T) {
 		err := deleteExistingCRD(t, crdName)
 		if err != nil {
 			t.Errorf("failed to delete crd %s: %v", crdName, err)
+		}
+	}
+}
+
+// ensureExperimentalCRDs creates experimental Gateway API custom resource definitions.
+// This function temporarily disables the ingress operator's VAP to allow CRD creation.
+// The VAP is re-enabled before the function returns.
+func ensureExperimentalCRDs(t *testing.T) {
+	vm := newVAPManager(t, gwapiCRDVAPName)
+	if err, recoverFn := vm.disable(); err != nil {
+		defer recoverFn()
+		t.Fatalf("failed to disable vap: %v", err)
+	}
+	defer vm.enable()
+
+	for _, crdName := range xcrdNames {
+		if _, err := createCRD(crdName); err != nil {
+			t.Fatalf("failed to create experimental crd %q: %v", crdName, err)
+		} else {
+			t.Logf("created experimental crd %q", crdName)
 		}
 	}
 }
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -271,6 +271,21 @@ func createGatewayClass(name, controllerName string) (*gatewayapiv1.GatewayClass
 	return gatewayClass, nil
 }
 
+// createCRD creates the CRD with the given name or retrieves it if already exists.
+func createCRD(name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	crd := buildGWAPICRDFromName(name)
+	if err := kclient.Create(context.Background(), crd); err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			if err := kclient.Get(context.Background(), types.NamespacedName{Name: name}, crd); err != nil {
+				return nil, fmt.Errorf("failed to get crd %q: %w", name, err)
+			}
+			return crd, nil
+		}
+		return nil, fmt.Errorf("failed to create crd %q: %w", name, err)
+	}
+	return crd, nil
+}
+
 // buildGatewayClass initializes the GatewayClass and returns its address.
 func buildGatewayClass(name, controllerName string) *gatewayapiv1.GatewayClass {
 	return &gatewayapiv1.GatewayClass{
@@ -348,6 +363,12 @@ func buildGWAPICRDFromName(name string) *apiextensionsv1.CustomResourceDefinitio
 	case "referencegrants":
 		kind = "ReferenceGrant"
 		versions = []map[string]bool{{"v1beta1": true}}
+	case "listenersets":
+		kind = "ListenerSet"
+		versions = []map[string]bool{{"v1alpha1": true}}
+	case "grpcroutes":
+		kind = "GRPCRoute"
+		versions = []map[string]bool{{"v1": true}}
 	}
 
 	crd := &apiextensionsv1.CustomResourceDefinition{


test/e2e/util_gatewayapi_test.go
```diff
@@ -271,6 +271,21 @@ func createGatewayClass(name, controllerName string) (*gatewayapiv1.GatewayClass
...
 }
 
+// createCRD creates the CRD with the given name or retrieves it if already exists.
+func createCRD(name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	crd := buildGWAPICRDFromName(name)
+	if err := kclient.Create(context.TODO(), crd); err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			if err = kclient.Get(context.TODO(), types.NamespacedName{Name: name}, crd); err == nil {
+				return crd, nil
+			}
```
Miciah, line 281:
```suggestion
			if err = kclient.Get(context.TODO(), types.NamespacedName{Name: name}, crd); err != nil {
				return nil, err
			}
			return crd, nil
```
I fixed a similar logic error here: https://github.com/openshift/cluster-ingress-operator/pull/1152/commits/30b576c4293b2962b137bcb93ff77f278d7bf681

test/e2e/gateway_api_test.go
```diff
@@ -360,3 +366,23 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
 
 	return errs
 }
+
+// createExperimentalCRDs creates experimental Gateway API custom resource definitions.
+func createExperimentalCRDs(t *testing.T) {
+	t.Helper()
```
Miciah, line 372: Is `createExperimentalCRDs` really a helper?  It looks more like a test.  If it failed, I'd want to get a line number of the point of failure within `createExperimentalCRDs`, not the call site of `createExperimentalCRDs`.  
```suggestion
// testExperimentalCRDs creates experimental Gateway API custom resource definitions.
func testExperimentalCRDs(t *testing.T) {
```

alebedev87:
Gateway wasn't accepted and istio operator didn't come up:
```
=== RUN   TestAll/serial/TestGatewayAPI/testGatewayAPIObjects
    . . .
    util_gatewayapi_test.go:564: found gateway class openshift-default, but it is not yet Accepted. Retrying...
    util_gatewayapi_test.go:586: failed to get gateway test-gateway, retrying...
=== RUN   TestAll/serial/TestGatewayAPI/testGatewayAPIIstioInstallation
    . . .
    util_gatewayapi_test.go:479: failed to get deployment openshift-operators/istio-operator, retrying...
    gateway_api_test.go:95: failed to find expected Istio operator: error finding deployment openshift-operators/istio-operator: context deadline exceeded
```

Due to `connection refused` errors against `redhat-operators` catalog source service:
```
                {
                        "message": "error using catalogsource openshift-marketplace/redhat-operators: failed to list bundles: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 172.30.3.25:50051: connect: connection refused\"",
                        "reason": "ErrorPreventedResolution",
                        "status": "True",
                        "type": "ResolutionFailed"
```
The redhat-operators pod seemed to come up ready later because nothing can be found in [its logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1200/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview/1900314521021452288/artifacts/e2e-aws-operator-techpreview/gather-extra/artifacts/pods/openshift-marketplace_redhat-operators-22wf4_registry-server.log).

test/e2e/gateway_api_test.go
alebedev87: I tried to stay in "the traditions" of this test file which uses `t.Helper()` for all the tests: [1](https://github.com/openshift/cluster-ingress-operator/blob/a41be6eed16361903e5b8ddd0131ac7585e87da3/test/e2e/gateway_api_test.go#L106-L107), [2](https://github.com/openshift/cluster-ingress-operator/blob/a41be6eed16361903e5b8ddd0131ac7585e87da3/test/e2e/gateway_api_test.go#L125-L126), etc. 

I think the wording `create` does sound like a helper indeed. While `test` doesn't quite reflect the logic of the function. I think `ensureExperimentalCRDs` can be the middle ground here, highlighting the testing nature of the function.

test/e2e/util_gatewayapi_test.go
alebedev87: Thanks! Fixed the logic.

alebedev87:
```
--- FAIL: TestAll/serial/TestGatewayAPI/testGatewayAPIObjects
    . . .
    util_gatewayapi_test.go:858: failed to get DNSRecord openshift-ingress/test-gateway-68966c4f68-wildcard: client rate limiter Wait returned an error: context deadline exceeded, retrying...
    gateway_api_test.go:95: failed to observe successful status of one or more gateway object/s: context deadline exceeded
```

alebedev87:
Trying to play with the timeouts of `TestGatewayAPI` - increased for the objects which failed the recent operator-techpreview jobs.

alebedev87:
`TestGatewayAPI: Assert OSSM OLM objects before GWAPI objects` + increase `DNSRecord` polling timeout to 5 minutes.

lihongan:
/test e2e-ibmcloud-operator

just a test

alebedev87:
`e2e-aws-operator-techpreview` is green after the increase of the timeouts.

alebedev87:
https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r2001054549

test/e2e/gateway_api_test.go
Miciah: > I tried to stay in "the traditions" of this test file which uses `t.Helper()` for all the tests: [1](https://github.com/openshift/cluster-ingress-operator/blob/a41be6eed16361903e5b8ddd0131ac7585e87da3/test/e2e/gateway_api_test.go#L106-L107), [2](https://github.com/openshift/cluster-ingress-operator/blob/a41be6eed16361903e5b8ddd0131ac7585e87da3/test/e2e/gateway_api_test.go#L125-L126), etc.

Yeah, I noticed those "traditions" and fixed them here: https://github.com/openshift/cluster-ingress-operator/pull/1152/commits/748bf50e7688f64a5ce5c2692553bf1835a1de73

The reason I don't want them to have `t.Helper()` is that it makes diagnosing failures more difficult.  While working on #1152, I would get failures with line numbers pointing me to the helper's call site, and I needed to search for the failure message or trace through the code to find the interesting part of the test.

alebedev87:
`e2e-aws-operator-techpreview` is green, second time in a row.

test/e2e/gateway_api_test.go
alebedev87: Removed `t.Helper()` from subtests.

alebedev87:
`e2e-aws-operator-techpreview` is green, third time in a row.

alebedev87:

Another run, keep testing the timeout change.

test/e2e/gateway_api_test.go
```diff
@@ -273,6 +279,18 @@ func ensureCRDs(t *testing.T) {
 	}
 }
 
+// ensureExperimentalCRDs tests that experimental Gateway API custom resource definitions exist.
+func ensureExperimentalCRDs(t *testing.T) {
```
candita, line 283: This is identical to `ensureCRDs`, so why not just parameterize the crd names?

test/e2e/util_gatewayapi_test.go
```diff
@@ -271,6 +271,21 @@ func createGatewayClass(name, controllerName string) (*gatewayapiv1.GatewayClass
...
+func createCRD(name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	crd := buildGWAPICRDFromName(name)
+	if err := kclient.Create(context.Background(), crd); err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			if err := kclient.Get(context.Background(), types.NamespacedName{Name: name}, crd); err != nil {
+				return nil, fmt.Errorf("failed to get crd: %w", err)
+			}
+			return crd, nil
+		}
+		return nil, fmt.Errorf("failed to create crd: %w", err)
```
candita, line 284:
```suggestion
		return nil, fmt.Errorf("failed to create crd %s: %w", name, err)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -348,6 +363,9 @@ func buildGWAPICRDFromName(name string) *apiextensionsv1.CustomResourceDefinitio
 	case "referencegrants":
 		kind = "ReferenceGrant"
 		versions = []map[string]bool{{"v1beta1": true}}
+	case "listenersets":
```
candita, line 366: Since you are adding a new CRD anyway, can you add the GRPCRoute CRD as well?

test/e2e/util_gatewayapi_test.go
```diff
@@ -529,8 +547,8 @@ func assertGatewayClassSuccessful(t *testing.T, name string) (*gatewayapiv1.Gate
 	nsName := types.NamespacedName{Namespace: "", Name: name}
 	recordedConditionMsg := "not found"
 
-	// Wait up to 2 minutes for the gateway class to be Accepted.
-	err := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 2*time.Minute, false, func(context context.Context) (bool, error) {
+	// Wait up to 4 minutes for the gateway class to be Accepted.
```
candita, line 550: Any explanation for why creating everything takes so much longer?  Should we be waiting for something else to happen first?

candita: Is it at all possible that adding another clause to the CEL validation is causing increasing times?  If so, we wouldn't want to use a longer CEL validation.

test/e2e/gateway_api_test.go
```diff
@@ -273,6 +279,18 @@ func ensureCRDs(t *testing.T) {
...
+	t.Helper()
```
candita, line 284: You probable don't want the t.Helper() here either if we're getting rid of them to make it easier to debug/troubleshoot CI issues.

test/e2e/gateway_api_test.go
```diff
@@ -297,6 +315,26 @@ func deleteCRDs(t *testing.T) {
 	}
 }
 
+// createExperimentalCRDs creates experimental Gateway API custom resource definitions.
+func createExperimentalCRDs(t *testing.T) {
+	t.Helper()
```
candita, line 320:
```suggestion
```

test/e2e/gateway_api_test.go
```diff
@@ -273,6 +288,18 @@ func ensureCRDs(t *testing.T) {
...
 
+// ensureExperimentalCRDs tests that experimental Gateway API custom resource definitions exist.
+func ensureExperimentalCRDs(t *testing.T) {
+	t.Helper()
+	for _, crdName := range xcrdNames {
+		crdVersion, err := assertCrdExists(t, crdName)
+		if err != nil {
+			t.Fatalf("failed to find crd %s: %v", crdName, err)
+		}
+		t.Logf("found crd %s at version %s", crdName, crdVersion)
```
Miciah, line 290: "CRD" should be all upper-case, and  log messages should start with a capital letter (see https://github.com/openshift/cluster-ingress-operator/pull/1152#discussion_r1994834172).
```suggestion
			t.Fatalf("Failed to find CRD %s: %v", crdName, err)
		}
		t.Logf("Found CRD %s at version %s", crdName, crdVersion)
```
But that's a trivial issue, not a blocker for merging this PR.

test/e2e/gateway_api_test.go
```diff
@@ -109,6 +114,16 @@ func testGatewayAPIResources(t *testing.T) {
 	ensureCRDs(t)
 }
 
+// testExperimentalGatewayAPIResources verifies that the experimental Gateway API Custom Resource Definitions are available.
+func testExperimentalGatewayAPIResources(t *testing.T) {
```
Miciah, line 117: When reading this function definition, I immediately wondered, "Doesn't the ValidatingAdmissionPolicy prevent this?".  The function godoc should note that the function scales down CVO and removes the VAP in order to create an experimental CRD.  

Understanding that the function removes the VAP raises another question:  Why test this?  The point of the VAP is to block creating experimental CRDs, so I would expect a test to try to create an experimental CRD, expect to get an error, and fail if it succeeds in creating the experimental CRD.  What's the point of having an E2E test that verifies that the user can do an action that the VAP blocks when the VAP is removed?  

I would guess that you want to test that creating the experimental CRD in the *absence* of the VAP succeeds in order to establish a baseline, so that if you also test that creating an experimental CRD in the *presence* of the VAP fails, you can thereby conclude that the creating would succeed **but for** the VAP.  Is that the idea?  

I see that `testGatewayAPIResourcesProtection` verifies that creating a CRD (standard or experimental) fails when the VAP is present.  If the purpose of `testExperimentalGatewayAPIResources` is to establish a baseline for `testGatewayAPIResourcesProtection`, can you add comment to that effect?  Without some explanation, the purpose of `testExperimentalGatewayAPIResources` is unclear, and the function is very confusing.

test/e2e/util_gatewayapi_test.go
alebedev87: > Is it at all possible that adding another clause to the CEL validation is causing increasing times? If so, we wouldn't want to use a longer CEL validation.

Well, yes, longer CEL expressions may slow down authorization process. However, the VAP we use is relatively simple and should not harm the elapsed time of tests, at least not to the degree of dozens of seconds. Also, the failures I saw were related to the fact that gatewayclass or other objects were not accepted. Which suggests that the custom resource objects were created which in turn implies that CRD existed.

I suspect the install plan approvals to be the cause of the failures. Slack thread which discusses this: [link](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1742421080814679).

test/e2e/gateway_api_test.go
alebedev87: The idea was to adapt the case of the experimental CRDs to the existing `testGatewayAPIResourcesProtection` test. `testGatewayAPIResourcesProtection` assumes that GWAPI CRDs already exist on the cluster. This assumption is important to be able to test `UPDATE` verb (mentioned in the VAP). Because for the update case we have to be able to get the CRD from the API server. That's why `testExperimentalGatewayAPIResources` creates the experimental CRD bypassing the VAP, it tries to mock a situation when an experimental CRD exist so that `testGatewayAPIResourcesProtection`  will be able to check that the update against it is blocked.
Also, for the future, I think that `testExperimentalGatewayAPIResources` can be used to verify that all experimental CRDs are removed by the gatewayapi-controller (part of https://issues.redhat.com//browse/NE-1969).

alebedev87: > Also, for the future, I think that testExperimentalGatewayAPIResources can be used to verify that all experimental CRDs are removed by the gatewayapi-controller (part of https://issues.redhat.com//browse/NE-1969).
The future case can be addressed in the future. I think it's better to return [my previous suggestion about `ensureExperimentalCRDs` function](https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r1995139137) because it may be a little less confusing. Also, I added a comment to explain the purpose.

test/e2e/gateway_api_test.go
alebedev87: I returned to [the previous idea of using `ensureExperimentalCRDs`](https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r2005192321), so this function doesn't exist anymore.

test/e2e/util_gatewayapi_test.go
alebedev87: I returned to [the previous idea of using `ensureExperimentalCRDs`](https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r2005192321), so this function doesn't exist anymore.

test/e2e/gateway_api_test.go
alebedev87: I returned to [the previous idea of using `ensureExperimentalCRDs`](https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r2005192321), so this function doesn't exist anymore.

test/e2e/gateway_api_test.go
alebedev87: I returned to [the previous idea of using `ensureExperimentalCRDs`](https://github.com/openshift/cluster-ingress-operator/pull/1200#discussion_r2005192321), so this function doesn't exist anymore.

test/e2e/gateway_api_test.go
alebedev87: Done.

test/e2e/util_gatewayapi_test.go
alebedev87: Done.

manifests/01-validating-admission-policy.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     # Consider only request to Gateway API CRDs.
     - name: "check-only-gateway-api-crds"
       # When the operation is DELETE, the "object" variable is null.
-      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io' || (request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.x-k8s.io'"
```
candita, line 23: I think you would be better off looking at the CRD channel rather than the group.  Everything in https://github.com/kubernetes-sigs/gateway-api/blob/main/config/crd/experimental/ is experimental, but only 2 CRDs there are using the special group.   See https://github.com/kubernetes-sigs/gateway-api/blob/main/config/crd/experimental/gateway.networking.k8s.io_backendlbpolicies.yaml#L7 for an example of the channel annotation.

Miciah: The reason this PR adds `gateway.networking.x-k8s.io` to the VAP is so that it doesn't miss CRDs that belong to that group.  That is, we're not trying to discriminate between standard and experimental CRDs, and the goal isn't to check specifically for experimental Gateway API CRDs; rather, the goal is to make sure that when the VAP checks Gateway API CRDs, we include standard _as well as_ experimental CRDs when checking Gateway API CRDs.  

Hypothetically, if Gateway API defined yet another API group for Gateway API CRDs, we would need to add that group here as well.  For the purposes of this VAP, we don't care about the channel (standard, experimental or some yet-to-be-defined channel); we just want to make sure the VAP covers _all_ Gateway API CRDs.

candita:
Per @Miciah, it's okay to
/unhold

PR #1201, opened 2025-03-21
NE-1969: Ensure only managed Gateway API CRDs exist

alebedev87:
Delete all unmanaged Gateway API CRDs present in the cluster. An unmanaged Gateway API CRD is one with `gateway.networking.k8s.io` or `gateway.networking.x-k8s.io` in its `spec.group` field, and not managed by the gatewayapi controller.

### DIFF
--- a/pkg/operator/controller/gatewayapi/controller.go
+++ b/pkg/operator/controller/gatewayapi/controller.go
@@ -2,6 +2,7 @@ package gatewayapi
 
 import (
 	"context"
+	"fmt"
 	"sync"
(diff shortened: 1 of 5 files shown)


alebedev87:
/hold

The unmanaged Gateway API CRDs will be handled via a Degraded cluster operator. Holding to close until the PR which takes over is posted.

alebedev87:
Closing in favor of https://github.com/openshift/cluster-ingress-operator/pull/1205.

PR #1202, opened 2025-03-24
NE-1969: Add "v1" version to OpenShift GatewayClass controller name

alebedev87:
Add the "v1" suffix to the GatewayClass controllerName to allow for alternative OpenShift Gateway API implementations in the future.


Miciah:
The Gateway API feature is behind a featuregate.  

PR #1204, opened 2025-03-24
NE-1994: Add E2E test for Istio manual deployment

Miciah:
Add a new test to verify that Istio is configured not to allow manual deployment.

### DIFF
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/openshift/api/features"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
@@ -18,8 +19,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -96,9 +99,10 @@ func TestGatewayAPI(t *testing.T) {
 	t.Run("testGatewayAPIResources", testGatewayAPIResources)
 	if gatewayAPIControllerEnabled {
 		t.Run("testGatewayAPIObjects", testGatewayAPIObjects)
+		t.Run("testGatewayAPIManualDeployment", testGatewayAPIManualDeployment)
 		t.Run("testGatewayAPIIstioInstallation", testGatewayAPIIstioInstallation)
 	} else {
-		t.Log("Gateway API Controller not enabled, skipping testGatewayAPIObjects and testGatewayAPIIstioInstallation")
+		t.Log("Gateway API Controller not enabled, skipping controller tests")
 	}
 	t.Run("testGatewayAPIResourcesProtection", testGatewayAPIResourcesProtection)
 }
@@ -188,6 +192,128 @@ func testGatewayAPIObjects(t *testing.T) {
 		t.Errorf("failed to observe successful status of one or more gateway object/s: %v", strings.Join(errs, ","))
 	} else {
 		t.Log("gateway class, gateway, and http route created successfully")
+	}
+}
+
+// testGatewayAPIManualDeployment verifies that Istio's "manual deployment"
+// feature is not enabled (see
+// <https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment>).
+// We only want Istio to allow "automated deployment" (see
+// <https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment>).
+//
+// When manual deployment is enabled, then Istio allows a gateway to use an
+// existing service (for example, another gateway's service) by specifying that
+// service in spec.addresses.  When a gateway using manual deployment specifies
+// another gateway's service, the resulting behavior is effectively the same
+// behavior as Gateway API's concept of gateway listener "merging" (see
+// <https://github.com/kubernetes-sigs/gateway-api/blob/v1.2.1/apis/v1/gateway_types.go#L181-L182>).
+//
+// Gateway listener merging is underspecified in Gateway API and is not
+// consistently implemented among Gateway API implementations, and so we do not
+// want to allow it or any similar behavior (such as Istio's "manual
+// deployment") until such a time as it is well defined, standard behavior.
+// Instead, for the time being, we expect Istio to provision a service for a
+// gateway ("automated deployment"), even if the gateway specifies some existing
+// service in spec.addresses.
+func testGatewayAPIManualDeployment(t *testing.T) {
+	gatewayClass, err := createGatewayClass("openshift-default", "openshift.io/gateway-controller/v1")
+	if err != nil {
+		t.Fatalf("Failed to create gatewayclass: %v", err)
+	}
+
+	gatewayName := types.NamespacedName{
+		Name:      "manual-deployment",
+		Namespace: "openshift-ingress",
+	}
+	// Use the router's internal service in order to ensure that the
+	// referent exists.  Using an existing service isn't strictly necessary
+	// in order to verify that Istio does not use manual deployment; if
+	// manual deployment *is* enabled, Istio rejects the gateway if it
+	// points to a non-existent referent.  However, using an existing
+	// service more closely reflects the way that manual deployment *would*
+	// be used if it were allowed.
+	const existingServiceHostname = "router-internal-default.openshift-ingress.svc.cluster.local"
+	gateway := gatewayapiv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      gatewayName.Name,
+			Namespace: gatewayName.Namespace,
+		},
+		Spec: gatewayapiv1.GatewaySpec{
+			GatewayClassName: gatewayapiv1.ObjectName(gatewayClass.Name),
+			Addresses: []gatewayapiv1.GatewayAddress{{
+				Type:  ptr.To(gatewayapiv1.HostnameAddressType),
+				Value: existingServiceHostname,
+			}},
+			Listeners: []gatewayapiv1.Listener{{
+				Name:     "http",
+				Hostname: ptr.To(gatewayapiv1.Hostname(fmt.Sprintf("*.manual-deployment.%s", dnsConfig.Spec.BaseDomain))),
+				Port:     80,
+				Protocol: "HTTP",
+			}},
+		},
+	}
+	t.Logf("Creating gateway %q...", gatewayName)
+	if err := kclient.Create(context.Background(), &gateway); err != nil {
+		t.Fatalf("Failed to create gateway %v: %v", gatewayName, err)
+	}
+	t.Cleanup(func() {
+		if err := kclient.Delete(context.Background(), &gateway); err != nil {
+			if !errors.IsNotFound(err) {
+				t.Errorf("Failed to delete gateway %v: %v", gatewayName, err)
+			}
+		}
+	})
+
+	interval, timeout := 5*time.Second, 1*time.Minute
+	t.Logf("Polling for up to %v to verify that the gateway is accepted...", timeout)
+	if err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, gatewayName, &gateway); err != nil {
+			t.Logf("Failed to get gateway %v: %v; retrying...", gatewayName, err)
+
+			return false, nil
+		}
+
+		for _, condition := range gateway.Status.Conditions {
+			if condition.Type == string(gatewayapiv1.GatewayConditionAccepted) {
+				t.Logf("Found %q status condition: %+v", gatewayapiv1.GatewayConditionAccepted, condition)
+
+				if condition.Status == metav1.ConditionTrue {
+					return true, nil
+				}
+			}
+		}
+
+		t.Logf("Observed that gateway %v is not yet accepted; retrying...", gatewayName)
+
+		return false, nil
+	}); err != nil {
+		t.Errorf("Failed to observe the expected condition for gateway %v: %v", gatewayName, err)
+	}
+
+	serviceName := types.NamespacedName{
+		Name:      fmt.Sprintf("%s-%s", gateway.Name, gatewayClass.Name),
+		Namespace: gateway.Namespace,
+	}
+	var service corev1.Service
+	t.Logf("Polling for up to %v to verify that service %q is created...", timeout, serviceName)
+	if err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, serviceName, &service); err != nil {
+			t.Logf("Failed to get service %s: %v; retrying...", serviceName, err)
+
+			return false, nil
+		}
+
+		// Just verify that the service is created.  No need to verify
+		// that a load balancer is provisioned.  Indeed, provisioning
+		// will likely fail because Istio copies the address hostname to
+		// the service spec.loadBalancerIP field, which at least some
+		// cloud provider implementations reject.
+
+		t.Logf("Found service %q", serviceName)
+
+		return true, nil
+	}); err != nil {
+		t.Errorf("Failed to observe the expected condition for service %v: %v", serviceName, err)
 	}
 }
 


Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/567bd5ebd2a0733bfe57d0d19fd52f6b10104a73..1f0444b996dfff19d6118aadb7309bc8e1293fa8 is just a rebase.

Miciah:

Per TRT, "Feature gated code is free to use this label."

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
 	}
 }
 
+// testGatewayAPIManualDeployment verifies that Istio's manual deployment is not
+// enabled.  When manual deployment is enabled, then Istio allows a gateway use
+// to another gateway's service by specifying that gateway's service in
```
candita, line 200:
```suggestion
// enabled.  When manual deployment is enabled, then Istio allows a gateway use
// to another gateway's service by specifying that gateway's service in
nit:
```
```suggestion
// enabled.  When manual deployment is enabled, then Istio allows a gateway to use
// another gateway's service by specifying that gateway's service in
```

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+// spec.addresses.  This results in behavior similar to gateway listener
+// merging, which we do not want to allow at this time.  Instead, we expect
+// Istio to provision a service for this specific gateway, even if it specifies
+// spec.addresses.
+func testGatewayAPIManualDeployment(t *testing.T) {
+	gatewayClass, err := createGatewayClass("openshift-default", "openshift.io/gateway-controller")
```
candita, line 206:
```suggestion
	gatewayClass, err := createGatewayClass("openshift-default", "openshift.io/gateway-controller/v1")
```

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+				t.Errorf("Failed to delete gateway %v: %v", gatewayName, err)
+			}
+		}
+	})
+
+	interval, timeout := 5*time.Second, 1*time.Minute
+	t.Logf("Polling for up to %v to verify that the gateway is accepted...", timeout)
+	if err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, gatewayName, &gateway); err != nil {
+			t.Logf("Failed to get gateway %s: %v", gatewayName, err)
```
candita, line 250: nit, it can be helpful to see "retrying..." in the log:
```suggestion
			t.Logf("Failed to get gateway %s: %v, retrying...", gatewayName, err)
```

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+		for _, condition := range gateway.Status.Conditions {
+			if condition.Type == string(gatewayapiv1.GatewayConditionAccepted) {
+				t.Logf("Found %q status condition: %+v", gatewayapiv1.GatewayConditionAccepted, condition)
+
+				if condition.Status == metav1.ConditionTrue {
+					return true, nil
+				}
+			}
+		}
+
```
candita, line 264: nit:
```suggestion
t.Logf("Gateway %s not yet Accepted, retrying...", gatewayname)
```

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+		return false, nil
+	}); err != nil {
+		t.Errorf("Failed to observe the expected condition: %v", err)
```
candita, line 267:
```suggestion
		t.Errorf("Failed to find gateway %s at the expected condition: %v", gatewayname, err)
```

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+
+	serviceName := types.NamespacedName{
+		Name:      fmt.Sprintf("%s-%s", gateway.Name, gatewayClass.Name),
+		Namespace: gateway.Namespace,
+	}
+	var service corev1.Service
+	t.Logf("Polling for up to %v to verify that service %q is created...", timeout, serviceName)
+	if err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, serviceName, &service); err != nil {
+			t.Logf("Failed to get service %s: %v", serviceName, err)
```
candita, line 278:
```suggestion
			t.Logf("Failed to get service %s: %v, retrying...", serviceName, err)
```

test/e2e/gateway_api_test.go
candita: Very odd - I don't understand how the test works without the correct controllerName.

Miciah: I haven't rerun the tests since #1202 merged.

test/e2e/gateway_api_test.go
Miciah: That seems redundant; do we need to include the gateway name in _every_ log message in the test?  I can add it though.

test/e2e/gateway_api_test.go
Miciah: It seems redundant (and possibly inaccurate for the last iteration when we reach the timeout), but I can add it.

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,105 @@ func testGatewayAPIObjects(t *testing.T) {
...
+		// that a load balancer is provisioned.  Indeed, provisioning
+		// will likely fail because Istio copies the address hostname to
+		// the service spec.loadBalancerIP field, which at least some
+		// cloud provider implementations reject.
+
+		t.Logf("Found service %q", serviceName)
+
+		return true, nil
+	}); err != nil {
+		t.Errorf("Failed to observe the expected condition: %v", err)
```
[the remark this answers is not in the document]
Miciah: Isn't this extrapolating, and does it add any useful information?  I think it's better to keep the error message to what we actually observed and know.  The preceding log lines will provide the necessary context to understand why the expected condition was not observed.

test/e2e/gateway_api_test.go
Miciah: Thanks!  Fixed in https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b.

test/e2e/gateway_api_test.go
Miciah: Fixed in  https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b.

test/e2e/gateway_api_test.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b.

test/e2e/gateway_api_test.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b.

test/e2e/gateway_api_test.go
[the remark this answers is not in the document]
Miciah: I added the service name to the message in https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/1f0444b996dfff19d6118aadb7309bc8e1293fa8..6734479ebfc5bbb83897bc51f7d7ca580f783e2b rebases, updates the controller name per #1202, and updates some comments and strings to address review comments.

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,107 @@ func testGatewayAPIObjects(t *testing.T) {
 	}
 }
 
+// testGatewayAPIManualDeployment verifies that Istio's manual deployment is not
```
alebedev87, line 198: I think that I'm struggling to understand what "Istio manual deployment" means. Is it an installation of `servicemeshoperator3` OLM operator made by a third party (not CIO)?

candita: Now that you mention it, I think this test might not be complete.

I think of manual deployment as the opposite of auto-deployment.  Auto-deployment means that Istio must create a service (for a Gateway?).  Manual deployment means it must NOT create the service automatically.  Manual deployment is apparently configured by adding a service's address in the Gateway's spec.Addresses.  We don't want manual deployment, so here we're checking that it created the service automatically even though a service is specified in the spec.Addresses.

I think to be complete, the test needs to check that the created service is not the same service specified in the spec.Addresses.  But also, maybe the service that is added to spec.Addresses needs to really exist, and the test needs to make sure the service that is created is not the service added in spec.Addresses.

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,107 @@ func testGatewayAPIObjects(t *testing.T) {
...
+	gateway := gatewayapiv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      gatewayName.Name,
+			Namespace: gatewayName.Namespace,
+		},
+		Spec: gatewayapiv1.GatewaySpec{
+			GatewayClassName: gatewayapiv1.ObjectName(gatewayClass.Name),
+			Addresses: []gatewayapiv1.GatewayAddress{{
+				Type:  ptr.To(gatewayapiv1.HostnameAddressType),
+				Value: "lb.example.com",
```
candita, line 224: Following up on https://github.com/openshift/cluster-ingress-operator/pull/1204#discussion_r2017371544, shouldn't the spec.Address point to an existing service?  Otherwise, how do we know that the manual deployment didn't just fail because the service didn't exist?

test/e2e/gateway_api_test.go
```diff
@@ -191,6 +195,107 @@ func testGatewayAPIObjects(t *testing.T) {
...
+	}
+
+	serviceName := types.NamespacedName{
+		Name:      fmt.Sprintf("%s-%s", gateway.Name, gatewayClass.Name),
+		Namespace: gateway.Namespace,
+	}
+	var service corev1.Service
+	t.Logf("Polling for up to %v to verify that service %q is created...", timeout, serviceName)
+	if err := wait.PollUntilContextTimeout(context.Background(), interval, timeout, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, serviceName, &service); err != nil {
```
candita, line 279: Hm, how do we know it didn't create a service of another name?

test/e2e/gateway_api_test.go
alebedev87: > I think of manual deployment as the opposite of auto-deployment. Auto-deployment means that Istio must create a service (for a Gateway?). Manual deployment means it must NOT create the service automatically. Manual deployment is apparently configured by adding a service's address in the Gateway's spec.Addresses. We don't want manual deployment, so here we're checking that it created the service automatically even though a service is specified in the spec.Addresses.

Got it, thank you! The test starts to make sense for me now.

> I think to be complete, the test needs to check that the created service is not the same service specified in the spec.Addresses. But also, maybe the service that is added to spec.Addresses needs to really exist, and the test needs to make sure the service that is created is not the service added in spec.Addresses.

Yes, that may complete the test indeed. We can create a service and set its FQDN as `Gateway.Spec.Address` (type `Hostname`): [Istio doc's example using service FQDN](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment).

Miciah: > I think of manual deployment as the opposite of auto-deployment. Auto-deployment means that Istio must create a service (for a Gateway?). Manual deployment means it must NOT create the service automatically. Manual deployment is apparently configured by adding a service's address in the Gateway's spec.Addresses. We don't want manual deployment, so here we're checking that it created the service automatically even though a service is specified in the spec.Addresses.

Right.  If it helps, I can add a link to https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment in the code comment.  

> I think to be complete, the test needs to check that the created service is not the same service specified in the spec.Addresses. But also, maybe the service that is added to spec.Addresses needs to really exist, and the test needs to make sure the service that is created is not the service added in spec.Addresses.

I'm not sure I understand.  If a service got created, then that implies that Istio did automated deployment.  We know what service the test specifies, and it is not a `.svc.cluster.local` host name: https://github.com/openshift/cluster-ingress-operator/blob/6734479ebfc5bbb83897bc51f7d7ca580f783e2b/test/e2e/gateway_api_test.go#L222-L225

> Yes, that may complete the test indeed. We can create a service and set its FQDN as `Gateway.Spec.Address` (type `Hostname`): [Istio doc's example using service FQDN](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment).

I could change the test to specify `spec.addresses[].value: router-internal-default.openshift-ingress.svc.cluster.local`.  I had that during one iteration of the PR, thought it was unnecessary and potentially confusing, and changed the value to "lb.example.com".  I can change it back if that would be less confusing.

test/e2e/gateway_api_test.go
Miciah: > Following up on [#1204 (comment)](https://github.com/openshift/cluster-ingress-operator/pull/1204#discussion_r2017371544), shouldn't the spec.Address point to an existing service? Otherwise, how do we know that the manual deployment didn't just fail because the service didn't exist?

We know that because the gateway was accepted and Istio created a service for it.

test/e2e/gateway_api_test.go
Miciah: I suppose we could list all services and then look for unexpected services, but is there a reason why this test should do that?

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/6734479ebfc5bbb83897bc51f7d7ca580f783e2b..e2eb4f098027f87d7761bea11198f9aacfae41e1 elaborates on some comments to explain the concepts of Istio's "manual deployment" and "automated deployment" as well as Gateway API's gateway listener "merging".  It also changes the test gateway to point to the router's internal service so that the gateway's referent is an existing service, which more closely resembles how someone would use manual deployment in practice.

test/e2e/gateway_api_test.go
Miciah: References to "manual deployment", "automated deployment", and gateway listener "merging" added here: https://github.com/openshift/cluster-ingress-operator/blob/e2eb4f098027f87d7761bea11198f9aacfae41e1/test/e2e/gateway_api_test.go#L198-L217

Changed to use an existing service here: https://github.com/openshift/cluster-ingress-operator/blob/e2eb4f098027f87d7761bea11198f9aacfae41e1/test/e2e/gateway_api_test.go#L228-L234

I struggled with describing why the test needs to use an existing service.  I _think_ what I wrote will make sense to future me.  Let me know what you think!

candita:
Cluster install failure:

> CatalogdClusterCatalogOpenshiftRedhatMarketplaceDegraded: Internal error occurred: failed calling webhook "inject-metadata-name.olm.operatorframework.io": failed to call webhook: Post "https://catalogd-service.openshift-catalogd.svc:9443/mutate-olm-operatorframework-io-v1-clustercatalog?timeout=10s": no endpoints available for service "catalogd-service"
> CatalogdClusterCatalogOpenshiftRedhatOperatorsDegraded: Internal error occurred: failed calling webhook "inject-metadata-name.olm.operatorframework.io": failed to call webhook: Post "https://catalogd-service.openshift-catalogd.svc:9443/mutate-olm-operatorframework-io-v1-clustercatalog?timeout=10s": no endpoints available for service "catalogd-service"
> 

/test e2e-aws-operator

candita:
Multus CNI error for single node:

> Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_simpletest-rc-to-be-deleted-wskfn_e2e-gc-5189_bda753e8-edc3-471b-a1cc-dd404222fc05_0(7ebb5667a9c8c1e09418ea9377b06a50cf153a5042fd56930080bef4fd6e7aae): error adding pod e2e-gc-5189_simpletest-rc-to-be-deleted-wskfn to CNI network "multus-cni-network": plugin type="multus-shim" name="multus-cni-network" failed (add): CmdAdd (shim): CNI request failed with status 400: ...
> ERRORED: error configuring pod [e2e-gc-5189/simpletest-rc-to-be-deleted-wskfn] networking: Multus: [e2e-gc-5189/simpletest-rc-to-be-deleted-wskfn/bda753e8-edc3-471b-a1cc-dd404222fc05]: error waiting for pod: Get "https://api-int.ci-op-cw14gdlz-b5a36.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/e2e-gc-5189/pods/simpletest-rc-to-be-deleted-wskfn?timeout=1m0s": context deadline exceeded

/test e2e-aws-ovn-single-node

PR #1206, opened 2025-03-27
NE-1912: Add aggregated ClusterRoles for Gateway API

shaneutt:
This adds aggregated ClusterRoles for Gateway API as per the "Security Policy" in the Gateway API EP.

### DIFF
--- a/pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
+++ b/pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
@@ -1,0 +1,30 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: system:openshift:gateway-api:aggregate-to-admin
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
+  - apiGroups:
+      - gateway.networking.k8s.io
+    resources:
+      - gateways
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - gateway.networking.k8s.io
+    resources:
+      - httproutes
+      - grpcroutes
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
+      - deletecollection
--- a/pkg/operator/controller/gatewayapi/rbac.go
+++ b/pkg/operator/controller/gatewayapi/rbac.go
@@ -1,0 +1,123 @@
+package gatewayapi
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+var managedClusterRoles = []*rbacv1.ClusterRole{
+	manifests.GatewayAPIAdminClusterRole(),
+	manifests.GatewayAPIViewClusterRole(),
+}
+
+// ensureGatewayAPIRBAC ensures that all RBAC resources for Gateway API are
+// deployed to the cluster, and up to date.
+func (r *reconciler) ensureGatewayAPIRBAC(ctx context.Context) error {
+	var errs []error
+
+	for _, clusterRole := range managedClusterRoles {
+		if err := r.ensureClusterRole(ctx, clusterRole); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// ensureClusterRole ensures that a specific ClusterRole RBAC resource is
+// deployed to the cluster, and up to date.
+func (r *reconciler) ensureClusterRole(ctx context.Context, desired *rbacv1.ClusterRole) error {
+	clusterRoleAlreadyExists, current, err := r.currentClusterRole(ctx, desired.Name)
+	if err != nil {
+		return err
+	}
+
+	if clusterRoleAlreadyExists {
+		if updated, err := r.updateClusterRole(ctx, current, desired); err != nil {
+			return err
+		} else if updated {
+			if _, _, err := r.currentClusterRole(ctx, desired.Name); err != nil {
+				return err
+			}
+		}
+	} else {
+		return r.createClusterRole(ctx, desired)
+	}
+
+	return nil
+}
+
+// currentClusterRole retrieves a ClusterRole from the cluster, given the name
+// of that role.
+func (r *reconciler) currentClusterRole(ctx context.Context, name string) (bool, *rbacv1.ClusterRole, error) {
+	var existingClusterRole rbacv1.ClusterRole
+
+	if err := r.client.Get(ctx, types.NamespacedName{Name: name}, &existingClusterRole); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil, nil
+		}
+		return false, nil, fmt.Errorf("failed to get ClusterRole %s: %w", name, err)
+	}
+
+	return true, &existingClusterRole, nil
+}
+
+// createClusterRole creates the desired ClusterRole on the cluster.
+func (r *reconciler) createClusterRole(ctx context.Context, desired *rbacv1.ClusterRole) error {
+	desired.ResourceVersion = ""
+
+	if err := r.client.Create(ctx, desired); err != nil {
+		return fmt.Errorf("failed to create ClusterRole %s: %w", desired.Name, err)
+	}
+
+	log.Info("Created ClusterRole", "name", desired.Name)
+
+	return nil
+}
+
+// updateClusterRole verifies whether the current ClusterRole is up to date with
+// what is desired, and performs an update if not.
+func (r *reconciler) updateClusterRole(ctx context.Context, current, desired *rbacv1.ClusterRole) (bool, error) {
+	changed, updated := clusterRoleChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+	if err := r.client.Update(ctx, updated); err != nil {
+		return false, fmt.Errorf("failed to update ClusterRole %s: %w", updated.Name, err)
+	}
+
+	log.Info("updated ClusterRole", "name", updated.Name, "diff", diff)
+
+	return true, nil
+}
+
+// clusterRoleChanged indicates whether there are changes between the current
+// ClusterRole and what is expected, and provides the updated version of the
+// ClusterRole if so.
+func clusterRoleChanged(current, expected *rbacv1.ClusterRole) (bool, *rbacv1.ClusterRole) {
+	if cmp.Equal(current.Rules, expected.Rules) &&
+		cmp.Equal(current.ObjectMeta.Labels, expected.ObjectMeta.Labels) &&
+		cmp.Equal(current.ObjectMeta.Annotations, expected.ObjectMeta.Annotations) {
+		return false, nil
+	}
+
+	updated := current.DeepCopy()
+	updated.ObjectMeta.Labels = expected.ObjectMeta.Labels
+	updated.ObjectMeta.Annotations = expected.ObjectMeta.Annotations
+	updated.Rules = expected.Rules
+
+	return true, updated
+}
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -105,6 +105,8 @@ func TestGatewayAPI(t *testing.T) {
 		t.Log("Gateway API Controller not enabled, skipping controller tests")
 	}
 	t.Run("testGatewayAPIResourcesProtection", testGatewayAPIResourcesProtection)
+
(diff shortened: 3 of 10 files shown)


pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,127 @@
...
+		cmp.Equal(current.ObjectMeta.Labels, expected.ObjectMeta.Labels) &&
+		cmp.Equal(current.ObjectMeta.Annotations, expected.ObjectMeta.Annotations) &&
+		cmp.Equal(current.ObjectMeta.Finalizers, expected.ObjectMeta.Finalizers) {
+		return false, nil
+	}
+
+	updated := current.DeepCopy()
+	updated.ObjectMeta.Labels = expected.ObjectMeta.Labels
+	updated.ObjectMeta.Annotations = expected.ObjectMeta.Annotations
+	updated.Rules = expected.Rules
```
Miciah, line 124: Do we really care about finalizers?  If so, this function needs to update `updated.ObjectMeta.Finalizers`:
```suggestion
	if cmp.Equal(current.Rules, expected.Rules) &&
		cmp.Equal(current.ObjectMeta.Labels, expected.ObjectMeta.Labels) &&
		cmp.Equal(current.ObjectMeta.Annotations, expected.ObjectMeta.Annotations) &&
		cmp.Equal(current.ObjectMeta.Finalizers, expected.ObjectMeta.Finalizers) {
		return false, nil
	}

	updated := current.DeepCopy()
	updated.ObjectMeta.Labels = expected.ObjectMeta.Labels
	updated.ObjectMeta.Annotations = expected.ObjectMeta.Annotations
	updated.Rules = expected.Rules
	updated.ObjectMeta.Finalizers = expected.ObjectMeta.Finalizers
```
And the it might also be appropriate to check the deletion timestamp to make sure we don't try to add finalizers to an object that is marked for deletion.

test/e2e/util_gatewayapi_test.go
```diff
@@ -941,3 +944,52 @@ func (m *vapManager) enable() {
...
+func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRoleName, srcClusterRoleName string) error {
+	t.Helper()
+
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		var destClusterRole rbacv1.ClusterRole
+		if err := kclient.Get(context.Background(), types.NamespacedName{Name: destClusterRoleName}, &destClusterRole); err != nil {
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
```
Miciah, line 957: We usually retry on all errors, and in this case, I think it is notable and worth logging if you get a not-found error for any of the standard clusterroles.  
```suggestion
			t.Logf("Failed to get destination ClusterRole %s; retrying...: %v", err)
			return false, nil
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -941,3 +944,52 @@ func (m *vapManager) enable() {
...
+		}
+
+		var srcClusterRole rbacv1.ClusterRole
+		if err := kclient.Get(context.Background(), types.NamespacedName{Name: srcClusterRoleName}, &srcClusterRole); err != nil {
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
```
Miciah, line 965:
```suggestion
			t.Logf("Failed to get source ClusterRole %s: %v; retrying...", err)
			return false, nil
```

pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
```diff
@@ -0,0 +1,28 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: system:gateway-api:aggregate-to-admin
```
Miciah, line 4: This is a fairly generic, non-OpenShift-specific name for a cluster-scoped resource.  What happens if the cluster-admin is already using this name for a clusterrole?

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,127 @@
...
+				return err
+			}
+		}
+	} else {
+		if err := r.createClusterRole(ctx, desired); err != nil {
+			return err
+		}
+		if _, _, err := r.currentClusterRole(ctx, desired.Name); err != nil {
+			return err
+		}
```
Miciah, line 59: Is it really useful to get the resource again after doing a create or an update?  Some of our `ensureFoo` functions to this so that they can return the created or updated object, but otherwise it seems unnecessary.

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,127 @@
...
+	return true, &existingClusterRole, nil
+}
+
+// createClusterRole creates the desired ClusterRole on the cluster.
+func (r *reconciler) createClusterRole(ctx context.Context, desired *rbacv1.ClusterRole) error {
+	if err := r.client.Create(ctx, desired); err != nil {
+		return fmt.Errorf("failed to create ClusterRole %s: %w", desired.Name, err)
+	}
+
+	log.Info("created ClusterRole", "name", desired.Name)
```
Miciah, line 86: I recently realized that the Kubernetes guidelines specify that every log message should start with a capital letter: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#message-style-guidelines
```suggestion
	log.Info("Created ClusterRole", "name", desired.Name)
```
Same applies to the `log.Info` in `updateClusterRole` and the instances of `t.Logf` in `test/e2e/gateway_api_test.go` in the last commit.

pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
shaneutt: Yeah, this is just draft cruft I was looking for a better name I think I found one based on looking at other ones other operators used :+1:

pkg/operator/controller/gatewayapi/rbac.go
shaneutt: Yeah, draft cruft :+1:

pkg/operator/controller/gatewayapi/rbac.go
shaneutt: Draft cruft, hadn't self-reviewed yet and we don't need to worry about finalizers :+1: 

```suggestion
	if cmp.Equal(current.Rules, expected.Rules) &&
		cmp.Equal(current.ObjectMeta.Labels, expected.ObjectMeta.Labels) &&
		cmp.Equal(current.ObjectMeta.Annotations, expected.ObjectMeta.Annotations) {
		return false, nil
	}

	updated := current.DeepCopy()
	updated.ObjectMeta.Labels = expected.ObjectMeta.Labels
	updated.ObjectMeta.Annotations = expected.ObjectMeta.Annotations
	updated.Rules = expected.Rules
```

candita:
Test failure in e2e-aws-operator-techpreview is unusual.  The dnsrecord isn't even created, even while trying to get it for one minute.
 
> util_gatewayapi_test.go:854: Failed to get DNSRecord openshift-ingress/test-gateway-f47b7d78-wildcard: dnsrecords.ingress.operator.openshift.io "test-gateway-f47b7d78-wildcard" not found; retrying...

knobunc:
/approve

Looks good to me too, but I'll wait for a second review

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,123 @@
...
+
+// ensureGatewayAPIRBAC ensures that all RBAC resources for Gateway API are
+// deployed to the cluster, and up to date.
+func (r *reconciler) ensureGatewayAPIRBAC(ctx context.Context) error {
+	var errs []error
+
+	for _, clusterRole := range managedClusterRoles {
+		if err := r.ensureClusterRole(ctx, clusterRole); err != nil {
+			errs = append(errs, err)
+		}
```
Miciah, line 31: `NewAggregate` ignores nil values.  
```suggestion
		errs = append(errs, r.ensureClusterRole(ctx, clusterRole))
```

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+
+	if clusterRoleAlreadyExists {
+		if updated, err := r.updateClusterRole(ctx, current, desired); err != nil {
+			return err
+		} else if updated {
+			if _, _, err := r.currentClusterRole(ctx, desired.Name); err != nil {
+				return err
+			}
```
Miciah, line 51: Is this get after update needed?  Related comment from my earlier review: https://github.com/openshift/cluster-ingress-operator/pull/1206#discussion_r2017755806

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		return false, nil, fmt.Errorf("failed to get ClusterRole %s: %w", name, err)
+	}
+
+	return true, &existingClusterRole, nil
+}
+
+// createClusterRole creates the desired ClusterRole on the cluster.
+func (r *reconciler) createClusterRole(ctx context.Context, desired *rbacv1.ClusterRole) error {
+	desired.ResourceVersion = ""
```
Miciah, line 77: Why is this needed?  The object came from the manifest, so it won't have `revisionVersion` set anyway, right?

pkg/operator/controller/gatewayapi/rbac.go
```diff
@@ -0,0 +1,123 @@
...
+		return false, nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+	if err := r.client.Update(ctx, updated); err != nil {
+		return false, fmt.Errorf("failed to update ClusterRole %s: %w", updated.Name, err)
+	}
+
+	log.Info("updated ClusterRole", "name", updated.Name, "diff", diff)
```
Miciah, line 102:
```suggestion
	log.Info("Updated ClusterRole", "name", updated.Name, "diff", diff)
```

test/e2e/gateway_api_test.go
```diff
@@ -402,6 +404,25 @@ func testGatewayAPIResourcesProtection(t *testing.T) {
...
+// aggregated ClusterRoles) are properly deployed and aggregated.
+func testGatewayAPIRBAC(t *testing.T) {
+	aggregationMapping := map[string][]string{
+		"system:openshift:gateway-api:aggregate-to-admin": {"admin", "edit"},
+		"system:openshift:gateway-api:aggregate-to-view":  {"view"},
+	}
+
+	for srcClusterRoleName, destClusterRoleNames := range aggregationMapping {
+		for _, destClusterRoleName := range destClusterRoleNames {
+			t.Logf("verifying that ClusterRole %s aggregates all PolicyRules from %s", destClusterRoleName, srcClusterRoleName)
```
Miciah, line 417:
```suggestion
			t.Logf("Verifying that ClusterRole %s aggregates all PolicyRules from %s", destClusterRoleName, srcClusterRoleName)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -947,3 +950,48 @@ func (m *vapManager) enable() {
...
 	}
 }
+
+func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRoleName, srcClusterRoleName string) error {
+	t.Helper()
+
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		var destClusterRole rbacv1.ClusterRole
+		if err := kclient.Get(context.Background(), types.NamespacedName{Name: destClusterRoleName}, &destClusterRole); err != nil {
+			t.Logf("Failed to get destination ClusterRole %s; retrying...: %v", destClusterRoleName, err)
```
Miciah, line 960:
```suggestion
			t.Logf("Failed to get destination ClusterRole %s: %v; retrying...", destClusterRoleName, err)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -947,3 +950,48 @@ func (m *vapManager) enable() {
 		m.t.Errorf("failed to find vap %q: %v", m.name, err)
 	}
 }
+
+func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRoleName, srcClusterRoleName string) error {
+	t.Helper()
+
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
```
Miciah, line 957: Yeesh, we have multiple tests using this `timeout` const, which is defined here: https://github.com/openshift/cluster-ingress-operator/blob/e60422dca3964c4f00b7cac64944d5ec2a874baa/test/e2e/configurable_route_test.go#L27

So we have multiple unrelated tests using the same generically named const, which is defined in an arbitrary location, with an arbitrary value, with no godoc.  We should clean that up at some point.  

I don't mean to pick on this PR though.  Evidently we've been cargo-culting this `timeout` for some time.  At some point, we should clean all these uses of `timeout` up.

Miciah:
Oh, does "merge-method-rebase" mean we get no merge commit?

pkg/operator/controller/gatewayapi/rbac.go
shaneutt: Not needed: a leftover from copying the general structure of the CRD logic.

shaneutt:
> Oh, does "merge-method-rebase" mean we get no merge commit?

Right, it will apply the commits on top as individual commits without a merge commit. It didn't occur to me this would be an issue, if you prefer merge commits I'll keep this in mind for any of my future contributions.

---

As for your other comments: it's merged now (since we're on a tight schedule), but _if_ there happened to be any spare time for some refactoring I was thinking it might be good to look at places where things can be made more generic between the CRD and RBAC logic, given the similarities. That might be worth spending a few cycles after all the required deliverables are in.

PR #1207, opened 2025-03-27
Gw detector

aslakknutsen:


pkg/operator/controller/status/controller.go
```diff
@@ -554,6 +561,23 @@ func computeOperatorAvailableCondition(ingresses []operatorv1.IngressController)
...
+		degradedCondition.Message = fmt.Sprintf("Unsupportd objects found: %s", status.Details)
+	} else {
+		degradedCondition.Status = configv1.ConditionFalse
+		degradedCondition.Reason = "SupportNotDegraded"
+		degradedCondition.Message = fmt.Sprintf("No unsupported usage found.")
+
+	}
+	return degradedCondition
+}
+
```
aslakknutsen, line 580: @Miciah @shaneutt Still "a few" rough edges here, but does this specific part look remotely like you had in mind? :)

(wanted to run the thing so the best idea at the time was to create a PR and launch with clusterbot. hehe)

vendor/github.com/openshift/api/config/v1/types_cluster_operator.go
```diff
@@ -183,6 +183,8 @@ const (
 	// expectation is that all unexpected errors are handled as operators mature.
 	OperatorDegraded ClusterStatusConditionType = "Degraded"
 
+	OperatorDegradedSupport ClusterStatusConditionType = "DegradedSupport"
+
```
aslakknutsen, line 187: obs, didn't notice this was in vendor :)

pkg/operator/controller/status/controller.go
Miciah: This is close.  However, I expected `computeOperatorDegradedCondition` to call `computeExternalStatus`.  The new input from the detectors would be aggregated with the input that `computeOperatorDegradedCondition` currently uses in order to decide ultimately whether to set `Degraded=False` or `Degraded=True`.  

`computeOperatorDegradedCondition` might need some refactoring; ultimately I expect it to call several functions that give it the inputs it needs to compute the overall `Degraded` status condition.  For an example of what I have in mind, see the `computeIngressUpgradeableCondition` function: https://github.com/openshift/cluster-ingress-operator/blob/288f96cf3ae11f539efc3ab4bbf805a6d06592ef/pkg/operator/controller/ingress/status.go#L648-L673

Adding a new status condition in the clusteroperator might raise some eyebrows, and it wouldn't raise the alarm the way that setting `Degraded=False` does.

aslakknutsen: Right, I was wondering about that; should a "support issue" declare the whole thing Degraded/Borked. Felt a bit harsh. :)

manifests/00-cluster-role.yaml
```diff
@@ -250,3 +250,29 @@ rules:
   verbs:
   - list
   - watch
+
+# Required for watching and detecting unsupported usage of the Gateway istio control plane
```
shaneutt, line 254: :+1:

pkg/operator/controller/detector/istio/detector_istio.go
```diff
@@ -0,0 +1,66 @@
...
+	{APIVersion: "security.istio.io/v1", Kind: "PeerAuthentication"},
+	{APIVersion: "security.istio.io/v1", Kind: "RequestAuthentication"},
+	{APIVersion: "telemetry.istio.io/v1", Kind: "Telemetry"},
+}
+
+func SetupDetectors(mgr manager.Manager, statusReporter detector.StatusReporter, mappings map[metav1.GroupKind][]crd.ControllerFunc) {
+
+	ignoreOwnerRefs := []metav1.OwnerReference{}
+	ignoreLabels := map[string]string{
+		"kuadrant.io/managed": "true", // kuadrant is allowed to configure our control plane
```
shaneutt, line 34: Is this surefire? cc @guicassolato

PR #1203, opened 2025-03-24
OCPBUGS-53424: Wait for install plans to enter the "Requires Approval" phase before approving them

rfredette:

PR #1212, opened 2025-04-03
 OCPBUGS-54966: Improve detection of missing DNSRecord for Gateway

alebedev87:
This PR aims at helping with gateway API e2e test failures when `DNSRecord` for a gateway cannot be found in time( [example](https://github.com/openshift/cluster-ingress-operator/pull/1203#issuecomment-2775521838)):
- Adds more logs to the logic of DNSRecord ensuring to see when DNSRecord is updated or left untouched.
- Adds more logs to the logic of generation of desired DNSRecord to see why Gateway service is not taken.
- Adds more logs to Gateway service enqueue logic to see when reconciling hits or does not hit.
- Ensures the `Programmed` condition is checked on the Gateway resource in the e2e test.
- Logs gateway service definition while polling for gateway acceptance in the e2e test.
- Logs the reason why DNSRecord is not ready during the assertion.

### DIFF
--- a/pkg/operator/controller/gateway-service-dns/controller.go
+++ b/pkg/operator/controller/gateway-service-dns/controller.go
@@ -78,7 +78,11 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 			// A DNSRecord CR needs to be updated if, and only if,
 			// the hostname has changed (a listener's port and
 			// protocol have no bearing on the DNS record).
-			return gatewayListenersHostnamesChanged(old, new)
+			changed := gatewayListenersHostnamesChanged(old, new)
+			if changed {
+				log.Info("Listener hostname changed", "gateway", e.ObjectNew.(*gatewayapiv1.Gateway).Name)
+			}
+			return changed
 		},
 	}
 	isInOperandNamespace := predicate.NewPredicateFuncs(func(o client.Object) bool {
@@ -103,6 +107,7 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 				},
 			}
 			requests = append(requests, request)
+			log.Info("Enqueuing service for gateway", "gateway", o.GetName(), "service", request.NamespacedName)
 		}
 		return requests
 	}
--- a/pkg/resources/dnsrecord/dns.go
+++ b/pkg/resources/dnsrecord/dns.go
@@ -77,15 +77,17 @@ func EnsureDNSRecord(client client.Client, name types.NamespacedName, dnsRecordL
 		if err := client.Create(context.TODO(), desired); err != nil {
 			return false, nil, fmt.Errorf("failed to create dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
 		}
-		log.Info("created dnsrecord", "dnsrecord", desired)
+		log.Info("Created dnsrecord", "dnsrecord", desired)
 		return CurrentDNSRecord(client, name)
 	case wantWC && haveWC:
 		if updated, err := updateDNSRecord(client, current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
 		} else if updated {
+			log.Info("Updated dnsrecord", "dnsrecord", desired)
 			return CurrentDNSRecord(client, name)
 		}
 	}
+	log.Info("No desired dnsrecord", "service", service)
 
 	return haveWC, current, nil
 }
@@ -130,6 +132,7 @@ func desiredWildcardDNSRecord(name types.NamespacedName, dnsRecordLabels map[str
 func desiredDNSRecord(name types.NamespacedName, dnsRecordLabels map[string]string, ownerRef metav1.OwnerReference, domain string, dnsPolicy iov1.DNSManagementPolicy, service *corev1.Service) (bool, *iov1.DNSRecord) {
 	// No LB target exists for the domain record to point at.
 	if len(service.Status.LoadBalancer.Ingress) == 0 {
+		log.Info("No load balancer target for dnsrecord", "dnsrecord", name, "service", service.Name)
 		return false, nil
 	}
 
@@ -138,6 +141,7 @@ func desiredDNSRecord(name types.NamespacedName, dnsRecordLabels map[string]stri
 	// Quick sanity check since we don't know how to handle both being set (is
 	// that even a valid state?)
 	if len(ingress.Hostname) > 0 && len(ingress.IP) > 0 {
+		log.Error(nil, "Both load balancer hostname and IP are set", "dnsrecord", name, "service", service.Name)
 		return false, nil
 	}
 
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -594,10 +594,11 @@ func assertGatewayClassSuccessful(t *testing.T, name string) (*gatewayapiv1.Gate
 		return false, nil
 	})
 	if err != nil {
+		t.Logf("[%s] Last observed gatewayclass:\n%s", time.Now().Format(time.DateTime), util.ToYaml(gwc))
 		return nil, fmt.Errorf("gatewayclass %s is not %v; last recorded status message: %s", name, gatewayapiv1.GatewayClassConditionStatusAccepted, recordedConditionMsg)
 	}
 
-	t.Logf("Observed that gatewayclass %s has been accepted: %+v", name, gwc.Status)
+	t.Logf("[%s] Observed that gatewayclass %s has been accepted: %+v", time.Now().Format(time.DateTime), name, gwc.Status)
 
 	return gwc, nil
 }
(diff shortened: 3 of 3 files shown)


alebedev87:
The gateway service didn't get its load balancer target in time (last transition: `17:27:57`):
```
          - type: Programmed
            status: "False"
            observedgeneration: 1
            lasttransitiontime: "2025-04-04T17:27:57Z"
            reason: AddressNotAssigned
            message: 'Assigned to service(s) test-gateway-openshift-default.openshift-ingress.svc.cluster.local:80,
              but failed to assign to all requested addresses: address pending for hostname
              "test-gateway-openshift-default.openshift-ingress.svc.cluster.local"'
```

Load balancer provisioning happened at `17:29:28`:
```
- apiVersion: v1                                                                                                                                     
  count: 1
  eventTime: null
  firstTimestamp: "2025-04-04T17:29:28Z"
  involvedObject:
    apiVersion: v1
    kind: Service
    name: test-gateway-openshift-default
    namespace: openshift-ingress
    resourceVersion: "95293"
    uid: 33f249bf-a0c3-4350-b3cc-70e4467c44a6
  kind: Event
  lastTimestamp: "2025-04-04T17:29:28Z"
  message: Ensured load balancer
```

The istiod created the service at `17:29:27`:
```
2025-04-04T17:29:27.435940790Z 2025-04-04T17:29:27.435926Z	info	model	Full push, new service openshift-ingress/test-gateway-openshift-default.openshift-ingress.svc.cluster.local
```

And itself, the istiod pod was created at  `17:29:18`:
```yaml
  creationTimestamp: "2025-04-04T17:29:18Z"
  generateName: istiod-openshift-gateway-57c5c5ccb9-
```

Who accepted the gatewayclass then?! [The check for its acceptance goes before the gateway check](https://github.com/openshift/cluster-ingress-operator/blob/6fe4db1fb589aabec6ce85c34180971ebf1e7089/test/e2e/gateway_api_test.go#L520-L526).

alebedev87:
Running `testGatewayAPIIstioInstallation` before `testGatewayAPIObjects` may help to avoid problems with LB provisioning.

pkg/resources/dnsrecord/dns.go
```diff
@@ -138,6 +141,7 @@ func desiredDNSRecord(name types.NamespacedName, dnsRecordLabels map[string]stri
 	// Quick sanity check since we don't know how to handle both being set (is
 	// that even a valid state?)
 	if len(ingress.Hostname) > 0 && len(ingress.IP) > 0 {
+		log.Info("no load balancer hostname or IP for dnsrecord", "dnsrecord", name, "service", service.Name)
```
candita, line 144: In this case it has **both** hostname and ip, which is unexpected.

test/e2e/util_gatewayapi_test.go
```diff
@@ -562,27 +563,42 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
 
 	gw := &gatewayapiv1.Gateway{}
 	nsName := types.NamespacedName{Namespace: namespace, Name: name}
-	recordedConditionMsg := "not found"
+	recordedAcceptedConditionMsg, recordedProgrammedConditionMsg := "not found", "not found"
```
Miciah, line 566: With your change below to use `%q` instead of `%s`, it would be clearer to initialize these variables to empty strings.

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -78,7 +78,11 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 			// A DNSRecord CR needs to be updated if, and only if,
 			// the hostname has changed (a listener's port and
 			// protocol have no bearing on the DNS record).
-			return gatewayListenersHostnamesChanged(old, new)
+			changed := gatewayListenersHostnamesChanged(old, new)
+			if changed {
+				log.Info("listener hostname changed", "gateway", e.ObjectNew.(*gatewayapiv1.Gateway).Name)
```
Miciah, line 83: Here and in the other log messages that you add, please capitalize the first letter.  See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#message-style-guidelines.

Miciah:
In some test failures that I have been investigating, there are no events to indicate that the service controller tried to provision an LB, which leads me to wonder whether Istio sometimes doesn't create the service or doesn't configure it correctly.  The istiod logs aren't very helpful for diagnosing these failures.  For this reason, I believe it could be helpful for `ensureGatewayObjectSuccess` to get and log the service definition.  Would you be interested in adding that to this PR?

alebedev87:
`e2e-aws-operator-techpreview` failed on `testGatewayAPIDNSListenerUpdate` most likely because [CCM was frozen for some time (~3 minutes) to acquire the lease](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1212/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview/1911809382308384768/artifacts/e2e-aws-operator-techpreview/gather-extra/artifacts/pods/openshift-cloud-controller-manager-operator_cluster-cloud-controller-manager-operator-cc466d9b5-4lpwt_cluster-cloud-controller-manager.log) which slowed down the LB deprovisioning for the gateway service.

https://github.com/openshift/cluster-ingress-operator/pull/1220 aims at helping signal such problems.

test/e2e/util_gatewayapi_test.go
```diff
@@ -1037,21 +1066,21 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
...
 					if condition.Type == v1.DNSRecordPublishedConditionType && condition.Status == string(metav1.ConditionTrue) {
-						t.Logf("Found DNSRecord %v %s=%s", recordName, condition.Type, condition.Status)
+						t.Logf("[%s] Found DNSRecord %v %s=%s", time.Now().Format(time.DateTime), recordName, condition.Type, condition.Status)
 						return true, nil
 					}
 				}
 			}
 		}
-		t.Logf("Found DNSRecord %v but could not determine its readiness; retrying...", recordName)
+		t.Logf("[%s] Found DNSRecord %v but could not determine its readiness; retrying...", time.Now().Format(time.DateTime), recordName)
```
lihongan, line 1083: Can we have a clear log when `len(dnsRecord.Status.Zones) == 0` ?
For example on OpenStack, the `Status` is empty even it has `Spec`.
```
$ oc -n openshift-ingress get dnsrecords gateway-586fb57b47-wildcard -oyaml | yq .spec
dnsManagementPolicy: Managed
dnsName: '*.gw.d5p8jv2-b5564.shiftstack.devcluster.openshift.com.'
recordTTL: 30
recordType: A
targets:
  - 38.102.83.6

$ oc -n openshift-ingress get dnsrecords gateway-586fb57b47-wildcard -oyaml | yq .status
null
```

alebedev87: Done.

rfredette:
Other than the one nit, I'm happy with this.

/approve

will leave lgtm to @candita

pkg/resources/dnsrecord/dns.go
```diff
@@ -77,15 +77,17 @@ func EnsureDNSRecord(client client.Client, name types.NamespacedName, dnsRecordL
...
 		return CurrentDNSRecord(client, name)
 	case wantWC && haveWC:
 		if updated, err := updateDNSRecord(client, current, desired); err != nil {
 			return true, current, fmt.Errorf("failed to update dnsrecord %s/%s: %v", desired.Namespace, desired.Name, err)
 		} else if updated {
+			log.Info("Updated dnsrecord", "dnsrecord", desired)
 			return CurrentDNSRecord(client, name)
 		}
 	}
+	log.Info("No desired dnsrecord")
```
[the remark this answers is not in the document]
alebedev87: I agree that the service name can be added. Let's see if the PR will get any other review comments, if so, I'll apply the suggestion.

test/e2e/util_gatewayapi_test.go
```diff
@@ -609,30 +610,58 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
 
 	gw := &gatewayapiv1.Gateway{}
 	nsName := types.NamespacedName{Namespace: namespace, Name: name}
-	recordedConditionMsg := "not found"
+	recordedAcceptedConditionMsg, recordedProgrammedConditionMsg := "", ""
 
-	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+	// Wait for the gateway to be accepted and programmed.
+	// Load balancer provisioning can take several minutes on some platforms.
```
candita, line 616: nit: it would be great in the future to set timeouts that vary based on platforms.

pkg/resources/dnsrecord/dns.go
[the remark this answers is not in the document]
candita: @alebedev87 I have some small nits, and the only one I would I feel strongly about is https://github.com/openshift/cluster-ingress-operator/pull/1212/files#r2277556893.  If you want to address that in a followup, just remove the hold.  

/hold
/lgtm

test/e2e/util_gatewayapi_test.go
Miciah: Is there an advantage to using a lower timeout on platforms that don't need the higher timeout?  

I would prefer to take the longest amount of time we expect provisioning to take, and double it.  We don't pay the cost unless provisioning actually takes that amount of time, but we do pay a cost if the test times out when waiting longer would have allowed it to pass.

test/e2e/util_gatewayapi_test.go
```diff
@@ -609,30 +610,58 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+			return false, nil
+		}
+		t.Logf("[%s] Found gateway service: %+v", time.Now().Format(time.DateTime), svc)
 		return false, nil
 	})
 	if err != nil {
-		t.Logf("Last observed gateway:\n%s", util.ToYaml(gw))
-		return nil, fmt.Errorf("gateway %v not %v, last recorded status message: %s", nsName, gatewayapiv1.GatewayConditionAccepted, recordedConditionMsg)
+		t.Logf("[%s] Last observed gateway:\n%s", time.Now().Format(time.DateTime), util.ToYaml(gw))
+		return nil, fmt.Errorf("gateway %v does not have all expected conditions, last recorded status messages: %q, %q", nsName, recordedAcceptedConditionMsg, recordedProgrammedConditionMsg)
```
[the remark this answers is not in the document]
Miciah: The previous log line, with `util.ToYaml(gw)`, should already log the Accepted and Programmed status conditions; is it useful to include them in the error message as well?

test/e2e/util_gatewayapi_test.go
candita: We pay the cost of higher timeouts when everything is failing.  I'm suggesting that we be more choosy with expected timeouts to take advantage of tests failing faster on platforms that don't take as long to provision.  It's just a nit though, so no big deal.

Miciah: > We pay the cost of higher timeouts when everything is failing.

Can we test for "everything is failing" in the polling loop, and exit early if the condition is detected?  That would fail faster, flake less, and provide useful diagnostic information right in the test output.

candita: Yes, but let's prioritize the merge so we can get better CI signal.

candita:
We know the root cause for failures in CI for e2e-gcp-operator and they are being handled by the cloud controller team.
/override e2e-gcp-operator

test/e2e/util_gatewayapi_test.go
Miciah: > Yes, but let's prioritize the merge so we can get better CI signal.

Agreed.  The PR has already merged, but in general when we have tests that are flaky because polling loops time out, I would argue for (1) increasing timeouts as a first pass in order to reduce flakiness and (2) refining the failure conditions as a second pass in order to fail faster and improve diagnostics.  Optimizing to reduce flakiness virtually always takes precedence over optimizing to fail fast.

PR #1213, opened 2025-04-03
NE-1957: Add Gateway API DNS Feature e2e tests

grzpiotrowski:
This PR adds basic e2e tests for the Gateway API DNS feature.

The tests create gateways in different configurations and check for expected DNSRecords to be present or not existing, depending on the test case.

The test cases covered are:
- Multiple gateways with listeners of the same hostname.
- Multiple gateways with listeners with overlapping hostnames.
- Creating a gateway with multiple listeners and then updating and deleting the listeners to ensure the DNSRecords are reconciled accordingly.
- Gateway with a listener with no hostname.

### DIFF
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/openshift/api/features"
+	iov1 "github.com/openshift/api/operatoringress/v1"
 	operatorclient "github.com/openshift/cluster-ingress-operator/pkg/operator/client"
 	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
 
@@ -101,6 +102,10 @@ func TestGatewayAPI(t *testing.T) {
 		t.Run("testGatewayAPIObjects", testGatewayAPIObjects)
 		t.Run("testGatewayAPIManualDeployment", testGatewayAPIManualDeployment)
 		t.Run("testGatewayAPIIstioInstallation", testGatewayAPIIstioInstallation)
+		t.Run("testGatewayAPIDNS", testGatewayAPIDNS)
+		t.Run("testGatewayAPIDNSListenerUpdate", testGatewayAPIDNSListenerUpdate)
+		t.Run("testGatewayAPIDNSListenerWithNoHostname", testGatewayAPIDNSListenerWithNoHostname)
+
 	} else {
 		t.Log("Gateway API Controller not enabled, skipping controller tests")
 	}
@@ -421,6 +426,279 @@ func testGatewayAPIRBAC(t *testing.T) {
 			}
 		}
 	}
+}
+
+func testGatewayAPIDNS(t *testing.T) {
+	domain := "gws." + dnsConfig.Spec.BaseDomain
+
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	testCases := []struct {
+		name                       string
+		createGateways             []testGateway
+		expectedListenerConditions []metav1.Condition
+		expectedDNSRecords         map[expectedDnsRecord]bool
+	}{
+		// TODO: In this case Gateway Listeners should be reported as conflicted. To be fixed in the future release.
+		{
+			name: "multipleGatewaysSameListenerHostname",
+			createGateways: []testGateway{
+				{gatewayName: "gw1",
+					namespace: operatorcontroller.DefaultOperandNamespace,
+					listeners: []testListener{
+						{
+							name:     "http",
+							hostname: ptr.To("abc." + domain),
+						},
+					},
+				},
+				{gatewayName: "gw2",
+					namespace: operatorcontroller.DefaultOperandNamespace,
+					listeners: []testListener{
+						{
+							name:     "http",
+							hostname: ptr.To("abc." + domain)},
+					},
+				},
+			},
+			expectedListenerConditions: []metav1.Condition{
+				{Type: "Accepted", Status: metav1.ConditionTrue},
+				{Type: "Conflicted", Status: metav1.ConditionFalse},
+				{Type: "Programmed", Status: metav1.ConditionTrue},
+				{Type: "ResolvedRefs", Status: metav1.ConditionTrue},
+			},
+			expectedDNSRecords: map[expectedDnsRecord]bool{
+				{dnsName: "abc." + domain + ".", gatewayName: "gw1"}: true,
+				{dnsName: "abc." + domain + ".", gatewayName: "gw2"}: true,
+			},
+		},
+		{
+			name: "gatewayListenersWithOverlappingHostname",
+			createGateways: []testGateway{
+				{gatewayName: "gw3",
+					namespace: operatorcontroller.DefaultOperandNamespace,
+					listeners: []testListener{
+						{
+							name:     "http",
+							hostname: ptr.To("qwe." + domain)},
+					},
+				},
+				{gatewayName: "gw4",
+					namespace: operatorcontroller.DefaultOperandNamespace,
+					listeners: []testListener{
+						{
+							name:     "http",
+							hostname: ptr.To("*." + domain),
+						},
+					},
+				},
+			},
+			expectedListenerConditions: []metav1.Condition{
+				{Type: "Accepted", Status: metav1.ConditionTrue},
+				{Type: "Conflicted", Status: metav1.ConditionFalse},
+				{Type: "Programmed", Status: metav1.ConditionTrue},
+				{Type: "ResolvedRefs", Status: metav1.ConditionTrue},
+			},
+			expectedDNSRecords: map[expectedDnsRecord]bool{
+				{dnsName: "qwe." + domain + ".", gatewayName: "gw3"}: true,
+				{dnsName: "*." + domain + ".", gatewayName: "gw4"}:   true,
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var gateways []*gatewayapiv1.Gateway
+
+			// Create gateways
+			for _, gateway := range tc.createGateways {
+				createdGateway, err := createGatewayWithListeners(gatewayClass, gateway.gatewayName, gateway.namespace, gateway.listeners)
+				gateways = append(gateways, createdGateway)
+				if err != nil {
+					t.Fatalf("failed to create gateway %s: %v", gateway.gatewayName, err)
+				}
+			}
+
+			t.Cleanup(func() {
+				for _, gateway := range gateways {
+					if err := kclient.Delete(context.TODO(), gateway); err != nil {
+						if errors.IsNotFound(err) {
+							continue
+						}
+						t.Errorf("Failed to delete gateway %q: %v", gateway.Name, err)
+					}
+				}
+			})
+
+			for _, gateway := range gateways {
+				gateway, err := assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, gateway.Name)
+				if err != nil {
+					t.Errorf("Failed to create %s gateway: %v", gateway.Name, err)
+				}
+			}
+
+			if err := assertExpectedDNSRecords(t, tc.expectedDNSRecords); err != nil {
+				t.Fatalf("dnsRecord expectations not met: %v", err)
+			}
+
+			t.Logf("Check if gateways listener conditions match expected state.")
+			for _, gateway := range gateways {
+				for _, listener := range gateway.Spec.Listeners {
+					if err := waitForGatewayListenerCondition(t, types.NamespacedName{Namespace: gateway.Namespace, Name: gateway.Name}, string(listener.Name), tc.expectedListenerConditions...); err != nil {
+						t.Fatalf("did not get expected listener %s condition: %v", listener.Name, err)
+					} else {
+						t.Logf("gateway: %s listener %s conditions match expected state.", gateway.Name, listener.Name)
+					}
+				}
+			}
+		})
+	}
+
+}
+
+func testGatewayAPIDNSListenerUpdate(t *testing.T) {
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	domain := "gws." + dnsConfig.Spec.BaseDomain
+
+	gateway, err := createGatewayWithListeners(gatewayClass, "test-gateway-update", operatorcontroller.DefaultOperandNamespace, []testListener{
+		{name: "http-listener1", hostname: ptr.To("foo." + domain)},
+		{name: "http-listener2", hostname: ptr.To("bar." + domain)},
+	})
+
+	t.Cleanup(func() {
+		if err := kclient.Delete(context.TODO(), gateway); err != nil {
+			if errors.IsNotFound(err) {
+				return
+			}
+			t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+		}
+	})
+
+	if err != nil {
+		t.Fatalf("failed to create gateway with multiple listeners: %v", err)
+	}
+	t.Logf("Created gateway %s with multiple hostnames", gateway.Name)
+
+	gateway, err = assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, "test-gateway-update")
+	if err != nil {
+		t.Fatalf("failed to accept/program gateway test-gateway-update: %v", err)
+	}
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "foo." + domain + ".", gatewayName: "test-gateway-update"}: true,
+	}); err != nil {
+		t.Fatalf("DNSRecord %s expectations not met: %v", "foo."+domain+".", err)
+	}
+
+	// Fetch the latest version of the Gateway resource.
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: operatorcontroller.DefaultOperandNamespace, Name: gateway.Name}, gateway); err != nil {
+		t.Fatalf("failed to get gateway resource: %v", err)
+	}
+
+	// Modify one gateway listener hostname
+	newHostname := gatewayapiv1.Hostname("baz." + domain)
+	gateway.Spec.Listeners[0].Hostname = &newHostname
+
+	if err := kclient.Update(context.TODO(), gateway); err != nil {
+		t.Fatalf("failed to update gateway %v: %v", gateway.Name, err)
+	}
+	t.Logf("Modified gateway %s's listener hostname from %s to: %s", gateway.Name, "foo."+domain+".", "baz."+domain+".")
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "foo." + domain + ".", gatewayName: "test-gateway-update"}: false,
+		{dnsName: "bar." + domain + ".", gatewayName: "test-gateway-update"}: true,
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: true,
+	}); err != nil {
+		t.Fatalf("DNSRecord %s expectations not met: %v", "baz."+domain+".", err)
+	}
+
+	// Fetch the latest version of the Gateway resource.
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: operatorcontroller.DefaultOperandNamespace, Name: gateway.Name}, gateway); err != nil {
+		t.Fatalf("failed to get gateway resource: %v", err)
+	}
+
+	// Delete one of the listeners
+	gateway.Spec.Listeners = gateway.Spec.Listeners[0 : len(gateway.Spec.Listeners)-1]
+
+	if err := kclient.Update(context.TODO(), gateway); err != nil {
+		t.Fatalf("failed to update gateway with removed listener: %v", err)
+	}
+	t.Logf("Deleted listener %s from gateway %s.", "http-listener2", gateway.Name)
+
+	t.Logf("Checking that the dnsRecord for %s gets removed after removing the listener.", "bar."+domain+".")
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: true,
+		{dnsName: "bar." + domain + ".", gatewayName: "test-gateway-update"}: false,
+	}); err != nil {
+		t.Fatalf("expected bar.%s. to be deleted, but it was not", domain)
+	}
+
+	if err := kclient.Delete(context.TODO(), gateway); err != nil {
+		t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+	}
+
+	t.Logf("Checking the remaining DNSRecord %s gets deleted after gateway deletion.", "baz."+domain+".")
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: false,
+	}); err != nil {
+		t.Fatalf("expected baz.%s. to be deleted, but it was not", domain)
+	}
+	t.Logf("Confirmed DNSRecord removed after gateway deletion.")
+}
+
+func testGatewayAPIDNSListenerWithNoHostname(t *testing.T) {
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	gateway, err := createGatewayWithListeners(gatewayClass, "test-nohost-gateway", operatorcontroller.DefaultOperandNamespace, []testListener{
+		{name: "http-listener-no-host", hostname: nil},
+	})
+
+	t.Cleanup(func() {
+		if err := kclient.Delete(context.TODO(), gateway); err != nil {
+			t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+		}
+	})
+
+	if err != nil {
+		t.Fatalf("failed to create gateway with a listener with no hostname: %v", err)
+	}
+
+	gateway, err = assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, "test-nohost-gateway")
+	if err != nil {
+		t.Fatalf("failed to accept/program gateway test-nohost-gateway: %v", err)
+	}
+
+	t.Logf("Created gateway %s with a listener with no hostname", gateway.Name)
+
+	err = wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+		dnsRecords := &iov1.DNSRecordList{}
+		// List all DNSRecords from the default operand namespace.
+		if err := kclient.List(context, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
+			t.Logf("failed to list DNSRecords: %v; retrying...", err)
+			return false, nil
+		}
+
+		for _, record := range dnsRecords.Items {
+			if record.Labels["gateway.networking.k8s.io/gateway-name"] == gateway.Name {
+				t.Fatalf("dnsrecord found while not expected: %v", err)
+			}
+		}
+		t.Logf("No DNSRecord for gateway listener with no hostname, continuing polling...")
+		return false, nil
+	})
+	t.Logf("Confirmed no DNSRecord created for gateway listener with no hostname.")
+
 }
 
 // ensureCRDs tests that the Gateway API custom resource definitions exist.
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
@@ -227,6 +228,52 @@ func createGateway(gatewayClass *gatewayapiv1.GatewayClass, name, namespace, dom
 			}
 		} else {
 			return nil, fmt.Errorf("failed to create gateway: %v", err.Error())
+		}
+	}
+	return gateway, nil
+}
+
+func createGatewayWithListeners(gatewayClass *gatewayapiv1.GatewayClass, name, namespace string, listeners []testListener) (*gatewayapiv1.Gateway, error) {
+
+	var gatewayListeners []gatewayapiv1.Listener
+	for _, spec := range listeners {
+		var hostname *gatewayapiv1.Hostname = nil
+		if spec.hostname != nil {
+			h := gatewayapiv1.Hostname(*spec.hostname)
+			hostname = &h
+		}
+
+		fromNamespace := gatewayapiv1.FromNamespaces(allNamespaces)
+		allowedRoutes := gatewayapiv1.AllowedRoutes{
+			Namespaces: &gatewayapiv1.RouteNamespaces{From: &fromNamespace},
+		}
+
+		l := gatewayapiv1.Listener{
+			Name:          gatewayapiv1.SectionName(spec.name),
+			Hostname:      hostname,
+			Port:          80,
+			Protocol:      "HTTP",
+			AllowedRoutes: &allowedRoutes,
+		}
+		gatewayListeners = append(gatewayListeners, l)
+	}
+
+	gateway := &gatewayapiv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: gatewayapiv1.GatewaySpec{
+			GatewayClassName: gatewayapiv1.ObjectName(gatewayClass.Name),
+			Listeners:        gatewayListeners,
+		},
+	}
+
+	if err := kclient.Create(context.TODO(), gateway); err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("cannot create gateway %s: %v", name, err.Error())
+		} else {
+			return nil, fmt.Errorf("failed to create gateway %s: %v", name, err.Error())
 		}
 	}
 	return gateway, nil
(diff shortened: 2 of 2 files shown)


test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+				createGateways: []struct {
+					gatewayName string
+					namespace   string
+					hostname    string
+				}{
+					{gatewayName: "gw1", namespace: operatorcontroller.DefaultOperandNamespace}, // TODO: Modify the gw creation function to make the hostname optional, empty hostname is invalid but non existing is ok
+				},
+				expectedListenerConditions: []metav1.Condition{
+					{Type: "Accepted", Status: metav1.ConditionFalse},
+					{Type: "Conflicted", Status: metav1.ConditionTrue},
```
candita, line 489: Is this case showing up as Conflicted?

test/e2e/util_gatewayapi_test.go
```diff
@@ -6,11 +6,13 @@ package e2e
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
candita, line 15: Move this out of the system imports.

test/e2e/util_gatewayapi_test.go
```diff
@@ -232,6 +234,37 @@ func createGateway(gatewayClass *gatewayapiv1.GatewayClass, name, namespace, dom
 	return gateway, nil
 }
 
+func createGatewayWithListener(gatewayClass *gatewayapiv1.GatewayClass, name, namespace, listenerHostname string) (*gatewayapiv1.Gateway, error) {
```
candita, line 237: It would be useful to have a function called `addListenerToGateway`.  Can you turn this into that?

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+func waitForGatewayListenerCondition(t *testing.T, gatewayName types.NamespacedName, listenerName string, conditions ...metav1.Condition) error {
+	t.Helper()
+
+	gateway := &gatewayapiv1.Gateway{}
+	expected := gatewayListenerConditionMap(conditions...)
+	current := map[string]string{}
+
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, gatewayName, gateway); err != nil {
+			t.Logf("failed to get gateway %s: %v", gatewayName.Name, err)
```
candita, line 635:
```suggestion
			t.Logf("failed to get gateway %s, retrying...", gatewayName.Name)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+				current = gatewayListenerConditionMap(listenerStatus.Conditions...)
+				return conditionsMatchExpected(expected, current), nil
+			}
+		}
+
+		if listenerStatus == nil {
+			t.Logf("listener %s not found in gateway %s", listenerName, gatewayName.Name)
+			return false, nil
+		}
+		return false, nil
```
candita, line 652: If you want to stop here do this:
```suggestion
		if listenerStatus == nil {
			t.Logf("listener %s not found in gateway %s", listenerName, gatewayName.Name)
			return false, fmt.Errorf("listener %s not found in gateway %s", listenerName, gatewayName.Name)
		}
```
If you want to retry, do this:
```suggestion
		if listenerStatus == nil {
			t.Logf("listener %s not found in gateway %s, retrying...", listenerName, gatewayName.Name)
			return false, nil
		}
```
Looks like we don't need the last `return false, nil`.

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+	if err != nil {
+		t.Errorf("Expected conditions: %v\n Current conditions: %v", expected, current)
+		if gateway != nil {
+			if gwStatusPretty, err := json.MarshalIndent(gateway.Status, "", "  "); err != nil {
+				t.Fatal(err)
+			} else {
+				t.Logf("Gateway %s/%s status: %s\n", gateway.Namespace, gateway.Name, string(gwStatusPretty))
+			}
+		}
+	}
```
candita, line 664: We don't need to print out the gateway status.
```suggestion
	if err != nil {
		t.Errorf("Expected conditions: %v\n Current conditions: %v", expected, current)
	}
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+// assertListenerDNSRecord checks to make sure a DNSRecord exists in a ready state for the specified gateway listener,
+// and returns an error if not.
+func assertListenerDNSRecord(t *testing.T, gatewayName, listenerHostname string) error {
+	t.Helper()
+
+	if !strings.HasSuffix(listenerHostname, ".") {
+		listenerHostname += "."
+	}
+
+	var dnsRecord *v1.DNSRecord
```
candita, line 685: This shouldn't be a pointer.

test/e2e/util_gatewayapi_test.go
```diff
@@ -849,7 +979,7 @@ func assertDNSRecord(t *testing.T, recordName types.NamespacedName) error {
 	t.Helper()
 	dnsRecord := &v1.DNSRecord{}
 
-	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 3*time.Minute, false, func(context context.Context) (bool, error) {
```
candita, line 982: You can remove this, you're not calling assertDNSRecord in these tests.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			},
+			expectDNSRecords: []struct {
+				dnsName     string
+				gatewayName string
+			}{
+				{dnsName: "foo.apps." + domain, gatewayName: "gw1"},
+				{dnsName: "*.apps." + domain, gatewayName: "gw2"},
+			},
+		},
+	}
```
candita, line 525: These are the 4+ tests listed in the description.  Can you add comments or matching names for these in your test cases?
1. Create multiple Gateways that have listeners with the same hostname, as well as with differing hostnames and test DNS flow.
2. Create a Gateway with Listeners that don't have a hostname
3. Gateway listener trying to claim a fully qualified hostname (abc.apps.<DNS config base domain>) that would match a *.apps.<DNS config base domain> ingress controller wildcard. 
4. DNSRecords handling when the associated Gateway is deleted or hostname was deleted or modified on the Gateway.

test/e2e/gateway_api_test.go
Miciah: We should enable the test case, but adjust the expectations as following:
```go
expectedListenerConditions: []metav1.Condition{
	{Type: "Accepted", Status: metav1.ConditionTrue},
	{Type: "Conflicted", Status: metav1.ConditionFalse},
	{Type: "Programmed", Status: metav1.ConditionTrue},
	{Type: "ResolvedRefs", Status: metav1.ConditionTrue},
```
But add a comment:
```go
// These expectations reflect a bug in Istio.  Istio should be marking the gateways as conflicted
// but currently does not.  See <insert Jira ticket here if we have one>.
```
We'll adjust the test when we bump to a version of Istio that behaves correctly.

candita: In this case, of a listener with no hostname, this is the expected outcome.  If I recall correctly, listeners with no hostname listen to all hosts.

Miciah: I see, I commented on the wrong test case.  Sorry!

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+	expected := gatewayListenerConditionMap(conditions...)
+	current := map[string]string{}
+
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
+		if err := kclient.Get(context, gatewayName, gateway); err != nil {
+			t.Logf("failed to get gateway %s: %v", gatewayName.Name, err)
+			return false, nil
+		}
+
+		var listenerStatus *gatewayapiv1.ListenerStatus
```
candita, line 639: This doesn't need to be a pointer, does it?

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +623,103 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 3*time.Minute, false, func(ctx context.Context) (bool, error) {
+		dnsRecords := &v1.DNSRecordList{}
+		if err := kclient.List(ctx, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
+			t.Logf("failed to list DNSRecords: %v; retrying...", err)
+			return false, nil
+		}
+
+		for _, record := range dnsRecords.Items {
+			if record.Labels["gateway.networking.k8s.io/gateway-name"] == gatewayName &&
+				record.Spec.DNSName == listenerHostname {
```
candita, line 696: Doublecheck on whether you need to add a trailing `.`.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		if err := kclient.Delete(context.TODO(), &gateway); err != nil {
+			if errors.IsNotFound(err) {
+				return
+			}
+			t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+		}
+	})
+
+	if err != nil {
+		t.Fatalf("failed to create first gateway with hostname foo.[...]: %v", err)
```
candita, line 599:
```suggestion
		t.Fatalf("failed to create gateway with a listener of hostname foo.%s: %v", domain, err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	}
+
+	// Modify gateway listener hostname
+	newHostname := gatewayapiv1.Hostname("bar." + domain)
+	gateway.Spec.Listeners[0].Hostname = &newHostname
+
+	if err := kclient.Update(context.TODO(), gateway); err != nil {
+		t.Fatalf("failed to update gateway %v: %v", gateway.Name, err)
+	}
+	t.Logf("Modified gateway's %s listener hostname to: %s", gateway.Name, "bar."+domain)
```
candita, line 625:
```suggestion
	t.Logf("Modified gateway %s's listener hostname to: %s", gateway.Name, "bar."+domain)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		*/
+		// stop here if needed
+		{
+			name: "gatewayListenersWithOverlappingHostname",
+			createGateways: []struct {
+				gatewayName string
+				namespace   string
+				hostname    string
+			}{
+				{gatewayName: "gw1", namespace: operatorcontroller.DefaultOperandNamespace, hostname: "foo.apps." + domain},
```
candita, line 508: Don't use "foo" again, use something new so we can make distinctions in the logs.  Also, why "foo.apps."+domain?  The domain already contains "apps".

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +426,210 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	// Use the dnsConfig base domain set up in TestMain.
+	domain := "gws." + dnsConfig.Spec.BaseDomain
+
+	// Create a Gateway with the listener hostname "foo.[...]"
+	gateway, err := createGatewayWithListener(gatewayClass, "test-gateway", operatorcontroller.DefaultOperandNamespace, "foo."+domain)
```
candita, line 586: Don't use "foo" again here.  The old dnsRecord could still be around, and we want to be able to review logs without confusion.

test/e2e/util_gatewayapi_test.go
candita: To debug, you can just add a line that prints out the record labels and spec.dnsname.

test/e2e/gateway_api_test.go
grzpiotrowski: We moved this test case outside of the test table to reduce some of the complexity.

test/e2e/util_gatewayapi_test.go
grzpiotrowski: For almost all of the test cases we were looking at gateways created with the listeners in initial conditions of the tests.
The currently existing util function `createGateway()` creates a listener with "*.<baseDomain>" hostname by default, so we didn't want to have it mix with what is set up for the tests.

test/e2e/gateway_api_test.go
grzpiotrowski: Added matching names now to identify the tests. All of the listed above are covered now.

test/e2e/gateway_api_test.go
```diff
@@ -6,6 +6,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	v1 "github.com/openshift/api/operatoringress/v1"
```
grzpiotrowski, line 9: To be moved.

test/e2e/util_gatewayapi_test.go
grzpiotrowski: Thank you. The intent was to retry there. Modified the log to reflect that.
Also about the last return, seeing: `Missing the 'return' statement at the end of the function`

test/e2e/gateway_api_test.go
grzpiotrowski: Done. Changed the hostnames to vary between tests.

test/e2e/gateway_api_test.go
grzpiotrowski: I think there was no particular reason for the "foo.apps".
Changed the hostnames to differ between tests.

test/e2e/gateway_api_test.go
grzpiotrowski: Sorry, looks like I missed this one.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	testCases := []struct {
+		name           string
+		createGateways []struct {
+			gatewayName string
+			namespace   string
+			hostname    *string
```
candita, line 444: nit:  For future reference it is unusual to use a pointer for a declaration like this.  We might use them for function signature variables but I don't see any reason a pointer is needed here.  Don't worry about it for now.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			var gateways []*gatewayapiv1.Gateway
+
+			// Create gateways
+			for _, gateway := range tc.createGateways {
+				createdGateway, err := createGatewayWithListeners(gatewayClass, gateway.gatewayName, gateway.namespace, []struct {
+					Name     string
+					Hostname *string
+				}{
+					{Name: "http", Hostname: gateway.hostname},
+				})
```
candita, line 505: nit: The anonymous struct parameter in the function call feels awkward to me, but there's probably nothing technically wrong with it.  Ordinarily you could do something like this first and use the gw parameter:
`gw := []struct{{Name:"http", Hostname: gateway.hostname},}
`

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
 	}
 }
 
+func testGatewayAPIDNS(t *testing.T) {
+	domain := "gws." + dnsConfig.Spec.BaseDomain
+
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
```
candita, line 435: I expect that you'll need to ignore an error saying that this already exists.  It's okay if it already exists.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+				}
+			}
+		})
+	}
+
+}
+
+func testGatewayAPIDNSListenerUpdate(t *testing.T) {
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
```
candita, line 549: https://github.com/openshift/cluster-ingress-operator/pull/1213/files#r2038438805

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		if err := kclient.Delete(context.TODO(), &gateway); err != nil {
+			if errors.IsNotFound(err) {
+				return
+			}
+			t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+		}
+	})
+
+	if err != nil {
+		t.Fatalf("failed to create gateway with a listener of hostname foo.%s: %v", domain, err)
```
candita, line 575: nit: 
```suggestion
		t.Fatalf("failed to create gateway with multiple listeners: %v", domain, err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	}
+
+	gateway, err = assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, "test-gateway-update")
+	if err != nil {
+		t.Fatalf("failed to create the test gateway: %v", err)
```
candita, line 580: nit:
```suggestion
		t.Fatalf("failed to accept/program gateway test-gateway-update: %v", err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	}
+
+	t.Logf("Created gateway %s with hostname %s", gateway.Name, "foo."+domain)
```
candita, line 583: nit:
```suggestion
	t.Logf("Created gateway %s with multiple hostnames", gateway.Name)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "foo." + domain + ".", gatewayName: "test-gateway-update"}: true,
+	}); err != nil {
+		t.Fatalf("DNSRecord expectations not met: %v", err)
```
candita, line 588: nit:
```suggestion
		t.Fatalf("DNSRecord %s expectations not met: %v", "foo."+domain+".", err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+
+	if err := kclient.Update(context.TODO(), gateway); err != nil {
+		t.Fatalf("failed to update gateway %v: %v", gateway.Name, err)
+	}
+	t.Logf("Modified gateway's %s listener hostname to: %s", gateway.Name, "baz."+domain+".")
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: true,
+	}); err != nil {
+		t.Fatalf("DNSRecord expectations not met: %v", err)
```
candita, line 608: nit:
```suggestion
		t.Fatalf("DNSRecord %s expectations not met: %v", "baz".+domain+".", err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		t.Fatalf("failed to update gateway with removed listener: %v", err)
+	}
+	t.Logf("Deleted listener %s from gateway %s", "http-listener2", gateway.Name)
+
+	t.Logf("Checking that the dnsRecord for %s gets removed.", "baz."+domain+".")
+
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: true,
+		{dnsName: "foo." + domain + ".", gatewayName: "test-gateway-update"}: false,
+		{dnsName: "bar." + domain + ".", gatewayName: "test-gateway-update"}: false,
```
candita, line 635: nit:
Maybe it would be better to check for each after making the changes, change the `assertExpectedDNSRecord` to return the bool.  Not a deal-breaker.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			return
+		}
+		t.Errorf("failed to delete gateway %q: %v", gateway.Name, err)
+	}
+
+	// Check the DNSRecord gets deleted for the remaining listener hostname
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: false,
+	}); err != nil {
+		t.Fatalf("DNSRecord expectations not met: %v", err)
```
candita, line 652:
```suggestion
		t.Fatalf("Expected baz.%s. to be deleted, but it wasn't", domain)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	}
+}
+
+func testGatewayAPIDNSListenerWithNoHostname(t *testing.T) {
+	gatewayClass, err := createGatewayClass(operatorcontroller.OpenShiftDefaultGatewayClassName, operatorcontroller.OpenShiftGatewayClassControllerName)
+	if err != nil {
```
candita, line 658: https://github.com/openshift/cluster-ingress-operator/pull/1213/files#r2038438805

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		t.Fatalf("failed to create gatewayclass: %v", err)
+	}
+
+	// Create a Gateway with the listener hostname "foo.[...]"
```
candita, line 662: nit:
```suggestion
	// Create a Gateway with an empty listener hostname
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	gateway, err := createGatewayWithListeners(gatewayClass, "test-nohost-gateway", operatorcontroller.DefaultOperandNamespace, []struct {
+		Name     string
+		Hostname *string
+	}{
+		{Name: "http-listener-no-host", Hostname: nil},
+	})
+
+	t.Cleanup(func() {
+		gateway := gatewayapiv1.Gateway{ObjectMeta: metav1.ObjectMeta{Name: "test-nohost-gateway", Namespace: operatorcontroller.DefaultOperandNamespace}}
```
candita, line 671: nit: you shouldn't need to recreate `gateway`, the original should still be around.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		}
+	})
+
+	if err != nil {
+		t.Fatalf("failed to create gateway with a listener with no hostname: %v", err)
+	}
+
+	gateway, err = assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, "test-nohost-gateway")
+	if err != nil {
+		t.Fatalf("failed to create the test gateway: %v", err)
```
candita, line 686: nit:
```suggestion
		t.Fatalf("failed to accept/program gateway test-nohost-gateway: %v", err)
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,286 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			return false, nil
+		}
+
+		for _, record := range dnsRecords.Items {
+			if record.Labels["gateway.networking.k8s.io/gateway-name"] == gateway.Name {
+				t.Fatalf("dnsrecord found while not expected: %v", err)
+			}
+		}
+
+		t.Logf("No DNSRecord for gateway listener as expected...")
```
candita, line 705: This is getting all the dnsRecords and trying to find the unexpected one.  It could possibly take a little while to delete it, so it would be better for you to just look for one dnsRecord and poll until it disappears, then fail if it isn't gone in 1 minute.

```suggestion
               dnsRecord := &v1.DNSRecord{}
               recordName := operatorcontroller.GatewayDNSRecordName(gateway, domain)
               
		// Try to get the dnsRecord, and if it's not found, exit the poll successfully.
		if err := kclient.Get(context, recordName, dnsRecord); err != nil {
                        if errors.IsNotFound(err) {
				return true, nil
			}
			// If it is found, then keep trying to see if it gets deleted.
			t.Logf("Unexpected failure getting dnsRecord, retrying....")
		}
	        return false, nil
		})
		
		if err != nil {
		    t.Logf("Expected dnsRecord to be deleted but it wasn't")
                }
		t.Logf("No dnsRecord for gateway listener as expected...")
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -232,6 +233,58 @@ func createGateway(gatewayClass *gatewayapiv1.GatewayClass, name, namespace, dom
...
+
+	if err := kclient.Create(context.TODO(), gateway); err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			name := types.NamespacedName{Namespace: namespace, Name: name}
+			if err = kclient.Get(context.TODO(), name, gateway); err != nil {
+				return nil, fmt.Errorf("failed to get the existing gateway: %v", err.Error())
+			}
+		} else {
+			return nil, fmt.Errorf("failed to create gateway: %v", err.Error())
+		}
```
candita, line 283: If it already exists, just return an error.  No need to get the existing one.

```suggestion
		if kerrors.IsAlreadyExists(err) {
				return nil, fmt.Errorf("cannot create gateway %s: %v", name, err.Error())
		} else {
			return nil, fmt.Errorf("failed to create gateway %s: %v", name, err.Error())
		}
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +643,132 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+
+		listenerStatus := gatewayapiv1.ListenerStatus{}
+		for _, ls := range gateway.Status.Listeners {
+			if string(ls.Name) == listenerName {
+				listenerStatus = ls
+				current = gatewayListenerConditionMap(listenerStatus.Conditions...)
+				return conditionsMatchExpected(expected, current), nil
+			}
+		}
+		if &listenerStatus == nil {
```
candita, line 667: Shouldn't this be:
```suggestion
		if listenerStatus == nil {
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +643,132 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+			t.Logf("listener %s not found in gateway %s, retrying...", listenerName, gatewayName.Name)
```
candita, line 668:
```suggestion
			t.Logf("listener %s has not been updated with status, retrying...", listenerName, gatewayName.Name)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +643,132 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+			return false, nil
+		}
+		return false, nil
+	})
+
+	if err != nil {
+		t.Errorf("Expected conditions: %v\n Current conditions: %v", expected, current)
```
candita, line 675:
```suggestion
		return fmt.Errorf("Expected conditions: %v don't match\n Current conditions: %v", expected, current)
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +643,132 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+
+// expectedDnsRecord is used to represent a DNSRecord expectation.
+type expectedDnsRecord struct {
+	dnsName     string
+	gatewayName string
+}
+
+// assertExpectedDNSRecords polls until the DNSRecords in the default operand namespace match the given expectations.
+// The expectations parameter is a map where keys are expectations for DNSRecord and values indicate whether a DNSRecord should be present.
+func assertExpectedDNSRecords(t *testing.T, expectations map[expectedDnsRecord]bool) error {
```
candita, line 696: This would be easier if you did looked at one dnsRecord at a time and returned the bool for it.

candita:
@grzpiotrowski don't worry about the comments marked as `nit` for now.  However, there seems to be an issue in testGatewayAPIDNSListenerUpdate as I mentioned in slack.  Please try to look into simplifying the assertexpectedDNSRecords function.

test/e2e/gateway_api_test.go
grzpiotrowski: Having this as a pointer was enabling us to treat the `hostname` field as optional by setting it to `nil` [(inspiration source)](https://github.com/openshift/cluster-ingress-operator/blob/e3ad4c57081c3236bde76d67cf0b1e1c47fe6eff/pkg/operator/controller/ingress/poddisruptionbudget_test.go#L13-L22).
For all test cases in [testGatewayAPIDNS](https://github.com/grzpiotrowski/cluster-ingress-operator/blob/920e509eb6c3f94a1500834c49eadeffa5ef09e8/test/e2e/gateway_api_test.go#L431) we have the hostname specified, so it is not really used here at this time.
However we are using `hostname=nil` in the [testGatewayAPIDNSListenerWithNoHostname](https://github.com/grzpiotrowski/cluster-ingress-operator/blob/920e509eb6c3f94a1500834c49eadeffa5ef09e8/test/e2e/gateway_api_test.go#L663-L668) test.

All of these tests are reusing [createGatewayWithListeners](https://github.com/grzpiotrowski/cluster-ingress-operator/blob/920e509eb6c3f94a1500834c49eadeffa5ef09e8/test/e2e/util_gatewayapi_test.go#L236C6-L286) function, so the use of pointer for `hostname` stems from this generalization.

test/e2e/gateway_api_test.go
grzpiotrowski: [createGatewayClass](https://github.com/openshift/cluster-ingress-operator/blob/920e509eb6c3f94a1500834c49eadeffa5ef09e8/test/e2e/util_gatewayapi_test.go#L298) returns the gatewayClass in case it already exists. `IsAlreadyExists` error is ignored within that function already.

grzpiotrowski:
Made some mistakes in the testGatewayAPIDNSListenerUpdate test flow by confusing expected hostnames, should be ok now.

Also the `assertExpectedDNSRecords` function was not setting the `found=true` flag in case it found a non expected DNSRecord.
And added more logging to this function to help with debugging and readability of this function.

Refrained from breaking it down into a few simpler checks for now. Want to check it working correctly now and perhaps see if it's more acceptable with additional logs.

test/e2e/util_gatewayapi_test.go
```diff
@@ -590,6 +643,132 @@ func assertGatewaySuccessful(t *testing.T, namespace, name string) (*gatewayapiv
...
+				return false, nil
+			}
+			// If the record is not expected but was found, return false to continue polling.
+			if !shouldBePresent && found {
+				t.Logf("DNSRecord for hostname %q (gateway: %s) is present but was expected to be absent; retrying...", exp.dnsName, exp.gatewayName)
+				expectationsMet = false
+				return false, nil
+			}
+		}
+		t.Logf("Continuing polling to ensure non-expected DNSRecords do not exist...")
```
grzpiotrowski, line 762:
```suggestion
		if haveExpectNotPresent {
			t.Logf("Continuing polling to ensure non-expected DNSRecords do not exist...")
		}
```

To make it not log this misleadingly a single time when there is no negative expectations.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create gateway with multiple listeners: %v", err)
+	}
+
+	gateway, err = assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, "test-gateway-update")
+	if err != nil {
+		t.Fatalf("failed to create the test gateway: %v", err)
+	}
+
+	t.Logf("Created gateway %s with hostname %s", gateway.Name, "foo."+domain)
```
alebedev87, line 582: This message should be logged after creation. At this point we ensured the gateway is accepted, so the message should say so.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "foo." + domain + ".", gatewayName: "test-gateway-update"}: true,
+	}); err != nil {
+		t.Fatalf("DNSRecord expectations not met: %v", err)
+	}
+
+	// Fetch the latest version of the Gateway resource.
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: operatorcontroller.DefaultOperandNamespace, Name: gateway.Name}, gateway); err != nil {
+		t.Fatalf("failed to get gateway resource: %v", err)
+	}
```
alebedev87, line 593: You get the gateway from `assertGatewaySuccessful` (it polls with `Get`s), do you really need to do it again?

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+
+	// Delete one of the listeners
+	var updatedListeners []gatewayapiv1.Listener
+	for _, l := range gateway.Spec.Listeners {
+		if l.Name == "http-listener2" { //
+			continue
+		}
+		updatedListeners = append(updatedListeners, l)
+	}
+	gateway.Spec.Listeners = updatedListeners
```
alebedev87, line 625:
```suggestion
	gateway.Spec.Listeners = gateway.Spec.Listeners[0:len(gateway.Spec.Listeners)-1]
```

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+	if err := assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: "baz." + domain + ".", gatewayName: "test-gateway-update"}: true,
+		{dnsName: "bar." + domain + ".", gatewayName: "test-gateway-update"}: false,
+	}); err != nil {
+		t.Fatalf("DNSRecord expectations not met: %v", err)
+	}
+
+	if err := kclient.Delete(context.TODO(), gateway); err != nil {
+		if errors.IsNotFound(err) {
+			return
```
alebedev87, line 642: Why just `return`? Your gateway should be there, if someone deleted it it's a test `Error()`.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+		Hostname *string
+	}{
+		{Name: "http-listener-no-host", Hostname: nil},
+	})
+
+	t.Cleanup(func() {
+		if err := kclient.Delete(context.TODO(), gateway); err != nil {
+			if errors.IsNotFound(err) {
+				return
+			}
```
alebedev87, line 673: In this test we don't delete gateway as part of the test, so it expect the gateway to be there.

alebedev87 (commented):
First review. I will need to re-iterate later.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			},
+			expectedListenerConditions: []metav1.Condition{
+				{Type: "Accepted", Status: metav1.ConditionTrue},
+				{Type: "Conflicted", Status: metav1.ConditionFalse},
+				{Type: "Programmed", Status: metav1.ConditionTrue},
+				{Type: "ResolvedRefs", Status: metav1.ConditionTrue},
+			},
+			expectedDNSRecords: map[expectedDnsRecord]bool{
+				{dnsName: "abc." + domain + ".", gatewayName: "gw1"}: true,
+				{dnsName: "abc." + domain + ".", gatewayName: "gw2"}: true,
```
alebedev87, line 468: So, it will be the same DNS record in the same zone. Which gateway's target will win then (on AWS, on GCP, on Azure)? I saw [this thread](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1743690816760349). What is the conclusion? Do we accept this a valid GatewayAPI (`Conflicted=False`) and buggy DNS management in CIO? Or we consider the latter as valid too (I would have a question about it then)?

candita: See the comment: 
`// TODO: In this case Gateway Listeners should be reported as conflicted. To be fixed in the future release.`
This is due to Istio, and it will need to be documented as an Istio-ism.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+			for _, gateway := range tc.createGateways {
+				createdGateway, err := createGatewayWithListeners(gatewayClass, gateway.gatewayName, gateway.namespace, []struct {
+					Name     string
+					Hostname *string
+				}{
+					{Name: "http", Hostname: gateway.hostname},
+				})
+				gateways = append(gateways, createdGateway)
+				if err != nil {
+					t.Fatalf("failed to create gateway %s: %v", gateway.gatewayName, err)
```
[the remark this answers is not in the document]
candita: It should be ok.  For these specific tests, there are only 2 gateways, and both must be created for the test to be valid.

test/e2e/gateway_api_test.go
```diff
@@ -423,6 +428,284 @@ func testGatewayAPIRBAC(t *testing.T) {
...
+						}
+						t.Fatalf("failed to delete gateway %q: %v", gateway.Name, err)
+					}
+				}
+			})
+
+			for _, gateway := range gateways {
+				gateway, err := assertGatewaySuccessful(t, operatorcontroller.DefaultOperandNamespace, gateway.Name)
+				if err != nil {
+					t.Fatalf("failed to create %s gateway: %v", gateway.Name, err)
```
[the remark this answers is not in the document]
candita: The test is invalid if only one gateway was created, so that's not needed.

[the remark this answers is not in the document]
alebedev87: The test verifies both gateways, so it's more productive to get the test results for both gateways in a single run than do 2 (or more runs) to see the problems all the gateways.

test/e2e/gateway_api_test.go
alebedev87: I think that there are 2 sides to this thing: Istio (`Conflicted` listener status) and cluster ingress operator (`DNSRecord`). Do we consider the `DNSRecord` logic valid? On AWS, for instance, two or more gateways will be overriding CNAME record. Is this the behavior we expect?

test/e2e/gateway_api_test.go
grzpiotrowski: At some stage I got hit by a `object has been modified` error on this update.
I did not investigate in detail what was updated in the meantime between the `assertGatewaySuccessful` and the gateway update attempt
But I suspect it could have something to do that `assertGatewaySuccessful` only cares about the `Accepted`  condition but while waiting for the DNSRecords before the update, the gateway also goes into `Programmed` state?

I reckon these [changes](https://github.com/openshift/cluster-ingress-operator/pull/1212/files#diff-e5602046dd05b4971c4090c6e11bf804af9c9311b091ed909452cdf94d7eae15R566-R620) from PR  #1212 would mitigate this?

test/e2e/gateway_api_test.go
grzpiotrowski: :bulb:

grzpiotrowski:
Changing back to draft before the changes push to then run the `e2e-aws-gatewayapi` test alone for quicker results.

grzpiotrowski:
:four_leaf_clover: 

/test e2e-aws-gatewayapi

test/e2e/gateway_api_test.go
candita: Yes, we should open a bug because the dnsRecord shouldn't be created for a Listener that is Conflicted=True.

test/e2e/gateway_api_test.go
grzpiotrowski: In this case our intent was to look if any DNSRecord gets wrongly created for a gateway which has only one listener, which does not have a hostname field.
I think that `GatewayDNSRecordName` would not be appropriate here as we would need to know the DNSRecord name. Also conciously avoided using this function throughout the tests as it is used by the controller and wanted to avoid usage of the tested code. But maybe that was being overly cautious.

Or did I misunderstand and for this test case a valid script would be to:
 - create a gateway with a listener with `foo.domain.` hostname
 - delete the listener/replace it with one with no hostname
 - assert the DNSRecord for `foo.domain.` was deleted
 ?

test/e2e/util_gatewayapi_test.go
grzpiotrowski: I think that this variable holds the `ListenerStatus` struct and to assert we didn't get it we'd check the pointer?

test/e2e/gateway_api_test.go
alebedev87: Right, `assertGatewaySuccessful` only checks for `Accepted` condition which is not the last modification before the gateway is ready. So, yes, adding a check for `Programmed` status to `assertGatewaySuccessful` can help to eliminate the need for this `Get()` here. You can keep it for my PR to follow up on this then.

test/e2e/gateway_api_test.go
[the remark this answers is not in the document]
alebedev87: Got it, yes, I agree on this point. Then you have to move `t.Cleanup()`  before the creation because here you exit the test without the cleanup.

test/e2e/util_gatewayapi_test.go
candita: I'm pretty sure `&listenerStatus` would always be a value because an address cannot itself be nil.  You can test this later by debug printing out the value of `&listenerStatus`.

test/e2e/util_gatewayapi_test.go
candita: In all honesty, I can't review this as-is and it needs to be clarified asap, but doesn't block the merge unless there are continuing failures.

candita:
Will approve as-is for early CI testing. Some minor PR fixes are still pending.
/lgtm
/approve

grzpiotrowski:
looks like infrastructure issues.

/retest-required

PR #1216, opened 2025-04-08
NE-2009: Move VAP to Default featureset

alebedev87:
The `openshift-ingress-operator-gatewayapi-crd-admission` Validating Admission Policy is bound to `GatewayAPI` featuregate, which has been promoted to GA (https://github.com/openshift/api/pull/2261).
This PR removes the `feature-set` annotation from manifests related to VAP, this effectively makes them part of the `Default` featureset.

### DIFF
--- a/manifests/01-validating-admission-policy-binding.yaml
+++ b/manifests/01-validating-admission-policy-binding.yaml
@@ -7,7 +7,6 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
-    release.openshift.io/feature-set: "TechPreviewNoUpgrade,CustomNoUpgrade"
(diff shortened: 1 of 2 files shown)


alebedev87:
/hold

Until https://github.com/openshift/api/pull/2261 is merged.

Miciah:
/approve
/lgtm

You might want to remove the link to https://github.com/openshift/api/pull/2261 from the commit message.  Otherwise GitHub adds a back-reference from the referenced PR to the commit itself.  This can be annoying in the case of backports (unlikely in this case), rebases, or anything else that creates a new commit with a new commit id but the same commit message.

Miciah:
Thanks!
/lgtm

alebedev87:
As discussed with Joel, since the VAP doesn't break existing `origin` tests and to avoid overriding failed jobs in https://github.com/openshift/api/pull/2261, I'm merging this PR first.

/unhold

PR #1217, opened 2025-04-09
OCPBUGS-54745: status: Conditionally add CRDs to relatedObjects

Miciah:
#### status: Omit istios namespace in `relatedObjects`

Omit the namespace for the istios resource in `relatedObjects` as the istios resource is cluster-scoped.

#### status: Omit gateways namespace in `relatedObjects`

Omit the namespace for the gateways resource in `relatedObjects` as Istio manages gateways in all namespaces.

#### `getOperatorState`: Add `context` parameter

Pass `ctx` from the controller's `Reconcile` method to `getOperatorState, and use it instead of `context.TODO()`.

#### status: Conditionally add CRDs to `relatedObjects`

Check whether the gatewayclasses, gateways, and istios CRDs actually exist before adding them to `relatedObjects`.

Watch customresourcedefinitions in the status controller so that it updates `relatedObjects` as these CRDs are created.

Check the "GatewayAPIController" featuregate to determine whether to add the gatewayclasses, gateways, istios, and subscriptions resources to relatedObjects, in addition to checking the "GatewayAPI" featuregate.

Before this change, the operator could add istios to `relatedObjects` even if the OSSM subscription failed to install.  By convention, an operator should only add resources to `relatedObjects` if those resources exist.

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -183,7 +183,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 	oldStatus := co.Status.DeepCopy()
 
-	state, err := r.getOperatorState(ingressNamespace, canaryNamespace, co)
+	state, err := r.getOperatorState(ctx, ingressNamespace, canaryNamespace, co)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to get operator state: %v", err)
 	}
@@ -320,11 +320,11 @@ type operatorState struct {
 
 // getOperatorState gets and returns the resources necessary to compute the
 // operator's current state.
-func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string, co *configv1.ClusterOperator) (operatorState, error) {
+func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, canaryNamespace string, co *configv1.ClusterOperator) (operatorState, error) {
 	state := operatorState{}
 
 	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ingressNamespace}}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: ingressNamespace}, ns); err != nil {
+	if err := r.client.Get(ctx, types.NamespacedName{Name: ingressNamespace}, ns); err != nil {
 		if !errors.IsNotFound(err) {
 			return state, fmt.Errorf("failed to get namespace %q: %v", ingressNamespace, err)
 		}
@@ -333,7 +333,7 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string, 
 	}
 
 	ns = &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: canaryNamespace}}
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: canaryNamespace}, ns); err != nil {
+	if err := r.client.Get(ctx, types.NamespacedName{Name: canaryNamespace}, ns); err != nil {
 		if !errors.IsNotFound(err) {
 			return state, fmt.Errorf("failed to get namespace %q: %v", canaryNamespace, err)
 		}
@@ -342,7 +342,7 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string, 
 	}
 
 	ingressList := &operatorv1.IngressControllerList{}
-	if err := r.cache.List(context.TODO(), ingressList, client.InNamespace(r.config.Namespace)); err != nil {
+	if err := r.cache.List(ctx, ingressList, client.InNamespace(r.config.Namespace)); err != nil {
 		return state, fmt.Errorf("failed to list ingresscontrollers in %q: %v", r.config.Namespace, err)
 	} else {
 		state.IngressControllers = ingressList.Items
@@ -352,7 +352,7 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string, 
 		if r.config.MarketplaceEnabled && r.config.OperatorLifecycleManagerEnabled {
 			var subscription operatorsv1alpha1.Subscription
 			subscriptionName := operatorcontroller.ServiceMeshOperatorSubscriptionName()
-			if err := r.cache.Get(context.TODO(), subscriptionName, &subscription); err != nil {
+			if err := r.cache.Get(ctx, subscriptionName, &subscription); err != nil {
 				if !errors.IsNotFound(err) {
 					return state, fmt.Errorf("failed to get subscription %q: %v", subscriptionName, err)
 				}
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -237,13 +237,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 				Group:    sailv1.GroupVersion.Group,
 				Resource: "istios",
 			})
-			if state.IngressNamespace != nil {
-				related = append(related, configv1.ObjectReference{
-					Group:     gatewayapiv1.GroupName,
-					Resource:  "gateways",
-					Namespace: state.IngressNamespace.Name,
-				})
-			}
+			related = append(related, configv1.ObjectReference{
+				Group:     gatewayapiv1.GroupName,
+				Resource:  "gateways",
+				Namespace: "", // Include all namespaces.
+			})
 		}
 	}
 
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -24,9 +24,11 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	utilclock "k8s.io/utils/clock"
 
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -50,13 +52,27 @@ const (
 
 	ingressesEqualConditionMessage = "desired and current number of IngressControllers are equal"
 
+	// gatewaysResourceName is the name of the Gateway API gateways CRD.
+	gatewaysResourceName = "gateways.gateway.networking.k8s.io"
+	// gatewayclassesResourceName is the name of the Gateway API
+	// gatewayclasses CRD.
+	gatewayclassesResourceName = "gatewayclasses.gateway.networking.k8s.io"
+	// istiosResourceName is the name of the Sail Operator istios CRD.
+	istiosResourceName = "istios.sailoperator.io"
+
 	controllerName = "status_controller"
 )
 
-var log = logf.Logger.WithName(controllerName)
+var (
+	log = logf.Logger.WithName(controllerName)
 
-// clock is to enable unit testing
-var clock utilclock.Clock = utilclock.RealClock{}
+	// clock is to enable unit testing
+	clock utilclock.Clock = utilclock.RealClock{}
+
+	// relatedObjectsCRDs is a set of names of CRDs that we add to
+	// relatedObjects if they exist.
+	relatedObjectsCRDs = sets.New[string](gatewaysResourceName, gatewayclassesResourceName, istiosResourceName)
+)
 
 // New creates the status controller. This is the controller that handles all
 // the logic for creating the ClusterOperator operator and updating its status.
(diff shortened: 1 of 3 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -331,13 +382,44 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string)
...
+		if r.config.GatewayAPIControllerEnabled {
+			var (
+				crd                                  apiextensionsv1.CustomResourceDefinition
+				gatewaysResourceNamespacedName       = types.NamespacedName{Name: gatewaysResourceName}
+				gatewayclassesResourceNamespacedName = types.NamespacedName{Name: gatewayclassesResourceName}
+				istiosResourceNamespacedName         = types.NamespacedName{Name: istiosResourceName}
+			)
+
+			if err := r.cache.Get(ctx, gatewaysResourceNamespacedName, &crd); err != nil {
+				if !errors.IsNotFound(err) {
```
candita, line 402: Does it matter much if there was a prior resource, then it was deleted?  E.g. someone removed the Istio or other CRDs we're checking here - do we then need to remove these from relatedObjects?  Or is it okay because this is reconciled periodically?

candita:
Cluster install failure

pkg/operator/controller/status/controller.go
Miciah: The watch on CRDs should trigger reconciliation if the istios CRD is deleted: https://github.com/openshift/cluster-ingress-operator/blob/0fa1281ffd86f48464e2f09c302f47f41bc4a8d6/pkg/operator/controller/status/controller.go#L135-L143

`getOperatorState` intializes `operatorState` leaving `haveIstiosResource` false and only sets it to true if a `Get` on the istios CRD returns a nil error value, meaning the CRD is present:
https://github.com/openshift/cluster-ingress-operator/blob/0fa1281ffd86f48464e2f09c302f47f41bc4a8d6/pkg/operator/controller/status/controller.go#L354-L355
https://github.com/openshift/cluster-ingress-operator/blob/0fa1281ffd86f48464e2f09c302f47f41bc4a8d6/pkg/operator/controller/status/controller.go#L415-L421

Then `Reconcile` only adds istios to `relatedObjects` if `haveIstiosResource` is true:
https://github.com/openshift/cluster-ingress-operator/blob/0fa1281ffd86f48464e2f09c302f47f41bc4a8d6/pkg/operator/controller/status/controller.go#L255-L260

So we should be fine.

candita:
E2E Test failure:
 operator_test.go:254: 
Expected [{Group: Resource:namespaces Namespace: Name:openshift-ingress-operator} 
{Group:operator.openshift.io Resource:ingresscontrollers Namespace:openshift-ingress-operator Name:} {Group:ingress.operator.openshift.io Resource:dnsrecords Namespace:openshift-ingress-operator Name:} 
{Group: Resource:namespaces Namespace: Name:openshift-ingress} 
{Group:ingress.operator.openshift.io Resource:dnsrecords Namespace:openshift-ingress Name:} 
{Group: Resource:namespaces Namespace: Name:openshift-ingress-canary} 
{Group:gateway.networking.k8s.io Resource:gatewayclasses Namespace: Name:}],

found [{Group: Resource:namespaces Namespace: Name:openshift-ingress-operator} 
{Group:operator.openshift.io Resource:ingresscontrollers Namespace:openshift-ingress-operator Name:} {Group:ingress.operator.openshift.io Resource:dnsrecords Namespace:openshift-ingress-operator Name:} 
{Group: Resource:namespaces Namespace: Name:openshift-ingress} 
{Group:ingress.operator.openshift.io Resource:dnsrecords Namespace:openshift-ingress Name:} 
{Group: Resource:namespaces Namespace: Name:openshift-ingress-canary} 
{Group:gateway.networking.k8s.io Resource:gatewayclasses Namespace: Name:} 
**{Group:gateway.networking.k8s.io Resource:gateways Namespace:openshift-ingress Name:}]**
    operator_test.go:264: Did not get expected status related objects: timed out waiting for the condition

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/0fa1281ffd86f48464e2f09c302f47f41bc4a8d6..c7db35d125f750f799fe98bf48a850089b1b09e2 makes two changes:

- Only add the "subscriptions" resource to `relatedObjects` if the "GatewayAPIController" featuregate is enabled.
- Adjust `TestClusterOperatorStatusRelatedObjects` to expect "gateways" in `relatedObjects` if the "GatewayAPIController" featuregate is enabled.

test/e2e/operator_test.go
```diff
@@ -237,8 +237,18 @@ func TestClusterOperatorStatusRelatedObjects(t *testing.T) {
...
 		})
-		// This test runs before TestGatewayAPI, so we do *not* expect
-		// to see subscriptions, istios, or gateways in relatedObjects.
+		if gatewayAPIControllerEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPIController); err != nil {
+			t.Fatalf("Failed to look up %q featuregate: %v", features.FeatureGateGatewayAPIController, err)
+		} else if gatewayAPIControllerEnabled {
+			expected = append(expected, configv1.ObjectReference{
+				Group:     "gateway.networking.k8s.io",
+				Resource:  "gateways",
+				Namespace: "openshift-ingress",
```
candita, line 246: Should the namespace be blank here?

test/e2e/operator_test.go
```diff
@@ -237,8 +237,18 @@ func TestClusterOperatorStatusRelatedObjects(t *testing.T) {
 			Group:    "gateway.networking.k8s.io",
 			Resource: "gatewayclasses",
 		})
-		// This test runs before TestGatewayAPI, so we do *not* expect
-		// to see subscriptions, istios, or gateways in relatedObjects.
+		if gatewayAPIControllerEnabled, err := isFeatureGateEnabled(features.FeatureGateGatewayAPIController); err != nil {
```
candita, line 240: Based on the test output, gateways is never appended to expected.

test/e2e/operator_test.go
Miciah: Possibly.  My thinking was that we were telling people to create gateways in the "openshift-ingress" namespace, so the platform-managed ones should be there, we're allowing gateways in other namespaces as well (we just won't manage DNS for them).  Based on that, I suppose it makes sense to leave the namespace blank.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/c7db35d125f750f799fe98bf48a850089b1b09e2..4ca2f3ee05bcafa57afef7ab12e18e29f934db4b makes these changes:

- Update the status controller to add "gatewayclasses" to `relatedObjects` if, and only if, *both* the "GatewayAPI" *and* "GatewayAPIController" featuregates are enabled.
- Omit the namespace for gateways in `relatedObjects` as Istio manages gateways in any namespace.  
- Update `TestClusterOperatorStatusRelatedObjects` accordingly.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/ad9275e5917f18ad9739586e627ba7cb22c1fce2..1c68435a08d7a3adac8744082ed74146548b9f7c deletes an unnecessary check for the operand namespace's existence, which is unnecessary when we don't specify the operand namespace for a resource in `relatedObjects`.

pkg/operator/controller/status/controller.go
```diff
@@ -355,6 +404,37 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string,
 			}
 			state.unmanagedGatewayAPICRDNames = extension.UnmanagedGatewayAPICRDNames
 		}
+
+		if r.config.GatewayAPIControllerEnabled {
```
alebedev87, line 408: I'm wondering about the logic to put the check for `Gateway`/`GatewayClass` CRDs into this condition while `Subscription` check remains in the condition above (`GatewayAPI` featuregate). `Subscription` and `Istio` CRD can only be created when `GatewayAPIController` is enabled while `Gateway`/`GatewayClass` CRDs are created when `GatewayAPI` is enabled.

Miciah:
/hold
We'll merge this after branch cut.

rhamini3:
Valid related objects test is now passing with the proposed changes in [AWS Tech Preview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1217/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/1912295818073739264)
Marking bug as verified
/label qe-approved

pkg/operator/controller/status/controller.go
```diff
@@ -219,18 +252,25 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+			related = append(related, configv1.ObjectReference{
+				Group:    gatewayapiv1.GroupName,
+				Resource: "gatewayclasses",
+			})
+		}
+		if state.haveGatewaysResource {
+			related = append(related, configv1.ObjectReference{
+				Group:     gatewayapiv1.GroupName,
+				Resource:  "gateways",
+				Namespace: "", // Include all namespaces.
```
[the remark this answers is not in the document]
Miciah: > Does it mean that we include all resources with this group to the related objects (not CRD)? 

Yes.  The reference specifies the "gateways" resource but doesn't specify the namespace, so the reference includes any gateway in any namespace.  

> From one hand, [`PILOT_ENABLE_GATEWAY_CONTROLLER_MODE` description](https://github.com/openshift/cluster-ingress-operator/blob/86f19bd34a6f24e6836d4496fc17a13671379af0/pkg/operator/controller/gatewayclass/istio.go#L86-L90) says that gateway api resources are watched from all namespaces (btw I failed to find this variable in [downstream istio](https://github.com/openshift-service-mesh/istio/blob/748c90d41d4b04cb0aa3a1251dd24e07a1f27bfc/pilot/pkg/features/pilot.go)).

`PILOT_ENABLE_GATEWAY_CONTROLLER_MODE` is referenced here: https://github.com/openshift-service-mesh/sail-operator/blob/release-3.0/chart/samples/gwControllerMode.yaml#L14

However, I think that this variable is actually ignored by OSSM 3, as you say, as it has been superseded by EnhancedResourceScoping.  @dgn, is this correct?  Is the example Helm chart out of date?  

I intend to remove setting `PILOT_ENABLE_GATEWAY_CONTROLLER_MODE` from this operator with https://github.com/openshift/cluster-ingress-operator/pull/1227/commits/43caa071bd32d2285fbce6f4cac7b0cff2c7dd15.

> From the other hand [we don't manage DNS for gateways from namespaces other than `openshift-ingress`](https://github.com/openshift/cluster-ingress-operator/blob/86f19bd34a6f24e6836d4496fc17a13671379af0/pkg/operator/controller/gateway-service-dns/controller.go#L109-L114). Do we consider gateways with BYO DNS are managed by us?

Yes, Istio provisions a proxy even if cluster-ingress-operator doesn't publish a DNS record, so I consider such gateways to be managed by the OpenShift platform.

pkg/operator/controller/status/controller.go
Miciah: That's a fair point.  I address it with https://github.com/openshift/cluster-ingress-operator/compare/1c68435a08d7a3adac8744082ed74146548b9f7c..98b6246a365277d9feb157ece380d5b20e55332b by conditionalizing more logic on `GatewayAPIControllerEnabled`.  This also relates to (and conflicted with) the changes in #1232.

pkg/operator/controller/status/controller.go
```diff
@@ -345,32 +392,61 @@ func (r *reconciler) getOperatorState(ingressNamespace, canaryNamespace string,
...
+			} else {
+				state.haveGatewaysResource = true
+			}
+			if err := r.cache.Get(ctx, gatewayclassesResourceNamespacedName, &crd); err != nil {
+				if !errors.IsNotFound(err) {
+					return state, fmt.Errorf("failed to get CRD %q: %v", gatewayclassesResourceName, err)
+				}
+			} else {
+				state.haveGatewayclassesResource = true
+			}
```
alebedev87, line 442: `GatewayClass`es and `Gateway`s can be created even if OLM capabilities are not present. Should we move them up? Out of `if r.config.GatewayAPIControllerEnabled && r.config.MarketplaceEnabled && r.config.OperatorLifecycleManagerEnabled` condition.

Miciah: I've gone back and forth on this matter, and I think there are reasonable arguments either way.  In my most recent update, I have conditionalized adding gatewayclasses and gateways to `relatedObjects` with the condition `r.config.GatewayAPIControllerEnabled && r.config.MarketplaceEnabled && r.config.OperatorLifecycleManagerEnabled` on the grounds that we are only adding these resources to `relatedObjects` as part of, and in support of, the controller feature.  That is, if we didn't provide our own controller and were only providing the CRDs, then we wouldn't bother gathering gatewayclasses or gateways; we only need these resources in order to diagnose issues with the gateway controller.  

If the operator could specify gateways and gatewayclasses by controller name in the `relatedObjects` specification, I would consider doing that.  As it is, the operator would need to check the controller name on each gatewayclass and then check the gatewayclass name on each gateway in order to add each individual gatewayclass or gateway that our controller managed, and I don't think that that would be appropriate or worth the effort.

alebedev87: > That is, if we didn't provide our own controller and were only providing the CRDs, then we wouldn't bother gathering gatewayclasses or gateways; we only need these resources in order to diagnose issues with the gateway controller.

Ack.

PR #1222, opened 2025-04-18
CNTRLPLANE-112: Remove ARO HCP MIv2 Authentication for Ingress Operator

bryan-cox:
This PR:
- removes the ARO HCP MIv2 authentication method. ARO HCP only uses UserAssignedIdentityCredentials to authentication with Azure
- removes the filewatcher package as it is no longer needed; it was associated with ARO HCP managed identity version 2 authentication, which was removed from the ingress operator

This PR needs https://github.com/openshift/hypershift/pull/6058 merged first.


pkg/dns/azure/client/auth.go
```diff
@@ -3,36 +3,38 @@ package client
...
 		}
 	}
 
 	// Fallback to using client ID from env variable if not set.
 	if strings.TrimSpace(config.ClientID) == "" {
 		config.ClientID = os.Getenv("AZURE_CLIENT_ID")
-		if strings.TrimSpace(config.ClientID) == "" {
+
+		// client ID is not populated when managed Azure HCP is using UserAssignedIdentityCredentials
+		if strings.TrimSpace(config.ClientID) == "" && userAssignedIdentityCredentialsFilePath == "" {
```
Miciah, line 37: Besides removing the option to use `ARO_HCP_MI_CLIENT_ID`, this change is the only functional difference in the PR: ARO HCP can now omit `TenantID` and `ClientID` when using `UserAssignedIdentityCredentials`—right?

Miciah:
/approve
/lgtm

I want to double-check that I understand correctly: https://github.com/openshift/cluster-ingress-operator/pull/1222#discussion_r2054672579

/hold

If my understanding is correct, feel free to `/hold cancel`!

pkg/dns/azure/client/auth.go
bryan-cox: Yeah, both tenant and client IDs are within the UserAssignedIdentityCredentials object so they are extracted during the authentication method from the msi-dataplane package.

PR #1226, opened 2025-04-29
NE-2032: Refactor GatewayAPI DNS e2e

grzpiotrowski:
This PR refactors the Gateway API DNS helper functions to improve maintainability and aid in further enhancing the tests, refactoring particularly around the
[cluster-ingress-operator/test/e2e/util_gatewayapi_test.go: assertExpectedDNSRecords](https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/util_gatewayapi_test.go#L701-L777) function.

This is a follow up to comments in PR #1213 which added Gateway API DNS e2e tests.

### DIFF
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -701,79 +701,69 @@ type testListener struct {
 func assertExpectedDNSRecords(t *testing.T, expectations map[expectedDnsRecord]bool) error {
 	t.Helper()
 
-	var expectationsMet bool
-
-	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false, func(context context.Context) (bool, error) {
-		haveExpectNotPresent := false
-		// expectationsMet starts true and gets set to false when some expectation is not met.
-		expectationsMet = true
-
-		dnsRecords := &v1.DNSRecordList{}
-		if err := kclient.List(context, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
-			return false, fmt.Errorf("failed to list DNSRecords: %v", err)
-		}
-
-		// Iterate over all expectations.
-		for exp, shouldBePresent := range expectations {
-			if !shouldBePresent {
-				haveExpectNotPresent = true
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false,
+		func(ctx context.Context) (bool, error) {
+			dnsRecords := &v1.DNSRecordList{}
+			if err := kclient.List(ctx, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
+				return false, fmt.Errorf("failed to list DNSRecords: %v", err)
 			}
 
-			// Reset the found and foundReady flags for each expectation.
-			found := false
-			foundReady := false
-			// Look for a DNSRecord that matches the expected gateway and DNS name.
-			for _, record := range dnsRecords.Items {
-				if record.Labels["gateway.networking.k8s.io/gateway-name"] == exp.gatewayName &&
-					record.Spec.DNSName == exp.dnsName {
-
-					if !shouldBePresent {
-						expectationsMet = false
-						found = true
-						t.Logf("DNSRecord %q (%s) found but should not be present.", record.Name, exp.dnsName)
-						return false, nil
-					}
-					found = true
-					// DNSRecord found and should be present, check if it is published
-					for _, zone := range record.Status.Zones {
-						for _, condition := range zone.Conditions {
-							if condition.Type == v1.DNSRecordPublishedConditionType && condition.Status == string(metav1.ConditionTrue) {
-								t.Logf("Found DNSRecord %q (%s) %s=%s as expected", record.Name, exp.dnsName, condition.Type, condition.Status)
-								foundReady = true
-							}
-						}
-					}
-					if !foundReady {
-						t.Logf("Found DNSRecord %v but could not determine its readiness; retrying...", record.Name)
-						expectationsMet = false
-						return false, nil
-					}
+			for exp, shouldBePresent := range expectations {
+				expectationMet, err := checkDNSRecordExpectation(t, *dnsRecords, exp, shouldBePresent)
+				if err != nil {
+					return false, err
+				}
+				if !expectationMet {
+					return false, nil
 				}
 			}
 
-			// If the record is expected but not found, return false to continue polling.
-			if shouldBePresent && !found {
-				t.Logf("DNSRecord for hostname %q (gateway: %s) is expected to be present but was not found; retrying...", exp.dnsName, exp.gatewayName)
-				expectationsMet = false
-				return false, nil
-			}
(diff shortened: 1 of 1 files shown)


grzpiotrowski:
```
"Error executing test process"
```
Unrelated, I think.

/retest

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
-					if !foundReady {
-						t.Logf("Found DNSRecord %v but could not determine its readiness; retrying...", record.Name)
-						expectationsMet = false
-						return false, nil
-					}
+			for exp, shouldBePresent := range expectations {
+				expectationMet, err := checkDNSRecordExpectation(t, *dnsRecords, exp, shouldBePresent)
+				if err != nil {
+					return false, err
+				}
```
alebedev87, line 715: Does `checkDNSRecordExpectation` even return an error? I can only see `nil`s returned. If no error can be returned from `checkDNSRecordExpectation` then it has to return only 1 value (`bool`).

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
-
-		// Iterate over all expectations.
-		for exp, shouldBePresent := range expectations {
-			if !shouldBePresent {
-				haveExpectNotPresent = true
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 1*time.Minute, false,
+		func(ctx context.Context) (bool, error) {
+			dnsRecords := &v1.DNSRecordList{}
+			if err := kclient.List(ctx, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
+				return false, fmt.Errorf("failed to list DNSRecords: %v", err)
```
alebedev87, line 708: In most of the polling cases we log the error and retry the API operator. Returning an error would break the polling loop.
```suggestion
			if err := kclient.List(ctx, dnsRecords, client.InNamespace(operatorcontroller.DefaultOperandNamespace)); err != nil {
			        t.Logf("Failed to list DNSRecords: %v", err)
				return false, nil
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
-							}
-						}
-					}
-					if !foundReady {
-						t.Logf("Found DNSRecord %v but could not determine its readiness; retrying...", record.Name)
-						expectationsMet = false
-						return false, nil
-					}
+			for exp, shouldBePresent := range expectations {
+				expectationMet, err := checkDNSRecordExpectation(t, *dnsRecords, exp, shouldBePresent)
```
alebedev87, line 712: I think it's fine to pass by pointer to gain some CPU cycles on copying of all DNSRecords of the cluster. 
```suggestion
				expectationMet, err := checkDNSRecordExpectation(t, dnsRecords, exp, shouldBePresent)
```
Unfortunately, go doesn't have pointers to const (hi, C++!) to benefit from the pointer parameter and ensure no modification on the pointed object. But it's quite common to pass by pointer even in read-only cases.

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
+
+func checkDNSRecordExpectation(t *testing.T, dnsRecords v1.DNSRecordList, exp expectedDnsRecord, shouldBePresent bool) (bool, error) {
+	for _, rec := range dnsRecords.Items {
+		if !dnsRecordMatches(rec, exp) {
+			continue
 		}
-		if haveExpectNotPresent {
-			t.Logf("Continuing polling to ensure non-expected DNSRecords do not exist...")
+		if !shouldBePresent {
+			t.Logf("DNSRecord %s found but expected it absent.", exp.dnsName)
```
alebedev87, line 731:
```suggestion
			t.Logf("DNSRecord %q found but expected to be absent; retrying...", exp.dnsName)
```

alebedev87 (commented):
Some minor remarks. There was a remark in the previous PR [about the cleanup function](https://github.com/openshift/cluster-ingress-operator/pull/1213#discussion_r2040256393). Do you think you can consider it too?

test/e2e/util_gatewayapi_test.go
candita: Agree.

test/e2e/util_gatewayapi_test.go
candita: I would like to see the message that is logged here, be returned to the caller so the caller can figure out what to do with the message, instead of logging it here.  Some messages work great for polling, e.g.  "DNSRecord %s found but not yet published, retrying...", but don't make sense if you wanted to reuse this function in the future for a one-time call instead of repeated polling.

test/e2e/util_gatewayapi_test.go
candita: See https://github.com/openshift/cluster-ingress-operator/pull/1226/files#r2093564183 on returning this error instead of logging it.

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
+			return false, nil
 		}
-		return !haveExpectNotPresent, nil
-	})
+		if !isPublished(rec) {
+			t.Logf("DNSRecord %s found but not yet published, retrying...", exp.dnsName)
```
candita, line 735: See https://github.com/openshift/cluster-ingress-operator/pull/1226/files#r2093564183 on returning this error instead of logging it.

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
+			return false, nil
+		}
+		t.Logf("DNSRecord %s found & published as expected", exp.dnsName)
```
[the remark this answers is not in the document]
candita: See https://github.com/openshift/cluster-ingress-operator/pull/1226/files#r2093564183 on returning this error instead of logging it.

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
+		return true, nil
+	}
 
-	if !expectationsMet {
-		return fmt.Errorf("failed to observe expected DNSRecords: %v", err)
+	if shouldBePresent {
+		t.Logf("DNSRecord %s not found, retrying...", exp.dnsName)
```
candita, line 743: See https://github.com/openshift/cluster-ingress-operator/pull/1226/files#r2093564183 on returning this error instead of logging it.

test/e2e/util_gatewayapi_test.go
```diff
@@ -701,79 +701,69 @@ type testListener struct {
...
+		return false, nil
 	}
-	return nil
+	t.Logf("DNSRecord %s correctly absent", exp.dnsName)
```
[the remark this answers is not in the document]
candita: See https://github.com/openshift/cluster-ingress-operator/pull/1226/files#r2093564183 on returning this error instead of logging it.

candita (commented):
Looks much better, thanks.  Just a few nits.

PR #1125, opened 2024-08-09
OCPBUGS-37491: Ingress operator status not degraded when canary route fails

candita:
The status was not being set to degraded because the lastTransitionTime was being set unnecessarily. It should only be set when the status changes.  See https://github.com/openshift/cluster-dns-operator/pull/375

The fix involves recognizing when the condition status changes, and only updating the transition time when this happens.

* pkg/operator/controller/condition.go - export the ConditionChanged func
  for use by all previous callers.
* pkg/operator/controller/dns/controller.go - use the new ConditionChanged,
  update logic for setting condition transition time.
* pkg/operator/controller/ingress/status.go - use the new ConditionChanged,
  update logic for setting condition transition time.
* pkg/operator/controller/status/controller.go - use the new ConditionChanged,
  update logic for setting condition transition time.

PR #1229, opened 2025-05-07
OCPBUGS-31521: Don't publish duplicate DNS records

rfredette:
Before attempting to publish a domain to a zone, check if that domain is already being published to the same zone.

### DIFF
--- a/pkg/operator/controller/dns/controller.go
+++ b/pkg/operator/controller/dns/controller.go
@@ -62,6 +62,8 @@ const (
 	kubeCloudConfigName = "kube-cloud-config"
 	// cloudCABundleKey is the key in the kube cloud config ConfigMap where the custom CA bundle is located
 	cloudCABundleKey = "ca-bundle.pem"
+	// dnsRecordIndexFieldName is the key for the DNSRecord index, used to identify conflicting domain names.
+	dnsRecordIndexFieldName = "Spec.DNSName"
 )
 
 var log = logf.Logger.WithName(controllerName)
@@ -79,6 +81,19 @@ func New(mgr manager.Manager, config Config) (runtimecontroller.Controller, erro
 		return nil, err
 	}
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &iov1.DNSRecord{}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})); err != nil {
+		return nil, err
+	}
+	// When a DNS record is deleted, there may be a conflicting record that is
+	// waiting to be published. We need a different map function than usual,
+	// which uses the domain name from the deleted record to find the oldest
+	// record with the same domain name (if any exist), and queue that record
+	// for reconciliation.
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &iov1.DNSRecord{}, handler.EnqueueRequestsFromMapFunc(reconciler.mapOnRecordDelete), predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	})); err != nil {
 		return nil, err
 	}
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.DNS{}, handler.EnqueueRequestsFromMapFunc(reconciler.ToDNSRecords))); err != nil {
@@ -102,7 +117,15 @@ func New(mgr manager.Manager, config Config) (runtimecontroller.Controller, erro
 	})); err != nil {
 		return nil, err
 	}
+	if err := operatorCache.IndexField(context.Background(), &iov1.DNSRecord{}, dnsRecordIndexFieldName, indexDNSRecords); err != nil {
+		return nil, err
+	}
 	return c, nil
+}
+
+func indexDNSRecords(o client.Object) []string {
+	dnsRecord := o.(*iov1.DNSRecord)
+	return []string{dnsRecord.Spec.DNSName}
 }
 
 // Config holds all the things necessary for the controller to run.
@@ -178,7 +201,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	if dnsConfig.Spec.PublicZone != nil {
 		zones = append(zones, *dnsConfig.Spec.PublicZone)
 	}
-	requeue, statuses := r.publishRecordToZones(zones, record)
+	requeue, statuses := r.publishRecordToZones(ctx, zones, record)
 
 	// Requeue if publishing records failed.
 	result := reconcile.Result{}
@@ -318,7 +341,7 @@ func (r *reconciler) publishRecord(zone configv1.DNSZone, record *iov1.DNSRecord
 
 // publishRecordToZones attempts to publish records and returns a bool
 // indicating if we need to requeue due to errors and list of latest DNS Zone status.
-func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1.DNSRecord) (bool, []iov1.DNSZoneStatus) {
+func (r *reconciler) publishRecordToZones(ctx context.Context, zones []configv1.DNSZone, record *iov1.DNSRecord) (bool, []iov1.DNSZoneStatus) {
 	var statuses []iov1.DNSZoneStatus
 	var requeue bool
 	dnsPolicy := record.Spec.DNSManagementPolicy
@@ -333,27 +356,43 @@ func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1
 			continue
 		}
 
-		var err error
 		var condition iov1.DNSZoneCondition
 		if dnsPolicy == iov1.UnmanagedDNS {
-			log.Info("DNS record not published", "record", record.Spec)
+			log.Info("DNS record not published: DNS management policy is unmanaged", "record", record.Spec)
 			condition = iov1.DNSZoneCondition{
-				Message:            "DNS record is currently not being managed by the operator",
-				Reason:             "UnmanagedDNS",
-				Status:             string(operatorv1.ConditionUnknown),
-				Type:               iov1.DNSRecordPublishedConditionType,
-				LastTransitionTime: metav1.Now(),
+				Message: "DNS record is currently not being managed by the operator",
+				Reason:  "UnmanagedDNS",
+				Status:  string(operatorv1.ConditionUnknown),
+				Type:    iov1.DNSRecordPublishedConditionType,
 			}
 		} else if isRecordPublished {
+			var err error
 			condition, err = r.replacePublishedRecord(zones[i], record)
+			if err != nil {
+				requeue = true
+			}
+		} else if isOldestRecord, err := isOldestRecordForDomain(ctx, r.cache, record); err != nil {
+			log.Error(err, "failed to validate DNS record", "record", record.Spec)
+			condition = iov1.DNSZoneCondition{
+				Message: "Pre-publish validation failed",
+				Reason:  "InternalError",
+				Status:  string(operatorv1.ConditionFalse),
+				Type:    iov1.DNSRecordPublishedConditionType,
+			}
+			requeue = true
+		} else if !isOldestRecord {
+			log.Info("DNS record not published: domain name already used by another DNS record", "record", record.Spec)
+			condition = iov1.DNSZoneCondition{
+				Message: "Domain name is already in use",
+				Reason:  "DomainAlreadyInUse",
+				Status:  string(operatorv1.ConditionFalse),
+				Type:    iov1.DNSRecordPublishedConditionType,
+			}
 		} else {
 			condition, err = r.publishRecord(zones[i], record)
-		}
-
-		// Check if replacing or publishing record resulted in an error.
-		// If it did, re-enqueue the request for processing.
-		if err != nil {
-			requeue = true
+			if err != nil {
+				requeue = true
+			}
 		}
 
 		statuses = append(statuses, iov1.DNSZoneStatus{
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/dns/controller.go
```diff
@@ -846,3 +873,32 @@ func getIbmDNSProvider(dnsConfig *configv1.DNS, creds *corev1.Secret, instanceCR
...
+	}
+
+	if len(records.Items) == 0 {
+		return false, nil
+	}
+
+	for _, existingRecord := range records.Items {
+		// we only care if the domain name is published by a different record, so ignore the matching record if it
+		// already exists.
+		// TODO: There's got to be a better way to match the same object
```
Miciah, line 892: Compare by `UID` instead of name.

pkg/operator/controller/dns/controller.go
```diff
@@ -352,6 +361,24 @@ func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1
 			}
 		} else if isRecordPublished {
 			condition, err = r.replacePublishedRecord(zones[i], record)
+		} else if isDomainPublished, err = domainIsAlreadyPublishedInZone(context.Background(), r.cache, record, &zones[i]); err != nil {
```
Miciah, line 364: You don't need to declare `isDomainPublished` outside the `else if` clause.  Better to keep `isDomainPublished` and `err` scoped to the `else if` clauses.  
```suggestion
		} else if isDomainPublished, err := domainIsAlreadyPublishedInZone(context.Background(), r.cache, record, &zones[i]); err != nil {
```

pkg/operator/controller/dns/controller.go
```diff
@@ -352,6 +361,24 @@ func (r *reconciler) publishRecordToZones(zones []configv1.DNSZone, record *iov1
...
+				LastTransitionTime: metav1.Now(),
+			}
+		} else if isDomainPublished {
+			log.Info("DNS record not published: domain name already used by another DNS record", "record", record.Spec)
+			condition = iov1.DNSZoneCondition{
+				Message:            "Domain name is already in use",
+				Reason:             "DomainAlreadyInUse",
+				Status:             string(operatorv1.ConditionUnknown),
+				Type:               iov1.DNSRecordPublishedConditionType,
+				LastTransitionTime: metav1.Now(),
```
Miciah, line 380: I believe `mergeConditions` will set `LastTransitionTime`, so you don't need to set it here (lines 371 and 380).

Miciah: https://github.com/openshift/cluster-ingress-operator/pull/1125/commits/523461a512528f13675de1204e3d6ad381974dbd updated `mergeConditions` to set `LastTransitionTime`.  That's a fairly recent change, so older code (e.g. line 360) might be setting `LastTransitionTime` because it used to be necessary but no longer is.

pkg/operator/controller/dns/controller.go
```diff
@@ -580,6 +646,37 @@ func (r *reconciler) ToDNSRecords(ctx context.Context, o client.Object) []reconc
 	return requests
 }
 
+func (r *reconciler) MapOnRecordDelete(ctx context.Context, o client.Object) []reconcile.Request {
+	deletedRecord, ok := o.(*iov1.DNSRecord)
+	if !ok {
+		log.Info("failed to read DNS record")
```
Miciah, line 652:
```suggestion
		log.Infof("Got unexpected object; expected type DNSRecord, got type %T", o)
```

pkg/operator/controller/dns/controller.go
```diff
@@ -81,6 +83,16 @@ func New(mgr manager.Manager, config Config) (runtimecontroller.Controller, erro
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &iov1.DNSRecord{}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})); err != nil {
 		return nil, err
 	}
+	// When a DNS record is deleted, there may be a conflicting record that should be published. Watch exclusively for
+	// deletes, and queue a reconcile request for the appropriate conflicting record, if applicable.
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &iov1.DNSRecord{}, handler.EnqueueRequestsFromMapFunc(reconciler.MapOnRecordDelete), predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return true },
```
Miciah, line 90: Add a comment with your findings that the delete event happens when the object is actually deleted, not when it is merely marked for deletion (that is, when `deletionTimestamp` is set).

pkg/operator/controller/dns/controller.go
rfredette: This is still a relevant comment, but I moved the function, which made github think it's outdated. I'll make this change in my next update

pkg/operator/controller/dns/controller.go
Miciah: Or something more like this:
```suggestion
		log.Error(nil, "Got unexpected type of object", "expected", "DNSRecord", "actual", fmt.Sprintf("%T", o))
```

pkg/operator/controller/dns/controller.go
```diff
@@ -580,6 +646,37 @@ func (r *reconciler) ToDNSRecords(ctx context.Context, o client.Object) []reconc
 	return requests
 }
 
+func (r *reconciler) MapOnRecordDelete(ctx context.Context, o client.Object) []reconcile.Request {
```
Miciah, line 649:
```suggestion
func (r *reconciler) mapOnRecordDelete(ctx context.Context, o client.Object) []reconcile.Request {
```

rfredette, line 649: Needs a godoc comment

pkg/operator/controller/dns/controller_test.go
```diff
@@ -840,3 +847,26 @@ func Test_customCABundle(t *testing.T) {
 		})
 	}
 }
+
+func buildFakeCache(t *testing.T) *testutil.FakeCache {
```
rfredette, line 851: Also could use a godoc comment

pkg/operator/controller/dns/controller_test.go
```diff
@@ -128,9 +132,9 @@ func Test_publishRecordToZones(t *testing.T) {
 	}
 }
 
-// TestPublishRecordToZonesMergesStatus verifies that publishRecordToZones
+// Test_publishRecordToZonesMergesStatus verifies that publishRecordToZones
 // correctly merges status updates.
-func TestPublishRecordToZonesMergesStatus(t *testing.T) {
+func Test_publishRecordToZonesMergesStatus(t *testing.T) {
```
Miciah, line 137: `TestPublishRecordToZonesMergesStatus` is an appropriate for the test as there is no `publishRecordToZonesMergesStatus` function.

pkg/operator/controller/dns/controller.go
```diff
@@ -390,6 +427,35 @@ func recordIsAlreadyPublishedToZone(record *iov1.DNSRecord, zoneToPublish *confi
...
+	}
+
+	if len(records.Items) == 0 {
+		return false, nil
+	}
+
+	for _, existingRecord := range records.Items {
+		// we only care if the domain name is published by a different record, so ignore the matching record if it
+		// already exists.
+		// TODO: There's got to be a better way to match the same object
```
rikatz, line 445: You can match the UID here. I would say that during admission webhooks matching UID is a bad idea, as the UID will be added once persisted by apiserver, but in case of a reconciliation process it means that the UID was already generated and cannot be changed anymore.

pkg/operator/controller/dns/controller_test.go
```diff
@@ -840,3 +847,26 @@ func Test_customCABundle(t *testing.T) {
...
 	}
 }
+
+func buildFakeCache(t *testing.T) *testutil.FakeCache {
+	t.Helper()
+	scheme := runtime.NewScheme()
+	iov1.AddToScheme(scheme)
+	fakeClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithIndex(&iov1.DNSRecord{}, dnsRecordIndexFieldName, func(o client.Object) []string {
```
rikatz, line 857: nit (no need to fix now!): do we care about making this indexer function some sort of utils/specific function that can be used both on the `operatorCache.IndexField` and on fakeCache to keep consistency?

Miciah: Do you mean a util function that the controller logic and test logic would share?  That does make sense, though I do caution against re-using controller logic in tests if doing so could mask a defect in the controller logic.

pkg/operator/controller/dns/controller.go
```diff
@@ -580,6 +646,37 @@ func (r *reconciler) ToDNSRecords(ctx context.Context, o client.Object) []reconc
...
+		log.Error(err, "failed to list DNS records")
+		return []reconcile.Request{}
+	}
+	if len(otherRecords.Items) == 0 {
+		// Nothing to do.
+		return []reconcile.Request{}
+	}
+	oldestExistingRecord := iov1.DNSRecord{}
+	for _, existingRecord := range otherRecords.Items {
+		if oldestExistingRecord.CreationTimestamp.IsZero() || existingRecord.CreationTimestamp.Before(&oldestExistingRecord.CreationTimestamp) {
```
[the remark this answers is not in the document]
Miciah: That seems reasonable.  On the one hand, as noted in https://github.com/openshift/cluster-ingress-operator/pull/1229#discussion_r2091816974, the reconciliation triggers when the DNSRecord CR is actually deleted, not merely marked for deletion.  On the other hand, we could be in a situation where multiple dnsrecords are being deleted at the same time, in which case your suggestion might avoid some unnecessary reconciliations.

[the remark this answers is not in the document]
rikatz: yeap, I should have been more explicit but my idea here is to ignore any other DNS record that would be considered older but also being on the deletion queue.

candita:
E2E test failure seen on this, reported by @rikatz in https://issues.redhat.com/browse/OCPBUGS-64675.

>      --- FAIL: TestAll/serial/TestGatewayAPI/testGatewayAPIDNS (341.29s)
>                 --- PASS: TestAll/serial/TestGatewayAPI/testGatewayAPIDNS/multipleGatewaysSameListenerHostname (161.16s)
>                 --- FAIL: TestAll/serial/TestGatewayAPI/testGatewayAPIDNS/gatewayListenersWithOverlappingHostname (180.09s)
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1e2adcf]
>

rikatz:
/cc

will review again soon!

rikatz:
@coderabbitai given it has been too long since you reviewed this PR, I want  you to review again. You must report any critical or major finding, you must do a review line by line. Please report any possible inconsistency or behavior that can cause disruption, corruption. Report as much findings as possible, do not leave any review to be verified on a future commit, I want you to do a single report with EVERYTHING that must be fixed on a single review shot

pkg/operator/controller/dns/controller_test.go
```diff
@@ -117,9 +123,10 @@ func Test_publishRecordToZones(t *testing.T) {
 			r := &reconciler{
 				// TODO To write a fake provider that can return errors and add more test cases.
 				dnsProvider: &dns.FakeProvider{},
+				cache:       buildFakeCache(t, nil),
 			}
 
-			_, actual := r.publishRecordToZones(test.zones, record)
+			_, actual := r.publishRecordToZones(context.TODO(), test.zones, record)
```
rikatz, line 129: followup: could use t.Context

rikatz:
@rfredette overall lgtm. If possible I am missing some e2e test for this case, that's the only missing item before we can merge.

rikatz:
tested this on my cluster, and it is working correctly and as expected.

Creating two gateways with duplicated names will fail (on the newer one) for DNSRecord ensuring, given it is duplicated. 

```
  - conditions:
    - message: Domain name is already in use
      reason: DomainAlreadyInUse
      status: "False"
      type: Published
    dnsZone:
      tags:
        Name: ci-ln-ybj1b62-76ef8-wzlgc-int
        kubernetes.io/cluster/ci-ln-ybj1b62-76ef8-wzlgc: owned
  - conditions:
    - message: Domain name is already in use
      reason: DomainAlreadyInUse
      status: "False"
      type: Published
    dnsZone:
      id: Z0764857NINJVO7A6ST2
```

Deleting the old Gateway will make this progress. The gateway also contains the proper DNSReady condition:

```
  - lastTransitionTime: "2026-07-29T20:45:51Z"
    message: One or more listeners have DNS provisioning issues
    observedGeneration: 1
    reason: SomeListenersNotReady
    status: "False"
    type: DNSReady
listeners:
    - lastTransitionTime: "2026-07-29T20:45:51Z"
      message: 'The record failed to provision in some zones: [{ map[Name:ci-ln-ybj1b62-76ef8-wzlgc-int
        kubernetes.io/cluster/ci-ln-ybj1b62-76ef8-wzlgc:owned]} {Z0764857NINJVO7A6ST2
        map[]}]'
      observedGeneration: 1
      reason: FailedZones
      status: "False"
      type: DNSReady
```

melvinjoseph86:
marking bug as verified, since it is fixed through pre-merge testing

Creating two gateways with duplicated names will fail (on the newer one) for DNSRecord ensuring, given it is duplicated.
```

➜  oc -n openshift-ingress get gateway
NAME      CLASS               ADDRESS                                                                   PROGRAMMED   AGE
gateway   openshift-default   aec4b006bd6b645e68c9032c1ce86ea5-1647689892.us-east-2.elb.amazonaws.com   True         10s

➜  oc get dnsrecord -A       
NAMESPACE                    NAME                          AGE
openshift-ingress-operator   default-wildcard              72m
openshift-ingress            gateway-7988756fb8-wildcard   2m7s

  spec:
    dnsManagementPolicy: Managed
    dnsName: '*.apps.ci-ln-23sy3xt-76ef8.aws-4.ci.openshift.org.'
    recordTTL: 30
    recordType: CNAME
    targets:
    - aec4b006bd6b645e68c9032c1ce86ea5-1647689892.us-east-2.elb.amazonaws.com
  status:
    observedGeneration: 1
    zones:
    - conditions:
      - message: Domain name is already in use
        reason: DomainAlreadyInUse
        status: "False"
        type: Published
      dnsZone:
        tags:
          Name: ci-ln-23sy3xt-76ef8-4bclz-int
          kubernetes.io/cluster/ci-ln-23sy3xt-76ef8-4bclz: owned
    - conditions:
      - message: Domain name is already in use
        reason: DomainAlreadyInUse
        status: "False"
        type: Published
      dnsZone:
        id: Z0764857NINJVO7A6ST2
kind: List
metadata:
  resourceVersion: ""
```

 ```
       message: 'The record failed to provision in some zones: [{ map[Name:ci-ln-23sy3xt-76ef8-4bclz-int
          kubernetes.io/cluster/ci-ln-23sy3xt-76ef8-4bclz:owned]} {Z0764857NINJVO7A6ST2
          map[]}]'
        observedGeneration: 1
        reason: FailedZones
        status: "False"
        type: DNSReady
      name: http
      supportedKinds:
      - group: gateway.networking.k8s.io
        kind: HTTPRoute
      - group: gateway.networking.k8s.io
        kind: GRPCRoute
```

    Delete the gateway and confirm that upstream dnsrecord and routes are not affected
```

➜  oc -n openshift-ingress delete gateway gateway          
gateway.gateway.networking.k8s.io "gateway" deleted
➜  oc get dnsrecord -A                                     
NAMESPACE                    NAME                          AGE
openshift-ingress-operator   default-wildcard              66m
openshift-ingress            gateway-7988756fb8-wildcard   2m5s
➜  curl -I oauth-openshift.apps.ci-ln-23sy3xt-76ef8.aws-4.ci.openshift.org
HTTP/1.1 302 Found
content-length: 0
location: https://oauth-openshift.apps.ci-ln-23sy3xt-76ef8.aws-4.ci.openshift.org/
cache-control: no-cache

➜  curl -I canary-openshift-ingress-canary.apps.ci-ln-23sy3xt-76ef8.aws-4.ci.openshift.org
HTTP/1.1 302 Found
content-length: 0
location: https://canary-openshift-ingress-canary.apps.ci-ln-23sy3xt-76ef8.aws-4.ci.openshift.org/
cache-control: no-cache
```

/verified by @mjoseph

rikatz:
/lgtm
/approve
/hold

I want the pipeline to run, and I want to check with @rfredette if he will be able to do the e2e here, otherwise I will open a followup OCPBUGS to the e2e tests to be written later

rikatz:
/hold cancel
Spoke with Mike, we need this fixed and running. I will open a followup story for us (or a bug) to address the comment on adding e2e test

rikatz:
opened https://redhat.atlassian.net/browse/OCPBUGS-100424 to add the e2e tests

PR #1246, opened 2025-07-29
NE-2108: Add OSSM channel and version override annotations

Miciah:
Define annotations for the gatewayclass by which the channel and version for the OSSM subscription can be overridden.

When creating the subscription, the operator now checks the following annotations:

    unsupported.do-not-use.openshift.io/ossm-channel
    unsupported.do-not-use.openshift.io/ossm-version

One or both annotations may be specified to override the channel or version from the config.

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -130,6 +130,8 @@ type Config struct {
 	GatewayAPIOperatorChannel string
 	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
 	GatewayAPIOperatorVersion string
+	// IstioVersion is the version of Istio to configure on the Istio CR.
+	IstioVersion string
 }
 
 // reconciler reconciles gatewayclasses.
@@ -206,7 +208,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx); err != nil {
 		errs = append(errs, err)
 	}
-	if _, _, err := r.ensureIstio(ctx, &gatewayclass); err != nil {
+	if _, _, err := r.ensureIstio(ctx, &gatewayclass, r.config.IstioVersion); err != nil {
 		errs = append(errs, err)
 	} else {
 		// The OSSM operator installs the istios.sailoperator.io CRD.
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -126,6 +126,8 @@ type Config struct {
 	OperatorNamespace string
 	// OperandNamespace is the namespace in which Istio should be deployed.
 	OperandNamespace string
+	// GatewayAPIOperatorCatalog is the catalog source to use to install the Gateway API implementation.
+	GatewayAPIOperatorCatalog string
 	// GatewayAPIOperatorChannel is the release channel of the Gateway API implementation to install.
 	GatewayAPIOperatorChannel string
 	// GatewayAPIOperatorVersion is the name and release of the Gateway API implementation to install.
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -39,6 +39,38 @@ const (
 	// inferencepoolExperimentalCrdName is the name of the experimental
 	// (alpha version) InferencePool CRD.
 	inferencepoolExperimentalCrdName = "inferencepools.inference.networking.x-k8s.io"
+
+	// subscriptionCatalogOverrideAnnotationKey is the key for an
+	// unsupported annotation on the gatewayclass using which a custom
+	// catalog source can be specified for the OSSM subscription.  This
+	// annotation is only intended for use by OpenShift developers.  Note
+	// that this annotation is intended to be used only when initially
+	// creating the gatewayclass and subscription; changing the catalog
+	// source on an existing subscription will likely have no effect or
+	// cause errors.
+	subscriptionCatalogOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/ossm-catalog"
+	// subscriptionChannelOverrideAnnotationKey is the key for an
+	// unsupported annotation on the gatewayclass using which a custom
+	// channel can be specified for the OSSM subscription.  This annotation
+	// is only intended for use by OpenShift developers.  Note that this
+	// annotation is intended to be used only when initially creating the
+	// gatewayclass and subscription; changing the channel on an existing
+	// subscription will likely have no effect or cause errors.
+	subscriptionChannelOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/ossm-channel"
+	// subscriptionVersionOverrideAnnotationKey is the key for an
+	// unsupported annotation on the gatewayclass using which a custom
+	// version of OSSM can be specified.  This annotation is only intended
+	// for use by OpenShift developers.  Note that this annotation is
+	// intended to be used only when initially creating the gatewayclass and
+	// subscription; OLM will not allow downgrades, and upgrades are
+	// generally restricted to the next version after the currently
+	// installed version.
+	subscriptionVersionOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/ossm-version"
+	// istioVersionOverrideAnnotationKey is the key for an unsupported
+	// annotation on the gatewayclass using which a custom version of Istio
+	// can be specified.  This annotation is only intended for use by
+	// OpenShift developers.
+	istioVersionOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/istio-version"
 )
 
 var log = logf.Logger.WithName(controllerName)
@@ -204,13 +236,29 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	}
 
 	var errs []error
-	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
+	ossmCatalog := r.config.GatewayAPIOperatorCatalog
+	if v, ok := gatewayclass.Annotations[subscriptionCatalogOverrideAnnotationKey]; ok {
+		ossmCatalog = v
+	}
+	ossmChannel := r.config.GatewayAPIOperatorChannel
+	if v, ok := gatewayclass.Annotations[subscriptionChannelOverrideAnnotationKey]; ok {
+		ossmChannel = v
+	}
+	ossmVersion := r.config.GatewayAPIOperatorVersion
+	if v, ok := gatewayclass.Annotations[subscriptionVersionOverrideAnnotationKey]; ok {
+		ossmVersion = v
+	}
+	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion); err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
 	}
-	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx); err != nil {
+	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
 		errs = append(errs, err)
 	}
-	if _, _, err := r.ensureIstio(ctx, &gatewayclass, r.config.IstioVersion); err != nil {
+	istioVersion := r.config.IstioVersion
+	if v, ok := gatewayclass.Annotations[istioVersionOverrideAnnotationKey]; ok {
+		istioVersion = v
+	}
+	if _, _, err := r.ensureIstio(ctx, &gatewayclass, istioVersion); err != nil {
 		errs = append(errs, err)
 	} else {
 		// The OSSM operator installs the istios.sailoperator.io CRD.
--- a/pkg/operator/controller/gatewayclass/controller_test.go
+++ b/pkg/operator/controller/gatewayclass/controller_test.go
@@ -1,0 +1,279 @@
+package gatewayclass
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	sailv1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	configv1 "github.com/openshift/api/config/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/stretchr/testify/assert"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache/informertest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	testutil "github.com/openshift/cluster-ingress-operator/pkg/operator/controller/test/util"
+)
+
+func Test_Reconcile(t *testing.T) {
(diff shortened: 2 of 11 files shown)


pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -178,7 +187,15 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	var errs []error
-	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx); err != nil {
+	channel := r.config.GatewayAPIOperatorChannel
+	if v, ok := gatewayclass.Annotations[subscriptionChannelOverrideAnnotationKey]; ok {
+		channel = v
+	}
+	version := r.config.GatewayAPIOperatorVersion
+	if v, ok := gatewayclass.Annotations[subscriptionVersionOverrideAnnotationKey]; ok {
+		version = v
+	}
```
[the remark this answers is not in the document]
Miciah: Mutating the config doesn't seem right to me, for a few reasons:

- The annotation could be set and then removed.  The behavior gets confusing if the annotations are subsequently removed.  Do we keep using the override value even after the annotation has been removed, or do we need to update the config again to restore the original value?
- The operator initialization produces the config, and the controller consumes it.  If the controller then mutates the config, that seems surprising to me.  

Your question did prompt me to wonder what happens if you have two gatewayclasses with different annotations, and that raises the question of whether the gatewayclass is the right place for these annotations.  Unfortunately, I don't have a better idea.

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -31,6 +31,15 @@ import (
...
 	controllerName = "gatewayclass_controller"
+
+	// subscriptionChannelOverrideAnnotationKey is the key for an
+	// unsupported annotation on the gatewayclass using which a custom
+	// channel of OSSM can be specified.
+	subscriptionChannelOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/ossm-channel"
+	// subscriptionVersionOverrideAnnotationKey is the key for an
+	// unsupported annotation on the gatewayclass using which a custom
+	// version of OSSM can be specified.
+	subscriptionVersionOverrideAnnotationKey = "unsupported.do-not-use.openshift.io/ossm-version"
```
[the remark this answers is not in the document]
alebedev87: > Once [NE-2097](https://issues.redhat.com/browse/NE-2097) is implemented, does that limitation go away?

Partially, but some version changes will remain impossible to achieve. Like downgrades, or upgrade to versions which are skipped from the upgrade graph (e.g. `3.0.0` -> `3.0.2`) or versions which are beyond the latest version (e.g. `3.5.0`). I'm trying to cover those cases in the operator status (`Progressing`, `Degraded`) but I'm seeing OLM limitations which prevent me from detecting some situations. Here we let the user set the version, we can expect more unexpected inputs comparing to us (NI&D) hard coding the "known to exist" versions.

I think that, at minimum, we should leave some instructions about which input is expected here:
1. No downgrades are possible.
2. Only existing versions are expected.
3. No upgrades greater than N+1 is possible.
The #3 only makes sense until https://issues.redhat.com/browse/NE-2097 is implemented.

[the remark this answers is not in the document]
Miciah: > > Once [NE-2097](https://issues.redhat.com/browse/NE-2097) is implemented, does that limitation go away?
> 
> Partially, but some version changes will remain impossible to achieve - such as downgrades, upgrades that skip versions in the upgrade graph (e.g., `3.0.0` -> `3.0.2`), or upgrades to versions beyond the latest available (e.g., `3.5.0`). I'm trying to handle these cases using the operator status (`Progressing`, `Degraded`), but there are OLM limitations that prevent reliable detection in some situations.

Understood.  

> Since we allow users to set the version manually, we must anticipate more unexpected inputs compared to when we (NI&D) hard-code known, valid versions.

The annotation is explicitly unsupported.  If we get a ticket for an issue that results from using any of these annotations, we should immediately close the ticket.  

> At a minimum, we should provide clear instructions to users about the expected inputs:
> 
>     1. Downgrades are not supported.
> 
>     2. Only existing, shipped versions should be used.
> 
>     3. Upgrades to versions greater than N+1 are not supported.
> 
> 
> This third point applies until [NE-2097](https://issues.redhat.com/browse/NE-2097) is implemented.

None of these are supported.  The only people who should use or even know about these annotations are OpenShift Networking developers and OpenShift AI developers.  However, I can add a code comment with this information.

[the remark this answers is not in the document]
Miciah: > Only existing, shipped versions should be used.

I don't believe that this is true; the annotation could be useful to test unreleased versions.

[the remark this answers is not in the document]
alebedev87: > > Only existing, shipped versions should be used.

> I don't believe that this is true; the annotation could be useful to test unreleased versions.

Agree "shipped" is not required.

pkg/operator/controller/gatewayclass/controller_test.go
```diff
@@ -0,0 +1,222 @@
...
+			name:    "Gatewayclass with OSSM and Istio overrides",
+			request: req("openshift-default"),
+			existingObjects: []runtime.Object{
+				&gatewayapiv1.GatewayClass{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{
+							"unsupported.do-not-use.openshift.io/ossm-catalog":  "foo",
+							"unsupported.do-not-use.openshift.io/ossm-channel":  "bar",
+							"unsupported.do-not-use.openshift.io/ossm-version":  "baz",
+							"unsupported.do-not-use.openshift.io/istio-version": "quux",
```
alebedev87, line 139: Did you intent to showcase that it's possible to set any values for those annotations? Or it was just to gain some time? If it's latter, may be it would make sense to set them to values which look more real.

pkg/operator/controller/gatewayclass/controller_test.go
```diff
@@ -0,0 +1,222 @@
...
+				assert.Contains(t, err.Error(), tc.expectError)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, reconcile.Result{}, res)
+			cmpOpts := []cmp.Option{
+				cmpopts.EquateEmpty(),
+				cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "OwnerReferences"),
+				cmpopts.IgnoreFields(metav1.TypeMeta{}, "Kind", "APIVersion"),
+				cmpopts.IgnoreFields(apiextensionsv1.CustomResourceDefinition{}, "Spec"),
```
alebedev87, line 207: I don't see where we use CRDs in this test.

alebedev87 (commented):
Overall LGTM, just 2 small remarks.

pkg/operator/controller/gatewayclass/controller_test.go
Miciah: I could come up with something for the OSSM version and Istio version, but I don't know what would be a realistic value for the catalog or channel other than the default values, and I didn't think it mattered for the test case.  What would be some more real looking values?  Do you just want an example that has a similar length and uses similar characters to the default values?

Miciah:
I tested the latest changes using a Cluster Bot cluster.  First, I created a subscription with catalog "foo", channel "bar", and version "baz"; cluster-ingress-operator created the subscription with those parameters, and the subscription failed to install (as expected).
<details>

```text
% oc create -f -<<'EOF'
apiVersion: gateway.networking.k8s.io/v1beta1
kind: GatewayClass
metadata:
  name: openshift-default
  annotations:
    unsupported.do-not-use.openshift.io/ossm-catalog: foo
    unsupported.do-not-use.openshift.io/ossm-channel: bar
    unsupported.do-not-use.openshift.io/ossm-version: baz
    unsupported.do-not-use.openshift.io/istio-version: quux
spec:
  controllerName: openshift.io/gateway-controller/v1
EOF
gatewayclass.gateway.networking.k8s.io/openshift-default created
% oc -n openshift-operators get subscriptions/servicemeshoperator3 -o yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  annotations:
    ingress.operator.openshift.io/owned: ""
  creationTimestamp: "2025-07-31T20:04:09Z"
  generation: 1
  labels:
    operators.coreos.com/servicemeshoperator3.openshift-operators: ""
  name: servicemeshoperator3
  namespace: openshift-operators
  resourceVersion: "32150"
  uid: 9059c142-5e6c-4b1f-a748-ef21e40138bd
spec:
  channel: bar
  config:
    annotations:
      openshift.io/required-scc: restricted-v2
      target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
    resources:
      requests:
        cpu: 10m
        memory: 64Mi
  installPlanApproval: Manual
  name: servicemeshoperator3
  source: foo
  sourceNamespace: openshift-marketplace
  startingCSV: baz
status:
  catalogHealth:
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: certified-operators
      namespace: openshift-marketplace
      resourceVersion: "31048"
      uid: b020a0e9-85a4-41f9-a27f-0eca17d0b281
    healthy: true
    lastUpdated: "2025-07-31T20:04:12Z"
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: community-operators
      namespace: openshift-marketplace
      resourceVersion: "31208"
      uid: a2552e99-ab22-4692-b2b7-bcb0695658e2
    healthy: true
    lastUpdated: "2025-07-31T20:04:12Z"
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: redhat-marketplace
      namespace: openshift-marketplace
      resourceVersion: "30987"
      uid: f465d361-1bda-4454-8fd4-f73713db2573
    healthy: true
    lastUpdated: "2025-07-31T20:04:12Z"
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: redhat-operators
      namespace: openshift-marketplace
      resourceVersion: "31093"
      uid: 44ebacf8-da11-4c2f-85c7-a6a854ac7905
    healthy: true
    lastUpdated: "2025-07-31T20:04:12Z"
  conditions:
  - lastTransitionTime: "2025-07-31T20:04:13Z"
    message: targeted catalogsource openshift-marketplace/foo missing
    reason: UnhealthyCatalogSourceFound
    status: "True"
    type: CatalogSourcesUnhealthy
  - message: 'constraints not satisfiable: no operators found from catalog foo in
      namespace openshift-marketplace referenced by subscription servicemeshoperator3,
      subscription servicemeshoperator3 exists'
    reason: ConstraintsNotSatisfiable
    status: "True"
    type: ResolutionFailed
  lastUpdated: "2025-07-31T20:04:13Z"
%
```
</details>

Next, I cleaned up the previous gatewayclass and subscription and created a new subscription with catalog "redhat-operators", channel "stable", OSSM version "servicemeshoperator3.v3.0.3", and Istio version "v1.24-latest"; cluster-ingress-operator created the expected subscription and configured the Istio CR with the expected version.
<details>

```text
% oc delete gatewayclasses/openshift-default
gatewayclass.gateway.networking.k8s.io "openshift-default" deleted
% oc -n openshift-operators delete subscriptions/servicemeshoperator3
subscription.operators.coreos.com "servicemeshoperator3" deleted
% oc create -f -<<'EOF'
apiVersion: gateway.networking.k8s.io/v1beta1
kind: GatewayClass
metadata:
  name: openshift-default
  annotations:
    unsupported.do-not-use.openshift.io/ossm-catalog: redhat-operators
    unsupported.do-not-use.openshift.io/ossm-channel: stable
    unsupported.do-not-use.openshift.io/ossm-version: servicemeshoperator3.v3.0.3
    unsupported.do-not-use.openshift.io/istio-version: v1.24-latest
spec:
  controllerName: openshift.io/gateway-controller/v1
EOF

gatewayclass.gateway.networking.k8s.io/openshift-default created
% oc get istios/openshift-gateway -o yaml
apiVersion: sailoperator.io/v1
kind: Istio
metadata:
  creationTimestamp: "2025-07-31T20:06:18Z"
  generation: 1
  name: openshift-gateway
  ownerReferences:
  - apiVersion: gateway.networking.k8s.io/v1
    kind: GatewayClass
    name: openshift-default
    uid: d7a09806-79cf-4e34-8d43-eb08b4a937d2
  resourceVersion: "33789"
  uid: 5d0b1a87-f7fa-466b-ad08-4db5448690d7
spec:
  namespace: openshift-ingress
  updateStrategy:
    type: InPlace
  values:
    global:
      defaultPodDisruptionBudget:
        enabled: false
      istioNamespace: openshift-ingress
      priorityClassName: system-cluster-critical
    meshConfig:
      accessLogFile: /dev/stdout
      defaultConfig:
        proxyHeaders:
          envoyDebugHeaders:
            disabled: true
          metadataExchangeHeaders:
            mode: IN_MESH
          server:
            disabled: true
      ingressControllerMode: "OFF"
    pilot:
      cni:
        enabled: false
      enabled: true
      env:
        ENABLE_GATEWAY_API_MANUAL_DEPLOYMENT: "false"
        PILOT_ENABLE_ALPHA_GATEWAY_API: "false"
        PILOT_ENABLE_GATEWAY_API: "true"
        PILOT_ENABLE_GATEWAY_API_CA_CERT_ONLY: "true"
        PILOT_ENABLE_GATEWAY_API_COPY_LABELS_ANNOTATIONS: "false"
        PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER: "true"
        PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER: "false"
        PILOT_ENABLE_GATEWAY_API_STATUS: "true"
        PILOT_GATEWAY_API_CONTROLLER_NAME: openshift.io/gateway-controller/v1
        PILOT_GATEWAY_API_DEFAULT_GATEWAYCLASS_NAME: openshift-default
        PILOT_MULTI_NETWORK_DISCOVER_GATEWAY_API: "false"
      podAnnotations:
        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
    sidecarInjectorWebhook:
      enableNamespacesByDefault: false
  version: v1.24-latest
status:
  activeRevisionName: openshift-gateway
  conditions:
  - lastTransitionTime: "2025-07-31T20:06:19Z"
    status: "True"
    type: Reconciled
  - lastTransitionTime: "2025-07-31T20:06:19Z"
    message: not all istiod pods are ready
    reason: IstiodNotReady
    status: "False"
    type: Ready
  observedGeneration: 1
  revisions:
    inUse: 0
    ready: 0
    total: 1
  state: IstiodNotReady
%
```
</details>

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -210,7 +210,7 @@ func (r *reconciler) currentInstallPlan(ctx context.Context) (bool, *operatorsv1
 			continue
 		}
 		for _, csvName := range installPlan.Spec.ClusterServiceVersionNames {
-			if csvName == r.config.GatewayAPIOperatorVersion {
+			if csvName == version {
```
alebedev87, line 213: Good catch! That's what I was talking about in [this comment](https://github.com/openshift/cluster-ingress-operator/pull/1246#discussion_r2243053351):

> For the latter I was worried that we may use r.config elsewhere in the code.

How can we make sure there are no other places like this?

pkg/operator/controller/gatewayclass/controller_test.go
alebedev87: I thought more of giving bad examples of how the annotation values can look like. It's our testing which doesn't oblige us to anything.

> I don't know what would be a realistic value for the catalog or channel other than the default values

Channel can be `stable-3.0`, catalog can be any catalog source from the cluster something like `community-operators`?

pkg/operator/controller/gatewayclass/subscription.go
Miciah: > How can we make sure there are no other places like this? Well, that's rather rhetorical, I hope the manual testing already caught all.

In my mind, this is an argument for passing specific inputs to functions (for example, `version string` for `ensureServiceMeshOperatorInstallPlan`) rather than relying on large state objects (such as `r.config`).  Using `r.config` everywhere is reminiscent of using global variables; it complicates the code unnecessarily and non-obviously.  

> Btw, this line was puzzling me when I was looking at it for the upgrade implementation. I was thinking why we don't use `subscription.spec.startingCSV` here because `startingCSV` contains the desired version which was written down before in the reconcile loop.

@rfredette might know the answer.  However, it seems to me that `startingCSV` isn't quite the right thing to use, based on its semantics; although in practice cluster-ingress-operator updates `startingCSV` to be the version we want to install, that use of `startingCSV` does not seem to be the intended use.

PR #1253, opened 2025-08-04
NE-2108: docs - correct Istio version format in ossm-overrides.md

alebedev87:
[Istio versions use `v` prefix](https://github.com/openshift-service-mesh/sail-operator/blob/c2c0b245714353bfb720b7cfdfe9c5dd21f64fe4/bundle/manifests/sailoperator.io_istios.yaml#L9925); without it, the Istio CRD will complain about unsupported version:

```
failed to create istioi /openshift-gateway: Istio.sailoperator.io "openshift-gateway" is invalid:
[spec.version: Unsupported value: "1.26-latest": supported values: "v1.26-latest"]
```


lihongan:
/lgtm
Thank you!

PR #1261, opened 2025-08-10
Trt 2243 ingress disruption

neisw:
Note - This is starting as an investigation due to the timing of the inclusion of [cluster-ingress-operator#1246](https://github.com/openshift/cluster-ingress-operator/pull/1246/files) and noted increase in ingress disruption along with payload failures.  We want to collect signal from the revert to evaluate further.

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are **testing** reverting this **potentially** breaking change to get CI and/or nightly payloads flowing again.

Multiple payload failures due to ingress disruption noted in [TRT-2243](https://issues.redhat.com/browse/TRT-2243)

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Before merging the unrevert, please run these jobs on the PR and check the result of these jobs to confirm the fix has corrected the problem:

```
/payload-aggregate periodic-ci-openshift-release-master-ci-4.20-upgrade-from-stable-4.19-e2e-gcp-ovn-rt-upgrade 10
```

CC: @Miciah


neisw:
Second go around had more failures

PR #1263, opened 2025-08-11
NE-1476: Added network policies for the router

knobunc:
Added the framework for network policies for the router.

The operator has a deny all network policy that for the openshift-ingress-operator namespace and an allow policy for egress to the apiserver and dns ports at any IP.

The operator installs a deny all network policy for the openshift-ingress namespace.

Then for each ingresscontroller that it manages it installs an allow policy for ingress for http and https traffic and metrics.

It has to allow ingress from the router pods to any IP because the route endpoints can be at any ip or pod.

It also needs access to the api server, but that is covered by the wildcard allow policy.

https://issues.redhat.com/browse/NE-1476

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -155,6 +155,7 @@ rules:
   - networking.k8s.io
   resources:
   - ingressclasses
+  - networkpolicies
   verbs:
   - '*'
 
--- a/manifests/01-network-policy.yaml
+++ b/manifests/01-network-policy.yaml
@@ -1,0 +1,43 @@
+# We control the namespace, deny anything we do not explicitly allow
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ingress-operator-deny-all
+  namespace: openshift-ingress-operator
+  annotations:
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+### Allow the operator needs to talk to the apiserver and dns,
+### but it also needs to be able to configure external DNS and the endpoints
+### are only known at run time.  So it needs wildcard egress.
+### Allow access to the metrics ports on the operators
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: ingress-operator-allow
+  namespace: openshift-ingress-operator
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  podSelector:
+    matchLabels:
+      name: ingress-operator
+  policyTypes:
+  - Egress
+  - Ingress
+  egress:
+  - to:
+    - ipBlock:
+        cidr: 0.0.0.0/0
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 9393
--- a/pkg/operator/controller/canary/networkpolicy_test.go
+++ b/pkg/operator/controller/canary/networkpolicy_test.go
@@ -1,0 +1,80 @@
+package canary
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test_desiredCanaryNetworkPolicy verifies the desired NetworkPolicy for canary pods.
+func Test_desiredCanaryNetworkPolicy(t *testing.T) {
+	np := desiredCanaryNetworkPolicy()
+
+	if np.Name != "ingress-canary" || np.Namespace != "openshift-ingress-canary" {
+		t.Fatalf("unexpected name/namespace: %s/%s", np.Namespace, np.Name)
+	}
+	if np.Labels["ingress.openshift.io/canary"] != canaryControllerName {
+		t.Fatalf("missing or wrong owner label: %#v", np.Labels)
+	}
+	// Ensure pod selector is set via helper.
+	if np.Spec.PodSelector.MatchLabels == nil {
+		t.Fatalf("pod selector should be set")
+	}
+	// Ensure ingress rules include ports 8888 and 8443.
+	want := map[int32]struct{}{8888: {}, 8443: {}}
+	got := map[int32]struct{}{}
+	for _, r := range np.Spec.Ingress {
+		for _, p := range r.Ports {
+			if p.Protocol != nil && *p.Protocol == corev1.ProtocolTCP && p.Port != nil && p.Port.IntVal != 0 {
+				got[p.Port.IntVal] = struct{}{}
+			}
+		}
+	}
+	for port := range want {
+		if _, ok := got[port]; !ok {
+			t.Fatalf("expected ingress port %d to be present", port)
+		}
+	}
+}
+
+// Test_canaryNetworkPolicyChanged verifies change detection and update behavior.
+func Test_canaryNetworkPolicyChanged(t *testing.T) {
+	original := &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-ingress-canary",
+			Name:      "ingress-canary",
+			Labels: map[string]string{
+				"ingress.openshift.io/canary": canaryControllerName,
+			},
+		},
+		Spec: networkingv1.NetworkPolicySpec{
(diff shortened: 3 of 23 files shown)


knobunc:
@candita thanks.  I've been focusing on the e2e-aws-ovn tests and those are consistently failing on:

lihongan:
cc @melvinjoseph86

manifests/00-cluster-role.yaml
```diff
@@ -155,6 +155,7 @@ rules:
   - networking.k8s.io
   resources:
   - ingressclasses
+  - networkpolicies
   verbs:
   - '*'
```
Miciah, line 160: If possible, avoid adding wildcard rules.  We have complaints about existing wildcards in RBAC rules, and we should avoid adding more.

manifests/01-network-policy.yaml
```diff
@@ -0,0 +1,43 @@
...
+    capability.openshift.io/name: Ingress
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  podSelector: {}
+  policyTypes:
+  - Ingress
+  - Egress
+---
+### Allow the operator needs to talk to the apiserver and dns,
```
Miciah, line 17: Do you mean the cluster DNS service (that is, CoreDNS) here?  
```suggestion
### The operator needs to talk to the apiserver and cluster DNS service,
```

pkg/operator/controller/canary/networkpolicy_test.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Test_desiredCanaryNetworkPolicy verifies the desired NetworkPolicy for canary pods.
+func Test_desiredCanaryNetworkPolicy(t *testing.T) {
+	np := desiredCanaryNetworkPolicy()
+
+	if np.Name != "ingress-canary" || np.Namespace != "openshift-ingress-canary" {
+		t.Fatalf("unexpected name/namespace: %s/%s", np.Namespace, np.Name)
+	}
```
Miciah, line 17: Consider using testify for these assertions.

pkg/operator/controller/canary/networkpolicy_test.go
```diff
@@ -0,0 +1,80 @@
...
+	if np.Spec.PodSelector.MatchLabels == nil {
+		t.Fatalf("pod selector should be set")
+	}
+	// Ensure ingress rules include ports 8888 and 8443.
+	want := map[int32]struct{}{8888: {}, 8443: {}}
+	got := map[int32]struct{}{}
+	for _, r := range np.Spec.Ingress {
+		for _, p := range r.Ports {
+			if p.Protocol != nil && *p.Protocol == corev1.ProtocolTCP && p.Port != nil && p.Port.IntVal != 0 {
+				got[p.Port.IntVal] = struct{}{}
```
Miciah, line 31: `struct{}` is passé.  Consider using `"k8s.io/apimachinery/pkg/util/sets".Set[int32]`.

pkg/operator/controller/canary/networkpolicy_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// Change in owner label should be detected but preserve other labels.
+	mutated = original.DeepCopy()
+	mutated.Labels["ingress.openshift.io/canary"] = "other"
+	if changed, updated := canaryNetworkPolicyChanged(original, mutated); !changed {
+		t.Fatalf("expected change detected for label diff")
+	} else if updated.Labels["ingress.openshift.io/canary"] != "other" {
+		t.Fatalf("expected owner label updated")
+	}
```
Miciah, line 79: Despite the comment, the test logic doesn't actually check that other labels are preserved.

pkg/operator/controller/gatewayapi/controller.go
```diff
@@ -144,6 +149,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	// Ensure the allow NetworkPolicy for Gateway API components.
+	if _, _, err := r.ensureGatewayAPINetworkPolicy(ctx); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to ensure gateway-api network policy: %w", err)
+	}
```
Miciah, line 155: This logic to create the NetworkPolicy for gateway proxies seems like it belongs in the operator's gatewayclass controller with the logic that installs the Istio control plane.

pkg/operator/controller/gatewayapi/networkpolicy.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	updated := current.DeepCopy()
+	updated.Spec = expected.Spec
+	// Merge/overwrite labels from expected; preserve any unrelated labels.
+	if updated.Labels == nil {
+		updated.Labels = map[string]string{}
+	}
+	for k, v := range expected.Labels {
+		updated.Labels[k] = v
+	}
```
Miciah, line 100: This function is not fixed-point.  That is, you can get in an update loop where `current` has some unexpected label that isn't in `expected`.  Then the function merges `updated` into `expected`, but it keeps the unexpected label, and then the `cmp.Equal(current.Labels, expected.Labels, ...)` detects this as a change, and you get a loop.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1078,6 +1082,10 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
 		return fmt.Errorf("failed to ensure namespace: %v", err)
 	}
 
+	if _, _, err := r.ensureRouterNamespaceNetworkPolicy(ci); err != nil {
+		return fmt.Errorf("failed to ensure default namespacenetwork policy: %v", err)
```
Miciah, line 1086:
```		return fmt.Errorf("failed to ensure default namespace network policy: %w", err)
```

pkg/operator/controller/ingress/namespace.go
```diff
@@ -155,3 +157,17 @@ func (r *reconciler) ensureRouterClusterRoleBinding() error {
...
 }
+
+func (r *reconciler) ensureRouterNamespaceNetworkPolicy(ci *operatorv1.IngressController) (bool, *networkingv1.NetworkPolicy, error) {
+	np := manifests.RouterNetworkPolicyDenyAll()
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: np.Namespace, Name: np.Name}, np); err != nil {
+		if !errors.IsNotFound(err) {
+			return false, nil, fmt.Errorf("failed to get router network policy %s/%s: %v", np.Namespace, np.Name, err)
+		}
+		if err := r.client.Create(context.TODO(), np); err != nil {
+			return false, nil, fmt.Errorf("failed to create router network policy %s/%s: %v", np.Namespace, np.Name, err)
```
Miciah, line 168: This is the "default namespace network policy", not the "router network policy".  Anyway, `ensureIngressController` wraps these errors, so the wrapping here seems redundant.

pkg/operator/controller/ingress/networkpolicy.go
```diff
@@ -0,0 +1,121 @@
...
+	controller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	networkingv1 "k8s.io/api/networking/v1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+// ensureRouterNetworkPolicy ensures the per-ingresscontroller NetworkPolicy that
+// allows ingress to router pods and egress to the network exists and is up to date.
+func (r *reconciler) ensureRouterNetworkPolicy(ic *operatorv1.IngressController) error {
```
Miciah, line 21: Any reason this `ensure` function doesn't return the object or Boolean value?  It doesn't strictly need to do so, so I don't object if you prefer to keep it simple, but it deviates from the established pattern for `ensure` methods.

pkg/operator/controller/ingress/networkpolicy_test.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// Change in owner label should be detected but preserve other labels.
+	mutated = original.DeepCopy()
+	mutated.Labels["ingresscontroller.operator.openshift.io/owning-ingresscontroller"] = "other"
+	if changed, updated := routerNetworkPolicyChanged(original, mutated); !changed {
+		t.Fatalf("expected change detected for label diff")
+	} else if updated.Labels["ingresscontroller.operator.openshift.io/owning-ingresscontroller"] != "other" {
+		t.Fatalf("expected owner label updated")
+	}
```
Miciah, line 83: This isn't checking that other labels are preserved.

Miciah:
Please also add the new networkpolicies to `relatedObjects`: https://github.com/openshift/cluster-ingress-operator/blob/cf93ea59b523455d4651bba3dc1858ca60c9ccc7/pkg/operator/controller/status/controller.go#L226-L289

See also <https://issues.redhat.com//browse/OCPBUGS-65476>.

rikatz:
/close
implemented via #1392

PR #1247, opened 2025-07-29
NE-2097: Add logic to upgrade OSSM Operator

alebedev87:
This PR introduces initial upgrade logic for the OSSM Operator.

The implemented logic supports automatic approval of the next install plan from the upgrade graph, progressing sequentially toward the desired version. If the next available install plan version is semantically greater than the desired version, the upgrade stops.

**Status handling updates**

- The Progressing condition is set to True during an upgrade (i.e., when the installed version does not match the desired version), and remains so until the desired version is successfully installed.
- The Degraded condition is set to True if any install plan in the upgrade path fails.

**Notes**

- This PR does not include any logic for upgrading Istio CR. A temporary safeguard is already in place: the Istio CRD contains a list of supported OSSM versions. Therefore, until the desired OSSM version is installed, any Istio update will fail due to the CRD validation.

**Manual test**
```
$ oc get co ingress -w
NAME      VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.19.0-0.nightly-2025-07-29-065329   True        False         False      24m
# GatewayClass creation: 3.0.0 version is hard coded
ingress   4.19.0-0.nightly-2025-07-29-065329   True        True          False      24m     OSSM operator is upgrading to version "servicemeshoperator3.v3.0.0"
ingress   4.19.0-0.nightly-2025-07-29-065329   True        True          False      24m     OSSM operator is upgrading to version "servicemeshoperator3.v3.0.0" (installed(-ing):"servicemeshoperator3.v3.0.0",next:"servicemeshoperator3.v3.0.0")
ingress   4.19.0-0.nightly-2025-07-29-065329   True        True          False      24m     OSSM operator is upgrading to version "servicemeshoperator3.v3.0.0" (installed(-ing):"servicemeshoperator3.v3.0.0",next:"servicemeshoperator3.v3.0.1")
# 3.0.0 is installed
ingress   4.19.0-0.nightly-2025-07-29-065329   True        False         False      25m
# Desired OSSM version changed to 3.0.2
ingress   4.19.0-0.nightly-2025-07-29-065329   True        True          False      25m     OSSM operator is upgrading to version "servicemeshoperator3.v3.0.2" (installed(-ing):"servicemeshoperator3.v3.0.0",next:"servicemeshoperator3.v3.0.1")
[...]

PR #1268, opened 2025-08-13
NE-2066: Set degraded=true when OSSM 3 can't be installed

rfredette:
Detect subscriptions that would prevent the Ingress Operator from installing OSSM 3, and set the operator's degraded condition to true when any of those subscriptions are present.

This is the implementation of [NE-2066](https://issues.redhat.com//browse/NE-2066)

### DIFF
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/google/go-cmp/cmp"
@@ -72,6 +73,14 @@ var (
 	// relatedObjectsCRDs is a set of names of CRDs that we add to
 	// relatedObjects if they exist.
 	relatedObjectsCRDs = sets.New[string](gatewaysResourceName, gatewayclassesResourceName, istiosResourceName)
+
+	// ossmSubscriptions lists the package names for all OLM subscriptions that are known to conflict with the
+	// subscription to OSSM3 created by the Ingress Operator.
+	ossmSubscriptions = sets.New[string](
+		"sailoperator",
+		"servicemeshoperator",
+		"servicemeshoperator3",
+	)
 )
 
 // New creates the status controller. This is the controller that handles all
@@ -83,10 +92,20 @@ var (
 // in case something else updates or deletes it.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	operatorCache := mgr.GetCache()
+	// The status controller needs to be aware of conflicting OLM subscriptions in any namespace. In order to
+	// prevent ballooning the main cache to watch all namespaces on all its informers, create a separate cache for
+	// tracking OLM subscriptions across all namespaces.
+	var subscriptionCache cache.Cache
+	var err error
+	if subscriptionCache, err = cache.New(mgr.GetConfig(), cache.Options{}); err != nil {
+		return nil, err
+	}
+	mgr.Add(subscriptionCache)
 	reconciler := &reconciler{
-		config: config,
-		client: mgr.GetClient(),
-		cache:  operatorCache,
+		config:            config,
+		client:            mgr.GetClient(),
+		cache:             operatorCache,
+		subscriptionCache: subscriptionCache,
 	}
 	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: reconciler})
 	if err != nil {
@@ -156,6 +175,17 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		})); err != nil {
 			return nil, err
 		}
+
+		isOSSMSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
+			subscription, ok := o.(*operatorsv1alpha1.Subscription)
+			if !ok || subscription.Spec == nil {
+				return false
+			}
+			return ossmSubscriptions.Has(subscription.Spec.Package)
+		})
+		if err := c.Watch(source.Kind[client.Object](reconciler.subscriptionCache, &operatorsv1alpha1.Subscription{}, handler.EnqueueRequestsFromMapFunc(toDefaultIngressController), isOSSMSubscription)); err != nil {
+			return nil, err
+		}
 	}
 
 	return c, nil
@@ -178,6 +208,7 @@ type Config struct {
 	CanaryImage                     string
 	OperatorReleaseVersion          string
 	Namespace                       string
+	GatewayAPIOperatorVersion       string
 }
 
 // IngressOperatorStatusExtension holds status extensions of the ingress cluster operator.
@@ -192,8 +223,9 @@ type IngressOperatorStatusExtension struct {
 type reconciler struct {
 	config Config
 
-	client client.Client
-	cache  cache.Cache
+	client            client.Client
+	cache             cache.Cache
+	subscriptionCache cache.Cache
 }
 
 // Reconcile computes the operator's current status and therefrom creates or
@@ -366,6 +398,14 @@ type operatorState struct {
 	// haveGatewayclassesResource means that the
 	// "gatewayclasses.gateway.networking.k8s.io" CRD exists.
 	haveGatewayclassesResource bool
+	// ossmSubscriptions contains all subscriptions that may conflict with the operator-created ossm subscription.
+	ossmSubscriptions []operatorsv1alpha1.Subscription
+	// expectedGatewayAPIOperatorVersion reflects the expected OSSM 3 version. It is used in determining if a
+	// user-supplied OSSM 3 subscription would cause the operator's installation of OSSM 3 to fail.
+	expectedGatewayAPIOperatorVersion string
+	// shouldInstallOSSM reflects whether the ingress operator should install OSSM. Currently, this happens when a
+	// gateway class with Spec.ControllerName=operatorcontroller.OpenShiftGatewayClassControllerName is created.
+	shouldInstallOSSM bool
 }
 
 // getOperatorState gets and returns the resources necessary to compute the
@@ -447,6 +487,27 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
 			} else {
 				state.haveIstiosResource = true
 			}
+
+			state.expectedGatewayAPIOperatorVersion = r.config.GatewayAPIOperatorVersion
+			subscriptionList := operatorsv1alpha1.SubscriptionList{}
+			if err := r.subscriptionCache.List(ctx, &subscriptionList); err != nil {
+				return state, fmt.Errorf("failed to get subscriptions: %w", err)
+			}
+			for _, subscription := range subscriptionList.Items {
+				if subscription.Spec != nil && ossmSubscriptions.Has(subscription.Spec.Package) {
+					state.ossmSubscriptions = append(state.ossmSubscriptions, subscription)
+				}
+			}
+
+			gatewayClassList := gatewayapiv1.GatewayClassList{}
+			if err := r.cache.List(ctx, &gatewayClassList, client.MatchingFields{
+				operatorcontroller.GatewayClassIndexFieldName: operatorcontroller.OpenShiftGatewayClassControllerName,
+			}); err != nil {
+				return state, fmt.Errorf("failed to list gateway classes: %w", err)
+			}
+			// If one or more gateway classes have ControllerName=operatorcontroller.OpenShiftGatewayClassControllerName,
+			// the ingress operator should try to install OSSM.
+			state.shouldInstallOSSM = (len(gatewayClassList.Items) > 0)
 		}
 	}
 
@@ -504,6 +565,7 @@ func computeOperatorDegradedCondition(state operatorState) configv1.ClusterOpera
 	for _, fn := range []func(state operatorState) configv1.ClusterOperatorStatusCondition{
 		computeIngressControllerDegradedCondition,
 		computeGatewayAPICRDsDegradedCondition,
+		computeGatewayAPIInstallDegradedCondition,
 	} {
 		degradedCondition = joinConditions(degradedCondition, fn(state))
 	}
@@ -565,6 +627,64 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
 		degradedCondition.Status = configv1.ConditionTrue
 		degradedCondition.Reason = "GatewayAPICRDsDegraded"
 		degradedCondition.Message = fmt.Sprintf("Unmanaged Gateway API CRDs found: %s.", state.unmanagedGatewayAPICRDNames)
+	}
+
+	return degradedCondition
+}
+
+// computeGatewayAPIInstallDegradedCondition computes the degraded condition for the Gateway API OSSM subscription. It
+// checks for known conflicting subscriptions, as well as already existing subscription(s) to OSSM3, and reports
+// degraded when any of those subscriptions would prevent the installation of an appropriate Istio control plane.
+func computeGatewayAPIInstallDegradedCondition(state operatorState) configv1.ClusterOperatorStatusCondition {
+	degradedCondition := configv1.ClusterOperatorStatusCondition{}
+
+	// If OSSM doesn't need to be installed, or there are no possible conflicting subscriptions, return degraded=false.
+	if !state.shouldInstallOSSM || len(state.ossmSubscriptions) == 0 {
+		return degradedCondition
+	}
+
+	conflicts := []string{}
+	warnings := []string{}
+	for _, subscription := range state.ossmSubscriptions {
+		if subscription.Spec.Package == "servicemeshoperator3" {
+			if _, found := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; found {
+				// The subscription that the ingress operator creates naturally does not conflict with itself.
+				continue
+			}
+			if subscription.Status.InstalledCSV == "" {
+				// The subscription hasn't finished its install. We will get another reconcile request once the
+				// installation is complete, so we can ignore this for now.
+				continue
+			}
+			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion)
+			switch {
+			case err != nil:
+				warnings = append(warnings, fmt.Sprintf("failed to compare installed OSSM version to expected: %v", err))
+			case versionDiff < 0:
+				// Installed version is newer than expected. Gateway API install may still work if the correct Istio
+				// version is supported. Warn the user that the installed OSSM version may be incompatible.
+				warnings = append(warnings, fmt.Sprintf("Found version %s, but operator-managed Gateway API expects version %s. Operator-managed Gateway API may not work as intended.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion))
+			case versionDiff > 0:
+				// Installed version is older than expected. Gateway API install will not work, since the correct Istio
+				// version won't be supported.
+				conflicts = append(conflicts, fmt.Sprintf("Installed version %s does not support operator-managed Gateway API. Install version %s or uninstall %s/%s to enable functionality.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion, subscription.Namespace, subscription.Name))
+			case versionDiff == 0:
+				// Installed version is exactly as expected. Nothing to do.
+			}
+		} else {
+			conflicts = append(conflicts, fmt.Sprintf("Package %s from subscription %s/%s prevents enabling operator-managed Gateway API. Uninstall %s/%s to enable functionality.", subscription.Spec.Package, subscription.Namespace, subscription.Name, subscription.Namespace, subscription.Name))
+		}
+	}
+	if len(conflicts) > 0 {
+		degradedCondition.Status = configv1.ConditionTrue
+		degradedCondition.Reason = "GatewayAPIInstallConflict"
+		degradedCondition.Message = strings.Join(conflicts, "\n")
+	} else if len(warnings) > 0 {
+		// Warnings are not enough to set degraded=true, but should still be included in the status message to warn
+		// users to possible issues. Leave status=false and reason unset, but put warning messages in the message field.
+		degradedCondition.Status = configv1.ConditionFalse
+		degradedCondition.Reason = "GatewayAPIInstallWarnings"
+		degradedCondition.Message = strings.Join(warnings, "\n")
 	}
 
 	return degradedCondition
@@ -834,7 +954,7 @@ func joinConditions(currCond, newCond configv1.ClusterOperatorStatusCondition) c
 	case configv1.ConditionTrue:
 		if currCond.Status == configv1.ConditionTrue {
 			currCond.Reason += "And" + newCond.Reason
-			currCond.Message += newCond.Message
+			currCond.Message += "\n" + newCond.Message
 		} else {
 			// Degraded=True status overrides other statuses.
 			currCond.Status = newCond.Status
@@ -844,7 +964,7 @@ func joinConditions(currCond, newCond configv1.ClusterOperatorStatusCondition) c
 	case configv1.ConditionUnknown:
 		if currCond.Status == configv1.ConditionUnknown {
 			currCond.Reason += "And" + newCond.Reason
-			currCond.Message += newCond.Message
+			currCond.Message += "\n" + newCond.Message
 		} else if currCond.Status != configv1.ConditionTrue {
 			// Degraded=Unknown overrides false and empty statuses.
 			currCond.Status = newCond.Status
@@ -854,7 +974,7 @@ func joinConditions(currCond, newCond configv1.ClusterOperatorStatusCondition) c
 	case configv1.ConditionFalse:
 		if currCond.Status == configv1.ConditionFalse {
 			currCond.Reason += "And" + newCond.Reason
-			currCond.Message += newCond.Message
+			currCond.Message += "\n" + newCond.Message
 		} else if currCond.Status == "" {
 			// Degraded=False overrides empty status.
 			currCond.Status = newCond.Status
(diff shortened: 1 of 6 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVerison)
+			if err != nil {
+				log.Error(err, "failed to compare installed OSSM version to expected", "installed", subscription.Status.InstalledCSV, "expected", state.expectedGatewayAPIOperatorVerison)
+				conflicts = append(conflicts, fmt.Sprintf("Installed version %s may prevent enabling operator-managed Gateway API. Install version %s or remove subscription %s/%s to enable.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVerison, subscription.Namespace, subscription.Name))
+			}
+			switch {
+			case versionDiff < 0:
+				// Installed version is newer than expected. Gateway API install may still work if the correct Istio
+				// version is supported.
+				log.Info("found newer OSSM version than expected. Gateway API install may not work as intended", "installed", subscription.Status.InstalledCSV, "expected", state.expectedGatewayAPIOperatorVerison)
```
Miciah, line 651: Better to put something in the status condition message rather than logging it on every status reconciliation.  (This might require changes to `joinConditions` to preserve the message even if the status is `False`.)

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
 	return degradedCondition
 }
 
+// computeGatewayAPIInstallDegradedCondition computes the degraded condition for the Gateway API OSSM subscription. It
+// checks for known conflicting subscriptions, as well as already existing subscription(s) to OSSM3, and reports
+// degraded when any of those subscriptions would prevent the installation of an appropriate Istio control plane.
```
Miciah, line 615: Can we make this conditional on having Gateway API enabled (that is, having a gatewayclass with our controller name)?

pkg/operator/controller/status/controller.go
```diff
@@ -448,6 +474,19 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
...
+
+		state.expectedGatewayAPIOperatorVerison = r.config.GatewayAPIOperatorVersion
+		subscriptionList := operatorsv1alpha1.SubscriptionList{}
+		if err := r.config.SubscriptionCache.List(ctx, &subscriptionList); err != nil {
+			return state, fmt.Errorf("failed to get subscriptions: %w", err)
+		}
+		for _, subscription := range subscriptionList.Items {
+			log.Info("evaluating subscription", "name", subscription.Name)
+			if conflictingSubscriptions.Has(subscription.Spec.Package) {
+				log.Info("found relevant subscription", "name", subscription.Name)
```
rfredette, line 486: remove this debugging log line too

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+				// The subscription hasn't finished its install. We will get another reconcile request once the
+				// installation is complete, so we can ignore this for now.
+				continue
+			}
+			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVerison)
+			if err != nil {
+				log.Error(err, "failed to compare installed OSSM version to expected", "installed", subscription.Status.InstalledCSV, "expected", state.expectedGatewayAPIOperatorVerison)
+				conflicts = append(conflicts, fmt.Sprintf("Installed version %s may prevent enabling operator-managed Gateway API. Install version %s or remove subscription %s/%s to enable.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVerison, subscription.Namespace, subscription.Name))
+			}
+			switch {
```
Miciah, line 647: No need to log the error *and* update `conflicts`; might as well just put the error message in the `conflicts` message.  Also, you don't want to continue to the `switch` if you got an error:
```suggestion
			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVerison)
			switch {
			case err != nil:
				conflicts = append(conflicts, fmt.Sprintf("failed to compare installed OSSM version to expected: %w", err))
```

pkg/operator/controller/status/controller.go
```diff
@@ -156,6 +164,18 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		})); err != nil {
 			return nil, err
 		}
+
+		isConflictingSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
```
Miciah, line 168:
```suggestion
		isOSSMSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
```

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+func computeGatewayAPIInstallDegradedCondition(state operatorState) configv1.ClusterOperatorStatusCondition {
+	degradedCondition := configv1.ClusterOperatorStatusCondition{}
+
+	if len(state.Subscriptions) == 0 {
+		return degradedCondition
+	}
+
+	conflicts := []string{}
+	for _, subscription := range state.Subscriptions {
+		if subscription.Spec.Package == "servicemeshoperator3" {
```
Miciah, line 625: Consider introducing a named const for "servicemeshoperator3".

pkg/operator/controller/status/controller.go
```diff
@@ -864,3 +966,32 @@ func joinConditions(currCond, newCond configv1.ClusterOperatorStatusCondition) c
...
+		return 0, fmt.Errorf("%q and %q are different packages. cannot compare version numbers", a, b)
+	}
+	if aX != bX {
+		return bX - aX, nil
+	}
+	if aY != bY {
+		return bY - aY, nil
+	}
+	return bZ - aZ, nil
+}
```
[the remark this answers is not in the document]
alebedev87: Interesting, I didn't see this indirect dependency that's why I chose [the coreos's semver module](https://github.com/openshift/cluster-ingress-operator/pull/1247/files#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63R83-R84) in my PR. The blang's module has more GitHub stars but looks a little less maintained.

pkg/operator/controller/status/controller.go
alebedev87: So, it will be `Degraded=False` but with `Found newer OSSM version than expected` message? Will this sound contradictory?

alebedev87:
@rfredette : I forgot to mention that `Test_computeOperatorDegradedCondition` unit test should be updated to covert the changes.

pkg/operator/controller/status/controller.go
rfredette: I intend to keep the `Gateway API install may not work as intended` part of the message, which hopefully at least gives warning that something may be wrong.

In the future, we can be more certain if the ingress operator needs to be degraded by checking if the version of Istio that it needs is still supported by the newer OSSM version. With that, we can do away with the vague warning about gateway API possibly working or possibly not. I don't think I have the time to include that in this PR, but I created [NE-2119](https://issues.redhat.com/browse/NE-2119) to track that work.

Miciah: I think it is fine to have `Degraded` with status `False` and a message that says it isn't necessarily degraded _but_ the subscription isn't exactly what we expected.  I think that's good enough.  

If we get ambitious, maybe it would make sense to add a metric and an alerting rule?  That would make the communication a little more explicit and difficult to miss, but it could also be annoying, and it might be more work than we have time to do right now.

pkg/operator/controller/status/controller.go
Miciah: It might be safer to check the gatewayclass, in case a conflicting subscription or some other issue is preventing the operator from creating the expected subscription.

alebedev87 (commented):
Another look after some comments were addressed.

rfredette:
test failures appear unrelated

/test e2e-aws-operator

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+					isOurSubscription = true
+					break
+				}
+			}
+			if isOurSubscription {
+				// The subscription that the ingress operator creates naturally does not conflict with itself.
+				continue
+			}
+			if subscription.Status.InstalledCSV == "" {
+				// The subscription hasn't finished its install. We will get another reconcile request once the
```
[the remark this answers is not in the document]
alebedev87: [As discussed in Slack](https://redhat-internal.slack.com/archives/CBWMXQJKD/p1755532729080009), let's follow up on this point later. Because checking for a non conflicting installation (additionally to Subscription) adds complexity which may be blokcing while we are approaching the branch cut.

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +610,68 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+			}
+		} else {
+			log.Info(fmt.Sprintf("Package %s from subscription %s/%s prevents enabling operator-managed Gateway API", subscription.Spec.Package, subscription.Namespace, subscription.Name))
+			conflicts = append(conflicts, fmt.Sprintf("Package %s from subscription %s/%s prevents enabling operator-managed Gateway API. Remove subscription %s/%s to enable.", subscription.Spec.Package, subscription.Namespace, subscription.Name, subscription.Namespace, subscription.Name))
+		}
+	}
+	if len(conflicts) > 0 {
+		log.Info("Setting degraded to True", "conflicts", conflicts)
+		degradedCondition.Status = configv1.ConditionTrue
+		degradedCondition.Reason = "GatewayAPIInstallBlocked"
```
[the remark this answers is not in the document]
alebedev87: > I feel like having the user-facing names reference Gateway API rather than OSSM is the better option. I don't know if it's clear to users what service mesh has to do with the ingress operator, and I think referring to gateway API rather than OSSM helps to clarify what's possibly not working.

We mention the OSSM operator in the documentation but indeed there is no big emphasis on this.

In the context of [the other comment](https://github.com/openshift/cluster-ingress-operator/pull/1268#discussion_r2288184124): what do you think about `GatewayAPIInstallConflict` instead of `GatewayAPIInstalledBlocked`?

pkg/operator/operator.go
```diff
@@ -227,6 +236,8 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		GatewayAPIControllerEnabled:     gatewayAPIControllerEnabled,
 		MarketplaceEnabled:              marketplaceEnabled,
 		OperatorLifecycleManagerEnabled: olmEnabled,
+		SubscriptionCache:               subscriptionCache,
+		GatewayAPIOperatorVersion:       config.GatewayAPIOperatorVersion,
```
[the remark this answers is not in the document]
Miciah: @rfredette and I have previously discussed whether we need to take the override into account.  We decided that it isn't strictly necessary.  Because `desiredSubscription` sets the `ingress.operator.openshift.io/owned` annotation and `computeGatewayAPIInstallDegradedConditionwe` checks for the same annotation, this PR will *not* cause the operator to report `Degraded=False` if the operator has created the subscription, with or without overrides.  

Aside from that, it is debatable whether it makes sense to report `Degraded=False` when an override is in use.  I would prefer not to report `Degraded=False` in this case, especially considering that layered products could use this annotation for tech-preview releases.  However, because we have the annotation, we generally shouldn't run into a scenario where the operator reports `Degraded=False` due to overrides.

rfredette:
e2e-aws-ovn failure seems unrelated.

/test e2e-aws-ovn

pkg/operator/operator.go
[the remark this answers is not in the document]
alebedev87: Sorry, I saw this message only after I posted my comment and refreshed the page:

> @rfredette pointed out to me today that we wouldn't actually set Degraded=True if the override were used to specify a newer version; you would get Degraded=False with a message on the status condition along the lines of, "No problem detected, but you are using an untested configuration." (The discussion at https://github.com/openshift/cluster-ingress-operator/pull/1268#discussion_r2274529711 is relevant.)

To me, this sounds coherent with what I posted and makes it even better: the user will be warned in a soft way.

pkg/operator/operator.go
```diff
@@ -217,6 +218,14 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
 		return nil, fmt.Errorf("failed to create configurable-route controller: %v", err)
 	}
 
+	// The status controller needs to be aware of conflicting OLM subscriptions in any namespace. In order to prevent
+	// ballooning the main cache to watch all namespaces on all its informers, create a separate cache for tracking OLM
+	// subscriptions across all namespaces.
```
candita, line 223: Why do we need to track subscriptions across all namespaces? Aren't there a limited number of namespaces that subscriptions should be installed to?  Or are we just being prepared for anything here?

pkg/operator/operator.go
```diff
@@ -398,16 +410,23 @@ func (o *Operator) Start(ctx context.Context) error {
...
-	errChan := make(chan error)
+	mgrErrChan := make(chan error)
 	go func() {
-		errChan <- o.manager.Start(ctx)
+		mgrErrChan <- o.manager.Start(ctx)
+	}()
+
+	cacheErrChan := make(chan error)
+	go func() {
+		cacheErrChan <- o.subscriptionCache.Start(ctx)
```
candita, line 420: Why do we need to explicitly start a cache in this case?  Because it is not contained in the manager?  Do we also need to explicitly stop it?

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +623,64 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+			}
+			if subscription.Status.InstalledCSV == "" {
+				// The subscription hasn't finished its install. We will get another reconcile request once the
+				// installation is complete, so we can ignore this for now.
+				continue
+			}
+			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion)
+			switch {
+			case err != nil:
+				conflicts = append(conflicts, fmt.Sprintf("failed to compare installed OSSM version to expected: %v", err))
```
candita, line 653: It would be better to future-proof this so we didn't start getting Degraded=true if a version formatting change happened without our knowledge (or we forgot to check this with every update).  How about if there's an error, we set this to be a warning?

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +623,64 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+				warnings = append(warnings, fmt.Sprintf("Found version %s, but operator-managed Gateway API expects version %s. operator-managed Gateway API may not work as intended.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion))
+			case versionDiff > 0:
+				// Installed version is older than expected. Gateway API install will not work, since the correct Istio
+				// version won't be supported.
+				conflicts = append(conflicts, fmt.Sprintf("Installed version %s is too old to support operator-managed Gateway API. Install version %s or uninstall %s/%s to enable.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion, subscription.Namespace, subscription.Name))
+			case versionDiff == 0:
+				// Installed version is exactly as expected. Nothing to do.
+			}
+		} else {
+			conflicts = append(conflicts, fmt.Sprintf("Package %s from subscription %s/%s prevents enabling operator-managed Gateway API. Uninstall %s/%s to enable.", subscription.Spec.Package, subscription.Namespace, subscription.Name, subscription.Namespace, subscription.Name))
```
candita, line 666: See https://github.com/openshift/cluster-ingress-operator/pull/1268/files#r2289426439

How about:
```suggestion
			conflicts = append(conflicts, fmt.Sprintf("Package %s from subscription %s/%s conflicts with operator-managed Gateway API. Uninstall %s/%s to restore or enable functionality.", subscription.Spec.Package, subscription.Namespace, subscription.Name, subscription.Namespace, subscription.Name))
```

pkg/operator/operator.go
candita: It's fine to be prepared for anything.

pkg/operator/controller/status/controller.go
```diff
@@ -447,6 +478,27 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
...
 			}
+
+			state.expectedGatewayAPIOperatorVersion = r.config.GatewayAPIOperatorVersion
+			subscriptionList := operatorsv1alpha1.SubscriptionList{}
+			if err := r.config.SubscriptionCache.List(ctx, &subscriptionList); err != nil {
+				return state, fmt.Errorf("failed to get subscriptions: %w", err)
+			}
+			for _, subscription := range subscriptionList.Items {
+				if ossmSubscriptions.Has(subscription.Spec.Package) {
+					state.ossmSubscriptions = append(state.ossmSubscriptions, subscription)
```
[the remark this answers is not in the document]
alebedev87: > Is there anything that resets this list of ossmSubscriptions, or otherwise cleans up when a subscription is deleted?

The operator state is computed from scratch for each reconciliation loop.

> do we care if the subscription is approved or not?

Subscriptions cannot be approved, InstallPlans can. For the degraded status we only care for Subscriptions.

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +623,64 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
 
+// computeGatewayAPIInstallDegradedCondition computes the degraded condition for the Gateway API OSSM subscription. It
+// checks for known conflicting subscriptions, as well as already existing subscription(s) to OSSM3, and reports
+// degraded when any of those subscriptions would prevent the installation of an appropriate Istio control plane.
+func computeGatewayAPIInstallDegradedCondition(state operatorState) configv1.ClusterOperatorStatusCondition {
+	degradedCondition := configv1.ClusterOperatorStatusCondition{}
+
+	// If OSSM doesn't need to be installed, or there are no possible conflicting subscriptions, return degraded=false.
+	if !state.shouldInstallOSSM || len(state.ossmSubscriptions) == 0 {
+		return degradedCondition
```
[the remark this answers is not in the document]
alebedev87: Degraded conditions are joined together. Adding an empty one is not going to impact any other condition from the joined list. It's unit tested in `Test_computeOperatorDegradedCondition`: any testcases which existed before this change are not impacted by this empty condition.

pkg/operator/controller/status/controller.go
```diff
@@ -156,6 +165,18 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		})); err != nil {
 			return nil, err
 		}
+
+		isOSSMSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
+			subscription, ok := o.(*operatorsv1alpha1.Subscription)
+			if !ok {
+				return false
+			}
+			return ossmSubscriptions.Has(subscription.Spec.Package)
```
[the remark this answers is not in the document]
rfredette: Actually, `Spec` can be `nil`. I'll add checks for that.

pkg/operator/operator.go
Miciah: Now that you mention it, the logic is unclear to me as well.  I figure it has to do with the fact that the subscriptions resource might not exist when the operator starts, but I'm not entirely sure.  

In the case of, say, the route-metrics controller, the controller similarly creates its own cache so that it can watch all namespaces without impacting the main cache.  However, the route-metrics controller initializes the cache in its `New` method and registers the cache with the manager so that the manager starts it:  https://github.com/openshift/cluster-ingress-operator/blob/15bdd86b9bf96f1b6a23c34c670f442a9a018a41/pkg/operator/controller/route-metrics/controller.go#L43-L51

Ryan, can you add a code comment to explain why special handling is needed for the status controller's subscriptions cache?

pkg/operator/controller/status/controller_test.go
```diff
@@ -793,7 +796,86 @@ func Test_computeOperatorDegradedCondition(t *testing.T) {
...
+					sub("sailoperator", "sailoperator", "sailoperator.v1.0.0", false),
+				},
+				shouldInstallOSSM:                 true,
+				expectedGatewayAPIOperatorVersion: "servicemeshoperator3.v3.1.0",
+			},
+			expectCondition: configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorDegraded,
+				Status:  configv1.ConditionTrue,
+				Reason:  "GatewayAPIInstallConflict",
+				Message: "Installed version servicemeshoperator3.v3.0.0 is too old to support operator-managed Gateway API. Install version servicemeshoperator3.v3.1.0 or uninstall foo/servicemeshoperator3 to enable.\nPackage sailoperator from subscription foo/sailoperator prevents enabling operator-managed Gateway API. Uninstall foo/sailoperator to enable.",
```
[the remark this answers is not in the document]
rfredette: Technically I don't believe they can both be installed simultaneously; ossm3 and sail both claim the same CRDs, and OLM prevents installing any operator that claims CRDs already claimed by an installed operator. I should probably adjust this test scenario to only have one of the installs successful, but I think the actual message is fine

pkg/operator/controller/status/controller.go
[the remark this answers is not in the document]
candita: Nit: a small update to the comment would be helpful then:

// If OSSM doesn't need to be installed, or there are no possible conflicting subscriptions, return.

pkg/operator/controller/status/controller.go
```diff
@@ -72,6 +73,16 @@ var (
...
+
+	// ossmSubscriptions lists the package names for all OLM subscriptions that are known to conflict with the
+	// subscription to OSSM3 created by the Ingress Operator.
+	ossmSubscriptions = sets.New[string](
+		"sailoperator",
+		"servicemeshoperator",
+		"servicemeshoperator3",
+	)
+
+	subscriptionCache cache.Cache
```
candita, line 85: Instead of a global variable, I was expecting to see `subscriptionCache` as a field of `reconciler` in `func New`.  If you make other changes, can you change this too?

pkg/operator/controller/status/controller.go
```diff
@@ -864,3 +984,38 @@ func joinConditions(currCond, newCond configv1.ClusterOperatorStatusCondition) c
...
+	if len(bSplit) != 4 {
+		return 0, fmt.Errorf("%q does not match expected format", b)
+	}
+	bName = bSplit[0]
+	bX, _ = strconv.Atoi(bSplit[1][1:]) // X has format "v%d", so [1:] cuts out the "v"
+	bY, _ = strconv.Atoi(bSplit[2])
+	bZ, _ = strconv.Atoi(bSplit[3])
+	if aName != bName {
+		return 0, fmt.Errorf("%q and %q are different packages. cannot compare version numbers", a, b)
+	}
```
candita, line 1013: It would be better to check for the same package name before we even call this function.

rfredette: This is more of a safety check. We shouldn't hit this in actual use.

pkg/operator/controller/status/controller.go
```diff
@@ -156,6 +165,26 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+			return ossmSubscriptions.Has(subscription.Spec.Package)
+		})
+
+		// The status controller needs to be aware of conflicting OLM subscriptions in any namespace. In order to
+		// prevent ballooning the main cache to watch all namespaces on all its informers, create a separate cache for
+		// tracking OLM subscriptions across all namespaces.
+		if reconciler.subscriptionCache, err = cache.New(mgr.GetConfig(), cache.Options{}); err != nil {
+			return nil, err
+		}
+		mgr.Add(reconciler.subscriptionCache)
```
candita, line 183: I expect this is a part of the initialization you need to do before calling `controller.New` in line 100.

pkg/operator/controller/status/controller.go
rfredette: Yeah. Currently, `compareVersionNums` only used in the case where `subscription.Spec.Package` is `servicemeshoperator3`, and it's compared against the expected version, which will be `servicemeshoperator3.v<version>`. It's possible someone could change the expected version, since it's a field set in the ingress operator deployment, but that's not really supported.

pkg/operator/controller/status/controller.go
```diff
@@ -570,6 +631,64 @@ func computeGatewayAPICRDsDegradedCondition(state operatorState) configv1.Cluste
...
+			}
+			if subscription.Status.InstalledCSV == "" {
+				// The subscription hasn't finished its install. We will get another reconcile request once the
+				// installation is complete, so we can ignore this for now.
+				continue
+			}
+			versionDiff, err := compareVersionNums(subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion)
+			switch {
+			case err != nil:
+				warnings = append(warnings, fmt.Sprintf("failed to compare installed OSSM version to expected: %v", err))
```
candita, line 661: Can you add a unit test for this?

pkg/operator/controller/status/controller.go
rfredette: It works either way, but I moved this before `controller.New` for better readability

pkg/operator/controller/status/controller_test.go
```diff
@@ -793,7 +796,86 @@ func Test_computeOperatorDegradedCondition(t *testing.T) {
...
+					sub("sailoperator", "sailoperator", "sailoperator.v1.0.0", false),
+				},
+				shouldInstallOSSM:                 true,
+				expectedGatewayAPIOperatorVersion: "servicemeshoperator3.v3.1.0",
+			},
+			expectCondition: configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorDegraded,
+				Status:  configv1.ConditionTrue,
+				Reason:  "GatewayAPIInstallConflict",
+				Message: "Package sailoperator from subscription foo/sailoperator prevents enabling operator-managed Gateway API. Uninstall foo/sailoperator to enable.",
```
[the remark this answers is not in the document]
alebedev87: I didn't mean to check the `installedCVS`. What I meant is the wording: "enabling" and "to enable" work only in the case when the CIO managed subscription tries to be created **after** a conflicting subscription. There is still a case when, for instance, a `sailoperator` is trying t be installed after the CIO managed OSSM was already installed. In that case the wording "to enable" is inaccurate because the CIO is already enabled.

candita:
e2e-gcp-operator is failing with the designated override reason, so I will override it here.

/override ci/prow/e2e-gcp-operator

PR #1269, opened 2025-08-14
OCPBUGS-60492: Skip Gateway Service DNS record creation on platforms with ClusterHostedDNS is configured

sadasu:
When in-cluster DNS/custom-DNS is configured on AWS and GCP, skip creating Gateway Service DNS record.

### DIFF
--- a/pkg/operator/controller/gateway-service-dns/controller.go
+++ b/pkg/operator/controller/gateway-service-dns/controller.go
@@ -247,6 +247,9 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 		if dnsrecord.ManageDNSForDomain(domain, infraConfig.Status.PlatformStatus, dnsConfig) {
 			dnsPolicy = iov1.ManagedDNS
 		}
+		if checkClusterHostedDNS(infraConfig) {
+			dnsPolicy = iov1.UnmanagedDNS
+		}
 		_, _, err := dnsrecord.EnsureDNSRecord(r.client, name, labels, ownerRef, domain, dnsPolicy, service)
 		errs = append(errs, err)
 	}
@@ -281,4 +284,24 @@ func (r *reconciler) deleteStaleDNSRecordsForGateway(ctx context.Context, gatewa
 		errs = append(errs, dnsrecord.DeleteDNSRecord(r.client, name))
 	}
 	return errs
+}
+
+// checkClusterHostedDNS returns true if the platform supports in-cluster DNS and if
+// that DNS solution is currently enabled in place of the Cloud provider's default DNS.
+func checkClusterHostedDNS(infraConfig *configv1.Infrastructure) bool {
+	status := infraConfig.Status.PlatformStatus
+	switch status.Type {
+	case configv1.AWSPlatformType:
+		if status.AWS != nil && status.AWS.CloudLoadBalancerConfig != nil && status.AWS.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
+			return true
+		}
+		return false
+	case configv1.GCPPlatformType:
+		if status.GCP != nil && status.GCP.CloudLoadBalancerConfig != nil && status.GCP.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
+			return true
+		}
+		return false
+	default:
+		return false
+	}
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -199,6 +199,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
+	if !checkPlatformSupport(infraConfig) {
+		log.Info("infrastructure 'cluster' has unsupported configuration; reconciliation will be skipped")
+		return reconcile.Result{}, nil
+	}
+
 	domains := getGatewayHostnames(&gateway)
 	var errs []error
 	errs = append(errs, r.ensureDNSRecordsForGateway(ctx, &gateway, &service, domains.List(), infraConfig, dnsConfig)...)
```
Miciah, line 209: It can be useful to create the DNSRecord CR, but with `dnsManagementState: Unmanaged`, to serve as a reference for what DNS record the cluster-admin is _expected_ to create.  So I would suggest still calling `ensureDNSRecordsForGateway` but changing it to to set `dnsManagementState: Unmanaged` when the platform is not supported.  Does that make sense?

candita:
It would be good to investigate this CI failure.  It seems related to me, being GCP and DNS.

> === NAME  TestAll/parallel/TestInternalLoadBalancerGlobalAccessGCP
>     operator_test.go:1295: Expected conditions: map[Admitted:True Available:True DNSManaged:True DNSReady:True LoadBalancerManaged:True LoadBalancerReady:True]
>          Current conditions: map[Admitted:True Available:False DNSManaged:True DNSReady:False Degraded:True DeploymentAvailable:True DeploymentReplicasAllAvailable:True DeploymentReplicasMinAvailable:True DeploymentRollingOut:False EvaluationConditionsDetected:False LoadBalancerManaged:True LoadBalancerProgressing:False LoadBalancerReady:False Progressing:False Upgradeable:True]
> ...
>   "message": "One or more other status conditions indicate a degraded state: LoadBalancerReady=False (SyncLoadBalancerFailed: The service-controller component is reporting SyncLoadBalancerFailed events like: Error syncing load balancer: failed to ensure load balancer: googleapi: Error 400: Resource 'projects/XXXXXXXXXXXXXXXXXXXXXX/zones/us-central1-c/instances/ci-op-b4v9j4c3-76b3b-j7ls4-worker-c-sz59r' is expected to be in the subnetwork 'projects/XXXXXXXXXXXXXXXXXXXXXX/regions/us-central1/subnetworks/ci-op-b4v9j4c3-76b3b-j7ls4-master-subnet' but is in the subnetwork 'projects/XXXXXXXXXXXXXXXXXXXXXX/regions/us-central1/subnetworks/ci-op-b4v9j4c3-76b3b-j7ls4-worker-subnet'., wrongSubnetwork\nThe cloud-controller-manager logs may contain more details.)"
>             },

pkg/operator/controller/gateway-service-dns/controller.go
Miciah: > My reasoning for putting the check here vs `ensureDNSRecordsForGateway` was that this check would give the same outcome for each domain within `ensureDNSRecordsForGateway`.

If you're concerned about the performance impact, you can call `checkClusterHostedDNS` (previously `checkPlatformSupport`) in `Reconcile` and pass the Boolean result to `ensureDNSRecordsForGateway`.

pkg/operator/controller/gateway-service-dns/controller.go
```diff
@@ -247,6 +247,9 @@ func (r *reconciler) ensureDNSRecordsForGateway(ctx context.Context, gateway *ga
 		if dnsrecord.ManageDNSForDomain(domain, infraConfig.Status.PlatformStatus, dnsConfig) {
 			dnsPolicy = iov1.ManagedDNS
 		}
+		if checkClusterHostedDNS(infraConfig) {
+			dnsPolicy = iov1.UnmanagedDNS
+		}
```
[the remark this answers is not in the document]
sadasu: custom-dns is enabled by the customer when they do not want OpenShift to use the cloud provider's default DNS. So, setting the `dnsPolicy` to `iov1.UnmanagedDNS` so that the Gateway API controller does not configure the cloud's DNS.

PR #1272, opened 2025-08-25
Set servicemesh operator catalog for OKD

khdo:


pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -49,6 +53,13 @@ func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context)
 
 // desiredSubscription returns the desired subscription.
 func desiredSubscription(name types.NamespacedName) (*operatorsv1alpha1.Subscription, error) {
+	packageName := "servicemeshoperator"
+	catalogSource := "redhat-operators"
+
+	if OKD {
+		packageName = "servicemeshoperator3"
+		catalogSource = "okderators"
+	}
```
Prashanth684, line 62: does this require OKDerators catalog to be installed? what happens if you trigger this feature without installing OKDerators?

candita:
Thanks for the update.  Once this has proved successful and merges, we can investigate removing the OKD skip in:
https://github.com/openshift/origin/blob/main/test/extended/router/gatewayapicontroller.go#L72

PR #1275, opened 2025-08-27
no-jira: Ingress operator must search for the GCP Custom Endpoints Installer feature gate

barbacbd:
Update the reference from FeatureGateGCPCustomAPIEndpoints to FeatureGateGCPCustomAPIEndpointsInstall due to new procedures. This indicates that the feature gate originates as an OpenShift installer feature gate.


barbacbd:
Closing in favor of a solution that does not require explicit api endpoint overrides for each new service.

PR #1277, opened 2025-09-05
use safe tls-cipher-suites

lance5890:
change  the `TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256` and `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256`  insecure cipher to more safer suites


candita:
@lance5890 what is the reasoning for making these changes?  Was there an issue with the metrics that you were trying to solve?

candita:
@lance5890 the config you're changing is specifically for Prometheus's metric retrieval, is this PR to fix an issue with Prometheus?  Please let us know, and update the commit message with details.   

This PR is pending acceptance while waiting for your answer and we will drop review soon without an answer.

candita:
@lance5890 we are going to close this because we think the change should be made in https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go.  If you'd like to file an issue there, that would be very helpful.

PR #1284, opened 2025-09-24
[WIP][NE-2131] OTE - Implement initial framework and tests

rikatz:
This change implements  the OTE (Openshift Tests Extension) on Cluster Ingress Operator, and migrates 2 tests suites to it:
* GatewayAPI - Test assertion for GatewayAPI capability
* GatewayAPI Controller - Test for GatewayAPI controller + Sail integrations

As the tests were originally based on openshift/origin, a part of the framework had to be migrated/adapted to the new approach. This was necessary to avoid importing the whole openshift/origin and kubernetes/kubernetes tree just for test helper functions.

There is still pending work:
* Generate the binary and add to the container image - To be called by openshift/origin tests
* Migrate / add the capability of non privileged client to the client framework

### DIFF
--- a/cmd/ingress-operator-ext-tests/main.go
+++ b/cmd/ingress-operator-ext-tests/main.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/openshift-eng/openshift-tests-extension/pkg/cmd"
 	e "github.com/openshift-eng/openshift-tests-extension/pkg/extension"
+	et "github.com/openshift-eng/openshift-tests-extension/pkg/extension/extensiontests"
 	g "github.com/openshift-eng/openshift-tests-extension/pkg/ginkgo"
 	"github.com/spf13/cobra"
 
@@ -51,6 +52,11 @@ func main() {
 	// Initialize the environment before running any tests.
 	specs.AddBeforeAll(func() {
 		framework.Init()
+	})
+
+	// TODO: Remove this, it is just to verify that openshift-origin is able to fetch OTE tests
+	specs = specs.Walk(func(spec *et.ExtensionTestSpec) {
+		spec.Name = spec.Name + " using OTE"
 	})
 
 	ext.AddSpecs(specs)
--- a/.openshift-tests-extension/openshift_payload_cluster-ingress-operator.json
+++ b/.openshift-tests-extension/openshift_payload_cluster-ingress-operator.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure required CRDs should already be installed",
+    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure required CRDs should already be installed using OTE",
     "labels": {},
     "resources": {
       "isolation": {}
@@ -10,7 +10,7 @@
     "environmentSelector": {}
   },
   {
-    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure existing CRDs can not be deleted",
+    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure existing CRDs can not be deleted using OTE",
     "labels": {},
     "resources": {
       "isolation": {}
@@ -20,7 +20,7 @@
     "environmentSelector": {}
   },
   {
-    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure existing CRDs can not be updated",
+    "name": "[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io] Verify Gateway API CRDs and ensure existing CRDs can not be updated using OTE",
     "labels": {},
     "resources": {
       "isolation": {}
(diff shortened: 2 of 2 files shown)


test/extended/framework/cluster.go
```diff
@@ -0,0 +1,171 @@
+package framework
+
+// This file is based on operator-framework-controller client init, but without
```
rikatz, line 3: This comment is wrong, need to fix it

test/extended/framework/client.go
```diff
@@ -0,0 +1,358 @@
+package framework
```
rikatz, line 1: Add here where the client came from (this is a smaller set from the original openshift/origin repo, without the Ginkgo specifics)

test/extended/framework/client.go
```diff
@@ -0,0 +1,358 @@
...
+// AdminKubeClient provides a Kubernetes client for the cluster admin user.
+func (c *CLI) AdminKubeClient() kubernetes.Interface {
+	return kubernetes.NewForConfigOrDie(c.restCfg)
+}
+
+func (c *CLI) AdminOperatorClient() operatorv1client.Interface {
+	return operatorv1client.NewForConfigOrDie(c.restCfg)
+}
+
+// TODO: this should be a user and not an admin, need to check!
```
rikatz, line 183: This is the TODO to use the regular user and not the admin user

test/extended/specs/gatewayapi.go
```diff
@@ -0,0 +1,157 @@
...
+var _ = g.Describe("[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io]", func() {
+	defer g.GinkgoRecover()
+
+	var (
+		oc  *framework.CLI
+		err error
+	)
+
+	g.BeforeEach(func() {
+		oc, err = framework.NewCLI("gateway-api", admissionapi.LevelBaseline, framework.Get().K8sClient, framework.Get().RestCfg, framework.Get().IsOpenShift)
```
rikatz, line 25: while it may be tempting to abstract this and add this whole "framework.XXX" inside the client creation, I would prefer to receive these as parameters of NewCLI so this can be eventually reused

test/extended/specs/gatewayapi.go
```diff
@@ -0,0 +1,157 @@
+package specs
```
rikatz, line 1: Add where this file came from, so there may be a comparision (maybe a permalink from the original repo)

test/extended/specs/gatewayapicontroller.go
```diff
@@ -0,0 +1,743 @@
+package specs
```
rikatz, line 1: Add where this file came from, a permalink, etc

cmd/ingress-operator-ext-tests/main.go
```diff
@@ -0,0 +1,70 @@
+/*
```
rikatz, line 1: Need to add a README so others can know how to use it

test/extended/framework/client.go
```diff
@@ -0,0 +1,358 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	admissionapi "k8s.io/pod-security-admission/api"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1client "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+)
+
+type CLI struct {
```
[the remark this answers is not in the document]
rikatz: And https://github.com/openshift/operator-framework-operator-controller/blob/main/openshift/tests-extension/pkg/helpers/dump.go

Makefile
```diff
@@ -24,6 +24,13 @@ endif
 build: generate
 	$(GO_BUILD_RECIPE)
 
+.PHONY: build-tests-ext
```
rikatz, line 27: Dockerfile build example: https://github.com/openshift/machine-api-operator/blob/255bf58e9af927944f20495fd53eb65e09dd9224/Dockerfile#L16

cmd/ingress-operator-ext-tests/main.go
```diff
@@ -0,0 +1,70 @@
...
+This command is used to run the OLMv1 tests extension for OpenShift.
+It registers the OLMv1 tests with the OpenShift Tests Extension framework
+and provides a command-line interface to execute them.
+
+For further information, please refer to the documentation at:
+https://github.com/openshift-eng/openshift-tests-extension/blob/main/cmd/example-tests/main.go
+*/
+package main
```
rikatz, line 9: Execute a test with https://docs.ci.openshift.org/docs/how-tos/multi-pr-presubmit-testing/

cmd/ingress-operator-ext-tests/main.go
```diff
@@ -0,0 +1,70 @@
...
+	ext.AddSuite(
+		e.Suite{
+			Name:    "cluster-ingress-operator/serial",
+			Parents: []string{"openshift/conformance/serial"},
+			Qualifiers: []string{
+				`(name.contains("[Serial]") || name.contains("[Slow]"))`,
+			},
+		})
+
+	specs, err := g.BuildExtensionTestSpecsFromOpenShiftGinkgoSuite()
```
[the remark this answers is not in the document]
rikatz: skipping this for now to keep the test name stable

test/extended/specs/gatewayapi.go
```diff
@@ -0,0 +1,157 @@
...
+	g "github.com/onsi/ginkgo/v2"
+	o "github.com/onsi/gomega"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	admissionapi "k8s.io/pod-security-admission/api"
+
+	"github.com/openshift/cluster-ingress-operator/test/extended/framework"
+)
+
+var _ = g.Describe("[sig-network][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io]", func() {
```
lihongan, line 16:
```suggestion
var _ = g.Describe("[sig-network-edge][OCPFeatureGate:GatewayAPI][Feature:Router][apigroup:gateway.networking.k8s.io]", func() {
```
It is to keep consistent with the "GatewayAPIController" tests. 
I used wrong tag when adding tests to Origin repo.

PR #1287, opened 2025-09-30
NE-2737: Add support for TLS curves in router deployment configuration

richardsonnick:
To support new PQC curves, let's add a config for explicitly setting the supported groups. 

ROUTER_CURVES is piped through OpenShift/router to HAProxy where it will be used by `ssl-default-bind-curves`

For more context see: https://github.com/openshift/router/pull/678

### DIFF
--- a/manifests/03-cluster-operator.yaml
+++ b/manifests/03-cluster-operator.yaml
@@ -30,3 +30,25 @@ status:
     - group: ingress.operator.openshift.io
       namespace: openshift-ingress-operator
       resource: dnsrecords
+    - group: rbac.authorization.k8s.io
+      name: openshift-ingress-operator
+      resource: clusterroles
+    - group: rbac.authorization.k8s.io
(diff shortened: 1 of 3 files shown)


vendor/github.com/openshift/api/config/v1/types_tlssecurityprofile.go
alebedev87: We don't directly modify the vendored modules. There should be a corresponding `go.mod`/`go.sum` bump for this change.

davidesalerno (changes requested):

davidesalerno:
@richardsonnick I think that we are missing the bump of the API version

richardsonnick:
> @richardsonnick I think that we are missing the bump of the API version
Yep, bumping this here: https://github.com/openshift/cluster-ingress-operator/pull/1473

richardsonnick:
Fixing blocking tests...

PR #1299, opened 2025-10-31
OCPBUGS-62627: cluster operator ingress reported Progressing=True wit…

davidesalerno:
…h reason=Reconciling for a node reboot

This PR will avoid Cluster Ingress Operator reporting `Progressing=True` condition if there is a node reboot or scale up as per Cluster Operator new requirement ([link](https://github.com/openshift/api/blob/61248d910ff74aef020492922d14e6dadaba598b/config/v1/types_cluster_operator.go#L163-L164)).

The Cluster Ingress Operator will continue to report `DeploymentRollingOut=true` condition with a different `Reason` if there is a node reboot or scale up and we will exclude these reasons (`ReplicasStabilizing` and `PodsStarting`) from the ones causing the operator `Progressing=True` condition.

In order to detect the a Deployment has been fully rolled out in the past we will use `Deployment` condition `Progressing` with reason `NewReplicaSetAvailable` as similarly done in library-go (see [PR](https://github.com/openshift/library-go/pull/2034))

### DIFF
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -47,7 +48,24 @@ type expectedCondition struct {
 	// or else the condition is not checked.
 	ifConditionsTrue []string
 	gracePeriod      time.Duration
+	// ignoreReasons specifies condition reasons that should be ignored when
+	// determining if the expected condition is met. This allows distinguishing
+	// between significant state changes (configuration rollouts) and transient
+	// infrastructure events (node reboots, pod restarts).
+	// ignoreReasons are only consulted when the condition exists and its Status
+	// differs from expected.status; they are not consulted if the condition is
+	// entirely missing.
+	ignoreReasons []string
 }
+
+const (
+	ReasonDeploymentRollingOut = "DeploymentRollingOut"
+	ReasonNotRollingOut        = "DeploymentNotRollingOut"
+	ReasonPodsStarting         = "PodsStarting"
+	ReasonReplicasStabilizing  = "ReplicasStabilizing"
+
+	deploymentNewRSAvailableReason = "NewReplicaSetAvailable"
+)
 
 // syncIngressControllerStatus computes the current status of ic and
 // updates status upon any changes since last sync.
@@ -332,7 +350,7 @@ func checkConditions(expectedConds []expectedCondition, conditions []operatorv1.
 		if !haveCondition {
 			continue
 		}
-		if condition.Status == expected.status {
+		if condition.Status == expected.status || slices.Contains(expected.ignoreReasons, condition.Reason) {
 			continue
 		}
 		failedPredicates := false
@@ -505,12 +523,13 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
 // of expected or available replicas.
 // See Reference: https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go
 func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	// If have replicas is less than want replicas, then we are waiting for replicas to be updated.
+	// If updated replicas is less than desired replicas, then we are waiting for updated replicas to be created.
 	if deployment.Spec.Replicas != nil && deployment.Status.UpdatedReplicas < *deployment.Spec.Replicas {
+		reason := computeDeploymentRollingOutReason(deployment)
 		return operatorv1.OperatorCondition{
 			Type:   IngressControllerDeploymentRollingOutConditionType,
 			Status: operatorv1.ConditionTrue,
-			Reason: "DeploymentRollingOut",
+			Reason: reason,
 			Message: fmt.Sprintf(
 				"Waiting for router deployment rollout to finish: %d out of %d new replica(s) have been updated...\n",
 				deployment.Status.UpdatedReplicas, *deployment.Spec.Replicas),
@@ -518,10 +537,11 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
 	}
 	// If have replicas greater than updated replicas, then we are waiting for old replicas to terminate.
 	if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
+		reason := computeDeploymentRollingOutReason(deployment)
 		return operatorv1.OperatorCondition{
 			Type:   IngressControllerDeploymentRollingOutConditionType,
 			Status: operatorv1.ConditionTrue,
-			Reason: "DeploymentRollingOut",
+			Reason: reason,
 			Message: fmt.Sprintf(
 				"Waiting for router deployment rollout to finish: %d old replica(s) are pending termination...\n",
 				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
@@ -529,10 +549,11 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
 	}
 	// If available replicas less than updated replicas, then we are waiting for updated replicas to become available.
 	if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
+		reason := computeDeploymentRollingOutReason(deployment)
 		return operatorv1.OperatorCondition{
 			Type:   IngressControllerDeploymentRollingOutConditionType,
 			Status: operatorv1.ConditionTrue,
-			Reason: "DeploymentRollingOut",
+			Reason: reason,
 			Message: fmt.Sprintf(
 				"Waiting for router deployment rollout to finish: %d of %d updated replica(s) are available...\n",
 				deployment.Status.AvailableReplicas, deployment.Status.UpdatedReplicas),
@@ -542,9 +563,58 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
 	return operatorv1.OperatorCondition{
 		Type:    IngressControllerDeploymentRollingOutConditionType,
 		Status:  operatorv1.ConditionFalse,
-		Reason:  "DeploymentNotRollingOut",
+		Reason:  ReasonNotRollingOut,
 		Message: "Deployment is not actively rolling out",
 	}
+}
+
+// computeDeploymentRollingOutReason computes the Reason for the DeploymentRollingOut
+// condition by checking the deployment's "NewReplicaSetAvailable" condition, pod
+// replica counts, and generation metadata to distinguish between an ingress
+// controller configuration rollout and an infrastructure-driven change (e.g.
+// node reboot, scale up/down).
+func computeDeploymentRollingOutReason(deployment *appsv1.Deployment) string {
+	// Check if the deployment has the "NewReplicaSetAvailable" condition.
+	// This indicates that Kubernetes has successfully completed a rollout.
+	hasNewReplicaSetAvailable := false
+	for _, cond := range deployment.Status.Conditions {
+		if cond.Type == appsv1.DeploymentProgressing &&
+			cond.Status == corev1.ConditionTrue &&
+			cond.Reason == deploymentNewRSAvailableReason {
+			hasNewReplicaSetAvailable = true
+			break
+		}
+	}
+
+	var (
+		wantReplicas      = int32(1)
+		haveReplicas      = deployment.Status.Replicas
+		updatedReplicas   = deployment.Status.UpdatedReplicas
+		availableReplicas = deployment.Status.AvailableReplicas
+	)
+
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+
+	currentGen := deployment.Generation
+	observedGen := deployment.Status.ObservedGeneration
+	if hasNewReplicaSetAvailable && currentGen == observedGen && updatedReplicas == wantReplicas {
+		if haveReplicas != wantReplicas {
+			return ReasonReplicasStabilizing
+		}
+	}
+
+	// If rollout was previously completed and now only availability is lagging,
+	// treat it as infrastructure-driven pod startup.
+	if hasNewReplicaSetAvailable &&
+		currentGen == observedGen &&
+		updatedReplicas == wantReplicas &&
+		haveReplicas == wantReplicas &&
+		availableReplicas < wantReplicas {
+		return ReasonPodsStarting
+	}
+	return ReasonDeploymentRollingOut
 }
 
 // computeIngressDegradedCondition computes the ingresscontroller's "Degraded"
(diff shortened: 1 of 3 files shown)


lihongan:
per-merge tested and passed
/label qe-approved
/verified by @lihongan 
```
## watching co/ingress status while rebooting the node

## with the fix:
$ oc get co/ingress -w
NAME      VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.21.0-0-2025-11-17-075629-test-ci-ln-scc84qk-latest   True        False         False      23m     


## without the fix
$ oc get co/ingress -w
NAME      VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.21.0-0.nightly-2025-11-13-042845   True        False         False      34m     
ingress   4.21.0-0.nightly-2025-11-13-042845   True        True          False      35m     ingresscontroller "default" is progressing: IngressControllerProgressing: One or more status conditions indicate progressing: DeploymentRollingOut=True (DeploymentRollingOut: Waiting for router deployment rollout to finish: 1 of 2 updated replica(s) are available......
ingress   4.21.0-0.nightly-2025-11-13-042845   True        False         False      36m     

```

pkg/operator/controller/ingress/status_test.go
```diff
@@ -514,7 +518,7 @@ func Test_computeDeploymentRollingOutCondition(t *testing.T) {
 			expectMessageContains: "Deployment is not actively rolling out",
 		},
 		{
-			name:                  "Router pod replicas have/updated < want",
+			name:                  "Router pod replicas have/updated < wantGrace",
```
bentito: It seems like from the code under test that this should be named `wantReplicas` or `replicasWanted` to avoid confusion with something like a grace period (time) which I don't think is being tested?

pkg/operator/controller/ingress/status.go
```diff
@@ -765,6 +847,12 @@ func computeIngressProgressingCondition(conditions []operatorv1.OperatorConditio
 		{
 			condition: IngressControllerDeploymentRollingOutConditionType,
 			status:    operatorv1.ConditionFalse,
+			// Ignore infrastructure-driven deployment rollouts when computing
+			// the IngressController's Progressing condition.
+			ignoreReasons: []string{
+				ReasonReplicasStabilizing.String(), // Node reboots, pod evictions
```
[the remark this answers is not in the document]
davidesalerno: In this case the user should change the Ingress CR to set the nuber of replicas would like to have and so the check `deployment.Status.ObservedGeneration == deployment.Generation` we are doing into `computeDeploymentRollingOutStatusAndReason` will fail and the operator status `Progressing=True` will be maintained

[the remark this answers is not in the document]
bentito: I see what you mean about the generation check failing initially, but I think that window is super brief (just until the deployment controller observes the spec update).

The bulk of the time is usually spent waiting for the new pods to actually become ready. In that state (ObservedGeneration == Generation but AvailableReplicas < Spec.Replicas), the logic returns ReasonPodsStarting. Since ReasonPodsStarting is in the ignoreReasons list, the operator will flip back to Progressing=False almost immediately, even while the new pods are still spinning up.

If we want users to see Progressing=True for the duration of the scale-up (which I think they'd expect), we might need to reconsider ignoring ReasonPodsStarting or find a way to distinguish 'pods starting due to explicit scale' vs 'pods starting due to restart'

pkg/operator/controller/ingress/status.go
```diff
@@ -499,40 +532,56 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
+		if cond.Type == appsv1.DeploymentProgressing &&
+			cond.Status == corev1.ConditionTrue &&
+			cond.Reason == deploymentNewRSAvailableReason {
+			hasNewReplicaSetAvailable = true
+
+			break
+		}
+	}
+
+	// If have replicas is less than wantGrace replicas, then we are waiting for replicas to be updated.
```
Miciah: What is "wantGrace replicas"?  Do you just mean "wanted replicas"?
```suggestion
	// If updated replicas is less than desired replicas, then we are waiting for updated replicas to be created.
```

pkg/operator/controller/ingress/status.go
```diff
@@ -542,11 +591,49 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
+			return ReasonPodsStarting.String()
+		}
+	}
+
+	// Additional safety check: Even without NewReplicaSetAvailable, if everything
+	// matches except readiness, treat it as pods starting up
+	if deployment.Status.ObservedGeneration == deployment.Generation &&
+		updatedReplicas == wantReplicas &&
+		haveReplicas == wantReplicas &&
+		(readyReplicas < wantReplicas || availableReplicas < wantReplicas) {
```
Miciah: `availableReplicas < readyReplicas` should be impossible, so you should be able to simplify `readyReplicas < wantReplicas || availableReplicas < wantReplicas` as `availableReplicas < wantReplicas`.  Is there any reason we cannot do that?

pkg/operator/controller/ingress/status.go
```diff
@@ -542,11 +591,49 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
+		return ReasonPodsStarting.String()
+	}
+	return ReasonDeploymentRollingOut.String()
+}
```
Miciah: `computeDeploymentRollingOutStatusAndReason` can be simplified to the following, tests still pass, and the logic is much clearer in my opinion:
```go
func computeDeploymentRollingOutReason(deployment *appsv1.Deployment) string {
	if deployment.Status.ObservedGeneration != deployment.Generation {
		return ReasonDeploymentRollingOut
	}

	wantReplicas := int32(1)
	if deployment.Spec.Replicas != nil {
		wantReplicas = *deployment.Spec.Replicas
	}

	switch {
	case deployment.Status.UpdatedReplicas != wantReplicas:
		return ReasonDeploymentRollingOut
	case deployment.Status.AvailableReplicas < wantReplicas:
		return ReasonPodsStarting
	case deployment.Status.Replicas != wantReplicas:
		return ReasonReplicasStabilizing
	default:
		return ReasonDeploymentRollingOut
	}
}
```
Is this logic correct, or are we missing some test case where it would fail?  

For that matter, we can refactor `computeDeploymentRollingOutCondition` as follows:
```go
func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
	if deployment.Spec.Replicas != nil && deployment.Status.UpdatedReplicas < *deployment.Spec.Replicas {
			Type:   IngressControllerDeploymentRollingOutConditionType,
			Status: operatorv1.ConditionTrue,
			Reason: ReasonDeploymentRollingOut,
			Message: fmt.Sprintf(
				"Waiting for router deployment rollout to finish: %d out of %d new replica(s) have been updated...\n",
				deployment.Status.UpdatedReplicas, *deployment.Spec.Replicas),
		}
	}
	if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
			Type:   IngressControllerDeploymentRollingOutConditionType,
			Status: operatorv1.ConditionTrue,
			Reason: ReasonReplicasStabilizing,
			Message: fmt.Sprintf(
				"Waiting for router deployment rollout to finish: %d old replica(s) are pending termination...\n",
				deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
		}
	}
	if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
			Type:   IngressControllerDeploymentRollingOutConditionType,
			Status: operatorv1.ConditionTrue,
			Reason: ReasonPodsStarting,
			Message: fmt.Sprintf(
				"Waiting for router deployment rollout to finish: %d of %d updated replica(s) are available...\n",
				deployment.Status.AvailableReplicas, deployment.Status.UpdatedReplicas),
		}
	}

		Type:    IngressControllerDeploymentRollingOutConditionType,
		Reason:  ReasonNotRollingOut,
		Message: "Deployment is not actively rolling out",
	}
}
```
`hasNewReplicaSetAvailable` and `computeDeploymentRollingOut(StatusAnd)Reason` seem like unnecessary complications.  If they are necessary, can you add a test case to illustrate why, and maybe some code comments to explain as well?

pkg/operator/controller/ingress/status.go
```diff
@@ -542,11 +591,49 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
+		availableReplicas = deployment.Status.AvailableReplicas
+		readyReplicas     = deployment.Status.ReadyReplicas
+	)
+
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+
+	if hasNewReplicaSetAvailable && deployment.Status.ObservedGeneration == deployment.Generation &&
+		updatedReplicas == wantReplicas {
```
Miciah: Would the condition be clearer without the line wrapping?
```suggestion
	if hasNewReplicaSetAvailable && deployment.Status.ObservedGeneration == deployment.Generation && updatedReplicas == wantReplicas {
```
or to avoid making the line longer
```suggestion
	currentGen := deployment.Generation
	observedGen := deployment.Status.ObservedGeneration
	if hasNewReplicaSetAvailable && currentGen == observedGen && updatedReplicas == wantReplicas {
```

pkg/operator/controller/ingress/status.go
```diff
@@ -499,40 +532,56 @@ func computeDeploymentReplicasAllAvailableCondition(deployment *appsv1.Deploymen
...
 // See Reference: https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go
 func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
-	// If have replicas is less than want replicas, then we are waiting for replicas to be updated.
+	// Check if the deployment has the "NewReplicaSetAvailable" condition.
+	// This indicates that Kubernetes has successfully completed a rollout.
+	hasNewReplicaSetAvailable := false
+	for _, cond := range deployment.Status.Conditions {
+		if cond.Type == appsv1.DeploymentProgressing &&
+			cond.Status == corev1.ConditionTrue &&
+			cond.Reason == deploymentNewRSAvailableReason {
```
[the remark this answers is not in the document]
davidesalerno: `k8s.io/kubernetes `is not currently vendored in this project, and importing it as a library dependency is generally discouraged by the Kubernetes project itself due to its large transitive dependency graph. Pulling in that module just for a single string constant ("NewReplicaSetAvailable") would significantly increase the dependency footprint. I think a local constant is the more pragmatic choice here,

pkg/operator/controller/ingress/status.go
```diff
@@ -542,11 +591,49 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
 	}
 }
 
+// computeDeploymentRollingOutStatusAndReason computes the ingresscontroller's Status and Reason for DeploymentRollingOut
+// condition type checking if the deployment has the "NewReplicaSetAvailable" condition and the status of the pod replicas
+// and the generation metadata, in order to detect if the deployment is rolling out for am ingresscontroller configuration
+// change or due to an infrastructure change (i.e. node reboot due to update or scale up or down)
+func computeDeploymentRollingOutStatusAndReason(hasNewReplicaSetAvailable bool, deployment *appsv1.Deployment) string {
+	var (
+		wantReplicas      = int32(0)
```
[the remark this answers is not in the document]
davidesalerno: Thanks for the explanation, I updated the code accordingly

pkg/operator/controller/ingress/status.go
```diff
@@ -542,11 +563,59 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
+	)
+
+	if deployment.Spec.Replicas != nil {
+		wantReplicas = *deployment.Spec.Replicas
+	}
+
+	currentGen := deployment.Generation
+	observedGen := deployment.Status.ObservedGeneration
+	if hasNewReplicaSetAvailable && currentGen == observedGen && updatedReplicas == wantReplicas {
+		if haveReplicas != wantReplicas {
```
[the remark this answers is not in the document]
Miciah: `haveReplicas` is the total number of replicas.  It is impossible to have fewer _total_ replicas (`haveReplicas`) than there are _updated_ replicas (`updatedReplicas`).  In other words, `haveReplicas >= updatedReplicas` is an invariant.  

This logic is saying this: If we have rolled out the desired number of updated replicas (`updatedReplicas == wantReplicas`), do we have the wrong number of replicas (`haveReplicas != wantReplicas`)?  If so, then we must still be "stabilizing", meaning we are waiting for some old replicas to be deleted.  

So, `haveReplicas != wantReplicas` is the correct condition, and it could even be replaced with `haveReplicas > wantReplicas` if that would be clearer, but using `haveReplicas < wantReplicas` would mean we would never reach the `return ReasonReplicasStabilizingstabilizing` statement.

pkg/operator/controller/ingress/status_test.go
```diff
@@ -577,6 +581,54 @@ func Test_computeDeploymentRollingOutCondition(t *testing.T) {
...
+			expectStatus:          operatorv1.ConditionTrue,
+			replicasHave:          pointer.Int32(4),
+			replicasWanted:        pointer.Int32(4),
+			replicasUpdated:       pointer.Int32(4),
+			replicasAvailable:     pointer.Int32(1),
+			expectMessageContains: "Waiting for router deployment rollout to finish: 1 of 4 updated replica(s) are available...\n",
+			expectReasonContains:  ReasonPodsStarting,
+			differentGeneration:   false,
+			hasNewReplicaSet:      true,
+		},
```
Miciah: "Router pods replicas available < updated and generation is equal to observedGeneration" and "Router pod replicas have/updated < replicasWanted, generation is equal to observedGeneration and NewReplicaSet created" are identical test cases.

pkg/operator/controller/ingress/status.go
```diff
@@ -543,11 +563,56 @@ func computeDeploymentRollingOutCondition(deployment *appsv1.Deployment) operato
...
+			return ReasonReplicasStabilizing
+		}
+	}
+
+	// Additional safety check: Even without NewReplicaSetAvailable, if everything
+	// matches except availability, treat it as pods starting up.
+	if currentGen == observedGen && updatedReplicas == wantReplicas && haveReplicas == wantReplicas && availableReplicas < wantReplicas {
+		return ReasonPodsStarting
```
bentito: I think coderabbit is still on track here:

This block returns `ReasonPodsStarting` without checking `hasNewReplicaSetAvailable`. Because `ReasonPodsStarting` is included in the `ignoreReasons` list for the Progressing condition, won't this cause the Operator to incorrectly report `Progressing=False` during the final phase of a configuration-driven rollout (when the old pods have been deleted and the new ones are starting up but not yet available)?

I think the fix is to merge the checks into the `hasNewReplicaSetAvailable` block above.

bentito: And we should probably add a test case to prevent regressions for this specific scenario (a config rollout where old pods have terminated but new pods are not yet available.

lihongan:
/verified by @lihongan 

re-verify again with latest change and looks good

PR #1256, opened 2025-08-05
Azure: Add Ingress LB IPs to Infra CR when in-cluster DNS is enabled

sadasu:

PR #1302, opened 2025-11-05
CORS-4229: Revert "dns/gcp: Allow configuring custom endpoints"

barbacbd:
This reverts commit 95b36d0270996954500ee1840203c305b66f5431.

### DIFF
--- a/pkg/dns/gcp/provider.go
+++ b/pkg/dns/gcp/provider.go
@@ -31,30 +31,13 @@ type Provider struct {
 }
 
 type Config struct {
-	Project                   string
-	UserAgent                 string
-	CredentialsJSON           []byte
-	Endpoints                 []configv1.GCPServiceEndpoint
-	GCPCustomEndpointsEnabled bool
+	Project         string
+	UserAgent       string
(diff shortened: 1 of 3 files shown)


patrickdillon:
This 
/lgtm

As Brent mentioned, we changed the implementation [to use DNS records](https://docs.cloud.google.com/vpc/docs/configure-private-service-connect-apis) rather than [WithEndpoint](https://docs.cloud.google.com/vpc/docs/configure-private-service-connect-apis#configure-p-dns). This was necessary to support STS & oauth API calls, but also turned out to be a simpler implementation which also supports all day-2 operators. A better solution all around, but we realized it late.

alebedev87:
/hold

Holding until https://github.com/openshift/cluster-ingress-operator/pull/1256 and https://github.com/openshift/cluster-ingress-operator/pull/1297 are merged.

Miciah:
/hold cancel
since #1297 merged.

rikatz:
/test e2e-aws-ovn-serial-1of2

one last try, otherwise we will need to investigate this

PR #1316, opened 2025-12-02
OCPBUGS-65482: Add e2e test for secure redirect port stripping

bentito:
This adds a new e2e test 'TestSecureRedirectStripsPort' that verifies that the router strips the port from the Host header when performing a redirect due to insecureEdgeTerminationPolicy: Redirect.

~It temporarily updates the operator deployment to use a custom router image 'quay.io/btofel/router:test' which contains the fix.~

### DIFF
--- a/test/e2e/redirect_port_stripping_test.go
+++ b/test/e2e/redirect_port_stripping_test.go
@@ -1,0 +1,209 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+)
+
+func TestSecureRedirectCorrectness(t *testing.T) {
+	t.Parallel()
+
+	// Ensure we can get the kube config, although kclient is used for most operations.
+	_, err := config.GetConfig()
+	if err != nil {
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+
+	name := "redirect-correctness"
+	icName := types.NamespacedName{Namespace: operatorNamespace, Name: name}
+	domain := name + "." + dnsConfig.Spec.BaseDomain
+	ic := newPrivateController(icName, domain)
+	if err := kclient.Create(context.TODO(), ic); err != nil {
+		t.Fatalf("failed to create ingresscontroller %s: %v", icName, err)
+	}
+	defer assertIngressControllerDeleted(t, kclient, ic)
+
+	conditions := []operatorv1.OperatorCondition{
+		{Type: operatorv1.IngressControllerAvailableConditionType, Status: operatorv1.ConditionTrue},
+	}
+	if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, icName, conditions...); err != nil {
+		t.Fatalf("failed to observe expected conditions: %v", err)
+	}
+
+	deployment := &appsv1.Deployment{}
+	if err := kclient.Get(context.TODO(), controller.RouterDeploymentName(ic), deployment); err != nil {
+		t.Fatalf("failed to get ingresscontroller deployment: %v", err)
+	}
+	service := &corev1.Service{}
+	if err := kclient.Get(context.TODO(), controller.InternalIngressControllerServiceName(ic), service); err != nil {
+		t.Fatalf("failed to get ingresscontroller service: %v", err)
+	}
+
+	echoPod := buildEchoPod(name+"-echo", deployment.Namespace)
+	if err := kclient.Create(context.TODO(), echoPod); err != nil {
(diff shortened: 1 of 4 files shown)


test/e2e/redirect_port_stripping_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Fatalf("failed to get kube config: %v", err)
+	}
+	kclient, err := client.New(kubeConfig, client.Options{})
+	if err != nil {
+		t.Fatalf("failed to create runtime client: %v", err)
+	}
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		t.Fatalf("failed to create kube client: %v", err)
+	}
```
alebedev87: There is already a global `kclient` set up for the whole test suite: [link](https://github.com/openshift/cluster-ingress-operator/blob/b62dd27c901d495e6e78b11968872c002cc2416f/test/e2e/operator_test.go#L120). Is there a reason you need a dedicated one?

test/e2e/redirect_port_stripping_test.go
```diff
@@ -0,0 +1,139 @@
...
+		defer readCloser.Close()
+
+		scanner := bufio.NewScanner(readCloser)
+		for scanner.Scan() {
+			line := scanner.Text()
+			// Looking for "Location: https://<host><path>"
+			// We want to ensure NO ":80" is present.
+			if strings.HasPrefix(strings.ToLower(line), "location:") {
+				t.Logf("Found location header: %s", line)
+				if strings.Contains(line, "https://"+echoRoute.Spec.Host) && !strings.Contains(line, ":80") {
```
alebedev87: `redirect sheme` is appending `/` for requests without any path, we may want to make sure `http-request redirect` will be doing the same thing:
```suggestion
				if strings.Contains(line, "https://"+echoRoute.Spec.Host+"/") && !strings.Contains(line, ":80") {
```

test/e2e/redirect_port_stripping_test.go
bentito: Nope, we don't, ty. The test will now use the global kclient.

bentito:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/cluster-ingress-operator#1316 https://github.com/openshift/router/pull/696

manifests/02-deployment.yaml
```diff
@@ -80,7 +80,7 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: IMAGE
-              value: openshift/origin-haproxy-router:v4.0
+              value: quay.io/btofel/router:test
```
Thealisyed: This was probably not meant to be pushed to the PR?

test/e2e/redirect_port_stripping_test.go
```diff
@@ -0,0 +1,143 @@
...
+			// We want to ensure NO ":80" is present.
+			if strings.HasPrefix(strings.ToLower(line), "location:") {
+				t.Logf("Found location header: %s", line)
+				if strings.Contains(line, "https://"+echoRoute.Spec.Host+"/") && !strings.Contains(line, ":80") {
+					foundLocation = true
+				}
+			}
+		}
+		if found302 && foundLocation {
+			return true, nil
```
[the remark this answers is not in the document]
bentito: Normally passing would give us:
```
--- PASS: TestAll/parallel/TestSecureRedirectStripsPort (85.33s)
```
if we're debugging the test, that log line might help but otherwise would just be wordy, IMO.

manifests/02-deployment.yaml
bentito: it was meant, I'll push a reverting commit for this and the verify test will start passing. I'll push the revert commit once reviewers all get a chance to see the success output on the `/testwith` multi PR test run that proves this worked.

bentito:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/cluster-ingress-operator#1316 openshift/router#696

test/e2e/redirect_port_stripping_test.go
```diff
@@ -0,0 +1,194 @@
...
+					t.Logf("Did not find 302 Found in output:\n%s", output)
+					return false, nil
+				}
+
+				// Check Location header
+				lines := strings.Split(output, "\n")
+				foundLocation := false
+				for _, line := range lines {
+					// Use Contains to be case-insensitive for "Location:"
+					// We use Contains instead of HasPrefix because curl -v output might start with "< "
```
alebedev87: I think it always starts with `< ` for a response. You can check `< location` using HAsPrefix to be 100% sure that `location` header is from the response and not from the request.

test/e2e/redirect_port_stripping_test.go
```diff
@@ -0,0 +1,194 @@
...
+					if strings.Contains(strings.ToLower(line), "location:") {
+						t.Logf("Found location header: %s", line)
+						// Verify matches expected
+						if strings.Contains(line, tc.expectedLoc) {
```
alebedev87: `strings.Contains` is fine but it can give false negatives if the returned url has more at the end than the expected loc.

alebedev87 (commented):
LGTM, just 2 last nit picks and I think you can fixup commits.

test/e2e/redirect_port_stripping_test.go
bentito: done in https://github.com/openshift/cluster-ingress-operator/pull/1316/commits/85ff827ddad04597865a13d7c64c20b6ad0556e4

alebedev87:

The main change was in the router (https://github.com/openshift/router/pull/696) and it was merged some time ago. This PR only adds an new e2e presubmit for the change. We are letting it follow the change as it should not impact the OCP payload in any way.

PR #1326, opened 2025-12-16
NE-2396: gatewayclass: Enable Horizontal Pod Autoscaling

Miciah:
Enable Horizontal Pod Autoscaling (HPA) on Istio.  Hard-code the autoscaling parameters based on the cluster infrastructure config:

- If the infrastructure topology is "SingleReplica", set minimum to 1.
- Otherwise, set minimum to 2.
- In any case, set maximum to 10.

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -98,6 +98,7 @@ const (
 
 type extraIstioConfig struct {
 	proxyConfig *configv1.Proxy
+	infraConfig *configv1.Infrastructure
 }
 
 var log = logf.Logger.WithName(controllerName)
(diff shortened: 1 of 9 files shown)


pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -146,6 +147,12 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 		return nil, err
 	}
 
+	// Watch the cluster infrastructure config in case the infrastructure
+	// topology changes.
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &configv1.Infrastructure{}, reconciler.enqueueRequestForSomeGatewayClass())); err != nil {
```
rikatz: question here is, if you change Sail Operator parameter will it reflect on Pilot config?
Also, in case this parameter is also used for Gateway replica deployment, will Sail reconcile all the gateways and HPA? 

(Giving some thoughts for test)

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/d4b675d3ab9c5e4543adc3872d8b94986c8d1034..fc6e70e0073410a750b5782b0ef385c2ff0eee4a makes `assertHorizontalPodAutoscalerEnabled` more robust in case the gateway isn't found or the HPA is created without the expected parameters and is subsequently updated to have the expected parameters.

PR #1279, opened 2025-09-05
NE-2139: Kubernetes to 1.33.4 and controller-runtime to 0.21

rikatz:
This change bumps Kubernetes libraries to 1.33.4  and controller-runtime to 0.21.0

This change bumps Kubernetes libraries to 1.33.4 and controller-runtime to v0.21.0

Due to some breaking changes, additional modifications were required:
* Fixed calls to controller.NewUnmanaged
* Forcing the unregistration of Go metrics of controller-runtime due
to conflicts
* Fixing calls to EventRecorder
* Bumping Dockerfile images for Openshift 4.21

PR #1329, opened 2025-12-23
OCPBUGS-70211: Fix logging for unmanaged controllers

Miciah:
Set default values, such as the logger, in the controller options for unmanaged controllers, namely the gateway-labeler, gateway-service-dns, and gatewayclass controllers.

Before controller-runtime v0.21.0, controller-runtime implicitly set default values for controller options for managed and unmanaged controllers alike.  Controller-runtime internally used the `DefaultFromConfig` method to do so.  Since controller-runtime v0.21.0, these default values are not implicitly set for unmanaged controllers (see https://github.com/kubernetes-sigs/controller-runtime/commit/d9ff283bfe844e8e3806eb2d264b2a6fa7815f66).  In particular, this means that the logger is not initialized, and so the controller initialization and any reconciliation errors are not logged.  This PR explicitly sets default values by calling `DefaultFromConfig` for unmanaged controllers.

Follow-up to #1279.

### DIFF
--- a/pkg/operator/controller/gateway-labeler/controller.go
+++ b/pkg/operator/controller/gateway-labeler/controller.go
@@ -50,7 +50,9 @@ func NewUnmanaged(mgr manager.Manager) (controller.Controller, error) {
 		client:   mgr.GetClient(),
 		recorder: mgr.GetEventRecorderFor(controllerName),
 	}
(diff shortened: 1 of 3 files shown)


Thealisyed (commented):
/lgtm

Miciah:
I am fine with waiting for https://github.com/coreos/rhel-coreos-config/pull/134 and a clean CI run.

PR #1332, opened 2026-01-02
OCPBUGS-70211: Add missing logger to unmanaged controllers

rikatz:
After bumping to newer controller-runtime the logger is not set anymore on unmanaged controllers.

This change adds the current logger from each controller back to the controller.Options, making controller-runtime use it when constructing the manager instance


rikatz:
Miciah already fixed it on https://github.com/openshift/cluster-ingress-operator/pull/1329 :man_facepalming:

PR #1334, opened 2026-01-05
OCPBUGS-9037, OCPBUGS-64565: Ensure canary cert matches the default ingress controller's cert

rfredette:
Add the canary certificate controller, which makes sure the certificate used for the ingress canary is the same as the default ingress controller's certificate.

This is part of the fix for OCPBUGS-9037 and OCPBUGS-64565

### DIFF
--- a/pkg/operator/controller/canary-certificate/controller.go
+++ b/pkg/operator/controller/canary-certificate/controller.go
@@ -1,0 +1,255 @@
+package canarycertificate
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	logf "github.com/openshift/cluster-ingress-operator/pkg/log"
+	"github.com/openshift/cluster-ingress-operator/pkg/manifests"
+
+	operatorcontroller "github.com/openshift/cluster-ingress-operator/pkg/operator/controller"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	canaryCertControllerName = "canary_certificate_controller"
+)
+
+var (
+	log = logf.Logger.WithName(canaryCertControllerName)
+)
+
+// Config holds all the things necessary for the controller to run.
+type Config struct {
+	OperatorNamespace string
+	OperandNamespace  string
+}
+
+// reconciler handles the actual canary certificate reconciliation logic in
+// response to events.
+type reconciler struct {
+	config   Config
+	client   client.Client
+	recorder record.EventRecorder
+}
+
+// New creates the canary certificate controller
+//
+// The canary certificate controller mirrors the default ingress controller's
+// certificate into the canary's namespace. It watches the canary's certificate
+// and the default ingress controller's certificate.
+func New(mgr manager.Manager, config Config) (controller.Controller, error) {
+	operatorCache := mgr.GetCache()
+	reconciler := &reconciler{
+		config:   config,
+		client:   mgr.GetClient(),
+		recorder: mgr.GetEventRecorderFor(canaryCertControllerName),
+	}
+	c, err := controller.New(canaryCertControllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+	// Watch for updates on the canary's certificate.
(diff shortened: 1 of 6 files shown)


rfredette:
> Does this replace #1155?

Yes, it does. It was failing all e2e tests, so I wanted to wait until it's not totally unsound before obsoleting that PR.

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,228 @@
...
+	c, err := controller.New(canaryCertControllerName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return nil, err
+	}
+	// Watch for updates on the canary's certificate. Also watch for updates on
+	// the default ingress controller's certificate. Because the default ingress
+	// controller's certificate name can be set at runtime, a Get() would need
+	// to be done to determine the exact certificate, so reconcile on any
+	// certificate changes in the operator's namespace.
+	canaryCertPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
```
candita: I feel like there needs to be a deeper inspection of the certificate contents, because the name and namespace  could stay the same, but the certificate contents themselves could change.  Is that a concern?

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		return false, nil, err
+	}
+	return true, currentCanaryCert, nil
+
+}
+
+// desiredCanaryCertificate returns the desired canary certificate secret, based
+// on the default ingress controller's certificate.
+func desiredCanaryCertificate(canaryDaemonSet appsv1.DaemonSet, defaultIngressCertificate corev1.Secret) *corev1.Secret {
```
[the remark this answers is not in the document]
Miciah: Or just pass in an OwnerReference value rather than the DaemonSet itself.

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,228 @@
...
+		changed = true
+	}
+	if current.Type != desired.Type {
+		updated.Type = desired.Type
+		changed = true
+	}
+	if !cmp.Equal(current.Annotations, desired.Annotations) {
+		updated.Annotations = desired.Annotations
+		changed = true
+	}
```
Miciah: Is this necessary?  Should we have equivalent logic for labels?

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,228 @@
...
+		case o.GetName() == operatorcontroller.CanaryCertificateName().Name && o.GetNamespace() == operatorcontroller.CanaryCertificateName().Namespace:
+			return true
+		case o.GetNamespace() == config.OperandNamespace:
+			return true
+		}
+		return false
+	})
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &corev1.Secret{}, &handler.EnqueueRequestForObject{}, canaryCertPredicate)); err != nil {
+		return nil, err
+	}
```
Miciah: We discussed this on a call.  We'll change this from 1 predicate and a watch into 2 predicates in a disjunction, 1 mapfunc, and a watch:

- One predicate that matches the canary secret in the "openshift-ingress-canary" namespace.  
- One predicate that matches the default IngressController's default certificate in the "openshift-ingress" namespace.  
- A mapfunc that always maps to the canary secret.

This has the following benefits:

- We still have a single watch.
- The watch has a predicate that matches exactly the events we care about (namely updates to either the original secret _or_ the canary's copy).
- The reconciliation request is always for the canary's secret, which is logically the thing this controller reconciles.

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,228 @@
...
+func (r *reconciler) updateCanaryCertificate(current, desired *corev1.Secret) (bool, error) {
+	changed, updated := canaryCertificateChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+	if err := r.client.Update(context.TODO(), updated); err != nil {
+		return false, err
+	}
+	log.Info("updated canary certificate", "namespace", updated.Namespace, "name", updated.Name)
+	return false, nil
```
candita:
```suggestion
	return true, nil
```

candita:
/approve

Hold for @Miciah to tag with lgtm.
/hold

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	if current.Type != desired.Type {
+		updated.Type = desired.Type
+		changed = true
+	}
+	if !cmp.Equal(current.Annotations, desired.Annotations) {
+		updated.Annotations = desired.Annotations
+		changed = true
+	}
+	if !cmp.Equal(current.Labels, desired.Labels) {
```
Miciah:
```suggestion
	if !cmp.Equal(current.Labels, desired.Labels, cmpopts.EquateEmpty()) {
```

pkg/operator/operator.go
```diff
@@ -232,6 +233,14 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
...
 	}
 
+	// Set up the canary certificate controller
+	if _, err := canarycertcontroller.New(mgr, canarycertcontroller.Config{
+		OperatorNamespace: config.Namespace,
+		OperandNamespace:  operatorcontroller.DefaultOperandNamespace,
+	}); err != nil {
+		return nil, fmt.Errorf("failed to create canary cert controller: %w", err)
+	}
+
```
Miciah: As you noted on our call, it would make sense to move the canarycertcontroller initialization after the canarycontroller initialization.  It doesn't _really_ matter though—canarycertcontroller cannot create the secret until the daemonset exists because of the owner reference, but the daemonset pods cannot start until the secret exists.  Everything eventually works out (_Eventual Consistency_!) irrespective of which controller starts first.

Miciah:
The lack of `EquateEmpty` is the only thing I noticed that might have some functional impact.  Other comments and suggestions concern trivial stylistic issues.  Feel free to accept or drop those suggestions.  

/lgtm

PR #1342, opened 2026-01-15
OCPBUGS-85148, NE-2395: Fix e2e tests to work on platforms with unmanaged DNS

rikatz:
On on-prem platforms we cannot assure that the DNS record will always work. On some platforms the DNS controller is created with a fake controller, and DNSRecords never gets ready. In these cases, what we need is that the tests rely on the Gateway status address as a name and connect to it directly instead of trying to resolve the DNS.

This change creates a new function to verify if the test is running on a platform with unmanaged DNS, and in this case it skips verifying DNSRecord and makes the resolution of DNS records be hardcoded to use the gateway address on the http client

### DIFF
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -693,8 +693,19 @@ func isDNSManagementSupported(t *testing.T) bool {
 			t.Logf("error getting dns config %v: %v, retrying...", name, err)
 			return false, nil
 		}
-		dnsManaged = (dnsConfig.Spec.PrivateZone != nil && dnsConfig.Spec.PrivateZone.ID != "") ||
-			(dnsConfig.Spec.PublicZone != nil && dnsConfig.Spec.PublicZone.ID != "")
+
+		// On the majority of platforms, the *Zone not being null is enough
+		// On Azure, a platform can be null and be empty so we need to validate (this means unmanaged as well)
+		// On AWS, spec.ID is optional as AWS may use tags
+
+		managedPrivateZone := dnsConfig.Spec.PrivateZone != nil &&
+			(dnsConfig.Spec.PrivateZone.ID != "" || len(dnsConfig.Spec.PrivateZone.Tags) > 0)
+
+		managedPublicZone := dnsConfig.Spec.PublicZone != nil &&
+			(dnsConfig.Spec.PublicZone.ID != "" || len(dnsConfig.Spec.PublicZone.Tags) > 0)
+
+		dnsManaged = managedPrivateZone || managedPublicZone
+
 		return true, nil
 	})
 	if err != nil {
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -681,6 +681,27 @@ func updateAndVerifyInfrastructureConfigWithRetry(t *testing.T, name types.Names
 	})
 }
 
+func isDNSManagementSupported(t *testing.T) bool {
+	t.Helper()
+	dnsConfig := &configv1.DNS{}
+	dnsManaged := false
+	name := types.NamespacedName{
+		Name: "cluster",
+	}
+	err := wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, timeout, false, func(ctx context.Context) (done bool, err error) {
+		if err := kclient.Get(ctx, types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
+			t.Logf("error getting infrastructure config %v: %v, retrying...", name, err)
+			return false, nil
+		}
+		dnsManaged = dnsConfig.Spec.PrivateZone != nil || dnsConfig.Spec.PublicZone != nil
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to define if Managed DNS is supported: %s", err)
+	}
+	return dnsManaged
+}
+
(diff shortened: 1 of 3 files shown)


rikatz:
Executed test on a vsphere cluster:

```
kubectl exec -it launch-clusterbot-wait -c test -- /tmp/e2e -test.run TestGatewayAPI -test.v
--- PASS: TestGatewayAPI (381.23s)
    --- PASS: TestGatewayAPI/testGatewayAPIResources (46.41s)
    --- PASS: TestGatewayAPI/testGatewayAPIObjects (53.42s)
    --- PASS: TestGatewayAPI/testGatewayAPIManualDeployment (10.07s)
    --- PASS: TestGatewayAPI/testGatewayAPIIstioInstallation (14.15s)
    --- SKIP: TestGatewayAPI/testGatewayAPIDNS (1.01s)
    --- SKIP: TestGatewayAPI/testGatewayAPIDNSListenerUpdate (1.01s)
    --- PASS: TestGatewayAPI/testGatewayAPIDNSListenerWithNoHostname (96.05s)
    --- PASS: TestGatewayAPI/testGatewayAPIResourcesProtection (87.86s)
        --- PASS: TestGatewayAPI/testGatewayAPIResourcesProtection/Ingress_operator_service_account_required (10.33s)
        --- PASS: TestGatewayAPI/testGatewayAPIResourcesProtection/Pod_binding_required (10.37s)
    --- PASS: TestGatewayAPI/testGatewayAPIRBAC (0.05s)
    --- PASS: TestGatewayAPI/testOperatorDegradedCondition (67.17s)
```

How to execute the test:
* Compile the e2e test with `go test -c -o e2e -tags e2e ./test/e2e/`
* Get a cluster of type vsphere, static and get the IPs  (as per https://github.com/openshift/release/pull/73349)
* Install MetalLB with the static IPs on this cluster (still check the pipeline above)
* Get the kubeconfig of the build cluster
* With that kubeconfig, copy the e2e and run:
```
kubectl cp e2e launch-clusterbot-wait:/tmp/ -c test
kubectl exec -it launch-clusterbot-wait -c test -- /tmp/e2e -test.run TestGatewayAPI -test.v
```

gcs278:
I'd like to take a look
/assign

test/e2e/util_test.go
```diff
@@ -681,6 +681,27 @@ func updateAndVerifyInfrastructureConfigWithRetry(t *testing.T, name types.Names
 	})
 }
 
+func isDNSManagementSupported(t *testing.T) bool {
```
gcs278: *nit* We usually write a short header comment of the function description

test/e2e/util_test.go
```diff
@@ -681,6 +681,27 @@ func updateAndVerifyInfrastructureConfigWithRetry(t *testing.T, name types.Names
...
+	dnsManaged := false
+	name := types.NamespacedName{
+		Name: "cluster",
+	}
+	err := wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, timeout, false, func(ctx context.Context) (done bool, err error) {
+		if err := kclient.Get(ctx, types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
+			t.Logf("error getting infrastructure config %v: %v, retrying...", name, err)
+			return false, nil
+		}
+		dnsManaged = dnsConfig.Spec.PrivateZone != nil || dnsConfig.Spec.PublicZone != nil
```
gcs278: This is more of a correctness comment because it's a generic helper function, but consider not only checking for `nil`, but consider checking for nil and for empty IDs, as for the Azure platform installation with `UserProvisionedDns=True` would have non-nil `PrivateZone` and `PublicZones`, but empty ID strings:
https://github.com/openshift/installer/blob/0324f33df063228208250c67d26a315877f335f4/pkg/asset/manifests/dns.go#L147-L150

test/e2e/util_test.go
```diff
@@ -681,6 +681,27 @@ func updateAndVerifyInfrastructureConfigWithRetry(t *testing.T, name types.Names
...
 
+func isDNSManagementSupported(t *testing.T) bool {
+	t.Helper()
+	dnsConfig := &configv1.DNS{}
+	dnsManaged := false
+	name := types.NamespacedName{
+		Name: "cluster",
+	}
+	err := wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, timeout, false, func(ctx context.Context) (done bool, err error) {
+		if err := kclient.Get(ctx, types.NamespacedName{Name: "cluster"}, dnsConfig); err != nil {
```
gcs278:
```suggestion
		if err := kclient.Get(ctx, name, dnsConfig); err != nil {
```

gcs278 (commented):
Looks good! All of my comments are mostly nits about style, feel free to take or leave them. I agree with the approach overall.

gcs278:
Thanks for the updates! Yea @lihongan is right, the `TestAWSLBTypeChange` failure is due to https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/362 merging, and our fix needs to merge: https://github.com/openshift/cluster-ingress-operator/pull/1142. Though, I'm not sure about the `TestCanaryRoute`, that seems unrelated.

/lgtm

grzpiotrowski (approved):
/approve

PR #1349, opened 2026-01-22
OCPBUGS-54702: change default timeout tunnel if using AWS NLB

jcmoraisjr:
AWS NLB has its timeout tunnel as 350s by default. Currently we don't
have a way to change this configuration, so changing instead default
timeout tunnel on the HAProxy side to 349s. This configuration avoids
TCP resets on the client side.

The LB type is being read from a Service annotation if provided,
otherwise it comes from the Status. This avoids to use a wrong type, not
applied yet, still on a pending state.

https://issues.redhat.com/browse/OCPBUGS-54702

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -131,12 +131,12 @@ const (
 
 // ensureRouterDeployment ensures the router deployment exists for a given
 // ingresscontroller.
-func (r *reconciler) ensureRouterDeployment(ci *operatorv1.IngressController, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer, networkConfig *configv1.Network, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy, proxyNeeded bool) (bool, *appsv1.Deployment, error) {
+func (r *reconciler) ensureRouterDeployment(ci *operatorv1.IngressController, infraConfig *configv1.Infrastructure, ingressConfig *configv1.Ingress, apiConfig *configv1.APIServer, networkConfig *configv1.Network, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy, currentLBService *corev1.Service, proxyNeeded bool) (bool, *appsv1.Deployment, error) {
 	haveDepl, current, err := r.currentRouterDeployment(ci)
 	if err != nil {
 		return false, nil, err
 	}
-	desired, err := desiredRouterDeployment(ci, &r.config, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig)
+	desired, err := desiredRouterDeployment(ci, &r.config, ingressConfig, infraConfig, apiConfig, networkConfig, currentLBService, proxyNeeded, haveClientCAConfigmap, clientCAConfigmap, clusterProxyConfig)
 	if err != nil {
 		return haveDepl, current, fmt.Errorf("failed to build router deployment: %v", err)
 	}
@@ -258,7 +258,7 @@ func headerValues(values []operatorv1.IngressControllerHTTPHeader) string {
 }
 
 // desiredRouterDeployment returns the desired router deployment.
-func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, apiConfig *configv1.APIServer, networkConfig *configv1.Network, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy) (*appsv1.Deployment, error) {
+func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, apiConfig *configv1.APIServer, networkConfig *configv1.Network, currentLBService *corev1.Service, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy) (*appsv1.Deployment, error) {
 	deployment := manifests.RouterDeployment()
 	name := controller.RouterDeploymentName(ci)
 	deployment.Name = name.Name
@@ -650,6 +650,33 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 	}
 	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
+	// Check for AWS deployment, and if so and exposed via NLB, need to change default timeout tunnel to less than 350s
+	// https://issues.redhat.com/browse/OCPBUGS-54702
+	var tunnelTimeout *time.Duration
+	if ci.Spec.TuningOptions.TunnelTimeout != nil && ci.Spec.TuningOptions.TunnelTimeout.Duration > 0*time.Second {
+		// honor any configuration provided by the user
+		tunnelTimeout = &ci.Spec.TuningOptions.TunnelTimeout.Duration
+	} else {
+		// no config from the user, checking for NLB
+		isAWS := infraConfig.Status.PlatformStatus != nil &&
+			infraConfig.Status.PlatformStatus.Type == configv1.AWSPlatformType
+		if isAWS {
+			var lbType operatorv1.AWSLoadBalancerType
+			if currentLBService != nil {
+				lbType = getAWSLoadBalancerTypeFromServiceAnnotation(currentLBService)
+			} else {
+				lbType = getAWSLoadBalancerTypeInStatus(ci)
+			}
+			if lbType == operatorv1.AWSNetworkLoadBalancer {
+				// NLB at AWS, need to use less than 350s as the timeout
+				tunnelTimeout = ptr.To((awsNLBDefaultTunnelTimeoutSeconds - 1) * time.Second)
+			}
+		}
+	}
+	if tunnelTimeout != nil {
+		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_TUNNEL_TIMEOUT", Value: durationToHAProxyTimespec(*tunnelTimeout)})
+	}
+
 	if ci.Spec.HTTPHeaders != nil && len(ci.Spec.HTTPHeaders.Actions.Response) != 0 {
 		env = append(env, corev1.EnvVar{Name: RouterHTTPResponseHeaders, Value: headerValues(ci.Spec.HTTPHeaders.Actions.Response)})
 	}
@@ -669,9 +696,6 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 	}
 	if ci.Spec.TuningOptions.ServerFinTimeout != nil && ci.Spec.TuningOptions.ServerFinTimeout.Duration > 0*time.Second {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ServerFinTimeout.Duration)})
-	}
-	if ci.Spec.TuningOptions.TunnelTimeout != nil && ci.Spec.TuningOptions.TunnelTimeout.Duration > 0*time.Second {
-		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_TUNNEL_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.TunnelTimeout.Duration)})
 	}
 	if ci.Spec.TuningOptions.ConnectTimeout != nil && ci.Spec.TuningOptions.ConnectTimeout.Duration > 0*time.Second {
 		env = append(env, corev1.EnvVar{Name: "ROUTER_DEFAULT_CONNECT_TIMEOUT", Value: durationToHAProxyTimespec(ci.Spec.TuningOptions.ConnectTimeout.Duration)})
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -175,7 +175,7 @@ func TestTuningOptions(t *testing.T) {
 	ic.Spec.TuningOptions.ReloadInterval = metav1.Duration{Duration: 30 * time.Second}
 	ic.Spec.TuningOptions.HTTPKeepAliveTimeout = &metav1.Duration{Duration: 30 * time.Second}
 
-	deployment, err := desiredRouterDeployment(ic, &Config{IngressControllerImage: ingressControllerImage}, ingressConfig, infraConfig, apiConfig, networkConfig, false, false, nil, clusterProxyConfig)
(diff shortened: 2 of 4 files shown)


pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -201,6 +201,49 @@ func TestTuningOptions(t *testing.T) {
 	checkDeploymentHasEnvSorted(t, deployment)
 }
 
+func TestTunningOptionsForAWSNLB(t *testing.T) {
```
gcs278:
```suggestion
func TestTuningOptionsForAWSNLB(t *testing.T) {
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -637,6 +637,32 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	if ci.Spec.TuningOptions.TunnelTimeout != nil && ci.Spec.TuningOptions.TunnelTimeout.Duration > 0*time.Second {
+		// honor any configuration provided by the user
+		tunnelTimeout = &ci.Spec.TuningOptions.TunnelTimeout.Duration
+	} else {
+		// no config from the user, checking for NLB
+		eps := ci.Spec.EndpointPublishingStrategy
+		isAWS := eps != nil &&
+			eps.LoadBalancer != nil &&
+			eps.LoadBalancer.ProviderParameters != nil &&
+			eps.LoadBalancer.ProviderParameters.Type == operatorv1.AWSLoadBalancerProvider
```
gcs278: Hmm you should take a look at the impact of https://github.com/openshift/cluster-ingress-operator/pull/1142 on reading `Type` in the spec. You should also rebase to get these updates (at the time of writing this).

#1142 was a response to the cloud team making the LB Type annotation immutable. We have an existing pattern where, if you change a spec value that is immutable on the LB Service (or requires service deletion to effectuate), we put the IC into a `Progressing` state with a message telling the use to manually delete the service. So `Type` here has been added as one of these effectuation-required API fields (I need to think of a better name for this 😄).

The implication is that: during these pending states, spec will not have the correct/actual LB Type. Using IC status doesn't help here either because it represents the _desired state (aka spec) + defaulting_. 

So I think we have 2 options:
1. Keep your PR how it is - and allow a temporary gap for incorrect configuration
   - Tunnel configuration would be inaccurate during this NLB --> CLB or CLB --> NLB pending state 
2. Pass in the LB service, and use `getAWSLoadBalancerTypeFromServiceAnnotation` (introduced in #1142)
   - Another unfortunate dependency that makes our logic more intertwined

pkg/operator/controller/ingress/deployment.go
```diff
@@ -637,6 +637,32 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 	}
 	env = append(env, corev1.EnvVar{Name: RouterHAProxyThreadsEnvName, Value: strconv.Itoa(threads)})
 
+	// (jomorais) Check for AWS deployment, and if so and exposed via NLB, need to change default timeout tunnel to less than 350s
```
gcs278: Sorry this is a minor nit - but generally don't need author name here - we can use git history

```suggestion
	// Check for AWS deployment, and if so and exposed via NLB, need to change default timeout tunnel to less than 350s
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -637,6 +637,32 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		isAWS := eps != nil &&
+			eps.LoadBalancer != nil &&
+			eps.LoadBalancer.ProviderParameters != nil &&
+			eps.LoadBalancer.ProviderParameters.Type == operatorv1.AWSLoadBalancerProvider
+		if isAWS {
+			awsNLB := eps.LoadBalancer.ProviderParameters.AWS != nil &&
+				eps.LoadBalancer.ProviderParameters.AWS.Type == operatorv1.AWSNetworkLoadBalancer
+			if awsNLB {
+				// NLB at AWS, need to use less than 350s as the default value
+				tunnelTimeout = ptr.To(time.Duration(349 * time.Second))
```
gcs278: *nit* redundant type conversion, it's already time.Duration 

```suggestion
				tunnelTimeout = ptr.To(349 * time.Second)
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -201,6 +201,49 @@ func TestTuningOptions(t *testing.T) {
...
+	setupAndCheck := func(userTimeoutTunnel *time.Duration, providerParams *operatorv1.ProviderLoadBalancerParameters, expectPresent bool, expectedValue string) {
```
gcs278: I'm not sure if I can say this is a "team convention" yet, but most new unit tests we write, we use `t.Run` (at least where it makes sense). This test has 8 cases or so - consider using table-driven tests with `t.Run` subtests.

We get better failure reporting, the ability to run subtests individually, and it provides a bit of common structure for writing unit tests.

gcs278 (commented):
Nice PR @jcmoraisjr! No major comments, except for LB Type pending implications, let's talk about that more.

pkg/operator/controller/ingress/deployment_test.go
jcmoraisjr: Sure, this is a pattern I use to use as well. Just tried to follow the same approach of other tests, but in fact moving to a table-driven makes it much more clear and having better output in case of failures.

pkg/operator/controller/ingress/deployment.go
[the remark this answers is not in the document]
gcs278: Sorry for the major delay - I think the only reasonable solution is to pass service down via function parameters. It feels a bit messy though having so many function parameters, and intertwining logic.

I'm not sure what the right solution is. The source of truth is dual: service annotation when it exists (actual), IC spec when it doesn't (future). Any code needing the 'real' LB type must handle both cases.

I think having an IC status that represents the _real_ LB Type would _may be_ be helpful so we have 1 source of truth. Unfortunately, the API team has said in the past that we can't change the existing behavior of status as that's a breaking change.

We've explored adding a new field in status, `status.observedEndpointPublishingStrategy.LoadBalancer.ProviderParameters.Type` that would be a new field that ONLY represents the real/actual/observed value. That would at least allow us to decouple this function (and others like `IsProxyProtocolNeeded`) from service.

But I think this introduces another problem: if the observed status is updated AFTER the LB Type change (which we currently updates status after service), then the consumers of the observed status are delayed by 1 reconcile.  Maybes that's not a big deal since LB Type changes take time and are disruptive anyways. But also, I think the K8S community may frown upon using status as a source of truth for internal logic.

Another solution is the nuclear answer: rethink the "delay effectuation" pattern. Not having `spec` as the single source of truth of desired state makes our logic more intertwined. The fields being immutable has been explored, but deleting an entire IC to change one field is painful. Maybe we prevent changing LB Type in the spec with API rejection, but only allow when you have a field called `acknowledgeServiceRecreation: true` set. Like a "break glass" field. This would allow the spec, to be always true (or eventually true). But this is a non-trivial change, and we still have to support old behavior, so maybe for a redesign in the future 😸 

Anyways, for the sake of this bug, I'd recommend just trying to pass the service down through the parameters. But feel free to suggest anything else.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -633,6 +633,36 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		eps := ci.Status.EndpointPublishingStrategy
+		isAWS := eps != nil &&
+			eps.LoadBalancer != nil &&
+			eps.LoadBalancer.ProviderParameters != nil &&
+			eps.LoadBalancer.ProviderParameters.Type == operatorv1.AWSLoadBalancerProvider
+		if isAWS {
+			var lbType operatorv1.AWSLoadBalancerType
+			if currentLBService != nil {
+				lbType = getAWSLoadBalancerTypeFromServiceAnnotation(currentLBService)
+			} else {
```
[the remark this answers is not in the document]
gcs278: I see code rabbits point - `isAWS` is fragile, because `spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.Type` is optional. However, `status.eps.lb.pp.Type` **should** always be set by this [logic](https://github.com/openshift/cluster-ingress-operator/blob/70bab8473945ba7ceef86c8f5f4146211e868ba5/pkg/operator/controller/ingress/controller.go#L787-L790). However, I'm not 100% sure that an old IC introduced before that code change couldn't sneak through upgrades without `status.eps.lb.pp.AWS.Type` set (unless it gets [re-admitted](https://github.com/openshift/cluster-ingress-operator/blob/70bab8473945ba7ceef86c8f5f4146211e868ba5/pkg/operator/controller/ingress/controller.go#L427)), and that could be a problem for future code.

It doesn't matter for your NLB-defaulting fix here, as users must set  `spec.eps.lb.pp.Type` if they set `spec.eps.lb.pp.AWS.Type` to `NLB` and that propagates to `status`.

BUT, say a future bug says "set default timeout for CLB". Well, this is a very small chance that `status.eps.lb.pp.Type` is not set, and adjusting this logic for CLB default _might_ not work. The safest thing to do is pass in `platformStatus *configv1.PlatformStatus` as a parameter and check `platformStatus *configv1.PlatformStatus platformStatus.Type`. Yet another parameter, but it's probably best as defensive code (aka status is a minefield). We also generally use `platformStatus` as the source of truth for platform in our code.

Make sense?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -633,6 +633,36 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+				lbType = getAWSLoadBalancerTypeInStatus(ci)
+			}
+			if lbType == operatorv1.AWSNetworkLoadBalancer {
+				// NLB at AWS, need to use less than 350s as the default value
+				tunnelTimeout = ptr.To(349 * time.Second)
```
gcs278: *nit* it's a nit, but did you consider declaring this as a constant (e.g. `NLB_DEFAULT_TUNNEL_TIMEOUT=350` or something) and doing `NLB_DEFAULT_TUNNEL_TIMEOUT - 1`?

A const might make it more visible, and more clear how you arrived at 349, and in case someone refactors in the future and accidentally removes your code comments.

pkg/operator/controller/ingress/deployment.go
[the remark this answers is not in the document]
jcmoraisjr: Got it. I did a small refactor in the unit test to cover scenarios like NLB configuration and missing type, as well as configured to other platforms.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -201,11 +201,151 @@ func TestTuningOptions(t *testing.T) {
...
+			providerParamsAWS:    awsCLB, // LB service annotation should win and consider NLB, but skipping due to platform type
+			lbService:            &corev1.Service{ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{AWSLBTypeAnnotation: AWSNLBAnnotation}}},
+			userTimeoutTunnel:    nil,
+			expectedTimeoutValue: "",
+		},
+		{
+			name:                 "should skip timeout if using NLB from Status and CLB from service annotation",
+			providerParamsType:   awsPlatform,
+			providerParamsAWS:    awsNLB, // LB service annotation should win, consider CLB, and skip timeout configuration
+			lbService:            &corev1.Service{ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{AWSLBTypeAnnotation: AWSCLBAnnotation}}},
```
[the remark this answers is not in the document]
jcmoraisjr: Looks good, just updated both annotation removal and cleaning up service annotations.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -201,11 +201,151 @@ func TestTuningOptions(t *testing.T) {
 	checkDeploymentHasEnvSorted(t, deployment)
 }
 
+func TestTunningOptionForAWSNLB(t *testing.T) {
```
gcs278: sorry, *nit* spelling
```suggestion
func TestTuningOptionForAWSNLB(t *testing.T) {
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1029,7 +1169,9 @@ func TestDesiredRouterDeploymentFIPS(t *testing.T) {
 		},
 	}
 
-	deployment, err := desiredRouterDeployment(ic, &Config{IngressControllerImage: ingressControllerImage}, ingressConfig, infraConfig, apiConfig, networkConfig, proxyNeeded, false, nil, clusterProxyConfig)
+	// func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, ingressConfig *configv1.Ingress, infraConfig *configv1.Infrastructure, apiConfig *configv1.APIServer, networkConfig *configv1.Network, currentLBService *corev1.Service, proxyNeeded bool, haveClientCAConfigmap bool, clientCAConfigmap *corev1.ConfigMap, clusterProxyConfig *configv1.Proxy) (*appsv1.Deployment, error) {
```
gcs278: I think this was left over.
```suggestion
```

jcmoraisjr: Oops removed.

pkg/operator/controller/ingress/deployment_test.go
jcmoraisjr: No worries, updated.

gcs278:
Thanks for being responsive on this one. We've covered the gap in effectutation (where the user needs to delete the service in order to implement the LB Type change). Though it's messy, I think it's more correct to handle it. This tricky situation is a candidate to document in a `.coderabbit.yaml` or README in our repo so code rabbit can learn and apply to a review.

/approve

@grzpiotrowski did you want to take a look as a second reviewer?

grzpiotrowski (approved):
/lgtm

melvinjoseph86:
```
➜  oc get clusterversion
version   4.22.0-0-2026-04-29-050247-test-ci-ln-36f15lb-latest   True        False         9h      Cluster version is 4.22.0-0-2026-04-29-050247-test-ci-ln-36f15lb-latest

On default check
➜  oc -n openshift-ingress get deployment/router-default -oyaml | grep TIME -A1
    - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
     value: 349s

➜  oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"tuningOptions":{"tunnelTimeout":"200s"}}}'
ingresscontroller.operator.openshift.io/default patched
➜  oc -n openshift-ingress get deployment/router-default -oyaml | grep TIME -A1
    - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
     value: 200s

➜  oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"tuningOptions":{"tunnelTimeout":"350s"}}}'
ingresscontroller.operator.openshift.io/default patched
➜  oc -n openshift-ingress get deployment/router-default -oyaml | grep TIME -A1
    - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
     value: 350s

➜  oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"tuningOptions":{"tunnelTimeout":"351s"}}}'
ingresscontroller.operator.openshift.io/default patched
➜  oc -n openshift-ingress get deployment/router-default -oyaml | grep TIME -A1
    - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
     value: 351s

➜  oc -n openshift-ingress-operator patch ingresscontroller/default --type=merge -p '{"spec":{"tuningOptions":{"tunnelTimeout":null}}}'
ingresscontroller.operator.openshift.io/default patched
➜  oc -n openshift-ingress get deployment/router-default -oyaml | grep TIME -A1
    - name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
     value: 349s
```
     
Pre merge verified 
/verified by @mjoseph

PR #1362, opened 2026-02-13
NE-2664: parameterize haproxy version

jcmoraisjr:
add two new annotations used to parameterize the haproxy version to be used in the router.

* `haproxy-version`: defines the haproxy version to use. I must be one of the installed versions in the default router image. If missing, it defaults to the older installed version.

* `haproxy-image`: URL of the haproxy image to use. If configured, ingress operator deploys haproxy as a sidecar and configures it instead of the embedded haproxy from the router image.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/url"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -1198,6 +1201,94 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		haproxyImage = ci.Annotations["haproxy-image"]
+	}
+
+	if haproxyVersion != "" {
+		env = append(env, corev1.EnvVar{
+			Name:  "ROUTER_HAPROXY_VERSION",
+			Value: haproxyVersion,
+		})
+	}
+
```
Thealisyed: What happens if an invalid version string is given. Is there a fallback or some validation?

Thealisyed (commented):
It might also be worth noting that we should consider adding some test coverage for the annotations (version, image...) maybe?
Wdyt??

jcmoraisjr:
Closing this branch on behalf of https://github.com/openshift/cluster-ingress-operator/pull/1439

/close

PR #1374, opened 2026-03-03
WIP: OCPBUGS-77056: add external certs benchmark scripts

bentito:
This PR adds an orchestrator (`00-run-all.sh`) and supporting scripts to benchmark the OpenShift router's startup time when handling `spec.tls.externalCertificate` routes, aimed at validating fixes for OCPBUGS-77056.

The benchmark harness tests how quickly the router can load N external certificates, comparing the existing bug (where a global write lock causes serial registrations) against a patched image that allows concurrent processing.

### DIFF
--- a/hack/benchmark-external-certs/00-run-all.sh
+++ b/hack/benchmark-external-certs/00-run-all.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+# 00-run-all.sh
+# Full OCPBUGS-77056 benchmark orchestrator.
+#
+# Phases (each starts with a wipe for idempotency):
+#   0. quay login + preflight checks
+#   1. Wipe + create test routes (extcert-bench namespace)
+#   2. BASELINE: restart existing cluster router, measure startup
+#   3. PATCHED: deploy pre-built patched image, restart, measure startup
+#   4. Print comparison table
+#   5. Cleanup prompt
+#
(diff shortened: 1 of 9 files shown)


bentito:
The fix for OCPBUGS-77056 that enables faster startup behavior benchmarked in this PR is implemented and proposed in the following linked pulls:

* openshift/library-go#2132 (Async secret registration / non-blocking cache sync)
* openshift/router#745 (Asynchronous and cached SubjectAccessReview evaluation)

bentito:
This PR is primarily to tie together changes in router and library-go and to illustrate a benchmarking scheme for the Perf/Scale team. 

@SachinNinganure please let me know if the benchmarking scripts helps test changes to the router made in the linked PRs.

We are hoping you're able to work testing with these PRs in place into some of your work on https://issues.redhat.com/browse/CORENET-6842 

Primarily we're looking to make sure that parallelizing these calls to speed up the router's handling of adding routes with external certs, is not simply shifting the problem to the API server or elsewhere. So when we save time in the router pod are we causing CPU or memory problems elsewhere on the cluster. Thanks!

bentito:
Closing this PR. We have finalized a router-only fix (openshift/router#745) that optimizes the SAR checks and route status writing. The benchmark scripts here are no longer needed.

PR #1380, opened 2026-03-11
OCPBUGS-25739: Fix PDB MaxUnavailable for 2-replica ingress controllers

davidesalerno:
Use an integer MaxUnavailable of 1 instead of "50%" for 2-replica ingress controllers to avoid ambiguous percent-based rounding behavior in the disruption controller.

Also add isSecondReplicaUnschedulable helper to detect pods that cannot be scheduled, and update the test expectations to match the new logic.

### DIFF
--- a/pkg/operator/controller/ingress/poddisruptionbudget.go
+++ b/pkg/operator/controller/ingress/poddisruptionbudget.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 // ensureRouterPodDisruptionBudget ensures the pod disruption budget exists for
@@ -68,14 +69,23 @@ func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deployme
 		return false, nil, nil
 	}
 
-	maxUnavailable := "50%"
+	// OCPBUGS-25739 - For 2-replica ingress controllers, use an integer MaxUnavailable
+	// of 1 instead of "50%" to provide a deterministic value regardless of topology,
+	// avoiding any dependency on percent-based rounding in the disruption controller.
+	//
 	// OCPBUGS-7546 - make sure number of available pods is always 2 when there are only 3 replicas.
-	if ic.Spec.Replicas != nil && int(*ic.Spec.Replicas) >= 3 {
-		maxUnavailable = "25%"
+	var maxUnavailable intstr.IntOrString
+	replicas := ptr.Deref(ic.Spec.Replicas, 0)
+	switch {
+	case replicas == 2:
+		maxUnavailable = intstr.FromInt(1)
+	case replicas >= 3:
+		maxUnavailable = intstr.FromString("25%")
+	default:
+		maxUnavailable = intstr.FromString("50%")
 	}
 
 	name := controller.RouterPodDisruptionBudgetName(ic)
-	pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
 	pdb := policyv1.PodDisruptionBudget{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name.Name,
@@ -84,7 +94,7 @@ func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deployme
 		Spec: policyv1.PodDisruptionBudgetSpec{
 			// The disruption controller rounds MaxUnavailable up.
 			// https://github.com/kubernetes/kubernetes/blob/65dc445aa2d581b4fa829258e46e4faf44e999b6/pkg/controller/disruption/disruption.go#L539
-			MaxUnavailable: pointerTo(intstr.FromString(maxUnavailable)),
+			MaxUnavailable: &maxUnavailable,
 			Selector:       controller.IngressControllerDeploymentPodSelector(ic),
 		},
 	}
--- a/pkg/operator/controller/ingress/poddisruptionbudget_test.go
+++ b/pkg/operator/controller/ingress/poddisruptionbudget_test.go
@@ -30,10 +30,10 @@ func Test_desiredRouterPodDisruptionBudget(t *testing.T) {
 			expectMaxUnavailable: intstr.FromString("50%"),
 		},
 		{
-			description:          "if replicas is 2, PDB should be 50%",
+			description:          "if replicas is 2, PDB should be 1",
 			replicas:             pointerTo(2),
 			expectPDB:            true,
-			expectMaxUnavailable: intstr.FromString("50%"),
+			expectMaxUnavailable: intstr.FromInt(1),
 		},
 		{
 			description:          "if replicas is 3, PDB should be 25%",
@@ -44,12 +44,6 @@ func Test_desiredRouterPodDisruptionBudget(t *testing.T) {
 		{
 			description:          "if replicas is 4, PDB should be 25%",
 			replicas:             pointerTo(4),
-			expectPDB:            true,
-			expectMaxUnavailable: intstr.FromString("25%"),
-		},
-		{
-			description:          "if replicas is 5, PDB should be 25%",
-			replicas:             pointerTo(5),
 			expectPDB:            true,
 			expectMaxUnavailable: intstr.FromString("25%"),
 		},
--- a/pkg/operator/controller/gatewayapi/controller_test.go
+++ b/pkg/operator/controller/gatewayapi/controller_test.go
@@ -228,7 +228,7 @@ func Test_Reconcile(t *testing.T) {
 					StatusWriter: fakeClient.Status(),
 				},
 			}
-			ctrl := &testutil.FakeController{t, false, nil}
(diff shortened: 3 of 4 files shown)


pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -68,10 +68,21 @@ func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deployme
...
+	// OCPBUGS-25739 - For HostNetwork ingress controllers with 2 replicas, use an integer
+	// MaxUnavailable of 1 instead of "50%" to avoid ambiguous percent-based rounding behavior
+	// in the disruption controller.
+	//
 	// OCPBUGS-7546 - make sure number of available pods is always 2 when there are only 3 replicas.
-	if ic.Spec.Replicas != nil && int(*ic.Spec.Replicas) >= 3 {
-		maxUnavailable = "25%"
+	var maxUnavailable intstr.IntOrString
+	isHostNetwork := ic.Status.EndpointPublishingStrategy != nil &&
+		ic.Status.EndpointPublishingStrategy.Type == operatorv1.HostNetworkStrategyType
```
jcmoraisjr: This is not a topology issue I think, but instead a floating point issue that should happen on some architectures, which means that this can happen on other topologies as well. So my suggestion instead is to hardcode "`2` replicas" => "`1` unavailable" exactly like you did, but without restricting to baremetal / hostNetwork.

In case this makes sense, some changes can be skipped in the pdb_test.go file.

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -68,10 +68,21 @@ func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deployme
...
+	switch {
+	case isHostNetwork && ic.Spec.Replicas != nil && int(*ic.Spec.Replicas) == 2:
+		maxUnavailable = intstr.FromInt(1)
+	case ic.Spec.Replicas != nil && int(*ic.Spec.Replicas) >= 3:
+		maxUnavailable = intstr.FromString("25%")
+	default:
+		maxUnavailable = intstr.FromString("50%")
```
jcmoraisjr, line 85: nit: you can simplify the pointer check and walking through the spec struct by adding this before the switch declaration:

```go
replicas := ptr.Deref(ic.Spec.Replicas, 0)
```

`0` also means unassigned and both goes to the default case, leaving the logic the same.

pkg/operator/controller/ingress/poddisruptionbudget.go
```diff
@@ -84,7 +95,7 @@ func desiredRouterPodDisruptionBudget(ic *operatorv1.IngressController, deployme
 		Spec: policyv1.PodDisruptionBudgetSpec{
 			// The disruption controller rounds MaxUnavailable up.
 			// https://github.com/kubernetes/kubernetes/blob/65dc445aa2d581b4fa829258e46e4faf44e999b6/pkg/controller/disruption/disruption.go#L539
-			MaxUnavailable: pointerTo(intstr.FromString(maxUnavailable)),
+			MaxUnavailable: pointerTo(maxUnavailable),
```
jcmoraisjr: You don't need the `pointerTo()` anymore, since you can now use a pointer to `maxUnavailable`


```suggestion
			MaxUnavailable: &maxUnavailable,
```

jcmoraisjr (commented):
Looks good, just a doubt regarding the need of checking for baremetal/hostNetwork.

pkg/operator/controller/ingress/poddisruptionbudget.go
davidesalerno: Updated

pkg/operator/controller/ingress/poddisruptionbudget.go
davidesalerno: I introduced the `replicas` variable as suggested and used it to simplify the code

pkg/operator/controller/ingress/poddisruptionbudget.go
davidesalerno: Regarding this floating-point concern, the computation `ceil(50 * 2 / 100)` is actually exact in IEEE 754 on all architectures. `50.0 * 2.0 = 100.0 `(exact), `100.0 / 100.0 = 1.0` (exact), `ceil(1.0) = 1` (exact). All these values are perfectly representable in binary floating point — no repeating fractions or precision loss. This holds on x86_64, ARM64, ppc64le, and s390x (all architectures OpenShift supports).  

So I think that it is better to limit the fix to HostNetwork ingress controllers on clusters with only 2 worker nodes because looking at the Jira it seems  that it is happening on this specific setup.

jcmoraisjr: My rationale on this is that:

* the e2e failure should not be happening if the calculation is precise on all the supported platforms
* I cannot see a relationship between the network topology and floating point calculation

So I was wondering if we should simply skip the calculation when we can provide an accurate result - and this is exactly what you did. However limiting this update to a specific topology means to me like cover the only known e2e test exposing the problem. Otoh, removing the restriction means a common, predictable and correct behavior on all architectures and topologies, which also simplifies the understanding of the code.

davidesalerno: Sorry @jcmoraisjr  I probably misinterpreted your observation.

I limited the fix to cluster using `HostNetwork` and having 2 nodes for this reason because with `HostNetwork`, each router pod binds to host ports (80/443), so only one router pod can run per node. 

On a 2-node HA cluster:                                                                                                  
  - When a node is drained/rebooted during an upgrade, its router pod goes down                                                                                                                                                     
  - The replacement pod cannot be scheduled on the remaining node (port conflict)                                                                                                                                                   
  - You're stuck with 1 available pod until the original node comes back

This scheduling constraint doesn't exist for other endpoint publishing strategies. With those, a replacement pod can immediately schedule on the remaining node, so the unavailability window is very  short and the Available=False condition resolves quickly. 

Could it be useful to add a comment to clarify this decision into the code?

pkg/operator/controller/ingress/poddisruptionbudget_test.go
```diff
@@ -14,42 +14,49 @@ func Test_desiredRouterPodDisruptionBudget(t *testing.T) {
 	testCases := []struct {
 		description          string
 		replicas             *int32
+		strategyType         operatorv1.EndpointPublishingStrategyType
```
jcmoraisjr: This looks to be not used anymore.

jcmoraisjr:
/approve
/lgtm

Looks great! Simple and straightforward.

pkg/operator/controller/ingress/poddisruptionbudget_test.go
```diff
@@ -47,12 +47,6 @@ func Test_desiredRouterPodDisruptionBudget(t *testing.T) {
 			expectPDB:            true,
 			expectMaxUnavailable: intstr.FromString("25%"),
 		},
-		{
-			description:          "if replicas is 5, PDB should be 25%",
-			replicas:             pointerTo(5),
-			expectPDB:            true,
-			expectMaxUnavailable: intstr.FromString("25%"),
-		},
```
jcmoraisjr, line 55: Hmm wondering why it was removed, it should be passing like the 3 and 4 replicas counterpart.

davidesalerno: I removed it because it is the same test case of 3 and 4 and so it was a duplicate.
Do you think it could be useful?

jcmoraisjr: Looks good, just curious. Thanks!

melvinjoseph86:
/pj-rehearse periodic-ci-openshift-release-main-nightly-4.22-upgrade-from-stable-4.21-e2e-metal-ipi-ovn-upgrade

melvinjoseph86:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-upgrade-from-stable-4.21-e2e-metal-ipi-ovn-upgrade

melvinjoseph86:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-upgrade-from-stable-4.21-e2e-metal-ipi-upgrade-ovn-ipv6  periodic-ci-openshift-release-main-nightly-4.23-upgrade-from-stable-4.22-e2e-metal-ipi-upgrade-ovn-ipv6   periodic-ci-openshift-release-main-nightly-4.23-upgrade-from-stable-4.22-e2e-metal-ipi-ovn-upgrade

melvinjoseph86:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-upgrade-from-stable-4.21-e2e-metal-ipi-upgrade-ovn-ipv6 periodic-ci-openshift-release-main-nightly-4.23-upgrade-from-stable-4.22-e2e-metal-ipi-upgrade-ovn-ipv6 periodic-ci-openshift-release-main-nightly-4.23-upgrade-from-stable-4.22-e2e-metal-ipi-ovn-upgrade periodic-ci-openshift-release-main-nightly-4.22-upgrade-from-stable-4.21-e2e-metal-ipi-ovn-upgrade

melvinjoseph86:
i am not able to pre image verify this PR, as clusterbot cluster is not coming up and no baremetal support in flexy, hence planning to verify this bug post merge.
/verified later @mjoseph

PR #1398, opened 2026-03-20
OCPBUGS-76609: Do not take over OSSM subscription

rikatz:

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -463,7 +463,8 @@ func (r *reconciler) reconcileWithOLM(ctx context.Context, request reconcile.Req
 	if subscription == nil {
 		log.Info("No OSSM subscription available; skipping install plan enforcement")
 	} else if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
-		log.Info("Found an existing OSSM subscription with another owner; installation skipped")
+		log.Info("Found an existing OSSM subscription with another owner; installation skipped",
+			"namespace", subscription.Namespace, "name", subscription.Name)
 	} else {
 		if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
 			errs = append(errs, err)
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -455,12 +455,11 @@ func (r *reconciler) reconcileWithOLM(ctx context.Context, request reconcile.Req
 	if v, ok := gatewayclass.Annotations[subscriptionVersionOverrideAnnotationKey]; ok {
 		ossmVersion = v
 	}
+
 	_, subscription, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
-	}
-
-	if subscription == nil {
+	} else if subscription == nil {
 		log.Info("No OSSM subscription available; skipping install plan enforcement")
 	} else if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
 		log.Info("Found an existing OSSM subscription with another owner; installation skipped",
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -460,7 +460,9 @@ func (r *reconciler) reconcileWithOLM(ctx context.Context, request reconcile.Req
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
 	}
 
-	if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+	if subscription == nil {
+		log.Info("No OSSM subscription available; skipping install plan enforcement")
+	} else if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
 		log.Info("Found an existing OSSM subscription with another owner; installation skipped")
 	} else {
 		if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -455,14 +455,13 @@ func (r *reconciler) reconcileWithOLM(ctx context.Context, request reconcile.Req
 	if v, ok := gatewayclass.Annotations[subscriptionVersionOverrideAnnotationKey]; ok {
 		ossmVersion = v
 	}
-	//
-	exists, existingSubscription, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion)
+	_, subscription, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
 	}
 
-	if exists && existingSubscription.Namespace != operatorcontroller.OpenshiftOperatorNamespace {
-		log.Info("another OSSM subscription already exists, installation will be skipped")
+	if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+		log.Info("Found an existing OSSM subscription with another owner; installation skipped")
 	} else {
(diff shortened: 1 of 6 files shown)


pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -456,12 +456,19 @@ func (r *reconciler) reconcileWithOLM(ctx context.Context, request reconcile.Req
...
-	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
-		errs = append(errs, err)
+
+	if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+		log.Info("Found an existing OSSM subscription with another owner; installation skipped")
+	} else {
+		if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
+			errs = append(errs, err)
+		}
 	}
```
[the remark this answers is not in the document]
Miciah: No, I am saying that I already pointed out the issue before you did, @coderabbitai.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -36,6 +36,30 @@ const (
...
+		// There is a subscription that is not owned by us. In this case we early return
+		// because we cannot support multiple existing OSSM subscriptions, so instead of
+		// trying to continue the workflow of making CIO take over the subscription
+		// the code is early returned without further update, and CIO can be marked
+		// with a degradation warning that allows the cluster admin to identify the
+		// other existing subscriptions, and decide further action.
+		// This does not block the rest of GatewayClass reconciliation, it just avoids
+		// CIO taking over subscriptions (or adding new ones) while other subscription
+		// exists.
+		if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; subscription.Name == name.Name && !ok {
```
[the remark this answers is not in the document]
rikatz: @Miciah I've found out this can cause bugs. When installing the subscription using the CLI you are not enforced to use the same name of OSSM, but instead it relies on the spec.Package field. In fact the rest of our code relies on it. So I just pushed a change that actually lists all subscriptions, but get the ones that contains `Spec.Package == "servicemeshoperator3"` only

Also, given there is no guarantee that spec.Package will be indexed (we would need a new indexer) I've just went with the simpler approach

rikatz:
test failing here is due to github.com/openshift/cluster-ingress-operator/pull/1408

rikatz:
@coderabbitai please do another review on this PR, and let me know any critical or major findings that are must fixes

rikatz:
/retest-required
weird flake on hypershift-e2e-aks infra

lihongan:
/retest

seeing `api error AuthFailure: AWS was not able to validate the provided access credentials]` in aws jobs

rikatz:
@coderabbitai please do another review on this PR, and let me know any critical or major findings that are must fixes

bentito:
PR looks in excellent shape. I had a minor concern about the all namespaces reconciliation in the status controller, specifically the direct `r.client.List` calls for `Subscriptions`. Since these hit the API server on every reconcile instead of using the dedicated `subscriptionCache`, it's technically a bit more load than necessary. However, given that the total number of `Subscriptions` is usually low and we're prioritizing absolute data freshness for the `ClusterOperator` status, it's likely a negligible performance trade-off in practice.

/lgtm

lihongan:
cc @rhamini3

pkg/operator/controller/status/controller_test.go
```diff
@@ -896,7 +896,27 @@ func Test_computeOperatorDegradedCondition(t *testing.T) {
...
+				},
+				ossmSubscriptions: []operatorsv1alpha1.Subscription{
+					sub("servicemeshoperator3", "servicemeshoperator3", "servicemeshoperator3.v3.1.0", false),
+				},
+				shouldInstallOSSM:                 true,
+				expectedGatewayAPIOperatorVersion: "servicemeshoperator3.v3.1.0",
+			},
+			expectCondition: configv1.ClusterOperatorStatusCondition{
+				Type:    configv1.OperatorDegraded,
+				Status:  configv1.ConditionFalse,
```
gcs278, line 917: TIL we add messages with `Degraded: False` situations (I know it was previously `Degraded: False` and you didn't change that). Does this mean, "it's not degraded yet, things should work, but we're not in control" type of situation?

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -251,6 +263,38 @@ func (r *reconciler) updateInstallPlan(ctx context.Context, current, desired *op
 	return true, nil
 }
 
+// findConflictingSubscription tries to find any OSSM/servicemesh subscription that may have conflicts with the one
+// that CIO will install
+func (r *reconciler) findConflictingSubscription(ctx context.Context) (*operatorsv1alpha1.Subscription, error) {
+	var subscriptionList operatorsv1alpha1.SubscriptionList
+	// r.client is being used here so we can scan all namespaces without relying/requiring them to be on the cache
+	if err := r.client.List(ctx, &subscriptionList, &client.ListOptions{
```
gcs278, line 271: Edge Case: Since we are not watching all subscriptions from every namespace, what happens if the user has a BYO-Subscription in a random namespace, but then the user changes their mind, and deletes the BYO-Subscription?

The user could possibly get stuck since the GatewayClass controller won't trigger any new reconcilation due to no watches firing a new reconcile to tell the CIO to take over.

I don't have a good solution to this, as we need to watch all subscriptions from all namespaces. I suppose you need to add a new separate cache/informer just for subscriptions in all namespaces (pretty chunky). Or, more primitively, you could add a RequeueAfter period of 5 minutes.

Given that this is a rare scenario, I'm okay to defer.

gcs278:
Nice update @rikatz well done. I agree with the changes here and have no blocker comments.

/approve

pkg/operator/controller/gatewayclass/subscription.go
rikatz: We actually are: 

https://github.com/openshift/cluster-ingress-operator/blob/8008be81974c8aa0add770087908e40a2138f344/pkg/operator/controller/gatewayclass/controller.go#L168-L176

The problem is not the namespace, but the name. As I was telling Miciah on a thread above, we care about the `sub.Spec.Package` while our previous assumption was about name, and we don't check it on predicates.

I just didn't wanted to make too many changes, but IIUC as soon as we have any subscription change with the name servicemeshoperator3 it will kick the Gateway Class reconciliation, and it will use r.client (which is not cached) to do a full list

gcs278: That watches subscriptions in the `operatorCache`, which is restricted to a subset of namespaces: https://github.com/openshift/cluster-ingress-operator/blob/eeb09ed6471af5e30666b46dc9414e7b6d01917f/pkg/operator/operator.go#L157-L164

But I also agree that you'd need to remove the predicate to not filter with the name `servicemeshoperator3` as well (or change to package).

My point being that the GatewayClass controller won't kick a reconcilation if a subscription is outside of the `operatorCache` subset of our watched namespaces (regardless of it's name), and if it's removed, it's possible the CIO will hesitate to take over. We should test this out in real life to verify if it's a problem, but anyways it's a minor edge case.

rhamini3:
Verified the behaviour is expected on a pre-merge cluster

Installed OSSM on the `default` namespace
```
 % oc -n default get sub,csv,pod
NAME                                                     PACKAGE                SOURCE             CHANNEL
subscription.operators.coreos.com/servicemeshoperator3   servicemeshoperator3   redhat-operators   stable

NAME                                                                     DISPLAY                            VERSION   RELEASE   REPLACES                      PHASE
clusterserviceversion.operators.coreos.com/servicemeshoperator3.v3.3.1   Red Hat OpenShift Service Mesh 3   3.3.1               servicemeshoperator3.v3.3.0   Succeeded

NAME                                         READY   STATUS    RESTARTS   AGE
pod/servicemesh-operator3-7847cb57b6-zpkkr   1/1     Running   0          16m
```

Install the gatewayclass
```
% oc get gc                        
openshift-default   openshift.io/gateway-controller/v1   True       9m32s
```

Check the `openshift-operators` namespace for the subscription
```
% oc -n openshift-operators get sub        
No resources found in openshift-operators namespace.
```

Check the ingress-operator logs to see the message preview
```
2026-04-07T18:59:41.946Z	INFO	operator.gatewayclass_controller	gatewayclass/controller.go:406	reconciling with OLM	{"request": {"name":"openshift-default"}}
2026-04-07T18:59:41.949Z	INFO	operator.gatewayclass_controller	gatewayclass/controller.go:406	Found an existing OSSM subscription with another owner; installation skipped	{"namespace": "default", "name": "servicemeshoperator3"}
```

/verified by @rhamini3

rikatz:
@coderabbitai did the tests of this PR failed because of any change here or were they flakes?

rikatz:
@Coderabbitai but analyze the previous failures from @openshift-ci-robot comment

pkg/operator/controller/status/controller_test.go
rikatz: when we were discussing this, the point of making degraded false but adding a message was to not block upgrades.

So it is one of those situations where we say "well, things will work but not as expected"

PR #1408, opened 2026-03-31
OCPBUGS-81442: TestUnsupportedConfigOverride: Ignore featuregate and defaults

Miciah:
#### `waitForDeploymentEnvVar`: Fix godoc

Update the godoc for the `waitForDeploymentEnvVar` test helper func to start with the func name and to mention that the helper specifically checks the "router" container.

#### Add `waitForDeploymentFunc` test helper func

Factor the polling loop logic out of the `waitForDeploymentEnvVar` helper into the new `waitForDeploymentFunc` helper.

#### `TestUnsupportedConfigOverride`: Ignore featuregate

Ignore the featuregate for the Dynamic Configuration Manager when testing `unsupportedConfigOverrides` options related to DCM, and don't verify defaults.  Overrides take precedence over the featuregate, and the test's purpose is to verify that overrides work properly, not verify the default settings.

Follow-up to #1385.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -4242,26 +4242,45 @@ func waitForDeploymentComplete(t *testing.T, cl client.Client, deployment *appsv
 // value, or unset if the provided value is the empty string.
 func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.Deployment, timeout time.Duration, name, value string) error {
 	t.Helper()
-	deploymentName := types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name}
-	err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
-		deployment := &appsv1.Deployment{}
-		if err := kclient.Get(context.TODO(), deploymentName, deployment); err != nil {
-			t.Logf("error getting deployment %s/%s: %v", deploymentName.Namespace, deploymentName.Name, err)
-			return false, nil
-		}
+
+	return waitForDeploymentFunc(t, deployment, timeout, func(deployment *appsv1.Deployment) bool {
 		for _, container := range deployment.Spec.Template.Spec.Containers {
 			if container.Name == "router" {
 				for _, v := range container.Env {
 					if v.Name == name {
-						return v.Value == value, nil
+						return v.Value == value
 					}
 				}
-				return len(value) == 0, nil
+
+				return len(value) == 0
 			}
 		}
-		return false, nil
+
+		return false
 	})
-	return err
+}
+
+// waitForDeploymentFunc polls the given deployment until the given condition
+// func returns true or the timeout elapses.
+func waitForDeploymentFunc(t *testing.T, deployment *appsv1.Deployment, timeout time.Duration, fn func(deployment *appsv1.Deployment) bool) error {
+	t.Helper()
+
+	deploymentName := types.NamespacedName{
+		Namespace: deployment.Namespace,
+		Name:      deployment.Name,
+	}
+
+	return wait.PollUntilContextTimeout(context.Background(), 1*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		var deployment appsv1.Deployment
+
+		if err := kclient.Get(ctx, deploymentName, &deployment); err != nil {
+			t.Logf("Failed to get deployment %s: %v", deploymentName, err)
+
+			return false, nil
+		}
+
+		return fn(&deployment), nil
+	})
 }
 
 // waitForDeploymentCompleteWithOldPodTermination waits for a deployment to
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -4237,9 +4237,9 @@ func waitForDeploymentComplete(t *testing.T, cl client.Client, deployment *appsv
 	return nil
 }
 
-// Wait for the provided deployment to have the specified environment variable
-// set with the provided value, or unset if the provided value is the empty
-// string.
+// waitForDeploymentEnvVar waits for the provided router deployment to have the
+// specified environment variable set in the router container with the provided
+// value, or unset if the provided value is the empty string.
 func waitForDeploymentEnvVar(t *testing.T, cl client.Client, deployment *appsv1.Deployment, timeout time.Duration, name, value string) error {
 	t.Helper()
 	deploymentName := types.NamespacedName{Namespace: deployment.Namespace, Name: deployment.Name}
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -27,7 +27,6 @@ import (
 	"gopkg.in/yaml.v2"
 
 	configv1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/api/features"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	iov1 "github.com/openshift/api/operatoringress/v1"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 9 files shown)


ShudiLi:
Checked the log of [ci/prow/e2e-aws-operator-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1408/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview/2039349701983080448), could not find the "set ROUTER_HAPROXY_CONFIG_MANAGER=true: timed out"

```
% oc get clusterversion
version   4.22.0-0-2026-04-02-061236-test-ci-ln-h3pknmt-latest   True        False         115m    Cluster version is 4.22.0-0-2026-04-02-061236-test-ci-ln-h3pknmt-latest

% oc -n openshift-ingress get deployment router-default -oyaml | grep ROUTER_HAPROXY_CONFIG_MANAGER  -A1    
        - name: ROUTER_HAPROXY_CONFIG_MANAGER
          value: "true"

% oc -n openshift-ingress logs router-default-77bcc95d78-4gr76  | grep time
% 
```

test/e2e/operator_test.go
```diff
@@ -3620,66 +3619,81 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
+		}, {
+			name:                       "max-dynamic-servers",
+			unsupportedConfigOverrides: `{"dynamicConfigManager":"true","maxDynamicServers":"2"}`,
+			expectedEnv: map[string]string{
+				"ROUTER_HAPROXY_CONFIG_MANAGER": "true",
+				"ROUTER_HAPROXY_CONTSTATS":      "",
+				"ROUTER_MAX_DYNAMIC_SERVERS":    "2",
+			},
+		},
+	}
```
jcmoraisjr, line 3677: I missed a test for load balance, and also wondering if adding an attribute without enabling dynamic config is doable, e.g. `{"maxDynamicServers":"3"}`

test/e2e/operator_test.go
```diff
@@ -4237,31 +4251,94 @@ func waitForDeploymentComplete(t *testing.T, cl client.Client, deployment *appsv
...
+			return container
+		}
+	}
+
+	return corev1.Container{}
+}
+
+// waitForDeploymentFunc polls the given deployment until the given condition
+// func returns true or the timeout elapses.
+func waitForDeploymentFunc(t *testing.T, cl client.Client, deployment *appsv1.Deployment, timeout time.Duration, fn func(deployment *appsv1.Deployment) bool) error {
```
jcmoraisjr: `cl` is not being used.

test/e2e/operator_test.go
```diff
@@ -3620,66 +3619,81 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
-		dcmExpectedValue = ""
-		maxServersDefaultValue = "1"
-		maxServersUpdateValue = "2"
-		maxServersExpectedValue = "2"
+	icName := types.NamespacedName{
+		Namespace: operatorNamespace,
+		Name:      names.SimpleNameGenerator.GenerateName("unsupported-config-override-"),
+	}
+	domain := fmt.Sprintf("%s.%s", icName.Name, dnsConfig.Spec.BaseDomain)
+	ic := newPrivateController(icName, domain)
```
jcmoraisjr, line 3684: Any value on moving it outside of the inner test?

test/e2e/operator_test.go
```diff
@@ -3620,66 +3619,81 @@ func TestLoadBalancingAlgorithmUnsupportedConfigOverride(t *testing.T) {
...
 				t.Fatalf("failed to update ingresscontroller: %v", err)
 			}
-			if err := waitForDeploymentEnvVar(t, kclient, deployment, 1*time.Minute, tt.env, tt.expectedValue); err != nil {
-				t.Fatalf("expected updated deployment to set %s=%s: %v", tt.env, tt.expectedValue, err)
+
+			var deployment appsv1.Deployment
+
+			if err := kclient.Get(context.Background(), controller.RouterDeploymentName(ic), &deployment); err != nil {
+				t.Fatalf("failed to get ingresscontroller deployment: %v", err)
+			}
```
jcmoraisjr, line 3710: In case I'm not missing anything, this call is not needed, `waitForDeploymentEnv()` below just needs the deployment name and it will get the content by itself.

jcmoraisjr (commented):
Looks good, just nitpicking here and there.

test/e2e/operator_test.go
Miciah: Fair point.  The test was previously focused on DCM overrides, but it is in fact named just `TestUnsupportedConfigOverride`, so I'll add a test case for the `loadBalancingAlgorithm` override.  I'll also add a test case that specifies `maxDynamicServers` but also turns off DCM (in which case the `maxDynamicServers` override should be ignored).

test/e2e/operator_test.go
Miciah: Good point.  I'll remove the unused parameter from `waitForDeploymentFunc`, `waitForDeploymentEnv`, and `waitForDeploymentEnvVar` (this one in a separate commit since removing the parameter requires updating a lot of other tests).

test/e2e/operator_test.go
Miciah: Yes:

- It means the test now covers updates from one nonempty set of overrides to another set of overrides.  The old logic only tested updates from an empty set of overrides.  
- It means that the test should run faster and with lower resource usage as it re-uses the ingresscontroller resource and router deployment instead of creating a new one for each test case.

test/e2e/operator_test.go
Miciah: `deployment` is used in two ways:

- `waitForDeploymentEnv` reads the name from `deployment`.  I could change `waitForDeploymentEnv` to take a `types.NamespacedName` argument instead of an `appsv1.Deployment` argument, but that would make it inconsistent with `waitForDeploymentEnvVar`.  Alternatively, I could stub out an `appsv1.Deployment` value with just the `Name` and `Namespace` fields filled in, which is consistent with what we do in some other tests.  I'll take that second approach.  Is that fine?
- The error logging uses `deployment` to print the actual environment variables from the deployment... though now I realize that we really need to do a `Get` _after_ `waitForDeploymentEnv` returns an error so that we have the current environment—`waitForDeploymentEnv` uses `waitForDeploymentFunc`, which updates its local `deployment` variable, but neither `waitForDeploymentEnv` nor `waitForDeploymentFunc` updates the `deployment` variable that is declared inside the test.  I'll fix the test to do a `Get` before logging.

test/e2e/operator_test.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/1be2d350d8a8948114c23963deca73ba0e2d5f19..9d2f210939366b2bc00cc739f2dec3e8e3d649e0.

test/e2e/operator_test.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/1be2d350d8a8948114c23963deca73ba0e2d5f19..9d2f210939366b2bc00cc739f2dec3e8e3d649e0, and https://github.com/openshift/cluster-ingress-operator/pull/1408/commits/a84698847d42367856b448ace60ae0a7e342c847 for `waitForDeploymentEnvVar`.

test/e2e/operator_test.go
Miciah: Done in https://github.com/openshift/cluster-ingress-operator/compare/1be2d350d8a8948114c23963deca73ba0e2d5f19..9d2f210939366b2bc00cc739f2dec3e8e3d649e0.

rhamini3:
reverified bug and confirmed that featuregate enablement does not time out

```
 % oc -n openshift-ingress get deployment router-default -oyaml | grep ROUTER_HAPROXY_CONFIG_MANAGER  -A1                                                                            
        - name: ROUTER_HAPROXY_CONFIG_MANAGER
          value: "true"
iamin@iamin-mac openshift-tests-private % oc -n openshift-ingress get pods
NAME                              READY   STATUS    RESTARTS   AGE
router-default-756b694859-hfq2l   1/1     Running   0          9m8s
router-default-756b694859-t5dqm   1/1     Running   0          9m8s
iamin@iamin-mac openshift-tests-private % oc -n openshift-ingress logs router-default-756b694859-hfq2l | grep time
iamin@iamin-mac openshift-tests-private % 
```

/verified by @rhamini3

gcs278:
flake on TestManagedDNSToUnmanagedDNSIngressController in e2e-azure-operator.
```
unmanaged_dns_test.go:187: DNSRecord zone expected to have status=Unknown but got status=True
```

Claude says:

> The test creates an IngressController with managed DNS, then switches it to dnsManagementPolicy=Unmanaged. After the switch, it expects the DNSRecord condition status to become Unknown    
>   (since DNS is no longer managed), but the condition remained True.
> 
> This is a timing/reconciliation race — the DNS record controller hadn't reconciled the status change before the test's polling timed out.

only 1 flake I see, probably not work pursuing.

/unhold
/test e2e-azure-operator

Miciah:
Merging of this PR missed the release-4.22 branch cut.  We need this fix in the release-4.22 branch to fix CI failures there.  

/cherry-pick release-4.22

PR #1401, opened 2026-03-24
OCPBUGS-3917: filter non-FIPS TLS 1.3 ciphers from ROUTER_CIPHERSUITES on FIPS clusters

bentito:
## Summary

On FIPS-enabled clusters, `TLS_CHACHA20_POLY1305_SHA256` is not FIPS-compliant (OpenSSL's FIPS policy excludes CHACHA20). When this cipher appears in HAProxy's `ssl-default-bind-ciphersuites` and a TLS 1.3 client offers it first, HAProxy fails the handshake rather than negotiating the next offered cipher.

The default `Intermediate`, `Old`, and `Modern` TLS profiles all include this cipher. The CIO unconditionally passed it to the router via `ROUTER_CIPHERSUITES`, causing reencrypt route TLS handshake failures on FIPS clusters when clients offered CHACHA20 first.

## Fix

**`deployment.go`**: Added a package-level `isFIPSEnabled` var (a `func() bool`) that detects FIPS mode by checking whether `TLS_CHACHA20_POLY1305_SHA256` is present in `crypto/tls.CipherSuites()` — it is absent when the Go runtime operates under FIPS constraints. When FIPS mode is detected, the cipher is filtered from `tls13Ciphers` before `ROUTER_CIPHERSUITES` is set. The two remaining FIPS-compliant TLS 1.3 ciphers (`TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384`) are retained, so TLS 1.3 continues to work normally.

**`controller.go`**: `validateTLSSecurityProfile` now rejects custom profiles whose *only* TLS 1.3 cipher suites are non-FIPS-compliant on a FIPS cluster, providing a clear error message so users understand why their configuration isn't effective.

## Tests

- `TestDesiredRouterDeploymentFIPS` — verifies CHACHA20 is stripped and FIPS-compliant ciphers are kept in FIPS mode; verifies non-FIPS mode passes all ciphers through untouched
- `Test_validateTLSSecurityProfileFIPS` — 4 sub-cases covering rejection/acceptance of custom profiles on a simulated FIPS cluster

## Related

OCPBUGS-3917 — TLS connection to reencrypt route fails on FIPS cluster depending on certain client cipher order

PR #1409, opened 2026-04-01
NE-2332: Implement ROUTER_CURVES environment variable

candita:
Set a default ROUTER_CURVES value at deploy time to support PQC compliance. If FIPS is enabled on the cluster, we cannot use the ML-KEM TLS curve because the OpenSSL FIPS provider in RHEL9 does not support ML-KEM.  Make the default ROUTER_CURVES observe this constraint.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -116,6 +116,7 @@ const (
 	RouterEnableCompression    = "ROUTER_ENABLE_COMPRESSION"
 	RouterCompressionMIMETypes = "ROUTER_COMPRESSION_MIME"
 	RouterBackendCheckInterval = "ROUTER_BACKEND_CHECK_INTERVAL"
+	RouterTLSCurves            = "ROUTER_CURVES"
 
 	RouterServiceHTTPPort  = "ROUTER_SERVICE_HTTP_PORT"
 	RouterServiceHTTPSPort = "ROUTER_SERVICE_HTTPS_PORT"
@@ -996,6 +997,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 			Value: strings.Join(tls13Ciphers, ":"),
 		})
 	}
+
+	// The default TLS supportedGroups (curves) include X25519MLKEM768,
+	// x25519, P-256, P-384, and P-521.
+	tlsCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
+	// If FIPS is enabled on this cluster, we cannot use
+	// ML-KEM or X25519 in the TLS supportedGroups (aka curves).
+	// ML-KEM and X25519 are not supported by OpenSSL FIPS 140-3.
+	// See https://redhat.atlassian.net/browse/TRT-2597 and Appendix D of
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
+	if isFIPSEnabled {
+		tlsCurves = "P-256:P-384:P-521"
+	}
+	env = append(env, corev1.EnvVar{
+		Name:  RouterTLSCurves,
+		Value: tlsCurves,
+	})
 
 	var minTLSVersion string
 	switch tlsProfileSpec.MinTLSVersion {
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -2781,3 +2781,55 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
 		})
 	}
 }
+
+// TestDesiredRouterDeploymentTLSCurves verifies that desiredRouterDeployment
+// sets ROUTER_CURVES as expected with and without FIPS enabled.
+func TestDesiredRouterDeploymentTLSCurves(t *testing.T) {
+	testCases := []struct {
+		name        string
+		fipsEnabled bool
+		expectedEnv []envData
+	}{
+		{
+			name:        "Include ML-KEM and X25519 when FIPS is not enabled",
+			fipsEnabled: false,
+			expectedEnv: []envData{
+				{"ROUTER_CURVES", true, "X25519MLKEM768:X25519:P-256:P-384:P-521"},
+			},
+		}, {
+			name:        "Exclude ML-KEM and X25519 when FIPS is enabled",
+			fipsEnabled: true,
+			expectedEnv: []envData{
+				{"ROUTER_CURVES", true, "P-256:P-384:P-521"},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ic := &operatorv1.IngressController{
+				Status: operatorv1.IngressControllerStatus{
+					EndpointPublishingStrategy: &operatorv1.EndpointPublishingStrategy{
+						Type: operatorv1.PrivateStrategyType,
+					},
+				},
+			}
+
+			wasFIPSEnabled := isFIPSEnabled
+			// Restore the real value of isFIPSEnabled in case it is used by other tests.
+			defer func() {
+				isFIPSEnabled = wasFIPSEnabled
+			}()
+			// Use the temporary test case status of FIPS enablement for testing.
+			isFIPSEnabled = tc.fipsEnabled
+
+			deployment, err := desiredRouterDeployment(ic, &Config{IngressControllerImage: ingressControllerImage}, &configv1.Ingress{}, &configv1.Infrastructure{}, &configv1.APIServer{}, &configv1.Network{}, false, false, nil, &configv1.Proxy{})
+			if err != nil {
+				t.Error(err)
+			}
+
+			if err := checkDeploymentEnvironment(t, deployment, tc.expectedEnv); err != nil {
+				t.Error(err)
+			}
+		})
+	}
+}


pkg/operator/controller/ingress/deployment.go
```diff
@@ -51,6 +51,9 @@ var isFIPSEnabled = func() bool {
 	return len(data) > 0 && data[0] == '1'
 }()
 
+// We only need to set this once.
+var fipsIsEnabled = isFIPSEnabled
```
Miciah: Why do we need `fipsIsEnabled`?

pkg/operator/controller/ingress/deployment.go
```diff
@@ -997,6 +1001,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 		})
 	}
 
+	// The default TLS supportedGroups (curves) include X25519MLKEM768,
+	// x25519, P-256, P-384, and P-521.
+	tlsCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
```
Miciah, line 1003: Is there a documentation reference for this?  https://github.com/openshift/router/pull/678 only enabled X25519MLKEM768, X25519, and P-256.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -423,6 +423,10 @@ func Test_desiredRouterDeployment(t *testing.T) {
 	if !reflect.DeepEqual(deployment.Spec.Template.Spec.Tolerations, expectedTolerations) {
 		t.Errorf("router Deployment has unexpected tolerations, expected: %#v, got: %#v", expectedTolerations, deployment.Spec.Template.Spec.Tolerations)
 	}
+	expectedCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
+	if isFIPSEnabled {
+		expectedCurves = "X25519:P-256:P-384:P-521"
+	}
```
Miciah: Maybe better to add a new test, something like this:
```go
// TestDesiredRouterDeploymentTLSCurves verifies that desiredRouterDeployment
// sets ROUTER_CURVES as expected with and without FIPS enabled.
func TestDesiredRouterDeploymentTLSCurves(t *testing.T) {
	testCases := []struct {
		name        string
		fipsEnabled bool
		expectedEnv []envData
	}{
		{
			name:       "Include ML-KEM when FIPS is not enabled",
			fipsEnabled: false,
			expectedEnv: []envData{
				{"ROUTER_CURVES", true, "X25519MLKEM768:X25519:P-256:P-384:P-521"},
			},
		}, {
			name:       "Exclude ML-KEM when FIPS is enabled",
			fipsEnabled: true,
			expectedEnv: []envData{
				{"ROUTER_CURVES", true, "X25519:P-256:P-384:P-521"},
			},
		},
	}
		t.Run(tc.name, func(t *testing.T) {
			ic := &operatorv1.IngressController{
				Status: operatorv1.IngressControllerStatus{
						Type: operatorv1.PrivateStrategyType,
					},
				},
			}

			wasFIPSEnabled := isFIPSEnabled
			defer func() {
				isFIPSEnabled = wasFIPSEnabled
			}()
			isFIPSEnabled = tc.fipsEnabled
			
			deployment, err := desiredRouterDeployment(ic, &Config{IngressControllerImage: ingressControllerImage}, &configv1.Ingress{}, &configv1.Infrastructure{}, &configv1.APIServer{}, &configv1.Network{}, false, false, nil, &configv1.Proxy{})
			if err != nil {
				t.Error(err)
			}

			if err := checkDeploymentEnvironment(t, deployment, tc.expectedEnv); err != nil {
				t.Error(err)
			}
		})
	}
}
```
This example assumes you remove `fipsIsEnabled`; change it to `isFIPSEnabled` within the test if you choose to keep `isFIPSEnabled`.

lihongan:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-fips

pkg/operator/controller/ingress/deployment.go
```diff
@@ -997,6 +1001,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	// If FIPS is enabled on this cluster, we cannot use
+	// ML-KEM in the TLS supportedGroups (aka curves).
+	// ML-KEM is not supported by the OpenSSL FIPS provider in RHEL9.
+	// See https://redhat.atlassian.net/browse/TRT-2597
+	if fipsIsEnabled {
+		tlsCurves = "X25519:P-256:P-384:P-521"
```
lihongan: `X25519` is not approved by FIPS yet, we could set `P-256:P-384:P-521` for FIPS currently.

Tested on FIPS cluster and got below erro:
```
	[ALERT]    (14) : config : Proxy 'fe_no_sni': unable to set SSL curves list to 'X25519:P-256:P-384:P-521' for bind 'unix@/var/lib/haproxy/run/haproxy-no-sni.sock' at [/var/lib/haproxy/conf/haproxy.config:170].
	Proxy 'fe_no_sni': unable to set SSL curves list to 'X25519:P-256:P-384:P-521' for bind 'unix@/var/lib/haproxy/run/haproxy-no-sni.sock' at [/var/lib/haproxy/conf/haproxy.config:170].
	[ALERT]    (14) : config : Fatal errors found in configuration.
```

pkg/operator/controller/ingress/deployment.go
lihongan: [NIST SP 800-52 Rev. 2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-52r2.pdf) explicitly recommends P-384 alongside P-256, but not mentioned P-521. 
And P-256, P-384, P-521 are all acceptable and listed in [NIST SP 800-186](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf).

pkg/operator/controller/ingress/deployment.go
candita: The FIPS value is set once on the cluster and cannot change over time, so I was attempting to save os.Read calls.  There may be a better way to do this.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -51,6 +51,9 @@ var isFIPSEnabled = func() bool {
 	return len(data) > 0 && data[0] == '1'
 }()
 
+// We only need to set this once.
+var fipsIsEnabled = isFIPSEnabled
+
```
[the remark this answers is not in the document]
candita: @coderabbitai The content in the file that is read by os.Read doesn't change.  Fix this so the function is never called more than once.

pkg/operator/controller/ingress/deployment.go
candita: I'll remove it, it was a misguided attempt.

pkg/operator/controller/ingress/deployment.go
candita: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf Appendix D has P-521 and I was also trying to use both https://github.com/haproxy/haproxy/blob/master/src/ssl_utils.c#L988 (which has P-521 and X25519, and more) and the recommendations made to @davidesalerno in https://github.com/openshift/api/pull/2583 (which has P-521).

I agree with @lihongan that X25519 is probably not supported in FIPS since it isn't listed in the NIST FIPS documentation.

pkg/operator/controller/ingress/deployment.go
Miciah: Ah, `isFIPSEnabled` was a function in an earlier iteration of #1401, but he changed it to a `bool` value that is initialized just once.

pkg/operator/controller/ingress/deployment.go
Miciah: So the set of curves comes from the predefined TLS security profiles (with the updates in https://github.com/openshift/api/pull/2583)?   And those profiles come from Mozilla: 
```go
	// The profiles are based on version 5.7 of the Mozilla Server Side TLS
	// configuration guidelines. The cipher lists consist of the configuration's
	// "ciphersuites" followed by the Go-specific "ciphers" from the guidelines.
	// See: https://ssl-config.mozilla.org/guidelines/5.7.json
```
Source: https://github.com/openshift/api/blob/ad9eb11110b6e704efa5000c12c5fb5db1685e8a/config/v1/types_tlssecurityprofile.go#L10-L13

It might be helpful to have a similar code comment here, something like, "This list of curves comes from the Mozilla Service Side TLS configuration guidelines at <https://ssl-config.mozilla.org/guidelines/5.7.json>."

candita: There are no PQC curves in the Mozilla guidelines.

pkg/operator/controller/ingress/deployment.go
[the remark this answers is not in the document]
candita: Fixed

pkg/operator/controller/ingress/deployment_test.go
candita: Fixed, thanks.

pkg/operator/controller/ingress/deployment.go
candita: Fixed.

pkg/operator/controller/ingress/deployment.go
Miciah: > There are no PQC curves in the Mozilla guidelines.

Oh, then https://github.com/openshift/api/pull/2583 should have a similar citation.  I'll post a comment there.  Thanks for pointing that out!

Miciah: But wait, https://ssl-config.mozilla.org/guidelines/5.7.json does mention curves:
```json
{
    "version": 5.7,
    "href": "https://ssl-config.mozilla.org/guidelines/5.7.json",
    "configurations": {
        "modern": {
            "certificate_curves": ["prime256v1", "secp384r1"],
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        },
        "intermediate": {
            "certificate_curves": ["prime256v1", "secp384r1"],
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        },
        "old": {
            "certificate_curves": ["prime256v1", "secp384r1"],
            /* ... */
            "tls_curves": ["X25519", "prime256v1", "secp384r1"],
            /* ... */
        }
    }
}
```

Miciah: Following up here: https://github.com/openshift/api/pull/2583#discussion_r3033991019

candita: Also, HAProxy will not parse `X25519` or any ML-KEM based curves, not even in the master branch.  https://github.com/haproxy/haproxy/blob/master/src/ssl_utils.c#L988  

~~And, FIPS enabled or not doesn't matter in terms of the parsing.~~
Update: FIPS doesn't support X25519 or X25519MLKEM768.  When we move to golang 1.26, we will pick up two new MKLEM groups.

Miciah: Are you sure that that code is relevant?  As far as I can tell, the `curves_list` table is only used by the `curves2nid` and `curveid2str` functions, which are only used in the ACME, JWS, and SSL trace code.  Also, you are referencing code in the main branch, which is very different from the HAProxy 2.8 code.  

From what I can tell, HAProxy passes the argument for `ssl-default-bind-curves` straight through to OpenSSL:
```c
/* parse the "ssl-default-bind-ciphers" / "ssl-default-server-ciphers" keywords
 * in global section. Returns <0 on alert, >0 on warning, 0 on success.
 */
static int ssl_parse_global_ciphers(char **args, int section_type, struct proxy *curpx,
                                    const struct proxy *defpx, const char *file, int line,
                                    char **err)
{
	char **target;

	target = (args[0][12] == 'b') ? &global_ssl.listen_default_ciphers : &global_ssl.connect_default_ciphers;

	if (too_many_args(1, args, err, NULL))
		return -1;

	if (*(args[1]) == 0) {
		memprintf(err, "global statement '%s' expects a cipher suite as an argument.", args[0]);
		return -1;
	}

	free(*target);
	*target = strdup(args[1]);
	return 0;
}
```
Source: https://github.com/haproxy/haproxy/blob/aaba8d090ec663906af1fecbad83c26bbccf0761/src/cfgparse-ssl.c#L244-L266

```c
#if defined(SSL_CTX_set1_curves_list)
	if (global_ssl.listen_default_curves && !conf->ssl_conf.curves)
		conf->ssl_conf.curves = strdup(global_ssl.listen_default_curves);
#endif
```
Source: https://github.com/haproxy/haproxy/blob/aaba8d090ec663906af1fecbad83c26bbccf0761/src/cfgparse-ssl.c#L1248-L1251
```c
	conf_curves = (ssl_conf && ssl_conf->curves) ? ssl_conf->curves : bind_conf->ssl_conf.curves;
	if (conf_curves) {
		if (!SSL_CTX_set1_curves_list(ctx, conf_curves)) {
```
Source: https://github.com/haproxy/haproxy/blob/aaba8d090ec663906af1fecbad83c26bbccf0761/src/ssl_sock.c#L4750-L4752

Miciah:
Thanks!
/approve
/lgtm

Miciah:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-fips

gcs278:
Just adding my LGTM for completeness. CI is looking good (at least no complete router failure) for FIPs and non-FIPs.
/lgtm

pkg/operator/controller/ingress/deployment.go
candita: You're right.  @gcs278 found the same.

In case you didn't already know, the PQC team chose to use only curves/groups supported for the current Golang version, from https://pkg.go.dev/crypto/tls#CurveID.  So, when we move to 1.26, we will pick up two new curves.  And those two new curves are supported in FIPS, per the FIPS team.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -997,6 +998,22 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	// The default TLS supportedGroups (curves) include X25519MLKEM768,
+	// x25519, P-256, P-384, and P-521.
+	tlsCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
+	// If FIPS is enabled on this cluster, we cannot use
+	// ML-KEM or X25519 in the TLS supportedGroups (aka curves).
+	// ML-KEM and X25519 are not supported by OpenSSL FIPS 140-3.
+	// See https://redhat.atlassian.net/browse/TRT-2597 and Appendix D of
+	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
+	if isFIPSEnabled {
+		tlsCurves = "P-256:P-384:P-521"
```
candita, line 1010: Also note that the curves are supposed to be selected in order, meaning in a FIPS situation P-256 is selected first.  It is faster and less resource intensive than P-384 and P-521, but less secure.

candita:
> fail [github.com/openshift/origin/test/extended/authorization/scc.go:76]: 1 pods failed before test on SCC errors
> Error creating: pods "operator-controller-controller-manager-85dd86c94c-" is forbidden: unable to validate against any security context constraint: provider "privileged": Forbidden: not usable by user or serviceaccount for ReplicaSet.apps/v1/operator-controller-controller-manager-85dd86c94c -n openshift-operator-controller happened 11 times}

/test e2e-aws-ovn

candita:
e2e-aws-operator-techpreview success is dependent on
https://github.com/openshift/cluster-ingress-operator/pull/1408

lihongan:
/verified by @lihongan 

the CI look good and no issue with FIPS during pre-merge testing
```
$ oc get clusterversion
version   4.22.0-0-2026-04-07-013150-test-ci-ln-13wqfbk-latest   True        False         7m19s   Cluster version is 4.22.0-0-2026-04-07-013150-test-ci-ln-13wqfbk-latest

// check FIPS mode on node
sh-5.1# fips-mode-setup --check
FIPS mode is enabled.

$ oc -n openshift-ingress get deployment/router-default -oyaml | grep CURVES -A1
        - name: ROUTER_CURVES
          value: P-256:P-384:P-521

$ oc -n openshift-ingress rsh router-default-5f59f5c59-b2mbp
sh-5.1$ grep curves haproxy.config 
  ssl-default-bind-curves P-256:P-384:P-521
```

pkg/operator/controller/ingress/deployment.go
Miciah: Does that mean that the source of truth for the list of curves is the intersection of what NIST recommends and what Go supports?

candita: Right now it seems to be the combination of what Go 1.25 supports minus whether the cluster is FIPS, which doesn't support X25519 or X25519MLKEM768.

PR #1410, opened 2026-04-02
CORS-4372: Create IPv6 records for GCP

barbacbd:
pkg/util/aws/dualstack.go:

- Removed the "aws specific" capability of determining stack type

pkg/util/ipfamily/ipfamily.go:

- Created the common function to determine stack type dynamically.

pkg/util/ipfamily/ipfamily_test.go:

- Generated tests for the stack determination

pkg/dns/gcp/provider.go:

- Added dual-stack DNS support for GCP *.apps records

pkg/operator/controller/dns/controller.go:

- Add IP family to the controller call.

### DIFF
--- a/pkg/dns/gcp/provider.go
+++ b/pkg/dns/gcp/provider.go
@@ -3,6 +3,7 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


gcs278:
@barbacbd are you looking for a review for this PR or is it still on hold?

PR #1411, opened 2026-04-02
OCPBUGS-81550: AWS, Azure and GCP ClusterHostedDNS: Add Ingress LB IPs to Infra CR after masters up

sadasu:
Before the Ingress default Load Balancer's IP address is added to Infrastructure CR, make sure MCO is done bringing up the control plane nodes. This ensures that the Infra CR used by MCO to genarate master ignition would not be different from the master ignition files generated by the installer.

Background: MCO uses the infrastructure CR to generate its master ignition. Adding the ingress LB IPs before MCO has done that would cause the ignition file generated by the MCO to be different from the one generated by the Installer (without the ingress LB IPs). This race condition prevents masters (usually master 1 or 2) from becoming ready. This race condition has been observed on both AWS and Azure CI runs with userProvisionedDNS enabled.
This fix would cause a reboot of the master nodes when the Infra CR is eventually mutated but we are OK with that.

Fixes:  https://redhat.atlassian.net/browse/OCPBUGS-81550


pkg/operator/controller/ingress/controller.go
```diff
@@ -1416,8 +1416,36 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
 	return true, nil
 }
 
-// computeUpdatedInfraFromService updates PlatformStatus for AWS, Azure and GCP with Ingress LB IPs when the DNSType is `ClusterHosted`.
-func computeUpdatedInfraFromService(service *corev1.Service, infraConfig *configv1.Infrastructure) (bool, error) {
+// isMCOReady checks if the machine-config-operator is ready by examining its
+// ClusterOperator conditions. Returns true if machine-config-operator is
+// Available=True and Progressing=False.
+func isMCOReady(ctx context.Context, client client.Client) bool {
```
sadasu, line 1460: @yuqi-zhang @patrickdillon  Is this a correct way to check if MCO is done with generating ignition for Master nodes so that it is safe to add the Ingress LB IPs to the Infra CR?

gpei:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

pkg/operator/controller/ingress/controller.go
patrickdillon: sorry I missed this ping originally. It does seem like this would work, MCO does not go Progressing=False until after `syncRequiredMachineConfigPools` is completed, which is where we get this error.

jinyunma:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

jinyunma:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

pkg/operator/controller/ingress/controller.go
```diff
@@ -1416,8 +1416,36 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+	mco := &configv1.ClusterOperator{}
+	if err := client.Get(ctx, types.NamespacedName{Name: "machine-config-operator"}, mco); err != nil {
```
jinyunma: Seems here uses incorrect operator name, should be `machine-config`

[Job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-cluster-ingress-operator-1411-nightly-4.22-e2e-azure-custom-dns-techpreview/2041364025949818880) failed as ingress degraded, in ingress controller logs,
```
2026-04-07T05:36:52.725Z	INFO	operator.ingress_controller	ingress/controller.go:1331	Skipping Azure Infrastructure status update: machine-config-operator not yet ready
```
Actually, machine-config operator was already ready.

sadasu: Thanks @jinyunma ! Yes, I had the name of MCO wrong here. Corrected.

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

jinyunma:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

gpei:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

gpei:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-custom-dns-techpreview

gpei:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

gpei:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

gpei:
@sadasu Over the past several test runs, the installation Azure custom-dns job has been consistently stable and the e2e tests are performing well: https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/openshift-cluster-ingress-operator-1411-nightly-4.22-e2e-azure-custom-dns-techpreview

I've also noticed that in the AWS custom-dns job, there's one test case that consistently fails:                                
`[sig-instrumentation] Prometheus [apigroup:image.openshift.io] when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured`

I used AI to help analyze this failure, and it suggested that the issue stems from the CoreDNS configuration file. Specifically, the Corefile is adding IPv6 records for *.apps domain, but these records are pointing to IPv4 addresses instead. This mismatch is causing numerous IPv6 resolution failures.                                                                                                 This might be an area where we need to fix the Corefile configuration to properly handle the address records.

sadasu:
> I used AI to help analyze this failure, and it suggested that the issue stems from the CoreDNS configuration file. Specifically, the Corefile is adding IPv6 records for *.apps domain, but these records are pointing to IPv4 addresses instead. This mismatch is causing numerous IPv6 resolution failures. This might be an area where we need to fix the Corefile configuration to properly handle the address records.

@gpei Thanks for looking into this failure. I found that the errors were reported in the CoreDNS pod started by the cluster-dns-operator. Any changes I make to their CoreFile would have to be in collaboration with them.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1455,12 +1490,20 @@ func computeUpdatedInfraFromService(service *corev1.Service, infraConfig *config
...
 				platformStatus.AWS.CloudLoadBalancerConfig.ClusterHosted.IngressLoadBalancerIPs = ingressLBIPs
+				log.Info("Updating AWS Infrastructure status with Ingress Load Balancer IPs after machine-config is ready")
 				return true, nil
 			}
 		}
 		return false, nil
 	case configv1.AzurePlatformType:
 		if platformStatus.Azure != nil && platformStatus.Azure.CloudLoadBalancerConfig != nil && platformStatus.Azure.CloudLoadBalancerConfig.DNSType == configv1.ClusterHostedDNSType {
+			// For Azure with userProvisionedDNS (ClusterHosted DNS), only update Infrastructure API
+			// after machine-config-operator is ready to avoid installation issues.
```
patrickdillon: nit: perhaps we want to make this comment (and the duplicate one) more explicit that we are waiting to mutate the infrastructure CR until after the MCO has rendered its master ignition in order to avoid the MCO's validation against this. We know that this will cause a reboot, but we are ok with that in the case of custom dns.

sadasu:
/verified by CI

Updated a comment in the code and so lost the verified label. Adding it back.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1416,8 +1416,36 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+		if condition.Type == configv1.OperatorProgressing {
+			progressingCondition = (condition.Status == configv1.ConditionTrue)
+		}
+	}
+
+	// MCO is ready when Available=True and Progressing=False
+	return availableCondition && !progressingCondition
+}
+
+// computeUpdatedInfraFromService updates PlatformStatus for Azure, GCP and AWS with Ingress LB IPs when the DNSType is `ClusterHosted`.
```
Miciah: Is there a reason you changed "AWS, Azure and GCP" to "Azure, GCP and AWS"?  It seems like a gratuitous change that makes the diff noisier.

pkg/operator/controller/ingress/controller.go
```diff
@@ -1416,8 +1416,36 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
-// computeUpdatedInfraFromService updates PlatformStatus for AWS, Azure and GCP with Ingress LB IPs when the DNSType is `ClusterHosted`.
-func computeUpdatedInfraFromService(service *corev1.Service, infraConfig *configv1.Infrastructure) (bool, error) {
+// isMCOReady checks if the machine-config-operator is ready by examining its
+// ClusterOperator conditions. Returns true if machine-config-operator is
+// Available=True and Progressing=False.
+func isMCOReady(ctx context.Context, client client.Client) bool {
+	mco := &configv1.ClusterOperator{}
+	if err := client.Get(ctx, types.NamespacedName{Name: "machine-config"}, mco); err != nil {
+		// If machine-config is not available, assume it's not ready
+		log.V(2).Info("failed to get machine-config ClusterOperator, assuming not ready", "error", err)
```
Miciah: Would it make more sense to distinguish between not-found errors and other errors?
```suggestion
		if kerrors.IsNotFound(err) {
			log.V(2).Info("failed to get machine-config ClusterOperator, assuming not ready", "error", err)
		} else {
			log.Error(err, "failed to get machine-config ClusterOperator")
		}
```
You could even add a return error value, return the error in the latter case, and bubble it up to the reconciliation loop, but using `log.Error` at least raises the prominence of the log message.

pkg/operator/controller/ingress/controller.go
Miciah: I am curious why this issue does not affect GCP.  

- Are we certain that it does *not* affect GCP?
- Would the be any harm in doing the `isMCOReady` check always, irrespective of the platform?  
- If we indeed do not need the `isMCOReady` check on GCP, can you add a comment in the "GCP" case explaining why it doesn't have the check when AWS and Azure do need it?

Miciah: On a related note, I would prefer if we didn't need to pass the context and client to `computeUpdatedInfraFromService`.  If you did change the logic to perform the `isMCOReady` check on all platforms, you could then move the `isMCOReady` check from inside `computeUpdatedInfraFromService` to `ensureIngressController`, and then you wouldn't need to pass the context and client to `computeUpdatedInfraFromService`.

Miciah:
Regarding the commit messages:

- You have a typo: "genarate" → "generate".
- Please add a reference to https://redhat.atlassian.net/browse/OCPBUGS-81550.

pkg/operator/controller/ingress/controller.go
sadasu: @Miciah currently I don't have evidence of this race condition occurring in GCP. I do like your suggestion. So, please look for an update.

pkg/operator/controller/ingress/controller.go
sadasu: I am happy to log an error message here using `log.Error`. I am not sure if there is any specific action we would take (other than logging it) if the error was bubbled up.

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-custom-dns

pkg/operator/controller/ingress/controller.go
```diff
@@ -1416,6 +1429,62 @@ func (r *reconciler) allRouterPodsDeleted(ingress *operatorv1.IngressController)
...
+
+	availableCondition := false
+	progressingCondition := true // Default to true (progressing) if not found
+
+	for _, condition := range mco.Status.Conditions {
+		if condition.Type == configv1.OperatorAvailable {
+			availableCondition = (condition.Status == configv1.ConditionTrue)
+		}
+		if condition.Type == configv1.OperatorProgressing {
+			progressingCondition = (condition.Status == configv1.ConditionTrue)
```
Miciah: If either status condition has status "Unknown", this logic treats that the same as "False".  So if "Available" is "Unknown", then we might assume that MCO is not ready and stall; and if "Progressing" is "Unknown", then we might assume MCO is ready when it might not be.  It might be safer to do something like this:
```suggestion
		if condition.Type == configv1.OperatorAvailable && condition.Status == configv1.ConditionTrue {
			availableCondition = true
		}
		if condition.Type == configv1.OperatorProgressing && condition.Status == configv1.ConditionFalse {
			progressingCondition = false
```
However, if MCO's status conditions never have "Unknown" status in practice, or if it is all right to treat "Unknown" as "True" for our purposes here, then the current logic should be fine.

Miciah:
/approve

If https://github.com/openshift/cluster-ingress-operator/pull/1411#discussion_r3060338432 isn't a problem,
/lgtm

pkg/operator/controller/ingress/controller.go
Miciah: Bubbling the error up would make the controller retry the reconciliation immediately, with exponential backoff.

pkg/operator/controller/ingress/controller.go
Miciah: I'm a little surprised that CodeRabbit didn't comment on this.  Hm, reviews seem to be paused.  

@coderabbitai review

Miciah: Thanks, @coderabbitai!  Yeah, I noticed the `context.TODO()` too, but I didn't comment on it as the function already uses `context.TODO()` in other places.  As a separate followup, we should indeed replace `context.TODO()` throughout by threading `ctx` through from the top-level `Reconcile` method.

Miciah: Thanks, @coderabbitai!  No need to open a GitHub issue.  We use Jira to track these sorts of issues.  I have stubbed out a story in our "Refactoring and code cleanups in 4.23" epic: https://redhat.atlassian.net/browse/NE-2584

pkg/operator/controller/ingress/controller.go
```diff
@@ -1328,13 +1328,26 @@ func (r *reconciler) ensureIngressController(ci *operatorv1.IngressController, d
...
+		// For platforms with ClusterHosted DNS, only update Infrastructure API after machine-config is ready.
+		// Background: MCO uses the infrastructure CR to generate its master ignition. Adding the ingress
+		// LB IPs before MCO has done that would cause the ignition file generated by the MCO to be
+		// different from the one generated by the Installer (without the ingress LB IPs). This race
+		// prevents masters (usually master 1 or 2) from becoming ready.
+		// This fix would cause a reboot of the master nodes when the Infra CR is eventually mutated but
+		// we are OK with that.
+		if needsClusterHostedDNS(infraConfig) {
+			if !isMCOReady(context.TODO(), r.client) {
+				log.Info("Skipping Infrastructure status update: machine-config not yet ready")
```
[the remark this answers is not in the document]
Miciah: Related discussion: https://github.com/openshift/cluster-ingress-operator/pull/1411#discussion_r3059343877

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-custom-dns

pkg/operator/controller/ingress/controller.go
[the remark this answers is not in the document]
sadasu: @Miciah which of the above 2 options is preferable, The retry logic or adding a watch for when MCO becomes Available=True, Progressing=False?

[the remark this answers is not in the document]
Miciah: Hm, good question.  In general, I prefer an event-driven approach to polling, so I am leaning towards Option 2: watch clusteroperators, with two predicates: (1) the name matches "machine-config", and (2) the "Available" or "Progressing" status condition has changed status.

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-azure-custom-dns-techpreview

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-custom-dns

Miciah:
The PR in its current form is strictly better than the PR in the form that I previously `/lgtm`'d, so
/lgtm

We agreed that adding a watch as discussed in https://github.com/openshift/cluster-ingress-operator/pull/1411#discussion_r3060430719 can be done in a follow-up.

sadasu:
Worked for AWS and Azure. Waiting for GCP.

sadasu:
/payload-job periodic-ci-openshift-release-main-nightly-4.22-e2e-gcp-custom-dns

PR #1414, opened 2026-04-08
[release-4.21] OCPBUGS-82146: Do not take over OSSM subscription

rikatz:

This is a manual backport of https://github.com/openshift/cluster-ingress-operator/pull/1398 because release-4.22/master are already using noOLM approach and cherry-picking the fix would be conflicting

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -103,7 +103,9 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 	}
 
 	isServiceMeshSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
-		return o.GetName() == operatorcontroller.ServiceMeshOperatorSubscriptionName().Name
+		subscription, ok := o.(*operatorsv1alpha1.Subscription)
+		return ok && subscription.Spec != nil &&
+			subscription.Spec.Package == operatorcontroller.ServiceMeshOperatorSubscriptionPackage
 	})
 	if err = c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.Subscription{},
 		reconciler.enqueueRequestForSomeGatewayClass(), isServiceMeshSubscription)); err != nil {
--- a/pkg/operator/controller/gatewayclass/controller.go
+++ b/pkg/operator/controller/gatewayclass/controller.go
@@ -248,12 +248,21 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	if v, ok := gatewayclass.Annotations[subscriptionVersionOverrideAnnotationKey]; ok {
 		ossmVersion = v
 	}
-	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion); err != nil {
+
+	_, subscription, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion)
+	if err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
+	} else if subscription == nil {
+		log.Info("No OSSM subscription available; skipping install plan enforcement")
+	} else if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+		log.Info("Found an existing OSSM subscription with another owner; installation skipped",
+			"namespace", subscription.Namespace, "name", subscription.Name)
+	} else {
+		if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
+			errs = append(errs, err)
+		}
 	}
-	if _, _, err := r.ensureServiceMeshOperatorInstallPlan(ctx, ossmVersion); err != nil {
-		errs = append(errs, err)
-	}
+
 	istioVersion := r.config.IstioVersion
 	if v, ok := gatewayclass.Annotations[istioVersionOverrideAnnotationKey]; ok {
 		istioVersion = v
--- a/pkg/operator/controller/gatewayclass/subscription.go
+++ b/pkg/operator/controller/gatewayclass/subscription.go
@@ -35,6 +35,14 @@ const (
 // for servicemeshoperator is present and returns a Boolean indicating whether
 // it exists, the subscription if it exists, and an error value.
 func (r *reconciler) ensureServiceMeshOperatorSubscription(ctx context.Context, catalog, channel, version string) (bool, *operatorsv1alpha1.Subscription, error) {
+	conflictingSubscription, err := r.findConflictingSubscription(ctx)
+	if err != nil {
+		return false, nil, err
+	}
+
+	if conflictingSubscription != nil {
+		return true, conflictingSubscription, nil
+	}
 	name := operatorcontroller.ServiceMeshOperatorSubscriptionName()
 	have, current, err := r.currentSubscription(ctx, name)
 	if err != nil {
@@ -92,7 +100,7 @@ func desiredSubscription(name types.NamespacedName, gwapiOperatorCatalog, gwapiO
 				},
 			},
 			InstallPlanApproval:    operatorsv1alpha1.ApprovalManual,
-			Package:                "servicemeshoperator3",
+			Package:                operatorcontroller.ServiceMeshOperatorSubscriptionPackage,
 			CatalogSource:          gwapiOperatorCatalog,
 			CatalogSourceNamespace: "openshift-marketplace",
 			StartingCSV:            gwapiOperatorVersion,
@@ -182,6 +190,9 @@ func (r *reconciler) currentInstallPlan(ctx context.Context, version string) (bo
 	if err != nil {
 		return false, nil, err
 	}
+	if subscription == nil {
+		return false, nil, nil
+	}
 	installPlans := &operatorsv1alpha1.InstallPlanList{}
 	if err := r.client.List(ctx, installPlans, client.InNamespace(operatorcontroller.OpenshiftOperatorNamespace)); err != nil {
 		return false, nil, err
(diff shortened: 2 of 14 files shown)


rikatz:
@coderabbitai review

rikatz:
@coderabbitai this is a manual backport of https://github.com/openshift/cluster-ingress-operator/pull/1398 please verify if it adherent with the original code.

pkg/operator/controller/gatewayclass/subscription.go
```diff
@@ -251,6 +262,38 @@ func (r *reconciler) updateInstallPlan(ctx context.Context, current, desired *op
...
+		// exists.
+		sub := &subscriptionList.Items[i]
+		if sub.Spec != nil && sub.Spec.Package == operatorcontroller.ServiceMeshOperatorSubscriptionPackage {
+			if _, ok := sub.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+				return sub, nil
+			}
+		}
+	}
+	// No conflicting subscription found, return null
+	return nil, nil
```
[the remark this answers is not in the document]
rikatz: fixed, good catch

rhamini3:
marking PR as pre-merge-verified

Installed OSSM on default namespace, installed gatewayClass, checked operator logs to make sure OSSM installation was skipped
```
% oc -n default get sub,csv,pod
NAME                                                     PACKAGE                SOURCE             CHANNEL
subscription.operators.coreos.com/servicemeshoperator3   servicemeshoperator3   redhat-operators   stable

NAME                                                                     DISPLAY                            VERSION   REPLACES                      PHASE
clusterserviceversion.operators.coreos.com/servicemeshoperator3.v3.3.1   Red Hat OpenShift Service Mesh 3   3.3.1     servicemeshoperator3.v3.3.0   Succeeded

NAME                                         READY   STATUS    RESTARTS   AGE
pod/servicemesh-operator3-67b6c797c5-5kmnf   1/1     Running   0          38s


% oc -n openshift-operators get sub,pod
No resources found in openshift-operators namespace.
iamin@iamin-mac ~ % oc get gc
openshift-default   openshift.io/gateway-controller/v1   True       28s


2026-04-10T16:25:28.011Z	INFO	operator.gatewayclass_controller	controller/controller.go:119	Found an existing OSSM subscription with another owner; installation skipped	{"namespace": "default", "name": "servicemeshoperator3"}
2026-04-10T16:25:32.341Z	INFO	operator.gatewayclass_controller	controller/controller.go:119	reconciling	{"request": {"name":"openshift-default"}}
2026-04-10T16:25:32.346Z	INFO	operator.gatewayclass_controller	controller/controller.go:119	Found an existing OSSM subscription with another owner; installation skipped	{"namespace": "default", "name": "servicemeshoperator3"}
2026-04-10T16:25:36.736Z	INFO	operator.gatewayclass_controller	controller/controller.go:119	reconciling	{"request": {"name":"openshift-default"}}
2026-04-10T16:25:36.740Z	INFO	operator.gatewayclass_controller	controller/controller.go:119	Found an existing OSSM subscription with another owner; installation skipped	{"namespace": "default", "name": "servicemeshoperator3"}
```

/verified by @rhamini3

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -103,7 +103,9 @@ func NewUnmanaged(mgr manager.Manager, config Config) (controller.Controller, er
 	}
 
 	isServiceMeshSubscription := predicate.NewPredicateFuncs(func(o client.Object) bool {
-		return o.GetName() == operatorcontroller.ServiceMeshOperatorSubscriptionName().Name
+		subscription, ok := o.(*operatorsv1alpha1.Subscription)
+		return ok && subscription.Spec != nil &&
+			subscription.Spec.Package == operatorcontroller.ServiceMeshOperatorSubscriptionPackage
```
gcs278, line 108: Good catch. Noted that this predicate change isn't in 4.22+'s OLM path, but it's dead code there.

Generally I would ask you to add this to the main branch and backport this additionally, but as long as we never re-enable OLM in 4.22, it shouldn't matter too much.

gcs278:
LGTM - identical changes to https://github.com/openshift/cluster-ingress-operator/pull/1398, the only addition is the subscription predicate change (watching by package instead of name), which isn't in 4.22 but is dead code there, so not blocking on a separate bug for it. 

/approve
/lgtm

pkg/operator/controller/gatewayclass/controller.go
```diff
@@ -248,12 +250,21 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-	if _, _, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion); err != nil {
+
+	_, subscription, err := r.ensureServiceMeshOperatorSubscription(ctx, ossmCatalog, ossmChannel, ossmVersion)
+	if err != nil {
 		errs = append(errs, fmt.Errorf("failed to ensure ServiceMeshOperatorSubscription: %w", err))
+	} else if subscription == nil {
+		log.Info("No OSSM subscription available; skipping install plan enforcement")
+	} else if _, ok := subscription.Annotations[operatorcontroller.IngressOperatorOwnedAnnotation]; !ok {
+		log.Info("Found an existing OSSM subscription with another owner; installation skipped",
+			"namespace", subscription.Namespace, "name", subscription.Name)
```
candita, line 261: Don't we also want to skip everything else here?  Why do we ensure the Istio and start the Istio watch if this isn't our subscription?

pkg/operator/controller/gatewayclass/controller.go
candita: What if the conflicting subscription is for a previous package, e.g. OSSM 2.x ?

rikatz: Nothing, as nothing happens before. Having OSSM 2.x is not part of our scope and in fact IIRC we tell that they are conflicting. We didn't replaced it before, we do not replace it now

pkg/operator/controller/gatewayclass/controller.go
rikatz: Because the subscription can be removed and then we want to ensure ours.

BTW we just watch OSSM subscription, and this assertion is related to see if the subscription contains our annotation or not (if contains ours, means we installed it, otherwise someone else installed OSSM 3.x and we don't do anything)

candita: But we do go on to ensure istio and start a istio watch.  Why don't we skip that too?

rikatz: Because we are still installing Istio with the current OSSM subscription from the user

gcs278:
There's a discussion about backporting the no-OLM feature gate (as disabled) to older versions. If that lands first, this would cherry-pick cleanly. And we'd avoid having to manually resolve conflicts later if the no-OLM backport does end up happening.

gcs278:
/hold

As a part of the NoOLM assessment - I'm also going to do a quick assessment of this solution (as it's the alternative solution to backporting NO-OLM). I'd like to discuss before we merge this in 4.21.

pkg/operator/controller/status/controller.go
```diff
@@ -663,13 +652,13 @@ func computeGatewayAPIInstallDegradedCondition(state operatorState) configv1.Clu
...
 				// version is supported. Warn the user that the installed OSSM version may be incompatible.
-				warnings = append(warnings, fmt.Sprintf("Found version %s, but operator-managed Gateway API expects version %s. Operator-managed Gateway API may not work as intended.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion))
+				warnings = append(warnings, fmt.Sprintf("Found version %s on %s/%s, but operator-managed Gateway API expects version %s. Operator-managed Gateway API may not work as intended.", subscription.Status.InstalledCSV, subscription.Namespace, subscription.Name, state.expectedGatewayAPIOperatorVersion))
 			case versionDiff > 0:
 				// Installed version is older than expected. Gateway API install will not work, since the correct Istio
 				// version won't be supported.
 				conflicts = append(conflicts, fmt.Sprintf("Installed version %s does not support operator-managed Gateway API. Install version %s or uninstall %s/%s to enable functionality.", subscription.Status.InstalledCSV, state.expectedGatewayAPIOperatorVersion, subscription.Namespace, subscription.Name))
```
gcs278, line 659: I put together a compatibility matrix for detached subscriptions after the z-stream bumps land:

|                  | OSSM 3.1 (1.24-1.26) | OSSM 3.2 (1.24-1.27) | OSSM 3.3 (1.26-1.28) | OSSM 3.4 (1.28-1.30) |
| ---------------- | -------------------- | -------------------- | -------------------- | -------------------- |
| 4.19.z (v1.26.8) | ✓                    | ✓                    | ✓                    | N/A                  |
| 4.20.z (v1.28.5) | ✗                    | ✗                    | ✓                    | ✓                    |
| 4.21.z (v1.28.5) | ✗                    | ✗                    | ✓                    | ✓                    |

✗ = CRD validation rejects (Istio version not in enum) or Istio EOL Condition

After the 4.20 and 4.21 z-stream bumps to v1.28.5, a customer managing their own subscription needs to be on at least OSSM 3.3.x for it to work. OSSM 3.2 and earlier don't include Istio 1.28.x in their CRD schema, so CIO's Istio CR creation will be rejected by the API server or get an EOL condition. The current `Degraded=True` for "version too old" is correct for these y-stream mismatches.

However, within a compatible OSSM minor (e.g., customer on 3.3.1 vs CIO expecting 3.3.3), using `v1.28-latest` instead of pinning to an exact Istio z-stream avoids unnecessary failures. I think we should only set `Degraded=True` when the customer is a **y-stream behind, not a z-stream behind**. This could be a follow-up — not blocking this PR.

TLDR: Switching to `v1.y-latest` gives customers managing their own subscriptions more room to work. We should only flag a hard `Degraded=True` when the OSSM minor version is incompatible, not on z-stream differences.

gcs278:
/unhold

I've done a lot of thinking about the impact of this. A few things to consider:

1. The `v1.y-latest` tag being introduced in the first z-stream bumps (#1444) makes the detached subscription state much easier to manage — customers won't break on every z-stream backport, and they can run a compatible minor+1 OSSM version without hitting an exact version mismatch.
2. Opting out of CIO-managed subscription management isn't obvious — customers can remove the `ingress.operator.openshift.io/owned` annotation from the subscription to go self-managed. We should document this as a supported escape hatch.
3. There are two failure modes for unsupported Istio versions: the CRD enum rejects versions that have been fully removed (hard API error), but existing CRs with a previously-valid version survive CRD upgrades due to ratcheting — in that case, the sail-operator reports an EOL condition on the Istio CR status, which CIO currently doesn't surface. This is a consideration for future status improvements.

But overall, it makes sense and is an improvement from the existing behavior, though I do think we should follow up with:
1. Status improvements (catch EOL condition and/or Istio CR Istio Version API rejections and bubble that up to `Degraded=True`)
2. Prometheus alert for greater visibility 
3. Only go `Degraded=True` for less than minor stream differences (not z-streams).

gcs278:
This bug will be resolved via https://github.com/openshift/cluster-ingress-operator/pull/1442 and make this PR obsolete as we are now pursuing removing OLM entirely in 4.21. @rikatz mind if we close?

rikatz:
don't mind at all, thanks Grant!

PR #1415, opened 2026-04-08
[release-4.20] OCPBUGS-82147: Do not take over OSSM subscription

rikatz:

This is a manual backport of https://github.com/openshift/cluster-ingress-operator/pull/1398 because release-4.22/master are already using noOLM approach and cherry-picking the fix would be conflicting


gcs278:
@rikatz mind if we close this since we are starting the no olm backport for 4.20 now?

rikatz:
not at all

PR #1426, opened 2026-04-29
OCPBUGS-63219: Support NLB protocol to configure proxy protocol and client IP preservation

gcs278:
## Summary

AWS NLBs have `preserve_client_ip.enabled=true` by default on their target groups. This causes hairpin connection failures on internal NLBs: when a pod sends traffic through the NLB and it routes back to the same node, the return packet's destination matches the pod's own IP, breaking the connection. This impacts ROSA private deployments and any internal NLB setup where the client pod and router pod share a node.

This PR adds a `protocol` field to `AWSNetworkLoadBalancerParameters` with two values:

- **`PROXY`** — disables native client IP preservation and enables PROXY protocol v2 (`preserve_client_ip.enabled=false,proxy_protocol_v2.enabled=true`). HAProxy parses PROXY protocol headers to obtain the original client IP. Hairpin connections work.
- **`TCP`** — uses AWS's native client IP preservation (`preserve_client_ip.enabled=true`). This is the current behavior. Hairpin connections may fail on internal NLBs.

The field reuses the same name and type already used by four other endpoint publishing strategies (HostNetwork, NodePort, Private, IBM), providing a consistent API surface: regardless of strategy or platform, `protocol: PROXY` means "enable PROXY protocol on the router and configure the infrastructure to match."

## What this PR does

- **Service annotation**: When `PROXY`, sets `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=false,proxy_protocol_v2.enabled=true` on the NLB Service. When `TCP`, sets `preserve_client_ip.enabled=true,proxy_protocol_v2.enabled=false`. When protocol is empty (pre-existing NLB), no annotation is set — the operator does not manage the annotation for these ICs.
- **Router Deployment**: When `PROXY`, sets `ROUTER_USE_PROXY_PROTOCOL=true` on the router so HAProxy parses PROXY protocol headers. The existing `IsProxyProtocolNeeded` function is updated to return true for NLBs with PROXY protocol.
- **Proxy protocol annotation separation**: The CLB proxy protocol annotation (`aws-load-balancer-proxy-protocol: "*"`) is now only set for CLBs. NLBs use the target-group-attributes annotation instead. A switch statement in `desiredLoadBalancerService` routes to the right annotation based on effective LB type and protocol.
- **LB type transition safety**: Uses `getEffectiveAWSLoadBalancerType` to read the LB type from the current service annotation (not status) during pending CLB↔NLB transitions, preventing annotation stomping on the still-live service.
- **Status sync**: `IngressStatusesEqual` is updated to compare protocol, and `updatePublishingStrategy` syncs spec changes to status. It also handles two additional scenarios:
  - **Cleared spec**: When a user removes a previously-set protocol from spec, the operator re-applies the PROXY default (status being non-empty proves the field was previously managed).
  - **CLB→NLB transition**: When the status LB type transitions to NLB and no protocol is set, the operator defaults to PROXY (detected via `previousStatusLBType`). Pre-existing NLBs on upgrade are excluded because their status type was already NLB (`isPreExistingNLB`).
- **Defaulting**: `setDefaultProviderParameters` defaults new NLB IngressControllers to `PROXY`, gated by `isNewIngressController` (`ic.Status.EndpointPublishingStrategy == nil`). This is more reliable than `!alreadyAdmitted` because `alreadyAdmitted` can flip on re-admission after a validation failure.
- **Auto-delete safety**: When the `auto-delete-load-balancer` annotation is set, the desired service and proxy protocol determination are computed from status rather than the current service. This prevents NLB-specific annotations from being applied to a new CLB service during LB type transitions, which the CCM would reject.
- **Hairpin risk alert**: Emits an informational `NLBHairpinRisk` alert for existing internal NLB IngressControllers that have protocol unset, nudging admins to set PROXY (fix hairpin) or TCP (acknowledge current behavior). The alert description includes an `oc edit` command with the IC name populated from `{{ $labels.name }}`.

## Upgrade compatibility

- **Existing IngressControllers are not modified on upgrade.** The `isNewIngressController` check (status not yet initialized) ensures the PROXY default is only applied to brand new IngressControllers. Re-admitted ICs are not affected because their status is already initialized.
- **Empty protocol in status means "not managed."** When an existing NLB has no protocol set (the pre-upgrade state), `IsProxyProtocolNeeded` returns false, the desired service gets no target-group-attributes annotation, and the one-directional reconciliation in `loadBalancerServiceChanged` only adds or updates the annotation — never removes it. Any user-set annotation (e.g., the KCS hairpin workaround) is left untouched.
- **CLB→NLB transitions get PROXY defaulted.** When an already-admitted IC changes from CLB to NLB without explicitly setting the protocol, `updatePublishingStrategy` detects the type transition via `previousStatusLBType` and defaults to PROXY. This does not fire on upgrade because the status type was already NLB.
- **Mutability**: The field is mutable and applied in-place — changing it does not require deleting or recreating the load balancer Service. The CCM supports updating NLB target group attributes in-place. There is a brief window during transition where connection failures may occur as the NLB attribute change and router rollout happen independently.

## Design

- [Design: NLB protocol for client IP preservation](https://docs.google.com/document/d/14UVY8U-exch30-tYloX_-DMhM7hJX8226VA2q7bTJeo/edit?tab=t.0)

## Dependencies

- openshift/api#2661 — adds the `protocol` field to the API
- openshift/cloud-provider-aws#112 — adds CCM support for `service.beta.kubernetes.io/aws-load-balancer-target-group-attributes` annotation


go.mod
```diff
@@ -225,3 +225,5 @@ replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.5
 // Use OSSM 3.3.1 for Sail Library integration.
 // Replace needed because the fork keeps upstream module name for compatibility.
 replace github.com/istio-ecosystem/sail-operator => github.com/openshift-service-mesh/sail-operator v0.0.0-20260327145107-d4059b965c68
+
+replace github.com/openshift/api => github.com/gcs278/api v0.0.0-20260429000454-cff0427099ea
```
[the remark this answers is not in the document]
gcs278: @coderabbit I know - I'm waiting for https://github.com/openshift/api/pull/2661 to merge

test/e2e/nlb_client_ip_preservation_test.go
```diff
@@ -0,0 +1,192 @@
...
+	if err := waitForDeploymentEnvVar(t, kclient, deployment, 1*time.Minute, "ROUTER_USE_PROXY_PROTOCOL", "true"); err != nil {
+		t.Fatalf("expected ROUTER_USE_PROXY_PROTOCOL=true on router deployment: %v", err)
+	}
+}
+
+// waitForIngressControllerClientIPPreservationMode waits for the
+// IngressController status to have the expected clientIPPreservationMode.
+func waitForIngressControllerClientIPPreservationMode(t *testing.T, name types.NamespacedName, expected operatorv1.ClientIPPreservationMode) error {
+	t.Helper()
+	return wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 2*time.Minute, false, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
gcs278: Done

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -454,8 +460,15 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
 		switch platform.Type {
 		case configv1.AWSPlatformType:
 			service.Annotations[awsLBHealthCheckIntervalAnnotation] = awsLBHealthCheckIntervalDefault
+			effectiveLBType := getEffectiveAWSLoadBalancerType(ci, currentService)
 			if proxyNeeded {
-				service.Annotations[awsLBProxyProtocolAnnotation] = "*"
+				if effectiveLBType == operatorv1.AWSNetworkLoadBalancer {
+					service.Annotations[awsLBTargetGroupAttributesAnnotation] = "preserve_client_ip.enabled=false,proxy_protocol_v2.enabled=true"
+				} else {
```
aswinsuryan: nit: Consider defining the target group attribute values as constants — unit tests can reuse them directly.

aswinsuryan:
Had an initial look — the approach looks solid. Left one minor nit.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Good idea to add them as constants - done. However i generally try to avoid referencing constants from a unit test, as it couples the implementation and the test - if someone changes the constant value, the test silently passes instead of catching the regression.

gcs278:
@jcmoraisjr @aswinsuryan ready for review when you get a chance.

test/e2e/operator_test.go
```diff
@@ -1678,6 +1678,301 @@ func TestAWSLBTypeDefaulting(t *testing.T) {
...
+				// The NLB type annotation should NOT be on the service yet
+				// (it's still a CLB).
+				waitForLBAnnotation(t, ic, ingresscontroller.AWSLBTypeAnnotation, false, "")
+			},
+			verifyAfterTransition: func(t *testing.T, ic *operatorv1.IngressController) {
+				t.Helper()
+				// After transition to NLB with defaulted PROXY protocol,
+				// the target-group-attributes annotation should be set and
+				// the CLB proxy annotation should be absent.
+				waitForLBAnnotation(t, ic, awsLBTargetGroupAttributesAnnotation, true, "preserve_client_ip.enabled=false,proxy_protocol_v2.enabled=true")
```
aswinsuryan, line 1723: For CLB→NLB with no explicit protocol in spec, status protocol stays empty and no target-group-attributes annotation is set., but e2e expects  PROXY annotation after transition. Should it be 
`waitForLBAnnotation(t, ic, awsLBTargetGroupAttributesAnnotation, false, "")` here ?

test/e2e/operator_test.go
```diff
@@ -1678,6 +1678,301 @@ func TestAWSLBTypeDefaulting(t *testing.T) {
...
+				aws.NetworkLoadBalancerParameters = &operatorv1.AWSNetworkLoadBalancerParameters{
+					Protocol: operatorv1.NLBProtocolTCP,
+				}
+			},
+			targetLBType: operatorv1.AWSClassicLoadBalancer,
+			verifyWhilePending: func(t *testing.T, ic *operatorv1.IngressController) {
+				t.Helper()
+				// The NLB target-group-attributes annotation should still have TCP
+				// values on the old NLB service while the type change is pending.
+				waitForLBAnnotation(t, ic, awsLBTargetGroupAttributesAnnotation, true, "preserve_client_ip.enabled=true")
```
aswinsuryan: waitForLBAnnotation requires an exact match. This expects "preserve_client_ip.enabled=true", but the operator sets "preserve_client_ip.enabled=true,proxy_protocol_v2.enabled=false".

gcs278: ah yea good catch, i missed this

test/e2e/operator_test.go
gcs278: Ah yes, good catch! This was actually a bug. We weren't defaulting to proxy on a CLB to NLB switch - that's a mild bug. I've added a small snippet of code to default on the transition, so this e2e stays the same.

gcs278:
Whoops, need to debug this [failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1426/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator/2078269520924905472)
/hold

```
=== NAME  TestAll/parallel/TestAWSNLBUpgradeAnnotationPreservation
    nlb_client_ip_preservation_test.go:293: failed to verify connectivity with workload with reqURL http://a4754258d58ec41a1922b75cd55f933e-deccd0838a836f13.elb.us-east-1.amazonaws.com using external client: timed out waiting for the condition
```

gcs278:
no related failures:
/retest

gcs278:
no failures related
/retest

gcs278:
EnsureNoCrashingPods - I've seen this many times already on other PRs
/test hypershift-e2e-aks

pkg/operator/controller/ingress/metrics.go
```diff
@@ -31,10 +31,16 @@ var (
 		Help: "Report the number of active NLBs on AWS clusters.",
 	}, []string{"name"})
 
+	nlbHairpinRisk = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "ingress_controller_aws_nlb_hairpin_risk",
+		Help: "Reports whether an IngressController using an internal AWS NLB has no explicit protocol setting and may be affected by hairpin connection failures. 0 is no risk, 1 is at risk.",
+	}, []string{"name"})
```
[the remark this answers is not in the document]
gcs278: @coderabbit  - it works, {{ $labels.namespace }} is injected by Prometheus.

gcs278:
@codrabbit have you done a full review?

gcs278:
Not related
/retest

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -454,7 +468,18 @@ func desiredLoadBalancerService(ci *operatorv1.IngressController, deploymentRef
...
+			// loadBalancerServiceChanged does not stomp any user-set annotation.
+			nlbProtocol := getAWSNLBProtocol(ci.Status.EndpointPublishingStrategy)
+			switch {
+			case effectiveLBType == operatorv1.AWSNetworkLoadBalancer && nlbProtocol == operatorv1.NLBProtocolProxy:
+				service.Annotations[awsLBTargetGroupAttributesAnnotation] = awsNLBTargetGroupAttributesProxy
+			case effectiveLBType == operatorv1.AWSNetworkLoadBalancer && nlbProtocol == operatorv1.NLBProtocolTCP:
+				service.Annotations[awsLBTargetGroupAttributesAnnotation] = awsNLBTargetGroupAttributesTCP
+			case effectiveLBType == operatorv1.AWSClassicLoadBalancer:
 				service.Annotations[awsLBProxyProtocolAnnotation] = "*"
 			}
```
jcmoraisjr, line 493: Wondering if part of this code can be derived from `proxyNeeded` as a way to deduplicate the logic, since a similar code is already behind it. On a quick exercise I ended up with the code below, but I'm not sure if it makes sense and if it's in fact simpler.

```go
			switch effectiveLBType {
			case operatorv1.AWSNetworkLoadBalancer:
				target := awsNLBTargetGroupAttributesTCP
				if proxyNeeded {
					target = awsNLBTargetGroupAttributesProxy
				}
				service.Annotations[awsLBTargetGroupAttributesAnnotation] = target
			case operatorv1.AWSClassicLoadBalancer:
				service.Annotations[awsLBProxyProtocolAnnotation] = "*"
			}
```

pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -1416,6 +1469,20 @@ func getAWSNLBSubnets(eps *operatorv1.EndpointPublishingStrategy) *operatorv1.AW
 	return nil
 }
 
+// getAWSNLBProtocol returns the protocol for the NLB reported in the
+// endpoint publishing strategy.
+func getAWSNLBProtocol(eps *operatorv1.EndpointPublishingStrategy) operatorv1.NLBProtocol {
```
jcmoraisjr, line 1483: Does `getAWSNLBProtocol()` deserves a `getEffectiveAWSNLBProtocol()` for a service based result, just like `getEffectiveAWSLoadBalancerType()` already does?

test/e2e/nlb_client_ip_preservation_test.go
```diff
@@ -0,0 +1,471 @@
...
+			AWS: &operatorv1.AWSLoadBalancerParameters{
+				Type: operatorv1.AWSNetworkLoadBalancer,
+				NetworkLoadBalancerParameters: &operatorv1.AWSNetworkLoadBalancerParameters{
+					Protocol: operatorv1.NLBProtocolProxy,
+				},
+			},
+		},
+	}
+
+	if err := createWithRetryOnError(t, context.TODO(), ic, 2*time.Minute); err != nil {
```
jcmoraisjr: Does `t.Context()` fit here? It is cancelled in the end of the test, e.g. when it times out.

test/e2e/util_test.go
```diff
@@ -660,6 +660,32 @@ func updateIngressConfigStatusWithRetryOnConflict(t *testing.T, name types.Names
...
+		mutateStatusFn(&ic.Status)
+		if err := kclient.Status().Update(ctx, &ic); err != nil {
+			if errors.IsConflict(err) {
+				t.Logf("conflict when updating ingress controller status %v: %v, retrying...", name, err)
+				return false, nil
+			}
+			return false, err
+		}
+		return true, nil
+	})
```
jcmoraisjr, line 686: `retry.RetryOnConflict()` uses to be the pragmatic way to retry conflicting apiserver updates. I can see the same polling pattern used elsewhere, so just wondering if it adds value compared with `RetryOnConflict()`.

jcmoraisjr (commented):
The whole update looks great. Nice handling of the state, good test coverage, awesome PR description. Consider the comments below just as shared thoughts and some sort of curiosity. I can approve/lgtm it the way it is, just after some feedback.

jcmoraisjr:
Let me approve it since I should be ooo until tomorrow eod. It's on hold so it won't get merged.

/approve
/lgtm

melvinjoseph86:
i am marking this as verified for now and will test the same after merge.
/verified later @mjoseph

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: ah yea interesting point. I debated this style myself too.

I can do a modified version of this, but we must have a `if len(nlbProtocol) > 0 {` to protect the upgrade case (no status == no annotation change).

I think it's cleaner. Done.

pkg/operator/controller/ingress/load_balancer_service.go
gcs278: Good question. `getEffectiveAWSLoadBalancerType()` exists because LB type is a delete-and-recreate-service-style configuration: status and service can drift during a pending transition, and we need the service annotation to know the actual LB type of the still-live service.

NLB protocol changes are applied in-place by the CCM, so status and service are always in sync. There's no window where we'd need to read the protocol from the service.

test/e2e/nlb_client_ip_preservation_test.go
gcs278: Good point - a lot of our existing e2e use `context.TODO()` but I can switch over new uses to `t.Context()`.

test/e2e/util_test.go
gcs278: Yea good observation. `retry.RetryOnConflict` would work for the conflict case. Though, this helper follows the same pattern as `updateIngressControllerWithRetryOnConflict` and `updateIngressConfigStatusWithRetryOnConflict` in the same file for consistency.

The polling approach also retries transient Get errors and provides a configurable timeout, which is useful in CI environments where API server hiccups can occur.

gcs278 (commented):
thanks for the quick review @jcmoraisjr

jcmoraisjr:
unrelated flakiness

/retest-required

jcmoraisjr:
Thanks for the comments @gcs278! The changes looks solid.

/lgtm

gcs278:
@jcmoraisjr Sorry for the extra review round: the proxy annotation switch refactor you suggested actually exposed a bug in the auto-delete workflow.

When auto-delete is set and the LB type changes from NLB→CLB, `desiredLoadBalancerService` was computed using the old NLB service. This caused NLB-specific annotations (target-group-attributes) to bleed onto the new CLB service. The same staleness affected the router's proxy protocol determination, but the difference is that a wrong env var on the router self-heals on the next reconcile — the CCM does not:
```
"message": "The service-controller component is reporting SyncLoadBalancerFailed events like: 
Error syncing load balancer: failed to ensure load balancer: target group attributes annotation 
is only supported for NLB"
```

The fix: when auto-delete is set, pass nil for `currentService` so the desired service and proxy protocol determination are computed from status (the intended state) rather than from the soon-to-be-deleted service.

Also reduced `TestAWSLBTypeTransitionSafety` from 6 to 4 test cases — the dropped cases were testing slightly different flavors of the same code path, and creating that many NLBs in parallel was adding unnecessary load that concerns me regarding our E2E quotas.

jcmoraisjr:
I don't know whether I'm glad to help the finding or sorry to have added the noise :slightly_smiling_face: 

/lgtm

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-upgrade

gcs278:
install failure
/test e2e-gcp-operator

gcs278:
`NAME  TestAll/parallel/TestContainerLoggingMinLength
    operator_test.go:2887: failed to create network policy syslog-netpol: networkpolicies.networking.k8s.io "syslog-netpol" already exists` known bug:
/test e2e-aws-operator

gcs278:
this job is known to be flaky today:
`        --- FAIL: TestAutoscaling/ValidateHostedCluster/EnsureNoCrashingPods (0.30s)`

/test hypershfit-e2e-aks

gcs278:
Copying @melvinjoseph86's verified:
/verified later @mjoseph

gcs278:
`  * could not run steps: step e2e-gcp-operator failed: "e2e-gcp-operator" post steps failed: "e2e-gcp-operator" pod "e2e-gcp-operator-gather-gcp-console" failed: could not watch pod: the pod ci-op-j897g344/e2e-gcp-operator-gather-gcp-console failed after 1m2s (failed containers: test): ContainerFailed one or more containers exited`

/test e2e-gcp-operator

gcs278:
`{  failed to acquire lease for "hypershift-aks-quota-slice": resources not found}`:
/test hypershift-e2e-aks

gcs278:
I think we need to proceed merging with this one as is. It's not a perfect solution, but it is a pragmatic one, this fix is in high demand. In the future, I may try to add some sort of one-off migration logic so that we don't have these subtle status behaviors.

I also would like to add E2E tests for the next metric/alert, but since we are out of time, we can merge as is, and I'll provide a follow up.

/unhold

PR #1427, opened 2026-04-29
OCPBUGS-84551: isFIPSEnabled: Check FIPS_ENABLED env. var. first

Miciah:
When detecting whether the cluster has FIPS enabled, check the FIPS_ENABLED environment variable before checking procfs.  HyperShift sets this environment variable so that the operator can correctly detect the FIPS status of hosted clusters, which can be different from the FIPS status of the management cluster where the operator runs.

--- 

https://github.com/openshift/hypershift/pull/8375 sets `FIPS_ENABLED`, so the complete fix depends on both that PR and this one.  However, because this PR falls back to procfs, it doesn't strictly depend on the hypershift PR.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -40,16 +40,47 @@ import (
 	securityv1 "github.com/openshift/api/security/v1"
 )
 
-// isFIPSEnabled reports whether the current node is operating in FIPS
-// mode by checking /proc/sys/crypto/fips_enabled. This is a
-// package-level variable so tests can override it to simulate FIPS mode.
-var isFIPSEnabled = func() bool {
-	data, err := os.ReadFile("/proc/sys/crypto/fips_enabled")
-	if err != nil {
-		return false
+// isFIPSEnabled reports whether the cluster has FIPS enabled.
+var isFIPSEnabled = detectFIPS()
+
+// lookupEnv is an alias for os.LookupEnv that can be overridden in unit tests.
+var lookupEnv = os.LookupEnv
+
+// readFile is an alias for os.ReadFile that can be overridden in unit tests.
+var readFile = os.ReadFile
+
+// detectFIPS reports whether the cluster is operating in FIPS
+// mode by checking the FIPS_ENABLED environment variable if it set or
+// the /proc/sys/crypto/fips_enabled file otherwise. This is a package-
+// level variable so tests can override it to simulate FIPS mode.
+func detectFIPS() bool {
+	if v, ok := lookupEnv("FIPS_ENABLED"); ok {
+		if result, err := strconv.ParseBool(v); err != nil {
+			log.Error(err, "Failed to parse FIPS_ENABLED environment variable; falling back to procfs")
+		} else {
+			log.Info("Found FIPS_ENABLED environment variable", "value", v, "result", result)
+
+			return result
+		}
 	}
-	return len(data) > 0 && data[0] == '1'
-}()
+
+	result := false
+	if data, err := readFile("/proc/sys/crypto/fips_enabled"); err != nil {
+		log.Error(err, "Failed to read /proc/sys/crypto/fips_enabled; assuming FIPS is not enabled", "result", result)
+
+		return result
+	} else if len(data) == 0 {
+		log.Error(nil, "Got empty /proc/sys/crypto/fips_enabled; assuming FIPS is not enabled", "result", result)
+
+		return result
+	} else {
+		result = data[0] == '1'
+
+		log.Info("Read /proc/sys/crypto/fips_enabled", "data", data, "result", result)
+
+		return result
+	}
+}
 
 const (
 	WildcardRouteAdmissionPolicy = "ROUTER_ALLOW_WILDCARD_ROUTES"
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -2,6 +2,7 @@ package ingress
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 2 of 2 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -41,14 +41,36 @@ import (
...
-	if err != nil {
-		return false
+	if v, ok := os.LookupEnv("FIPS_ENABLED"); ok {
+		if result, err := strconv.ParseBool(v); err != nil {
+			log.Error(err, "Failed to parse FIPS_ENABLED environment variable; falling back to profs")
+		} else {
+			log.Info("Found FIPS_ENABLED environment variable", "value", v, "result", result)
+
+			return result
+		}
```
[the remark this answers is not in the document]
Miciah: Line 50 logs the error, so it isn't a silent failure.  The alternative to logging and falling back to procfs is for the operator to crashloop.  That certainly would be more visible, but I think it is preferable for the operator to start so that the cluster has a chance of having functional ingress.

Miciah:
https://github.com/openshift/cluster-ingress-operator/compare/33700ce200d0d75eb2da1078aa62f7f15300e2f5..fd75881bdbdeca4180f6170641a970da5d217afc moves the FIPS-detection logic to a named function and adds a unit test for it.

Miciah:
[e2e-aws-operator-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1427/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview/2049649320675774464) failed because of [OCPBUGS-81442 e2e-aws-operator-techpreview failing TestUnsupportedConfigOverride/dynamic-config-manager after PR #1385](https://redhat.atlassian.net/browse/OCPBUGS-81442).

[e2e-aws-ovn-upgrade](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1427/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-upgrade/2049649321023901696) failed because these tests failed:

- `[Monitor:node-lifecycle][sig-node] node-lifecycle detects unexpected not ready node`
- `[Monitor:legacy-test-framework-invariants-pathological][sig-arch] events should not repeat pathologically for ns/openshift-network-diagnostics`

@coderabbitai, please analyze these test failures in e2e-aws-ovn-upgrade and tell me whether there are any related Jira tickets.

Miciah:
@coderabbitai, to compute the failure rates, are you reading directly from Sippy?

Miciah:
@coderabbitai, please provide a Mermaid flowchart showing your reasoning that connects the "[Monitor:node-lifecycle][sig-node] node-lifecycle detects unexpected not ready node" failure to OCPBUGS-55713 and OCPBUGS-44757.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -40,16 +40,47 @@ import (
...
+var readFile = os.ReadFile
+
+// detectFIPS reports whether the cluster is operating in FIPS
+// mode by checking the FIPS_ENABLED environment variable if it set or
+// the /proc/sys/crypto/fips_enabled file otherwise. This is a package-
+// level variable so tests can override it to simulate FIPS mode.
+func detectFIPS() bool {
+	if v, ok := lookupEnv("FIPS_ENABLED"); ok {
+		if result, err := strconv.ParseBool(v); err != nil {
+			log.Error(err, "Failed to parse FIPS_ENABLED environment variable; falling back to profs")
```
bentito:
```suggestion
			log.Error(err, "Failed to parse FIPS_ENABLED environment variable; falling back to procfs")
```

bentito:
/lgtm
just had the one typo `procfs` in the log line. Other than that, this seems like clean way to make the FIPS determination ~once~ now that the HCP PR with the env var has merged.

pkg/operator/controller/ingress/deployment.go
Miciah: Fixed in https://github.com/openshift/cluster-ingress-operator/compare/fd75881bdbdeca4180f6170641a970da5d217afc..042145624eddf76bb21c0e273b99c96f66bf7b9e.  I'm surprised @coderabbitai didn't catch this!  Meantime, I'll oil my 'c' key.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -40,16 +40,47 @@ import (
...
+		} else {
+			log.Info("Found FIPS_ENABLED environment variable", "value", v, "result", result)
+
+			return result
+		}
+	}
+
+	result := false
+	if data, err := readFile("/proc/sys/crypto/fips_enabled"); err != nil {
+		log.Error(err, "Failed to read /proc/sys/crypto/fips_enabled; assuming FIPS is not enabled", "result", result)
```
candita, line 69: It would be great to see the error that was logged.  If we encounter another environment where FIPS is enabled but neither the env variable or procfs indicates that, knowing the error message might help.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -40,16 +40,47 @@ import (
...
+
+		return result
+	} else if len(data) == 0 {
+		log.Error(nil, "Got empty /proc/sys/crypto/fips_enabled; assuming FIPS is not enabled", "result", result)
+
+		return result
+	} else {
+		result = data[0] == '1'
```
candita, line 77: I know we don't expect procfs to contain anything other than 0 or 1, but what if it did?  Should we give a different result if it the value was `001` than if it was `25` ?

candita (approved):
I had 2 comments, both of which can be handled in a followup if you agree they are useful.

pkg/operator/controller/ingress/deployment.go
Miciah: It is hard to anticipate the conditions under which we would get an error.  I imagine the most likely errors would be "no such file or directory" or "permission denied" if the pod somehow got scheduled on a non-Linux node or if the kubelet didn't initialize the filesystem correctly.

pkg/operator/controller/ingress/deployment.go
Miciah: I just kept the old logic for determining the result in that case, and I did add a log message that includes the data, but we could log more explicitly about unexpected values.  I'd prefer to leave that for a followup if that's all right.

rhamini3:
Confirmed that FIPS env var exists before checking procfs and [unit tests](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1427/pull-ci-openshift-cluster-ingress-operator-master-unit/2049934714046779392) marking bug as verified

/verified by @rhamini3

pkg/operator/controller/ingress/deployment.go
```diff
@@ -40,16 +40,47 @@ import (
...
+
+		log.Info("Read /proc/sys/crypto/fips_enabled", "data", data, "result", result)
```
Miciah, line 79: It would improve readability to convert the `[]byte` value to `string` for the logger:
```suggestion
		log.Info("Read /proc/sys/crypto/fips_enabled", "data", string(data), "result", result)
```
Otherwise the `[]byte` value gets logged as base64-encoded bytes.  For example, the string value `"0\n"` gets logged as "MAo=" and `"1\n"` gets logged as "MQo=".

PR #1434, opened 2026-05-06
OCPBUGS-85679: Improve e2e tests reliability

rikatz:
**What**

  Fix multiple correctness bugs and resource leak patterns in e2e tests that cause flakes, hangs, and test infrastructure pollution.

  **Why**

  A full codebase review identified several categories of e2e test defects:
  - Inverted error checks that silently swallow real failures or retry on the wrong conditions, masking bugs in the operator.
  - An unbounded wait.PollInfinite call that can hang a test run forever if pod deletion stalls.
  - A data race on a shared mutable counter (podCount) accessed by parallel tests without synchronization.
  - 75 instances of t.Fatalf called inside defer func() blocks, which invokes runtime.Goexit() and prevents subsequent defers from executing — leaking pods, services, routes, and namespaces in the test cluster.

  **Changes**

  - Fix inverted error check in deleteWithRetryOnError (util_test.go): Changed !errors.IsAlreadyExists(err) to !errors.IsNotFound(err) so that delete retries correctly ignore "not found" (already deleted) rather than "already
   exists" (nonsensical for delete).
  - Fix inverted error check in verifyInternalIngressController (util_test.go): Added missing negation ! before errors.IsNotFound(err) so cleanup delete correctly tolerates already-gone resources.
  - Replace wait.PollInfinite with bounded timeout (util_test.go): Switched to wait.PollUntilContextTimeout with a 2-minute deadline to prevent infinite hangs when pod deletion stalls.
  - Fix data race on shared podCount (set_delete_test.go): Changed var podCount int with non-atomic podCount++ to atomic.Int32 with podCount.Add(1), eliminating the race between parallel
  TestSetIngressControllerResponseHeaders and TestSetRouteResponseHeaders.
  - Replace t.Fatalf with t.Errorf in all defer cleanup blocks (10 files, 75 locations): t.Fatalf in a deferred function calls runtime.Goexit(), which aborts the goroutine and skips all remaining defers — leaking test
  resources. Switched to t.Errorf which logs the failure but allows cleanup to complete.

### DIFF
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -277,7 +277,7 @@ func TestCanaryWithMTLS(t *testing.T) {
 		t.Fatalf("Failed to create configmap %s: %v", clientCAConfigmap.Name, err)
 	}
 
-	t.Cleanup(func() { assertDeleted(t, kclient, clientCAConfigmap) })
+	t.Cleanup(func() { deleteWithRetryOnError(t, context.Background(), clientCAConfigmap, 2*time.Minute) })
 
 	defaultIC := &operatorv1.IngressController{}
 	if err := kclient.Get(context.TODO(), defaultName, defaultIC); err != nil {
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -83,7 +83,7 @@ func TestCanaryRoute(t *testing.T) {
 
 	clientPodName := types.NamespacedName{Name: "canary-route-check", Namespace: canaryRoute.Namespace}
 	clientNetworkPolicy := buildTestPodNetworkPolicy(clientPodName)
-	if err := kclient.Create(context.TODO(), clientNetworkPolicy); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientNetworkPolicy, 2*time.Minute); err != nil {
 		t.Fatalf("Failed to create network policy %s/%s: %v", clientNetworkPolicy.Namespace, clientNetworkPolicy.Name, err)
 	}
 	t.Cleanup(func() {
@@ -97,7 +97,7 @@ func TestCanaryRoute(t *testing.T) {
 
 	image := deployment.Spec.Template.Spec.Containers[0].Image
 	clientPod := buildCanaryCurlPod(clientPodName.Name, clientPodName.Namespace, image, canaryRouteHost)
-	if err := kclient.Create(context.TODO(), clientPod); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientPod, 2*time.Minute); err != nil {
 		t.Fatalf("Failed to create pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
 	}
 	t.Cleanup(func() {
@@ -273,7 +273,7 @@ func TestCanaryWithMTLS(t *testing.T) {
 		},
 	}
 
-	if err := kclient.Create(context.TODO(), clientCAConfigmap); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientCAConfigmap, 2*time.Minute); err != nil {
 		t.Fatalf("Failed to create configmap %s: %v", clientCAConfigmap.Name, err)
 	}
 
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -303,7 +303,7 @@ func TestCanaryWithMTLS(t *testing.T) {
 		if err := updateIngressControllerWithRetryOnConflict(t, defaultName, timeout, func(defaultIC *operatorv1.IngressController) {
 			defaultIC.Spec.ClientTLS = *originalClientTLS
 		}); err != nil {
-			t.Fatalf("Failed to restore default ingress configuration: %v", err)
+			t.Errorf("Failed to restore default ingress configuration: %v", err)
 		}
 		t.Log("Restored clientTLS config for default ingresscontroller.")
 	})
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -83,7 +83,7 @@ func TestCanaryRoute(t *testing.T) {
 
 	clientPodName := types.NamespacedName{Name: "canary-route-check", Namespace: canaryRoute.Namespace}
 	clientNetworkPolicy := buildTestPodNetworkPolicy(clientPodName)
-	if err := createWithRetryOnError(t, context.Background(), clientNetworkPolicy, 2*time.Minute); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientNetworkPolicy, DefaultRetryTimeout); err != nil {
 		t.Fatalf("Failed to create network policy %s/%s: %v", clientNetworkPolicy.Namespace, clientNetworkPolicy.Name, err)
 	}
 	t.Cleanup(func() {
@@ -97,7 +97,7 @@ func TestCanaryRoute(t *testing.T) {
 
 	image := deployment.Spec.Template.Spec.Containers[0].Image
 	clientPod := buildCanaryCurlPod(clientPodName.Name, clientPodName.Namespace, image, canaryRouteHost)
-	if err := createWithRetryOnError(t, context.Background(), clientPod, 2*time.Minute); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientPod, DefaultRetryTimeout); err != nil {
 		t.Fatalf("Failed to create pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
 	}
 	t.Cleanup(func() {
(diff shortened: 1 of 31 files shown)


test/e2e/configurable_route_test.go
```diff
@@ -36,7 +36,7 @@ func TestConfigurableRouteRBAC(t *testing.T) {
 	}
 	defer func() {
 		if err := kclient.Get(context.TODO(), types.NamespacedName{Namespace: "", Name: "cluster"}, ingress); err != nil {
-			t.Fatalf("failed to get ingress resource: %v", err)
+			t.Errorf("failed to get ingress resource: %v", err)
```
Miciah: If it is safe to continue even if `Get` failed, do we need the `Get` at all?

Please consider using `updateIngressConfigSpecWithRetryOnConflict` instead.  If we can ultimately delete `eventuallyUpdateIngressSpec`, so much the better.

test/e2e/set_delete_test.go
```diff
@@ -26,9 +26,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/client-go/kubernetes"
+	"sync/atomic"
```
Miciah: `"sync/atomic"` should be grouped with the other standard package imports.

test/e2e/util_test.go
```diff
@@ -1004,12 +1004,14 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
 				t.Fatalf("failed to delete pod %q: %v", clientPodName, err)
 			}
-			// Wait for deletion to prevent a race condition. Use PollInfinite since we are already in a Poll.
-			wait.PollInfinite(5*time.Second, func() (bool, error) {
-				err = kclient.Get(context.TODO(), clientPodName, clientPod)
+			// Wait for deletion to prevent a race condition.
+			deleteCtx, deleteCancel := context.WithTimeout(context.Background(), 2*time.Minute)
+			defer deleteCancel()
+			wait.PollUntilContextTimeout(deleteCtx, 5*time.Second, 2*time.Minute, true, func(ctx context.Context) (bool, error) {
+				err = kclient.Get(ctx, clientPodName, clientPod)
```
Miciah, line 1047: This change could use some explanation in the commit message and possibly a code comment.  

Also, while you are changing the inner polling loop to use a context, would it make sense to change the outer polling loop to use a context as well, and have the inner polling loop's context derive from the outer loop's?  

@coderabbitai, please explain this change.

test/e2e/forwarded_header_policy_test.go
```diff
@@ -62,7 +62,7 @@ func testRouteHeaders(t *testing.T, image string, route *routev1.Route, address
 	defer func() {
 		if err := kclient.Delete(context.TODO(), clientPod); err != nil {
 			if !errors.IsNotFound(err) {
-				t.Fatalf("failed to delete pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
+				t.Errorf("failed to delete pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
```
Miciah: I'm tempted to suggest replacing every `defer`-`Delete`-`t.Error` with `deleteWithRetryOnError`.  Also, you could make `deleteWithRetryOnError` call `t.Error` itself to save the caller the trouble.  Last, we use `t.Cleanup` instead of `defer` in new code, so all together, we could change this:
```go
	defer func() {
		if err := kclient.Delete(context.TODO(), foo); err != nil {
			if !errors.IsNotFound(err) {
				t.Errorf("failed to delete foo %s/%s: %v", foo.Namespace, foo.Name, err)
			}
		}
	}()
```
into this:
```go
	t.Cleanup(func() { deleteWithRetryOnError(t, ctx, foo, 2*time.Minute) })
```
But I get if you want to keep this PR simple.  We can do further cleanups in a followup.

test/e2e/util_test.go
Miciah: @coderabbitai, the original comment, "Use PollInfinite since we are already in a Poll.", suggests that the outer polling loop's timing out would cause both the outer polling loop and the inner polling loop to be terminated.  Is it not the case that the outer `PollImmediate`'s timeout will terminate the inner `PollInfinite`?

Miciah: Thanks for explaining, @coderabbitai!  But one thing—

> Having the inner context derive from the outer loop's context (as I suggested earlier)

Heyyy, *I* made that suggestion and then asked *you* about the change.  * grin *.

test/e2e/set_delete_test.go
rikatz: fixed

test/e2e/forwarded_header_policy_test.go
rikatz: no no, I agree this is a good cleanup

test/e2e/configurable_route_test.go
rikatz: I don't think this means it is safe, I think this means that it will stop the current text, not do a fatal exit

rikatz: I did some changes to improve the cleanup resiliency, while I tried to avoid for now touching other parts of the logic of the code

rikatz:
@coderabbitai I want you do to a full assessment, line by line of my changes and check if something will be breaking, if some critical or major finding requires some addressing. Do an adversarial and detailed review. This PR goal is to fix and stabilize our e2e tests, so let me know if there is some blocker on merging it.

rikatz:
@coderabbitai are the failures related with the changes here?

rikatz:
@coderabbitai on your previous assessment, L1 and L3 seems wrong. L3 seems fixed, and L1 is wrong as I can find multiple usages of the function.  Do again a full assessment, line by line of my changes and check if something will be breaking, if some critical or major finding requires some addressing. Do an adversarial and detailed review. This PR goal is to fix and stabilize our e2e tests, so let me know if there is some blocker on merging it.

rikatz:
@coderabbitai before reviewing, clean your cache. You keep telling me that "eventuallyUpdateIngressSpec" is not used anymore but I can see it on configurable_route_test.go.  Do again a full assessment, line by line of my changes and check if something will be breaking, if some critical or major finding requires some addressing. Do an adversarial and detailed review. You don't need to make me happy, you need to be sure that I am sending a consistent, stable and improving PR. This PR goal is to fix and stabilize our e2e tests, so let me know if there is some blocker on merging it.

rikatz:
@coderabbitai on R1 your findings on operator_test.go seems very wrong. The blocks are out of t.Cleanup, you seem to be reading it incorrectly

rikatz:
@coderabbitai all good now, right?

rikatz:
@coderabbitai the test e2e-aws-operator seems to be permafailing after this PR, can you verify if it broke something?

rikatz:
@coderabbitai you were wrong. The failure was caused by my PR, specifically on:

```
 Root cause: verifyExternalIngressController and verifyInternalIngressController are called multiple times within the same test (e.g., TestAWSEIPAllocationsForNLB and TestAWSLBSubnets both call
  verifyExternalIngressController twice). Our change from defer to t.Cleanup moved cleanup from function-return scope to test-end scope, so the second call found resources still existing from the first call.

  Fix: Replaced bare kclient.Create with createWithRetryOnError in both functions. This helper already handles AlreadyExists gracefully (treats it as success), so:
  - First call: creates resources fresh, registers t.Cleanup
  - Second call: AlreadyExists is treated as success, the existing workload is still running and usable — only the LB configuration changed between calls, not the backend workload
```

rikatz:
@coderabbitai I want you to do a new review, line by line and the surroundings, on an adversarial way. You don't need to cheer to me, but rather find any issue that this PR and all it changes may bring. Try to find problems on this PR, or unintended behavior changes on the tests. Do a deep aggressive review of possible findings, and report all of them with a conclusion if this PR is ready to be merged or not

rikatz:
@coderabbitai I want you to do exactly the same assessment you did before, but based on my last commit that was fixing the findings or adding a proper comment on when functions should be used and let me know if this is better

test/e2e/util_test.go
```diff
@@ -955,21 +990,19 @@ func verifyInternalIngressController(t *testing.T, name types.NamespacedName, ho
...
 
 	var curlPodLogs string
-	err = wait.PollImmediate(10*time.Second, 10*time.Minute, func() (bool, error) {
-		if err := kclient.Get(context.TODO(), clientPodName, clientPod); err != nil {
+
+	// This loop will try, for 10 minutes, to get the client pod used to test the connectivity
+	// and execute the test until it passes.
+	// In case there is an error with the Pod creation, it will enter on another loop
+	// of maximum 5 minutes to delete and wait for this pod to be deleted before continuing
+	err = wait.PollUntilContextTimeout(t.Context(), 10*time.Second, 10*time.Minute, false, func(ctx context.Context) (bool, error) {
```
davidesalerno, line 1004: Nice improvement switching the inner delete wait to a bounded context. One follow-up thought: since this helper now has nested polling, would it be cleaner to derive the inner delete-wait context from the outer retry loop context all the way through, including the log stream/delete/recreate path? That would make deadline/cancellation handling more consistent and easier to reason about.

test/e2e/canary_test.go
```diff
@@ -83,7 +83,7 @@ func TestCanaryRoute(t *testing.T) {
 
 	clientPodName := types.NamespacedName{Name: "canary-route-check", Namespace: canaryRoute.Namespace}
 	clientNetworkPolicy := buildTestPodNetworkPolicy(clientPodName)
-	if err := kclient.Create(context.TODO(), clientNetworkPolicy); err != nil {
+	if err := createWithRetryOnError(t, context.Background(), clientNetworkPolicy, 2*time.Minute); err != nil {
```
davidesalerno: Since we are using 2 minute as a default timeout, I think it could be useful defining a constant and using it so that if we would like in future change it it will be easier and quicker

davidesalerno:
Just a couple of minor comments

test/e2e/util_test.go
rikatz: yes, good call! I forgot to replace the internal context to use the parent one created to the poll function, changing here!

rikatz:
@coderabbitai please verify if the failures on both pipelines were caused by my changes

rhamini3:
since all e2e tests are passing marking as verified

/verified by CI

PR #1300, opened 2025-11-04
OCPBUGS-85115, NE-2203: Add PrometheusRule for Gateway API

rikatz:
This PR adds a PromtheusRule to collect the usage of Gateway on an Openshift cluster, based on the GatewayClass. It reduces the metric to:
* Gateways that are accepted
* Tags by the GatewayClass (openshift/not-openshift)

As the PrometheusRule can be added without any caveat to the cluster (eg.: even if no Gateway is enabled), it is being added as its own rule, but as part of CIO installation.

This should be merged after https://github.com/openshift/cluster-monitoring-operator/pull/2734

PR #1436, opened 2026-05-06
[release-4.20] OCPBUGS-85130: Add PrometheusRule for Gateway API gathering

rikatz:
This is a backport of https://github.com/openshift/cluster-ingress-operator/pull/1300

This PR adds a PromtheusRule to collect the usage of Gateway on an Openshift cluster, based on the GatewayClass. It reduces the metric to:

* Gateways that are accepted
* Tags by the GatewayClass (openshift/not-openshift)

As the PrometheusRule can be added without any caveat to the cluster (eg.: even if no Gateway is enabled), it is being added as its own rule, but as part of CIO installation.

### DIFF
--- a/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
@@ -85,4 +85,30 @@ spec:
             summary: Route owned by an Ingress no longer managed
             description: "This alert fires when there is a Route owned by an unmanaged Ingress."
             message: "Route {{ $labels.namespace }}/{{ $labels.name }} is owned by an unmanaged Ingress."
-
+    - name: gateway-api-telemetry.rules
+      rules:
+        - record: openshift:gateway_api_usage:count
+          expr: |
+            (
+              sum by(gateway_class_type) (
+                label_replace(
+                  (
+                    kube_customresource_gateway_info{programmed="True"}
+                    and on(gateway_class) kube_customresource_gateway_class_info{accepted="True", controller="openshift.io/gateway-controller/v1"}
+                  ),
+                  "gateway_class_type", "openshift", "", ""
+                )
+              )
+            )
+            or
+            (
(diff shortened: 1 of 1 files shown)


gcs278:
Backport is clean and up-to-date. Sounds like we can merge this without https://github.com/openshift/cluster-monitoring-operator/pull/2920, but I'll leave that up to you.

/approve
/lgtm

rikatz:
Thanks @gcs278 

@Miciah we need the backport-risk-assessed label here as well, please!

melvinjoseph86:
verified using the existing automation test case ocp-86110 running on the cluster created using this PR and [PR2920](https://github.com/openshift/cluster-monitoring-operator/pull/2920).
```
passed: (2m54s) 2026-05-07T05:21:34 "[sig-network-edge] Network_Edge Component_Router Author:iamin-ROSA-OSD_CCS-ARO-High-86110-View metrics showing who is using Gateway API"

Writing JUnit report to junit_e2e_20260507-052134.xml

1 pass, 0 skip (2m54s)
```
Screenshot 2026-05-07 at 10 51 31 AM


Hence marking as verified
/verified by @mjoseph

Miciah:
@rikatz, what happens if the metric doesn't exist?

rikatz:
> @rikatz, what happens if the metric doesn't exist?

nothing. From gemini:
```
In Prometheus, if you query a metric that does not exist, it does not return an error. Instead, it returns an empty result set (an empty vector).

For a PrometheusRule resource, this means the rule will technically be "valid" and active, but because the query returns nothing, the rule will never trigger an alert or record any data.
```

I remember also testing it and it worked

Miciah:
Thanks!

This is a low-risk backport.  It is a clean cherry pick to add a Prometheus alerting rule.  As Ricardo says in https://github.com/openshift/cluster-ingress-operator/pull/1436#issuecomment-4398848344, there is no impact if the metric is absent.   

/label backport-risk-assessed

PR #1438, opened 2026-05-06
[release-4.20] OCPBUGS-85155: Fix e2e tests to work on platforms with unmanaged DNS

rikatz:
This is a manual backport of https://github.com/openshift/cluster-ingress-operator/pull/1342

On on-prem platforms we cannot assure that the DNS record will always work. On some platforms the DNS controller is created with a fake controller, and DNSRecords never gets ready. In these cases, what we need is that the tests rely on the Gateway status address as a name and connect to it directly instead of trying to resolve the DNS.

This change creates a new function to verify if the test is running on a platform with unmanaged DNS, and in this case it skips verifying DNSRecord and makes the resolution of DNS records be hardcoded to use the gateway address on the http client

### DIFF
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -460,6 +460,10 @@ func testGatewayAPIRBAC(t *testing.T) {
 }
 
 func testGatewayAPIDNS(t *testing.T) {
+	if !isDNSManagementSupported(t) {
+		t.Skip("this test can be executed just on platforms that support managed DNS")
+	}
+
 	domain := "gws." + dnsConfig.Spec.BaseDomain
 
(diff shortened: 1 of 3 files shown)


melvinjoseph86:
The TestGatewayAPI passed in the e2e-aws-operator job
        --- PASS: TestAll/serial/TestGatewayAPI (1041.90s)

/verified by e2e

rikatz:
/hold
for https://github.com/openshift/origin/pull/31262

gcs278:
/approve
/lgtm

This is a clean backport of https://github.com/openshift/cluster-ingress-operator/pull/1342 and the changes are identical. It's E2E only so, therefore no-risk to production code.

rikatz:
/close
closing in favor of automated cherry-pick at https://github.com/openshift/cluster-ingress-operator/pull/1514/changes

PR #1439, opened 2026-05-07
NE-2664: deploy haproxy as sidecar

jcmoraisjr:
Adding a new container to deploy HAProxy. This is a first but fully functional implementation that allows starting e2e and exploratory testsusing the new topology.

Key differences from the final version:

* The same image is used both on router and haproxy containers; and
* It is not possible to select HAProxy version.

This is due to some other and still pending stories that will handle both a new API field and building HAProxy image for all supported versions.

This update depends on https://github.com/openshift/router/pull/772

e2e tests updated in order to send a HTTP request to the router, because on the single container / embedded HAProxy, it is the reload-haproxy script the one sending the HTTP request the e2e test expects, and the new implementation does not use the reload script anymore.

https://redhat.atlassian.net/browse/NE-2664

## Why

[enhancements#1965](https://github.com/openshift/enhancements/pull/1965/changes) describes a new feature allowing users to choose a HAProxy version from IngressController API. Separating HAProxy on its own container is a simple and secure way to achieve this, with minimal disruption on the topology.

### DIFF
--- a/pkg/manifests/assets/router/deployment.yaml
+++ b/pkg/manifests/assets/router/deployment.yaml
@@ -10,17 +10,75 @@ spec:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     spec:
+      # manually configure the service account, so we can configure it only in the router container.
+      automountServiceAccountToken: false
       serviceAccountName: router
       # nodeSelector is set at runtime.
       priorityClassName: system-cluster-critical
+      shareProcessNamespace: true
+      initContainers:
+        - name: init-router
+          # image is set at runtime.
+          imagePullPolicy: IfNotPresent
+          terminationMessagePolicy: FallbackToLogsOnError
+          command: ["/bin/bash", "-c", "cp -R -p /var/lib/haproxy/* /mnt/config/"]
+          volumeMounts:
+          - mountPath: /mnt/config
+            name: haproxy-config
+        - name: haproxy
+          # image is set at runtime.
+          imagePullPolicy: IfNotPresent
+          restartPolicy: Always
+          securityContext:
+            # See https://bugzilla.redhat.com/2007246
+            allowPrivilegeEscalation: true
+            readOnlyRootFilesystem: true
+          terminationMessagePolicy: FallbackToLogsOnError
+          command: ["/var/lib/haproxy/start-haproxy"]
+          # admin.sock should follow routerHAProxyAdminSocket const
+          args: ["-W", "-db", "-S", "/var/lib/haproxy/run/admin.sock,mode,600", "-f", "/var/lib/haproxy/conf/haproxy.config"]
+          livenessProbe:
+            exec:
+              command: ["/bin/sh", "-c", "echo show version | socat - /var/lib/haproxy/run/haproxy.sock"]
+            failureThreshold: 3
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+            terminationGracePeriodSeconds: 10
+          readinessProbe:
+            failureThreshold: 3
+            exec:
+              command: ["/bin/sh", "-c", "echo show version | socat - /var/lib/haproxy/run/haproxy.sock"]
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          startupProbe:
+            failureThreshold: 120
+            exec:
+              # admin.sock should follow routerHAProxyAdminSocket const
+              command: ["/bin/sh", "-c", "echo show version | socat - /var/lib/haproxy/run/admin.sock"]
+            periodSeconds: 1
+            successThreshold: 1
+            timeoutSeconds: 1
+          resources:
+            requests:
+              cpu: 100m
+              memory: 256Mi
+          volumeMounts:
+          - mountPath: /var/lib/haproxy
+            name: haproxy-config
+          - mountPath: /etc/pki/tls/private
+            name: default-certificate
+            readOnly: true
+          - mountPath: /var/run/configmaps/service-ca
+            name: service-ca-bundle
+            readOnly: true
       containers:
         - name: router
           # image is set at runtime.
           imagePullPolicy: IfNotPresent
           securityContext:
-            # See https://bugzilla.redhat.com/2007246
-            allowPrivilegeEscalation: true
-            readOnlyRootFilesystem: false
+            readOnlyRootFilesystem: true
           terminationMessagePolicy: FallbackToLogsOnError
           # Merged at runtime.
           env:
@@ -60,14 +118,19 @@ spec:
             timeoutSeconds: 1
           resources:
             requests:
-              cpu: 100m
-              memory: 256Mi
+              cpu: 25m
+              memory: 64Mi
           volumeMounts:
+          - mountPath: /var/lib/haproxy
+            name: haproxy-config
           - mountPath: /etc/pki/tls/private
             name: default-certificate
             readOnly: true
           - mountPath: /var/run/configmaps/service-ca
             name: service-ca-bundle
+            readOnly: true
+          - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+            name: kube-api-access
             readOnly: true
       tolerations:
       # Ensure the pod isn't deleted during serial NoExecuteTaintManager tests.
@@ -82,6 +145,8 @@ spec:
         secret:
           defaultMode: 420
           # SecretName is set at run-time.
+      - name: haproxy-config
+        emptyDir: {}
       - name: service-ca-bundle
         configMap:
           defaultMode: 420
@@ -90,3 +155,20 @@ spec:
             path: service-ca.crt
           name: service-ca-bundle
           optional: false
+      - name: kube-api-access
+        projected:
+          defaultMode: 0400
+          sources:
+          - serviceAccountToken:
+              expirationSeconds: 3600
+              path: token
+          - configMap:
+              name: kube-root-ca.crt
+              items:
+              - key: ca.crt
+                path: ca.crt
+          - downwardAPI:
+              items:
+              - path: namespace
+                fieldRef:
+                  fieldPath: metadata.namespace
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -159,6 +159,12 @@ const (
 	StatsPortName = "metrics"
 
 	haproxyMaxTimeoutMilliseconds = 2147483647 * time.Millisecond
+
+	// routerHAProxyAdminSocket has the path to the unix socket of the master process CLI.
+	// This socket is used to issue reload, grab HAProxy version, list and manage active processes.
+	//
+	// NOTE: it is hardcoded in the deployment.yaml manifest, update there if updating here.
+	routerHAProxyAdminSocket = "/var/lib/haproxy/run/admin.sock"
 )
 
 // ensureRouterDeployment ensures the router deployment exists for a given
@@ -321,6 +327,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 
 	volumes := deployment.Spec.Template.Spec.Volumes
 	routerVolumeMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
+	haproxyVolumeMounts := deployment.Spec.Template.Spec.InitContainers[1].VolumeMounts
 
 	desiredReplicas := determineDeploymentReplicas(ci, ingressConfig, infraConfig)
 	deployment.Spec.Replicas = &desiredReplicas
@@ -550,6 +557,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 			MountPath: "/var/lib/haproxy/conf/error_code_pages",
 		}
 		routerVolumeMounts = append(routerVolumeMounts, httpErrorCodeVolumeMount)
+		haproxyVolumeMounts = append(haproxyVolumeMounts, httpErrorCodeVolumeMount)
 		if len(configmapName.Name) != 0 {
 			env = append(env, corev1.EnvVar{
 				Name:  "ROUTER_ERRORFILE_503",
@@ -813,7 +821,17 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 		env = append(env, corev1.EnvVar{Name: "ROUTE_LABELS", Value: routeSelector.String()})
 	}
 
+	// init image: just needs bash, reusing router's image is fine since we already need to download it anyway.
+	deployment.Spec.Template.Spec.InitContainers[0].Image = config.IngressControllerImage
+
+	// router image.
 	deployment.Spec.Template.Spec.Containers[0].Image = config.IngressControllerImage
+
+	// haproxy image: using router's image for now, but it needs its own image.
+	// HAProxy images: https://redhat.atlassian.net/browse/NE-2218
+	// API field: https://redhat.atlassian.net/browse/NE-2217
+	deployment.Spec.Template.Spec.InitContainers[1].Image = config.IngressControllerImage
+
 	deployment.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirst
 
 	var (
@@ -966,6 +984,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 			)
 			volumes = append(volumes, rsyslogConfigVolume, rsyslogSocketVolume)
 			routerVolumeMounts = append(routerVolumeMounts, rsyslogSocketVolumeMount)
+			haproxyVolumeMounts = append(haproxyVolumeMounts, rsyslogSocketVolumeMount)
 			deployment.Spec.Template.Spec.Containers = append(deployment.Spec.Template.Spec.Containers, syslogContainer)
 
 		case accessLogging.Destination.Type == operatorv1.SyslogLoggingDestinationType:
@@ -1271,6 +1290,7 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 
 	deployment.Spec.Template.Spec.Volumes = volumes
 	deployment.Spec.Template.Spec.Containers[0].VolumeMounts = routerVolumeMounts
+	deployment.Spec.Template.Spec.InitContainers[1].VolumeMounts = haproxyVolumeMounts
 
 	// If MIMETypes were supplied, expose the RouterEnableCompression and RouterCompressionMIMETypes
 	// environment variables.
@@ -1318,6 +1338,12 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 			Value: "true",
 		})
 	}
+
+	// ROUTER_HAPROXY_ADMIN_UNIX_SOCKET envvar defines where the router should find the HAProxy's admin socket.
+	env = append(env, corev1.EnvVar{
+		Name:  "ROUTER_HAPROXY_ADMIN_UNIX_SOCKET",
+		Value: routerHAProxyAdminSocket,
+	})
 
 	// TODO: The only connections from the router that may need the cluster-wide proxy are those for downloading CRLs,
 	// which, as of writing this, will always be http. If https becomes necessary, the router will need to mount the
@@ -1600,6 +1626,30 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 		return containers[i].Name < containers[j].Name
 	})
 	hashableDeployment.Spec.Template.Spec.Containers = containers
+	initContainers := make([]corev1.Container, len(deployment.Spec.Template.Spec.InitContainers))
+	for i, initContainer := range deployment.Spec.Template.Spec.InitContainers {
+		env := initContainer.Env
+		sort.Slice(env, func(i, j int) bool {
+			return env[i].Name < env[j].Name
+		})
+		initContainers[i] = corev1.Container{
+			Command:         initContainer.Command,
+			Env:             env,
+			Image:           initContainer.Image,
+			ImagePullPolicy: initContainer.ImagePullPolicy,
+			Name:            initContainer.Name,
+			LivenessProbe:   hashableProbe(initContainer.LivenessProbe),
+			ReadinessProbe:  hashableProbe(initContainer.ReadinessProbe),
+			StartupProbe:    hashableProbe(initContainer.StartupProbe),
+			SecurityContext: initContainer.SecurityContext,
+			Ports:           initContainer.Ports,
+			RestartPolicy:   initContainer.RestartPolicy,
+		}
+	}
+	sort.Slice(initContainers, func(i, j int) bool {
+		return initContainers[i].Name < initContainers[j].Name
+	})
+	hashableDeployment.Spec.Template.Spec.InitContainers = initContainers
 	hashableDeployment.Spec.Template.Spec.DNSPolicy = deployment.Spec.Template.Spec.DNSPolicy
 	hashableDeployment.Spec.Template.Spec.HostNetwork = deployment.Spec.Template.Spec.HostNetwork
 	volumes := make([]corev1.Volume, len(deployment.Spec.Template.Spec.Volumes))
@@ -1782,6 +1832,7 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 		containers[i+1] = *container.DeepCopy()
 	}
 	updated.Spec.Template.Spec.Containers = containers
+	updated.Spec.Template.Spec.InitContainers = expected.Spec.Template.Spec.InitContainers
 	updated.Spec.Template.Spec.DNSPolicy = expected.Spec.Template.Spec.DNSPolicy
 	updated.Spec.Template.Labels = expected.Spec.Template.Labels
 
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -11,6 +12,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -31,6 +33,11 @@ import (
 const (
 	ingressControllerImage = "quay.io/openshift/router:latest"
 	routerContainerName    = "router"
+
+	routerInitContainerName        = "init-router"
+	routerHAProxyContainerName     = "haproxy"
+	routerHAProxyConfigVolume      = "haproxy-config"
+	routerServiceAccountVolumeName = "kube-api-access"
 )
 
 var toleration = corev1.Toleration{
@@ -59,10 +66,10 @@ func checkDeploymentHasEnvSorted(t *testing.T, deployment *appsv1.Deployment) {
 	}
 }
 
-func checkDeploymentHasContainer(t *testing.T, deployment *appsv1.Deployment, name string, expect bool) {
+func checkDeploymentHasContainer(t *testing.T, containers []corev1.Container, name string, expect bool) {
 	t.Helper()
 
-	for _, container := range deployment.Spec.Template.Spec.Containers {
+	for _, container := range containers {
 		if container.Name == name {
 			if expect {
 				return
@@ -75,20 +82,38 @@ func checkDeploymentHasContainer(t *testing.T, deployment *appsv1.Deployment, na
 	}
 }
 
-func checkRouterContainerSecurityContext(t *testing.T, deployment *appsv1.Deployment) {
+func getHAProxyContainer(t *testing.T, deployment *appsv1.Deployment) *corev1.Container {
 	t.Helper()
-	checkDeploymentHasContainer(t, deployment, routerContainerName, true)
+	checkDeploymentHasContainer(t, deployment.Spec.Template.Spec.InitContainers, routerHAProxyContainerName, true)
 
-	for _, container := range deployment.Spec.Template.Spec.Containers {
-		if container.Name == routerContainerName {
-			if allowPrivEsc := container.SecurityContext.AllowPrivilegeEscalation; allowPrivEsc == nil || !*allowPrivEsc {
-				t.Errorf("%s container does not have securityContext.allowPrivilegeEscalation: true", routerContainerName)
-			}
-			if readOnlyFS := container.SecurityContext.ReadOnlyRootFilesystem; readOnlyFS == nil || *readOnlyFS {
-				t.Errorf("%s container does not have securityContext.readOnlyRootFilesystem: false", routerContainerName)
-			}
+	for _, container := range deployment.Spec.Template.Spec.InitContainers {
+		if container.Name == routerHAProxyContainerName {
+			return &container
 		}
 	}
+
+	t.Errorf("haproxy container not found")
+	return nil
+}
+
+func checkHAProxyContainerSecurityContext(t *testing.T, deployment *appsv1.Deployment) {
+	t.Helper()
+	container := getHAProxyContainer(t, deployment)
+
+	if allowPrivEsc := container.SecurityContext.AllowPrivilegeEscalation; allowPrivEsc == nil || !*allowPrivEsc {
+		t.Errorf("%s container does not have securityContext.allowPrivilegeEscalation: true", routerContainerName)
+	}
+	if readOnlyFS := container.SecurityContext.ReadOnlyRootFilesystem; readOnlyFS == nil || !*readOnlyFS {
+		t.Errorf("%s container does not have securityContext.readOnlyRootFilesystem: true", routerContainerName)
+	}
+}
+
+func checkHAProxyContainerSocketRef(t *testing.T, deployment *appsv1.Deployment) {
+	t.Helper()
+	container := getHAProxyContainer(t, deployment)
+
+	require.Equal(t, container.Args, []string{"-W", "-db", "-S", routerHAProxyAdminSocket + ",mode,600", "-f", "/var/lib/haproxy/conf/haproxy.config"}, "containers[haproxy].Args")
+	require.Equal(t, container.StartupProbe.Exec.Command, []string{"/bin/sh", "-c", "echo show version | socat - " + routerHAProxyAdminSocket}, "containers[haproxy].startupProbe.exec.command")
 }
 
 func checkDeploymentHash(t *testing.T, deployment *appsv1.Deployment) {
@@ -540,7 +565,8 @@ func Test_desiredRouterDeployment(t *testing.T) {
 		t.Fatalf("invalid router Deployment: %v", err)
 	}
 
-	checkRouterContainerSecurityContext(t, deployment)
+	checkHAProxyContainerSecurityContext(t, deployment)
+	checkHAProxyContainerSocketRef(t, deployment)
 	checkDeploymentHash(t, deployment)
 
 	if deployment.Spec.Replicas == nil {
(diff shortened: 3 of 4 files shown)


pkg/operator/controller/ingress/deployment.go
```diff
@@ -1591,6 +1662,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 		return containers[i].Name < containers[j].Name
 	})
 	hashableDeployment.Spec.Template.Spec.Containers = containers
+	hashableDeployment.Spec.Template.Spec.InitContainers = deployment.Spec.Template.Spec.InitContainers
```
bentito: If k8s injects defaults into `initContainer` could there be mismatches? I was comparing to what we do in `hashableDeployment` for `Containers` where we seem to be stripping out fields we don't want to trigger updates on.

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,6 +10,8 @@ spec:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     spec:
+      # manually configure the service account, so we can configure it only in the router container.
+      automountServiceAccountToken: false
```
bentito, line 14: This is a great security improvement

bentito:
/lgtm 
just the one comment, might need handling, up to you

gcs278:
I'm reviewing the enhancement now, and I'd like to also take a look at this approach:
/cc

gcs278:
@jcmoraisjr does this depend on https://github.com/openshift/router/pull/772? could you add a note in the PR description about this?

jcmoraisjr:
Nice catch, updated. The merge itself is not dependent on router's counterpart, since using sidecar is an opt-in. But in fact, if using HAProxy as a sidecar, the router should also be deployed with https://github.com/openshift/router/pull/772

jcmoraisjr:
/hold

this depends on https://github.com/openshift/router/pull/772 being merged first.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1310,6 +1317,64 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		Args:            []string{"-W", "-db", "-S", routerHAProxyAdminSocket + ",mode,600", "-f", "/var/lib/haproxy/conf/haproxy.config"},
+		Resources:       deployment.Spec.Template.Spec.Containers[0].Resources,
+		VolumeMounts:    haproxyVolumeMounts,
+	})
+
+	// ROUTER_HAPROXY_ADMIN_UNIX_SOCKET envvar defines where the router should find the HAProxy's admin socket.
+	env = append(env, corev1.EnvVar{
+		Name:  "ROUTER_HAPROXY_ADMIN_UNIX_SOCKET",
+		Value: routerHAProxyAdminSocket,
+	})
```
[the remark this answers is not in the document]
jcmoraisjr: In fact we changed the approach, there is not an opt-in anymore, and I also updated the description. Maybe you reviewed just before my update. Can you double check and let me know if there is some other pending?

gcs278:
Just curious to kick off some CI combo jobs with openshift/router#772 to see if it works - the default CI jobs are failing with because the default router isn't coming up (needs the changes in openshift/router#772):
```
level=error msg=Cluster operator ingress Available is False with IngressUnavailable: The "default" ingress controller reports Available=False: IngressControllerUnavailable: One or more status conditions indicate unavailable: DeploymentAvailable=False (DeploymentUnavailable: The deployment has Available status condition set to False (reason: MinimumReplicasUnavailable) with message: Deployment does not have minimum availability.)
```

This should in theory should start getting us some signal:
/testwith e2e-aws-operator openshift/router#772
/testwith e2e-aws-ovn openshift/router#772
/testwith e2e-aws-ovn-serial-1of2 openshift/router#772
/testwith e2e-aws-ovn-serial-2of2 openshift/router#772

gcs278:
Hmmm maybe I can't do multiple at once:
/testwith e2e-aws-operator openshift/router#772

gcs278:
ah i forgot you need the full specifier for the test
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn openshift/router#772
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-1of2 openshift/router#772
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-2of2 openshift/router#772

gcs278:
Okay maybe I had the formatting wrong AND you can't do multiple at once:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772

gcs278:
Not working so far
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772

gcs278:
Hey it finally [started](https://github.com/openshift/cluster-ingress-operator/pull/1439/checks?check_run_id=77202385799)! I guess I can't include any other text in the comment? :shrug:

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1310,6 +1317,64 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, corev1.Volume{
+		Name: routerHAProxyConfigVolume,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	})
+
+	// The initContainer is responsible for copying static config data from router's container to the shared volume
+	deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, corev1.Container{
+		Name:            routerInitContainerName,
```
gcs278: see the [failure](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-ovn/2057493651105255424) `: [Monitor:termination-message-policy][sig-arch] all containers in ns/openshift-ingress must have terminationMessagePolicy=FallbackToLogsOnError `

You'll need to add `terminationMessagePolicy=FallbackToLogsOnError` to the init container.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1310,6 +1317,64 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+
+	// HAProxy sidecar mount all the router volumes, except the service account token
+	haproxyVolumeMounts := slices.Clone(deployment.Spec.Template.Spec.Containers[0].VolumeMounts)
+	haproxyVolumeMounts = slices.DeleteFunc(haproxyVolumeMounts, func(mount corev1.VolumeMount) bool {
+		return mount.Name == routerServiceAccountVolumeName
+	})
+
+	// Finally, adding haproxy as a sidecar container
+	deployment.Spec.Template.Spec.Containers = append(deployment.Spec.Template.Spec.Containers, corev1.Container{
+		Name:            routerHAProxyContainerName,
```
gcs278: We should think about the impacts to the liveness probe.

Right now, the HAProxy sidecar has no liveness probe. If HAProxy crashes, it's the router container's liveness probe that fails which causes kubelet to restart the router container, not the HAProxy sidecar that actually died.

The sidecar stays dead. We should consider adding a basic liveness probe to the HAProxy container so kubelet restarts the right thing.

It's actually minor improvement - we can restart a failed HAProxy without restarting/reloading the router. However, we should verify that the router and HAProxy handle sidecar restarts gracefully. Specifically, can HAProxy rebind stale socket files on startup, and does the router recover when connecting to the recreated sockets? I think it does, but worth some manual testing.

jcmoraisjr: Just added probes checking the dynamic update API.

Also, moved all the configuration to the manifest, since we are always adding the init and sidecar containers, this is not conditional anymore.

> HAProxy rebind stale socket files on startup

Yep the file is truncated/recreated when HAProxy starts, the problem would be being locked or not having permission to change it.

Just gave it a quick try by sigkill-ing the haproxy process, simulating a crash - I can see its container restarting, haproxy is able to recreate the socket, I can manually access it (socat) and router can issue API calls as well.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -804,7 +806,17 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	// init image: just needs bash, reusing router's image is fine since we already need to download it anyway.
+	deployment.Spec.Template.Spec.InitContainers[0].Image = config.IngressControllerImage
+
+	// router image.
 	deployment.Spec.Template.Spec.Containers[0].Image = config.IngressControllerImage
+
+	// haproxy image: using router's image for now, but it needs its own image.
+	// HAProxy images: https://redhat.atlassian.net/browse/NE-2218
+	// API field: https://redhat.atlassian.net/browse/NE-2217
+	deployment.Spec.Template.Spec.Containers[1].Image = config.IngressControllerImage
```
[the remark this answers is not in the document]
jcmoraisjr: It is expected that the sourcing manifest always have one init container and two regular ones, and since this is a premise of the deployment, we change values directly without safeguards for simplicity. Our unit tests will catch any changes in the manifest side that does not have its counterpart in the code.

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-1of2 openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-2of2 openshift/router#772

gcs278:
@jcmoraisjr it's worth looking at those multi-pr failures ([e2e-aws-ovn](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-ovn/2057493651105255424), [e2e-aws-operator](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-operator/2057493631576576000)). 

I see logging failures that seem related to setting up syslog and container logging in the sidecar model. `TestSyslogLogging`, `TestContainerLoggingMaxLength`, and `TestContainerLoggingMinLength` all timed out waiting for HAProxy access log output. Worth investigating whether HAProxy in the sidecar is correctly writing to the rsyslog socket/UDP endpoint.

davidesalerno:
The PR describes HAProxy as a sidecar, but the implementation models it as a regular container under spec.containers, not as a Kubernetes native sidecar under `spec.initContainers` with `restartPolicy: Always` (see [here](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#sidecar-example)). Is that intentional? If yes, it may be worth clarifying in the PR description/comments that this uses a regular multi-container Pod pattern rather than the native sidecar-container feature.

jcmoraisjr:
> The PR describes HAProxy as a sidecar, but the implementation models it as a regular container under spec.containers

When I call it sidecar, I mean the well-known pattern of having a service container along with the main one, not the terminology used on init containers that runs along with the regular ones.

Now, talking about the native sidecar approach, I think it fits well for haproxy and should make the start script simpler, maybe it could even be dropped. What do you folks think? So we could adjust the topology instead of adjusting the naming. Let me invest some time on validating how/if this can simplify our deployment.

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

jcmoraisjr:
> Worth investigating whether HAProxy in the sidecar is correctly writing to the rsyslog socket/UDP endpoint.

Tks @gcs278. Just moved haproxy to initContainer / native sidecar model, which seems a better fit since it starts before and finishes after the router itself, and also does not touch the containers[] list, removing the risk of breaking some premise.

I've just restarted the jobs you suggested and pointed the failures, lets see how it behaves now.

pkg/operator/controller/ingress/deployment.go
gcs278, line 989: You'll need to append the volume mount to the haproxy container - this will fix a variety of sys log-related CI [failures](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-operator/2059993076897681408):

```suggestion
			deployment.Spec.Template.Spec.InitContainers[1].VolumeMounts = append(
                deployment.Spec.Template.Spec.InitContainers[1].VolumeMounts,
                rsyslogSocketVolumeMount,
            )
```

gcs278, line 559: You will also need this httpErrorCodeVolumeMount in the HAProxy container - you may want to consider making a `haproxyVolumeMounts` equivalent and follow the same pattern - up to you.

jcmoraisjr: Good catch, thanks! I also revisited all the `routerVolumeMounts`, updated both syslog and http pages, all the others seem not to be used by haproxy. Lets run new tests and see what happens.

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

gcs278:
I was looking at the `TestSyslogLogging` failure [here](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-operator/2060426125649121280) - that baffles me. The loki [logs](https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fmulti-pr-openshift-cluster-ingress-operator-1439-openshift-router-772-e2e-aws-operator%2F2060426125649121280%5C%22%7D%20%7C%20unpack%20%7C%20pod%3D%5C%22syslog%5C%22%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221780073421107%22,%22to%22:%221780109421109%22%7D%7D) for that container show no readiness probe from the router (should be hitting `http://localhost:80/_______internal_router_healthz`) every 10 seconds. That's the assumption that this test is relying on - the access logs are working, but not showing any readinessProbes (but the router pods is going ready).

Not sure - just thinking out loud.

pkg/manifests/assets/router/deployment.yaml
gcs278: I think this can be removed - router shouldn't need to bind to any privileged port like HAProxy does.

gcs278: this also may be able to be set to true now that we write into a shared volume - but I'm not sure - something to follow up on later.

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial1of2 openshift/router#772

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-1of2 openshift/router#772

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,17 +10,74 @@ spec:
...
           securityContext:
             # See https://bugzilla.redhat.com/2007246
             allowPrivilegeEscalation: true
-            readOnlyRootFilesystem: false
+            readOnlyRootFilesystem: true
+          terminationMessagePolicy: FallbackToLogsOnError
+          command: ["/var/lib/haproxy/start-haproxy"]
+          # admin.sock should follow routerHAProxyAdminSocket const
+          args: ["-W", "-db", "-S", "/var/lib/haproxy/run/admin.sock,mode,600", "-f", "/var/lib/haproxy/conf/haproxy.config"]
+          livenessProbe:
```
gcs278, line 40: We should also look at the router's liveness probe. It currently it checks `haproxy.sock`, so if HAProxy dies, the router's liveness fails too and kubelet may restart the router container unnecessarily.

Best practice with sidecars is to only restart the container that actually failed. The router's readiness and startup probes should still check HAProxy (since the pod shouldn't serve traffic without it), but liveness should only reflect the router's own health.

What do you think?

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,17 +10,74 @@ spec:
...
+            failureThreshold: 120
+            exec:
+              command: ["/bin/sh", "-c", "echo show version | socat - /var/lib/haproxy/run/admin.sock"]
+            periodSeconds: 1
+            successThreshold: 1
+            timeoutSeconds: 1
+          resources:
+            requests:
+              cpu: 100m
+              memory: 256Mi
```
gcs278, line 66: Hmm. This is effectively doubling the amount of CPU & Memory request in a cluster (haproxy + router requests). That's potentially a concern for large clusters running many router replicas.

The impacts might be felt in a variety of places like managed services or our CI clusters. It's something to think about - I don't know the right answer. I'd be a risk to try to determine a split to keep it too 100m and 256Mi total. @Miciah might have some better insights on this.

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: Good point. I've just walked into the operator and router code, wondering if I'm not missing anything.

* Liveness is configured against `/healthz` according to the [manifest](https://github.com/openshift/cluster-ingress-operator/blob/140e0bf13b3d01c369672c766c44b4be0b4ec78c/pkg/manifests/assets/router/deployment.yaml#L33-L42);
* Router's `:1936/healthz` can be configured either in HAProxy's monitor-uri or router's listener. This seems to be always configured in router's listener since [ROUTER_METRICS_TYPE](https://github.com/openshift/cluster-ingress-operator/blob/140e0bf13b3d01c369672c766c44b4be0b4ec78c/pkg/operator/controller/ingress/deployment.go#L564) and [STATS_PORT](https://github.com/openshift/cluster-ingress-operator/blob/140e0bf13b3d01c369672c766c44b4be0b4ec78c/pkg/operator/controller/ingress/deployment.go#L878-L881) are hard-coded;
* Router configures liveness in [template.go](https://github.com/openshift/router/blob/676113436feb61e5c89376d6a7ae66fdaefe8e98/pkg/cmd/infra/router/template.go#L639-L644) and configures `/healthz` in [metrics.go](https://github.com/openshift/router/blob/676113436feb61e5c89376d6a7ae66fdaefe8e98/pkg/router/metrics/metrics.go#L47);

So it seems that the missing part is to conditionally configure `checkSocket` since this is already being done in the sidecar. I just did this in https://github.com/openshift/router/pull/772, [here](https://github.com/jcmoraisjr/openshift-router/blob/909baf1f3931357bc3f23c7b4e7ff1ee62812eb5/pkg/cmd/infra/router/template.go#L652-L656).

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn https://github.com/openshift/router/pull/772

gcs278 (commented):
Oh darn I forgot to hit the "Submit Review" button, sorry!

pkg/manifests/assets/router/deployment.yaml
gcs278: Nice fix - didn't think it'd be that simple! The router readiness is based on haproxy still (can we pass traffic) and the router liveness if only based on the router process being responsive. And it makes sense to keep the startup the same as readiness here (passing traffic means we are initialized). LGTM.

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: This is just about requests (no limits applied), wondering if simply splitting the old metric by dividing both mem and cpu by 2 makes sense here.

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: Thanks!

Yep exactly - readiness being the ability of the router to configure the proxy, and liveness/start as the ability to be running despite of further configurations.

pkg/manifests/assets/router/deployment.yaml
gcs278: yea that's a simple solution, but I feel it doesn't quite reflect the processing requirements of router vs. haproxy. And arguably a starved HAProxy is worse - it impacts live traffic, whereas a starved router just slows route reconciliation.

Some options are:
1. **50m/128Mi each** — your suggestion - total unchanged. May underrepresent HAProxy's actual usage.
2. **75m/192Mi HAProxy + 25m/64Mi router** — total unchanged. Better reflects actual usage split.
3. **100m/256Mi HAProxy + 25m/64Mi router** — total goes up ~25%. HAProxy keeps what it had before, router gets a realistic minimum.

Option 2 avoids increasing the total pod request, while option 3 avoids reducing HAProxy's allocation from what it has today. I lean towards option 2 to be a bit more "fair" for representing the processing requirements of each without impacting the total footprint of all our users.

I'll leave it up to you though.

jcmoraisjr: In fact you helped me to refresh my memory that CPU requests are in fact shares of CPU, so making it lower will give less shares of CPU to each of the containers, which makes them to run slower in the same proportion if the node hits 100% CPU utilization. In the end, the hardcoded 100m is already a bad number on this regard so we shouldn't make it worse.

That said I'm more inclined to choose your option 3, which makes the most critical process (the proxy) to receive the same configuration as before, provided that the increase on 25m/64Mi in requests isn't a problem. The number are quite low btw if we think about the role of the component, so my guess is that option 3 should work well for us without side effect.

gcs278: Yea ultimately this might lead us to have to do a survey on the baseline resources HAProxy and the router use. We do have some old investigations in to baseline (static) memory usage in HAProxy: https://gist.github.com/frobware/2b527ce3f040797909eff482a4776e0b (also might be of interest for your DCM work).

For 1000 routes (assuming backends==routes) and no dynamic server slots, HAProxy's static memory usage is 18MB. Of course, we need more overhead than that, but 256Mi is probably a reasonable minimum overhead.

I don't quite know what baseline of memory the router uses. 64Mb seems reasonable enough with an educated guess.

I think option 3 is fine - what do you think about adding this to the EP as an implementation detail to document our thought process here? Doesn't have to be detailed - just more of a PSA to folks and a papertrail of our decision and why.

jcmoraisjr:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator https://github.com/openshift/router/pull/772

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,17 +10,74 @@ spec:
...
+          restartPolicy: Always
           securityContext:
             # See https://bugzilla.redhat.com/2007246
             allowPrivilegeEscalation: true
-            readOnlyRootFilesystem: false
+            readOnlyRootFilesystem: true
+          terminationMessagePolicy: FallbackToLogsOnError
+          command: ["/var/lib/haproxy/start-haproxy"]
+          # admin.sock should follow routerHAProxyAdminSocket const
+          args: ["-W", "-db", "-S", "/var/lib/haproxy/run/admin.sock,mode,600", "-f", "/var/lib/haproxy/conf/haproxy.config"]
```
gcs278, line 39: More of a minor point, but this is hardcoded and not using the offiical `ROUTER_HAPROXY_ADMIN_UNIX_SOCKET` env variable, so if that ENV was changed on the deployment, this would break - defeating the purpose of the ENV variable. Though, I know it's a bit messy doing it via GO.

gcs278:
@jcmoraisjr Generally LGTM. I have a couple of comments - my biggest concern is the resource usage. We should probably get a second opinion on that. Otherwise, it's looking good.

gcs278:
https://github.com/openshift/router/pull/772 has merged, CI jobs should be working now.

/retest

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: Updated to option 3, so HAProxy inherits the former router requests, and the router container uses now 25m/64Mb

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: Yea this is one of the hardcoded configs placed in the manifest. It is a convenience since it makes the configuration clear and simpler. I tried to circumvent this by adding comments about this dependency, and also added an unit test that should fail if only one side is updated. How does it sound?

gcs278:
My concern about the resource usage has been addressed. I have no other comments. Sorry I jumped into reviewing this one even though I wasn't assigned.

I'll reserve the LGTM for @davidesalerno and/or @bentito 
/approve

jcmoraisjr:
https://github.com/openshift/router/pull/772 already merged

/unhold

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,17 +10,75 @@ spec:
...
       priorityClassName: system-cluster-critical
-      containers:
-        - name: router
+      shareProcessNamespace: true
+      initContainers:
+        - name: init-router
+          # image is set at runtime.
+          imagePullPolicy: IfNotPresent
+          terminationMessagePolicy: FallbackToLogsOnError
+          command: ["/bin/bash", "-c", "cp -R -p /var/lib/haproxy/* /mnt/config/"]
```
bentito, line 24: I think the `haproxy/*` might be a problem if the dir is completely empty. This is safer I think:
```suggestion
          command: ["/bin/bash", "-c", "cp -R -p /var/lib/haproxy/. /mnt/config/"]
```

pkg/manifests/assets/router/deployment.yaml
```diff
@@ -10,17 +10,75 @@ spec:
...
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     spec:
+      # manually configure the service account, so we can configure it only in the router container.
+      automountServiceAccountToken: false
       serviceAccountName: router
       # nodeSelector is set at runtime.
       priorityClassName: system-cluster-critical
-      containers:
-        - name: router
+      shareProcessNamespace: true
```
bentito, line 18: I think this is pre-sidecar and can go away?

jcmoraisjr: In fact this is post-sidecar. We need it so the router process can find the haproxy process for its metrics. There's already the router counterpart code ps-ing haproxy from there. Without the shared process namespace, the haproxy process are missing from the router metrics.

pkg/manifests/assets/router/deployment.yaml
jcmoraisjr: Both router and haproxy depends on `/var/lib/haproxy/` populated. Making it work here would just postpone this problem.

pkg/manifests/assets/router/deployment.yaml
bentito: ah, okay, I get it now

pkg/manifests/assets/router/deployment.yaml
bentito: makes sense

rhamini3:
tested operator changes on 5.0 cluster

```
NAME                              READY   STATUS    RESTARTS      AGE
router-default-785d657597-cwjks   2/2     Running   0             71m
router-default-785d657597-r4c6d   2/2     Running   3 (73m ago)   85m

router container logs
% oc -n openshift-ingress logs router-default-785d657597-cwjks -c router 
I0624 18:58:07.631632      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:58:12.629498      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:58:21.088146      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:58:30.952481      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:58:35.953332      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:59:13.749098      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 18:59:18.749535      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 19:00:29.529021      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 19:00:34.530284      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0624 19:00:49.258766      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
```
3 containers in the router pod
```
Init Containers:
  init-router:
    Container ID:  cri-o://875973a800ff2d3f07717affe86a1502ab3c76e3360f82a1796243b3eb2832b8
    Image:         registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Image ID:      registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/bash
      -c
      cp -R -p /var/lib/haproxy/* /mnt/config/
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Wed, 24 Jun 2026 14:58:04 -0400
      Finished:     Wed, 24 Jun 2026 14:58:04 -0400
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /mnt/config from haproxy-config (rw)
  haproxy:
    Container ID:  cri-o://d86468bdab401239d27534a3eeed3aca2c0d916ca6a60eb938ed06a8eedf0480
    Image:         registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Image ID:      registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Port:          <none>
    Host Port:     <none>
    Command:
      /var/lib/haproxy/start-haproxy
    Args:
      -W
      -db
      -S
      /var/lib/haproxy/run/admin.sock,mode,600
      -f
      /var/lib/haproxy/conf/haproxy.config
    State:          Running
      Started:      Wed, 24 Jun 2026 14:58:05 -0400
    Ready:          True
    Restart Count:  0
    Requests:
      cpu:        100m
      memory:     256Mi
    Liveness:     exec [/bin/sh -c echo show version | socat - /var/lib/haproxy/run/haproxy.sock] delay=0s timeout=1s period=10s #success=1 #failure=3
    Readiness:    exec [/bin/sh -c echo show version | socat - /var/lib/haproxy/run/haproxy.sock] delay=0s timeout=1s period=10s #success=1 #failure=3
    Startup:      exec [/bin/sh -c echo show version | socat - /var/lib/haproxy/run/admin.sock] delay=0s timeout=1s period=1s #success=1 #failure=120
    Environment:  <none>
    Mounts:
      /etc/pki/tls/private from default-certificate (ro)
      /var/lib/haproxy from haproxy-config (rw)
      /var/run/configmaps/service-ca from service-ca-bundle (ro)
Containers:
  router:
    Container ID:   cri-o://14e6447d8121c42b91255868706f0271fd4f37c7c52fb5a9938044b912705842
    Image:          registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Image ID:       registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945
    Ports:          80/TCP, 443/TCP, 1936/TCP
    Host Ports:     0/TCP, 0/TCP, 0/TCP
    State:          Running
      Started:      Wed, 24 Jun 2026 14:58:07 -0400
    Ready:          True
    Restart Count:  0

  Type    Reason          Age   From               Message
  ----    ------          ----  ----               -------
  Normal  Scheduled       94m   default-scheduler  Successfully assigned openshift-ingress/router-default-785d657597-cwjks to ip-10-0-30-196.us-east-2.compute.internal
  Normal  AddedInterface  94m   multus             Add eth0 [10.128.2.16/23] from ovn-kubernetes
  Normal  Pulling         94m   kubelet            Pulling image "registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945"
  Normal  Pulled          94m   kubelet            Successfully pulled image "registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945" in 1.805s (1.805s including waiting). Image size: 305841524 bytes.
  Normal  Created         94m   kubelet            Container created
  Normal  Started         94m   kubelet            Container started
  Normal  Pulled          94m   kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945" already present on machine and can be accessed by the pod
  Normal  Created         94m   kubelet            Container created
  Normal  Started         94m   kubelet            Container started
  Normal  Pulled          94m   kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-vcdjlft/stable@sha256:f4a74c11482efa74a1bd2698fb434aa565782b2e23c16702dc94c73ab9234945" already present on machine and can be accessed by the pod
  Normal  Created         94m   kubelet            Container created
  Normal  Started         94m   kubelet            Container started
```
confirmed that haproxy container contains all routes and expected version
```
% oc -n openshift-ingress logs router-default-785d657597-cwjks -c haproxy
Starting HAProxy. SIGUSR1 timeout is 30s
[NOTICE]   (4) : Reloading HAProxy
[WARNING]  (4636) : Proxy public stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy public_ssl stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy fe_sni stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy fe_no_sni stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_sni stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_no_sni stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy openshift_default stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_edge_http:default:edge-route stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:default:reen-route stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_http:default:unsec-route stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_tcp:openshift-authentication:oauth-openshift stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:openshift-console:console stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_edge_http:openshift-console:downloads stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_tcp:openshift-ingress-canary:canary stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:openshift-monitoring:alertmanager-main stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:openshift-monitoring:prometheus-k8s stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:openshift-monitoring:prometheus-k8s-federate stopped (cumulated conns: FE: 0, BE: 0).
[WARNING]  (4636) : Proxy be_secure:openshift-monitoring:thanos-querier stopped (cumulated conns: FE: 0, BE: 0).
[NOTICE]   (4) : New worker (4645) forked
[NOTICE]   (4) : Loading success.
[NOTICE]   (4) : haproxy version is 2.8.18-ae90be6
[NOTICE]   (4) : path to executable is /usr/sbin/haproxy
[WARNING]  (4) : Former worker (4636) exited with code 0 (Exit)
```
confirmed that haproxy is functioning as intended

/verified by @rhamini3

gcs278:
unrelated failures
/test e2e-aws-ovn
/test e2e-azure-operator

bentito:
@coderabbitai can you display what you've been told to learn so far for this repo

PR #1440, opened 2026-05-07
OCPBUGS-55050: Remove resources limits from Gateway API proxy containers

rikatz:
When a Gateway is deployed, Istio will create a deployment containing resources limits set, which is not expected on Openshift components. These resource limits are added as part of the default helm template used by Istio controller when creating the deployments and Pods.

An attempt to make Istio/Sail Library to have a default configuration on resource limits is not possible, given the Kubernetes resources struct has omitempty, which means when something is null omit, and when it ommits it considers that a user passing null has 'no opinion' as opposed to 'I want to set it null'.

This way, the  only way of fixing it is by telling the Gateway customization, via the Gateway Class that we want to enforce the resource to be null, and then when Istio deploys this Gateway instance it will remove the resources.limits field from the proxy pods

The majority of this PR was originally copied from https://github.com/openshift/cluster-ingress-operator/pull/1428

### DIFF
--- a/pkg/operator/controller/gatewayclass/istio_sail_installer.go
+++ b/pkg/operator/controller/gatewayclass/istio_sail_installer.go
@@ -210,10 +210,10 @@ func openshiftValues(enableInferenceExtension bool, operandNamespace string, gat
 			}
 		}
 		if extraConfig.infraConfig != nil {
-			if hpaConfig, err := buildHorizontalPodAutoscalerConfig(extraConfig.infraConfig, gatewayclasses); err != nil {
-				return nil, fmt.Errorf("failed to build HPA config: %w", err)
+			if gwClassConfig, err := buildGatewayClassesConfig(extraConfig.infraConfig, gatewayclasses); err != nil {
+				return nil, fmt.Errorf("failed to build gateway class config: %w", err)
 			} else {
-				val.GatewayClasses = hpaConfig
+				val.GatewayClasses = gwClassConfig
 			}
 		}
 	}
@@ -244,10 +244,9 @@ func buildProxyMetadata(proxyConfig *configv1.Proxy) map[string]string {
 	return proxyMetadata
 }
 
-// buildHorizontalPodAutoscalerConfig returns Istio configuration for the
-// horizontal pod autoscaler given an infrastructure config and a slice of
-// gatewayclasses.
-func buildHorizontalPodAutoscalerConfig(infraConfig *configv1.Infrastructure, gatewayclasses []gatewayapiv1.GatewayClass) (json.RawMessage, error) {
+// buildGatewayClassesConfig returns Istio per-gatewayclass configuration
+// overlays given an infrastructure config and a slice of gatewayclasses.
+func buildGatewayClassesConfig(infraConfig *configv1.Infrastructure, gatewayclasses []gatewayapiv1.GatewayClass) (json.RawMessage, error) {
 	const maxReplicas = 10
 
 	var minReplicas = 2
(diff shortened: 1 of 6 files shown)


test/e2e/util_gatewayapi_test.go
```diff
@@ -811,6 +811,46 @@ func assertHorizontalPodAutoscalerEnabled(t *testing.T, namespaceName, gatewayNa
...
+
+		// We need to be sure that the istio-proxy exists, and that it has no resources.limits.
+		// If we cannot find istio-proxy, we must try again
+		if !foundIstioProxy {
+			t.Log("Container 'istio-proxy' was not found; retrying...")
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Errorf("Failed to verify terminationMessagePolicy on deployment %s: %v", deploymentName, err)
```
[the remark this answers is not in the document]
rikatz: fixed on my last commit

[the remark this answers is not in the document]
rikatz: @coderabbitai this is fixed

test/e2e/util_gatewayapi_test.go
```diff
@@ -811,6 +811,46 @@ func assertHorizontalPodAutoscalerEnabled(t *testing.T, namespaceName, gatewayNa
 	}
 }
 
+// assertProxyDeployCustomConfigurations verifies that the deployment sets
+// the right custom configurations on pods.
+func assertProxyDeployCustomConfigurations(t *testing.T, namespaceName, gatewayName, gatewayclassName string) {
+	t.Helper()
```
Thealisyed, line 817: The function mention "custom configurations" but it only checks one thing, the resource limits are nil. 
Would renaming it as something like `assertProxyHasNoResourceLimits` would be more precise?

Thealisyed: Actually on second thought if we think in terms of future proofing, your original name would make more sense if we intend to add more custom configs.. 🤔

rikatz: yeap, exactly. @gcs278 was working on some other similar issue: https://github.com/openshift/cluster-ingress-operator/pull/1428/changes

and at the moment we need to check terminationPolicy it wont be ProxyResourceLimits anymore

pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -262,6 +261,26 @@ func buildHorizontalPodAutoscalerConfig(infraConfig *configv1.Infrastructure, ga
...
+					"spec": map[string]any{
+						"containers": []map[string]any{
+							{
+								"name": gatewayProxyContainerName,
+								// OCPBUGS-55050: Managed cluster should set requests but not limits
+								// Istio sets resource limits by default, that come from the helm chart.
+								// This change enforces the resources.limits to be null, so we drop this value
+								// during the Proxy deployment
+								"resources": map[string]any{
+									"limits": nil,
```
gcs278, line 276: Looks like Envoy bases it's number of threads off of the number of cores that it sees (see John's [comment](https://github.com/istio/istio/issues/51456#issuecomment-2155220061)).

If we do this, users with large 256 core nodes will have each Envoy spin up 256 threads - probably not desirable. Anyways, I think we need to set `meshConfig.defaultConfig.concurrency = 2` to effectively achieve the same number of threads that was being provisions by these limits.

This is something we should test to make sure that `meshConfig.defaultConfig.concurrency = 2` is effectively the same as `resources.limits.cpu: 2000m` - I think it's still an improvement because it's not being hard-capped at 2 CPU-seconds per period anymore.  We should understand the impact either way. 

In the future, we may want to talk more about Envoy threading when we look into performance - HAProxy is [defaulted to 4](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L139) threads, and users can adjust via `Spec.TuningOptions.ThreadCount`. And Envoy is defaulted to 2.

Thealisyed (commented):
/lgtm
Leaving the approve for Grant :)

rikatz:
/close
Won't fix. Removing CPU limit from Envoy is a bad move. I will update the exception code/comment with explanation on why we are not removing the limits

PR #1446, opened 2026-05-14
OCPBUGS-74373: Remove restriction of unmanaged x-k8s.io

rikatz:
Previously, cluster-ingress-operator would manage and block any updates to CRDs that were from group gateway.networking.x-k8s.io.

After some analysis we identified that OSSM/Istio does filter these resources in case it does not have a flag to enable alpha APIs, which means  it is safe to allow users to deploy these CRDs.

This way, this change removes any restriction on x-k8s.io group, and removes any further reconciliation from CIO on x-k8s.io group.

This is a manual cherry-pick of #1336.

### DIFF
--- a/manifests/01-validating-admission-policy-ibm-cloud-managed.yaml
+++ b/manifests/01-validating-admission-policy-ibm-cloud-managed.yaml
@@ -20,7 +20,7 @@ spec:
     # Consider only request to Gateway API CRDs.
     - name: "check-only-gateway-api-crds"
       # When the operation is DELETE, the "object" variable is null.
-      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io' || (request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.x-k8s.io'"
+      expression: "(request.operation == 'DELETE' ? oldObject : object).spec.group == 'gateway.networking.k8s.io'"
   # Validations are evaluated in the the order of their declaration.
   validations:
     # Verify that the request was sent by the ingress operator's service account.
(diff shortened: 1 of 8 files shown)


rikatz:
depends on https://github.com/openshift/origin/pull/30718 which is already merged

Miciah:
Thanks!  This appears to be a clean cherry-pick modulo some fuzz in the test code.  

/approve
/lgtm

This is a low-risk change that (1) relaxes the ValidatingAdmissionPolicy so that it doesn't block operations on resources in the gateway.networking.**x-k8s.io** API group and (2) stops the operator from reconciling CRDs in this API group, by removing them from the operator's index of CRDs that it manages.  Istio ignores these x-k8s.io resources, so allowing the cluster-admin to create them does not in any way affect the OpenShift gateway controller.  

/label backport-risk-assessed

rhamini3:
tested on cluster bot

```
% oc get clusterversion
version   4.21.0-0-2026-06-02-153646-test-ci-ln-29i8jgb-latest   True        False         5m24s   Cluster version is 4.21.0-0-2026-06-02-153646-test-ci-ln-29i8jgb-latest
iamin@iamin-mac gateway-api % oc apply -f config/crd/experimental/gateway.networking.x-k8s.io_xlistenersets.yaml 
customresourcedefinition.apiextensions.k8s.io/xlistenersets.gateway.networking.x-k8s.io created
iamin@iamin-mac gateway-api % oc apply -f config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml 
Error from server (Forbidden): error when creating "config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml": customresourcedefinitions.apiextensions.k8s.io "tlsroutes.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'openshift-ingress-operator-gatewayapi-crd-admission' with binding 'openshift-ingress-operator-gatewayapi-crd-admission' denied request: Gateway API Custom Resource Definitions are managed by the Ingress Operator and may not be modified
iamin@iamin-mac gateway-api % 
```
VAP blocks `gateway.networking.k8s.io` CRDs and allows `gateway.networking.x-k8s.io` CRDs
/verified by @rhamini3

PR #1451, opened 2026-05-21
NO-JIRA: Add testing guidelines for stability

rikatz:
This PR adds a testing guideline to be used by implementer and (human and AI) reviewers on e2e tests on CIO repository. These guidelines should be used to bring consistency and stability on e2e tests, avoiding silent error drop, flaky cleanups, bad fatal testing calls.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -20,4 +20,65 @@ reviews:
   path_filters:
   - "!**/vendor/**"
   - "!vendor/**"
+  path_instructions:
+  - path: "test/e2e/**"
+    instructions: |
+      Review e2e test code for these reliability patterns (see test/e2e/TESTING.md for full guide):
+
+      1. CLEANUP: Resource deletion MUST use `t.Cleanup(func() { deleteWithRetryOnError(t, context.Background(), obj, 2*time.Minute) })`.
+         Flag any of these anti-patterns:
+         - `defer func() { kclient.Delete(...) }()` — use `t.Cleanup` + `deleteWithRetryOnError` instead
+         - `defer assertDeleted(...)` or `t.Cleanup(func() { assertDeleted(...) })` — use `deleteWithRetryOnError` instead (assertDeleted fatals on transient errors)
+         - Inline `kclient.Delete` in cleanup without retry logic
+         - IngressController cleanup: use `t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ic) })` not `defer`
+
+      2. CLEANUP SEVERITY: `t.Fatalf()` MUST NOT be used inside `t.Cleanup()` or `defer` functions. Use `t.Errorf()` instead.
+         A fatal in cleanup calls runtime.Goexit(), masks the original failure, and prevents other cleanup functions from running.
+
+      3. RESOURCE UPDATES: Kubernetes resource mutations MUST use retry-on-conflict helpers that re-fetch before update:
+         - `updateIngressControllerWithRetryOnConflict()`
+         - `updateIngressConfigSpecWithRetryOnConflict()` / `updateIngressConfigStatusWithRetryOnConflict()`
+         - `updateRouteWithRetryOnConflict()`
+         Flag any pattern that does `Get()` then `Update()` without conflict retry.
+
+      4. TIMEOUTS: Controller condition polling (`waitForIngressControllerCondition`) MUST use at least `3*time.Minute`.
+         Flag any timeout under 1 minute for condition checks (e.g., `10*time.Second`).
+         Standard cleanup timeout is `2*time.Minute`.
+
+      5. ERROR WRAPPING: `fmt.Errorf` SHOULD use `%w` (not `%v`) when the error will be inspected with `errors.Is`/`errors.As`.
+
+      6. ERROR TYPE CHECKS: Delete operations MUST check `errors.IsNotFound` (not `IsAlreadyExists`).
+         Create operations MUST check `errors.IsAlreadyExists` (not `IsNotFound`).
+         Flag inverted error type guards — they silently swallow real errors.
+
+      7. NO UNBOUNDED POLLING: `wait.PollInfinite` MUST NOT be used — it hangs the test suite forever.
+         Prefer `wait.PollUntilContextTimeout` over deprecated `wait.PollImmediate` (context-aware, cancellable).
+
+      8. SHARED MUTABLE STATE: Package-level variables shared between parallel tests MUST use `atomic.Int32` or `sync.Mutex`.
+         Never mutate package-level globals (`dnsConfig`, `infraConfig`) from individual tests — read into local variables.
+
+      9. NIL SAFETY: Guard `infraConfig.Status.PlatformStatus` before accessing `.Type` — it can be nil on some platforms.
+         Flag `&value == nil` comparisons (address-of-value is always non-nil, this is dead code).
+         Validate slice length before indexing (e.g., `record.Spec.Targets[0]` without length check).
+
+      10. FUNCTION CONTRACTS: If a helper returns `error`, it MUST NOT call `t.Fatalf` internally (the error return becomes dead code).
+          If a helper calls `t.Fatalf`, it should not have an `error` return type.
+
+      11. DEFER IN LOOPS: `defer` inside a loop accumulates until function exit, causing resource leaks (file descriptors, log streams).
+          Close resources explicitly within each loop iteration or use `t.Cleanup`.
+
+      12. LOG ACCURACY: `t.Log` / `t.Logf` messages MUST accurately describe the operation being performed.
+          Flag log messages that contradict the code (e.g., logging "Classic" while setting NLB).
+
+      13. CONTEXT PROPAGATION: Inner polling loops MUST derive their context from the outer loop's context.
+          Flag `context.Background()` inside a poll callback when an outer context parameter is available.
+          Flag `defer cancel()` inside poll callbacks — it accumulates deferred calls across iterations.
+
+      14. TEST HELPERS: Functions accepting `*testing.T` that are not test functions MUST call `t.Helper()`.
+          Flag return values of `wait.PollUntilContextTimeout` that are silently discarded (not checked or returned).
+
+      15. RESOURCE LIFECYCLE: Resources MUST be created (with error check) BEFORE registering `t.Cleanup` for deletion.
+          Flag patterns where `buildEcho*`/`buildRoute*` is followed by `t.Cleanup` without an intervening `kclient.Create`.
+
+      See test/e2e/TESTING.md for the full patterns guide.
 
--- a/test/e2e/TESTING.md
+++ b/test/e2e/TESTING.md
@@ -1,0 +1,603 @@
+# E2E Testing Patterns — CIO Correctness Guidelines
+
+This document describes the required patterns for writing reliable e2e tests in this repository. These patterns prevent flaky test failures caused by transient Kubernetes API errors, resource version conflicts, cleanup ordering issues, race conditions, and logic bugs.
+
+All rules apply to new code **and** to modifications of existing code. When touching a test, fix any violations in the lines you modify.
+
+---
+
+## 1. Resource Cleanup
+
+### 1.1 Use `t.Cleanup()` instead of `defer`
+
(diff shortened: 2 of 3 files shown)


Thealisyed:
/retest

e2e-aws-operator:  AWS credentials weren't injected, installer hung on interactive prompt until 2h timeout.

 e2e-hypershift: break-glass-credentials/independent_signers CRR timed out waiting for cert revocation

.coderabbit.yaml
```diff
@@ -20,4 +20,65 @@ reviews:
...
+          Flag `context.Background()` inside a poll callback when an outer context parameter is available.
+          Flag `defer cancel()` inside poll callbacks — it accumulates deferred calls across iterations.
+
+      14. TEST HELPERS: Functions accepting `*testing.T` that are not test functions MUST call `t.Helper()`.
+          Flag return values of `wait.PollUntilContextTimeout` that are silently discarded (not checked or returned).
+
+      15. RESOURCE LIFECYCLE: Resources MUST be created (with error check) BEFORE registering `t.Cleanup` for deletion.
+          Flag patterns where `buildEcho*`/`buildRoute*` is followed by `t.Cleanup` without an intervening `kclient.Create`.
+
+      See test/e2e/TESTING.md for the full patterns guide.
```
Miciah, line 83: Is it necessary to summarize `test/e2e/TESTING.md` here in addition to referencing `TESTING.md`?  How do we keep the CodeRabbit summary and `TESTING.md` synchronized?

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+}
+```
+
+### 2.2 Never use `wait.PollInfinite` or unbounded polling
+
+Unbounded polling can hang the entire test suite if a resource gets stuck (finalizer, webhook, etc.). Always use `wait.PollUntilContextTimeout` or `wait.PollImmediate` with an explicit timeout:
+
+```go
+// CORRECT — bounded timeout
+wait.PollUntilContextTimeout(ctx, 5*time.Second, 2*time.Minute, true, func(ctx context.Context) (bool, error) {
```
Miciah, line 129: Where does `ctx` come from?  It might be useful to have some guidance on whether it is appropriate to use a new `context.Background()` for each polling loop, whether it is better to create a single context for the whole test, or to do something else.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+Package-level variables shared between parallel tests must use `sync/atomic` or `sync.Mutex`:
+
+```go
+// CORRECT — atomic counter shared between parallel tests
+var podCount atomic.Int32
+
+func testHeaders(t *testing.T, ...) {
+    count := podCount.Add(1)
+    podName := fmt.Sprintf("test-pod-%d", count)
+}
```
Miciah, line 205: This more verbose example might be clearer:
```suggestion
// testHeaders is a helper that may be used in in multiple tests.
func testHeaders(t *testing.T, ...) {
	t.Helper()

    count := podCount.Add(1)
    podName := fmt.Sprintf("test-pod-%d", count)
	...
}
```

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+### 4.2 Never mutate global config variables
+
+Package-level variables like `dnsConfig` and `infraConfig` are initialized in `TestMain` and shared across all tests. Never re-read into these globals from individual tests:
+
+```go
+// CORRECT — read into a local variable
+localDNS := configv1.DNS{}
+if err := kclient.Get(ctx, types.NamespacedName{Name: "cluster"}, &localDNS); err != nil { ... }
```
Miciah, line 223: This example is missing the mutation.  If you add the mutation, is this approach really correct?  It leaves the global `dnsConfig` inconsistent with the persisted value.  Maybe it would be better to update the global value, but protect it with a mutex.

Miciah: Maybe a mutex isn't even necessary; any test that mutates cluster config probably needs to be a serial test anyway.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+})
+```
+
+---
+
+## 5. Nil Safety
+
+### 5.1 Guard `PlatformStatus` before access
+
+`infraConfig.Status.PlatformStatus` can be nil on some platforms. Always guard:
```
Miciah, line 253: This is false.  See <https://bugzilla.redhat.com/show_bug.cgi?id=1890038#c26>.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+| Operation | Minimum Timeout | Rationale |
+|-----------|----------------|-----------|
+| Controller condition polling | `3*time.Minute` | Controllers need time to reconcile, especially under load |
+| Resource deletion cleanup | `2*time.Minute` | Allows for finalizer processing |
+| Load balancer readiness | `5*time.Minute` | Cloud provisioning is slow |
+| DNS resolution | `10*time.Minute` (`dnsResolutionTimeout`) | Propagation + negative cache TTL |
+| IBM Cloud DNS warmup | `7*time.Minute` | Platform-specific negative caching |
+| Quick API fetches | `1*time.Minute` | For simple Get/List operations |
+
+Never use `10*time.Second` or similar tiny timeouts for controller condition checks. These cause flaky failures under CI load.
```
Miciah, line 310: Short timeouts that cause flakiness is a recurring issue, so maybe it is worth elaborating on this point.  

Generally, a timeout should be *at least* twice the expected worst case.  For example, if provisioning a load balancer has been observed to take up to 5 minutes when CI is under load, make the timeout *at least* 10 minutes.  Using a longer timeout does not make the test take more time unless it actually reaches the timeout, and in the case of parallel tests, a long timeout doesn't necessarily increase the overall CI job run time at all.  In contrast, timing out early on an operation that would have succeeded with a longer timeout makes the test flakier and causes developers to rerun the CI job with the flaky test, which takes far more time and wastes far more resources than having a longer timeout.  

If there is a need to verify that an operation completes within a *specific* period of time, make a dedicated test *specifically* to check how long the operation takes.  For example, an Origin monitor test could verify that SLB provisioning time does not exceed some threshold based on historical data.  However, the purpose of operator E2E tests is not to verify that SLBs do not take too long to provision, and so these tests should use timeouts that are as long as necessary to avoid failure under load.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+
+---
+
+## 7. Error Wrapping
+
+Use `%w` (not `%v`) in `fmt.Errorf` when the returned error may be inspected with `errors.Is()` or `errors.As()`:
```
Miciah, line 316: Using `errors.Wrap` or `errors.Wrapf` is also fine.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+return fmt.Errorf("failed to update ingresscontroller: %v", err)
+```
+
+---
+
+## 8. Functions That Return Errors vs Call t.Fatal
+
+### 8.1 Consistent function contract
+
+If a helper function has an `error` return, it must NOT call `t.Fatalf` internally. Either return the error for the caller to handle, or call `t.Fatalf` and return nothing:
```
Miciah, line 332: It is overly complicated to have a helper that *can* call `t.Error` or `t.Fatal` *and also* can return an error value.  Better to do only one or only the other.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+// PREFERRED — unique per run
+name := names.SimpleNameGenerator.GenerateName("test-gateway-")
+
+// ACCEPTABLE — descriptive but ensure cleanup is robust
+name := "test-gateway-update"
+```
+
+### 9.2 Log messages must match the actual operation
+
+Verify that `t.Log` messages accurately describe what the code does. Misleading log messages make debugging flaky tests significantly harder:
```
Miciah, line 370: I would also suggest generally preferring log messages over code comments.  Maybe a reasonable guidelines would be this: Use logs to describe *what* each step of the test does, and use code comments as necessary to explain *why* the test is dong something.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+spec.LoadBalancer.Platform.AWS.Type = configv1.NLB
+```
+
+---
+
+## 10. Resource Lifecycle in Loops
+
+### 10.1 Never `defer` resource cleanup inside a loop
+
+`defer` in a loop accumulates until function exit. Use explicit cleanup or `t.Cleanup`:
```
Miciah, line 388: This is confusing.  How is using `t.Cleanup` better than using `defer` in the case of a loop?  

(Using `t.Cleanup` instead of `defer` for cleanups in general is good practice, of course.)

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// CORRECT — close in each iteration
+for _, pod := range pods.Items {
+    logs, _ := getLogStream(pod)
+    buf := new(bytes.Buffer)
+    buf.ReadFrom(logs)
+    logs.Close()
+    // ...
+}
```
Miciah, line 405: This is fine for simple cases, but for more involved situations, you might need something like this:
```go
for _, pod := range pods.Items {
    func() {
        logs, _ := getLogStream(pod)
        defer logs.Close()
        // The deferred cleanup will happen at the bottom of the loop.  
        // It is safe to `break` or to call `t.Fatal` from inside the loop.
        // ...
    }()
}

File: test/e2e/TESTING.md [hunk 12]
```diff
@@ -0,0 +1,603 @@
...
+    // ...
+})
+
+// WRONG — inner context is independent, won't cancel when outer times out
+deleteCtx, deleteCancel := context.WithTimeout(context.Background(), 2*time.Minute)
+```
+
+### 11.2 Call cancel explicitly instead of deferring inside poll callbacks
+
+`defer` inside a poll callback accumulates deferred calls across iterations (the callback is a function called repeatedly, but defers don't fire until the parent function returns). Call cancel explicitly after the inner poll completes:
```
Miciah, line 434: Are you sure about that?  That doesn't make sense to me; the callback is a func like any other, so any deferred funcs in the callback func should be called when the callback func returns.

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+defer deleteCancel()  // won't fire until parent function returns
+```
+
+---
+
+## 12. Test Helper Conventions
+
+### 12.1 Always call `t.Helper()` in test helper functions
+
+Every function that accepts `*testing.T` and is not a test itself must call `t.Helper()` at the top. This makes failure output point to the caller, not the helper:
```
Miciah, line 452: This is sometimes confusing though.  For a more involved helper, knowing the callsite of the helper might be useless; it is helpful to see the point of failure *within* the helper.  (Arguably, some of our helpers are overly complicated.)

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+---
+
+## 13. Test Structure
+
+### Parallel vs Serial
+
+- **Parallel tests** (~90): Use `t.Parallel()`, create their own `IngressController`, independent of each other
+- **Serial tests** (~50): Modify cluster-wide resources (default IC, cluster ingress config, infrastructure config)
```
Miciah, line 506: And each test must be registered as either a parallel test or a serial test in `all_tests.go` (we have `hack/verify-e2e-test-all-presence.sh` to verify this).

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+
+```go
+func TestSomething(t *testing.T) {
+    t.Parallel()
+
+    // 1. Create IngressController
+    ic := newPrivateController(name, domain)
+    if err := kclient.Create(context.TODO(), ic); err != nil {
+        t.Fatalf("failed to create ingresscontroller: %v", err)
+    }
```
Miciah, line 526: Should we promote use of `createWithRetryOnError` in this example?

test/e2e/TESTING.md
```diff
@@ -0,0 +1,603 @@
...
+    t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ic) })
+
+    // 2. Wait for readiness
+    if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name, conditions...); err != nil {
+        t.Fatalf("failed to observe expected conditions: %v", err)
+    }
+
+    // 3. Create test workload (pod + service + route)
```
Miciah, line 534: The code comments are great examples of code comments that could be log messages.

PR #1452, opened 2026-05-21
RFE-2195: expose ROUTER_METRICS_HAPROXY_SERVER_THRESHOLD value via IngressController API

himank98:
## Summary

`ROUTER_METRICS_HAPROXY_SERVER_THRESHOLD` controls the maximum number of HAProxy backends (servers) per shard for which per-server Prometheus metrics  are collected. When the number of active backends exceeds this threshold,
per-server metrics are suppressed to prevent high-cardinality metric explosions in the monitoring stack.

Previously this variable was hardcoded to a default of 500 inside the router image with no way to tune it via the IngressController API. This PR exposes it as `spec.tuningOptions.haproxyServerThreshold` so cluster administrators can adjust the value — for example, raising it to 1000 to match the typical number of routes per shard and improve metrics coverage in large clusters.

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -149,6 +149,12 @@ const (
 	RouterBackendCheckInterval = "ROUTER_BACKEND_CHECK_INTERVAL"
 	RouterTLSCurves            = "ROUTER_CURVES"
 
+	// RouterMetricsHAProxyServerThreshold is the env var that controls the maximum
+	// number of servers (backends) per HAProxy process for which per-server Prometheus
+	// metrics are reported. Exceeding this threshold disables per-server metrics to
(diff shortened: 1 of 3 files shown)


gcs278:
@himank98 thanks for the contribution!

There are a couple things about this PR:

- This PR is directly modifying the vendored IngressController API - it's required to add all API changes via openshift/api, then vendor those changes here via the go.mod. We can't accept directly modified vendor files.
- We are unable to accept new features tied directly to RFEs without an enhancement proposal. This [doc](https://github.com/openshift/enhancements/blob/master/dev-guide/feature-zero-to-hero.md) is a good place to start with the process.

My suggestion - if you'd like to drive this forward, please work with our team's stakeholders to get a feature created and open an enhancement PR.

PR #1457, opened 2026-06-02
OCPBUGS-36723: Add missing include annotations to IBM Cloud and PowerVS ingress CredentialsRequests

asadawar:
## Summary

- Add `include.release.openshift.io` annotations and `controller-tools.k8s.io` label to IBM Cloud and PowerVS CredentialsRequest documents in `manifests/00-ingress-credentials-request.yaml`
- Matches the annotations already present on the AWS, Azure, and GCP documents in the same file

## Why this approach

The ingress credentials request file contains 5 YAML documents (one per cloud provider). The AWS, Azure, and GCP documents have:

```yaml
labels:
  controller-tools.k8s.io: "1.0"
annotations:
  capability.openshift.io/name: CloudCredential+Ingress
  include.release.openshift.io/ibm-cloud-managed: "true"
```

The IBM Cloud and PowerVS documents only have the `capability.openshift.io/name` annotation.

When `oc adm release extract --credentials-requests --included` runs, it checks each document's `include.release.openshift.io/self-managed-high-availability` annotation. Documents without it are excluded with:

```
Excluding CredentialsRequest "openshift-ingress-ibmcloud": include.release.openshift.io/self-managed-high-availability unset
```

This means every IBM Cloud and PowerVS IPI installation using Manual credentials mode is missing the ingress credential and must work around it by extracting without `--included` and manually copying the file.

## Verification

Confirmed the filtering behavior using `oc adm release extract -v=4`:
- With this fix, IBM Cloud and PowerVS ingress CRs pass the `--included` filter
- Without this fix, they are excluded, producing only 4 of the required 5 credential secrets

Bug: https://issues.redhat.com/browse/OCPBUGS-36723

### DIFF
--- a/manifests/00-ingress-credentials-request.yaml
+++ b/manifests/00-ingress-credentials-request.yaml
@@ -90,10 +90,15 @@ spec:
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
   name: openshift-ingress-ibmcloud
   namespace: openshift-cloud-credential-operator
   annotations:
     capability.openshift.io/name: CloudCredential+Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
 spec:
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
@@ -120,10 +125,15 @@ spec:
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
   name: openshift-ingress-powervs
   namespace: openshift-cloud-credential-operator
   annotations:
     capability.openshift.io/name: CloudCredential+Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
 spec:
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1


asadawar:
## Verification

Tested the fix locally against the filtering logic that `oc adm release extract --included` uses.

**Before (4.22.0-rc.5 release payload):**

```
$ oc adm release extract --credentials-requests --included --cloud=ibmcloud -v=4 2>&1 | grep ingress
Excluding CredentialsRequest "openshift-ingress-ibmcloud": include.release.openshift.io/self-managed-high-availability unset
Excluding CredentialsRequest "openshift-ingress-powervs": include.release.openshift.io/self-managed-high-availability unset

$ ls /tmp/creds/
# 0 files (4.22), 4 files without ingress (4.21)
```

**After (with this fix applied):**

```
Annotation check on fixed manifest:

Document 0: openshift-ingress (AWSProviderSpec)           - PASS (all annotations present)
Document 1: openshift-ingress-azure (AzureProviderSpec)   - PASS (all annotations present)
Document 2: openshift-ingress-gcp (GCPProviderSpec)       - PASS (all annotations present)
Document 3: openshift-ingress-ibmcloud (IBMCloudProviderSpec)   - PASS (all annotations present)
Document 4: openshift-ingress-powervs (IBMCloudPowerVSProviderSpec) - PASS (all annotations present)

--included filter simulation: 5/5 documents pass
--included + --cloud=ibmcloud: extracts openshift-ingress-ibmcloud correctly
```

**Unit tests:** `make test` passes with no regressions.

**How the root cause was found:**

Running `oc adm release extract -v=4` showed the exact exclusion reason. The IBM Cloud and PowerVS documents in this file were the only ones missing `include.release.openshift.io/self-managed-high-availability`. The `--included` filter in `oc` (via `library-go/pkg/manifest/manifest.go:Include()`) checks this annotation at line 355 and excludes documents without it. AWS, Azure, and GCP documents in the same file pass because they have all three `include.release.openshift.io` annotations.

Tested against both 4.21.17 and 4.22.0-rc.5 release payloads. Both exhibit the same behavior.

manifests/00-ingress-credentials-request.yaml
```diff
@@ -90,10 +90,15 @@ spec:
 apiVersion: cloudcredential.openshift.io/v1
 kind: CredentialsRequest
 metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
```
Miciah, line 94: From what I can find, old versions of controller-tools added this label to denote generated files, and the label seems to have been quietly deprecated.  See for example https://github.com/kubernetes-sigs/kubebuilder/issues/764.  (I cannot find an official deprecation notice.)  As the label is deprecated, and as this file was not generated using controller-tools, it doesn't make sense to add it here.  It should be harmless though.

manifests/00-ingress-credentials-request.yaml
```diff
@@ -90,10 +90,15 @@ spec:
...
   name: openshift-ingress-ibmcloud
   namespace: openshift-cloud-credential-operator
   annotations:
     capability.openshift.io/name: CloudCredential+Ingress
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
```
Miciah, line 101: `single-node-developer` and `self-managed-high-availability` shouldn't be necessary, but they should be harmless.

Miciah:
Thanks!

/approve
/lgtm

asadawar:
@Miciah All 18 CI jobs are passing. Could you please add `/ok-to-test` so the PR is eligible for merge? (It will retrigger CI, but everything passed cleanly on the last run.)

The fix has also been verified locally as described in the PR description (tested with `oc adm release extract --credentials-requests --included --cloud=ibmcloud` confirming the ingress CredentialsRequests are now correctly included). A `/verified` would be appreciated as well.

Thanks!

rhamini3:
confirmed that the oc adm runs successfully and ingress co is not degraded, aligning with @asadawar 

```
% oc adm release extract --credentials-requests --included --cloud=ibmcloud --to=/tmp/creds
Extracted release payload created at 2026-06-04T18:09:49Z
iamin@iamin-mac cluster-ingress-operator % oc adm release extract --credentials-requests --cloud=ibmcloud --to=/tmp/creds
warning: if you intend to pass CredentialsRequests to ccoctl, you should use --included to filter out requests that your cluster is not expected to need.
Extracted release payload created at 2026-06-04T18:09:49Z

 % oc get co ingress
NAME      VERSION                                               AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   5.0.0-0-2026-06-04-180941-test-ci-ln-wbsxh62-latest   True        False         False       90m 
```
/verified by @rhamini3

PR #1467, opened 2026-06-10
OCPBUGS-86841: Add BackendTLSPolicy and ReferenceGrant e2e test coverage helpers

Thealisyed:
## Summary

- Add `assertAllGatewayAPICRDsCovered` — lists all Gateway API CRDs on the
  cluster and verifies each is in the e2e `crdNames` list, catching cases where
  a new CRD is added to the operator but not to the test suite
- Add `assertClusterRoleCoversGatewayAPICRDs` — verifies the aggregated
  ClusterRoles have resource entries for all namespace-scoped Gateway API CRDs,
  catching RBAC gaps like OCPBUGS-85690
- Add `createBackendTLSPolicy` — creates a BackendTLSPolicy CR instance and
  verifies the API server accepts and persists it
- Add `createReferenceGrant` — creates a ReferenceGrant CR instance using
  unstructured objects (v1beta1 not vendored), with dynamic version lookup via
  `referenceGrantStorageVersion()` to handle API graduation

## Context

PR #1523 (merged) added `backendtlspolicies` to `crdNames` so the CRD is
exercised by the existing lifecycle and VAP protection tests. This PR extends
coverage by:

1. **Reverse coverage guardrail** — `assertAllGatewayAPICRDsCovered` prevents
   future CRDs from silently missing test coverage
2. **RBAC guardrail** — `assertClusterRoleCoversGatewayAPICRDs` prevents future
   RBAC gaps (catches the class of bug fixed in OCPBUGS-85690)
3. **CR creation validation** — `createBackendTLSPolicy` and `createReferenceGrant`
   are called from `ensureGatewayObjectCreation` to verify the CRD schemas are
   functional and objects can be persisted end-to-end

A full operational BackendTLSPolicy traffic test (TLS backend + actual traffic
through the gateway) is tracked as a follow-up in OCPBUGS-99912.

### DIFF
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -158,9 +158,7 @@ func testGatewayAPIResources(t *testing.T) {
 	ensureCRDs(t)
 
 	// Verify that all Gateway API CRDs on the cluster are in our test list.
-	if err := assertAllGatewayAPICRDsCovered(t, crdNames); err != nil {
-		t.Errorf("CRD coverage gap: %v", err)
-	}
+	assertAllGatewayAPICRDsCovered(t, crdNames)
 
 	// Deleting CRDs to ensure they gets recreated again
 	bypassVAP(t, deleteCRDs)
@@ -1653,31 +1651,11 @@ func ensureGatewayObjectCreation(t *testing.T, ns *corev1.Namespace) error {
 	}
 	// The http route is cleaned up when the namespace is deleted.
 
-	// Create BackendTLSPolicy and ReferenceGrant if their CRDs are available.
-	// These CRDs are installed by the operator but not reconciled, so this
-	// validates the CRD schemas are functional and objects can be persisted.
-	// The CRD existence check allows safe backporting to older branches.
-	btlsCRD := &apiextensionsv1.CustomResourceDefinition{}
-	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "backendtlspolicies.gateway.networking.k8s.io"}, btlsCRD); kerrors.IsNotFound(err) {
-		t.Log("BackendTLSPolicy CRD not found, skipping creation test")
-	} else if err != nil {
-		return fmt.Errorf("failed to check for BackendTLSPolicy CRD: %w", err)
-	} else {
-		if err := createBackendTLSPolicy(t, ns); err != nil {
-			return fmt.Errorf("BackendTLSPolicy object could not be created: %w", err)
-		}
-	}
-
-	refGrantCRD := &apiextensionsv1.CustomResourceDefinition{}
-	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "referencegrants.gateway.networking.k8s.io"}, refGrantCRD); kerrors.IsNotFound(err) {
-		t.Log("ReferenceGrant CRD not found, skipping creation test")
-	} else if err != nil {
-		return fmt.Errorf("failed to check for ReferenceGrant CRD: %w", err)
-	} else {
-		if err := createReferenceGrant(t, ns); err != nil {
-			return fmt.Errorf("ReferenceGrant object could not be created: %w", err)
-		}
-	}
+	// Create BackendTLSPolicy and ReferenceGrant CR instances. The operator
+	// installs these CRDs but does not reconcile them, so this validates
+	// the CRD schemas are functional and objects can be persisted.
+	createBackendTLSPolicy(t, ns)
+	createReferenceGrant(t, ns)
 
 	return nil
 }
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -157,6 +157,9 @@ func testGatewayAPIResources(t *testing.T) {
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
 	ensureCRDs(t)
 
+	// Verify that all Gateway API CRDs on the cluster are in our test list.
+	assertAllGatewayAPICRDsCovered(t, crdNames)
+
 	// Deleting CRDs to ensure they gets recreated again
 	bypassVAP(t, deleteCRDs)
 
@@ -557,6 +560,9 @@ func testGatewayAPIRBAC(t *testing.T) {
 	}
 
 	for srcClusterRoleName, destClusterRoleNames := range aggregationMapping {
+		t.Logf("verifying that ClusterRole %s covers all namespaced Gateway API CRDs", srcClusterRoleName)
+		assertClusterRoleCoversGatewayAPICRDs(t, srcClusterRoleName)
+
 		for _, destClusterRoleName := range destClusterRoleNames {
 			t.Logf("verifying that ClusterRole %s aggregates all PolicyRules from %s", destClusterRoleName, srcClusterRoleName)
 
@@ -1644,6 +1650,12 @@ func ensureGatewayObjectCreation(t *testing.T, ns *corev1.Namespace) error {
 		return fmt.Errorf("feature gate was enabled, but http route object could not be created: %v", err)
 	}
 	// The http route is cleaned up when the namespace is deleted.
+
+	// Create BackendTLSPolicy and ReferenceGrant CR instances. The operator
+	// installs these CRDs but does not reconcile them, so this validates
+	// the CRD schemas are functional and objects can be persisted.
+	createBackendTLSPolicy(t, ns)
+	createReferenceGrant(t, ns)
 
 	return nil
 }
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -125,6 +125,7 @@ func TestGatewayAPI(t *testing.T) {
 	})
 
 	t.Run("testGatewayAPIResources", testGatewayAPIResources)
+	t.Run("testGatewayAPICRDValidation", testGatewayAPICRDValidation)
 	t.Run("testGatewayAPIObjects", testGatewayAPIObjects)
 	t.Run("testGatewayAPIManualDeployment", testGatewayAPIManualDeployment)
 	if gatewayAPIWithoutOLMEnabled {
@@ -284,6 +285,21 @@ func testGatewayAPIIstioUninstallSailLibrary(t *testing.T) {
 	if err := assertIstioCRDs(t); err != nil {
 		t.Fatalf("Istio CRDs should persist after uninstall: %v", err)
 	}
+}
+
+// testGatewayAPICRDValidation tests that BackendTLSPolicy and ReferenceGrant
+// custom resources can be created and are accepted by the API server. These
+// CRDs are installed by the operator but not reconciled, so this validates
+// the CRD schemas are functional.
+func testGatewayAPICRDValidation(t *testing.T) {
+	ns := createNamespace(t, names.SimpleNameGenerator.GenerateName("test-e2e-gwapi-crd-"))
+
+	t.Run("BackendTLSPolicy", func(t *testing.T) {
+		testBackendTLSPolicyCreation(t, ns)
+	})
+	t.Run("ReferenceGrant", func(t *testing.T) {
+		testReferenceGrantCreation(t, ns)
+	})
 }
 
 // testGatewayAPIObjects tests that Gateway API objects can be created successfully.
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -157,6 +157,9 @@ func testGatewayAPIResources(t *testing.T) {
 	// Make sure all the *.gateway.networking.k8s.io CRDs are available since the FeatureGate is enabled.
 	ensureCRDs(t)
 
+	// Verify that all Gateway API CRDs on the cluster are in our test list.
+	assertAllGatewayAPICRDsCovered(t, crdNames)
+
 	// Deleting CRDs to ensure they gets recreated again
 	bypassVAP(t, deleteCRDs)
 
@@ -557,6 +560,9 @@ func testGatewayAPIRBAC(t *testing.T) {
 	}
 
 	for srcClusterRoleName, destClusterRoleNames := range aggregationMapping {
+		t.Logf("verifying that ClusterRole %s covers all namespaced Gateway API CRDs", srcClusterRoleName)
+		assertClusterRoleCoversGatewayAPICRDs(t, srcClusterRoleName)
+
 		for _, destClusterRoleName := range destClusterRoleNames {
 			t.Logf("verifying that ClusterRole %s aggregates all PolicyRules from %s", destClusterRoleName, srcClusterRoleName)
 
--- a/test/e2e/gateway_api_test.go
+++ b/test/e2e/gateway_api_test.go
@@ -125,7 +125,6 @@ func TestGatewayAPI(t *testing.T) {
 	})
 
 	t.Run("testGatewayAPIResources", testGatewayAPIResources)
-	t.Run("testGatewayAPICRDValidation", testGatewayAPICRDValidation)
 	t.Run("testGatewayAPIObjects", testGatewayAPIObjects)
 	t.Run("testGatewayAPIManualDeployment", testGatewayAPIManualDeployment)
 	if gatewayAPIWithoutOLMEnabled {
@@ -159,7 +158,9 @@ func testGatewayAPIResources(t *testing.T) {
 	ensureCRDs(t)
 
 	// Verify that all Gateway API CRDs on the cluster are in our test list.
-	assertAllGatewayAPICRDsCovered(t, crdNames)
+	if err := assertAllGatewayAPICRDsCovered(t, crdNames); err != nil {
+		t.Errorf("CRD coverage gap: %v", err)
+	}
 
 	// Deleting CRDs to ensure they gets recreated again
 	bypassVAP(t, deleteCRDs)
@@ -285,21 +286,6 @@ func testGatewayAPIIstioUninstallSailLibrary(t *testing.T) {
 	if err := assertIstioCRDs(t); err != nil {
 		t.Fatalf("Istio CRDs should persist after uninstall: %v", err)
 	}
-}
-
-// testGatewayAPICRDValidation tests that BackendTLSPolicy and ReferenceGrant
-// custom resources can be created and are accepted by the API server. These
-// CRDs are installed by the operator but not reconciled, so this validates
-// the CRD schemas are functional.
-func testGatewayAPICRDValidation(t *testing.T) {
-	ns := createNamespace(t, names.SimpleNameGenerator.GenerateName("test-e2e-gwapi-crd-"))
-
-	t.Run("BackendTLSPolicy", func(t *testing.T) {
-		testBackendTLSPolicyCreation(t, ns)
-	})
-	t.Run("ReferenceGrant", func(t *testing.T) {
-		testReferenceGrantCreation(t, ns)
-	})
 }
 
 // testGatewayAPIObjects tests that Gateway API objects can be created successfully.
@@ -1666,6 +1652,32 @@ func ensureGatewayObjectCreation(t *testing.T, ns *corev1.Namespace) error {
 		return fmt.Errorf("feature gate was enabled, but http route object could not be created: %v", err)
 	}
 	// The http route is cleaned up when the namespace is deleted.
+
+	// Create BackendTLSPolicy and ReferenceGrant if their CRDs are available.
+	// These CRDs are installed by the operator but not reconciled, so this
+	// validates the CRD schemas are functional and objects can be persisted.
+	// The CRD existence check allows safe backporting to older branches.
+	btlsCRD := &apiextensionsv1.CustomResourceDefinition{}
+	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "backendtlspolicies.gateway.networking.k8s.io"}, btlsCRD); kerrors.IsNotFound(err) {
+		t.Log("BackendTLSPolicy CRD not found, skipping creation test")
+	} else if err != nil {
+		return fmt.Errorf("failed to check for BackendTLSPolicy CRD: %w", err)
+	} else {
+		if err := createBackendTLSPolicy(t, ns); err != nil {
+			return fmt.Errorf("BackendTLSPolicy object could not be created: %w", err)
+		}
+	}
+
+	refGrantCRD := &apiextensionsv1.CustomResourceDefinition{}
+	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "referencegrants.gateway.networking.k8s.io"}, refGrantCRD); kerrors.IsNotFound(err) {
+		t.Log("ReferenceGrant CRD not found, skipping creation test")
+	} else if err != nil {
+		return fmt.Errorf("failed to check for ReferenceGrant CRD: %w", err)
+	} else {
+		if err := createReferenceGrant(t, ns); err != nil {
+			return fmt.Errorf("ReferenceGrant object could not be created: %w", err)
+		}
+	}
 
 	return nil
 }
--- a/test/e2e/util_gatewayapi_test.go
+++ b/test/e2e/util_gatewayapi_test.go
@@ -33,6 +33,9 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -1591,12 +1594,10 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
 	})
 }
 
-func assertAllGatewayAPICRDsCovered(t *testing.T, wantCRDs []string) error {
+func assertAllGatewayAPICRDsCovered(t *testing.T, wantCRDs []string) {
 	t.Helper()
 	haveCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
-	if err := kclient.List(t.Context(), haveCRDs); err != nil {
-		t.Fatalf("failed to list CRDs: %v", err)
-	}
+	require.NoError(t, kclient.List(t.Context(), haveCRDs), "failed to list CRDs")
 	var extraCRDs []string
 	for _, crd := range haveCRDs.Items {
 		if crd.Spec.Group == "gateway.networking.k8s.io" {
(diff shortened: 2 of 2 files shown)


Thealisyed:
Infra failures unrelated to our changes

/retest

test/e2e/gateway_api_test.go
rikatz, line 60: I think we should probably have a test as well that verifies if every found CRD on "gateway.networking.k8s.io" on the cluster exists on this list.

Something like 

```
* list all CRDs, filter by the ones on group "gateway.networking.k8s.io"
* for every item on this list: assert.Contains(t, crdNames, crdName)
```

This way we will get also the reverse case: did we added a new CRD, but never tested it?

rikatz: which also makes me think:
- we can separate the crd list here on the categories 'admin crds' and 'user crds'
- we can check during the test if our RBAC for each case contains exactly the same list here (admin must have access to all 'gateway.networking.k8s.io' as defined in https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml so doing a test that does a "kubectl get clusterrole system:openshift:gateway-api:aggregate-to-admin" and check if the existing roles here matches the installed CRDs
- Same thing for user, but with the limited access to its APIs

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1529,60 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+	for _, crd := range crdList.Items {
+		if crd.Spec.Group != "gateway.networking.k8s.io" {
+			continue
+		}
+		if crd.Spec.Scope == apiextensionsv1.ClusterScoped {
+			continue
+		}
+		plural := crd.Spec.Names.Plural
+		if !slices.Contains(roleResources, plural) {
+			t.Errorf("ClusterRole %s missing rule for Gateway API resource %q (CRD %s)", clusterRoleName, plural, crd.Name)
```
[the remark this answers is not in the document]
Thealisyed: If someone adds "*" to a Gateway API ClusterRole rule, we'd want this test to flag it for investigation rather than silently accepting it no? Since the aggregated ClusterRole lists each resource by name, wildcards would bypass these checks we are enforcing.. what do you think? @coderabbitai

feel free to push back or discuss this more with me!

test/e2e/gateway_api_test.go
Thealisyed: I was actually talking to Grant about this (prior to seeing your comment) xD

Thealisyed: I've added the following now:
1. `assertAllGatewayAPICRDsCovered` — lists all CRDs on the cluster with group `gateway.networking.k8s.io` and asserts each is in `crdNames`. Called in `testGatewayAPIResources` after `ensureCRDs`.
2. assertClusterRoleCoversGatewayAPICRDs — gets the ClusterRole from the cluster, collects its `gateway.networking.k8s.io` resources, and checks each namespace scoped Gateway API CRD is covered.

- Instead of separate "admin" and "user" CRD lists, the RBAC check queries the cluster and skips cluster scoped CRD's automatically thus no hardcoded lists to maintain.

gcs278:
I think we forgot to assign some PRs after PR Scrub

/assign @candita

gcs278:
🔔 Automated NID PR Reminder — High Priority

• Open for: 29 days (since Jun 10)
• Primary Reviewer: @candita
• Secondary Reviewer: @rikatz
• Last human activity: Jun 23 (15 days ago, @gcs278 commented)
• Waiting on: Author (@Thealisyed) — reviewer commented 15 days ago with no author response

No lgtm or approved labels yet.

---
_This is an experimental automated reminder. Contact @gcs278 if there are any issues._

test/e2e/util_gatewayapi_test.go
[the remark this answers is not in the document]
candita: @Thealisyed the failures on wildcards would be good to have in the godoc/comments.

candita:
@Thealisyed the idea behind OCPBUGS-86841 was to try to create the BackendTLSPolicy and ReferenceGrant within the existing e2e suite and make sure everything passes.  Start with `ensureGatewayObjectCreation` - create the BTLS policy and then ensure it works as described.  Even though we do this in the conformance testing, we are not running conformance testing as an e2e step or required PR CI step and may not have signal on all branches or all platforms.

Note that `ReferenceGrant` should have been a part of this work, not just `BackendTLSPolicy`.

Also see https://redhat.atlassian.net/browse/OCPBUGS-86841?focusedCommentId=17230208 

/hold for updates.

Thealisyed:
[Note to self](https://github.com/openshift/cluster-ingress-operator/pull/1482#issuecomment-4996615012)

test/e2e/gateway_api_test.go
```diff
@@ -270,6 +275,21 @@ func testGatewayAPIIstioUninstallSailLibrary(t *testing.T) {
...
+// the CRD schemas are functional.
+func testGatewayAPICRDValidation(t *testing.T) {
+	ns := createNamespace(t, names.SimpleNameGenerator.GenerateName("test-e2e-gwapi-crd-"))
+
+	t.Run("BackendTLSPolicy", func(t *testing.T) {
+		testBackendTLSPolicyCreation(t, ns)
+	})
+	t.Run("ReferenceGrant", func(t *testing.T) {
+		testReferenceGrantCreation(t, ns)
+	})
```
candita: @Thealisyed these should be included in the `ensureGatewayObjectCreation` function with all the other crd tests.  You should use the namespace created there so that these objects get cleaned up when that namespace is deleted.

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,60 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+		t.Fatalf("failed to list CRDs: %v", err)
+	}
+	for _, crd := range crdList.Items {
+		if crd.Spec.Group == "gateway.networking.k8s.io" {
+			if !slices.Contains(crdNames, crd.Name) {
+				t.Errorf("Gateway API CRD %q exists on cluster but is not in crdNames; add it to ensure test coverage", crd.Name)
+			}
+		}
+	}
+}
```
candita: I suggest collecting any extra CRDs and reporting an error if there are any.  Something like this (untested):

```suggestion
func assertAllGatewayAPICRDsCovered(t *testing.T, wantCrds []string) error {
	t.Helper()
	haveCrds := &apiextensionsv1.CustomResourceDefinitionList{}
	extraCrds := []string{}
	
	if err := kclient.List(context.Background(), haveCrds); err != nil {
		t.Fatalf("failed to list CRDs: %v", err)
	}
	for _, crd := range haveCrds.Items {
		if crd.Spec.Group == "gateway.networking.k8s.io" {
			if !slices.Contains(wantCrds, crd.Name) {
			    extraCrds = append(extraCrds, crd.Name)
			}
		}
	}
   if len(extraCrds) > 0 {
			return fmt.Errorf("Extra Gateway API CRDs were installed but are not being tested: %q", extraCrds)
   }
   return nil
}
```

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,60 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := kclient.List(context.Background(), crdList); err != nil {
+		t.Fatalf("failed to list CRDs: %v", err)
+	}
+	for _, crd := range crdList.Items {
+		if crd.Spec.Group != "gateway.networking.k8s.io" {
+			continue
+		}
+		if crd.Spec.Scope == apiextensionsv1.ClusterScoped {
```
candita: Can you add a comment on why you check the scope?

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,60 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("failed to get ClusterRole %s: %v", clusterRoleName, err)
+	}
+
+	var roleResources []string
+	for _, rule := range clusterRole.Rules {
+		if slices.Contains(rule.APIGroups, "gateway.networking.k8s.io") {
+			roleResources = append(roleResources, rule.Resources...)
```
candita: Do we care about the subresources here; should we exclude them?  E.g. gatewayclasses/status and gateways/status.

Also, there should be no wildcards under resources, so you could either comment that we would never expect that, or else emit a fatal error when you create the roleResources if it contains a wildcard under resources.

test/e2e/gateway_api_test.go
candita, line 1: Please add skips or conditions if the Gateway API version is less than 1.5 so these files can still be backported and run without breaking. 

If there's time, or it it's easier,  splitting out any 1.5 CRD tests into their own file (e.g. gateway_api_1_5_test.go) would be preferred.

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1617,140 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+				}},
+				Hostname: gatewayapiv1.PreciseHostname(hostname),
+			},
+		},
+	}
+}
+
+func testBackendTLSPolicyCreation(t *testing.T, ns *corev1.Namespace) {
+	t.Helper()
+	ctx := context.Background()
```
candita: Should this be `t.Context`?

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1617,140 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+			Ports: []corev1.ServicePort{{
+				Name:     "https",
+				Port:     443,
+				Protocol: corev1.ProtocolTCP,
+			}},
+			Selector: map[string]string{"app": "btlspolicy-target"},
+		},
+	}
+	if err := createWithRetryOnError(t, ctx, svc, DefaultRetryTimeout); err != nil {
+		t.Fatalf("failed to create target service: %v", err)
```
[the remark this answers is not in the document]
candita: @Thealisyed please fix or address why you don't need to.

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1617,140 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+		invalidPolicy := buildBackendTLSPolicy("test-btlspolicy-invalid", ns.Name, svc.Name, cm.Name, "backend.example.com")
+		wellKnown := gatewayapiv1.WellKnownCACertificatesSystem
+		invalidPolicy.Spec.Validation.WellKnownCACertificates = &wellKnown
+
+		err := kclient.Create(ctx, invalidPolicy)
+		if err == nil {
+			t.Fatal("expected CEL validation to reject BackendTLSPolicy with both CACertificateRefs and WellKnownCACertificates, but creation succeeded")
+		}
+		t.Logf("CEL validation correctly rejected invalid BackendTLSPolicy: %v", err)
+	})
```
candita: We probably don't need the explicit CEL check, but it would be good to have a test of the operational aspect of BackendTLSPolicy

test/e2e/gateway_api_test.go
Thealisyed: We used CRD existence checks instead of a file split since the code moved into `ensureGatewayObjectCreation` per the other request.

rikatz: I am trying to balance here between @candita request and your answer, so let me see if we can find a common ground:

I agree this PR makes it harder to be backported (and in fact, I have asked that the e2e tests were split on a different PR on https://github.com/openshift/cluster-ingress-operator/pull/1482#issuecomment-4996615012) and while I would not consider this a big deal 3 months ago, now that we are considering backporting Gateway API as well I think it is important that we keep things separated even if this adds slightly more work for us.

So:
* split the e2e tests changes on a different PR that should be backported after https://github.com/openshift/cluster-ingress-operator/pull/1482 so we get a state where "1 - the missing grpcroutes is fixed" -> "2 - and we are testing in case we start backporting Gateway API CRDs"
* split this addition to be only backendtlspolicies (I know I have asked you to add the tests, my bad :) but we need to make the state for backporting clean)

I think that splitting into their own test files is probably too much as they will properly belong to the backport process (eg.: the moment we backport TLSRoute, the e2e tests should be running and the Gateway API backport for 4.20 should also add this to crdNames as it is being done on its own PR already)

test/e2e/gateway_api_test.go
```diff
@@ -1532,6 +1541,32 @@ func ensureGatewayObjectCreation(t *testing.T, ns *corev1.Namespace) error {
 	}
 	// The http route is cleaned up when the namespace is deleted.
 
+	// Create BackendTLSPolicy and ReferenceGrant if their CRDs are available.
+	// These CRDs are installed by the operator but not reconciled, so this
+	// validates the CRD schemas are functional and objects can be persisted.
+	// The CRD existence check allows safe backporting to older branches.
+	btlsCRD := &apiextensionsv1.CustomResourceDefinition{}
```
rikatz: so, I think that at the moment we add referenceGrants and BackendTLSPolicies to the crdNames, there is no such thing as "if they exist". I think it should belong to the process of adding a new CRD trying to create it, and a failure here means "something went wrong on the CRD installation"

test/e2e/gateway_api_test.go
```diff
@@ -1532,6 +1541,32 @@ func ensureGatewayObjectCreation(t *testing.T, ns *corev1.Namespace) error {
...
+	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "backendtlspolicies.gateway.networking.k8s.io"}, btlsCRD); kerrors.IsNotFound(err) {
+		t.Log("BackendTLSPolicy CRD not found, skipping creation test")
+	} else if err != nil {
+		return fmt.Errorf("failed to check for BackendTLSPolicy CRD: %w", err)
```
rikatz: I don't really understand why we do a return of error in this function instead of calling t.Error and marking the test as failed. In fact, the caller of this test calls a t.Fatal over the error return, which is the worst case scenario as it may fully stop all of the tests.

So, for your e2e PR, it is worth to fix this as well, make this function stop returning an error, and in case an error is found do a t.Error (a t.Fatal doesn't even give a chance for cleanup functions)

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,76 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
 	})
 }
 
+func assertAllGatewayAPICRDsCovered(t *testing.T, wantCRDs []string) error {
```
rikatz: same comment on the error here. On `assertClusterRoleCoversGatewayAPICRDs` you do not return an error and instead call a t.Fatal (oops) inside it, why not calling a t.Error here directly? You do not return anything anyway from here, so maybe remove the error.

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,76 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+	t.Helper()
+	haveCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := kclient.List(t.Context(), haveCRDs); err != nil {
+		t.Fatalf("failed to list CRDs: %v", err)
```
rikatz: I would like us to start using less the t.Something() and more the testify library. So if possible, on your e2e PR, do a "require.NoError(t, err)" here and everywhere

rikatz: (otherwise please do not use t.Fatal)

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,76 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+	var extraCRDs []string
+	for _, crd := range haveCRDs.Items {
+		if crd.Spec.Group == "gateway.networking.k8s.io" {
+			if !slices.Contains(wantCRDs, crd.Name) {
+				extraCRDs = append(extraCRDs, crd.Name)
+			}
+		}
+	}
+	if len(extraCRDs) > 0 {
+		return fmt.Errorf("Gateway API CRDs installed but not tested: %q", extraCRDs)
```
rikatz: similar comment from above, these 3 lines could be changed to assert.Len(t, extraCRDs, 0)

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,76 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+
+	var clusterRole rbacv1.ClusterRole
+	if err := wait.PollUntilContextTimeout(context.Background(), 2*time.Second, 30*time.Second, false, func(ctx context.Context) (bool, error) {
+		if err := kclient.Get(ctx, types.NamespacedName{Name: clusterRoleName}, &clusterRole); err != nil {
+			t.Logf("failed to get ClusterRole %s: %v; retrying...", clusterRoleName, err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("failed to get ClusterRole %s: %v", clusterRoleName, err)
```
rikatz: once more, let's avoid t.Fatal

rikatz: (consider this comment for everywhere else we are using t.Fatal)

test/e2e/util_gatewayapi_test.go
```diff
@@ -1526,6 +1531,76 @@ func eventuallyClusterRoleContainsAggregatedPolicies(t *testing.T, destClusterRo
...
+	if err := kclient.List(t.Context(), crdList); err != nil {
+		t.Fatalf("failed to list CRDs: %v", err)
+	}
+	for _, crd := range crdList.Items {
+		if crd.Spec.Group != "gateway.networking.k8s.io" {
+			continue
+		}
+		// Cluster-scoped CRDs (e.g. GatewayClass) use separate RBAC and
+		// are not expected in namespace-scoped aggregated ClusterRoles.
+		if crd.Spec.Scope == apiextensionsv1.ClusterScoped {
```
rikatz: or you could just put all of your check from slices.Contains inside this if, if it was "if scope != ClusterScoped" or "if scope == NamespacedScope"

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1633,132 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+
+	retrieved := &gatewayapiv1.BackendTLSPolicy{}
+	if err := kclient.Get(ctx, types.NamespacedName{Namespace: ns.Name, Name: "test-btlspolicy"}, retrieved); err != nil {
+		return fmt.Errorf("failed to get created BackendTLSPolicy: %w", err)
+	}
+	t.Logf("successfully created and retrieved BackendTLSPolicy %s/%s", retrieved.Namespace, retrieved.Name)
+	return nil
+}
+
+func buildReferenceGrantUnstructured(name, namespace string) *unstructured.Unstructured {
```
rikatz: why unstructured? ReferenceGrant is also inside Gateway API library

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1633,132 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+						"kind":  "Service",
+					},
+				},
+			},
+		},
+	}
+	return refGrant
+}
+
+// createReferenceGrant creates a ReferenceGrant using unstructured objects (since
```
rikatz, line 1796: ah ok, you explain it here. Did you found this to be a problem or just a precaution?

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1633,132 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+
+	refGrant := buildReferenceGrantUnstructured("test-referencegrant", ns.Name)
+	if err := createWithRetryOnError(t, ctx, refGrant, DefaultRetryTimeout); err != nil {
+		return fmt.Errorf("failed to create ReferenceGrant: %w", err)
+	}
+
+	retrieved := &unstructured.Unstructured{}
+	retrieved.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "gateway.networking.k8s.io",
+		Version: "v1beta1",
```
rikatz: (I think in this case we should probably check on the CRD annotation which Gateway API version is, and assert if v1 or v1beta1 should be used on the unstrctured instead)

test/e2e/gateway_api_test.go
candita: Yes, I think Ali did this to address https://github.com/openshift/cluster-ingress-operator/pull/1467/changes#r3624402005, but I agree this is not the right way to check whether BTLS tests should be run in case of a backport.  @rikatz what is a better way to check whether BTLS tests should be run in a backport-compatible manner?

test/e2e/util_gatewayapi_test.go
candita, line 1: @Thealisyed when I wrote https://github.com/openshift/cluster-ingress-operator/pull/1467#issuecomment-4929903917, I meant that you should do something like `assertHttpRouteConnection` for BTLSPolicy functional testing:  https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/util_gatewayapi_test.go#L1038

Create the BTLSPolicy, HTTPRoute, and anything else needed to make sure that BTLSPolicy relays simple requests properly.  You can use the normative conformance test as an example for setting it up.  Stretch goal would be refactoring `assertHttpRouteConnection` to use an HTTPRoute with a BTLSPolicy.

test/e2e/util_gatewayapi_test.go
Thealisyed: ReferenceGrant's Go type lives in `sigs.k8s.io/gateway-api/apis/v1beta1` which is not vendored here as only `v1` is vendored.
Using unstructured avoids pulling in a whole new vendored package for a single test type.

test/e2e/util_gatewayapi_test.go
Thealisyed: Precaution as importing it directly would be a compile error since `v1beta1` isn't vendored in this repo.

test/e2e/gateway_api_test.go
Thealisyed: For backporting, the intent is that this commit should only be cherry-picked to branches where these CRDs exist so a missing CRD should be a test failure not a silent skip.

test/e2e/util_gatewayapi_test.go
Thealisyed: Acknowledged. Candita is on PTO this week — raising with Ricardo to align on whether a full end-to-end BackendTLSPolicy traffic test is required in this PR or can be tracked as a follow-up, given we're a week from pencils-down.

test/e2e/util_gatewayapi_test.go
Thealisyed: Cool, flattened to a single `if crd.Spec.Group == "gateway.networking.k8s.io" && crd.Spec.Scope == NamespaceScoped` condition.

test/e2e/util_gatewayapi_test.go
Thealisyed: Done — switched to `t.Context()` throughout.

test/e2e/gateway_api_test.go
Thealisyed: Done — both `createBackendTLSPolicy` and `createReferenceGrant` are now called from within `ensureGatewayObjectCreation`, using the same namespace and automatic cleanup.

test/e2e/util_gatewayapi_test.go
Thealisyed: Done — adopted your approach with an `extraCRDs` slice. In the latest version we use `assert.Len(t, extraCRDs, 0, ...)` from testify (per Ricardo's feedback) rather than returning error.

test/e2e/util_gatewayapi_test.go
Thealisyed: Done — added a comment explaining that cluster-scoped CRDs like GatewayClass are handled by separate RBAC and are not expected in namespace-scoped aggregated ClusterRoles.

test/e2e/util_gatewayapi_test.go
Thealisyed: Done — subresources (e.g. `gatewayclasses/status`) are now filtered out via `strings.Contains(res, "/")`. Wildcards are rejected with `require.NotEqual(t, "*", res, ...)`.

test/e2e/gateway_api_test.go
```diff
@@ -54,9 +54,9 @@ const (
 var crdNames = []string{
 	"gatewayclasses.gateway.networking.k8s.io",
 	"gateways.gateway.networking.k8s.io",
-	"grpcroutes.gateway.networking.k8s.io",
 	"httproutes.gateway.networking.k8s.io",
 	"referencegrants.gateway.networking.k8s.io",
+	"backendtlspolicies.gateway.networking.k8s.io",
```
rikatz, line 66: as you are splitting, you probably need to remove this line from here, otherwise it will get a conflict

rikatz:
@Thealisyed some requests:
* Per @candita please open a new story to implement a full e2e test with BTLSPolicy as well
* After https://github.com/openshift/cluster-ingress-operator/pull/1523 merges you need to fix the title, description and rebase this PR
Thanks!

Thealisyed:
Scoping the full operational BackendTLSPolicy e2e test (TLS-enabled backend + actual traffic through the gateway) as a follow-up per discussion with Ricardo. Tracking in OCPBUGS-99912.

Thealisyed:
/hold 
For [PR](https://github.com/openshift/cluster-ingress-operator/pull/1523) to merge

test/e2e/util_gatewayapi_test.go
```diff
@@ -1558,3 +1621,137 @@ func updateGatewaySpecWithRetry(t *testing.T, name types.NamespacedName, timeout
...
+			"kind":       "ReferenceGrant",
+			"metadata": map[string]interface{}{
+				"name":      name,
+				"namespace": namespace,
+			},
+			"spec": map[string]interface{}{
+				"from": []interface{}{
+					map[string]interface{}{
+						"group":     "gateway.networking.k8s.io",
+						"kind":      "HTTPRoute",
```
candita, line 1764: We probably don't need the explicit CEL check, but it would be good to have a test of the operational aspect of BackendTLSPolicy, e.g. make sure it works to deliver the traffic to where it needs to go.

test/e2e/util_gatewayapi_test.go
candita: ReferenceGrant is v1 in Gateway API 1.5:  https://github.com/kubernetes-sigs/gateway-api/blob/release-1.5/apis/v1/referencegrant_types.go#L17

test/e2e/gateway_api_test.go
Thealisyed: Sounds good — once PR #1523 merges to master the rebase of this PR will drop that commit automatically since the `backendtlspolicies` addition will already be in master. Will rebase immediately after #1523 lands.

test/e2e/util_gatewayapi_test.go
Thealisyed: Raised with Ricardo while you were on PTO and he confirmed to scope the full operational test out of this PR and create a Jira to track it. Created OCPBUGS-99912 for the e2e BackendTLSPolicy traffic test. Also added a comment to this PR linking to the Jira.

test/e2e/util_gatewayapi_test.go
Thealisyed: Raised with Ricardo while you were on PTO and he confirmed to scope the full operational test out of this PR and create a Jira to track it. Created OCPBUGS-99912 for the e2e BackendTLSPolicy traffic test. Also added a comment to this PR linking to the Jira.

test/e2e/util_gatewayapi_test.go
Thealisyed: Good catch — and our implementation already handles this correctly. `referenceGrantStorageVersion()` reads the storage version dynamically from the CRD's `spec.versions` at runtime rather than hardcoding `v1beta1`. On a cluster running Gateway API 1.5+ where ReferenceGrant has graduated to v1, it will automatically use `v1` for the unstructured object.

test/e2e/util_gatewayapi_test.go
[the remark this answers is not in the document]
Thealisyed: We added `require.NotEqual` which will fail loudly if a wildcard appears under `gateway.networking.k8s.io` resources in any ClusterRole. The subresource filtering is also commented inline explaining why `/`-containing entries are skipped.

test/e2e/util_gatewayapi_test.go
rikatz: yes, let's keep this out of the scope of this test. I think conformance is enough to guarantee the correct behavior of ReferenceGrant and BackendTLSPolicy

rikatz:
all of my requested changes were made, let's kick the tests for now

/lgtm
/approve

rikatz:

last failure happened on the deprovision, not even on the test, which is passing.

PR #1475, opened 2026-06-16
OCPBUGS-86833: Detect orphaned OSSM subscription after noOLM migration

aswinsuryan:
When migrating from OLM-based Gateway API to the Sail Library path, CIO stops managing the OLM subscription but does not clean it up or notify the cluster-admin. The subscription remains with the ingress.operator.openshift.io/owned annotation, consuming resources and potentially confusing the cluster-admin who may not know the migration happened.

Add a Prometheus gauge metric (ingress_operator_orphaned_ossm_subscription), a ClusterOperator status condition (OrphanedOSSMSubscription), and a PrometheusRule alert to inform the cluster-admin to either remove the annotation and take ownership of the subscription, or delete it entirely.

### DIFF
--- a/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
@@ -60,9 +60,9 @@ spec:
         labels:
           severity: info
         annotations:
-          summary: Orphaned OSSM subscription detected after Gateway API migration
-          description: "This alert fires when the servicemeshoperator3 OLM subscription with the 'ingress.operator.openshift.io/owned' annotation persists after migration to Sail Library (noOLM). The subscription is no longer managed by the Cluster Ingress Operator. To take ownership, remove the annotation: `oc annotate subscription servicemeshoperator3 -n openshift-operators ingress.operator.openshift.io/owned-`. To delete it: `oc delete subscription servicemeshoperator3 -n openshift-operators`."
-          message: "The servicemeshoperator3 subscription is orphaned after migration to Sail Library and is no longer managed by the Cluster Ingress Operator."
+          summary: Orphaned OpenShift Service Mesh operator subscription detected after Gateway API migration
+          description: "This alert fires when the servicemeshoperator3 OLM subscription with the 'ingress.operator.openshift.io/owned' annotation persists after the Cluster Ingress Operator migrated Gateway API from OLM-managed to direct installation. The subscription is no longer managed by the Cluster Ingress Operator. To take ownership, remove the annotation: `oc annotate subscription servicemeshoperator3 -n openshift-operators ingress.operator.openshift.io/owned-`. To uninstall the operator entirely, follow the procedure in the OpenShift documentation for deleting operators from a cluster."
+          message: "An OpenShift Service Mesh operator subscription is orphaned after Gateway API migration and is no longer managed by the Cluster Ingress Operator."
       # Recording rules related to route metrics for sending via telemetry
       - expr: min(route_metrics_controller_routes_per_shard)
         record: cluster:route_metrics_controller_routes_per_shard:min
--- a/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
+++ b/manifests/0000_90_ingress-operator_03_prometheusrules.yaml
@@ -54,6 +54,15 @@ spec:
           message: |
             The {{ $labels.namespace }}/{{ $labels.name }} ingresscontroller is
             unavailable: {{ $labels.reason }}.
+      - alert: OrphanedOSSMSubscription
+        expr: ingress_operator_orphaned_ossm_subscription == 1
+        for: 1h
+        labels:
+          severity: info
+        annotations:
+          summary: Orphaned OSSM subscription detected after Gateway API migration
+          description: "This alert fires when the servicemeshoperator3 OLM subscription with the 'ingress.operator.openshift.io/owned' annotation persists after migration to Sail Library (noOLM). The subscription is no longer managed by the Cluster Ingress Operator. To take ownership, remove the annotation: `oc annotate subscription servicemeshoperator3 -n openshift-operators ingress.operator.openshift.io/owned-`. To delete it: `oc delete subscription servicemeshoperator3 -n openshift-operators`."
+          message: "The servicemeshoperator3 subscription is orphaned after migration to Sail Library and is no longer managed by the Cluster Ingress Operator."
       # Recording rules related to route metrics for sending via telemetry
       - expr: min(route_metrics_controller_routes_per_shard)
         record: cluster:route_metrics_controller_routes_per_shard:min
--- a/pkg/operator/controller/status/controller.go
+++ b/pkg/operator/controller/status/controller.go
@@ -69,7 +69,7 @@ const (
 
 var orphanedOSSMSubscriptionMetric = prometheus.NewGauge(prometheus.GaugeOpts{
 	Name: "ingress_operator_orphaned_ossm_subscription",
-	Help: "Set to 1 when a CIO-owned OSSM subscription persists after migration to Sail Library (noOLM).",
+	Help: "Set to 1 when an orphaned OpenShift Service Mesh operator OLM subscription is detected that is no longer managed by the Cluster Ingress Operator.",
 })
 
 func RegisterMetrics() error {
@@ -151,10 +151,8 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	// status when the OSSM subscription is created or updated.  Note that
 	// the subscriptions resource only exists if the
 	// "OperatorLifecycleManager" capability is enabled, so we cannot watch
-	// it if the capability is not enabled.  Additionally, the default
-	// catalog only exists if the "marketplace" capability is enabled, so we
-	// cannot install OSSM without that capability.
-	if config.MarketplaceEnabled && config.OperatorLifecycleManagerEnabled {
+	// it if the capability is not enabled.
+	if config.OperatorLifecycleManagerEnabled {
 		if err := c.Watch(source.Kind[client.Object](operatorCache, &operatorsv1alpha1.Subscription{}, handler.EnqueueRequestsFromMapFunc(toDefaultIngressController), predicate.Funcs{
 			CreateFunc: func(e event.CreateEvent) bool {
 				return e.Object.GetNamespace() == operatorcontroller.OpenshiftOperatorNamespace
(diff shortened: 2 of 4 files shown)


pkg/operator/controller/status/controller.go
```diff
@@ -65,6 +67,15 @@ const (
 	controllerName = "status_controller"
 )
 
+var orphanedOSSMSubscriptionMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+	Name: "ingress_operator_orphaned_ossm_subscription",
+	Help: "Set to 1 when a CIO-owned OSSM subscription persists after migration to Sail Library (noOLM).",
```
gcs278: *nit* Since this is user facing, "CIO" isn't really a publicly defined acronym, i always prefer to spell it out.

Also, Sail Library (noOLM) is a bit of internal jargon - I wonder if we should make that a bit more generic:

```suggestion
	Help: "Set to 1 when an orphaned OpenShift Service Mesh operator OLM subscription is detected that is no longer managed by the Cluster Ingress Operator.",
```

manifests/0000_90_ingress-operator_03_prometheusrules.yaml
```diff
@@ -54,6 +54,15 @@ spec:
...
             The {{ $labels.namespace }}/{{ $labels.name }} ingresscontroller is
             unavailable: {{ $labels.reason }}.
+      - alert: OrphanedOSSMSubscription
+        expr: ingress_operator_orphaned_ossm_subscription == 1
+        for: 1h
+        labels:
+          severity: info
+        annotations:
+          summary: Orphaned OSSM subscription detected after Gateway API migration
+          description: "This alert fires when the servicemeshoperator3 OLM subscription with the 'ingress.operator.openshift.io/owned' annotation persists after migration to Sail Library (noOLM). The subscription is no longer managed by the Cluster Ingress Operator. To take ownership, remove the annotation: `oc annotate subscription servicemeshoperator3 -n openshift-operators ingress.operator.openshift.io/owned-`. To delete it: `oc delete subscription servicemeshoperator3 -n openshift-operators`."
```
gcs278: Removing the subscription doesn't uninstall OSSM. It leaves behind the ossm operator pod, this link provides info on how to remove an OLM Operator: https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/operators/administrator-tasks#olm-deleting-operators-from-a-cluster

You to delete the subscription and CSV, but there can be multiple CSV, so you need to look up the CSV.

*nit* it's a minor point, but `NoOLM` is just a jargon I use, and sail library is also a bit in the implementation details, what about something like this?

```suggestion
          description: "This alert fires when the servicemeshoperator3 OLM subscription with the 'ingress.operator.openshift.io/owned' annotation persists after the Cluster Ingress Operator migrated Gateway API from OLM-managed to direct installation. The subscription is no longer managed by the Cluster Ingress Operator. To take ownership, remove the annotation: `oc annotate subscription servicemeshoperator3 -n openshift-operators ingress.operator.openshift.io/owned-`. To uninstall the operator entirely, follow the procedure in the OpenShift documentation for deleting operators from a cluster."
```

pkg/operator/controller/status/controller.go
```diff
@@ -698,6 +727,24 @@ func computeGatewayAPIInstallDegradedCondition(state operatorState) configv1.Clu
...
+	names := make([]string, 0, len(state.orphanedOSSMSubscriptions))
+	for _, sub := range state.orphanedOSSMSubscriptions {
+		names = append(names, fmt.Sprintf("%s/%s", sub.Namespace, sub.Name))
+	}
+	degradedCondition.Status = configv1.ConditionFalse
+	degradedCondition.Reason = "OrphanedOSSMSubscription"
+	degradedCondition.Message = fmt.Sprintf(
+		"Orphaned OSSM subscription(s) found after migration to Sail Library: %s.",
+		strings.Join(names, ", "),
+	)
```
gcs278, line 741: I think we should also explain to a user here what they need to do for visibility.

```suggestion
        degradedCondition.Message = fmt.Sprintf(
           "Orphaned OSSM subscription(s) found after the Cluster Ingress Operator migrated Gateway API from OLM-managed to direct installation: %s. "+
           "The subscription(s) are no longer managed by the Cluster Ingress Operator. "+
           "To take ownership, remove the 'ingress.operator.openshift.io/owned' annotation. "+
            "To uninstall the operator entirely, follow the procedure in the OpenShift documentation for deleting operators from a cluster.",
         strings.Join(names, ", "),
    )
```

pkg/operator/controller/status/controller.go
```diff
@@ -485,9 +504,10 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
 
 		state.expectedGatewayAPIOperatorVersion = r.config.GatewayAPIOperatorVersion
 
-		// List OSSM subscriptions (OLM-specific only).
-		// In Sail Library mode, we don't check for subscription conflicts, so no need to list them.
-		if useOLM {
+		// List OSSM subscriptions.
+		// In OLM mode, check for subscription conflicts.
+		// In Sail Library mode, detect orphaned CIO-owned subscriptions.
+		if useOLM || useSailLibrary {
```
gcs278: Okay, so I realize `useOLM` was a bad name, it's confusing things, including me. It should be called `olmCapabilityEnabled`. There is an existing bug here actually - the condition should have been:
```
		if !useSailLibrary {
```

So I think the PR's change is incorrect. The context: clusters without OLM capabilities might not have the `Subscription` CRD, so this would actually cause all `r.client.List` for `Subscriptions` to fail hard and it'd block all status updates to the cluster operator.

In a ironic turn of events, you actually need this bug for this PR (now we want to check subscriptions for sail library). So I'd just leave it be:

```suggestion
		if useOLM {
```

And see my comment below about fixing the existing bug by using `!useSailLibrary` inside the loop.

pkg/operator/controller/status/controller.go
```diff
@@ -496,9 +516,17 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
 				return state, fmt.Errorf("failed to get subscriptions: %w", err)
 			}
 			for _, subscription := range subscriptionList.Items {
-				if subscription.Spec != nil && ossmSubscriptions.Has(subscription.Spec.Package) {
+				if subscription.Spec == nil {
+					continue
+				}
+				if useOLM && ossmSubscriptions.Has(subscription.Spec.Package) {
```
gcs278: Per my other comment, I think you mean `!useSailLibrary`.
`useOLM` means that OLM capabilities are enabled, which is almost always true.

So this will fix the bug I talk about above:

```suggestion
				if !useSailLibrary && ossmSubscriptions.Has(subscription.Spec.Package) {
```

gcs278 (commented):
nice PR @aswinsuryan thanks! A couple of comments.

Your PR also led me to find a related bug, so I have opened it https://redhat.atlassian.net/browse/OCPBUGS-90505

manifests/0000_90_ingress-operator_03_prometheusrules.yaml
rhamini3: should we also link the documentation to provide the steps on how to correctly delete the subscription and its related CSV similar to how the oc commands are provided above?

gcs278: Yea we could.  I'd suggest using the `latest` version tag for the docs link (e.g. `https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/...`) so we don't have to maintain bumping the version every release or hardcode a version like `5.0` that doesn't exist yet and will be outdated by `5.1`.

On the backports, we can hardcode it to the specific OCP version for better precision, but that's extra credit. We could also derive it from `RELEASE_VERSION` at runtime, but that feels like overkill for a docs link in a status message.

Though, one risk is that these links often break and we'd need to keep an eye on it. But it may still provide value in the hopefully limited time this alert is useful. I'll let @aswinsuryan make the call. I lean towards not including it due to risk of the link breaking in the future.

pkg/operator/controller/status/controller.go
aswinsuryan: Done.

pkg/operator/controller/status/controller.go
aswinsuryan: Good catch,  Yes I did not check the actual function of the flag, went by the name. Fixed it

pkg/operator/controller/status/controller.go
aswinsuryan: Done.

gcs278:
I think we forgot to assign some PRs after PR Scrub

/assign
/assign @rhamini3

manifests/0000_90_ingress-operator_03_prometheusrules.yaml
aswinsuryan: As the current wording should be enough to guide admins, I also think we can skip the docs link for now, considering the overhead of maintaining it

pkg/operator/controller/status/controller.go
```diff
@@ -453,7 +472,7 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
 	useOLM := r.config.MarketplaceEnabled && r.config.OperatorLifecycleManagerEnabled
 	useSailLibrary := r.config.GatewayAPIWithoutOLMEnabled
 	state.useSailLibrary = useSailLibrary
-	if useOLM || useSailLibrary {
+	if useOLM {
```
gcs278: I think this condition should be removed entirely. If you remove just the `|| useSailLibrary`, then it will no longer execute this logic for clusters without OLM capabilities (but with using Sail Library). 

```suggestion
```

pkg/operator/controller/status/controller.go
```diff
@@ -453,7 +472,7 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
 	useOLM := r.config.MarketplaceEnabled && r.config.OperatorLifecycleManagerEnabled
```
gcs278: Sorry I know this is adjacent, but this is also a bug - `r.config.MarketplaceEnabled` doesn't determine whether OLM is available, it only controls whether OLM catalog images are downloadable. A cluster can have OLM enabled without Marketplace, and this condition would incorrectly skip subscription checks on those clusters. This should just be based on OLM capability.

Feel free to skip - but it'd be easy to fix this too while you are in here. Also, `useOLM` is a bad name we should change (it's confusing me).

```suggestion
	olmCapabilityEnabled := r.config.OperatorLifecycleManagerEnabled
```

gcs278 (commented):
Thanks for the updates! Just two more comments.

pkg/operator/controller/status/controller.go
aswinsuryan: Done.

pkg/operator/controller/status/controller.go
aswinsuryan: oh ok got it,  removed.

pkg/operator/controller/status/controller.go
gcs278, line 492: *nit* Did you intentionally remove this? I think it's still a valid comment right?

aswinsuryan: is that one in line 509? Sorry, it was not intentional, added it back.

gcs278: Sorry - another one that is adjacent to your changes. This is also incorrect because we are intending to support clusters with marketplace disabled + OLM enabled.

It matters for your PR because for a MarketPlace-disabled + OLM-enabled cluster that installed OSSM via a subscription may NOT get your abandon alert in a timely manner as it wouldn't create these watches.

At the time of creating this code, we weren't thinking about supporting disconnected clusters like this, but now we do. Marketplace was never needed for disconnected clusters in the OLM path.

```suggestion
// it if the capability is not enabled. 
if config.OperatorLifecycleManagerEnabled {
```

aswinsuryan: oh ok .

pkg/operator/controller/status/controller.go
```diff
@@ -450,68 +469,73 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
...
+	} else {
+		state.haveGatewayclassesResource = true
+	}
+	if err := r.cache.Get(ctx, istiosResourceNamespacedName, &crd); err != nil {
+		if !errors.IsNotFound(err) {
+			return state, fmt.Errorf("failed to get CRD %q: %v", istiosResourceName, err)
 		}
+	} else {
+		state.haveIstiosResource = true
+	}
```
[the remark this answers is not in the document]
aswinsuryan: This is existing code not changed in this PR.

gcs278:
Thanks for all of the quick updates @aswinsuryan! looks great - I'm very interested in getting this backported ASAP as it's a very important signal to help avoid users wasting resources in the OLM to no-OLM transition.

/approve

@rhamini3 want to take a quick look and LGTM, then do a verify of this?

gcs278:
all unrelated
/retest

rhamini3:
PR looks good with unrelated failures
/lgtm

rhamini3:
Tested on a 5.0 cluster with CIO based provisioning for gatewayAPI

Subscription contains a ingress operator owned annoation, ingress CO shows expected message
```
 lastTransitionTime: "2026-06-25T16:18:19Z"
    message: |-
      The "default" ingress controller reports Degraded=False.
      Orphaned OSSM subscription(s) found after the Cluster Ingress Operator migrated Gateway API from OLM-managed to direct installation: openshift-operators/servicemeshoperator3. The subscription(s) are no longer managed by the Cluster Ingress Operator. To take ownership, remove the 'ingress.operator.openshift.io/owned' annotation. To uninstall the operator entirely, follow the procedure in the OpenShift documentation for deleting operators from a cluster.
    reason: IngressNotDegradedAndOrphanedOSSMSubscription
    status: "False"
    type: Degraded
```
Prometheus metrics active showing a value of 1
```
% oc -n openshift-ingress-operator logs ingress-operator-5594445674-ws6zr | grep registering
Defaulted container "ingress-operator" out of: ingress-operator, kube-rbac-proxy
2026-06-25T16:15:56.520Z	INFO	operator.main	ingress-operator/start.go:76	registering Prometheus metrics for canary_controller
2026-06-25T16:15:56.520Z	INFO	operator.main	ingress-operator/start.go:76	registering Prometheus metrics for ingress_controller
2026-06-25T16:15:56.520Z	INFO	operator.main	ingress-operator/start.go:76	registering Prometheus metrics for route_metrics_controller
2026-06-25T16:15:56.520Z	INFO	operator.main	ingress-operator/start.go:76	registering Prometheus metrics for status_controller

% curl -X GET -k -H "Authorization: Bearer $TOKEN" \
  "https://$THANOS_URL/api/v1/query?query=ingress_operator_orphaned_ossm_subscription"
{"status":"success","data":{"resultType":"vector","result":[{"metric":{"__name__":"ingress_operator_orphaned_ossm_subscription","container":"kube-rbac-proxy","endpoint":"metrics","instance":"10.128.0.38:9393","job":"metrics","namespace":"openshift-ingress-operator","pod":"ingress-operator-5594445674-ws6zr","prometheus":"openshift-monitoring/k8s","service":"metrics"},"value":[1782410551.120,"1"]}],"analysis":{}}}
```

After removing the annotation, `ingress_operator_orphaned_ossm_subscription` metric is cleared and ingress co condition is removed
```
 - lastTransitionTime: "2026-06-25T16:18:19Z"
    message: The "default" ingress controller reports Degraded=False.
    reason: IngressNotDegraded
    status: "False"
    type: Degraded

% curl -X GET -k -H "Authorization: Bearer $TOKEN" \
  "https://$THANOS_URL/api/v1/query?query=ingress_operator_orphaned_ossm_subscription"
{"status":"success","data":{"resultType":"vector","result":[{"metric":{"__name__":"ingress_operator_orphaned_ossm_subscription","container":"kube-rbac-proxy","endpoint":"metrics","instance":"10.128.0.38:9393","job":"metrics","namespace":"openshift-ingress-operator","pod":"ingress-operator-5594445674-ws6zr","prometheus":"openshift-monitoring/k8s","service":"metrics"},"value":[1782410744.302,"0"]}],"analysis":{}}}
```
/verified by @rhamini3

gcs278:
@aswinsuryan mind if we backport this 4.22, 4.21, 4.20 (and maybe 4.19 if NO-OLM lands there...)? should be a clean backport and easy with the `/cherry-pick` command.

gcs278:
Failure on `[sig-api-machinery] FieldValidation should detect duplicates in a CR when preserving unknown fields [Conformance] `

/test e2e-aws-ovn

gcs278:
@aswinsuryan let me know if you are ready to open that backport for release-4.22 - we should be good to begin that process.

PR #1478, opened 2026-06-18
NE-2737: Add support for TLS curves in router deployment configuration

richardsonnick:
To support new PQC curves, let's add a config for explicitly setting the supported groups. 

ROUTER_CURVES is piped through OpenShift/router to HAProxy where it will be used by `ssl-default-bind-curves`

For more context see: https://github.com/openshift/router/pull/678

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -1090,20 +1091,52 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 		})
 	}
 
-	// The default TLS supportedGroups (curves) include X25519MLKEM768,
-	// x25519, P-256, P-384, and P-521.
-	tlsCurves := "X25519MLKEM768:X25519:P-256:P-384:P-521"
-	// If FIPS is enabled on this cluster, we cannot use
-	// ML-KEM or X25519 in the TLS supportedGroups (aka curves).
-	// ML-KEM and X25519 are not supported by OpenSSL FIPS 140-3.
+	// Determine the candidate group list: use the explicitly configured groups
+	// when set (via a Custom TLS profile), otherwise use the defaults from the
+	// named TLS profiles. Component implementors do not need to check the
+	// TLSCurvePreferences feature gate; an absent field means the gate is
+	// disabled and components fall back to defaults automatically.
+	var tlsGroupList []string
+	if len(tlsProfileSpec.Groups) != 0 {
+		// When an explicit groups list is configured, pass it to the router
+		// as-is. FIPS filtering is applied below after the else clause.
+		for _, g := range tlsProfileSpec.Groups {
+			tlsGroupList = append(tlsGroupList, string(g))
+		}
+	} else {
+		// The default TLS supportedGroups (curves) include X25519MLKEM768,
+		// X25519, secp256r1 (P-256), secp384r1 (P-384), and secp521r1 (P-521).
+		// These use IANA names that HAProxy/OpenSSL accept directly.
+		tlsGroupList = []string{
+			string(configv1.TLSGroupX25519MLKEM768),
+			string(configv1.TLSGroupX25519),
+			string(configv1.TLSGroupSecP256r1),
+			string(configv1.TLSGroupSecP384r1),
+			string(configv1.TLSGroupSecP521r1),
+		}
+	}
+
+	// If FIPS is enabled on this cluster, we cannot use ML-KEM or X25519 in
+	// the TLS supportedGroups (aka curves). ML-KEM and X25519 are not
+	// supported by OpenSSL FIPS 140-3.
 	// See https://redhat.atlassian.net/browse/TRT-2597 and Appendix D of
 	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
+	// If filtering empties the list, fall back to the FIPS-compliant defaults
+	// (secp256r1:secp384r1:secp521r1) so the router always has at least one
+	// usable group.
 	if isFIPSEnabled {
-		tlsCurves = "P-256:P-384:P-521"
+		tlsGroupList = slices.DeleteFunc(tlsGroupList, nonFIPSGroups.Has)
+		if len(tlsGroupList) == 0 {
+			tlsGroupList = []string{
+				string(configv1.TLSGroupSecP256r1),
+				string(configv1.TLSGroupSecP384r1),
+				string(configv1.TLSGroupSecP521r1),
+			}
+		}
 	}
 	env = append(env, corev1.EnvVar{
 		Name:  RouterTLSCurves,
-		Value: tlsCurves,
+		Value: strings.Join(tlsGroupList, ":"),
 	})
 
 	var minTLSVersion string
@@ -1518,6 +1551,7 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 		ciphersString       string
 		cipherSuitesString  string
 		minTLSVersionString string
+		groupsString        string
 	)
 	for _, v := range env {
 		switch v.Name {
@@ -1527,6 +1561,8 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 			cipherSuitesString = v.Value
 		case "SSL_MIN_VERSION":
 			minTLSVersionString = v.Value
+		case RouterTLSCurves:
+			groupsString = v.Value
 		}
 	}
 
@@ -1550,9 +1586,17 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 		minTLSVersion = configv1.VersionTLS12
 	}
 
+	var groups []configv1.TLSGroup
+	if len(groupsString) != 0 {
+		for _, s := range strings.Split(groupsString, ":") {
+			groups = append(groups, configv1.TLSGroup(s))
+		}
+	}
+
 	profile := &configv1.TLSProfileSpec{
 		Ciphers:       ciphers,
 		MinTLSVersion: minTLSVersion,
+		Groups:        groups,
 	}
 
 	return profile
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -1156,7 +1156,7 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 		{RouterHTTPIgnoreProbes, true, "true"},
 		{"ROUTER_CIPHERS", true, "quux"},
 		{"ROUTER_CIPHERSUITES", true, "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256"},
-		{"ROUTER_CURVES", true, "X25519:secp256r1:secp384r1"},
+		{"ROUTER_CURVES", true, "X25519:secp256r1:secp521r1:secp384r1"},
 		{"SSL_MIN_VERSION", true, "TLSv1.2"},
 		{"ROUTER_IP_V4_V6_MODE", true, "v4v6"},
 		{RouterEnableCompression, true, "true"},
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -1051,6 +1051,12 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 					"TLS_AES_256_GCM_SHA384",
 					"TLS_CHACHA20_POLY1305_SHA256",
 				},
+				Groups: []configv1.TLSGroup{
+					configv1.TLSGroupX25519,
+					configv1.TLSGroupSecP256r1,
+					configv1.TLSGroupSecP521r1,
+					configv1.TLSGroupSecP384r1,
+				},
 				MinTLSVersion: configv1.VersionTLS12,
 			},
 		},
@@ -1150,6 +1156,7 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 		{RouterHTTPIgnoreProbes, true, "true"},
 		{"ROUTER_CIPHERS", true, "quux"},
 		{"ROUTER_CIPHERSUITES", true, "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256"},
+		{"ROUTER_CURVES", true, "X25519:secp256r1:secp384r1"},
 		{"SSL_MIN_VERSION", true, "TLSv1.2"},
 		{"ROUTER_IP_V4_V6_MODE", true, "v4v6"},
 		{RouterEnableCompression, true, "true"},
@@ -1915,6 +1922,10 @@ func Test_inferTLSProfileSpecFromDeployment(t *testing.T) {
 							Name:  "SSL_MIN_VERSION",
 							Value: "TLSv1.3",
 						},
+						{
+							Name:  "ROUTER_CURVES",
+							Value: "foo:bar",
+						},
 					},
 				},
 				{
@@ -1924,6 +1935,7 @@ func Test_inferTLSProfileSpecFromDeployment(t *testing.T) {
 			expected: &configv1.TLSProfileSpec{
 				Ciphers:       []string{"foo", "bar", "baz"},
 				MinTLSVersion: configv1.VersionTLS13,
+				Groups:        []configv1.TLSGroup{"foo", "bar"},
 			},
 		},
 	}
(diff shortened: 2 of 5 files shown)


richardsonnick:
Reopening this on a new branch. The old PR is [here](https://github.com/openshift/cluster-ingress-operator/pull/1287)

richardsonnick:
Depends on https://github.com/openshift/cluster-ingress-operator/pull/1473

candita:
@richardsonnick you can update the api within this PR, no need to wait for https://github.com/openshift/cluster-ingress-operator/pull/1473 to merge.  Keep in a separate commit.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,21 +1062,54 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
-	// ML-KEM and X25519 are not supported by OpenSSL FIPS 140-3.
-	// See https://redhat.atlassian.net/browse/TRT-2597 and Appendix D of
-	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
-	if isFIPSEnabled {
-		tlsCurves = "P-256:P-384:P-521"
+	// fipsCompliantDefaultGroups is the fallback used when FIPS mode is
+	// enabled and the explicit (or default) group list contains no
+	// FIPS-approved groups after filtering. secp256r1 and secp384r1 are the
+	// IANA names for P-256 and P-384, which HAProxy/OpenSSL also accept.
+	const fipsCompliantDefaultGroups = "secp256r1:secp384r1"
```
candita: Please don't change the group list that was approved initially, which includes P-521.  If you want to want to use the configv1.TLSGroup... constants, that's fine, but use them everywhere.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3014,13 +3063,20 @@ func TestDesiredRouterDeploymentTLSCurves(t *testing.T) {
...
+				}
+			}
 
 			wasFIPSEnabled := isFIPSEnabled
-			// Restore the real value of isFIPSEnabled in case it is used by other tests.
-			defer func() {
-				isFIPSEnabled = wasFIPSEnabled
-			}()
-			// Use the temporary test case status of FIPS enablement for testing.
+			defer func() { isFIPSEnabled = wasFIPSEnabled }()
```
candita: Please don't remove godoc.  That is there to help future maintainers.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,21 +1062,54 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		//
+		// On FIPS-enabled clusters, non-FIPS-approved groups (X25519 and
+		// X25519MLKEM768) must be removed from the explicit list, just as they
+		// are removed from the defaults below. If filtering empties the list,
+		// fall back to the FIPS-compliant defaults so the router always has at
+		// least one usable group.
+		var groupStrs []string
+		for _, g := range tlsProfileSpec.Groups {
+			if isFIPSEnabled && nonFIPSGroups.Has(string(g)) {
+				continue
```
candita: Please info log here that the group is being omitted.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,21 +1062,54 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+			}
+			groupStrs = append(groupStrs, string(g))
+		}
+		tlsGroups := strings.Join(groupStrs, ":")
+		if tlsGroups == "" {
+			tlsGroups = fipsCompliantDefaultGroups
```
candita: We don't know here if FIPS is enabled or not.  You should defer this check and setting the env until after the `else` clause.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,21 +1062,54 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		// X25519 and X25519MLKEM768 are not FIPS-approved; omit them when FIPS
+		// is enabled so that only FIPS-compliant EC groups are offered.
+		tlsGroups := "X25519MLKEM768:X25519:secp256r1:secp384r1"
+		if isFIPSEnabled {
+			tlsGroups = fipsCompliantDefaultGroups
+		}
+		env = append(env, corev1.EnvVar{
+			Name:  RouterTLSCurves,
+			Value: tlsGroups,
+		})
```
candita: Defer setting the env until after the `else` clause.  See https://github.com/openshift/cluster-ingress-operator/pull/1478/changes#r3438608659.

pkg/operator/controller/ingress/deployment.go
Miciah: Wouldn't that get logged on every reconciliation?  That could be noisy on FIPS-enabled clusters.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,20 +1062,61 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+			filtered = append(filtered, g)
+		}
+		tlsGroups = strings.Join(filtered, ":")
+		if tlsGroups == "" {
+			tlsGroups = strings.Join([]string{
+				string(configv1.TLSGroupSecP256r1),
+				string(configv1.TLSGroupSecP384r1),
+				string(configv1.TLSGroupSecP521r1),
+			}, ":")
+		}
```
Miciah: Seems wasteful to join, split, filter, and join again.  Any reason not to compute the `[]string` value, then filter it if needed for FIPS, and then join just once?

pkg/operator/controller/ingress/deployment.go
Miciah: If we could omit the logging, we could also simplify the filtering logic down to this:
```go
		filtered = slices.DeleteFunc(tlsGroups, nonFIPSGroups.Has)
```

Miciah:
Please update `inferTLSProfileSpecFromDeployment` so that the TLS profile in the IngressController resource's status reports the groups that are configured in the router deployment:
```diff
diff --git a/pkg/operator/controller/ingress/deployment.go b/pkg/operator/controller/ingress/deployment.go
index 94777fefd..7a068b0dd 100644
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1483,6 +1483,7 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 		ciphersString       string
 		cipherSuitesString  string
 		minTLSVersionString string
+		groupsString        string
 	)
 	for _, v := range env {
 		switch v.Name {
@@ -1492,6 +1493,8 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 			cipherSuitesString = v.Value
 		case "SSL_MIN_VERSION":
 			minTLSVersionString = v.Value
+		case RouterTLSCurves:
+			groupsString = v.Value
 		}
 	}
 
@@ -1515,9 +1518,17 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 		minTLSVersion = configv1.VersionTLS12
 	}
 
+	var groups []configv1.TLSGroup
+	if len(groupsString) != 0 {
+		for s := range strings.SplitSeq(groupsString, ":") {
+			groups = append(groups, configv1.TLSGroup(s))
+		}
+	}
+
 	profile := &configv1.TLSProfileSpec{
 		Ciphers:       ciphers,
 		MinTLSVersion: minTLSVersion,
+		Groups:        groups,
 	}
 
 	return profile
diff --git a/pkg/operator/controller/ingress/deployment_test.go b/pkg/operator/controller/ingress/deployment_test.go
index 7f9939d48..f1a7687a9 100644
--- a/pkg/operator/controller/ingress/deployment_test.go
+++ b/pkg/operator/controller/ingress/deployment_test.go
@@ -1656,6 +1656,10 @@ func Test_inferTLSProfileSpecFromDeployment(t *testing.T) {
 							Name:  "SSL_MIN_VERSION",
 							Value: "TLSv1.3",
 						},
+						{
+							Name:  "ROUTER_CURVES",
+							Value: "foo:bar",
+						},
 					},
 				},
 				{
@@ -1665,6 +1669,7 @@ func Test_inferTLSProfileSpecFromDeployment(t *testing.T) {
 			expected: &configv1.TLSProfileSpec{
 				Ciphers:       []string{"foo", "bar", "baz"},
 				MinTLSVersion: configv1.VersionTLS13,
+				Groups:        []configv1.TLSGroup{"foo", "bar"},
 			},
 		},
 	}
```

Miciah:
If you still have them, please add the commands that you used to do the vendor bump to the commit message, per team convention.  For an example of what I mean, please see the commit message here: https://github.com/openshift/cluster-ingress-operator/commit/5374dab985e9420c1ee9dbcfab55c550bbd0bbdf.

pkg/operator/controller/ingress/deployment.go
richardsonnick: Good point. This is unnecessary work. fixing.

Miciah:
By the way, it looks like you had literal ``` `groups` ``` in your commit message, and your shell expanded it:
```text
Adds support for setting the list of TLS supported groups (curves) used
during the TLS handshake via the Custom TLS security profile's new
ubuntu adm cdrom sudo dip lxd field, which maps to the router's ROUTER_CURVES environment
variable.
```

pkg/operator/controller/ingress/deployment.go
candita: Good point @Miciah - your idea about logging the configured groups in the status is better.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1062,20 +1062,59 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	} else {
+		// The default TLS supportedGroups (curves) include X25519MLKEM768,
+		// X25519, secp256r1 (P-256), and secp384r1 (P-384). These use IANA
+		// names that HAProxy/OpenSSL accept directly.
+		tlsGroupList = []string{
+			string(configv1.TLSGroupX25519MLKEM768),
+			string(configv1.TLSGroupX25519),
+			string(configv1.TLSGroupSecP256r1),
+			string(configv1.TLSGroupSecP384r1),
```
candita, line 1114: The default should still include P-521.

pkg/operator/controller/ingress/deployment.go
richardsonnick: Great idea thanks @Miciah will fix

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1474,9 +1510,17 @@ func inferTLSProfileSpecFromDeployment(deployment *appsv1.Deployment) *configv1.
 		minTLSVersion = configv1.VersionTLS12
 	}
 
+	var groups []configv1.TLSGroup
+	if len(groupsString) != 0 {
+		for _, s := range strings.Split(groupsString, ":") {
+			groups = append(groups, configv1.TLSGroup(s))
+		}
```
candita, line 1593: Nit:
```suggestion
	groups = append(groups, strings.Split(groupsString, ":"))
```

richardsonnick: Go only allows converting between slice types if the element types are identical, I think.

Since `groups = []configv1.TLSGroup` this doesnt seem to compile. The original does a cast to the correct type

davidesalerno:
@coderabbitai resume

davidesalerno:
@richardsonnick I think that we have to update the function `copyTLSSpec` to deep-copy `Groups` (pkg/operator/controller/tls.go:14):
Add `out.Groups = append([]configv1.TLSGroup(nil), in.Groups...) `after the `Ciphers` deep copy.
Without this, the function's defensive-copy contract is broken for the new field.

candita:
Miciah is too busy for this, taking it off his plate.
/unassign @Miciah

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -1817,6 +1827,7 @@ func Test_inferTLSProfileSpecFromDeployment(t *testing.T) {
 			expected: &configv1.TLSProfileSpec{
 				Ciphers:       []string{"foo", "bar", "baz"},
 				MinTLSVersion: configv1.VersionTLS13,
+				Groups:        []configv1.TLSGroup{"foo", "bar"},
```
candita, line 1938: Should this group list be empty since the groups are invalid?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -944,6 +944,11 @@ func TestDesiredRouterDeploymentSpecAndNetwork(t *testing.T) {
 					"TLS_AES_256_GCM_SHA384",
 					"TLS_CHACHA20_POLY1305_SHA256",
 				},
+				Groups: []configv1.TLSGroup{
+					configv1.TLSGroupX25519,
+					configv1.TLSGroupSecP256r1,
+					configv1.TLSGroupSecP384r1,
```
candita, line 1058: You could add P521 to a custom tls profile, please do so here.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+		name           string
+		fipsEnabled    bool
+		explicitGroups []configv1.TLSGroup // nil means no Custom profile; non-nil (even empty) sets a Custom profile
+		expectedEnv    []envData
 	}{
+		// ── Named profile (no explicit Custom profile set) ───────────────────
+		// Named profiles (Intermediate, etc.) have Groups populated by the API,
+		// so the if-branch is taken; the else-branch fallback (which includes
+		// secp521r1) is only reached when the TLSCurvePreferences feature gate
+		// is disabled and Groups is absent entirely.
```
candita: Please don't call this a "named profile". It should be either a custom profile or a pre-defined type (Old,Intermediate,Modern).  A custom profile can be created with custom groups, and include P521. 

```suggestion
		// Modern, Intermediate, and Old profiles all have the same pre-defined groups:
		//	TLSGroupX25519MLKEM768,
		//	TLSGroupX25519,
		//	TLSGroupSecP256r1,
		//	TLSGroupSecP384r1,
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+		{
+			name:        "Named profile: groups passed through when FIPS is not enabled",
+			fipsEnabled: false,
+			expectedEnv: []envData{
+				{"ROUTER_CURVES", true, "X25519MLKEM768:X25519:secp256r1:secp384r1"},
+			},
+		},
 		{
-			name:        "Include ML-KEM and X25519 when FIPS is not enabled",
+			name:        "Named profile: non-FIPS groups filtered when FIPS is enabled",
```
candita:
```suggestion
			name:        "non-FIPS groups are filtered when FIPS is enabled",
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+			fipsEnabled: true,
+			expectedEnv: []envData{
+				// X25519MLKEM768 and X25519 are filtered; named profile has no P-521.
```
candita:
```suggestion
				// X25519MLKEM768 and X25519 are filtered
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+				{"ROUTER_CURVES", true, "secp256r1:secp384r1"},
+			},
+		},
+		// ── Explicit groups (Custom TLS profile) ────────────────────────────
```
candita:
```suggestion
		// ── Custom groups (Custom TLS profile) ────────────────────────────
```

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+		{
+			name:        "Explicit: groups passed through unchanged when FIPS is not enabled",
 			fipsEnabled: false,
+			explicitGroups: []configv1.TLSGroup{
+				configv1.TLSGroupX25519MLKEM768,
+				configv1.TLSGroupX25519,
+				configv1.TLSGroupSecP256r1,
```
candita, line 3369: This is a good place to add P-521.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -3205,25 +3216,94 @@ func Test_ClosedClientConnectionPolicy(t *testing.T) {
...
+			explicitGroups: []configv1.TLSGroup{
+				configv1.TLSGroupX25519MLKEM768,
+				configv1.TLSGroupX25519,
+			},
+			expectedEnv: []envData{
+				// Entire list is non-FIPS; must fall back to secp256r1:secp384r1:secp521r1.
+				{"ROUTER_CURVES", true, "secp256r1:secp384r1:secp521r1"},
+			},
+		},
+		// ── Feature-gate disabled (Groups field absent) ──────────────────────
```
candita, line 3401: I think you're trying to simulate feature-gate disabled here, but this is just the same case as empty groups, which provide the default groups.

richardsonnick:
failing tests seem to be flakes [ci/prow/e2e-aws-operator-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1478/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-techpreview/2074855101507309568) [ci/prow/hypershift-e2e-aks](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1478/pull-ci-openshift-cluster-ingress-operator-master-hypershift-e2e-aks/2074855120750776320)

pkg/operator/controller/ingress/deployment_test.go
richardsonnick: The CRD handles verification before it hits this, this [`inferTLSProfilespecFromDeployment` test](https://github.com/openshift/cluster-ingress-operator/blob/7b7406ee0d4bf03de360d53c9cc4a83ee14332cc/pkg/operator/controller/ingress/deployment.go#L1455) is testing the parsing functionality. 

The CRD handles verification [here](https://github.com/richardsonnick/cluster-ingress-operator/blob/f9760ba418c0f818eb2f389dd7555a47e7d57850/manifests/00-custom-resource-definition-TechPreviewNoUpgrade.yaml#L2052). So this test is basically saying "whatever the CRD says goes".

What do you think? Should we add secondary validation for the enum values here?

candita:
@richardsonnick looks like you may need to fix a unit test.  You changed the input based on my comment, but didn't change the expectation.

> --- FAIL: TestDesiredRouterDeploymentSpecAndNetwork (0.00s)
>     deployment_test.go:1060: router Deployment has unexpected ROUTER_CURVES setting: expected "X25519:secp256r1:secp384r1", got "X25519:secp256r1:secp521r1:secp384r1"
>

pkg/operator/controller/ingress/deployment_test.go
Miciah: Yeah, the test is just verifying that the operator correctly parses a colon-delimited string into a slice.  It is a unit test, so it isn't subject to API validation.  We could substitute actual group names in the test inputs and outputs, but that isn't necessary for the test to serve its intended purpose, and using "foo" and "bar" makes the test easier to read in my opinion.

candita: No.  It's possible that a custom tls profile might contain something unexpected.

candita:
This error isn't related to the changes:
> ERROR: (ResourceGroupDeletionBlocked) Delete resource group MC_ci-op-pbnifxt2-f38cb-aks-rg_ci-op-pbnifxt2-f38cb-aks-cluster_uksouth failed. Deletion of resource group 'MC_ci-op-pbnifxt2-f38cb-aks-rg_ci-op-pbnifxt2-f38cb-aks-cluster_uksouth' failed as '14' resources could not be deleted. The provisioning state of the resource group will be rolled back. The tracking Id is '545c462d-e19b-4d7a-8991-7409be20cc2e'. Please check audit logs for more details.
> Code: ResourceGroupDeletionBlocked
> 

Let's see if it has been taken care of.

/test hypershift-e2e-aks

davidesalerno:
Hey @richardsonnick, the `e2e-aws-operator-techpreview` job is failing because `TestTLSSecurityProfile` doesn't account for TLS groups when the `TLSGroupPreferences` feature gate is enabled (which it is in TechPreview).

The test sets a custom profile with no `Groups` field, then asserts the full status matches with `reflect.DeepEqual`. But with your changes, the operator now populates default TLS groups in the status, so the assertion fails:

```
expected: groups: []
got:      groups: [X25519MLKEM768 X25519 secp256r1 secp384r1 secp521r1]
```

The fix needs three changes in `TestTLSSecurityProfile`:

1. **Check the feature gate** before asserting on groups for the Intermediate profile default:
```go
tlsGroupsEnabled, err := isFeatureGateEnabled(features.FeatureGateTLSGroupPreferences)
if err != nil {
    t.Fatalf("failed to check TLSGroupPreferences feature gate: %v", err)
}
if tlsGroupsEnabled {
    assert.Equal(t, intermediateProfileSpec.Groups, ic.Status.TLSProfile.Groups,
        "ingresscontroller status has unexpected TLS groups: expected %v, got %v",
        intermediateProfileSpec.Groups, ic.Status.TLSProfile.Groups)
}
```

2. **Add `Groups` to the custom profile spec** so the assertion has expected values:
```go
customProfileSpec := configv1.TLSProfileSpec{
    Ciphers:       []string{"ECDHE-ECDSA-AES256-GCM-SHA384"},
    Groups:        []configv1.TLSGroup{configv1.TLSGroupSecP256r1, configv1.TLSGroupSecP384r1},
    MinTLSVersion: configv1.VersionTLS12,
}
```

3. **Compare fields individually** instead of `reflect.DeepEqual` on the whole struct, and only assert groups when the feature gate is enabled:
```go
if !cmp.Equal(ic.Status.TLSProfile.Ciphers, customProfileSpec.Ciphers) {
    return false, nil
}
if ic.Status.TLSProfile.MinTLSVersion != customProfileSpec.MinTLSVersion {
    return false, nil
}
if tlsGroupsEnabled && !cmp.Equal(ic.Status.TLSProfile.Groups, customProfileSpec.Groups) {
    return false, nil
}
```

New imports needed: `"github.com/openshift/api/features"` and `"github.com/google/go-cmp/cmp"`. The `"gopkg.in/yaml.v2"` import can be dropped since the error formatting switches from YAML marshaling to `cmp.Diff`.

candita:
This PR is for a release blocking feature.

pkg/operator/controller/ingress/deployment_test.go
candita: I was wrong about that.  The tls profiile groups are types so they won't contain anything unexpected: https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L174-L175.

richardsonnick:
more aws infra issues

melvinjoseph86:
Tested with cluster bot

```
➜  oc get clusterversion
NAME      VERSION                                               AVAILABLE   PROGRESSING   SINCE   STATUS
version   5.0.0-0-2026-07-23-025516-test-ci-ln-82fvx2t-latest   True        False         3h2m    Cluster version is 5.0.0-0-2026-07-23-025516-test-ci-ln-82fvx2t-latest
➜  oc get deployment -n openshift-ingress router-default -o jsonpath='{.spec.template.spec.containers[0].env}' | jq '.[] | select(.name=="ROUTER_CURVES")'
{
  "name": "ROUTER_CURVES",
  "value": "X25519MLKEM768:X25519:secp256r1:secp384r1"
}

 Custom TLS profile with explicit groups
  oc patch ingresscontroller default -n openshift-ingress-operator --type=merge -p '{
    "spec": {
      "tlsSecurityProfile": {
        "type": "Custom",
        "custom": {
          "ciphers": ["ECDHE-ECDSA-AES256-GCM-SHA384"],
          "minTLSVersion": "VersionTLS12",
          "groups": ["secp256r1", "secp384r1"]
        }
      }
    }
  }'
➜   oc get deployment -n openshift-ingress router-default -o jsonpath='{.spec.template.spec.containers[0].env}' | jq '.[] | select(.name=="ROUTER_CURVES")'
{
  "name": "ROUTER_CURVES",
  "value": "secp256r1:secp384r1"
}
```

/verified by @mjoseph

PR #1480, opened 2026-06-19
NE-2796: Respect OpenShift TLS Profiles during Istio/GatewayAPI installation

rikatz:
This change makes Istio installation for Gateway API receive and respect the OpenShift TLS configuration once fetched from apiserver configuration.

The change will add a new watcher on apiservers.config.openshift.io, and generate a proper TLS configuration to be passed to sail-library that will both make Istio/Pilot and the provisioned Gateways to respect the centralized configurations.

Additionally, the changes from the commit https://github.com/openshift-service-mesh/sail-operator/commit/73fe9266d9e0dd7b04db344bf8331d7025bd04aa that fix how sail-library configures the TLS options is versioned on this change while OSSM 3.4.1, that contains this fix is not available.

### DIFF
--- a/pkg/operator/controller/gatewayclass/istio_sail_installer.go
+++ b/pkg/operator/controller/gatewayclass/istio_sail_installer.go
@@ -7,7 +7,9 @@ import (
 	"strings"
 
 	sailv1 "github.com/istio-ecosystem/sail-operator/api/v1"
+	"github.com/istio-ecosystem/sail-operator/pkg/config"
 	"github.com/istio-ecosystem/sail-operator/pkg/install"
+	openshifttls "github.com/openshift/controller-runtime-common/pkg/tls"
 	v1 "k8s.io/api/core/v1"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
@@ -134,10 +136,16 @@ func (r *reconciler) ensureIstio(ctx context.Context, istioVersion string, gatew
 		return fmt.Errorf("error verifying cluster proxy configuration: %w", err)
 	}
 
+	var apiserverConfig configv1.APIServer
+	if err := r.cache.Get(ctx, types.NamespacedName{Name: "cluster"}, &apiserverConfig); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("error verifying apiserver configuration: %w", err)
+	}
+
 	// Build options from current state
 	opts, err := r.buildInstallerOptions(enableInferenceExtension, istioVersion, gatewayclasses, &extraIstioConfig{
-		proxyConfig: &proxyConfig,
-		infraConfig: infraConfig,
+		proxyConfig:     &proxyConfig,
+		infraConfig:     infraConfig,
+		apiserverConfig: &apiserverConfig,
 	})
 	if err != nil {
 		return err
@@ -171,6 +179,14 @@ func (r *reconciler) buildInstallerOptions(enableInferenceExtension bool, istioV
 		return install.Options{}, fmt.Errorf("failed to merge Istio values: %w", err)
 	}
 
+	tlsOptions := &config.OpenShiftTLS{}
+	if extraConfig.apiserverConfig != nil {
+		tlsOptions, err = getTLSOptions(extraConfig.apiserverConfig.Spec.TLSSecurityProfile, extraConfig.apiserverConfig.Spec.TLSAdherence)
+		if err != nil {
+			return install.Options{}, fmt.Errorf("failed to get TLS Profile spec: %w", err)
+		}
+	}
+
 	return install.Options{
 		Namespace:      r.config.OperandNamespace,
 		Revision:       controller.IstioName("").Name,
@@ -178,7 +194,29 @@ func (r *reconciler) buildInstallerOptions(enableInferenceExtension bool, istioV
 		Version:        istioVersion,
 		ManageCRDs:     true,
 		IncludeAllCRDs: true,
+		OpenShiftTLS:   tlsOptions,
 	}, nil
+}
+
+func getTLSOptions(profile *configv1.TLSSecurityProfile, adherence configv1.TLSAdherencePolicy) (*config.OpenShiftTLS, error) {
+	tlsOptions := &config.OpenShiftTLS{}
+	if profile == nil {
+		return tlsOptions, nil
+	}
+
+	tlsspec, err := openshifttls.GetTLSProfileSpec(profile)
+	if err != nil {
+		return tlsOptions, fmt.Errorf("failed to get TLS Profile spec: %w", err)
+	}
+
+	tlsconfig, _ := openshifttls.NewTLSConfigFromProfile(tlsspec)
+
+	tlsOptions.TLSAdherencePolicy = adherence
+	tlsOptions.TLSProfileSpec = tlsspec
+	tlsOptions.TLSConfigFunc = tlsconfig
+
+	return tlsOptions, nil
+
 }
 
 // openshiftValues returns the OpenShift-specific value overrides for Istio.
--- a/test/e2e/gateway_api_tls_scanner_setup_test.go
+++ b/test/e2e/gateway_api_tls_scanner_setup_test.go
@@ -4,7 +4,6 @@
 package e2e
 
 import (
-	"context"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
@@ -89,6 +88,11 @@ func TestGatewayAPITLSScannerSetup(t *testing.T) {
 	_, err = assertGatewaySuccessful(t, gateway.Namespace, gateway.Name)
 	require.NoError(t, err, "Gateway was not accepted/programmed")
 
+	err = assertGatewayInfrastructureLabelsPropagated(t, gateway.Namespace, gateway.Name, string(gateway.Spec.GatewayClassName), map[string]string{
+		"app": gateway.Name,
+	})
+	require.NoError(t, err, "Gateway infrastructure labels were not propagated to Envoy pods")
+
 	err = assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
 		{dnsName: hostname + ".", gatewayName: gateway.Name}: true,
 	})
@@ -120,15 +124,14 @@ func ensureGatewayTLSSecret(t *testing.T, namespace, name, dnsName string) (*cor
 			"tls.key": []byte(keyPEM),
 		},
 	}
-	if err := createOrGetWithRetry(t, context.Background(), secret, DefaultRetryTimeout); err != nil {
+	if err := createOrGetWithRetry(t, t.Context(), secret, DefaultRetryTimeout); err != nil {
 		return nil, fmt.Errorf("failed to create TLS secret %s/%s: %w", namespace, name, err)
 	}
 	return secret, nil
 }
 
-// ensureHTTPSGateway creates a Gateway with an HTTPS listener and an
-// infrastructure "app" label matching the Gateway name so tls-scanner's
-// COMPONENT_FILTER can select the generated Envoy pods.
+// ensureHTTPSGateway creates a Gateway with an HTTPS listener and the given
+// TLS secret, including an infrastructure "app" label matching the Gateway name.
 func ensureHTTPSGateway(t *testing.T, gatewayClassName, name, namespace, hostname, secretName string) (*gatewayapiv1.Gateway, error) {
 	t.Helper()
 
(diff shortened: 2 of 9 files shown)


pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -171,16 +179,48 @@ func (r *reconciler) buildInstallerOptions(enableInferenceExtension bool, istioV
...
+		return tlsOptions, nil
+	}
+
+	tlsspec, err := openshifttls.GetTLSProfileSpec(profile)
+	if err != nil {
+		return tlsOptions, fmt.Errorf("failed to get TLS Profile spec: %w", err)
+	}
+
+	tlsconfig, _ := openshifttls.NewTLSConfigFromProfile(tlsspec)
+
```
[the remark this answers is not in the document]
rikatz: @coderabbitai I don't think I am ignoring an error. The return of NewTLSConfigFromProfile is an array of strings, containing unsupported ciphers, have you really checked it?

pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -171,16 +179,48 @@ func (r *reconciler) buildInstallerOptions(enableInferenceExtension bool, istioV
...
+	tlsOptions.TLSAdherencePolicy = adherence
+	tlsOptions.TLSProfileSpec = tlsspec
+	tlsOptions.TLSConfigFunc = tlsconfig
+
+	logTLS.Info("configuring tls with values", "tls", tlsOptions.TLSProfileSpec)
```
gcs278: *nit* I think this is a bit noisy to log the TLSProfile on every reconcile. Should probably be V(1) or just drop entirely.

```suggestion
```

gcs278 (commented):
This is a pretty straightforward PR that plumbs in the TLS Configuration from the API Server configuration. I cross checked various parts of this implementation to the ingresscontroller implementation and it checks out.

I think are missing E2E though, which concerns me. We should consider adding E2E as part of the tests introduced in https://github.com/openshift/origin/pull/30801. 

Just one nit, otherwise LGTM. I'll approve for now.
/approve

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1508 has merged, this can now have the vendor bump removed as we have used https://github.com/openshift-service-mesh/sail-operator/tree/release-3.4.0-nid.

rikatz:
@gcs278 updated. I agree, we must have a test on openshift/origin, I will raise a story for it

gcs278:
Nice. Took another look and everything looks good. Did some local testing with https://github.com/rikatz/cluster-ingress-operator/commit/256f4a7824347b7d6974f7f884c392c0bdd1c191 as best I could with claude and everything checked out.

As far as I know, this is effectively a NO-OP until https://github.com/openshift/api/pull/2883 is promoted (no adherence yet), which should be soon.
/lgtm

/hold
For @bentito if you want to take a look, and also please wait for https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/5.0.0-0.nightly/release/5.0.0-0.nightly-2026-07-22-160036 to be accepted (OSSM 3.4.0 got reverted last time, so clearing the nightly payload is a good idea to avoid another messy revert).

gcs278:
no concerning failures
/retest

gcs278:
https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/5.0.0-0.ci/release/5.0.0-0.ci-2026-07-22-191603 was accepted with OSSM 3.4.0, so I guess that's a good sign to move on with this.

Unfortunately, merge conflict - probably due to k8s 1.36 vendor bump landing in CIO.

rikatz:
but now I have conflicts to fix :D

gcs278:
No concern, but it would be great to get a early `/testwith` signal with https://github.com/openshift/origin/pull/31415 to prove it's working.

/lgtm

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-ovn-serial-1of2 openshift/api#2883 openshift/origin#31424

test/e2e/gateway_api_tls_scanner_setup_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	t.Logf("Gateway %s/%s is ready for TLS scanning (resources intentionally left in place)", gateway.Namespace, gateway.Name)
+}
+
+// ensureGatewayTLSSecret creates a self-signed TLS secret for the Gateway
+// HTTPS listener if it does not already exist.
+func ensureGatewayTLSSecret(t *testing.T, namespace, name, dnsName string) (*corev1.Secret, error) {
+	t.Helper()
+
+	certPEM, keyPEM, err := generateServerTLSKeyPair(dnsName)
```
gcs278, line 111: *nit* just a suggestion, but we already have `CreateTLSKeyCert` that is mostly the same as the new function you wrote, minus:
- No DNSName SAN - CreateTLSKeyCert puts the name in CommonName only, not DNSNames.
-  RSA vs ECDSA - CreateTLSKeyCert uses RSA 2048.

Not sure if either matter, but if possible it would be nice to reduce the amount of similar helpers like this. Was there a reason you couldn't use this?

test/e2e/gateway_api_tls_scanner_setup_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	err = assertExpectedDNSRecords(t, map[expectedDnsRecord]bool{
+		{dnsName: hostname + ".", gatewayName: gateway.Name}: true,
+	})
+	require.NoError(t, err, "DNSRecord never got ready")
+
+	assertProxyDeployCustomConfigurations(t, gateway.Namespace, gateway.Name, string(gateway.Spec.GatewayClassName))
+
+	t.Logf("Gateway %s/%s is ready for TLS scanning (resources intentionally left in place)", gateway.Namespace, gateway.Name)
+}
```
gcs278, line 104: If `Infrastructure.Labels` bugs out and doesn't propagate the `app` label to the Envoy pods, the tls-scanner will find no matching pods and exit successfully, silently passing without actually scanning Gateway API.

Think it's worth an extra validation that Istio is doing what it's told with `Infrastructure.Labels` before call it "ready"?

gcs278 (commented):
Generally LGTM - no major concerns. I've reviewed a couple times now, and my comments are e2e only.

pkg/operator/controller/gatewayclass/istio_sail_installer.go
rikatz: removed on previous commit

test/e2e/gateway_api_tls_scanner_setup_test.go
rikatz: nope, I asked cursor explicitly to reuse helpers but it decided it would be good to have its own function to not break existing ones. I can fix this

rikatz: (btw I imagine one of the reasons is to not break a public function signature)

rikatz: the change will be slightly bigger here, so I would say to stay as is for now and we improve as a cleanup

test/e2e/gateway_api_tls_scanner_setup_test.go
rikatz: done

test/e2e/gateway_api_tls_scanner_setup_test.go
gcs278: That's fine, thanks!

gcs278:
LGTM. These changes are effectively gated behind the TLSAdherence TechPreview feature gate. The sail library only applies TLS configuration to Envoy when `TLSAdherence: StrictAllComponents` is set, so on GA clusters where that field is not exposed, this code path is a no-op.

/lgtm

test/e2e/gateway_api_tls_scanner_setup_test.go
```diff
@@ -0,0 +1,227 @@
...
+			Organization: []string{"OpenShift E2E Testing"},
+			CommonName:   dnsName,
+		},
+		NotBefore:             time.Now().Add(-time.Hour),
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		KeyUsage:              x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		DNSNames:              []string{dnsName, "localhost"},
+	}
```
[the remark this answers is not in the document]
rikatz: this is totally fine, each CI execution is a prow job that runs, does the test and dies so we don't need a big expiration time

rikatz:
Verification executed:
* launch a cluster with techpreview and TLSAdherence promoted: `launch 5.0,openshift/cluster-ingress-operator#1480,openshift/api#2883 aws,techpreview` 
* Applied the TLSAdherence as StrictAllComponents and TLSProfile to modern
* Created the GatewayClass and the Gateway
* Verified via curl that TLS 1.2 requests were denied, while TLS 1.3 are accepted

melvinjoseph86:
Tested using clsuterbot on 5.0 techpreview  cluster

```
➜  oc get clusterversion
NAME      VERSION                                               AVAILABLE   PROGRESSING   SINCE   STATUS
version   5.0.0-0-2026-07-30-070658-test-ci-ln-tdgmnj2-latest   True        False         116m    Cluster version is 5.0.0-0-2026-07-30-070658-test-ci-ln-tdgmnj2-latest

➜    oc patch apiserver cluster --type merge -p '{
    "spec": {
      "tlsSecurityProfile": {
        "type": "Modern",
        "modern": {}
      },
      "tlsAdherence": "StrictAllComponents"
    }
  }'
apiserver.config.openshift.io/cluster patched

➜  oc get gatewayclass
openshift-default   openshift.io/gateway-controller/v1   True       21m

➜    DOMAIN=$(oc get dns cluster -o jsonpath='{.spec.baseDomain}')
➜   GATEWAY_HOST=$(oc get gateway tls-test-gateway -n openshift-ingress \
    -o jsonpath='{.status.addresses[0].value}')

➜     openssl req -x509 -nodes -days 1 -newkey ec -pkeyopt ec_paramgen_curve:P-256 \
    -keyout /tmp/tls.key -out /tmp/tls.crt \
    -subj "/CN=*.gwapi.${DOMAIN}" -addext "subjectAltName=DNS:*.gwapi.${DOMAIN}"
-----
➜    oc create secret tls test-tls-cert \
    --cert=/tmp/tls.crt --key=/tmp/tls.key \
    -n openshift-ingress --dry-run=client -o yaml | oc apply -f -
secret/test-tls-cert created

➜  cat <<EOF | oc apply -f -
kind: Gateway
metadata:
  name: tls-test-gateway
  namespace: openshift-ingress
spec:
  gatewayClassName: openshift-default
  listeners:
  - name: https
    hostname: "*.gwapi.${DOMAIN}"
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - name: test-tls-cert
    allowedRoutes:
      namespaces:
        from: All
EOF
gateway.gateway.networking.k8s.io/tls-test-gateway configured

➜  openssl s_client -connect "${GATEWAY_HOST}:443" \
    -servername "${LISTENER_HOST}" \
    -tls1_3 </dev/null 2>&1 | grep -E "Protocol|Cipher"
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Protocol: TLSv1.3

➜  curl -vk --tlsv1.3 \
    --resolve "${LISTENER_HOST}:443:${GATEWAY_HOST}" \
    "https://${LISTENER_HOST}:443/" 2>&1 | grep -E "SSL connection|TLSv1.3|HTTP/"

➜   curl -vk --tlsv1.2 --tls-max 1.2 \
    --resolve "${LISTENER_HOST}:443:${GATEWAY_HOST}" \
    "https://${LISTENER_HOST}:443/" 2>&1 | grep -E "alert|error|SSL|handshake|TLS"
* (304) (OUT), TLS handshake, Client hello (1):
* LibreSSL/3.3.6: error:1404B42E:SSL routines:ST_CONNECT:tlsv1 alert protocol version
curl: (35) LibreSSL/3.3.6: error:1404B42E:SSL routines:ST_CONNECT:tlsv1 alert protocol version
```

Hence marking as verified
/verified by @mjoseph

rikatz:
@coderabbitai analyze the failure, is it related with my PR?

rikatz:
@coderabbitai can you analyze past executions of this same job to see if infrastructure issues are consistent?

rikatz:
as per chai-bot analysis, coderabbit analysis and my analysis, hypershift-e2e-aks job is very flaky the last few days and failures are not related withthe current change.

Rather I am overriding this PR to move forward with the PR queue before pencils down

PR #1498, opened 2026-07-02
NE-2219: Select HAProxy version from IngressController API

jcmoraisjr:
Configures the HAProxy sidecar with the image URL representing the HAProxy version configured in the IngressController API.

`image-references` manifest updated with the new HAProxy images, their image names added in the deployment manifest, which configures new command-line options with the new HAProxy image URLs.

The haproxyVersion reading is under the FeatureGateIngressControllerMultipleHAProxyVersions feature gate, and the router image, which also has HAProxy installed, is used if the feature gate is disabled.

https://redhat.atlassian.net/browse/NE-2219


gcs278:
We need to manualk run this CI job:
/test e2e-aws-ovn-techpreview

manifests/02-deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
           - "$(WATCH_NAMESPACE)"
           - --image
           - "$(IMAGE)"
+          - --image-haproxy-previous
+          - "$(IMAGE_HAPROXY_PREVIOUS)"
+          - --image-haproxy-default
+          - "$(IMAGE_HAPROXY_DEFAULT)"
```
gcs278: I have a few concerns about the flag naming and approach:

1. The non-default image could also be used for forward testing a new HAProxy version, so `--image-haproxy-previous` is too specific: it implies a direction that may not always be true.
2. If we need to support more than two versions in the future, we'd have to keep adding new arguments.
3. In the current implementation, `--image-haproxy-default` doesn't actually determine the default - that's hardcoded as `defaultHAProxyVersion` in the code. So the flag name is misleading.

What do you think about using a repeatable key-value flag instead, like `--haproxy-image 2.8=<pullspec> --haproxy-image 3.2=<pullspec>`? It's a bit less user-friendly, but more flexible, adding or removing a version is just a manifest change with no new flags needed. Thoughts? Sorry for another suggestion - I'm iterating on ideas myself.

jcmoraisjr: Good one! Let's give it a try. It uses the former approach, but having the benefit of a fixed command-line option between OCP releases.

jcmoraisjr:
All the failures are infra related flakes. Hypershift tests btw are having quota related issues since a few days. Giving them another chance.

/retest

jcmoraisjr:
According to Hypershift folks, a cleanup was made on leaked resources that should make the e2e tests to work again. Giving another chance to all of them.

/retest

cmd/ingress-operator/start.go
```diff
@@ -97,6 +102,9 @@ func NewStartCommand() *cobra.Command {
 	if err := cmd.MarkFlagRequired("image"); err != nil {
 		panic(err)
 	}
+	if err := cmd.MarkFlagRequired("haproxy-image"); err != nil {
+		panic(err)
+	}
```
gcs278: Here are a couple of concerns with marking `--haproxy-image` as required:

1. This is a tech preview feature, so requiring the flag means mandating an input even for clusters that don't have the feature gate enabled.
2. For HyperShift, we need to coordinate adding the new flag to its CIO deployment. Making it optional with a fallback to the router image gives us a window to land the CIO changes first without immediately breaking HyperShift, while they work on adding the flag on their side.

cmd/ingress-operator/start.go
```diff
@@ -82,6 +86,7 @@ func NewStartCommand() *cobra.Command {
 
 	cmd.Flags().StringVarP(&options.OperatorNamespace, "namespace", "n", operatorcontroller.DefaultOperatorNamespace, "namespace the operator is deployed to (required)")
 	cmd.Flags().StringVarP(&options.IngressControllerImage, "image", "i", "", "image of the ingress controller the operator will manage (required)")
+	cmd.Flags().StringArrayVarP(&options.HAProxyImage, "haproxy-image", "", nil, "HAProxy images declared as '<version>=<image-url>'. Add a new parameter for every new image (required)")
```
gcs278: *nit* was there a reason you didn't use `StringToStringVarP` which also support commas `key=value,key=value` and makes the parsing a bit easier?

```suggestion
	cmd.Flags().StringToStringVarP(&options.HAProxyImages, "haproxy-image", "", nil, "HAProxy images as version=pullspec")
```

Then parsing would be something like:
```
haproxyImages := make(map[operatorv1.HAProxyVersion]string)
  for version, image := range opts.HAProxyImages {
      haproxyImages[operatorv1.HAProxyVersion(version)] = image
  }
```

But correct me if I'm missing something, I could be wrong.

test/e2e/maxconn_test.go
```diff
@@ -138,7 +139,7 @@ func TestTunableMaxConnectionsInvalidValues(t *testing.T) {
 		{"set maxconn 2000001", 2000001},
 	} {
 		t.Log(testCase.description)
-		const expectedErr string = `"spec.tuningOptions" must validate at least one schema (anyOf), spec.tuningOptions.maxConnections`
+		expectedErr := fmt.Sprintf("Invalid value: %d: maxConnections must be 0, -1, or between 2000 and 2000000", testCase.maxConnections)
```
gcs278, line 142: oh interesting that you had to adapt to https://github.com/openshift/api/pull/2861 in the bump. Looks good though.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -830,10 +832,25 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		}
+		var found bool
+		haproxyImage, found = config.HAProxyImages[haproxyVersion]
+		if !found {
+			// sanity check, an invalid version is prevented by API validation
+			return nil, fmt.Errorf("ingresscontroller %q has invalid spec.haproxyVersion: %v", ci.Name, ci.Spec.HAProxyVersion)
+		}
+	} else {
+		// use the router image if feature gate is disabled.
+		// Router image has HAProxy installed.
```
gcs278: to be clear, this is only until we "clean up" the router image (remove HAProxy), after the FG gets promoted. Might be good to track any post-GA cleanup this as a story so we don't forget.

pkg/operator/controller/ingress/status.go
```diff
@@ -103,6 +103,20 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
...
+		haproxyVersion = ic.Spec.HAProxyVersion
+		if haproxyVersion == "" {
+			haproxyVersion = defaultHAProxyVersion
+		}
+		if _, found := r.config.HAProxyImages[haproxyVersion]; !found {
+			// sanity check, an invalid version is prevented by API validation
+			haproxyVersion = ""
+		}
+	}
+	updated.Status.EffectiveHAProxyVersion = haproxyVersion
```
gcs278, line 111: *nit* the existing style in this `syncIngressControllerStatus` function has generally been to make this logic as a helper function to help keeping it from getting to unwieldy.

It could be something like this could be helpful:
`updated.Status.EffectiveHAProxyVersion = getEffectiveHAProxyVersion(ic)` and you could possibly share with `deployment.go` depending on how you want to handle errors.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -165,6 +165,8 @@ const (
 	//
 	// NOTE: it is hardcoded in the deployment.yaml manifest, update there if updating here.
 	routerHAProxyAdminSocket = "/var/lib/haproxy/run/admin.sock"
+
+	defaultHAProxyVersion = operatorv1.HAProxyVersion32
```
gcs278: I also think this should be an argument: `--default-haproxy-version`. Here's a situation why:

HyperShift updates its CIO deployment arguments separately from us. When we drop 2.8 and add 3.4 as the new default in 5.1, HyperShift can't make that change atomically: it requires multiple PRs. Adding `--default-haproxy-version` lets HyperShift update the images and the default together in a single PR, avoiding a CI implosion.

You can still default the argument to `operatorv1.HAProxyVersion32` if not provided, but having it will make for more flexible transitions with hypershift.

gcs278 (commented):
Wow - it's super satisfying to see how simple it is to implement this entire feature - no complicated defaulting logic. I'm surprised as I didn't think this would be that easy, but I think a well written EP made this a very easy review.

Nearly LGTM, just a couple of minor details and nits.

gcs278:
[Tech Preview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1498/pull-ci-openshift-cluster-ingress-operator-master-e2e-vsphere-static-metallb-operator-gwapi-techpreview/2074491246046351360) job are showing that this is working, I see this in the logs:
```
2026-07-07 14:21:04 [NOTICE]   (4) : haproxy version is 3.2.19-06868cd6e
```

gcs278:
Worth kicking off an upgrade TP job:

/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview-upgrade

@jcmoraisjr if you do push before this is finished, you can run `/payload abort` to stop it, and start it again with the new changes.

gcs278:
@jcmoraisjr I have updated https://github.com/openshift/release/pull/80171 to add `haproxy28` TP jobs so that we can test the alternate/previous version before merging. I will need your review on this. I've triggered a rehearsal.

Ultimately, we will want nightly jobs to test the previous (HAProxy 2.8) version, but I don't think it makes sense to add nightlys until we go GA (TP would pretty quickly be pointless). This first testing PR is a first step towards complete CI coverage.

manifests/02-deployment.yaml
```diff
@@ -81,6 +85,10 @@ spec:
                   fieldPath: metadata.namespace
             - name: IMAGE
               value: openshift/origin-haproxy-router:v4.0
+            - name: HAPROXY_28_IMAGE
+              value: quay.io/gspence/haproxy-router-haproxy28:latest
```
gcs278: Actually, is there a reason we can't go ahead and start using art's HAProxy 28 image in this PR? I noticed that's how https://github.com/openshift/cluster-ingress-operator/pull/1499 is working, and it seems like we are just waiting on ART for HAProxy 3.2.

cmd/ingress-operator/start.go
jcmoraisjr: Sure, changing `--haproxy-image` to optional instead. We can revisit this after adding the first Hypershift configuration.

cmd/ingress-operator/start.go
jcmoraisjr: Ow nice I wasn't aware of `StringToStringVarP` variant. Just moved to it, which seems to work really nice on my local tests.

pkg/operator/controller/ingress/status.go
jcmoraisjr: I just did it a try, found a common implementation for both spec and status having a single `getEffectiveHAProxyVersion()`, what do you think about the approach?

pkg/operator/controller/ingress/deployment.go
jcmoraisjr: Yep I wasn't with the Hypershift way in mind, just added a new and optional `--default-haproxy-version`.

manifests/02-deployment.yaml
jcmoraisjr: As we talked, although we're already pushing both `haproxy28` and `haproxy32`, neither of them are part of the nightly yet, so some of the tests should fail. The idea for now is to merge #1499 just with the image-references update, giving us the ability to create new clusters with this PR using the embedded images.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1385,6 +1391,39 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
 
+func getEffectiveHAProxyVersionAndImage(ic *operatorv1.IngressController, config *Config) (operatorv1.HAProxyVersion, string, error) {
+	if !config.FeatureMultiHAProxyEnabled {
+		// no effective version, and using router image in case FG is not enabled
+		return "", config.IngressControllerImage, nil
+	}
+
+	if len(config.HAProxyImages) == 0 {
+		// images map is currently optional, using router image as fallback
+		return operatorv1.HAProxyVersion28, config.IngressControllerImage, nil
```
gcs278: This tripped me up a bit. I realize that: if no `--haproxy-images` are provided, then we are forced to fallback on to the router image (there is no other option available). So it makes sense.

It's a nit pick, but I'd probably just lean towards removing the HAProxyVersion from this all together and return `""`. The version hardcoded here probably will go stale soon - since I want to merge https://github.com/openshift/router/pull/792 which bumps the router container to HAProxy 32 (to get the ball moving on HAProxy 32 CI exposure).

But either way, it's fine.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1385,6 +1391,39 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+	}
+
+	// version is user provided, fallback to the default one if unset/empty/missing, default can be overridden via command-line option
+	version := ic.Spec.HAProxyVersion
+	if version == "" {
+		version = operatorv1.HAProxyVersion(config.DefaultHAProxyVersion)
```
gcs278: *nit* unnecessary conversion.

```suggestion
		version = config.DefaultHAProxyVersion
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1385,6 +1391,39 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
+		if version == "" {
+			version = defaultHAProxyVersion
+		}
+	}
+
+	image, found := config.HAProxyImages[version]
+	if !found {
+		if ic.Spec.HAProxyVersion != "" {
+			// user provided an invalid one
+			return "", "", fmt.Errorf("ingresscontroller %q has invalid spec.haproxyVersion: %v", ic.Name, ic.Spec.HAProxyVersion)
```
gcs278: *nit* the word "invalid" is a bit generic. Arguably the API ensures all values are valid, but the core issue here is that the total set of images haven't been provided.

But it's not a big deal - this really should never happen. Maybe more helpful for future developers to understand the context.

```suggestion
			return "", "", fmt.Errorf("haproxy image for version %q is not available for ingresscontroller %q; ensure the operator is configured with --haproxy-image for this version", ic.Spec.HAProxyVersion, ic.Name)
```

manifests/02-deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
           - "$(WATCH_NAMESPACE)"
           - --image
           - "$(IMAGE)"
+          - --haproxy-image
+          - "2.8=$(HAPROXY_28_IMAGE)"
+          - --haproxy-image
+          - "3.2=$(HAPROXY_32_IMAGE)"
```
gcs278, line 66: It's another minor point, but what about:
1. Always providing `--default-haproxy-version` here
2. Making `--default-haproxy-version` required only when `--haproxy-image` is provided
3. Removing the hardcoding fallback logic for HAProxy32 (sorry I know, I previously suggested this)

This would do two things:
1. Remove hardcoded defaults in the code (given you agree with my other comment about making the fall back to `""` rather than HAProxy 2.8)
2. Force configuring the haproxy images and the default version all in the same place as variables/arguments for better visibility.

It still supports the hypershift overlap period as they won't be providing `--haproxy-image` at all, so `--default-haproxy-version` not required.

Thoughts?

gcs278 (commented):
Nearly LGTM. just a couple more minor details I wanted to point out. Thanks!

manifests/02-deployment.yaml
gcs278: Well, I just also realized we may still have `upgradeable=false` logic still hardcoded to HAProxy28 in the code (that _could_ be another argument, but that's a bit awkward as we/NID Team want to declare it, not allow people to change the deprecated version via arguments).

gcs278:
Now let's test our previous haproxy 28 jobs on this PR to simulate a pinned cluster:

/test e2e-aws-operator-haproxy28-techpreview
/test e2e-aws-ovn-haproxy28-techpreview
/test e2e-aws-ovn-serial-haproxy28-techpreview-1of2
/test e2e-aws-ovn-serial-haproxy28-techpreview-2of2
/test e2e-aws-ovn-upgrade-haproxy28-techpreview

manifests/02-deployment.yaml
jcmoraisjr: Hardcoding the default in the manifest wasn't my first option, but in the end it makes the rest of the validations easier and more predictable. Good approach, just applied.

pkg/operator/controller/ingress/deployment.go
jcmoraisjr: Yea I'm still missing a linter in my IDE

manifests/02-deployment.yaml
jcmoraisjr: > we may still have upgradeable=false logic still hardcoded to HAProxy28

Just added [NE-2809](https://redhat.atlassian.net/browse/NE-2809) so we don't forget to think about it.

gcs278:
@jcmoraisjr unfortunately something didn't work with our haproxy28 jobs. I see [evidence](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1498/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator-haproxy28-techpreview/2075336481601228800/artifacts/e2e-aws-operator-haproxy28-techpreview/gather-extra/artifacts/inspect/namespaces/openshift-ingress-operator/operator.openshift.io/ingresscontrollers/default.yaml) that the spec.haproxyVersion never made it to the default IngressController object and is still set to `3.2`.

I need to figure this one out. Also, I just remember a limitation of these `haproxy28` tests - it only sets `2.8` for the default ingressController. For all tests that use a custom IngressController (I think most of them) they will use `3.2` since there isn't a way to change the global default. We will need to think about this, it may change our testing approach - it does lessen the value of the `haproxy28` jobs.

pkg/operator/controller/ingress/status.go
```diff
@@ -103,6 +103,13 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 		updateIngressControllerFloatingIPOpenStackStatus(updated, service)
 	}
 
+	haproxyVersion, _, err := getEffectiveHAProxyVersionAndImage(ic, &r.config)
```
gcs278: Looks like `IngressStatusesEqual` needs to be updated and unit test added for `IngressStatusesEqual` changes.

jcmoraisjr: Good catch, just updated.

gcs278:
Thanks for all of the quick updates!

I think it's in good shape. We added new arguments, but left existing usages compatible (for hypershift) to not instantly break when we promote this. This is a pretty straightforward implementation.

I will note that we are lacking E2E. It's gated under TP, so if you wan to merge, I think that's okay. My personal recommendation (especially since we are under `acknowledge-critical-fixes-only`) is to wait for @rhamini3's origin E2E PR, and run a test with to make sure everything is working in E2E before merging this. We are ultimately gated on promotion by tests merging, not the implementation merging, so merging this PR now isn't going to speed up the promotion.

/approve
/lgtm

gcs278:
Nice, thanks for adding that annotation so we can fix the haproxy28 tests.

/lgtm

gcs278:
Let's see if https://github.com/openshift/release/pull/81980 works as expected:

/test e2e-aws-operator-haproxy28-techpreview
/test e2e-aws-ovn-haproxy28-techpreview
/test e2e-aws-ovn-serial-haproxy28-techpreview-1of2
/test e2e-aws-ovn-serial-haproxy28-techpreview-2of2
/test e2e-aws-ovn-upgrade-haproxy28-techpreview

gcs278:
Nice - it the haproxy28 jobs are working (grabbed this from one of the must gathers):
```
haproxy version is 2.8.18-ae90be6
```

@jcmoraisjr sorry you will need to rebase due to https://github.com/openshift/cluster-ingress-operator/pull/1507

gcs278:
@jcmoraisjr The haproxy28 jobs are working pretty well except there is a bug in the upgrade jobs. The installation step tries to run on the "pre-upgrade" cluster which doesn't have the `status.haproxyVersion` that I'm waiting for. So, i need to skip that verification so we set the annotation anyways for upgrade jobs.

Fix is here: https://github.com/openshift/release/pull/82032

rhamini3:
tested on a 5.0 TP cluster
Set the version as 3.2
```
% oc patch ingresscontroller default -n openshift-ingress-operator --type=merge -p '{"spec":{"haproxyVersion":"3.2"}}'
ingresscontroller.operator.openshift.io/default patched
% oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.spec.haproxyVersion}
3.2
% oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.status.effectiveHAProxyVersion}                  
3.2
% oc -n openshift-ingress exec -c haproxy deploy/router-default -- bash -c "echo show version |socat - /var/lib/haproxy/run/haproxy.sock"
3.2.19-06868cd6e
```
Set the version as 2.8
```
% oc patch ingresscontroller default -n openshift-ingress-operator --type=merge -p '{"spec":{"haproxyVersion":"2.8"}}'
ingresscontroller.operator.openshift.io/default patched
% oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.spec.haproxyVersion}                                      
2.8
 % oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.status.effectiveHAProxyVersion}
2.8
% oc -n openshift-ingress exec -c haproxy deploy/router-default -- bash -c "echo show version |socat - /var/lib/haproxy/run/haproxy.sock"
2.8.18-ae90be6
```

unset the version, the router is running 3.2
```
% oc patch ingresscontroller default -n openshift-ingress-operator --type=merge -p '{"spec":{"haproxyVersion":null}}'                    
ingresscontroller.operator.openshift.io/default patched
% oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.spec.HAProxyVersion}
%
% oc -n openshift-ingress-operator get ingresscontroller default -o jsonpath={.status.effectiveHAProxyVersion}
3.2
% oc -n openshift-ingress exec -c haproxy deploy/router-default -- bash -c "echo show version |socat - /var/lib/haproxy/run/haproxy.sock"
3.2.19-06868cd6e
```

Ran all Critical HaProxy tests and got pass

/verified by @rhamini3

gcs278:
ah prow died?

gcs278:
@jcmoraisjr I saw that @rhamini3's PR had a testwith [success](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-origin-31392-openshift-cluster-ingress-operator-1498-e2e-gcp-ovn-techpreview/2077860347825360896) (failed overall but passed his WIP tests).

I think that's good enough signal to start moving on getting this merged. What do you think? We are going to be fight CI for a while, so might as well start now.

/retest

jcmoraisjr:
@gcs278 fully agreed, most of the concern is adding regressions and we should be good on that regard. Any other improvements we eventually find can be added on a followup PR.

Thanks for all the tests! Really helpful, really appreciated.

gcs278:
infra pod scheduling failures:
/retest

gcs278:
Took another quick look, still looks good to me!

/lgtm

gcs278:
Looks like CI in rough shape after the rebase...

gcs278:
A lot of failures that are due to waiting for an update from [ART](https://redhat-internal.slack.com/archives/CB95J6R4N/p1784288935513879). Should be fixed hopefully soon.

jcmoraisjr:
I can see some jobs succeeding, how are them today?

/retest

gcs278:
@jcmoraisjr looks like e2e-aws-ovn [failed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1498/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn/2079179457007980544) on:
```
: [sig-network][Feature:Router] The HAProxy router should expose prometheus metrics for a route [apigroup:route.openshift.io] [Suite:openshift/conformance/parallel] expand_less	20s
{  fail [github.com/openshift/origin/test/extended/router/metrics.go:227]: Expected
    <[]float64 | len:0, cap:0>: nil
to consist of
    <[]*matchers.BeNumericallyMatcher | len:1, cap:1>: [
        {Comparator: ">", CompareTo: [<int>0]},
    ]
the missing elements were
    <[]*matchers.BeNumericallyMatcher | len:1, cap:1>: [
        {Comparator: ">", CompareTo: [<int>0]},
    ]}
```

But i see that this is already a [flake](https://sippy.dptools.openshift.org/sippy-ng/tests/5.0/analysis?test=%5Bsig-network%5D%5BFeature%3ARouter%5D%20The%20HAProxy%20router%20should%20expose%20prometheus%20metrics%20for%20a%20route%20%5Bapigroup%3Aroute.openshift.io%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fparallel%5D&filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22%5Bsig-network%5D%5BFeature%3ARouter%5D%20The%20HAProxy%20router%20should%20expose%20prometheus%20metrics%20for%20a%20route%20%5Bapigroup%3Aroute.openshift.io%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fparallel%5D%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22has%20entry%22%2C%22value%22%3A%22never-stable%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22has%20entry%22%2C%22value%22%3A%22aggregated%22%7D%5D%2C%22linkOperator%22%3A%22and%22%7D). I don't think it's related to this PR as it's already flakey. We should keep an eye on it though, the only real change here is the defaulting to haproxy 3.2 (which could possibly have some sort of impact, but feels unlikely).

/test e2e-aws-ovn

Hypershift failure doesn't seem related.
/test e2e-aws-ovn-hypershift-conformance

gcs278:
There's some sort of infra failure happenign

gcs278:
should be fixed
/retest

gcs278:
This test has been running for nearly 6 hours stuck on a teardown 

/test e2e-aws-ovn-hypershift-conformance

jcmoraisjr:
Hypershift cluster timed out to be created, let's give it another try.

/test e2e-aws-ovn-hypershift-conformance

gcs278:
@jcmoraisjr I think we might just want to override e2e-aws-ovn-hypershift-conformance. It's holding this PR up.

I did an analysis with claude, and it hasn't passed in [3 days in 20 runs or so](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-hypershift-conformance):

| Category                                                                                      | Count | %   |
| --------------------------------------------------------------------------------------------- | ----- | --- |
| **ImagePullBackOff** — `quay-proxy` returns `manifest unknown` for CI images                  | 12    | 60% |
| **Test failure** — varied conformance flakes (OAuthServer, EgressFirewall, RouteExternalCert) | 3     | 15% |
| **HostedCluster timeout** — hosted cluster never reaches `Available` within 30m               | 2     | 10% |
| **Payload image missing** — `release-latest` build fails, image not found after 5 retries     | 2     | 10% |
| **Lease exhaustion** — no `hypershift-aws-quota-slice` available                              | 1     | 5%  |

The biggest being the the image pull backoff, which is mostly fixed now. but everything else is unrelated failures, and we did get some decent signal when it installed successfully, but failed on the other tests. Our implementation is behind a FG anyways and backwards compatible. There is very little risk here and hypershift has been painfully flakey. Any concern with overriding?

gcs278:
We're having various issues with hypershift, having 3 days of solid failures. This PR is very low risk as the logic is mostly gated behind a TP FG. We have previously ensured that hypershift has passed enough tests, for any incompatibilities we might of accidentally introduced with the new CIO arguments. This [run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1498/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-hypershift-conformance/2079390485750747136) got the furthest, but failed on two PLR and NetworkPoilcy tests. That's suggests we have passing tests, and it looks like unrelated flakes.

I feel confident enough to suggest an override.

/override-sticky ci/prow/e2e-aws-ovn-hypershift-conformance

PR #1500, opened 2026-07-07
NE-2388: aws nlb security groups

aswinsuryan:
Adds implementation for the new IngressController AWS NLB security groups API. It takes the security groups specified on the IngressController and propagates them to the service.beta.kubernetes.io/aws-load-balancer-security-groups annotation on the LB-type Service.

When security groups are updated in the IngressController spec, the operator deletes and recreates the LB-type Service to effectuate the change.

This change is being introduced under the Tech Preview IngressControllerLBSecurityGroupsAWS feature gate and will be later promoted to GA.


pkg/operator/controller/ingress/load_balancer_service.go
```diff
@@ -1273,6 +1344,41 @@ func awsEIPAllocationsEqual(eipAllocations1, eipAllocations2 []operatorv1.EIPAll
 	return true
 }
 
+// awsSecurityGroupsEqual compares two SecurityGroupID slices and returns a boolean
+// whether they are equal or not. The order of the security groups is ignored.
+func awsSecurityGroupsEqual(securityGroups1, securityGroups2 []operatorv1.SecurityGroupID) bool {
```
gcs278, line 1349: just fyi, this has a bug with:
`["sg-a","sg-a"] vs ["sg-a","sg-b"] ` returning true.

However, the API prevents duplicate sgs. So, it's not reachable. You can fix if you want and add a unit test, but not required.

pkg/operator/controller/ingress/load_balancer_service_test.go
```diff
@@ -2113,3 +2206,160 @@ func Test_shouldRecreateLoadBalancer(t *testing.T) {
 		})
 	}
 }
+
+func TestAWSSecurityGroupsEqual(t *testing.T) {
```
gcs278, line 2210: *nit* this is a fine test, but also we should probably add some test cases in Test_IngressStatusEqual function (you'll notice EIP & Subnets test cases in there too) for completeness.

gcs278 (commented):
Very nice solidly written PR!

It was hard to fine anything to comment on - I really appreciate you using the helper functions and mirroring existing patterns with EIP & Subnets. It's makes the code review & risk much smaller.

I see no major issues. Just nit picks.

pkg/operator/controller/ingress/load_balancer_service.go
aswinsuryan: Done.

pkg/operator/controller/ingress/load_balancer_service_test.go
aswinsuryan: Done.

gcs278:
Thanks! 

/lgtm

I also realized you will need to add also 5 E2E tests in either openshift/origin, or, if possible, use the OTE https://github.com/openshift/cluster-ingress-operator/tree/master/tests-extension framework to put it in this repo. These can be very simple tests, like:
1. Create LB with security group
2. Update a security group
3. Remove a security group
4. Ensure status == spec
5. Unmanaged annotation preserved

Or anything even simpler. Unfortunately, our local CIO e2e doesn't count for this. It'd be super nice if we could use the existing helper functions in the OTE, but I'm not sure if that's possible. We can talk more later.

aswinsuryan:
> Thanks!
> 
> /lgtm
> 
> I also realized you will need to add also 5 E2E tests in either openshift/origin, or, if possible, use the OTE https://github.com/openshift/cluster-ingress-operator/tree/master/tests-extension framework to put it in this repo. These can be very simple tests, like:
> 
> 1. Create LB with security group
> 2. Update a security group
> 3. Remove a security group
> 4. Ensure status == spec
> 5. Unmanaged annotation preserved
> 
> Or anything even simpler. Unfortunately, our local CIO e2e doesn't count for this. It'd be super nice if we could use the existing helper functions in the OTE, but I'm not sure if that's possible. We can talk more later.

Thanks for pointing it out, I added the tests in OTE.

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	enabled, err := cl.isFeatureGateEnabled(ctx, featureGateLBSecurityGroupsAWS)
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to check IngressControllerLBSecurityGroupsAWS feature gate")
+	if !enabled {
+		g.Skip("test skipped because IngressControllerLBSecurityGroupsAWS feature gate is disabled")
+	}
+	return infra
+}
+
+var _ = g.Describe("[sig-network][OCPFeatureGate:IngressControllerLBSecurityGroupsAWS][Feature:Router][apigroup:operator.openshift.io] AWS NLB security groups", func() {
```
gcs278:
```suggestion
var _ = g.Describe("[sig-network-edge][OCPFeatureGate:IngressControllerLBSecurityGroupsAWS][Feature:Router][apigroup:operator.openshift.io] AWS NLB security groups", func() {
```

tests-extension/test/lb_security_groups_util.go
```diff
@@ -0,0 +1,487 @@
...
+	// featureGateLBSecurityGroupsAWS gates the securityGroups NLB API.
+	featureGateLBSecurityGroupsAWS configv1.FeatureGateName = "IngressControllerLBSecurityGroupsAWS"
+)
+
+// clients bundles the typed clients used by the security-group tests.
+type clients struct {
+	kube     kubernetes.Interface
+	config   configclient.Interface
+	operator operatorclient.Interface
+}
```
gcs278: what do you think about making this generic in a `util.go` or something common? along with any function that could be re-useable in this file?

I think you still keep this file for security groups specific functions (if there is enough utils left to justify). We just need to establish some commonality in OTE test patterns, otherwise we might have people use different frameworks and libraries, leading to unnecessary deltas in testing style.

tests-extension/test/lb_security_groups_util.go
```diff
@@ -0,0 +1,487 @@
...
+	autoDeleteLoadBalancerAnnotation = "ingress.operator.openshift.io/auto-delete-load-balancer"
+	// loadBalancerProgressingConditionType is the IngressController condition set when the LB Service must be recreated.
+	loadBalancerProgressingConditionType = "LoadBalancerProgressing"
+	// featureGateLBSecurityGroupsAWS gates the securityGroups NLB API.
+	featureGateLBSecurityGroupsAWS configv1.FeatureGateName = "IngressControllerLBSecurityGroupsAWS"
+)
+
+// clients bundles the typed clients used by the security-group tests.
+type clients struct {
+	kube     kubernetes.Interface
```
gcs278: Thoughts on just using controller-runtime instead of individual clientsets? that's what we do for our local `test/e2e`, and i double checked, the test extension vendor directory will only grow ~10mb.

We do have to watch for vendor bloat (will slow down our builds), but using controller-runtime might make it easier to port over helpers, and more familiar.

It's possible that @melvinjoseph86, @rikatz  or @Miciah might have an opinion on this.

tests-extension/test/lb_security_groups_util.go
```diff
@@ -0,0 +1,487 @@
...
+		awsconfig.WithCredentialsProvider(awscreds.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+	return ec2.NewFromConfig(cfg), nil
+}
+
+// getVPCID returns the VPC ID of the cluster.
+func getVPCID(ctx context.Context, ec2Client *ec2.Client, clusterName string) (string, error) {
```
gcs278: this could be separated into a `aws_util.go` file. Basically, I think we can mostly mirror the patterns in `test/e2e` as `test/e2e/aws_util_test.go` exists.

tests-extension/test/lb_security_groups_util.go
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// updateSecurityGroups updates the IngressController spec securityGroups, retrying on conflict.
+func (c *clients) updateSecurityGroups(ctx context.Context, name string, securityGroups []operatorv1.SecurityGroupID, autoDelete bool, timeout time.Duration) error {
+	return wait.PollUntilContextTimeout(ctx, 2*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		ic, err := c.operator.OperatorV1().IngressControllers(icNamespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+		nlbParams := ic.Spec.EndpointPublishingStrategy.LoadBalancer.ProviderParameters.AWS.NetworkLoadBalancerParameters
```
gcs278: *nit* might want to protect against nil pointer exceptions. I don't think it's an issue now, but may be a foot gun for future function re-use. Maybe use `getSpecSecurityGroups` if possible?

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,199 @@
...
+	defer g.GinkgoRecover()
```
gcs278: @coderabbitai is this unnecessary? I think this is in the wrong place as gets called when the gingko tree-building closure returns at suite setup.

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,199 @@
...
+				o.Expect(deleteSecurityGroup(ctx, ec2Client, sgID, sgDeleteTimeout)).To(o.Succeed())
+			})
+			g.DeferCleanup(func(ctx context.Context) {
+				if sgID2 != "" {
+					o.Expect(deleteSecurityGroup(ctx, ec2Client, sgID2, sgDeleteTimeout)).To(o.Succeed())
+				}
+			})
+
+			ic := newNLBIngressController(icName, icName+"."+domain, []operatorv1.SecurityGroupID{operatorv1.SecurityGroupID(sgID)})
+			_, err = cl.operator.OperatorV1().IngressControllers(icNamespace).Create(ctx, ic, metav1.CreateOptions{})
```
gcs278: *nit* I would consider pulling in `createWithRetryOnError` for object creation.

gcs278 (commented):
Nice tests, great helpers, very readable.

Overall, I don't have any major issues. However, you are the first one introducing new OTE tests, which is great! But I feel like we need to establish some patterns for helper functions so we don't have an exploding OTE directory with everyone writing tests in their own way.

I don't think it needs to be perfect, as it's very likely we will refactor OTE patterns in future PRs. IMO we should consider combining `go.mod`s so we can make test helpers common. But that's out of scope for this PR.

tests-extension/test/lb_security_groups.go
gcs278: coderabbit agreed - you don't need this.
```suggestion
```

gcs278:
@aswinsuryan see failures in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1500/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/2094524314232557568

gcs278:
Maybe a infra issue?
/test verify-deps

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,173 @@
...
+	infra, err := cl.getInfrastructure(ctx)
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to get infrastructure config")
+	if infra.Status.PlatformStatus == nil || infra.Status.PlatformStatus.Type != configv1.AWSPlatformType {
+		g.Skip("test skipped on non-AWS platform")
+	}
+	enabled, err := cl.isFeatureGateEnabled(ctx, featureGateLBSecurityGroupsAWS)
+	o.Expect(err).NotTo(o.HaveOccurred(), "failed to check IngressControllerLBSecurityGroupsAWS feature gate")
+	if !enabled {
+		g.Skip("test skipped because IngressControllerLBSecurityGroupsAWS feature gate is disabled")
+	}
```
gcs278: this is not required, and actually will likely cause it to break once the feature gate is removed. 

the `[OCPFeatureGate:IngressControllerLBSecurityGroupsAWS]` in the test name is already going to make sure it only runs on TP clusters, but also it ensures that it still runs after the FG goes default & gets removed.

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,173 @@
...
+	return infra
+}
+
+// These specs provision AWS Network Load Balancers and create non-default
+// IngressControllers whose per-router NetworkPolicy uses named ports. They are
+// marked [Serial] so they run after the parallel conformance phase, which keeps
+// the transient router NetworkPolicy from tripping the parallel
+// "[sig-network] NetworkPolicy should use numeric ports" platform check.
```
gcs278: I don't think this is the right approach. This is a legit bug: https://redhat.atlassian.net/browse/OCPBUGS-99494 that we need to fix. So problems with marking as serial:
1. It's intentionally avoiding tripping on a known OCPBUGs
2. Adding serial tests is getting expensive - they already shard to 1of1 and 1of2, and we need to be careful to only add required serial test

Also, be aware you tested `/test e2e-aws-ovn-techpreview` - that's the parallel job - I don't see the tests running in the [results](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1500/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/2095174393041784832) which means it's working, but serial jobs are `/test e2e-aws-ovn-serial-1of2`. It actually looks like we have no techpreview serial pre-submit jobs - I think you can use `/payload-job` but even then, i'm not [seeing](https://sippy.dptools.openshift.org/sippy-ng/jobs/5.1?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522variants%2522%252C%2522operatorValue%2522%253A%2522has%2520entry%2522%252C%2522value%2522%253A%2522never-stable%2522%252C%2522not%2522%253Atrue%257D%252C%257B%2522id%2522%253A99%252C%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522techpreview-1of2%2522%257D%255D%257D&sort=asc&sortField=net_improvement) any vanilla AWS jobs for techpreview + serial.

What happens when we put back to serial? Is this a new failure? If so, I wonder why it's not already failing for other parallel custom-IC tests.

tests-extension/test/util.go
```diff
@@ -0,0 +1,259 @@
...
+	return dns.Spec.BaseDomain, nil
+}
+
+// isFeatureGateEnabled returns whether the named feature gate is enabled for the cluster's desired version.
+func (c *clients) isFeatureGateEnabled(ctx context.Context, name configv1.FeatureGateName) (bool, error) {
+	cv := &configv1.ClusterVersion{}
+	if err := c.client.Get(ctx, crclient.ObjectKey{Name: "version"}, cv); err != nil {
+		return false, fmt.Errorf("failed to get cluster version: %w", err)
+	}
+	desiredVersion := cv.Status.Desired.Version
```
gcs278: if you plan to keep this, consider making this a helper that polls like we have:  https://github.com/openshift/cluster-ingress-operator/blob/c0cf37d73a15c1f5dae6978ff83fad711717cd9f/test/e2e/util_test.go#L1286

tests-extension/test/util.go
```diff
@@ -0,0 +1,259 @@
...
+	if len(desiredVersion) == 0 && len(cv.Status.History) > 0 {
+		desiredVersion = cv.Status.History[0].Version
+	}
+
+	fg := &configv1.FeatureGate{}
+	if err := c.client.Get(ctx, crclient.ObjectKey{Name: "cluster"}, fg); err != nil {
```
gcs278: if you keep this, consider making this poll for resiliency: https://github.com/openshift/cluster-ingress-operator/blob/c0cf37d73a15c1f5dae6978ff83fad711717cd9f/test/e2e/util_test.go#L1261

tests-extension/test/util.go
```diff
@@ -0,0 +1,259 @@
...
+
+// newNLBIngressController builds an external NLB IngressController with the given security groups.
+func newNLBIngressController(name, domain string, securityGroups []operatorv1.SecurityGroupID) *operatorv1.IngressController {
+	var nlbParams *operatorv1.AWSNetworkLoadBalancerParameters
+	if len(securityGroups) > 0 {
+		nlbParams = &operatorv1.AWSNetworkLoadBalancerParameters{SecurityGroups: securityGroups}
+	}
+	return &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{Namespace: icNamespace, Name: name},
+		Spec: operatorv1.IngressControllerSpec{
```
gcs278, line 108: consider setting `replicas: 1`, i think it defaults to 2, and it's unnecessarily using more resources than the test needs in CI.

tests-extension/test/lb_security_groups.go
gcs278, line 1: @melvinjoseph86 @rhamini3 (can't remember if I pinged you already):

it's worth for you all to take a look what @aswinsuryan and I are doing here. He's the first one to add a real OTE test for a feature gate implementation.

There might be some overlap here with https://github.com/openshift/cluster-ingress-operator/pull/1556. I don't think it needs to be perfect - as test helpers, and structure can evolve, I think we should merge something (get @aswinsuryan's FG test started) and you all can always adjust helpers, change structure, as you see fit.

tests-extension/test/lb_security_groups.go
gcs278: @coderabbitai why is this new OTE test causing `: [sig-network] NetworkPolicy should use numeric ports in all platform NetworkPolicies [apigroup:networking.k8s.io] [Suite:openshift/conformance/parallel] ` to fail in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1500/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/2094524314232557568, when other origin CI jobs (see https://github.com/openshift/origin/tree/main/test/extended/router) create custom ICs in parallel tests already? Are we doing something differently here that is causing this test failure to trigger? be aware, we know that it's a legit bug (https://redhat.atlassian.net/browse/OCPBUGS-99494), but I'm confused why it's not permafailing today, before the introduction of this PR.

gcs278: @aswinsuryan it may be because you are creating a custom IC and holding it for the life of the test in the `beforeall` type function? I know this is inconvenient, but I'd rather not add these as a serial tests.

> The new test holds its custom NLB IngressController for longer. It waits for NLB provisioning and performs Service recreation and security-group updates. This increases the overlap window.

I don't think you can merge with a broken CI test, but you could an alternate workaround. Maybe try creating/deleting an IC individually for each test  - which would be a little less efficient, but may escape the failure by only keeping an IC for a short period of time.

I'm 95% sure other tests do these and get away without breaking this test. Otherwise, the right solution is to  work with @rfredette and fix https://redhat.atlassian.net/browse/OCPBUGS-99494. If it's an easy fix, let's get it fixed as soon as we can.

tests-extension/test/lb_security_groups.go
aswinsuryan: good catch , Removed it

tests-extension/test/util.go
aswinsuryan: Removed

tests-extension/test/util.go
aswinsuryan: Removed

tests-extension/test/lb_security_groups.go
aswinsuryan: ok I made it back to parellel ,  each IC is now scoped to a single test via BeforeEach, not shared. will observe the run, and iterate if it still flakes.

gcs278: @aswinsuryan looks like it [passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1500/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/2097694261997735936).

Even if it's flakey (fails sometimes), that's fine - this PR _shouldn't_ get reverted as we should be able to argue we aren't introducing a regression and there's an open bug.

tests-extension/test/lb_security_groups.go
```diff
@@ -0,0 +1,171 @@
...
+
+			icName = uniqueICName("sgtest-ote")
+			ic := newNLBIngressController(icName, icName+"."+domain, []operatorv1.SecurityGroupID{operatorv1.SecurityGroupID(sgID)})
+			o.Expect(cl.createWithRetryOnError(ctx, ic, updateTimeout)).To(o.Succeed(), "failed to create ingresscontroller")
+			g.DeferCleanup(func(ctx context.Context) {
+				o.Expect(cl.deleteIngressController(ctx, icName, conditionTimeout)).To(o.Succeed())
+			})
+		})
+
+		g.It("provisions, updates, and removes the security groups over the LB lifecycle", func(ctx context.Context) {
```
gcs278: we now only have 2 `g.It` test cases - I think you previously had 5, but either way, we need 5 unique tests for FG promotion.

Was this something you recently changed? Can we split back to 5 tests?

tests-extension/test/lb_security_groups.go
aswinsuryan: Agree, we can continue with parallel mode and these tests.

tests-extension/test/lb_security_groups.go
aswinsuryan: I combined it to see if that reduced flakiness. Will split back into the 5 tests, since that's a hard requirement for FG promotion.

gcs278: ah sorry I understand now.

gcs278:
Thanks for all of the updates on this one. I think we are in good shape minus the network policy issue, which is not a regression introduced by this PR.

I want to say that we can "just merged it" even if this PR causes a new CI failure, but I'm not 100% sure what will happen. If it's a <50% flake, then let's merge it. If it's a >50% flake, let's work with @rfredette on https://redhat.atlassian.net/browse/OCPBUGS-99494 - it needs to be fixed, and I think it's likely we will get into the hot seat for breaking CI. Let me know what you think. 

/approve
/lgtm

gcs278:
ci was in rough shape yesterday
/retest

mtulio:
Per 5 tests passing for FG `IngressControllerLBSecurityGroupsAWS` on [e2e-aws-ovn-techpreview](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1500/pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-ovn-techpreview/2098493471420583936):

/lgtm

aswinsuryan:
/verified by "provisions a LoadBalancer service with the specified security group", "effectuates an update to the security groups", "removes the security groups using the auto-delete-load-balancer annotation", "reflects the effective security
  groups in the IngressController status", "preserves a security groups annotation and reconciles once the spec matches"

gcs278:
i can't access any of the prow logs - just going to run again
/test e2e-aws-ovn-hypershift-conformance

gcs278:
e2e-aws-operator-techpreview looks good (minus the expected subnets failure)

`  * could not run steps: step e2e-aws-ovn-hypershift-conformance failed: failed to acquire lease for "hypershift-aws-quota-slice": resources not found `
/test e2e-aws-ovn-hypershift-conformance

gcs278:
carrying over the verified:
/verified later @mjoseph

PR #1456, opened 2026-05-29
OCPBUGS-98310: Bump sail-operator install library to OSSM 3.4.0

gcs278:
## Summary

Vendors the sail-operator install library from the OSSM 3.4 release branch and adapts CIO to the updated API.

- **Vendor sail library** from `openshift-service-mesh/sail-operator` release-3.4
- **Adapt to API changes** in the 3.4 install library ([upstream](https://github.com/istio-ecosystem/sail-operator/pull/1857), [downstream backport](https://github.com/openshift-service-mesh/sail-operator/pull/901)) — new function signatures, CRD ownership model, values defaults
- **Bump OSSM version to 3.4.0** and Istio version to `v1.28-latest` (resolves to the latest 1.28 patch available in the catalog)
- **Add unit test** (`TestSailLibraryImageDigestsAreDownstream`) to catch upstream images slipping through — the sail library silently defaults to `registry.istio.io` images, and we've been tricked by this before

### API changes

- `install.New()` now takes a third argument `chart.CRDsFS` (the CRD filesystem) and variadic `LibraryOption` args.
- `install.Start()` now returns `(<-chan struct{}, error)` instead of just `<-chan struct{}`.
- `install.Apply()` now returns `error` instead of being void.
- `install.MergeValues()` now returns `(*sailv1.Values, error)` instead of just `*sailv1.Values`.
- `install.GatewayAPIDefaults()` now takes a `namespace` string argument.
- `install.Options.ManageCRDs` and `IncludeAllCRDs` changed from `*bool` to `bool`.

### CRD ownership / labeling

- CRDs continue to be labeled with `ingress.operator.openshift.io/owned=true` via `WithCRDOwnershipLabel`. This preserves backward compatibility with the old library and the origin upgrade test.

### Status conditions (CRDsReady)

The 3.4 sail library replaced ownership-based CRD states with readiness-based states and per-CRD `Managed`/`Ready` booleans. The `CRDInfo` struct now uses `Managed bool` and `Ready bool` fields instead of a single `State` enum. CIO maps these back to preserve the original condition reasons by inspecting per-CRD ownership via `countManagedCRDs`.

| Old reason | New reason | Status | Message change |
|---|---|---|---|
| `ManagedByCIO` | `ManagedByCIO` | `True` | No change (removed from library, now set by CIO) |
| `ManagedByOLM` | `ManagedByOLM` | `True` | No change (removed from library, now set by CIO) |
[...]

PR #1511, opened 2026-07-18
DO NOT MERGE: Validate OSSM 3.4 bump without lifecycle mutex fix

gcs278:
Recreating #1456 (without the lifecycle mutex fix from #1508) so we can validate testing assumptions — specifically that the istiod cleanup race condition reproduces without the fix.


gcs278:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

gcs278:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-upgrade 10

gcs278:
Couldn't reproduce here, but we have clear signal in https://github.com/openshift/cluster-ingress-operator/pull/1508
/close

PR #1502, opened 2026-07-08
NE-2688: Bump GWAPI to v1.5.1 and Istio v1.30.1

rhamini3:
Performing a y-stream Istio bump along with GatewayAPI CRD bump to v1.5.1

Adding in Conformance tests, introducing TLSRoutes and ListenerSets

PR #1513, opened 2026-07-21
NE-2836: Set Accepted=False and Prometheus alert for ListenerSets on managed Gateways

gcs278:
## Summary

Sets `Accepted=False` on ListenerSets targeting OpenShift-managed Gateways and exposes a per-ListenerSet Prometheus metric with an alert.

Istio 1.30 has a [conflict resolution bug](https://github.com/istio/istio/pull/60775) for ListenerSets that could allow hostname hijacking. CIO disables ListenerSet reconciliation via `PILOT_IGNORE_RESOURCES`, but users can still create ListenerSets that would become active on a future upgrade when the ignore is removed.

## Changes

- **New controller** (`listenerset-status`): per-ListenerSet reconciliation — watches ListenerSets, Gateways, and GatewayClasses. For any ListenerSet targeting an OpenShift-managed Gateway, sets `Accepted=False` with reason `UnsupportedByController`. Uses a field index on ListenerSets by parent Gateway for targeted lookups (index created in `ensureDependentControllers` after CRDs are installed). Predicates filter to only relevant changes (parentRef, gatewayClassName, create/delete).
- **Prometheus metric** (`ingress_operator_listenerset_on_managed_gateway`): GaugeVec with `listenerset_namespace` and `listenerset_name` labels, set to 1 per ListenerSet targeting a managed Gateway. Cleaned up when the ListenerSet is deleted or the parent Gateway is deleted.
- **PrometheusRule alert** (`ListenerSetOnManagedGateway`): fires per-ListenerSet after 5m, severity `info`. Message identifies the specific ListenerSet and warns about potential activation on future upgrade.
- **RBAC**: adds `listenersets` and `listenersets/status` to the CIO ClusterRole.
- **E2E test**: verifies a ListenerSet targeting a managed Gateway gets `Accepted=False` with `UnsupportedByController` reason.

Jira: [NE-2836](https://redhat.atlassian.net/browse/NE-2836)

### DIFF
--- a/manifests/00-cluster-role.yaml
+++ b/manifests/00-cluster-role.yaml
@@ -180,6 +180,8 @@ rules:
   - grpcroutes
   - referencegrants
   - backendtlspolicies
+  - listenersets
+  - listenersets/status
   verbs:
   - '*'
 
--- a/pkg/operator/controller/listenerset-status/controller.go
+++ b/pkg/operator/controller/listenerset-status/controller.go
@@ -4,8 +4,8 @@
 // ListenerSets, so CIO disables ListenerSet reconciliation via
 // PILOT_IGNORE_RESOURCES. This controller ensures users are informed
 // that their ListenerSets will not be reconciled. It also exposes a
-// Prometheus gauge to enable alerting when ListenerSets exist on
-// managed Gateways.
+// per-ListenerSet Prometheus gauge to enable alerting when ListenerSets
+// exist on managed Gateways.
 package listenerset_status
 
 import (
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -40,15 +39,19 @@ const (
 	// ReasonUnsupportedByController is the reason set on the Accepted
 	// condition when a ListenerSet targets an OpenShift-managed Gateway.
 	ReasonUnsupportedByController = "UnsupportedByController"
+
+	// listenerSetParentGatewayIndex is the field index key for looking up
+	// ListenerSets by their parent Gateway name (namespace/name).
+	listenerSetParentGatewayIndex = "spec.parentRef.gateway"
 )
 
 var (
 	log = logf.Logger.WithName(controllerName)
 
-	listenerSetOnManagedGatewayMetric = prometheus.NewGauge(prometheus.GaugeOpts{
+	listenerSetOnManagedGatewayMetric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "ingress_operator_listenerset_on_managed_gateway",
 		Help: "Set to 1 when a ListenerSet targets an OpenShift-managed Gateway. ListenerSets are not yet supported and may cause unexpected traffic behavior on upgrade.",
-	})
+	}, []string{"listenerset_namespace", "listenerset_name"})
 )
 
 func RegisterMetrics() error {
@@ -69,33 +72,78 @@ func NewUnmanaged(mgr manager.Manager) (controller.Controller, error) {
 		return nil, err
 	}
 
-	// Use a single synthetic key so that any ListenerSet or GatewayClass
-	// change triggers one global reconcile. The reconciler lists all
-	// ListenerSets and managed GatewayClasses to compute the full
-	// picture, so per-resource reconciliation would be redundant.
-	toReconcile := func(_ context.Context, _ client.Object) []reconcile.Request {
-		return []reconcile.Request{{
-			NamespacedName: types.NamespacedName{Name: "listenerset-check"},
-		}}
+	// Index ListenerSets by their parent Gateway (namespace/name) so that
+	// Gateway changes can enqueue only the affected ListenerSets.
+	if err := operatorCache.IndexField(context.Background(), &gatewayapiv1.ListenerSet{}, listenerSetParentGatewayIndex, func(o client.Object) []string {
+		ls := o.(*gatewayapiv1.ListenerSet)
+		parentNS := ls.Namespace
+		if ls.Spec.ParentRef.Namespace != nil {
+			parentNS = string(*ls.Spec.ParentRef.Namespace)
+		}
+		return []string{parentNS + "/" + string(ls.Spec.ParentRef.Name)}
+	}); err != nil {
+		return nil, fmt.Errorf("failed to create index for ListenerSets: %w", err)
 	}
 
-	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.ListenerSet{}, handler.EnqueueRequestsFromMapFunc(toReconcile))); err != nil {
+	// Watch ListenerSets directly — each ListenerSet reconciles itself.
+	// Only reconcile on create, delete, or parentRef changes.
+	listenerSetPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return true },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldLS, okOld := e.ObjectOld.(*gatewayapiv1.ListenerSet)
+			newLS, okNew := e.ObjectNew.(*gatewayapiv1.ListenerSet)
+			if !okOld || !okNew {
+				return false
+			}
+			return string(oldLS.Spec.ParentRef.Name) != string(newLS.Spec.ParentRef.Name) ||
+				parentRefNamespace(oldLS) != parentRefNamespace(newLS)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool { return true },
+	}
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.ListenerSet{}, &handler.EnqueueRequestForObject{}, listenerSetPredicate)); err != nil {
 		return nil, fmt.Errorf("failed to watch ListenerSets: %w", err)
 	}
 
+	// Watch Gateways — when a Gateway's class changes or it is deleted,
+	// enqueue the ListenerSets that target it via the index.
 	gatewayHasOurController := operatorcontroller.GatewayHasOurController(log, operatorCache, false)
+	gatewayToListenerSets := func(ctx context.Context, o client.Object) []reconcile.Request {
+		key := o.GetNamespace() + "/" + o.GetName()
+		var listenerSets gatewayapiv1.ListenerSetList
+		if err := operatorCache.List(ctx, &listenerSets, client.MatchingFields{listenerSetParentGatewayIndex: key}); err != nil {
+			log.Error(err, "failed to list ListenerSets for Gateway", "gateway", key)
+			return nil
+		}
+		var requests []reconcile.Request
+		for _, ls := range listenerSets.Items {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: ls.Namespace, Name: ls.Name},
+			})
+		}
+		return requests
+	}
 	gatewayPredicate := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool { return gatewayHasOurController(e.Object) },
 		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldGW, okOld := e.ObjectOld.(*gatewayapiv1.Gateway)
+			newGW, okNew := e.ObjectNew.(*gatewayapiv1.Gateway)
+			if !okOld || !okNew {
+				return false
+			}
+			if oldGW.Spec.GatewayClassName == newGW.Spec.GatewayClassName {
+				return false
+			}
 			return gatewayHasOurController(e.ObjectOld) || gatewayHasOurController(e.ObjectNew)
 		},
 		DeleteFunc:  func(e event.DeleteEvent) bool { return gatewayHasOurController(e.Object) },
 		GenericFunc: func(e event.GenericEvent) bool { return false },
 	}
-	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.Gateway{}, handler.EnqueueRequestsFromMapFunc(toReconcile), gatewayPredicate)); err != nil {
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.Gateway{}, handler.EnqueueRequestsFromMapFunc(gatewayToListenerSets), gatewayPredicate)); err != nil {
 		return nil, fmt.Errorf("failed to watch Gateways: %w", err)
 	}
 
+	// Watch GatewayClasses — when our GatewayClass is created or deleted,
+	// enqueue all ListenerSets so they can re-evaluate.
 	isOurGatewayClass := func(o client.Object) bool {
 		gc, ok := o.(*gatewayapiv1.GatewayClass)
 		if !ok {
@@ -103,14 +151,27 @@ func NewUnmanaged(mgr manager.Manager) (controller.Controller, error) {
 		}
 		return gc.Spec.ControllerName == operatorcontroller.OpenShiftGatewayClassControllerName
 	}
-	gatewayClassPredicate := predicate.Funcs{
-		CreateFunc: func(e event.CreateEvent) bool { return isOurGatewayClass(e.Object) },
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			return isOurGatewayClass(e.ObjectOld) || isOurGatewayClass(e.ObjectNew)
-		},
-		DeleteFunc: func(e event.DeleteEvent) bool { return isOurGatewayClass(e.Object) },
+	gatewayClassToListenerSets := func(ctx context.Context, _ client.Object) []reconcile.Request {
+		var listenerSets gatewayapiv1.ListenerSetList
+		if err := operatorCache.List(ctx, &listenerSets); err != nil {
+			log.Error(err, "failed to list ListenerSets for GatewayClass change")
+			return nil
+		}
+		var requests []reconcile.Request
+		for _, ls := range listenerSets.Items {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: ls.Namespace, Name: ls.Name},
+			})
+		}
+		return requests
 	}
-	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.GatewayClass{}, handler.EnqueueRequestsFromMapFunc(toReconcile), gatewayClassPredicate)); err != nil {
+	gatewayClassPredicate := predicate.Funcs{
+		CreateFunc:  func(e event.CreateEvent) bool { return isOurGatewayClass(e.Object) },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return isOurGatewayClass(e.Object) },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
+	}
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.GatewayClass{}, handler.EnqueueRequestsFromMapFunc(gatewayClassToListenerSets), gatewayClassPredicate)); err != nil {
 		return nil, fmt.Errorf("failed to watch GatewayClasses: %w", err)
 	}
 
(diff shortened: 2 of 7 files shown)


manifests/00-cluster-role.yaml
```diff
@@ -180,6 +180,8 @@ rules:
   - grpcroutes
   - referencegrants
   - backendtlspolicies
+  - listenersets
```
rikatz, line 183: not from here, but I think Ishmam's PR adds listenersets and also tlsroute, it will cause a conflict on there as well

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,202 @@
+// The listenerset-status controller watches for ListenerSet resources
+// targeting OpenShift-managed Gateways and sets Accepted=False on them.
+// Istio does not correctly implement hostname conflict resolution for
```
rikatz: Nit:
```suggestion
// Current Istio (1.30.1) does not correctly implement hostname conflict resolution for
```

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,202 @@
...
+		return []reconcile.Request{{
+			NamespacedName: types.NamespacedName{Name: "listenerset-check"},
+		}}
+	}
+
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.ListenerSet{}, handler.EnqueueRequestsFromMapFunc(toReconcile))); err != nil {
+		return nil, err
+	}
+
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.Gateway{}, handler.EnqueueRequestsFromMapFunc(toReconcile))); err != nil {
```
rikatz: given we just want the reconciliation to happen on Gateways that are managed by us, probably you need some indexer that just adds Gateways that we care.

rikatz: well, nevermind, my comment may not make sense here and may be over optimization. Just consider that now you are kicking a reconciliation for any Gateway, not sure if we can filter that out.

eg.: https://github.com/openshift/cluster-ingress-operator/blob/d1f37085d3d98418497f82adaa0ca1c4d8486dc9/pkg/operator/controller/gateway-status/controller.go#L103-L114

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,202 @@
...
+	// Since PILOT_IGNORE_RESOURCES prevents Istio from setting any status,
+	// this is the only signal the user gets that ListenerSets are unsupported.
+	found := false
+	for i := range listenerSetList.Items {
+		ls := &listenerSetList.Items[i]
+		parentNS := ls.Namespace
+		if ls.Spec.ParentRef.Namespace != nil {
+			parentNS = string(*ls.Spec.ParentRef.Namespace)
+		}
+		if !ourGateways[gatewayKey{string(ls.Spec.ParentRef.Name), parentNS}] {
```
rikatz: in theory, if this is not our Gateway, the implementation should clean it. Like let's say Envoy Gateway is running, it sees Accepted=False, it should clean and own to Accepted=True (unless we get hit by some SSA)

pkg/operator/operator.go
```diff
@@ -361,18 +363,31 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
...
+	dependentControllers := []controller.Controller{
+		gatewayClassController,
+		gatewayServiceDNSController,
+		gatewayLabelController,
+		gatewayStatusController,
+		gatewayNetworkPolicyController,
+	}
+
+	var listenerSetCRD apiextensionsv1.CustomResourceDefinition
+	if err := mgr.GetClient().Get(context.TODO(), types.NamespacedName{Name: "listenersets.gateway.networking.k8s.io"}, &listenerSetCRD); err != nil {
```
rikatz: why not use ctx here?

test/e2e/gateway_api_test.go
```diff
@@ -1603,3 +1604,128 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
 
 	return errs
 }
+
+// testListenerSetNotAccepted verifies that a ListenerSet targeting an
+// OpenShift-managed Gateway gets Accepted=False, and that a ListenerSet
+// targeting a non-existent gateway does not get any conditions set.
+func testListenerSetNotAccepted(t *testing.T) {
+	t.Helper()
```
rikatz: IMO this should be removed

test/e2e/gateway_api_test.go
```diff
@@ -1603,3 +1604,128 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
...
+
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Name: "listenersets.gateway.networking.k8s.io"}, &crd); err != nil {
```
rikatz:
```suggestion
	if err := kclient.Get(t.Context(), types.NamespacedName{Name: "listenersets.gateway.networking.k8s.io"}, &crd); err != nil {
```

rikatz: (here and everywhere if possible)

test/e2e/gateway_api_test.go
```diff
@@ -1603,3 +1604,128 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
...
+		if err := kclient.Delete(context.TODO(), ls); err != nil {
+			if !errors.IsNotFound(err) {
+				t.Logf("failed to delete ListenerSet: %v", err)
+			}
+		}
+	})
+
+	t.Log("verifying ListenerSet gets Accepted=False")
+	nsName := types.NamespacedName{Namespace: ls.Namespace, Name: ls.Name}
+	require.Eventually(t, func() bool {
```
rikatz:
```suggestion
	assert.Eventually(t, func() bool {
```

if you put require, the whole test may fail (unless this is your real intent)

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if len(gatewayClassList.Items) == 0 {
+		listenerSetOnManagedGatewayMetric.Set(0)
+		return reconcile.Result{}, nil
+	}
+	ourClassNames := make(map[string]bool, len(gatewayClassList.Items))
+	for _, gc := range gatewayClassList.Items {
+		ourClassNames[gc.Name] = true
+	}
+
```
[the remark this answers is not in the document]
gcs278: I'm removing the clearing in next push so this is irrelevant.

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,202 @@
...
+		if err := r.setListenerSetNotAccepted(ctx, ls); err != nil {
+			log.Error(err, "failed to set ListenerSet status", "listenerset", ls.Name, "namespace", ls.Namespace)
+		}
+	}
+
+	if found {
+		listenerSetOnManagedGatewayMetric.Set(1)
+	} else {
+		listenerSetOnManagedGatewayMetric.Set(0)
+	}
```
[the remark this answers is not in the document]
gcs278: Fixing in next push.

pkg/operator/operator.go
```diff
@@ -361,18 +363,31 @@ func New(config operatorconfig.Config, kubeConfig *rest.Config) (*Operator, erro
...
+		log.Info("ListenerSet CRD not found, skipping listenerset-status controller")
+	} else {
+		listenerSetStatusController, err := listenersetstatuscontroller.NewUnmanaged(mgr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create listenerset-status controller: %w", err)
+		}
+		dependentControllers = append(dependentControllers, listenerSetStatusController)
+	}
```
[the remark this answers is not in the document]
gcs278: Fixing in next push.

test/e2e/gateway_api_test.go
```diff
@@ -1602,3 +1604,128 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
...
+// testListenerSetNotAccepted verifies that a ListenerSet targeting an
+// OpenShift-managed Gateway gets Accepted=False, and that a ListenerSet
+// targeting a non-existent gateway does not get any conditions set.
+func testListenerSetNotAccepted(t *testing.T) {
+	t.Helper()
+
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Name: "listenersets.gateway.networking.k8s.io"}, &crd); err != nil {
+		t.Skip("ListenerSet CRD not installed, skipping")
+	}
```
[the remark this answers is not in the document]
gcs278: fixed in next push.

test/e2e/gateway_api_test.go
```diff
@@ -1602,3 +1604,128 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
...
+
+	t.Log("verifying Accepted condition is cleared after re-targeting")
+	require.Eventually(t, func() bool {
+		if err := kclient.Get(context.TODO(), nsName, ls); err != nil {
+			t.Logf("failed to get ListenerSet: %v", err)
+			return false
+		}
+		cond := condutils.FindStatusCondition(ls.Status.Conditions, string(gatewayapiv1.ListenerSetConditionAccepted))
+		return cond == nil || cond.Reason != listenersetstatuscontroller.ReasonUnsupportedByController
```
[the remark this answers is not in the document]
gcs278: Fixing in next push.

manifests/00-cluster-role.yaml
gcs278: He adds ListenerSet to the aggregated admin/view cluster roles, but we only add RBAC to `manifests/00-cluster-role.yaml` when CIO actually needs to create/edit/view the CRs, like in this PR. CIO already has CRD-level permissions separately.

This has been a bit confusing as we've unnecessarily added things like BackendTLSPolicy in this RBAC, in which the CIO does nothing with the CRs and probably should be removed.

pkg/operator/controller/listenerset-status/controller.go
gcs278: Done.

pkg/operator/controller/listenerset-status/controller.go
gcs278: Sure, it's a easy addition.

pkg/operator/controller/listenerset-status/controller.go
gcs278: Yes I agree, we don't need to clean up.

pkg/operator/operator.go
gcs278: fixed.

test/e2e/gateway_api_test.go
gcs278: Removed.

test/e2e/gateway_api_test.go
gcs278: Fixed.

test/e2e/gateway_api_test.go
gcs278: Done.

gcs278:
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/cluster-ingress-operator#1502

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &gatewayapiv1.ListenerSet{}, handler.EnqueueRequestsFromMapFunc(toReconcile))); err != nil {
+		return nil, fmt.Errorf("failed to watch ListenerSets: %w", err)
+	}
+
+	gatewayHasOurController := operatorcontroller.GatewayHasOurController(log, operatorCache, false)
+	gatewayPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return gatewayHasOurController(e.Object) },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return gatewayHasOurController(e.ObjectOld) || gatewayHasOurController(e.ObjectNew)
```
rikatz, line 123: nit: you don't need the old object, just the new one in this case, right? if the old object had our class but does not have anymore it means "we don't care"

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,207 @@
...
+		},
+		DeleteFunc:  func(e event.DeleteEvent) bool { return gatewayHasOurController(e.Object) },
```
rikatz, line 125: why would we care if a Gateway that has a ListenerSet pointing to us is deleted?

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,207 @@
...
+	type gatewayKey struct{ name, namespace string }
+	ourGateways := make(map[gatewayKey]bool)
+	for _, gw := range gatewayList.Items {
+		if ourClassNames[string(gw.Spec.GatewayClassName)] {
+			ourGateways[gatewayKey{gw.Name, gw.Namespace}] = true
+		}
+	}
+
+	listenerSetList := gatewayapiv1.ListenerSetList{}
+	if err := r.cache.List(ctx, &listenerSetList); err != nil {
```
rikatz: nit: it would be good if we indexed the ListenerSet by the parent Gateway, so we could ask directly here the list of ListenerSets that match our gateways

that said, this may be some over optimization to a controller that I hope will never see production

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,207 @@
...
+	return c, nil
+}
+
+type reconciler struct {
+	client client.Client
+	cache  cache.Cache
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log.Info("reconciling", "request", request)
```
rikatz: I think what concerns me here is that any reconciliation of a Gateway or a ListenerSet will list all of the GatewayClasses, then the Gateways, then the ListenerSets and this can exponentially grow. Imagine that if someone changes a Gateway to add, let's say a different certificate, or if someone tries to add a ListenerSet, it will do:
* A list of N resources on GatewayClass
* A list of M resources on Gateway
* A list of P resources on ListenerSet


I think that given the idea is to check if a ListenerSet points to a Gateway that is managed by our GatewayClass, I would suggest it to be a bit different:
* Reconcile on ListenerSet events only. Get the ListenerSet, check its parent, then just run a gatewayHasOurController on this resource/Gateway name
* In case you want to reconcile a Gateway change, just kick a new reconciliation in case the spec.gatewayClassName or spec.allowedListeners change, and with an index that for every ListenerSet it add the Gateway name (so you list just the resources you care).

I don't think there is much of a value of reconciling with GatewayClass (I may be missing something tbh) and I think we should limit the reconciliation here to the target object, wdyt?

gcs278: > Reconcile on ListenerSet events only. Get the ListenerSet, check its parent, then just run a gatewayHasOurController on this resource/Gateway name

Yea all of this design is mostly centered around the alert: it's designed as a global/singleton `gauge` in which it is `This alert fires when *any* ListenerSet resource targets an OpenShift-managed Gateway.`.

Doing a full reconcile of the state enables us to also remove this alert when no listenersets are attached. 

> I don't think there is much of a value of reconciling with GatewayClass (I may be missing something tbh) and I think we should limit the reconciliation here to the target object, wdyt?

If you uninstall Gateway API via deleting the gatewayclass, this watch allows us to remove the alert. An edge case, but out of completeness.

I've updated to refactor to per-listenerset reconciles & per-listenerset `guagevec` alert. I was debating whether it was worth it, but refactoring is  easy with claude, and the per-listenerset design is more elegant.

pkg/operator/controller/listenerset-status/controller.go
gcs278: Done.

pkg/operator/controller/listenerset-status/controller.go
gcs278: To clear the alert if the gateway changes gatewayclass.

pkg/operator/controller/listenerset-status/controller.go
gcs278: To clear the metric/alert.

gcs278 (commented):
Thanks for the review @rikatz! Refactored to use per-listenerset reconciles and per-listenerset metric/alert. Will do some self-review soon to catch anything i missed, but feel free to have a look.

pkg/operator/controller/listenerset-status/controller.go
rikatz: ah right

pkg/operator/controller/listenerset-status/controller.go
```diff
@@ -0,0 +1,266 @@
...
+		return gc.Spec.ControllerName == operatorcontroller.OpenShiftGatewayClassControllerName
+	}
+	gatewayClassToListenerSets := func(ctx context.Context, _ client.Object) []reconcile.Request {
+		var listenerSets gatewayapiv1.ListenerSetList
+		if err := operatorCache.List(ctx, &listenerSets); err != nil {
+			log.Error(err, "failed to list ListenerSets for GatewayClass change")
+			return nil
+		}
+		var requests []reconcile.Request
+		for _, ls := range listenerSets.Items {
```
rikatz, line 148: commenting for followup story: this means "a GatewayClass kicks the reconciliation for every ListenerSet" which is not really gatewayClassToListenerSets but more like a "reconcile it all".

As a followup, let's either filter this or rename the function and make it clear that any gatewayclass change (admin action) tries to reconcile every listenerset

test/e2e/gateway_api_test.go
```diff
@@ -1604,3 +1606,54 @@ func ensureGatewayObjectSuccess(t *testing.T, ns *corev1.Namespace) []string {
 
 	return errs
 }
+
+// testListenerSetNotAccepted verifies that a ListenerSet targeting an
+// OpenShift-managed Gateway gets Accepted=False.
+func testListenerSetNotAccepted(t *testing.T) {
```
rikatz, line 1716: another followup, if possible I think this test should:
* Create a GatewayClass (managed)
* Create a Gateway (managed)
* Point a ListenerSet to this Gateway
* Check that the Accepted = False AND our condition is set (you do it already)
* Check the Prometheus metric if possible (we are not checking it, maybe it deserves at least some unit test on the Reconcile to see if the registered metric is growing, etc)
* Delete the Gateway, check if the metric is gone (we don't care abouyt cleaning accepted now)
* Same for GatewayClass

rikatz:
/lgtm
/approve
The PR looks solid. The decision of not cleaning the Accepted condition from ListenerSet was made given we expect another controller to take over it so we don't want to conflict on this.

My only missing request is that an additional PR with some unit tests for the proper metric generation and deletion based on each reconciliation loop is created (a e2e would be better, but I am happy with unit tests)

test/e2e/gateway_api_test.go
gcs278: Sounds good - that should be reasonable. I'll add this in a follow up.

pkg/operator/controller/listenerset-status/controller.go
gcs278: Yea agree, it's a "reconcile all listenersets". Filtering felt a bit like a bit overkill, as uninstalling/reinstalling gateway api via add/delete gatewayclass is should be a rare occurrence.

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1530 has merged
/retest

gcs278:
`    --- FAIL: TestCreateCluster/ValidateHostedCluster (2452.10s)` - unrelated
/test e2e-hypershift

gcs278:
metal install failures
/retest

gcs278:
```
2026-07-30 21:41:58	
2026-07-30T21:41:58.550Z	ERROR	operator.main	cobra/command.go:1019	error starting	{"error": "failed to create operator: failed to create listenerset-status controller: failed to create index for ListenerSets: failed to get restmapping: no matches for kind \"ListenerSet\" in version \"gateway.networking.k8s.io/v1\""}
```

Moved the index creation into ensureDependentControllers to avoid the error when controller is created via `NewUnmanaged`

rikatz:
/lgtm
thanks for adding the tests and fixing the bug on indexer

melvinjoseph86:
Tested using clusterbot

```
  1. Verify RBAC — Operator Has ListenerSet Permissions
➜   oc get clusterrole openshift-ingress-operator -o yaml | grep -A2 "listenersets"
  - listenersets
  - listenersets/status
  verbs:
  - '*'

  2. ListenerSet Targeting a Managed Gateway Gets Accepted=False
  ➜  oc -n openshift-ingress get gateway
NAME              CLASS               ADDRESS                                                                  PROGRAMMED   AGE
ls-test-gateway   openshift-default   aeb3080089e1648caad07c0b2f626d64-868630333.us-east-2.elb.amazonaws.com   True         14s
➜    oc wait gateway/ls-test-gateway -n openshift-ingress --for=condition=Programmed --timeout=180s
gateway.gateway.networking.k8s.io/ls-test-gateway condition met

➜  cat <<EOF | oc apply -f -
kind: ListenerSet
metadata:
  name: test-ls-rejected
  namespace: openshift-ingress
spec:
  parentRef:
    name: ls-test-gateway
  listeners:
  - name: extra-listener
    port: 8080
    protocol: HTTP
EOF

➜    oc get listenerset test-ls-rejected -n openshift-ingress \
    -o jsonpath='{.status.conditions[?(@.type=="Accepted")]}' | python3 -m json.tool
{
    "lastTransitionTime": "2026-07-31T08:15:16Z",
    "message": "ListenerSets are not yet supported by the OpenShift Gateway API implementation. This ListenerSet will not be reconciled. On a future upgrade, this ListenerSet may become active and could cause unexpected traffic routing.",
    "observedGeneration": 1,
    "reason": "UnsupportedByController",
    "status": "False",
    "type": "Accepted"
}

  3. ListenerSet Targeting a Non-Managed Gateway Is NOT Affected
  ➜  cat <<EOF | oc apply -f -
kind: ListenerSet
metadata:
  name: test-ls-unmanaged
  namespace: openshift-ingress
spec:
  parentRef:
    name: some-other-gateway
  listeners:
  - name: extra-listener
    port: 8080
    protocol: HTTP
EOF

listenerset.gateway.networking.k8s.io/test-ls-unmanaged created
➜  oc get listenerset test-ls-unmanaged -n openshift-ingress \
>     -o jsonpath='{.status.conditions}' 
[{"lastTransitionTime":"1970-01-01T00:00:00Z","message":"Waiting for controller","reason":"Pending","status":"Unknown","type":"Accepted"},{"lastTransitionTime":"1970-01-01T00:00:00Z","message":"Waiting for controller","reason":"Pending","status":"Unknown","type":"Programmed"}]%                                                                                                                                                    

  4. Prometheus Alert Rule Is Installed
  ➜ oc get prometheusrule -n openshift-ingress-operator -o yaml | \
    grep -A10 "ListenerSetOnManagedGateway"
      - alert: ListenerSetOnManagedGateway
        annotations:
          description: This alert fires when a ListenerSet resource targets an OpenShift-managed
            Gateway. ListenerSets are not supported by the OpenShift Gateway API implementation.
            Remove the ListenerSet or retarget it to a non-OpenShift-managed Gateway.
          message: ListenerSet {{ $labels.listenerset_namespace }}/{{ $labels.listenerset_name
            }} is targeting an OpenShift-managed Gateway and will not be reconciled.
            On a future upgrade, this ListenerSet may become active and could cause
            unexpected traffic routing.
          summary: ListenerSet targeting an OpenShift-managed Gateway detected
        expr: ingress_operator_listenerset_on_managed_gateway == 1

```
/verified by @mjoseph

PR #1517, opened 2026-07-22
NE-2809: Add upgradeable logic for HAProxy version

jcmoraisjr:
Add the upgradeable logic checking if the HAProxy version being used is supported in the next release. A new command-line option provides the list of deprecated versions, since it cannot be inferred from the API or the current state.

The new command-line option defaults to "2.8", which is the currently deprecated version. Having a default value makes it easier to be adopted by deployments not configuring (yet) the option, e.g. Hypershift.

https://redhat.atlassian.net/browse/NE-2809

### DIFF
--- a/cmd/ingress-operator/start.go
+++ b/cmd/ingress-operator/start.go
@@ -56,6 +56,8 @@ type StartOptions struct {
 	HAProxyImages map[string]string
 	// DefaultHAProxyVersion is the default HAProxy version.
 	DefaultHAProxyVersion string
+	// DeprecatedHAProxyVersion is the list of HAProxy versions blocking upgrade.
+	DeprecatedHAProxyVersion []string
 	// CanaryImage is the pullspec of the ingress operator image
 	CanaryImage string
 	// ReleaseVersion is the cluster version which the operator will converge to.
@@ -89,6 +91,7 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&options.IngressControllerImage, "image", "i", "", "image of the ingress controller the operator will manage (required)")
 	cmd.Flags().StringToStringVarP(&options.HAProxyImages, "haproxy-image", "", nil, "HAProxy images as version=pullspec (optional)")
 	cmd.Flags().StringVarP(&options.DefaultHAProxyVersion, "default-haproxy-version", "", "", "defines the default HAProxy version, required if --haproxy-image is also provided (optional)")
+	cmd.Flags().StringSliceVarP(&options.DeprecatedHAProxyVersion, "deprecated-haproxy-version", "", []string{string(operatorv1.HAProxyVersion28)}, "defines the HAProxy version blocking upgrade if used (optional)")
 	cmd.Flags().StringVarP(&options.CanaryImage, "canary-image", "c", "", "image of the canary container that the operator will manage (optional)")
 	cmd.Flags().StringVarP(&options.ReleaseVersion, "release-version", "", statuscontroller.UnknownVersionValue, "the release version the operator should converge to (required)")
 	cmd.Flags().StringVarP(&options.MetricsListenAddr, "metrics-listen-addr", "", "127.0.0.1:60000", "metrics endpoint listen address (required)")
@@ -162,12 +165,21 @@ func start(opts *StartOptions) error {
 		}
 	}
 
+	if len(haproxyImages) > 0 {
+		for _, version := range opts.DeprecatedHAProxyVersion {
+			if _, found := haproxyImages[operatorv1.HAProxyVersion(version)]; !found {
+				return fmt.Errorf("deprecated HAProxy version %q not found in --haproxy-image", version)
+			}
+		}
+	}
+
 	operatorConfig := operatorconfig.Config{
 		OperatorReleaseVersion:    opts.ReleaseVersion,
 		Namespace:                 opts.OperatorNamespace,
 		IngressControllerImage:    opts.IngressControllerImage,
 		HAProxyImages:             haproxyImages,
 		DefaultHAProxyVersion:     defaultHAProxyVersion,
+		DeprecatedHAProxyVersion:  opts.DeprecatedHAProxyVersion,
 		CanaryImage:               opts.CanaryImage,
 		GatewayAPIOperatorCatalog: opts.GatewayAPIOperatorCatalog,
 		GatewayAPIOperatorChannel: opts.GatewayAPIOperatorChannel,
--- a/pkg/operator/controller/ingress/status.go
+++ b/pkg/operator/controller/ingress/status.go
@@ -122,7 +122,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	errs = append(errs, err)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressProgressingCondition(updated.Status.Conditions))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, platformStatus, secret))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(updated, r.config, deploymentRef, service, platformStatus, secret))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressEvaluationConditionsDetectedCondition(ic, service))
 
 	updated.Status.Conditions = PruneConditions(updated.Status.Conditions)
@@ -719,10 +719,12 @@ func computeIngressDegradedCondition(conditions []operatorv1.OperatorCondition, 
 }
 
 // computeIngressUpgradeableCondition computes the IngressController's "Upgradeable" status condition.
-func computeIngressUpgradeableCondition(ic *operatorv1.IngressController, deploymentRef metav1.OwnerReference, service *corev1.Service, platform *configv1.PlatformStatus, secret *corev1.Secret) operatorv1.OperatorCondition {
+func computeIngressUpgradeableCondition(ic *operatorv1.IngressController, config Config, deploymentRef metav1.OwnerReference, service *corev1.Service, platform *configv1.PlatformStatus, secret *corev1.Secret) operatorv1.OperatorCondition {
 	var errs []error
 
 	errs = append(errs, checkDefaultCertificate(secret, "*."+ic.Status.Domain))
+
+	errs = append(errs, checkDeprecatedHAProxyVersion(ic, config.DeprecatedHAProxyVersion))
 
 	if service != nil {
 		errs = append(errs, loadBalancerServiceIsUpgradeable(ic, deploymentRef, service, platform))
@@ -809,6 +811,15 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
 		}
 	}
 
+	return nil
+}
+
+// checkDeprecatedHAProxyVersion returns an error if the provided IngressController
+// resource references a HAProxy version unsupported in the next release.
+func checkDeprecatedHAProxyVersion(ic *operatorv1.IngressController, deprecatedHAProxyVersion []string) error {
+	if version := string(ic.Status.EffectiveHAProxyVersion); slices.Contains(deprecatedHAProxyVersion, version) {
+		return fmt.Errorf("haproxy %q is being removed in the next release; update or unset spec.haproxyVersion before upgrading", version)
+	}
 	return nil
 }
 
--- a/test/e2e/multi_haproxy_test.go
+++ b/test/e2e/multi_haproxy_test.go
@@ -1,0 +1,97 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apiserver/pkg/storage/names"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+func TestMultiHAProxyUpgradeableCondition(t *testing.T) {
+	enabled, err := isFeatureGateEnabled(features.FeatureGateIngressControllerMultipleHAProxyVersions)
+	require.NoError(t, err)
+	if !enabled {
+		t.Skip("Skipping TestMultiHAProxyUpgradeableCondition as FeatureGateIngressControllerMultipleHAProxyVersions is disabled")
+	}
+
+	ctx := context.Background()
+
+	testCases := map[string]struct {
+		haproxyVersion operatorv1.HAProxyVersion
+		expectedStatus bool
+	}{
+		"should block upgrade on deprecated version": {
+			haproxyVersion: operatorv1.HAProxyVersion28,
+			expectedStatus: false,
+		},
+		"should allow upgrade on supported version": {
+			haproxyVersion: operatorv1.HAProxyVersion32,
+			expectedStatus: true,
+		},
+		"should allow upgrade on default version": {
+			haproxyVersion: "",
+			expectedStatus: true,
+		},
+	}
+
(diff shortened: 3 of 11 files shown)


pkg/operator/controller/ingress/status.go
```diff
@@ -122,7 +122,7 @@ func (r *reconciler) syncIngressControllerStatus(ic *operatorv1.IngressControlle
 	errs = append(errs, err)
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressProgressingCondition(updated.Status.Conditions))
 	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, degradedCondition)
-	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, deploymentRef, service, platformStatus, secret))
+	updated.Status.Conditions = MergeConditions(updated.Status.Conditions, computeIngressUpgradeableCondition(ic, r.config, deploymentRef, service, platformStatus, secret))
```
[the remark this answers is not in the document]
jcmoraisjr: Good catch, just fixed. There is already a test exploring values both in spec and status, including initial (empty) ones.

test/e2e/multi_haproxy_test.go
```diff
@@ -0,0 +1,123 @@
...
+					HAProxyVersion: test.haproxyVersion,
+				},
+			}
+			err = kclient.Create(ctx, ic)
+			require.NoError(t, err)
+
+			t.Cleanup(func() {
+				err := kclient.Delete(ctx, ic)
+				assert.NoError(t, err)
+			})
```
gcs278: Be sure to leverage existing helpers - they are not only saving code, but also protect against flakes, like "createWithRetryOnError" for example.

```suggestion
  name := types.NamespacedName{Namespace: defaultName.Namespace, Name: names.SimpleNameGenerator.GenerateName("e2e-multi-haproxy-")}
  ic := newPrivateController(name, name.Name+".router.local")
  ic.Spec.HAProxyVersion = test.haproxyVersion
  
   if err := createWithRetryOnError(t, ctx, ic, DefaultRetryTimeout); err != nil {
      t.Fatalf("failed to create ingresscontroller %s: %v", name, err)
    }
  t.Cleanup(func() { assertIngressControllerDeleted(t, kclient, ic) })
  
  if err := waitForIngressControllerCondition(t, kclient, 5*time.Minute, name, availableConditionsForPrivateIngressController...); err != nil {
    t.Fatalf("failed to observe expected conditions: %v", err)
  }
```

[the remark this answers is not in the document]
gcs278: @jcmoraisjr i think that's a good point, we should fail it if we delete and the cluster operator status doesn't get cleared.

test/e2e/multi_haproxy_test.go
```diff
@@ -0,0 +1,123 @@
...
+				if !assert.GreaterOrEqual(collect, idx, 0, "condition %s not found", operatorv1.OperatorStatusTypeUpgradeable) {
+					return
+				}
+				condition := icCopy.Status.Conditions[idx]
+				if test.expectedStatus {
+					assert.Equal(collect, operatorv1.ConditionTrue, condition.Status)
+				} else {
+					assert.Equal(collect, operatorv1.ConditionFalse, condition.Status)
+				}
+			}, 60*time.Second, 2*time.Second)
```
gcs278: We do have a helper that already does this effectively:

```suggestion
	   expectedUpgradeable := operatorv1.ConditionTrue
        if !test.expectedStatus {
            expectedUpgradeable = operatorv1.ConditionFalse
        }
        err = waitForIngressControllerCondition(t, kclient, 60*time.Second, name,
            operatorv1.OperatorCondition{
                Type:   operatorv1.OperatorStatusTypeUpgradeable,
                Status: expectedUpgradeable,
            },
        )
        require.NoError(t, err)
```

test/e2e/multi_haproxy_test.go
```diff
@@ -0,0 +1,123 @@
...
+					return
+				}
+				condition := co.Status.Conditions[idx]
+				if test.expectedStatus {
+					assert.Equal(collect, configv1.ConditionTrue, condition.Status)
+				} else {
+					assert.Equal(collect, configv1.ConditionFalse, condition.Status)
+					assert.Contains(collect, condition.Message, fmt.Sprintf(`"%s"`, name))
+				}
+			}, 60*time.Second, 2*time.Second)
```
gcs278: If you don't mind about dropping the message check - then we also have a helper that does this too. I think the message check is fine to drop - that's something better handled in unit testing (e.g. someone changes the message and has to wait 2hr+ to realize they forgot to update this test). Your call though  

```suggestion
		expectedUpgradeable := configv1.ConditionTrue
        if !test.expectedStatus {
            expectedUpgradeable = configv1.ConditionFalse
        }
        err = waitForClusterOperatorConditions(t, kclient,
            configv1.ClusterOperatorStatusCondition{
                Type:   configv1.OperatorUpgradeable,
                Status: expectedUpgradeable,
            },
        )
        require.NoError(t, err)
```

cmd/ingress-operator/start.go
```diff
@@ -89,6 +91,7 @@ func NewStartCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&options.IngressControllerImage, "image", "i", "", "image of the ingress controller the operator will manage (required)")
 	cmd.Flags().StringToStringVarP(&options.HAProxyImages, "haproxy-image", "", nil, "HAProxy images as version=pullspec (optional)")
 	cmd.Flags().StringVarP(&options.DefaultHAProxyVersion, "default-haproxy-version", "", "", "defines the default HAProxy version, required if --haproxy-image is also provided (optional)")
+	cmd.Flags().StringSliceVarP(&options.DeprecatedHAProxyVersion, "deprecated-haproxy-version", "", []string{string(operatorv1.HAProxyVersion28)}, "defines the HAProxy version blocking upgrade if used (optional)")
```
gcs278, line 94: totally up to you, but adding something like:
```
for _, v := range opts.DeprecatedHAProxyVersion {
    if _, found := haproxyImages[operatorv1.HAProxyVersion(v)]; !found && len(haproxyImages) > 0 {
        return fmt.Errorf("deprecated HAProxy version %q is not in --haproxy-image", v)
    }
}
```
would make sure we don't accidentally provide a non-existent deprecated haproxy version.

test/e2e/operator_test.go
```diff
@@ -2646,7 +2633,7 @@ func TestContainerLoggingMinLength(t *testing.T) {
 	}
 	t.Cleanup(func() { deleteWithRetryOnError(t, context.Background(), clientPod, DefaultRetryTimeout) })
 
-	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "syslog-netpol", Namespace: clientPod.Namespace})
+	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "container-min-length-netpol", Namespace: clientPod.Namespace})
```
gcs278: is this an unrelated e2e bug fix? 😄  

I'm all for fixing e2e bugs, but keep in mind the backport process - mingling these changes together in other commits and/or PRs makes backporting a more tedious manual process (we can't use our automation tools).

I'd prefer to keep PRs scoped to what they are documented against - feel free to push up a small PR to fix this and tag it against a bug so we can backport it (yea I know, it's painful for such a small fix, but that keeps fixes backport-able and clean). Chai bot can create a bug, push a PR up for you if you ask it.

Let me know if I'm misunderstanding.

gcs278 (commented):
Overall LGTM - mainly just comments about using existing E2E helper functions (which often were created to avoid flakes).

test/e2e/operator_test.go
jcmoraisjr: Ouch, we talked about this fix, this is flaking due to the same name used on another parallel test. I fixed it locally and missed to revert. Let me create another PR to fix it.

cmd/ingress-operator/start.go
jcmoraisjr: Makes sense, added a validation for the deprecated haproxy version.

test/e2e/multi_haproxy_test.go
jcmoraisjr: Cool, I missed all these helpers, just changed as suggested.

The only part I dropped was the check for the IngressController condition - do you think it's better to have it? The Upgradeable condition works despite of it and the test runs much faster.

test/e2e/multi_haproxy_test.go
jcmoraisjr: The message check is an extra validation but already covered by the status, which is the most important one. I think we're good this way.

test/e2e/multi_haproxy_test.go
gcs278: Not a requirement, but I do think it's better to have more granular error conditions. Better to stop the the test if the IC is not going `Available` then to continue on to check `Upgradeable`. Establishes a baseline of "everything is working correctly" before checking for your condition.

Regardless - the addition of the available check effectively increased the timeout. Right now, looks like 1 minute for `Upgradable=False` which also includes the IC going ready. This is pretty tight. We generally allow 5 minutes just for the private IC to go available ([example](https://github.com/openshift/cluster-ingress-operator/blob/master/test/e2e/operator_test.go#L1945)).

So, I still recommend it, but either way, increasing the timeout from 1 minute is strong recommendation from me. Flakes are expensive - clusters / CIO can get really busy in parallel tests as we add more. Hence, I always advocate for generous timeouts.

test/e2e/multi_haproxy_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func TestMultiHAProxyUpgradeableCondition(t *testing.T) {
+	enabled, err := isFeatureGateEnabled(features.FeatureGateIngressControllerMultipleHAProxyVersions)
+	require.NoError(t, err)
+	if !enabled {
+		t.Skip("Skipping TestMultiHAProxyUpgradeableCondition as FeatureGateIngressControllerMultipleHAProxyVersions is disabled")
+	}
+
+	ctx := context.Background()
```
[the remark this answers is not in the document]
gcs278: I saw coderabbit commenting this on other tests. this is probably better done as a test-wide change. It's an existing problem, so I wouldn't block the PR on it.

test/e2e/multi_haproxy_test.go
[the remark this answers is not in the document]
jcmoraisjr: Fair point. I preserved the `t.Cleanup()` code as a fallback, in case the test fails, but also deleting the ic resource and waiting for the ClusterOperator resource to report Upgradeable condition as true.

@coderabbitai can you check my changes again and see it makes sense?

jcmoraisjr: I see your point. First first though was to check the Upgradeable condition only (hence the 1m timeout), but this is not a realistic scenario. Changed now to wait the cluster to report as available, and after that check the condition.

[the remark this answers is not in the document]
jcmoraisjr: Although tests are being asserted with `assert` on those lines, checking it on both ends make more sense in case the test is changed in the future. Just updated as suggested, please do another review.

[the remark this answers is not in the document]
jcmoraisjr: @coderabbitai although tests are being asserted with `assert` on those lines, checking it on both ends make more sense in case the test is changed in the future. Just updated as suggested, please do another review.

pkg/operator/controller/ingress/status.go
```diff
@@ -812,6 +814,18 @@ func checkDefaultCertificate(secret *corev1.Secret, domain string) error {
...
 
+// checkDeprecatedHAProxyVersion returns an error if the provided IngressController
+// resource references a HAProxy version unsupported in the next release.
+func checkDeprecatedHAProxyVersion(ic *operatorv1.IngressController, deprecatedHAProxyVersion []string) error {
+	if version := string(ic.Spec.HAProxyVersion); slices.Contains(deprecatedHAProxyVersion, version) {
+		return fmt.Errorf("haproxy %q is unsupported in the next release", version)
+	}
+	if version := string(ic.Status.EffectiveHAProxyVersion); slices.Contains(deprecatedHAProxyVersion, version) {
+		return fmt.Errorf("effective haproxy %q is unsupported in the next release", version)
+	}
```
gcs278: Minor point, but we should be a bit more descriptive in this error message, giving the user an action to take. Also, "unsupported in the next release" is true, but a bit vague. 

`haproxy %q is being removed in the next release; update or unset spec.haproxyVersion before upgrading`

Also, using spec & status is a bit redundant. Status should be the only thing you need to check, but it's not a big deal if you want to do it this way, as it technically works. Your choice.

jcmoraisjr: Updated as suggested.

My first though was to use spec in case status is stale, but the way it is implemented status is derived from the spec just before calculating Upgradeable, so it should always be up to date. Changed it as suggested as well.

gcs278:
Looks good! thanks for the e2e test addition.

/approve
/lgtm

gcs278:
Need a unit test update for the new message.

gcs278:
none related:
/retest

gcs278:
https://github.com/openshift/cluster-ingress-operator/pull/1524 has merged
/retest

gcs278:
@melvinjoseph86 will need your verification when you get a chance, thanks!

gcs278:
ah this failure again:
```
NAME  TestAll/parallel/TestContainerLoggingMinLength
    operator_test.go:2651: failed to create network policy syslog-netpol: networkpolicies.networking.k8s.io "syslog-netpol" already exists
```
/test e2e-aws-operator

melvinjoseph86:
Verified using cluster bot

```
➜  oc get clusterversion
oc geNAME      VERSION                                               AVAILABLE   PROGRESSING   SINCE   STATUS
version   5.0.0-0-2026-07-28-091829-test-ci-ln-rig31tt-latest   True        False         3h15m   Cluster version is 5.0.0-0-2026-07-28-091829-test-ci-ln-rig31tt-latest

  1. Verify the operator starts with the new flag
➜   oc get deployment -n openshift-ingress-operator ingress-operator -o yaml | grep -A1 deprecated-haproxy-version
        - --deprecated-haproxy-version
        - $(DEPRECATED_HAPROXY_VERSION)
        
 2. Verify upgrade is blocked with deprecated HAProxy version (2.8)
 ➜  oc create -f - <<EOF
kind: IngressController
metadata:
  name: test-haproxy28
spec:
  haproxyVersion: "2.8"
  domain: test28.example.com
  endpointPublishingStrategy:
    type: Private
EOF
ingresscontroller.operator.openshift.io/test-haproxy28 created  
➜   oc get clusteroperator ingress -o jsonpath='{.status.conditions}' | jq '.[] | select(.type=="Upgradeable")'
{
  "lastTransitionTime": "2026-07-28T13:09:43Z",
  "message": "Some ingresscontrollers are not upgradeable: ingresscontroller \"test-haproxy28\" is not upgradeable: OperandsNotUpgradeable: One or more managed resources are not upgradeable: haproxy \"2.8\" is being removed in the next release; update or unset spec.haproxyVersion before upgrading",
  "reason": "IngressControllersNotUpgradeable",
  "status": "False",
  "type": "Upgradeable"
}


  3. Verify upgrade is allowed with supported HAProxy version (3.2)
  ➜  oc create -f - <<EOF
kind: IngressController
metadata:
  name: test-haproxy32
spec:
  haproxyVersion: "3.2"
  domain: test32.example.com
  endpointPublishingStrategy:
    type: Private
EOF
ingresscontroller.operator.openshift.io/test-haproxy32 created
➜  oc get ingresscontroller test-haproxy32 -n openshift-ingress-operator -o jsonpath='{.status.conditions}' | jq '.[] | select(.type=="Upgradeable")'
{
  "lastTransitionTime": "2026-07-28T13:15:03Z",
  "message": "IngressController is upgradeable.",
  "reason": "Upgradeable",
  "status": "True",
  "type": "Upgradeable"
}

 4. Verify upgrade is allowed with default (unset) HAProxy version
 ➜  oc create -f - <<EOF
kind: IngressController
metadata:
  name: test-haproxy-default
spec:
  domain: testdefault.example.com
  endpointPublishingStrategy:
    type: Private
EOF
ingresscontroller.operator.openshift.io/test-haproxy-default created
➜  oc get ingresscontroller test-haproxy-default -n openshift-ingress-operator -o jsonpath='{.status.conditions}' | jq '.[] | select(.type=="Upgradeable")'
{
  "lastTransitionTime": "2026-07-28T13:15:03Z",
  "message": "IngressController is upgradeable.",
  "reason": "Upgradeable",
  "status": "True",
  "type": "Upgradeable"
}
```

Different scenarios are working as expected
/verified by @mjoseph

PR #1522, opened 2026-07-23
OCPBUGS-98088: Use /healthz for router startup probe

jcmoraisjr:
The startup probe previously used /healthz/ready, which requires the router to have synced its first route within the 2-minute failure threshold. On resource-constrained clusters or during slow provisioning, the bootstrap process can exceed this window, causing the router pod to crashloop.

Switching the startup probe to /healthz checks only that the router process is alive, allowing the readiness probe to gate traffic until the initial sync completes. This lets the router remain not-ready during a slow bootstrap rather than restarting repeatedly.

https://redhat.atlassian.net/browse/OCPBUGS-98088

### DIFF
--- a/pkg/manifests/assets/router/deployment.yaml
+++ b/pkg/manifests/assets/router/deployment.yaml
@@ -113,7 +113,7 @@ spec:
           startupProbe:
             failureThreshold: 120
             httpGet:
-              path: /healthz/ready
+              path: /healthz
               port: 1936
               scheme: HTTP
             periodSeconds: 1


jcmoraisjr:
Flaking distinct tests, let's give it another run

/test hypershift-e2e-aks

jcmoraisjr:
> image "quay-proxy.ci.openshift.org/openshift/ci@sha256:ed87740afa81fb3ee548ce223df59185895d821c4125334a5021e964d6d1717d" not found: manifest unknown: manifest unknown

/test hypershift-e2e-aks

jcmoraisjr:
Failed provisioning the cluster.

/test hypershift-e2e-aks

jcmoraisjr:
Provisioning failure:

> cp: cannot stat '/tmp/kubeconfig-726602482': No such file or directory

 /test e2e-aws-ovn-upgrade

gcs278:
I did some research on this. This is the correct fix. The startup probe was checking readiness when it should have been checking liveness. Though, this is not the root cause of this bug, but it should have a very good chance at resolving the bootstrap timeout. The bootstrap timeout properly exposed this as a bug.

Looking at one of logs of the [failures](https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22:%22PCEB727DF2F34084E%22,%22queries%22:%5B%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fperiodic-ci-openshift-release-main-ci-5.0-e2e-azure-ovn-techpreview%2F2074406747040124928%5C%22%7D%20%7C%20unpack%20%7C%20namespace%3D%5C%22openshift-ingress%5C%22%22,%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221783405091372%22,%22to%22:%221783441091374%22%7D%7D), it's pretty clear the router is spinning on the API server waiting for `"Failed to watch" err="failed to list *v1.Route: the server could not find the requested resource (get routes.route.openshift.io)`.

The startup probe's only job is to delay liveness. So checking readiness there means the container gets killed for not being ready, which is exactly what the readiness probe exists to handle without killing anything. We can be sure that when the startup probe passes, the liveness probe is also going to pass (both are now `/healthz`).

This lets the router stay alive during startup while it waits for the Route API to become available. The readiness probe handles gating traffic. The startup probe no longer kills the container for a dependency it can't control.

Now, the finger still points at the resource requests being downgraded as the root cause, which we should also revisit with some priority. But, this is still a correct fix regardless.

@coderabbit, can you confirm my research and if there are any risks or unintended consequences with a startup probe  change like this?

gcs278:
>The main intentional trade-off is that a router can now remain running but unready for longer when a dependency is unavailable or synchronization is otherwise stuck. That is preferable to a crash loop for this dependency-driven condition, and it does not send traffic to an unready router. 

Nicely put.

> One concrete follow-up: pkg/operator/controller/ingress/deployment_test.go still expects the startup-probe path to be /healthz/ready (around Line 2664). Since the controller directly carries this asset into the desired Deployment, that expected fixture needs to change to /healthz along with the manifest; otherwise the deployment unit test will fail.

Yea technically the unit test exists with:
```
  StartupProbe: &corev1.Probe{
  ProbeHandler: corev1.ProbeHandler{
	  HTTPGet: &corev1.HTTPGetAction{
		  Path: "/healthz/ready",
		  Port: intstr.IntOrString{
			  IntVal: int32(1936),
		  },
	  },
  },
  },
```

But it doesn't fail because it's just testing for changes - not a requirement to change this.

It's a release blocker, and @Miciah has informally ack'ed the approach in slack - I think we have enough consensus to move forward with this.

/approve
/lgtm

gcs278:
Actually, I need to correct myself after some more research. While I still agree this is a correctness fix, I don't see strong evidence that this will fix the regression.

The bootstrap failure is caused by the Route API (openshift-apiserver) not coming up. Whether the router restarts or retries internally with exponential backoff, neither fixes the fact that the Route API was never available. The resource reduction from the sidecar PR also doesn't explain this. The ingress operator logs confirm the same root cause: it was also failing on the Route API not being present.

So, unless someone points something out that I'm missing, I don't feel confident this will fix the problem. However, I still think it's a correctness fix that I agree with, that potentially could have some obscure improvement that I'm missing. I'll leave the LGTM for now, but keep this in mind.

gcs278:
Putting a preemptive hold, just because there is some uncertainty. If we find that the router isn't the root cause of these bootstrap failures, then we may want to 1. hold off on this change (despite it's correctness) and 2. file it under a different bug with lower priority (if it's likely not the fix)

/hold

jcmoraisjr:
Closing on behalf of https://github.com/openshift/cluster-ingress-operator/pull/1528

/close

PR #1532, opened 2026-07-29
OCPBUGS-100186: Rename network policy in TestContainerLoggingMinLength

jcmoraisjr:
The tests TestContainerLoggingMinLength and TestSyslogLogging are creating network policy resource with the same name. Whenever these tests run at the same time, one of them fails with "NetworkPolicy already exists" error.

Changing the name of the resource in TestContainerLoggingMinLength allows them to run at the same time without issues.

https://redhat.atlassian.net/browse/OCPBUGS-100186

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -2646,7 +2646,7 @@ func TestContainerLoggingMinLength(t *testing.T) {
 	}
 	t.Cleanup(func() { deleteWithRetryOnError(t, context.Background(), clientPod, DefaultRetryTimeout) })
 
-	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "syslog-netpol", Namespace: clientPod.Namespace})
+	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "minlength-netpol", Namespace: clientPod.Namespace})
 	if err := kclient.Create(context.TODO(), networkPolicy); err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/operator_test.go
```diff
@@ -2646,7 +2646,7 @@ func TestContainerLoggingMinLength(t *testing.T) {
 	}
 	t.Cleanup(func() { deleteWithRetryOnError(t, context.Background(), clientPod, DefaultRetryTimeout) })
 
-	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "syslog-netpol", Namespace: clientPod.Namespace})
+	networkPolicy := buildTestPodNetworkPolicy(types.NamespacedName{Name: "minlength-netpol", Namespace: clientPod.Namespace})
```
Miciah, line 2649: This change should be sufficient to fix the test failures.  However, as a follow-up, we should systematically replace hard-coded names like this through use of `GenerateName`.  I have created [NE-2835](https://redhat.atlassian.net/browse/NE-2835) to track this follow-up.  

For these NetworkPolicy resources specifically, it might make sense for `TestMain` to create a single NetworkPolicy resource for all tests, as they all use the same `type=test-pod` label.

Miciah (commented):
Thanks!  

/approve
/lgtm

jcmoraisjr:
unrelated failures

/retest-required

rikatz:
/verified by e2e
the test is running properly, and it signals that the right network policy is being created and removed

PR #1533, opened 2026-07-29
NE-2834: Replace deprecated math/rand.Read in e2e test

Miciah:
Replace the `math/rand` import with `crypto/rand` in the `randomString` helper.  `math/rand.Read` was deprecated in Go 1.20 in favor of `crypto/rand.Read`, which is functionally equivalent here.

### DIFF
--- a/test/e2e/http_header_buffer_test.go
+++ b/test/e2e/http_header_buffer_test.go
@@ -6,8 +6,8 @@ package e2e
 import (
 	"bufio"
 	"context"
+	"crypto/rand"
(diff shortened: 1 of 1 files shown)


pedjak (commented):
/lgtm

Miciah:
@pedjak, would you mind adding an `/approve` as well?

Miciah:
/approve
per https://github.com/openshift/cluster-ingress-operator/pull/1533#issuecomment-5620190485.

PR #1534, opened 2026-07-31
OCPBUGS-100435: Fix ListenerSet metric registry and add metric e2e verification

gcs278:
## Summary

Fixes the `ingress_operator_listenerset_on_managed_gateway` Prometheus metric to use the controller-runtime metrics registry instead of the default global Prometheus registry. After the metrics server migration ([NE-2742](https://redhat.atlassian.net/browse/NE-2742)), only metrics on `ctrlruntimemetrics.Registry` are served — the listenerset metric was being set correctly but never appeared in Prometheus.

Also adds e2e test coverage verifying the metric is set and cleaned up properly.

## Changes

- **Registry fix**: Replace `prometheus.Register()` with `ctrlruntimemetrics.Registry.Register()` in `pkg/operator/controller/listenerset-status/controller.go`
- **E2e verification**: Add `assertMetricValue` and `assertMetricGone` helpers to verify the Prometheus metric lifecycle in `testListenerSetNotAccepted`

## Test plan

- [ ] `testListenerSetNotAccepted` e2e passes — metric value is 1 when ListenerSet targets a managed Gateway
- [ ] Metric is cleaned up after ListenerSet deletion
- [ ] `ListenerSetOnManagedGateway` alert fires in the console when a ListenerSet targets a managed Gateway

Bug: https://issues.redhat.com/browse/OCPBUGS-100435

### DIFF
--- a/pkg/operator/controller/listenerset-status/controller.go
+++ b/pkg/operator/controller/listenerset-status/controller.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	ctrlruntimemetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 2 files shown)


gcs278:
ah darn missing a `}`.
/test e2e-aws-operator
/test e2e-gcp-operator

gcs278:
This is being taken over by https://github.com/openshift/cluster-ingress-operator/pull/1539
/close

PR #1538, opened 2026-08-02
OCPBUGS-101783: reconcile canary certificate on dependency creation

not-stbenjam:
The canary-certificate controller can permanently miss reconciliation when its dependencies are created in an unlucky order. It watches only Secrets, and its default-certificate predicate rejects the Secret event when the default IngressController does not exist yet. If the IngressController appears afterward, no watched event retries the controller.

A failed 5.0 HyperShift AKS job captured the exact race: the default certificate existed, its initial event was rejected at 02:25:32.434Z, and the default IngressController appeared about two seconds later. `canary-serving-cert` was never created; every canary pod remained Pending on the missing Secret, the Service had no endpoints, the admitted route returned EOF, and ingress degraded. A same-payload control with safe object ordering created the certificate successfully. This was the job's sole terminal cause, for a conservative historical impact of one green Prow job.

This change:
- watches the default IngressController and the canary DaemonSet, the two non-Secret reconciliation dependencies;
- maps their semantic create/update/delete events to the single canary-certificate request;
- filters status-only updates with the generation predicate;
- treats an expected not-yet-created IngressController in the Secret predicate as dependency-not-ready rather than a noisy error;
- adds focused predicate and event-mapping coverage, including unrelated-resource rejection.

Validation:
- `GOTOOLCHAIN=auto GOMAXPROCS=2 go test -mod=vendor -p=2 ./pkg/operator/controller/canary-certificate`
- `git diff --check`

Representative job: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aks/2082648859074367488

### DIFF
--- a/pkg/operator/controller/canary-certificate/controller.go
+++ b/pkg/operator/controller/canary-certificate/controller.go
@@ -85,7 +85,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		defaultIC := &operatorv1.IngressController{}
 		if err := reconciler.client.Get(context.Background(), defaultICName, defaultIC); err != nil {
 			if !errors.IsNotFound(err) {
-				log.Error(err, "Failed to get default IngressController")
+				log.Error(err, "Failed to get default IngressController", "namespace", defaultICName.Namespace, "name", defaultICName.Name)
 			}
 			return false
 		}
@@ -98,7 +98,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	// reconcile target: the canary certificate's secret.
 	enqueueRequestForCanaryCertificate := handler.EnqueueRequestsFromMapFunc(toCanaryCertificate)
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &corev1.Secret{}, enqueueRequestForCanaryCertificate, predicate.Or(isCanaryCert, isDefaultIngressCert))); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to watch Secrets: %w", err)
 	}
 
 	// The certificate Secret can be observed before the default IngressController
@@ -109,7 +109,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		return isDefaultIngressControllerDependency(o, config.OperatorNamespace)
 	})
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &operatorv1.IngressController{}, enqueueRequestForCanaryCertificate, predicate.And(isDefaultIngressController, predicate.GenerationChangedPredicate{}))); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to watch IngressControllers: %w", err)
 	}
 
 	// The canary DaemonSet supplies the certificate's controller owner reference.
@@ -117,7 +117,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	// controller-runtime's rate-limited error retry.
 	isCanaryDaemonSet := predicate.NewPredicateFuncs(isCanaryDaemonSetDependency)
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &appsv1.DaemonSet{}, enqueueRequestForCanaryCertificate, predicate.And(isCanaryDaemonSet, predicate.GenerationChangedPredicate{}))); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to watch DaemonSets: %w", err)
 	}
 	return c, nil
 }
--- a/pkg/operator/controller/canary-certificate/controller.go
+++ b/pkg/operator/controller/canary-certificate/controller.go
@@ -54,8 +54,9 @@ type reconciler struct {
 // New creates the canary certificate controller
 //
 // The canary certificate controller mirrors the default ingress controller's
-// certificate into the canary's namespace. It watches the canary's certificate
-// and the default ingress controller's certificate.
+// certificate into the canary's namespace. It watches the canary's certificate,
+// the default ingress controller's certificate, and the other resources that
+// reconciliation depends on.
 func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	operatorCache := mgr.GetCache()
 	reconciler := &reconciler{
@@ -68,9 +69,7 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		return nil, err
 	}
 	// Watch for updates on the canary's certificate.
-	isCanaryCert := predicate.NewPredicateFuncs(func(o client.Object) bool {
-		return o.GetName() == operatorcontroller.CanaryCertificateName().Name && o.GetNamespace() == operatorcontroller.CanaryCertificateName().Namespace
-	})
+	isCanaryCert := predicate.NewPredicateFuncs(isCanaryCertificate)
 	// Also watch for updates on the default ingress controller's certificate.
 	// Because the default ingress controller's certificate name can be set at
 	// runtime, a Get() needs to be done to determine the correct certificate.
@@ -85,7 +84,9 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		}
 		defaultIC := &operatorv1.IngressController{}
 		if err := reconciler.client.Get(context.Background(), defaultICName, defaultIC); err != nil {
-			log.Error(err, "Failed to get default IngressController")
+			if !errors.IsNotFound(err) {
+				log.Error(err, "Failed to get default IngressController")
+			}
 			return false
 		}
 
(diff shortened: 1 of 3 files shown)


bentito:
/lgtm
/assign @rfredette 
@rfredette I think our https://github.com/openshift/cluster-ingress-operator/pull/1334 introduced a kind of subtle race. I think these changes seem like a reasonable remediation. If you agree /approve ?

bentito:
@coderabbitai are the failing checks due to code or infra?

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -95,15 +96,57 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
+		Namespace: operatorNamespace,
+		Name:      manifests.DefaultIngressControllerName,
+	})
+}
+
+func isCanaryDaemonSetDependency(o client.Object) bool {
+	return hasNamespacedName(o, operatorcontroller.CanaryDaemonSetName())
+}
+
+func hasNamespacedName(o client.Object, name types.NamespacedName) bool {
```
davidesalerno: This helper is clean and reusable and so it could be a general utility. 
We can consider to move it into `pkg/operator/controller/names.go` alongside other name helpers.

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -95,15 +96,57 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 	})
 	// Regardless of which certificate changed, this controller only has one
 	// reconcile target: the canary certificate's secret.
-	enqueueRequestForCanaryCertificate := handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, a client.Object) []reconcile.Request {
-		return []reconcile.Request{{NamespacedName: operatorcontroller.CanaryCertificateName()}}
-	})
+	enqueueRequestForCanaryCertificate := handler.EnqueueRequestsFromMapFunc(toCanaryCertificate)
 	if err := c.Watch(source.Kind[client.Object](operatorCache, &corev1.Secret{}, enqueueRequestForCanaryCertificate, predicate.Or(isCanaryCert, isDefaultIngressCert))); err != nil {
```
davidesalerno, line 100: Could we wrap the `c.Watch(...)` errors with a bit more context here (for example, which watch registration failed)? Right now these return the raw error, which may make controller startup failures harder to diagnose if one of these watch setups breaks in the future.
We could setup failures with contextual `fmt.Errorf(... %w) ` messages so startup failures identify which watch registration failed.

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -85,7 +84,9 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		}
 		defaultIC := &operatorv1.IngressController{}
 		if err := reconciler.client.Get(context.Background(), defaultICName, defaultIC); err != nil {
-			log.Error(err, "Failed to get default IngressController")
+			if !errors.IsNotFound(err) {
```
davidesalerno, line 87: Nit: for the non-`NotFound` case, it may be worth including `defaultICName.Namespace` and `defaultICName.Name` in the log fields. The current message is fine, but having the resource identity in the log would make lookup failures easier to debug in-cluster.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: Good suggestion — if other controllers end up needing this helper, moving it to `names.go` would make sense. For now it's only used here, so keeping it scoped to the canary-certificate controller and we can promote it in a follow-up if reuse materializes.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: Good catch — wrapped all three `Watch()` calls with `fmt.Errorf` identifying the resource type (Secrets, IngressControllers, DaemonSets). Done in 48df4a6.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: Added `namespace` and `name` fields to the error log. Done in 48df4a6.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: Moved it to `pkg/operator/controller/names.go` as `HasNamespacedName` in 6364c97.

devguyio:
## Independent RCA: HyperShift AKS 4.22 e2e failures on PR openshift/hypershift#9211

We independently hit this exact bug while investigating CI failures on [openshift/hypershift#9211](https://github.com/openshift/hypershift/pull/9211) (an Azure CPO image override update — no code changes). Posting our full root cause analysis here as additional evidence supporting this fix.

### Affected Job

- **Job**: `pull-ci-openshift-hypershift-main-e2e-aks-4-22`
- **Build**: [2084667334202494976](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_hypershift/9211/pull-ci-openshift-hypershift-main-e2e-aks-4-22/2084667334202494976)
- **Payload**: `4.22.0-0.ci-2026-08-03-224338` (and `4.22.0-0.ci-2026-08-02-005640` for upgrade)
- **Failed 3/3 runs** on this PR, also hit on unrelated PR #9206 with the same payload

### Failed Tests

| Test | Cluster | Duration |
|------|---------|----------|
| `TestNodePool/HostedCluster0/EnsureHostedCluster` | `node-pool-qrhbc` (15 nodes) | 604s |
| `TestUpgradeControlPlane/ValidateHostedCluster` | `control-plane-upgrade-d6kk2` (2 nodes) | 2316s |

(+ 3 parent cascade failures. 355/360 tests passed.)

### Root Cause Chain

```
canary_certificate_controller predicate drops initial Secret event
  → canary-serving-cert secret never created
    → all ingress-canary DaemonSet pods stuck ContainerCreating (FailedMount)
      → canary route checks return EOF (no endpoints)
        → CanaryChecksSucceeding=False → IngressController Degraded=True
          → ingress ClusterOperator Degraded=True
            → ClusterVersionAvailable=False, ClusterVersionSucceeding=False
              → EnsureHostedCluster / ValidateHostedCluster timeout
```

### Evidence from Ingress Operator Logs

From the HCP namespace `e2e-clusters-sftsh-node-pool-qrhbc` on the management cluster:

```
17:14:48.889Z ERROR canary_certificate_controller predicate/predicate.go:116
    Failed to get default IngressController
    {"error": "ingresscontrollers.operator.openshift.io \"default\" not found"}

17:14:50.175Z INFO  Starting Controller {"controller": "canary_certificate_controller"}
17:14:50.175Z INFO  Starting workers    {"controller": "canary_certificate_controller", "worker count": 1}

# ↑ No reconciliation log from canary_certificate_controller ever again.
# Meanwhile, canary_controller logs this every ~60s for 40+ minutes:

17:14:50.269Z INFO  canary_controller canary/controller.go:258
    canary serving cert secret not found; skipping canary-serving-cert-hash annotation
[... repeated 90+ times until dump at 17:55:02Z ...]
```

### Comparative Evidence (Same Job Run)

| | Passing (`TestCreateCluster`) | Failing (`TestNodePool`) | Failing (`TestUpgradeControlPlane`) |
|---|---|---|---|
| Payload | `4.22.0-0.ci-2026-08-03-224338` | same | `4.22.0-0.ci-2026-08-02-005640` |
| Ingress operator image | identical | identical | identical |
| DaemonSet created | `17:15:26Z` | `17:14:50Z` | — |
| `canary-serving-cert` | Created `17:15:36Z` | **Never created** | **Never created** |
| Canary pods | 2/2 Running | 0/15 ContainerCreating | 0/2 ContainerCreating |
| Ingress CO | Degraded=False | **Degraded=True** | **Degraded=True** |

### Bug Mechanism Confirmed

Verified against the source at `pkg/operator/controller/canary-certificate/controller.go`:

1. **Line 101**: Controller watches only `Secret` objects, filtered by `predicate.Or(isCanaryCert, isDefaultIngressCert)`
2. **Lines 77-95** (`isDefaultIngressCert`): Predicate does a live `client.Get()` for the "default" IngressController on every Secret event. Returns `false` when the IngressController doesn't exist yet.
3. **Lines 71-73** (`isCanaryCert`): Only matches the `canary-serving-cert` secret — which doesn't exist because this controller is supposed to create it.
4. **No watches** on IngressController or DaemonSet objects, no periodic resync within a reasonable window (default is 10 hours).
5. **Result**: Once the initial Secret events during informer sync are filtered out, the controller goes permanently dormant.

The passing cluster won the timing race (IngressController existed before the cert secret event was processed). The failing clusters lost it.

### Impact

This bug blocks all `e2e-aks-4-22` runs non-deterministically. It affects any HyperShift-hosted cluster where the canary_certificate_controller starts before the default IngressController is created on the API server.

---
*Generated by [Claude Code](https://claude.ai/code), reviewed by @devguyio*

devguyio:
/pipeline required

unblocking 2nd stage, this bug is critical and causing multiple flakes in hypershift e2es

rfredette:
This looks good to me. Thanks!

/approve

candita:
`TestIngressControllerCustomEndpoints` seems to be permafailing in e2e-aws-operator with recent changes.

bentito:
@candita you asked me, out of band, why this bug stayed hidden for so long, so I did an RCA from that angle:

  1. Why it stayed hidden for so long
   * Silent Predicate Drops: The original predicate (isDefaultIngressCert) was doing a live API Get for the default IngressController. If the IngressController did not exist yet, the predicate logged an error and returned false. Because this happened at the event predicate filtering stage rather than during Reconcile(), controller-runtime dropped the event completely without placing it in the work queue. As a result, it never triggered the rate-limited retry/backoff loop that normally heals temporary errors.
   * Standard Cluster Timings: In standard OpenShift installations, the deployment orchestration order is highly predictable, and the default IngressController is almost always present by the time the secret events fire.

  2. Why it is surfacing in CI now
   * HyperShift/Hosted Control Plane Latency: As noted in the description, this was caught in a HyperShift AKS job. In hosted control plane environments, resource provisioning is highly parallelized, and API server availability/network latency can vary significantly. This dramatically widened the "unlucky" timing window where the Secret is created/updated before the default IngressController object is registered in the API server.
   * Once that initial event was silently discarded, the lack of a watch on IngressController and DaemonSet meant the controller had no way of knowing a dependency was now ready, resulting in a permanent sync failure.

Adding the explicit watches on the IngressController and DaemonSet dependencies is the correct fix here to guarantee eventual consistency regardless of timing

bentito:
@not-stbenjam isn't the only remaining failed e2e test the one this PR is trying to fix? if so, I can re-lgtm and override and we can get this merged.

stbenjam:
@redhat-chai-bot Can you look at *all* the failed jobs on this SHA and see if it has an occurence of the problem this PR is trying to fix? See https://prow.ci.openshift.org/pr-history/?org=openshift&repo=cluster-ingress-operator&pr=1538

pkg/operator/controller/canary-certificate/controller.go
```diff
@@ -95,15 +96,57 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
...
 	}
+
+	// The certificate Secret can be observed before the default IngressController
+	// exists. Watch the IngressController so that dependency creation order cannot
+	// permanently discard that Secret event. Generation filtering avoids
+	// reconciling on status-only updates.
+	isDefaultIngressController := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return isDefaultIngressControllerDependency(o, config.OperatorNamespace)
+	})
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &operatorv1.IngressController{}, enqueueRequestForCanaryCertificate, predicate.And(isDefaultIngressController, predicate.GenerationChangedPredicate{}))); err != nil {
```
Miciah, line 111: I thought we only cared about an out-of-order create event.  Why do we care about update events?

Miciah:
You need a rebase to eliminate the merge commit in https://github.com/openshift/cluster-ingress-operator/pull/1538/commits.  

Less importantly, it would be nice to put the refactoring (introducing the `isCanaryCertificate` and `toCanaryCertificate` helpers) and logging changes into separate commits from the actual fix (the added watches) and tests.

pkg/operator/controller/canary-certificate/controller.go
Miciah: `names.go` was originally for names of things, and we have `pkg/utils/` this sort of helper, but I suppose it isn't a big deal if you prefer to put `HasNamespacedName` in `names.go`.

pkg/operator/controller/canary-certificate/controller.go
Miciah: We don't wrap error values from `c.Watch` in other controllers.  Is the unwrapped error message really insufficient?  Should we update other controllers to wrap errors in a follow-up?

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: The create event is what fixes this bug, but update events matter too: the effective default-certificate Secret name is derived from the IngressController's `spec.defaultCertificate`. If that field is updated to point at a Secret that already exists, no Secret event fires (the `isDefaultIngressCert` predicate computes the effective name at event time, so it would only match a *future* event on the new Secret), and the canary certificate would keep mirroring the old certificate until something else nudged the controller. Watching spec updates closes that gap, and `GenerationChangedPredicate` keeps status-only updates from triggering reconciles.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: Fair point — since it's only used by this controller, I moved it back to an unexported `hasNamespacedName` helper in the canary-certificate package and reverted the `names.go` change. We can promote it to `pkg/util` if another controller needs it.

pkg/operator/controller/canary-certificate/controller.go
not-stbenjam: You're right that it's inconsistent with the rest of the repo — the controller name in the error context already identifies where the failure came from, so the wrapping added little. I reverted all three `c.Watch` calls to return the error unwrapped, matching the other controllers.

not-stbenjam:
@Miciah Rebased onto master to drop the merge commit and restructured the branch into three commits, with the fix first so it can be cherry-picked to release branches on its own (verified it applies cleanly to release-4.20 and passes unit tests there):

1. `canary-certificate: watch reconciliation dependencies` — the actual fix (IngressController + DaemonSet watches) and tests, self-contained against the pre-existing code
2. `canary-certificate: extract isCanaryCertificate predicate helper` — pure refactor (no behavior change)
3. `canary-certificate: improve default IngressController lookup logging` — stops logging expected NotFound as an error and adds namespace/name fields

While restructuring, I also addressed the inline comments: reverted the `c.Watch` error wrapping to match the rest of the repo, and moved `HasNamespacedName` out of `names.go` back to an unexported helper in the canary-certificate package. Replied on the update-events question in-thread.

PR #1542, opened 2026-08-04
OCPBUGS-86050: Update ROUTER_CURVES environment variable for go 1.26

davidesalerno:
Add support for NIST P-curve ML-KEM post-quantum hybrid groups (SecP256r1MLKEM768, SecP384r1MLKEM1024) that became available in Go 1.26 via crypto/tls.

Key changes:

-  Register SecP256r1MLKEM768 and SecP384r1MLKEM1024 in the tlsGroupToCurveID map so TLSGroupToCurveID resolves them to the corresponding crypto/tls CurveID constants.

-  Add fipsApprovedTLSGroups and nonFIPSGroups package-level variables to centralize the FIPS filtering logic. On FIPS-enabled clusters, X25519-based groups (X25519, X25519MLKEM768) are stripped because Curve25519 is excluded from the FIPS 140-3 approved list (NIST SP 800-56Ar3, Appendix D). The P-curve ML-KEM hybrids are FIPS-approved and remain in the list.
 
-  Use an explicit []string with deliberate ordering for the FIPS fallback group list (classical ECDH first, then PQ hybrids) instead of sets.NewString().List(), which produced an accidental ordering based on ASCII sort of the group names.

-  Update the FIPS fallback to include the PQ hybrid groups so that FIPS clusters using default settings still get post-quantum key exchange protection.

### DIFF
--- a/pkg/operator/controller/ingress/controller.go
+++ b/pkg/operator/controller/ingress/controller.go
@@ -93,6 +93,22 @@ var (
 		"TLS_AES_256_GCM_SHA384",
 	)
 
+	// fipsApprovedTLSGroups is the ordered list of TLS key exchange groups
+	// approved for use under FIPS 140-3 / NIST SP 800-56Ar3. It includes
+	// NIST P-curves (secp256r1, secp384r1, secp521r1) and their corresponding
+	// ML-KEM post-quantum hybrids (SecP256r1MLKEM768, SecP384r1MLKEM1024,
+	// approved as of Go 1.26). The order is deliberate: classical ECDH groups
+	// first for broad compatibility, then PQ hybrids.
+	// X25519-based groups (X25519, X25519MLKEM768) are excluded — see
+	// nonFIPSGroups.
+	fipsApprovedTLSGroups = []string{
+		string(configv1.TLSGroupSecP256r1),
+		string(configv1.TLSGroupSecP384r1),
+		string(configv1.TLSGroupSecP521r1),
+		string(configv1.TLSGroupSecP256r1MLKEM768),
+		string(configv1.TLSGroupSecP384r1MLKEM1024),
+	}
+
 	// nonFIPSGroups is the set of TLS groups that are NOT approved for use
 	// under FIPS 140-3 / NIST SP 800-56Ar3. X25519 and X25519MLKEM768 rely
 	// on Curve25519 arithmetic which is excluded from the FIPS-approved list
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1111,22 +1111,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
 		}
 	}
 
-	// If FIPS is enabled on this cluster, we cannot use ML-KEM or X25519 in
-	// the TLS supportedGroups (aka curves). ML-KEM and X25519 are not
-	// supported by OpenSSL FIPS 140-3.
+	// If FIPS is enabled on this cluster, we cannot use X25519-based groups
+	// (X25519, X25519MLKEM768) in the TLS supportedGroups (aka curves).
+	// These groups rely on Curve25519 arithmetic which is excluded from the
+	// FIPS-approved list. Note: ML-KEM groups based on NIST P-curves
+	// (SecP256r1MLKEM768, SecP384r1MLKEM1024) are FIPS-approved as of Go 1.26.
+	// X25519-based groups are not supported by OpenSSL FIPS 140-3.
 	// See https://redhat.atlassian.net/browse/TRT-2597 and Appendix D of
 	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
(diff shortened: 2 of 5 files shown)


pkg/operator/controller/tls.go
```diff
@@ -24,11 +24,13 @@ func copyTLSSpec(in *configv1.TLSProfileSpec) *configv1.TLSProfileSpec {
...
-	configv1.TLSGroupSecP384r1:      tls.CurveP384,
-	configv1.TLSGroupSecP521r1:      tls.CurveP521,
-	configv1.TLSGroupX25519MLKEM768: tls.X25519MLKEM768,
+	configv1.TLSGroupX25519:             tls.X25519,
+	configv1.TLSGroupSecP256r1:          tls.CurveP256,
+	configv1.TLSGroupSecP384r1:          tls.CurveP384,
+	configv1.TLSGroupSecP521r1:          tls.CurveP521,
+	configv1.TLSGroupX25519MLKEM768:     tls.X25519MLKEM768,
+	configv1.TLSGroupSecP256r1MLKEM768:  tls.SecP256r1MLKEM768,
+	configv1.TLSGroupSecP384r1MLKEM1024: tls.SecP384r1MLKEM1024,
```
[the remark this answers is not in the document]
davidesalerno: @coderabbitai This project is built using Dockerfile and the scope of Jira bug OCPBUGS-86050 it is to update ROUTER_CURVES not the golang version used that it is already at version 1.26 into the go.mod too

melvinjoseph86:
cc @rhamini3

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1111,22 +1111,21 @@ func desiredRouterDeployment(ci *operatorv1.IngressController, config *Config, i
...
 	// https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
-	// If filtering empties the list, fall back to the FIPS-compliant defaults
-	// (secp256r1:secp384r1:secp521r1) so the router always has at least one
-	// usable group.
+	// If filtering empties the list, fall back to the FIPS-approved groups
+	// (secp256r1:secp384r1:secp521r1:SecP256r1MLKEM768:SecP384r1MLKEM1024)
+	// so the router always has at least one usable group.
 	if isFIPSEnabled {
 		tlsGroupList = slices.DeleteFunc(tlsGroupList, nonFIPSGroups.Has)
 		if len(tlsGroupList) == 0 {
```
bentito, line 1127: Runs only when filtering empties the list. Predefined profiles retain `secp256r1:secp384r1`, while gate-disabled defaults retain three classical groups, so neither uses `fipsApprovedTLSGroups`. Only an `all-X25519` custom profile triggers it. This contradicts the PR’s stated goal of giving FIPS clusters using defaults post-quantum protection.

Is it just the PR description needs to change?

pkg/operator/controller/ingress/controller.go
```diff
@@ -93,6 +93,22 @@ var (
 		"TLS_AES_256_GCM_SHA384",
 	)
 
+	// fipsApprovedTLSGroups is the ordered list of TLS key exchange groups
+	// approved for use under FIPS 140-3 / NIST SP 800-56Ar3. It includes
+	// NIST P-curves (secp256r1, secp384r1, secp521r1) and their corresponding
+	// ML-KEM post-quantum hybrids (SecP256r1MLKEM768, SecP384r1MLKEM1024,
+	// approved as of Go 1.26). The order is deliberate: classical ECDH groups
```
bentito, line 100: The fallback deliberately puts classical groups before hybrids. OpenSSL treats this list as preference-ordered, so hybrid-capable peers may still negotiate classical ECDH. Go 1.26’s own preference order places hybrids first.

pkg/operator/controller/ingress/deployment.go
davidesalerno: Fixed — the FIPS block now unconditionally prepends any missing PQ hybrid groups (`SecP256r1MLKEM768`, `SecP384r1MLKEM1024`) after filtering, not just as a last-resort fallback when the list empties. Every FIPS cluster gets post-quantum protection regardless of  which profile is configure

pkg/operator/controller/ingress/controller.go
davidesalerno: Fixed — fipsApprovedTLSGroups now places the ML-KEM hybrids first (`SecP256r1MLKEM768`, `SecP384r1MLKEM1024`) to match Go 1.26's preference order, so hybrid-capable peers will negotiate PQ. The unconditional injection path also prepends them ahead of any surviving classical groups.

melvinjoseph86:
@rhamini3 Please test this PR, may be we need to change our automation accordingly after this PR merges

rhamini3:
/verified by "TestTLSGroupToCurveID|TestTLSConfigFromProfile|TestDesiredRouterDeploymentTLSGroups"

checked on clusterbot, router deploy has correct env var:
```
          - name: ROUTER_CURVES
            value: X25519MLKEM768:X25519:secp256r1:secp384r1
```

PR #1544, opened 2026-08-06
OCPBUGS-92835: add grace period to Available condition for deployment…

davidesalerno:
… blips

The IngressController Available condition immediately transitioned to False when the deployment reported DeploymentAvailable=False, even for transient blips lasting only ~30 seconds during pod rescheduling or node disruption. This caused the ClusterOperator Available condition to blip False, failing the CI invariant test
"clusteroperator/ingress should not change condition/Available".

The Degraded condition already had a 30-second grace period for DeploymentAvailable, but the Available condition had none. Add a 60-second grace period to computeIngressAvailableCondition for the DeploymentAvailable check, mirroring the existing grace period pattern. During the grace period, the IngressController remains Available=True and a requeue is scheduled to re-evaluate after the period expires.

Applying a blanket grace period to Available introduces two risks, which this change also addresses:

- A genuine full outage (zero available replicas) would otherwise be masked for up to 60 seconds instead of being reported immediately. The grace period is now only granted when at least one replica is still available; if availableReplicas is 0, DeploymentAvailable is evaluated with no grace period.

- A deployment that flaps quickly between available and unavailable (each unavailable period under 60s) can indefinitely avoid tripping Available=False, since the grace window resets on every transition back to True. This flapping is now independently observable via a new ingress_controller_deployment_available_transitions_total counter metric, incremented whenever DeploymentAvailable changes status, regardless of whether the grace period masks it in Available/Degraded.

### DIFF
--- a/pkg/operator/controller/ingress/status_test.go
+++ b/pkg/operator/controller/ingress/status_test.go
@@ -1957,76 +1957,143 @@ func Test_computeIngressProgressingCondition(t *testing.T) {
 }
 
 func Test_computeIngressAvailableCondition(t *testing.T) {
+	fakeClock := utilclocktesting.NewFakeClock(time.Time{})
+	clock = fakeClock
+	defer func() {
+		clock = utilclock.RealClock{}
+	}()
+
 	testCases := []struct {
-		description string
-		conditions  []operatorv1.OperatorCondition
-		expect      operatorv1.OperatorCondition
+		description       string
+		conditions        []operatorv1.OperatorCondition
+		availableReplicas int32
+		expect            operatorv1.OperatorCondition
+		expectRequeue     bool
+		expectAfter       time.Duration
 	}{
 		{
 			description: "deployment, dns, and lb available",
 			conditions: []operatorv1.OperatorCondition{
-				{Type: IngressControllerDeploymentAvailableConditionType, Status: operatorv1.ConditionTrue},
-				{Type: operatorv1.DNSManagedIngressConditionType, Status: operatorv1.ConditionTrue},
-				{Type: operatorv1.DNSReadyIngressConditionType, Status: operatorv1.ConditionTrue},
-				{Type: operatorv1.LoadBalancerManagedIngressConditionType, Status: operatorv1.ConditionTrue},
-				{Type: operatorv1.LoadBalancerReadyIngressConditionType, Status: operatorv1.ConditionTrue},
+				cond(IngressControllerDeploymentAvailableConditionType, operatorv1.ConditionTrue, "", clock.Now()),
+				cond(operatorv1.DNSManagedIngressConditionType, operatorv1.ConditionTrue, "", clock.Now()),
+				cond(operatorv1.DNSReadyIngressConditionType, operatorv1.ConditionTrue, "", clock.Now()),
+				cond(operatorv1.LoadBalancerManagedIngressConditionType, operatorv1.ConditionTrue, "", clock.Now()),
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/ingress/status_test.go
```diff
@@ -1957,76 +1957,143 @@ func Test_computeIngressProgressingCondition(t *testing.T) {
...
-			actual := computeIngressAvailableCondition(tc.conditions)
+			actual, err := computeIngressAvailableCondition(tc.conditions, tc.availableReplicas)
+			switch e := err.(type) {
+			case retryable.Error:
+				if !tc.expectRequeue {
+					t.Error("expected not to be told to requeue")
+				}
+				if tc.expectAfter.Seconds() != e.After().Seconds() {
+					t.Errorf("expected requeue after %s, got %s", tc.expectAfter.String(), e.After().String())
+				}
```
jcmoraisjr: nit: you can probably omit `.Seconds()` on both sides of the comparison, since `time.Duration` is a comparable type.

jcmoraisjr (commented):
This update sounds good: adding grace period by recalculating the ingress availability, implement retry reusing code, give visibility on short unavailability using a new metric, flag unavailable if missing all replicas, add new test scenarios and enlightening godocs.

My only finding is the one coderabbit already reported: the metric should be less accurate in case a failure happens in the status update, so my suggestion is to bump the metric only in case the update succeed.

There is also a really minor nit, only in the test and does not change its result, but worth to mention.

/approve

melvinjoseph86:
Tested with cluster bot
Transient pod deletion — Available should stay True (grace period)

```
➜  oc get deployment router-default -n openshift-ingress
NAME             READY   UP-TO-DATE   AVAILABLE   AGE
router-default   2/2     2            2           7h42m
➜  ROUTER_POD=$(oc get pods -n openshift-ingress \
  -l ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default \
  -o jsonpath='{.items[0].metadata.name}')
echo "Deleting $ROUTER_POD"
Deleting router-default-86d444c545-h2qsd
➜  oc delete pod $ROUTER_POD -n openshift-ingress
pod "router-default-86d444c545-h2qsd" deleted

➜  oc get ingresscontroller default -n openshift-ingress-operator -o json \
>   | jq '.status.conditions[] | select(.type=="Available" or .type=="DeploymentAvailable")'
{
  "lastTransitionTime": "2026-09-03T17:53:47Z",
  "message": "The deployment has Available status condition set to True",
  "reason": "DeploymentAvailable",
  "status": "True",
  "type": "DeploymentAvailable"
}
{
  "lastTransitionTime": "2026-09-03T17:53:47Z",
  "status": "True",
  "type": "Available"
}
➜  oc get co ingress -o json \
>   | jq '.status.conditions[] | select(.type=="Available")'
{
  "lastTransitionTime": "2026-09-03T17:53:47Z",
  "message": "The \"default\" ingress controller reports Available=True.",
  "reason": "IngressAvailable",
  "status": "True",
  "type": "Available"
}

➜  oc get deployment router-default -n openshift-ingress
NAME             READY   UP-TO-DATE   AVAILABLE   AGE
router-default   2/2     2            2           7h43m

Expected: co/ingress stays Available=True throughout the ~30s blip because the grace period (60s) absorbs the transient DeploymentAvailable=False.

➜  Downloads oc get co/ingress -w              
NAME      VERSION                                               AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   5.1.0-0-2026-09-03-075300-test-ci-ln-dsjjyn2-latest   True        False         False      86s   
```

It was not in progressing state. Hence marking as verified
/verified by @mjoseph

PR #1548, opened 2026-08-13
OCPBUGS-105442: Remove escalate/bind from Sail Library ClusterRole

Thealisyed:
## Summary

- Removes `\escalate\` and `\bind\` verbs from the Sail Library ClusterRole.
- Replaces them with an explicit least-privilege superset of all permissions istiod requiring **generated automatically** from the vendored istiod Helm chart templates
- `hack/update-sail-rbac.sh` reads the istiod chart templates across all vendored Istio version bands, skips conditionals not active in CIO's install path, takes the union and writes the rules into the manifest
- `hack/verify-sail-rbac.sh` re-reads the vendored charts and fails CI if the checked-in manifest is stale so that future vendor bumps are caught automatically
- Also covers OCPBUGS-87975: `coordination.k8s.io/leases`, `configmaps delete` and `secrets CRUD` were previously granted implicitly via escalate and are now explicit


Thealisyed:
/test e2e-aws-operator
/test e2e-aws-ovn

We want to look at the loki logs for Istiod. Look for any RBAC errors.

manifests/00-cluster-role.yaml
```diff
@@ -182,6 +182,7 @@ rules:
   - backendtlspolicies
   - listenersets
   - listenersets/status
+  - tlsroutes
```
gcs278: we spoke about this - this should be removed. This cluster role is representing CIO, and CIO doesn't reconcile, update, read, etc... TLSRoutes.

Only Istio needs tlsroute/status write permissions. That's already covered in the  `manifests/00-cluster-role-sail-library.yaml` role.

candita:
@Thealisyed don't forget to test the gateway api conformance too.

pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
Miciah, line 1: Why are you modifying the aggregated cluster-roles in `pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml` and `pkg/manifests/assets/gateway-api/aggregated-cluster-roles/view-cluster-role.yaml`?

The aggregated cluster-roles are aggregated into the admin and view cluster-roles in order to grant these permissions to end users who have these roles: namespace admins, namespace editors, namespace viewers, cluster admins, cluster editors, and cluster viewers.  Generally speaking, these are human users, not controllers.  

It makes sense to me that these cluster roles should have access to read (but not update) status and to create, read, update, and delete backendtlspolicies, routes resources, and typically listenersets as well (though a cluster-admin might want to restrict access to listenersets).

manifests/00-cluster-role-sail-library.yaml
Miciah, line 1: `01-cluster-role-binding-sail-library.yaml` defines a cluster role binding that binds this cluster role to the ingress-operator serviceaccount.  That means the ingress operator gets **cluster-wide** access for every permission enumerated in this cluster role.  Does istiod require all of these permissions in all namespaces because gateways can be created in arbitrary namespaces, or can this cluster role be split into two (or more): (1) a cluster role for which istiod truly requires a cluster role binding and (2) a cluster role for which istiod requires only a role binding for the "openshift-ingress" namespace (and possibly some constrained set of additional namespaces)?

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+  - networking.istio.io
+  - rbac.istio.io
+  - security.istio.io
+  - telemetry.istio.io
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
```
Miciah, line 150: Are permissions for istio.io required until [NE-2599](https://redhat.atlassian.net/browse/NE-2599) ("GatewayAPI only CRD mode" using `PILOT_IGNORE_RESOURCES`) is implemented?

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - gatewayclasses
+  verbs:
+  - create
+  - delete
+  - patch
+  - update
```
Miciah, line 222: Are these permissions necessary given that CIO sets `PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER` to "false"?

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+
+- apiGroups:
+  - gateway.networking.k8s.io
+  - gateway.networking.x-k8s.io
```
Miciah, line 226: Are permissions for `gateway.networking.x-k8s.io` necessary given that CIO sets `PILOT_ENABLE_ALPHA_GATEWAY_API` to "false"?  

Anyway, if we need it now, maybe we can remove it once [NE-2599](https://redhat.atlassian.net/browse/NE-2599) is implemented.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
+
+- apiGroups:
+  - gateway.networking.k8s.io
+  - networking.x-k8s.io
```
Miciah, line 236: Is permission for `gateways` in the `networking.x-k8s.io` API group (note the "x-") really needed?

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+  - watch
+
+- apiGroups:
+  - gateway.networking.x-k8s.io
+  resources:
+  - xbackendtrafficpolicies/status
+  - xlistenersets/status
+  verbs:
+  - patch
+  - update
```
Miciah, line 251: Is this needed?

manifests/00-cluster-role.yaml
```diff
@@ -179,9 +179,6 @@ rules:
   - httproutes
   - grpcroutes
   - referencegrants
-  - backendtlspolicies
-  - listenersets
-  - listenersets/status
```
gcs278, line 184: so, we can likely remove backendtlspolicy, but we can't remove listenerset/status.

Let's save this for a follow up PR that cleans up cluster role with unnecessary RBAC like backendtlspolicy.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,299 @@ rules:
...
+
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingressclasses
+  - ingresses
+  verbs:
+  - get
+  - list
+  - watch
```
Miciah, line 332: Are these permissions necessary given that CIO sets `ingressControllerMode: OFF` in the mesh config?

hack/verify-sail-rbac.sh
```diff
@@ -0,0 +1,175 @@
...
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+MANIFEST="${REPO_ROOT}/manifests/00-cluster-role-sail-library.yaml"
+RESOURCES_DIR="${REPO_ROOT}/vendor/github.com/istio-ecosystem/sail-operator/resources"
+
+LATEST_VERSION=$(ls "${RESOURCES_DIR}" | grep -v '\.go$' | sort -V | tail -1)
+CLUSTERROLE_TEMPLATE="${RESOURCES_DIR}/${LATEST_VERSION}/charts/istiod/templates/clusterrole.yaml"
+ROLE_TEMPLATE="${RESOURCES_DIR}/${LATEST_VERSION}/charts/istiod/templates/role.yaml"
+
+echo "Verifying Sail RBAC against Istio ${LATEST_VERSION}..."
```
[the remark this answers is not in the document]
Miciah: We need the union of the required permissions for any previous versions of Istio that CIO could be upgrading _from_  and the required permissions for whatever version of Istio that CIO is installing _now_.  

Consider that CIO could theoretically upgrade Istio from 1._n_ to 1.(_n_+1) where Istio 1._n_ requires access to _foos_ and Istio 1.(_n_+1) does not — or Istio 1.(_n_+1) requires access to _bars_ and Istio 1._n_ does not.  We don't want CIO to remove permission for _foos_ or _bars_ in the middle of an upgrade and thereby potentially break the old control-plane before it gets updated to the new Istio version or the new control-plane as soon as it gets updated.

hack/sail-rbac.py
```diff
@@ -0,0 +1,284 @@
...
+    if not s.startswith('['):
+        val = s.strip('"\'').strip()
+        return [val] if val else []
+    content = s[1:s.rfind(']')]
+    return [p.strip().strip('"\'') for p in content.split(',') if p.strip().strip('"\'')]
+
+
+def parse_rules(text):
+    """
+    Parse RBAC rules from YAML text without external dependencies.
```
Miciah, line 57: Would it be more robust and ergonomic to implement this in Go and re-use existing libraries for working with Helm charts and yaml syntax?  Instead of a Python script, it could be a sub-command of the ingress-operator executable.

hack/verify-sail-rbac.sh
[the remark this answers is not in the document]
Thealisyed: Done.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: I agree that is a legitimate security concern. Some permissions (leases, configmaps) are plausibly namespace-scoped to `openshift-ingress`, while things like watching Gateways across arbitrary namespaces genuinely need cluster scope. I can open a Jira ticket to track it properly. Getting the split wrong would silently break the Istio control plane mid-upgrade, so it warrants dedicated analysis rather than squeezing into a CVE fix.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: Yes, these are required regardless of NE-2599. Lines 29–31 of `clusterrole.yaml` grant `get/watch/list` on all `*.istio.io` groups unconditionally as there is no `{{- if }}` gate. 

Istiod reads these to reconcile its CRDs. NE-2599's `PILOT_IGNORE_RESOURCES` is a runtime flag that stops istiod from actioning on those resources but istiod's informer setup still needs watch permissions at startup. 
If NE-2599 ever changes the chart itself, the generate script will pick that up automatically on the next vendor bump.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: These come from the vendored chart unconditionally; there is no `{{- if PILOT_ENABLE_GATEWAY_API_GATEWAYCLASS_CONTROLLER }}` gate in the Helm template. 
The flag disables the controller at runtime but does not remove the RBAC rule. Since we generate directly from the chart, we include what the chart declares. 
Removing it would require a manual exclusion list that will be on every vendor bump. I can track that cleanup under NE-2599 once we can confirm it's dead code for our install path.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: Same reasoning as line 222 -  unconditional in the chart, no  `{{- if PILOT_ENABLE_ALPHA_GATEWAY_API }}` gate.  Agreed that NE-2599 is the right place to investigate removal.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: Same as line 222 — comes from `reader-clusterrole.yaml` unconditionally. `networking.x-k8s.io` is the pre-standard name for the Gateway API group; istiod watches it for backwards compatibility. 
NE-2599 is the right scope for investigating removal.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: These are the experimental Gateway API status subresources — istiod writes status back to them when reconciling extended Gateway API types. Comes from the chart unconditionally. 
NE-2599 is the right scope for investigating whether they can be dropped for our install path.

manifests/00-cluster-role-sail-library.yaml
Thealisyed: Unconditional in the chart — `ingressControllerMode: OFF` is a mesh config setting that disables istiod's ingress controller behaviour at runtime but the RBAC in the Helm template isn't gated on it. 

IIUC Istiod likely retains the permission for compatibility across install modes. NE-2599 is the right scope to confirm whether it can be dropped for our path.

hack/sail-rbac.py
Thealisyed: Agreed it would be more robust long-term. For this CVE fix I would say the rewrite is out of scope but I'll open a follow-up ticket.

manifests/00-cluster-role.yaml
Thealisyed: Noted — I'll track the `backendtlspolicy` removal in a follow-up PR so it doesn't block this. Agreed `listenerset/status` stays.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -36,7 +30,9 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
 rules:
-# Gateway autoscaling - Sail Library can potentially create HPAs
+# ── Sail Library: Helm-managed resources ─────────────────────────────────────
```
gcs278: try to avoid using non-ascii em dash - it might cause issues later.

manifests/00-cluster-role.yaml
gcs278: sorry - we spoke again - since escalate is being remove, we need to now include the superset of permissions for the admin and view aggregate cluster roles.

so disregard. keep it here.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,25 +63,379 @@ rules:
...
+  - create
+  - delete
+  - patch
+  - update
+
+- apiGroups:
+  - gateway.networking.k8s.io
+  - gateway.networking.x-k8s.io
+  resources:
+  - '*'
```
gcs278: this is like wack a mole 😓 we are fixing the escalate, but then we reintroduce https://redhat.atlassian.net/browse/OCPBUGS-6718.

I realize this is because we are generating from the Istio Helm chart - we are copying it's rbac, and it uses wildcards. 

I don't think we have a better option here. You cannot enumerate these resources because k8s MUST have the service account have the same exact cluster role wildcard, or you will get:
```
Error from server (Forbidden): clusterroles.rbac.authorization.k8s.io "wildcard-verbs-role" is forbidden:
user "system:serviceaccount:rbac-test:explicit-sa" (groups=["system:serviceaccounts" "system:serviceaccounts:rbac-test" "system:authenticated"])
is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["configmaps"], Verbs:["*"]}
```

So this is just an FYI for @bentito. Unless we can figure out a better way, you cannot enumerate these wildcards, or GWAPI/Istio will break (feel free to double check my work).

hack/sail-rbac.py
```diff
@@ -0,0 +1,319 @@
...
+    """
+    for cond in SKIP_CONDITIONS:
+        if cond not in s:
+            continue
+        if cond == 'istiodRemote.enabled':
+            # Skip only when the condition is not preceded by `not` in the same expression.
+            # `not .Values.istiodRemote.enabled` → this block is TRUE for CIO; don't skip.
+            if re.search(r'\bnot\b[^}]*' + re.escape(cond), s):
+                continue
+        return True
```
gcs278: this is fragile - it looks like we are hardcoding assumptions about certain conditionals here.

But as long as you commit to follow up with removing the python script in the near future, I won't block the PR on it.

manifests/00-cluster-role-sail-library.yaml
gcs278: yea, the istio cluster role always creates these [rules](https://github.com/openshift-service-mesh/sail-operator/blob/main/resources/v1.30.3/charts/istiod/templates/clusterrole.yaml#L29) regardless of any configuration.

CIO's cluster role MUST be a super set of istio's - or you will get a "403 forbidden" API-type error from k8s server on creation (and everything breaks).

manifests/00-cluster-role-sail-library.yaml
gcs278: i think this comes back to the fact that the generator scripts are using *all* of the istio versions charts available right?

xlistenerset is from 1.28, however, we don't necessarily know what Istio version we are using in the generator, so I think it's fair to union the rules from all istio chart versions.

manifests/00-cluster-role-sail-library.yaml
gcs278: agreed - if it's unconditional in the chart, it's required that CIO cluster role has it.

But I don't think NE-2599 has anything to do with ingress and ingressclasses.

gcs278 (commented):
Nice job @Thealisyed!

So a couple of things:
- I mostly agree with @Miciah - Python was probably not the best direction for generation here, and my apologies, I think I might have been the one to recommend it.
  - A Go generator as a CIO subcommand is likely the best path forward
- Someone might look at all this new RBAC and say "why are you adding unnecessary permissions?" but CIO's ClusterRole here MUST be a superset of whatever its service account creates (the istiod RBAC), otherwise you get a 403 Forbidden from the API server - that's important to note
  - aka it looks excessive, but it's a minimal set

Please do the following to verify:
1. do some "backport sanity checks". Does it work fine in 4.20? 4.21? 4.22? 5.0? Is it the same RBAC or different than 5.1?
3. Ensure we do upgrade testing via pre-submits and a full upgrade conformance (`/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade`)

I will approve & lgtm this PR because it production-wise (the RBAC changes) looks functional. However, I'd still much prefer to get a GO generator in. If you want to give it a shot - I'm willing to review any new attempts, but I can also understand we want to merge this quickly.

/approve
/lgtm

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade

gcs278:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

gcs278:
^ @Thealisyed please review the e2e jobs I started to make sure they pass (full upgrade conformance + hypershift).

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade


/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

cmd/ingress-operator/sail_rbac.go
```diff
@@ -0,0 +1,221 @@
...
+		if entry.IsDir() && strings.HasPrefix(entry.Name(), "v") {
+			versions = append(versions, entry.Name())
+		}
+	}
+	if len(versions) == 0 {
+		return "", fmt.Errorf("no versioned chart directories found")
+	}
+	sort.Strings(versions)
+
+	// Render every version and collect RBAC objects. Use the latest
```
gcs278, line 89: btw I had an update to my fork about this - https://github.com/openshift/cluster-ingress-operator/compare/master...gcs278:cluster-ingress-operator:sail-rbac-go-subcommand#diff-685083f34d37250cf3988f185403f53bb56373a187991f2dfe711c56b12a1716R90-R92

We can't just "use the latest". In some backports, we actually use non-latest istio versions. I think we need to use all Istio versions for now (to be on the safe side), inspired by your python script. There's no guarantee we will be running the latest version of Istio.

my fork current does a simple level of deduplication between the versions - but generally they are 95% the same. We can have a follow up that possibly detects the current istio version and only generates RBAC for that, but that might even be too restricted (technically istio version can be changed via CIO args...so we might want to stick with rendering all versions anyways).

I would just avoid the effort of parsing RBAC & trying to dedup perfectly (unless you can find a library to do it). I want to keep this as simple as possible.

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade

/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5
/test e2e-aws-operator
/test e2e-aws-ovn
/test e2e-aws-gatewayapi-conformance

cmd/ingress-operator/sail_rbac.go
```diff
@@ -0,0 +1,229 @@
...
+		parsed, err := extractRBACObjects(content)
+		if err != nil {
+			return nil, fmt.Errorf("parsing %s: %w", name, err)
+		}
+		objects = append(objects, parsed...)
+	}
+	return objects, nil
+}
+
+func isRBACTemplate(name string) bool {
```
gcs278, line 151: future follow up - this is a little fragile, we probably don't really even need to pre-filter files.

`extractRBACObjects` already skips non clusterrole and roles:
```
if cr.Kind != "ClusterRole" && cr.Kind != "Role" {
```

Not blocking, but I think we can just remove this.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -36,7 +30,9 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
 rules:
-# Gateway autoscaling - Sail Library can potentially create HPAs
+# --- Sail Library: Helm-managed resources ---
```
gcs278, line 33: follow up - I think in the future we should clean this up. This still makes sense to be separated from the generated Istio RBAC, but we should probably use different YAML files for clarity.

The difference between:
- Sail Library operational RBAC (current static section) : ermissions the Sail Library itself needs to install and manage Istio resources (e.g. creating/updating RBAC objects, managing webhooks, scaling HPAs). These are relatively stable and maintained by hand.
- Istiod RBAC superset (current generated section): the union of all permissions that istiod's own ClusterRoles request across all vendored versions.

I think the "sail library rbac" could probably just be pulled into `00-cluster-role.yaml` and we make this file 100% generated, and renamed to `00-cluster-role-istio.yaml`. You could even split generated to be 1:1 with istio rbac (`00-cluster-role-istio-reader.yaml` + `00-role-istio.yaml`) - don't quote me on the details, just throwing out some ideas for later. 😄

gcs278:
very nice update - I much prefer the minimalist GO library thanks @Thealisyed! Let's get this merged.
/lgtm

Please do a Loki log verification for any new errors / rbac warnings for the Istiod logs (`openshift-ingress` namespace) + CIO logs (`openshift-ingress-operator` namespace). Look if you happen to solve https://redhat.atlassian.net/browse/OCPBUGS-87975 by searching for the error message - then you can take credit for both (I'm not sure if this solves it or not).

gcs278:
all "could not watch pod"
/retest

gcs278:
could not watch pods
/retest

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5
/test e2e-aws-operator
/test e2e-aws-ovn
/test e2e-aws-gatewayapi-conformance

pkg/manifests/assets/gateway-api/aggregated-cluster-roles/admin-cluster-role.yaml
```diff
@@ -31,4 +31,3 @@ rules:
       - update
       - patch
       - delete
-      - deletecollection
```
gcs278, line 34: is there a particular reason we removed it from the RBAC? This could potentially break admin users (say if they are relying on the RBAC for scripts, etc...).

Any reason to not add it to the `00-cluster-role.yaml` instead?

Thealisyed: You’re right, there is no intended reason to remove those verbs from the aggregate-to-admin role.
I was narrowing the `CIO/Sail` operator permissions and over scoped the change into user-facing Gateway API admin RBAC. I’ll restore the aggregate role verbs and keep the Sail RBAC reduction scoped to the operator role.

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5
/test e2e-aws-operator
/test e2e-aws-ovn
/test e2e-aws-gatewayapi-conformance

pkg/operator/controller/status/controller.go
```diff
@@ -657,19 +597,17 @@ func (r *reconciler) getOperatorState(ctx context.Context, ingressNamespace, can
...
-		return state, fmt.Errorf("failed to list gateway classes: %w", err)
-	}
-	// If one or more gateway classes have ControllerName=operatorcontroller.OpenShiftGatewayClassControllerName,
-	// the ingress operator should try to install OSSM.
-	state.shouldInstallOSSM = (len(gatewayClassList.Items) > 0)
-
-	if r.config.ModeAccessor != nil {
-		state.modeTransition = r.config.ModeAccessor.GetTransitionState()
-		state.gatewayAPIManagementModeEnabled = r.config.ModeAccessor.GateEnabled()
+	if state.haveGatewayclassesResource {
```
gcs278: is this related to this rbac PR at all?

I try to be generous with unrelated changes (I dislike adding more jira overhead to fixes), but I'd much prefer to keep PRs targeted - especially since this involves backporting.

But since it's very easy to create bugs and get AI to push up changes, if this isn't related, can you remove and put into a separate PR?

manifests/00-cluster-role.yaml
```diff
@@ -248,14 +182,9 @@ rules:
...
+  - tlsroutes
   verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
+  - '*'
```
gcs278: ah this is very confusing - i was wondering why you had the `deleteCollection` RBAC failure when I saw this wildcard (it should work!).

This is a quirk of github. You aren't seeing the right "diff" with new context from https://github.com/openshift/cluster-ingress-operator/pull/1579 merging. It removed wildcards, but didn't add `deletecollection`.

so, rebase, and add `deletecollection` here.

gcs278:
changes look good!
/lgtm

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

gcs278:
Some TP features merged - we should with them:
/test e2e-aws-ovn-techpreview

gcs278:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

Thealisyed:
could not create pod

rhamini3:
confirmed the clusterrrole no longer contains escalate and bind, gateway resources are successfully installed and all related gatewayAPI tests pass

```
oc get clusterroles openshift-ingress-operator-sail-library -o yaml | grep escalate
iamin@iamin-mac cluster-ingress-operator % oc get clusterroles openshift-ingress-operator-sail-library -o yaml | grep bind    
  - clusterrolebindings
  - rolebindings
  - validatingadmissionpolicybindings

% oc get gc
openshift-default   openshift.io/gateway-controller/v1   True       13s

NAME                                       READY   STATUS    RESTARTS   AGE
istiod-openshift-gateway-58979b958-7bw4r   1/1     Running   0          29s

% oc -n openshift-ingress get gateway
NAME      CLASS               ADDRESS                                                                  PROGRAMMED   AGE
gateway   openshift-default   a81b5c3467de54044a26149b79875330-594748149.us-east-2.elb.amazonaws.com   True         11s
```
/verified by @rhamini3

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5
/test e2e-aws-operator
/test e2e-aws-ovn
/test e2e-aws-gatewayapi-conformance

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,29 +63,453 @@ rules:
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
 
-# Admission webhooks - Sail Library creates webhook configurations from istiod Helm chart
+# --- BEGIN GENERATED: do not edit below; run hack/update-sail-rbac.sh to regenerate ---
+
+# Union of istiod Helm chart RBAC across versions: v1.27.3, v1.27.5, v1.27.8, v1.27.9, v1.28.10, v1.28.4, v1.28.5, v1.28.6, v1.28.8, v1.30.1, v1.30.3, v1.30.4
```
gcs278: ah fyi this is the reason why the 3.4.2 bump changed the YAML (it was just this comment because `1.30.4` was added).

I still like this - it's pretty insightful to see the versions we are pulling (even if it requires regeneration every Istio bump). 👍

Thealisyed:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5
/test e2e-aws-operator
/test e2e-aws-ovn
/test e2e-aws-gatewayapi-conformance

cmd/ingress-operator/sail_rbac.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		for _, obj := range objects {
+			allRules, err = appendUniqueRules(allRules, obj.Rules, seen)
+			if err != nil {
+				return "", err
+			}
+		}
+	}
+
+	aggregationRules, err := aggregationAdminRules()
```
gcs278, line 112: IMO this should be a separate section in the cluster role - divided by a comment explaining to reviewers why this exists - not appended.

cmd/ingress-operator/sail_rbac.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+		if _, exists := seen[string(key)]; !exists {
+			seen[string(key)] = struct{}{}
+			allRules = append(allRules, rule)
+		}
+	}
+	return allRules, nil
+}
+
+func aggregationAdminRules() ([]rbacv1.PolicyRule, error) {
```
gcs278, line 138: we should have more documentation in the code - function comments ("why" is always better than "what").

It's not clear what and "admin rule" is here.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,29 +63,498 @@ rules:
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
 
-# Admission webhooks - Sail Library creates webhook configurations from istiod Helm chart
+# --- BEGIN GENERATED: do not edit below; run hack/update-sail-rbac.sh to regenerate ---
+
+# Union of Sail Library RBAC across versions: v1.27.3, v1.27.5, v1.27.8, v1.27.9, v1.28.10, v1.28.4, v1.28.5, v1.28.6, v1.28.8, v1.30.1, v1.30.3, v1.30.4
```
gcs278, line 85: this change from "istiod" to "sail library" is starting to mix a lot of thing together. The static rbac above is also "sail library".

Like I mentioned above, we need better traceability. Separate sections at a minimum with comment in-between explaining the source of the RBAC.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,29 +63,498 @@ rules:
...
+  - proxyconfigs
+  - serviceentries
+  - sidecars
+  - virtualservices
+  - workloadentries
+  - workloadgroups
+  verbs:
+  - '*'
+- apiGroups:
+  - sailoperator.io
```
gcs278, line 538: I think this is a sail library bug. https://github.com/openshift-service-mesh/sail-operator/pull/1083 added RBAC aggregates for istio + SailOperator CRDs via the sail library.

The sail library only ever installs istio.io CRDs. This is unnecessary for sail library to add this aggregate. But no action from you now - we must include this today or we will get failures.

manifests/00-cluster-role-sail-library.yaml
gcs278: BTW i think one file might be okay - but I think I'd prefer to split into generated sections for each source RBAC file we are making a superset for.

manifests/00-cluster-role-sail-library.yaml
```diff
@@ -67,29 +63,498 @@ rules:
...
+  - get
+  - list
+  - watch
+- apiGroups:
+  - extensions.istio.io
+  resources:
+  - trafficextensions
+  - wasmplugins
+  verbs:
+  - '*'
```
gcs278, line 522: just FYI - i agree with using `*` - it MUST match what sail library is doing (and sail library is using `*` because it's matching OLM's `*`).

Someone may complain later, but until then, this is fine.

gcs278 (commented):
You can leave these comments for a follow up - thanks for resolving the new 3.4.2 RBAC changes.

/lgtm

gcs278:
Created https://github.com/openshift-service-mesh/sail-operator/pull/1156 to follow up on the inconsistency in sail library i found.

Miciah:
> Someone might look at all this new RBAC and say "why are you adding unnecessary permissions?" but CIO's ClusterRole here MUST be a superset of whatever its service account creates (the istiod RBAC), otherwise you get a 403 Forbidden from the API server - that's important to note

Let's follow up with the Service Mesh team to discuss updating Helm charts to trim permissions that the controller as we configure it doesn't need.

cmd/ingress-operator/sail_rbac.go
Miciah: Related discussion: https://github.com/openshift/cluster-ingress-operator/pull/1548#discussion_r3875907701

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-5.1-e2e-aws-ovn-upgrade
/payload-aggregate periodic-ci-openshift-hypershift-release-5.1-periodics-e2e-aws-ovn-conformance-serial 5

rhamini3:
previously verified at [comment](https://github.com/openshift/cluster-ingress-operator/pull/1548#issuecomment-5722315016)
/verified by @rhamini3

PR #1555, opened 2026-08-19
WAF-3: Implement extension provider capability for WAF

rikatz:
As part of WAF observability story, Istio must be configured with an extra extension provider that has the capability of capture the state/attributes issued by the WAF engine and ship to an external OTEL collector.

This external OTEL collector will then be responsible to assembling and consolidating the logs into metrics, to be provided later to WAF users.

For Istio, this is mostly a no-op operation. The extensionProvider is configured, but only consumed when a Telemetry resource is created targetting the Gateway and this provider. Additionally, the implementation is kept simple as it will need to be backported, but also it may be removed/migrated once GatewayAPI implements TelemetryPolicy with logging capabilities

### DIFF
--- a/pkg/operator/controller/gatewayclass/controller_test.go
+++ b/pkg/operator/controller/gatewayclass/controller_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-logr/zapr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	sailv1 "github.com/istio-ecosystem/sail-operator/api/v1"
@@ -17,6 +18,9 @@ import (
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -276,6 +280,63 @@ func openshiftValues(enableInferenceExtension bool, operandNamespace string, gat
...
+			if err != nil {
+				log.Error(err, "invalid value for WAF collector host:port, skipping", "gatewayclass", gc, "value", wafOtelval)
+				break
+			}
+			wafCollectorPort, err := strconv.Atoi(port)
+			if err != nil {
+				log.Error(err, "invalid value for WAF collector port, skipping", "gatewayclass", gc, "value", wafOtelval)
+				break
+			}
+			val.MeshConfig.ExtensionProviders = []*sailv1.MeshConfigExtensionProvider{
```
rikatz: instead of setting, create or replace

pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -276,6 +280,70 @@ func openshiftValues(enableInferenceExtension bool, operandNamespace string, gat
...
+
+			if len(val.MeshConfig.ExtensionProviders) == 0 {
+				val.MeshConfig.ExtensionProviders = make([]*sailv1.MeshConfigExtensionProvider, 0)
+			}
+			val.MeshConfig.ExtensionProviders = append(val.MeshConfig.ExtensionProviders, &sailv1.MeshConfigExtensionProvider{
+
+				Name: new("waf-log-collector"),
+				EnvoyOtelAls: &sailv1.MeshConfigExtensionProviderEnvoyOpenTelemetryLogProvider{
+					Service: new(wafCollectorHost),
+					Port:    new(uint32(wafCollectorPort)),
```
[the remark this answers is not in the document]
rikatz: @coderabbitai a person with access to change a GatewayClass can do worst things. In this case, this is solely used internally, and it is a interhim solution. I don't think this should be fixed now

pkg/operator/controller/gatewayclass/istio_sail_installer.go
```diff
@@ -276,6 +280,71 @@ func openshiftValues(enableInferenceExtension bool, operandNamespace string, gat
...
+							"rule_id":        "%FILTER_STATE(wasm.io.coraza.waf.rule_id:PLAIN)%",
+							"action":         "%FILTER_STATE(wasm.io.coraza.waf.action:PLAIN)%",
+							"phase":          "%FILTER_STATE(wasm.io.coraza.waf.phase:PLAIN)%",
+							"waf_status":     "%FILTER_STATE(wasm.io.coraza.waf.status:PLAIN)%",
+							"severity":       "%FILTER_STATE(wasm.io.coraza.waf.severity:PLAIN)%",
+							"category":       "%FILTER_STATE(wasm.io.coraza.waf.category:PLAIN)%",
+						},
+					},
+				},
+			})
```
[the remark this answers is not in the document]
rikatz: @coderabbitai I don't know where this suggestion came from, but it is not right. This is OpenShift managed Gateway API, and this is where we can set the installation options.

pkg/operator/controller/gatewayclass/controller_test.go
```diff
@@ -267,6 +271,8 @@ func Test_Reconcile(t *testing.T) {
 		}
 	}
 
+	type sailInstallFunc func(version string, gieEnabled bool, proxyConfig map[string]string, gatewayclassesConfig json.RawMessage, openshiftTLS *config.OpenShiftTLS) *install.Options
```
gcs278: this isn't used, is this needed?

rikatz: probably no! I guess I forgot it when I was doing unit tests. Removed, squashed, pushed

gcs278 (approved):
No issues from me. `internal.do-not-use.openshift.io/...` is a new one. I suppose we are going for something like "it's supported for internal layered product use, but not supported for end user use".

It's understood that we have an idea for a more extensible solution, but IMO it's okay to merge this if WAF team would like to make incremental forward progress (aka I don't know how long an extensible config solution would take).

/lgtm

rikatz:
@gcs278 can I get approval here?

rikatz:
/verified by @rikatz

tested locally, and on a cluster, does as expected for WAF

PR #1557, opened 2026-08-19
OCPBUGS-112280: Preserve server-defaulted fields on init containers and volumes

jcmoraisjr:
deploymentConfigChanged() replaced InitContainers wholesale with the desired/expected value on every update, which reverts fields the API server defaults on the live object (e.g. TerminationMessagePath) and produces a noisy, misleading diff in the update log on every unrelated change. The primary container already avoided this by copying from current and selectively overlaying only the fields the operator manages; extract that logic into copyContainer() and apply it to every init container as well, syncing by index up to the length of the desired slice so any container beyond the two static ones (init-router, haproxy) still gets synced.

Since current is a live object, its InitContainers slice can differ in length from what's expected (e.g. right after an upgrade, or if it was manually edited). Guard against that by falling back to a verbatim copy of the desired InitContainers whenever the lengths don't match, before doing the selective per-field copy, so a mismatched live deployment self-heals instead of panicking on an out-of-range index.

hashableDeployment() built the hashable Containers and InitContainers independently, and the two had drifted: InitContainers included Args, RestartPolicy and VolumeMounts in the hash while Containers didn't, and neither hashed exactly the set of fields copyContainer() applies. That mismatch is exactly what caused the InitContainers bug above, and it silently allows the same failure to recur one field at a time. Extract the shared logic into hashableContainer(), used for both Containers and InitContainers, with a field list kept in sync with copyContainer() by a comment in each pointing at the other, so a field added to one and not the other fails loudly (detected as changed but never applied) instead of silently.

hashableDeployment() has the same class of issue for the projected service-account-token volume: DefaultMode and the DownwardAPI FieldRef.APIVersion are also defaulted by the API server, and were already handled for ConfigMap/Secret volumes but not for Projected ones, causing spurious hash mismatches. Normalize them the same way.

https://redhat.atlassian.net/browse/OCPBUGS-112280

### DIFF
--- a/pkg/operator/controller/ingress/deployment.go
+++ b/pkg/operator/controller/ingress/deployment.go
@@ -1697,23 +1697,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	hashableDeployment.Spec.Template.Spec.NodeSelector = deployment.Spec.Template.Spec.NodeSelector
 	containers := make([]corev1.Container, len(deployment.Spec.Template.Spec.Containers))
 	for i, container := range deployment.Spec.Template.Spec.Containers {
-		env := container.Env
-		sort.Slice(env, func(i, j int) bool {
-			return env[i].Name < env[j].Name
-		})
-		containers[i] = corev1.Container{
-			Command:         container.Command,
-			Env:             env,
-			Image:           container.Image,
-			ImagePullPolicy: container.ImagePullPolicy,
-			Name:            container.Name,
-			LivenessProbe:   hashableProbe(container.LivenessProbe),
-			ReadinessProbe:  hashableProbe(container.ReadinessProbe),
-			StartupProbe:    hashableProbe(container.StartupProbe),
-			Resources:       container.Resources,
-			SecurityContext: container.SecurityContext,
-			Ports:           container.Ports,
-		}
+		containers[i] = hashableContainer(container, false)
 	}
 	sort.Slice(containers, func(i, j int) bool {
 		return containers[i].Name < containers[j].Name
@@ -1721,26 +1705,7 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	hashableDeployment.Spec.Template.Spec.Containers = containers
 	initContainers := make([]corev1.Container, len(deployment.Spec.Template.Spec.InitContainers))
 	for i, initContainer := range deployment.Spec.Template.Spec.InitContainers {
-		env := initContainer.Env
-		sort.Slice(env, func(i, j int) bool {
-			return env[i].Name < env[j].Name
-		})
-		initContainers[i] = corev1.Container{
-			Args:            initContainer.Args,
-			Command:         initContainer.Command,
-			Env:             env,
-			Image:           initContainer.Image,
-			ImagePullPolicy: initContainer.ImagePullPolicy,
-			Name:            initContainer.Name,
-			LivenessProbe:   hashableProbe(initContainer.LivenessProbe),
-			ReadinessProbe:  hashableProbe(initContainer.ReadinessProbe),
-			StartupProbe:    hashableProbe(initContainer.StartupProbe),
-			SecurityContext: initContainer.SecurityContext,
-			Ports:           initContainer.Ports,
-			Resources:       initContainer.Resources,
-			RestartPolicy:   initContainer.RestartPolicy,
-			VolumeMounts:    initContainer.VolumeMounts,
-		}
+		initContainers[i] = hashableContainer(initContainer, true)
 	}
 	sort.Slice(initContainers, func(i, j int) bool {
 		return initContainers[i].Name < initContainers[j].Name
@@ -1760,6 +1725,26 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 		}
 		if vol.Secret != nil && vol.Secret.DefaultMode != nil && *vol.Secret.DefaultMode == int32(420) {
 			volumes[i].Secret.DefaultMode = nil
+		}
+		if vol.Projected != nil {
+			// 420 is the default value for DefaultMode for Projected volumes.
+			if vol.Projected.DefaultMode != nil && *vol.Projected.DefaultMode == int32(420) {
+				volumes[i].Projected.DefaultMode = nil
+			}
+			for _, source := range volumes[i].Projected.Sources {
+				// "v1" is the default value for FieldRef.APIVersion in DownwardAPI items.
+				if source.DownwardAPI != nil {
+					for _, item := range source.DownwardAPI.Items {
+						if item.FieldRef != nil && item.FieldRef.APIVersion == "v1" {
+							item.FieldRef.APIVersion = ""
+						}
+					}
+				}
+				// 3600 is the default value for ServiceAccountToken.ExpirationSeconds
+				if sat := source.ServiceAccountToken; sat != nil && sat.ExpirationSeconds != nil && *sat.ExpirationSeconds == 3600 {
+					sat.ExpirationSeconds = nil
+				}
+			}
 		}
 	}
 	sort.Slice(volumes, func(i, j int) bool {
@@ -1793,6 +1778,41 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 	hashableDeployment.Spec.Selector = deployment.Spec.Selector
 
 	return &hashableDeployment
+}
+
+// hashableContainer returns a copy of the given container containing only
+// the fields that should be used for computing the deployment hash. Keep
+// this in sync with copyContainer() — a field present here but missing
+// there is detected as changed but never actually applied.
+//
+// initContainer must be true for init containers: unlike the primary
+// router container, whose Args/Command users are allowed to customize by
+// hand, init containers are fully defined by the manifest, so their
+// Args/Command are also hashed/copied to keep them in sync with it.
+func hashableContainer(container corev1.Container, initContainer bool) corev1.Container {
+	env := container.Env
+	sort.Slice(env, func(i, j int) bool {
+		return env[i].Name < env[j].Name
+	})
+	hashContainer := corev1.Container{
+		Name:            container.Name,
+		Env:             env,
+		Image:           container.Image,
+		ImagePullPolicy: container.ImagePullPolicy,
+		LivenessProbe:   hashableProbe(container.LivenessProbe),
+		Ports:           container.Ports,
+		ReadinessProbe:  hashableProbe(container.ReadinessProbe),
+		Resources:       container.Resources,
+		RestartPolicy:   container.RestartPolicy,
+		SecurityContext: container.SecurityContext,
+		StartupProbe:    hashableProbe(container.StartupProbe),
+		VolumeMounts:    container.VolumeMounts,
+	}
+	if initContainer {
+		hashContainer.Args = container.Args
+		hashContainer.Command = container.Command
+	}
+	return hashContainer
 }
 
 // cmpMatchExpressions is a helper for hashableDeployment.
@@ -1933,7 +1953,17 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
 		containers[i+1] = *container.DeepCopy()
 	}
 	updated.Spec.Template.Spec.Containers = containers
-	updated.Spec.Template.Spec.InitContainers = expected.Spec.Template.Spec.InitContainers
+
+	// Only overrides init containers if current does not have the expected size.
+	expectedInitContainersCount := len(expected.Spec.Template.Spec.InitContainers)
+	if len(updated.Spec.Template.Spec.InitContainers) != expectedInitContainersCount {
+		initContainers := make([]corev1.Container, len(expected.Spec.Template.Spec.InitContainers))
+		for i, initContainer := range expected.Spec.Template.Spec.InitContainers {
+			initContainers[i] = *initContainer.DeepCopy()
+		}
+		updated.Spec.Template.Spec.InitContainers = initContainers
+	}
+
 	updated.Spec.Template.Spec.AutomountServiceAccountToken = expected.Spec.Template.Spec.AutomountServiceAccountToken
 	updated.Spec.Template.Spec.ShareProcessNamespace = expected.Spec.Template.Spec.ShareProcessNamespace
 	updated.Spec.Template.Spec.DNSPolicy = expected.Spec.Template.Spec.DNSPolicy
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
@melvinjoseph86 worth testing an upgrade scenario, from 4.22 to 5.0 with this patch: it touches the initContainers slice, which does not exist in 4.22.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1933,7 +1949,17 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
...
+
+	// Only overrides init containers if current does not have the expected size.
+	expectedInitContainersCount := len(expected.Spec.Template.Spec.InitContainers)
+	if len(updated.Spec.Template.Spec.InitContainers) != expectedInitContainersCount {
+		initContainers := make([]corev1.Container, len(expected.Spec.Template.Spec.InitContainers))
+		for i, initContainer := range expected.Spec.Template.Spec.InitContainers {
+			initContainers[i] = *initContainer.DeepCopy()
+		}
+		updated.Spec.Template.Spec.InitContainers = initContainers
+	}
```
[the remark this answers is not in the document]
jcmoraisjr: Fair point, just updated as suggested.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1761,6 +1726,22 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
 		if vol.Secret != nil && vol.Secret.DefaultMode != nil && *vol.Secret.DefaultMode == int32(420) {
 			volumes[i].Secret.DefaultMode = nil
 		}
+		if vol.Projected != nil {
+			// 420 is the default value for DefaultMode for Projected volumes.
+			if vol.Projected.DefaultMode != nil && *vol.Projected.DefaultMode == int32(420) {
```
gcs278, line 1731: It's a minor point, but there is a slight inconsistency here - it's missing another defensive nil-ing of 
`volumes[].projected.sources[].serviceAccountToken.expirationSeconds`.

We defensively nil out Projected.DefaultMode to guard against someone removing the explicit defaultMode: 0400 and inadvertently relying on the k8s default.

`expirationSeconds` also gets [defaulted](https://github.com/openshift/kubernetes/blob/master/pkg/apis/core/v1/defaults.go#L276) to 3600s. To be consistent, I think you would want to add a similar style nil for `expirationSeconds`.

It's not a problem because we explicitly set `expirationSeconds`, but if someone removed the explicit set and didn't update this, that would expose a bug. A code "foot gun" if you will.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1795,6 +1776,33 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
...
+// this in sync with copyContainer() — a field present here but missing
+// there is detected as changed but never actually applied.
+func hashableContainer(container corev1.Container) corev1.Container {
+	env := container.Env
+	sort.Slice(env, func(i, j int) bool {
+		return env[i].Name < env[j].Name
+	})
+	return corev1.Container{
+		Name:            container.Name,
+		Args:            container.Args,
```
gcs278: I think this might of been what @Miciah was referring to in bug status meeting. We previously allowed modification of Args. I actually used this quite a lot in my debugging/hack scripts: https://github.com/gcs278/openshift-hacks/blob/master/oc_functions#L303. I used command, but I probably could have easily used `args`.

I'd have to do some research, but there might be support folks relying on this "hack" to get verbose logging. We don't expose router logging verbosity via any API field, so this unsupported hack is actually pretty important.

Here's a Claude generated visual, which points out the inconsistencies that you are fixing:

 | Field           | 4.22 hashableDeployment | 4.22 deploymentConfigChanged | pre-PR hashableDeployment | pre-PR deploymentConfigChanged | post-PR hashableDeployment | post-PR deploymentConfigChanged |
| --------------- | ----------------------- | ---------------------------- | ----------------------------- | ---------------------------------- | ------------------------------ | ----------------------------------- |
| Args            | —                       | —                            | —                             | —                                  | ✓                              | ✓                                   |
| Command         | ✓                       | —                            | ✓                             | —                                  | ✓                              | ✓                                   |
| Env             | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| Image           | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| ImagePullPolicy | ✓                       | —                            | ✓                             | —                                  | ✓                              | ✓                                   |
| LivenessProbe   | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| Ports           | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| ReadinessProbe  | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| Resources       | —                       | —                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| RestartPolicy   | —                       | —                            | —                             | —                                  | ✓                              | ✓                                   |
| SecurityContext | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| StartupProbe    | ✓                       | ✓                            | ✓                             | ✓                                  | ✓                              | ✓                                   |
| VolumeMounts    | —                       | ✓                            | —                             | ✓                                  | ✓                              | ✓                                   |

As for other fields, like `Resources` or `RestartPolicy`, I don't have a solid reason to suggest to not reconcile. But I (maybe selfishly) would recommend to omit args/command for potential verbosity hack. @Miciah might have more historical information.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2284,11 +2360,35 @@ func Test_deploymentConfigChanged(t *testing.T) {
...
+			description: "if the current init container slice has a 3rd container",
+			mutateCurr: func(deployment *appsv1.Deployment) {
+				container := corev1.Container{}
+				deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, container)
+			},
+			expect: true,
+			assert: func(t *testing.T, deployment *appsv1.Deployment) {
+				assert.Len(t, deployment.Spec.Template.Spec.InitContainers, 2)
+			},
 		},
```
gcs278, line 2427: Should we add a unit test to verify the bug is fixed by changing a non-managed field?


```suggestion
		},
		{
        description: "init container k8s-defaulted fields are preserved after reconcile",
        mutateCurr: func(deployment *appsv1.Deployment) {
        // Simulate what k8s injects on save
        deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePath = "/dev/termination-log"
        deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePolicy = corev1.TerminationMessageReadFile
        },
        expect: false, // k8s-defaulted fields are not operator-managed, no change expected
        },
```

gcs278 (commented):
Nice fixes! sorry I didn't catch this in the initial side car changes. But now I understand slightly better after some digging.

What do you think about writing an E2E that watches for spurious deployment updates? Ironically, it could use a similar mechanism to the [router status conflict test:](https://github.com/openshift/origin/blob/main/test/extended/router/stress.go#L132). This seems very fragile to get right, and an E2E feels very valuable. But if you'd like to leave it for a follow up that's fine too.

melvinjoseph86:
Checked with upgrade from 4.22 to 5.0 worked fine and QE falling test cases on ingress-operator is passing [test upgrade 4.22.0-0.nightly-2026-08-19-204145 5.0.0-0.nightly-2026-08-18-203845,https://github.com/openshift/cluster-ingress-operator/pull/1557 aws](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2090463025529098240)

melvinjoseph86:
i tried upgrade using cluster bot only (which i mentioned above) as i cannot upgrade a 4.22 cluster to cluster with this PR.
Also i check the QE test cases and all are passing, hence marking as verified 
/verified by @mjoseph

pkg/operator/controller/ingress/deployment.go
jcmoraisjr: Thanks for building and sharing this table, really valuable to better understand how things are connected each other.

I do understand the arguments about Args and Command, in fact they'd break hacking in case we want to support them - even if unofficially. My arguments about adding them in the list is that they are being configured in the manifest, and if we remove them from this list, further updates wouldn't be reflected on a running, upgraded cluster. We can use a midterm approach here, creating distinct lists for container (avoid breaking legacy hacks) and initContainer (pretty new right now), this however would make the code a bit more complicated.

You folks have a really good understanding about the internals, so my vote is to simply follow your votes. :slightly_smiling_face:

gcs278: > My argument for keeping them in the list is that they are configured in the manifest, and if we remove them, future manifest updates wouldn't be reflected on a running, upgraded cluster.

Yea I suppose there is a slight difference in how router and haproxy28/32 containers configure their commands:
- Router containers use an [entrypoint](https://github.com/openshift/router/blob/master/images/router/haproxy/Dockerfile.ocp#L22) in the Dockerfile — so desired `command` and `args` are nil
- HAProxy sets them explicitly via [command](https://github.com/openshift/cluster-ingress-operator/blob/c8b5e9591e590f4b13d1d5d7acb5e08f7eb57a24/pkg/manifests/assets/router/deployment.yaml#L40) and [args](https://github.com/openshift/cluster-ingress-operator/blob/c8b5e9591e590f4b13d1d5d7acb5e08f7eb57a24/pkg/manifests/assets/router/deployment.yaml#L42) in the k8s deployment manifest

Neither is updated via `deployment.go` code. But since haproxy has real non-nil values for `command` and `args`, they probably _should_ be in the managed field set (though, it doesn't technically have to be) — if those values ever changed in a manifest update (e.g., a new haproxy script flag), the operator needs to detect and apply the change. You could argue a bit of a "footgun" for future developers if we didn't reconcile haproxy args.

The subtle but notable difference since haproxy is setting _something_ meaningful, while the router is setting _nothing_ in the deployment. But this seems to stem from the fact that `start-haproxy` isn't embedded in the haproxy container image, but instead copied from the router via init container startup.

What do you think about keeping your current helper structure, so that HAProxy container commands/args are fully managed, but we "restore" the router args/command so that they effectively unmanaged (if it looks reasonable to do in the code, i'm not sure...)? The benefit of that, is we can explicitly document that the router opts out of managing command/args in the code with a "why" comment instead and keep the code common. 

Side Note: Resources should definitely be managed - I just remember now that we updated the resources in the sidecar model. I have no good reason why restart policy should NOT be managed - in other words - my only concern is args/command on the router.

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1795,6 +1776,33 @@ func hashableDeployment(deployment *appsv1.Deployment, onlyTemplate bool) *appsv
...
 
+// hashableContainer returns a copy of the given container containing only
+// the fields that should be used for computing the deployment hash. Keep
+// this in sync with copyContainer() — a field present here but missing
+// there is detected as changed but never actually applied.
+func hashableContainer(container corev1.Container) corev1.Container {
+	env := container.Env
+	sort.Slice(env, func(i, j int) bool {
+		return env[i].Name < env[j].Name
+	})
```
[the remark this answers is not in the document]
jcmoraisjr: There is already a helper for that, `checkDeploymentHasEnvSorted()`, and a number of tests already call it after creating a desired router deployment. Also, although containers are passed by value, their Env field is a reference to the Env slice which is impacted by the `sort.Slice()` call in `hashableContainer()`. I'm wondering if I misunderstood your concern.

pkg/operator/controller/ingress/deployment.go
jcmoraisjr: Good one, just added.

pkg/operator/controller/ingress/deployment.go
jcmoraisjr: I parameterized and added unit tests that distinguish between containers and initContainers. How does it sound?

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2015,6 +2015,29 @@ func TestDeploymentHash(t *testing.T) {
...
+				deployment.Spec.Template.Spec.Volumes[0].Projected.Sources[0].DownwardAPI.Items[0].FieldRef.APIVersion = "v1"
+			},
+		},
+		{
+			description: "if projected volume defaultMode is defaulted to 420",
+			mutate: func(deployment *appsv1.Deployment) {
+				deployment.Spec.Template.Spec.Volumes[0].Projected.DefaultMode = ptr.To[int32](420)
+			},
+		},
 	}
```
gcs278, line 2050: maybe worth adding a unit test for the added expirationSeconds


```suggestion
	}
	{
      description: "if projected volume serviceAccountToken expirationSeconds is defaulted to 3600",
      mutate: func(deployment *appsv1.Deployment) {
          deployment.Spec.Template.Spec.Volumes[0].Projected.Sources[0].ServiceAccountToken = &corev1.ServiceAccountTokenProjection{
              ExpirationSeconds: ptr.To[int64](3600),
              Path:              "token",
          }
      },
  },
```

pkg/operator/controller/ingress/deployment.go
```diff
@@ -1981,6 +2006,32 @@ func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv
...
+	to.Env = from.Env
+	to.Image = from.Image
+	to.ImagePullPolicy = from.ImagePullPolicy
+	copyProbe(from.LivenessProbe, to.LivenessProbe, true)
+	to.Ports = from.Ports
+	copyProbe(from.ReadinessProbe, to.ReadinessProbe, true)
+	to.Resources = from.Resources
+	to.RestartPolicy = from.RestartPolicy
+	to.SecurityContext = from.SecurityContext
+	copyProbe(from.StartupProbe, to.StartupProbe, true)
```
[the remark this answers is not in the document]
jcmoraisjr: @coderabbitai I tend to agree with you, although this is the same behavior copied before this refactor, so changing it here would change the behavior before this PR. So I'd postpone it to another PR in case we want to fix it.

@gcs278 do you consider this something to be fixed?

https://github.com/openshift/cluster-ingress-operator/blob/d0f14ef2a0ff1c3a6baea6536096aec5511a3754/pkg/operator/controller/ingress/deployment.go#L1987-L1990

It seems that if we add (when missing) or remove (when assigned) we should be skipping to copy, which is not in sync with the hash counterpart.

pkg/operator/controller/ingress/deployment_test.go
jcmoraisjr: Yup, just added.

pkg/operator/controller/ingress/deployment.go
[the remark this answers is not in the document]
gcs278: @jcmoraisjr yea seems like a footgun for a future developer - and I think we have previous and future changes to the probes that make this more important.

@coderabbitai is this a problem for https://github.com/openshift/cluster-ingress-operator/pull/1528? Will our router containers not have their startupprobe updated on existing clusters? Or does it still update the deployment if other things change on upgrade (like container image)?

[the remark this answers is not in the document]
gcs278: ah so only when nil - yea @jcmoraisjr I suppose worth fixing for completeness, but not the biggest footgun - it's only if we ever remove a probe that it might be a problem. Feel free to follow up with it if you want, but I don't think it's critical.

pkg/operator/controller/ingress/deployment_test.go
```diff
@@ -2284,11 +2395,44 @@ func Test_deploymentConfigChanged(t *testing.T) {
...
+			},
+		},
+		{
+			description: "if the init container k8s-defaulted fields are preserved",
+			mutateCurr: func(deployment *appsv1.Deployment) {
+				// Simulate what k8s injects on save
+				deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePath = "/dev/termination-log"
+				deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePolicy = corev1.TerminationMessageReadFile
+			},
+			expect: false, // k8s-defaulted fields are not operator-managed, no change expected
```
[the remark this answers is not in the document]
gcs278: @coderabbitai something like this?

```
{
      description: "if the init container k8s-defaulted fields are preserved during an update",
      mutateCurr: func(deployment *appsv1.Deployment) {
          deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePath = "/dev/termination-log"
          deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePolicy = corev1.TerminationMessageReadFile
      },
      mutate: func(deployment *appsv1.Deployment) {
          deployment.Spec.Template.Spec.InitContainers[0].Image = "new-image:latest"
      },
      expect: true,
      assert: func(t *testing.T, deployment *appsv1.Deployment) {
          assert.Equal(t, "/dev/termination-log", deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePath)
          assert.Equal(t, corev1.TerminationMessageReadFile, deployment.Spec.Template.Spec.InitContainers[0].TerminationMessagePolicy)
      },
  },
```

pkg/operator/controller/ingress/deployment.go
gcs278: Yea that sounds like a reasonable approach. I know it's a bit inconsistent, but I think it's safer to not stomp on arg/commands for router (well, at least not instantly) to keep this verbosity workaround enabled.

We can reassess if/when the router deployment needs a command/arg change. Right now, they have never changed, and it's safe to exclude them.

pkg/operator/controller/ingress/deployment_test.go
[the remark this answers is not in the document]
jcmoraisjr: Good suggestion, just updated.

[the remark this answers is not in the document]
jcmoraisjr: @coderabbitai do you consider your report as resolved?

gcs278:
thank for all of the updates!
/approve
/lgtm

rhamini3:
@melvinjoseph86 previously [verified](https://github.com/openshift/cluster-ingress-operator/pull/1557#issuecomment-5369450313) this PR with QE tests and all CI and unit tests are passing
/verified by @rhamini3

PR #1547, opened 2026-08-10
NE-2779: Implement Gateway API management mode

rikatz:
This change implements the new Gateway API management mode, allowing users to fully disable managed Gateway API from OCP, and also to come back to a managed mode.

Additionally this change exposes the Gateway API CRDs on the container image, allowing support engineers to provide a clear rollback path to managed mode in case it is required.

PR #1560, opened 2026-08-21
OCPBUGS-91026: Add missing KAS-defaulted fields to ValidatingAdmissionPolicy manifest

natamm4:
Add matchPolicy, namespaceSelector, objectSelector, and scope to manifests/01-validating-admission-policy.yaml. These fields are injected as defaults by the API server, causing CVO to detect a diff on every reconciliation cycle and update loop indefinitely.

### DIFF
--- a/manifests/01-validating-admission-policy.yaml
+++ b/manifests/01-validating-admission-policy.yaml
@@ -14,7 +14,7 @@ spec:
       apiVersions: ["v1"]
       operations:  ["CREATE","UPDATE","DELETE"]
       resources:   ["customresourcedefinitions"]
-      scope: “*”
+      scope: "*"
     matchPolicy: Equivalent
     namespaceSelector: {}
     objectSelector: {}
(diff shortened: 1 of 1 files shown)


rikatz:
/approve
/lgtm

I have confirmed this is an issue, and given I am working on the GAteway API Management Knob it would be good to have this fixed first.

rikatz:
btw this was identified by Coderabbit on my PR for Management Knob as well: https://github.com/openshift/cluster-ingress-operator/pull/1547#discussion_r3752168702

rhamini3:
Verified on a 5.1 cluster that the VAP contains the updated fields
```
  matchConstraints:
    matchPolicy: Equivalent
    namespaceSelector: {}
    objectSelector: {}
    resourceRules:
    - apiGroups:
      - apiextensions.k8s.io
      apiVersions:
      - v1
      operations:
      - CREATE
      - UPDATE
      - DELETE
      resources:
      - customresourcedefinitions
      scope: '*'
```
/verified by @rhamini3

rikatz:
@natamm4 does this needs to be cherry-picked to earlier versions?

natamm4:
@rikatz I don't think there is a specific customer need to, so no for now I'm thinking.

PR #1564, opened 2026-08-26
[release-4.22] OCPBUGS-112634: Support Alternate Universe Domain

patrickdillon:
Manual cherry-pick of #1515 to get GCD support into 4.22


gcs278:
No major issue. I notice this implementation is slightly different than https://github.com/openshift/cluster-ingress-operator/pull/1515, presumably because in this backport we only bump to `google.golang.org/api v0.199.0` which is different than `v0.289.0` used in 5.0. That feels acceptable - if we are trying to do the bare minimum go mod bumps to get Universal Domain support. 

@bentito feel free to take a look too.

/approve
/lgtm

gcs278:
/payload-job periodic-ci-openshift-release-main-ci-4.22-upgrade-from-stable-4.21-e2e-gcp-ovn-upgrade

bentito:
I think it would be good to add unit coverage for the new New:
The existing `provider_test.go` file only tests `ParseZone`. It does not cover the initialization logic in `New(...)`

patrickdillon:
> I think it would be good to add unit coverage for the new New: The existing `provider_test.go` file only tests `ParseZone`. It does not cover the initialization logic in `New(...)`

You mean specifically for these changes or for the `New` function  as a whole? These changes are exercised by the e2e tests. Looking at the existing function signature of `New` it's not obvious to me what a unit test related to the GCP client initialization would look like. 

If we wanted unit tests for the GCP client initialization logic, we would need to encapsulate that logic into helper functions and put unit tests on those, but that would mean drift from the logic in later versions as well.

bentito:
> > I think it would be good to add unit coverage for the new New: The existing `provider_test.go` file only tests `ParseZone`. It does not cover the initialization logic in `New(...)`
> 
> You mean specifically for these changes or for the `New` function as a whole? These changes are exercised by the e2e tests. Looking at the existing function signature of `New` it's not obvious to me what a unit test related to the GCP client initialization would look like.
> 
> If we wanted unit tests for the GCP client initialization logic, we would need to encapsulate that logic into helper functions and put unit tests on those, but that would mean drift from the logic in later versions as well.

yeah fair points. I think you're right, e2e's are enough here.

Miciah:
/label backport-risk-assessed
per https://github.com/openshift/cluster-ingress-operator/pull/1564#issuecomment-5534427672.

Miciah:
Chai-bot has identified the hypershift-e2e-aks failures as attributable to [TRT-2950](https://redhat.atlassian.net/browse/TRT-2950).  

PR #1568, opened 2026-08-28
CORS-4530: Azure Gov Secret IL6 region support

jhixson74:
Azure Government Secret (IL6) environments require resolving endpoints  from the ARM endpoint URL rather than by name.


aswinsuryan:
Looks good to me. It would be nice to add a unit test covering the AzureUSSecCloud ARM-endpoint resolution path, but otherwise this looks good.

PR #1575, opened 2026-09-02
OCPBUGS-105398: refactor: remove Azure workload identity feature gate

damdo:
## What does this PR do?

Removes the `AzureWorkloadIdentity` feature-gate dependency from the
cluster ingress operator.

The feature is GA and enabled by default, so the previously gated Azure
workload identity behavior is now unconditional when no client secret is
available.

Changes include:
- Removing feature-gate observation and configuration plumbing.
- Removing the Azure DNS provider/client boolean parameter.
- Using workload identity whenever the Azure client secret is empty.

## Why is this needed?

This must merge before `openshift/api#3018`, which removes the feature-gate
definition. Remaining `.Enabled()` consumers would panic once the gate is
removed from the cluster-served feature-gate list.

Related:
- openshift/api#3018

### DIFF
--- a/pkg/dns/azure/client/client.go
+++ b/pkg/dns/azure/client/client.go
@@ -40,15 +40,11 @@ type Config struct {
 	// if Azure workload identity is not used.
 	ClientSecret string
 	// FederatedTokenFile is the path to a file containing a workload
-	// identity token.  If FederatedTokenFile is specified and
-	// AzureWorkloadIdentityEnabled is true, then Azure workload identity is
-	// used instead of using a client secret.
+	// identity token. If no client secret is available, Azure workload
+	// identity is used instead.
 	FederatedTokenFile string
 	// TenantID is the Azure tenant ID.
 	TenantID string
-	// AzureWorkloadIdentityEnabled indicates whether the
-	// "AzureWorkloadIdentity" feature gate is enabled.
-	AzureWorkloadIdentityEnabled bool
 }
 
 // ARecord is a DNS A record.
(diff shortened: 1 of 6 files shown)


Miciah:
Thanks!

/approve

@damdo, the PR description is detailed, but the commit message is missing a body.  Would you mind copying the description and a reference to https://redhat.atlassian.net/browse/OCPBUGS-105398 into the commit message?

Also, your PR might as well delete the reference to the featuregate in `AGENTS.md`: https://github.com/openshift/cluster-ingress-operator/blob/8e5b7c7c17890739041da5e8f28700d82b937aa7/AGENTS.md#L330

damdo:
@Miciah Thanks for the review and approval. I’ve updated the commit message with the PR description and the OCPBUGS-105398 reference, and removed the stale `AzureWorkloadIdentity` entry from `AGENTS.md`. The changes are now pushed.

Miciah:
Thanks!

/approve
/lgtm

pkg/dns/azure/client/client.go
```diff
@@ -40,15 +40,11 @@ type Config struct {
 	// if Azure workload identity is not used.
 	ClientSecret string
 	// FederatedTokenFile is the path to a file containing a workload
-	// identity token.  If FederatedTokenFile is specified and
```
davidesalerno, line 43: This wording looks directionally right after removing the feature gate, but it may be clearer to keep the relationship to `FederatedTokenFile` explicit. As written, “If no client secret is available” could read as though workload identity is used unconditionally, even though this field specifically represents the token file path. Consider something like: “FederatedTokenFile is the path to a file containing a workload identity token. If no client secret is available, this token file is used for Azure workload identity.”

davidesalerno (commented):
just one small nit about wording on a comment

pkg/dns/azure/client/client.go
damdo: I will address this in a follow-up, I prefer not to lose all the green ci passes I have here :)

damdo: Follow-up PR here: https://github.com/openshift/cluster-ingress-operator/pull/1575

PR #1585, opened 2026-09-10
OCPBUGS-114344: Detect unexpected router deployment updates

jcmoraisjr:
Once an ingresscontroller reports Available, its router deployment should stay untouched unless something explicitly needs to change it. There was no way to catch a regression that causes the operator to needlessly re-trigger a deployment update after readiness, e.g. a spurious rollout in reaction to an unrelated change to a dependent resource, or a misconfigured deployment hash after adding new fields in the deployment manifest.

Add generic captureObject/verifyGeneration helpers in util_test.go: snapshot an object's generation once the ingresscontroller is ready, then, after exercising the scenario under test, wait briefly for any async reconcile to land and assert the generation didn't change. Generation (rather than resource version) is used so status-only updates don't cause false failures.

https://redhat.atlassian.net/browse/OCPBUGS-114344

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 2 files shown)


test/e2e/operator_test.go
```diff
@@ -2111,9 +2112,16 @@ func TestNodePortServiceEndpointPublishingStrategy(t *testing.T) {
...
-		t.Errorf("failed to observe expected conditions: %v", err)
+		t.Fatalf("failed to observe expected conditions: %v", err)
 	}
 
+	// Capture the deployment's state now so we can assert the Ingress Controller
+	// deployment wasn't updated during the lifetime of the test. 5s is a heuristic
+	// margin for the ingress controller to react after the last Service update.
+	captured, err := captureObject(context.Background(), &appsv1.Deployment{}, controller.RouterDeploymentName(ing))
+	require.NoError(t, err)
+	defer captured.verifyGeneration(context.Background(), t, 5*time.Second)
```
jcmoraisjr, line 2123: This is an usage example, the idea is to incorporate it on more e2e tests.

test/e2e/util_test.go
```diff
@@ -850,6 +851,46 @@ func updateInfrastructureConfigStatusWithRetryOnConflict(t *testing.T, timeout t
...
+// verifyGeneration verifies that the object's generation is unchanged. It waits
+// `verifyAfter` before checking, to give asynchronous writers (e.g. a controller
+// reconcile loop) a chance to persist a change.
+//
+// This is a best-effort check: it only detects a change happening within
+// `verifyAfter` of the method call. A write queued behind a longer requeue/backoff
+// will not be caught. Choose `verifyAfter` based on the reconcile path under test
+// (e.g. a multiple of its expected requeue delay), and treat a pass as "no churn
+// observed within N seconds", not "no churn is possible".
+func (g *capturedObject) verifyGeneration(ctx context.Context, t *testing.T, verifyAfter time.Duration) {
```
jcmoraisjr, line 880: `g` seems misleading.

(in fact the type was renamed and I missed to update the varname)

PR #1586, opened 2026-09-11
OCPBUGS-105398: docs: clarify Azure workload identity token file

damdo:
## What does this PR do?

Clarifies the `FederatedTokenFile` documentation in the Azure DNS client.
The comment now explicitly states that the token file is used for Azure
workload identity when no client secret is available.

## Why is this needed?

This follow-up addresses review feedback from cluster-ingress-operator#1575:

https://github.com/openshift/cluster-ingress-operator/pull/1575#discussion_r3980393718

The wording keeps the relationship between `FederatedTokenFile` and Azure
workload identity explicit without implying that workload identity is used
without a token file.

Related:
- https://github.com/openshift/cluster-ingress-operator/pull/1575
- https://redhat.atlassian.net/browse/OCPBUGS-105398


davidesalerno (approved):
/approve
/lgtm

PR #1576, opened 2026-09-02
OCPBUGS-105317: Bump to OSSM 3.4.2 and istio 1.30.4

rhamini3:
This PR is bumping to latest OSSM 3.4.2 with latest istio. 

This bump helps address some CVE fixes. The bump to OSSM 3.4.2 also reverts https://github.com/openshift/cluster-ingress-operator/pull/1540 which was used for a workaround.

PR #1587, opened 2026-09-11
Enable ListenerSet conformance tests for Istio 1.30.4

gcs278:
## Summary
- Adds `ListenerSet` to `SUPPORTED_FEATURES` in `hack/gatewayapi-conformance.sh`
- Istio 1.30 added ListenerSet support, so we should run these conformance tests against the OSSM 3.4.2 / Istio 1.30.4 bump from #1576

Depends on #1576

## Test plan
- [ ] Run Gateway API conformance suite and verify ListenerSet tests pass


gcs278:
Verdict - ListenerSet is still failing `ListenerSetAllowedRoutesSupportedKinds` -  that appears like it should be fixed with an upstream fix coming in 1.30.5 - likely OSSM 3.4.3.

PR #1589, opened 2026-09-11
NE-2906: propagate IngressController tolerations to canary daemonset

davidesalerno:
## Description

The canary daemonset previously only tolerated `node-role.kubernetes.io/infra` (hardcoded in the manifest YAML). Clusters using custom node taints — e.g. for SR-IOV, GPU, or other workload-specialized nodes — configured those same taints as tolerations on the default IngressController so the router could schedule there. Without the same tolerations on the canary daemonset, canary pods could not be scheduled on those nodes, causing spurious degraded conditions on the default IngressController.

This change propagates `spec.nodePlacement.tolerations` from the default IngressController to the canary daemonset on every reconcile iteration:

- The hardcoded base toleration (`node-role.kubernetes.io/infra: Exists`) is always preserved.
- Custom tolerations are merged after the base set; exact duplicates (same key, operator, value, and effect) are not added twice.
- When `spec.nodePlacement` is unset or tolerations is empty, the daemonset behaves exactly as before (backward compatible).
- No new CRD field or API change is required; the existing `spec.nodePlacement.tolerations` field is reused.

## Jira

https://redhat.atlassian.net/browse/NE-2906

---

Always review AI generated responses prior to use.
AI-assisted response via openshift-developer plugin


pkg/operator/controller/canary/daemonset.go
```diff
@@ -195,9 +195,32 @@ func desiredCanaryDaemonSet(canaryImage string, certHash string, tlsProfileSpec
...
+// mergeTolerationsWithDedup returns the base tolerations with any custom tolerations
+// appended that are not already present, using cmpTolerations for equivalence.
+// The base tolerations are always preserved as-is.
+func mergeTolerationsWithDedup(base, custom []corev1.Toleration) []corev1.Toleration {
+	merged := make([]corev1.Toleration, len(base))
+	copy(merged, base)
+	for _, c := range custom {
+		found := false
+		for _, b := range merged {
+			if cmpTolerations(b, c) {
```
pedjak: **issue (bug):** `cmpTolerations` is an equality predicate, but line 249 passes it to `cmpopts.SortSlices` which requires a strict less-than comparator.

`cmpTolerations` returns `true` when two tolerations are equal. `cmpopts.SortSlices` interprets `true` as "a < b", so equal elements violate irreflexivity and distinct elements are all treated as equivalent — the sort is effectively a no-op.

This is pre-existing, but with this PR adding custom tolerations, multi-element lists become common. If the API server returns tolerations in a different order than `mergeTolerationsWithDedup` produces, `canaryDaemonSetChanged` will trigger a needless DaemonSet update every reconcile cycle.

Additionally, `mergeTolerationsWithDedup` correctly uses `cmpTolerations` for equality here, creating incompatible contracts on the same function — fixing it for `SortSlices` (to `a.Key < b.Key`) would silently break this dedup logic.

**Suggestion:** split into two functions — keep `cmpTolerations` for equality (used here in dedup), and add a `lessToleration` for `SortSlices`:
```go
func lessToleration(a, b corev1.Toleration) bool {
    return a.Key < b.Key
}
```

pkg/operator/controller/canary/daemonset_test.go
```diff
@@ -274,6 +277,160 @@ func Test_canaryDaemonsetChanged(t *testing.T) {
...
+	r := &reconciler{
+		config: Config{CanaryImage: "test-image"},
+		client: fake.NewClientBuilder().WithScheme(scheme).WithObjects(initial).Build(),
+	}
+	custom := corev1.Toleration{
+		Key:      "custom.io/workload",
+		Operator: corev1.TolerationOpExists,
+		Effect:   corev1.TaintEffectNoSchedule,
+	}
+	want := append(initial.Spec.Template.Spec.Tolerations, custom)
```
pedjak: **suggestion (test-robustness):** `append(initial.Spec.Template.Spec.Tolerations, custom)` may alias the backing array of `initial`, which is already stored in the fake client.

If a future change produces a base tolerations slice with `cap > len`, the `append` mutates `initial`'s backing array in-place. The fake client then returns the already-mutated object on subsequent `Get` calls, so the idempotency assertion (`resourceVersion` unchanged) passes vacuously.

**Suggestion:**
```go
want := slices.Concat(initial.Spec.Template.Spec.Tolerations, []corev1.Toleration{custom})
```

pkg/operator/controller/canary/controller.go
```diff
@@ -293,13 +305,7 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
 
-	// Get the canary route rotation annotation value
-	// from the default ingress controller.
-	ic := &operatorv1.IngressController{}
-	if err := r.client.Get(ctx, request.NamespacedName, ic); err != nil {
-		return result, fmt.Errorf("failed to get ingress controller %s: %w", request.NamespacedName.Name, err)
-	}
-
+	// Get the canary route rotation annotation value from the default ingress controller.
```
pedjak: is this comment really needed to be added in PR?

pkg/operator/controller/canary/controller.go
```diff
@@ -256,7 +256,19 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		tlsProfileSpec = operatorcontroller.TLSProfileSpecForSecurityProfile(apiConfig.Spec.TLSSecurityProfile)
 	}
 
-	haveDs, daemonset, err := r.ensureCanaryDaemonSet(ctx, tlsProfileSpec)
+	// Fetch the default IngressController to propagate its node placement tolerations
+	// to the canary daemonset so canary pods can be scheduled on nodes with custom taints.
+	ic := &operatorv1.IngressController{}
+	if err := r.client.Get(ctx, request.NamespacedName, ic); err != nil {
+		return result, fmt.Errorf("failed to get ingress controller %s: %w", request.NamespacedName.Name, err)
```
pedjak: **issue (bug):** The error handling here doesn't distinguish `NotFound` from transient errors. If the IngressController has been deleted or doesn't exist yet, this returns a wrapped error that will cause infinite requeue attempts.

The original code (line 298 before this PR) had the same issue, but it only blocked reading the rotation annotation — now it blocks all canary resource reconciliation (daemonset, service account, service, route).

**Suggestion:** handle `NotFound` explicitly — if the IC is gone, there's nothing to reconcile:
```go
if err := r.client.Get(ctx, request.NamespacedName, ic); err != nil {
	if apierrors.IsNotFound(err) {
		return result, nil
	}
	return result, fmt.Errorf("failed to get ingress controller %s: %w", request.NamespacedName.Name, err)
}
```

pkg/operator/controller/canary/daemonset.go
pedjak: **issue (simplification):** Use the upstream `Toleration.MatchToleration()` from `k8s.io/api/core/v1` instead of the custom `cmpTolerations` here:
```go
if b.MatchToleration(&c) {
```

`MatchToleration` compares by `<Key, Effect, Operator, Value>` — the canonical Kubernetes definition of toleration equivalence. This also untangles the dual-use problem noted in my earlier comment: `cmpTolerations` would only remain as the `SortSlices` comparator and could be safely rewritten as a proper less-than function without risking breakage here.

pedjak (changes requested):

PR #1591, opened 2026-09-14
OCPBUGS-53193: Update code references to use the default branch name 'main'

rfredette:


Miciah:
There are a few stale references to "master" in `HACKING.md` and `hack/ossm-overrides.md`, but we can get those in a followup.

PR #1599, opened 2026-09-18
NE-2913: Remove HAProxy 2.8 reference

jcmoraisjr:
Remove HAProxy 2.8 reference from the operator.

HAProxy 2.8 references in unit tests are updated in a way it does not need periodic maintenance, since the code and tests are version agnostic.

The e2e test is updated by skipping the block upgrade branch, since the current OCP version does not have any active version being deprecated.

https://redhat.atlassian.net/browse/NE-2913


gcs278:
Generally the approach looks good - straight forward. I'll defer to @pedjak for the approve/lgtm. However, you might want to let https://github.com/openshift/api/pull/3048 merge then re-vendor here to save yourself another PR.

